PR #78, opened 2022-02-10
Fixed typo in Container.html

Nini0la:
Misspelled 'Container'.


djl11:
Thanks for pointing out this typo! Actually the gh-pages branch is automatically pushed to as a github action during doc-building. But this has flagged a typo in the source code which I will fix now. Thanks @Nini0la!

PR #105, opened 2022-02-13
prod conformed for only numpy. 53 of 66 tests are working.

Volthai7us:


Volthai7us:
I only changed files for Numpy at start. But I am not sure about writing test and the files I did change. Could you check and give feedback @djl11

PR #158, opened 2022-02-14
test_cos_arrary_api

Ishticode:
Modified pytest bcz the already written one didn't handle rank one tensors well. Only torch backened required a mini tweak to deal with cos for float16.  
All tests passing for tensorflow, torch, jax, numpy.


Ishticode:
Check the following for updated version
- #259

PR #194, opened 2022-02-16
add unit test for ivy.ones

alierenak:
cc: @djl11
 I added corresponding unit test for ivy.ones. ivy generic function and backends functions are working without any change. Could you please take a look?


alierenak:
#282 after array-api-test module added, all codes are formatted

PR #260, opened 2022-02-21
ivy.asarray

MrTooOldDriver:
#255 adding asarray function in Array API
Current test passing status:

| framework | test_asarray_arrays | test_asarray_scalars |
| --- | --- | --- |
| jax | failed:x: [1] | pass:heavy_check_mark: |
| mxnet | failed:x: [2] | failed:x: [2] |
| numpy | pass:heavy_check_mark: | pass:heavy_check_mark: |
| tensorflow | failed:x:[3] | pass:heavy_check_mark: |
| torch | pass:heavy_check_mark: | pass:heavy_check_mark: |

[1] Error message:
`hypothesis.errors.MultipleFailures: Hypothesis found 3 distinct failures.`
Somehow hypothesis doesn't show me all error messages. Even I already tick do not add "--no-header --no-summary-q" option in pycharm advances settings.

[2] Since we don't care about mxnet. So not huge issues here. Error messages are related to data type conversion from ivy to mxnet.

[3] Error message:
`hypothesis.errors.InvalidArgument: Expected tuple but got shape=TensorShape([]) (type=TensorShape)`
Looks like hypothesis can't extract shape info from TensorShape type?

### DIFF
--- a/ivy/functional/backends/jax/core/general.py
+++ b/ivy/functional/backends/jax/core/general.py
@@ -3,6 +3,8 @@ Collection of Jax general functions, wrapped to fit Ivy syntax and signature.
 """
 
 # global
+from typing import Optional
+
 import jax as _jax
 import math as _math
 import numpy as _onp
@@ -86,7 +88,25 @@ def array(object_in, dtype=None, dev=None):
     return to_dev(_jnp.array(object_in, dtype=dtype_from_str(default_dtype(dtype, object_in))), default_device(dev))
 
 
-asarray = array
+def asarray(object_in, dtype: Optional[str] = None, dev: Optional[str] = None, copy: Optional[bool] = None):
+    if copy is None:
+        copy = False
+    if copy:
+        if dtype is None and isinstance(object_in, _jnp.ndarray):
+            return to_dev(object_in.copy(), dev)
+        if dtype is None and not isinstance(object_in, _jnp.ndarray):
+            return to_dev(_jnp.asarray(object_in).copy(), dev)
+        else:
+            dtype = dtype_to_str(default_dtype(dtype, object_in))
+            return to_dev(_jnp.asarray(object_in, dtype).copy(), dev)
+    else:
+        if dtype is None and isinstance(object_in, _jnp.ndarray):
+            return to_dev(object_in, dev)
+        if dtype is None and not isinstance(object_in, _jnp.ndarray):
+            return to_dev(_jnp.asarray(object_in), dev)
+        else:
+            dtype = dtype_to_str(default_dtype(dtype, object_in))
+            return to_dev(_jnp.asarray(object_in, dtype), dev)
 
 
 # noinspection PyUnresolvedReferences,PyProtectedMember
--- a/ivy/functional/backends/mxnet/core/general.py
+++ b/ivy/functional/backends/mxnet/core/general.py
@@ -3,6 +3,8 @@ Collection of MXNet general functions, wrapped to fit Ivy syntax and signature.
 """
 
 # global
+from typing import Optional
+
 import ivy
 _round = round
 import logging
@@ -124,7 +126,27 @@ def array(object_in, dtype=None, dev=None):
     return _mx.nd.array(object_in, cont, dtype=default_dtype(dtype, object_in))
 
 
-asarray = array
+def asarray(object_in, dtype: Optional[str] = None, dev: Optional[str] = None, copy: Optional[bool] = None):
+    #mxnet dont have asarray implementation
+    cont = _mxnet_init_context(default_device(dev))
+    if copy is None:
+        copy = False
+    if copy:
+        if dtype is None and isinstance(object_in, _mx.nd.NDArray):
+            return _mx.nd.array(object_in, cont).as_in_context(cont)
+        if dtype is None and not isinstance(object_in, _mx.nd.NDArray):
+            return _mx.nd.array(object_in, cont, dtype=default_dtype(dtype, object_in))
+        else:
+            dtype = dtype_to_str(default_dtype(dtype, object_in))
+            return _mx.nd.array(object_in, cont, dtype=default_dtype(dtype, object_in))
+    else:
+        if dtype is None and isinstance(object_in, _mx.nd.NDArray):
+            return object_in.as_in_context(cont)
+        if dtype is None and not isinstance(object_in, _mx.nd.NDArray):
+            return  _mx.nd.array(object_in, cont, dtype=default_dtype(dtype, object_in))
+        else:
+            dtype = dtype_to_str(default_dtype(dtype, object_in))
+            return _mx.nd.array(object_in, cont, dtype=default_dtype(dtype, object_in))
 
 
 def is_array(x, exclusive=False):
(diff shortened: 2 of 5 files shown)


djl11:
Thanks a lot for the PR!

I'll need to rebase this to the latest master branch before continuing. Could you please give me [write access](https://stackoverflow.com/questions/35278957/can-i-modify-someone-elses-pull-request-and-push-it-back-in-its-branch/49854013#:~:text=When%20creating%20a%20pull%20request,commit%20changes%20to%20your%20PR.&text=So%20if%20you%20want%20to,to%20give%20you%20this%20permission.) to the PR? 🙂

djl11:
Thanks a lot for your contribution. Could you update the methods to follow the formatting explained in the [contributor guide](https://lets-unify.ai/ivy/contributing.html) in the [docs](https://lets-unify.ai/ivy/)? I also go through some examples in the Array API [tutorial series](https://www.youtube.com/watch?v=HNV4eI0zp4M&list=PLwNuX3xB_tv96FaEUmTirlQLIsnUgvtTl) on YouTube.

Essentially:

- All methods should be defined with `def func_name` explicitly, no lambdas or direct binding.
- All Ivy methods should have type-hints and doc strings
- All backend methods should have type-hints, but no doc strings.
- All Array API methods should be moved to the correct sub-modules and files.

Please check the contributor guide in the docs for more details. Thanks!

Additionally, it may be necessary to resolve new conflicts with master since the PR was made. As a guide for resolving conflicts with upstream, I have created this short [YouTube series](https://www.youtube.com/watch?v=KCMceLBBlgs&list=PLwNuX3xB_tv8Gf2sELkXSqCCWcUq-AbrG).
Once there are no conflicts with upstream and the formatting is all correct, I'll then take a look at the tests and see if it's ready to be merged :)

djl11 (changes requested):
Could you please try running the tests from the command line rather than inside PyCharm. I expect this will enable the Hypothesis errors to print correctly. Try with and without the use of these flags.

MrTooOldDriver:
Hi @djl11 I been keep running into these strange issues that in numpy framework, default_dtype function doesn't return the correct datatype. It somehow thinks the type of bool array ([False]) is int and return the int default type int32. Is this normal or did I misunderstand the typing system in ivy? 
Screenshot from 2022-02-28 23-04-51
Screenshot from 2022-02-28 22-58-21

MrTooOldDriver:
Also the reason why numpy-array-api-tests failed is also because the default type isn't correct. The default now is int32 and float32 but test_ones wants float64 and test_full want int64
Screenshot from 2022-03-01 00-04-47

MrTooOldDriver:
> Could you please try running the tests from the command line rather than inside PyCharm. I expect this will enable the Hypothesis errors to print correctly. Try with and without the use of these flags.

Thank you for your advice! That solves the issue. Looks like all the error I got in jax is coming from __setitem__ function and type issues. JAX arrays is immutable and needs to use x = x.at[idx].set(y) for setitem. And the type ellipsis is not supported well.

>     def _unimplemented_setitem(self, i, x):
>       msg = ("'{}' object does not support item assignment. JAX arrays are "
>              "immutable. Instead of ``x[idx] = y``, use ``x = x.at[idx].set(y)`` "
>              "or another .at[] method: "
>              "https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html")
>      raise TypeError(msg.format(type(self)))
>  TypeError: '<class 'jaxlib.xla_extension.DeviceArray'>' object does not support item assignment. JAX arrays are immutable. Instead of ``x[idx] = y``, use ``x = x.at[idx].set(y)`` or another .at[] method: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html


>     def dtype_to_str(dtype_in):
>         if isinstance(dtype_in, str):
>             return dtype_in
>       return DTYPE_TO_STR[dtype_in]
>       KeyError: Ellipsis


>       if not _any(isinstance(leaf, ndarray_types) for leaf in leaves):
>         # TODO(jakevdp): falling back to numpy here fails to overflow for lists containing
>         # large integers; see discussion in https://github.com/google/jax/pull/6047.
>         # More correct would be to call coerce_to_array on each leaf, but this may have
>         # performance implications.
>        out = np.array(object, dtype=dtype, ndmin=ndmin, copy=False)
>        TypeError: int() argument must be a string, a bytes-like object or a number, not 'ellipsis'

MrTooOldDriver:
> Hi @djl11 I been keep running into these strange issues that in numpy framework, default_dtype function doesn't return the correct datatype. It somehow thinks the type of bool array ([False]) is int and return the int default type int32. Is this normal or did I misunderstand the typing system in ivy? Screenshot from 2022-02-28 23-04-51 Screenshot from 2022-02-28 22-58-21

**I have added a temporary fix to bypass the issue about default_type() didn't return the correct type for [False] array. The error in array-api test is now all fixed.**



> > Could you please try running the tests from the command line rather than inside PyCharm. I expect this will enable the Hypothesis errors to print correctly. Try with and without the use of these flags.
> 
> Thank you for your advice! That solves the issue. Looks like all the error I got in jax is coming from **setitem** function and type issues. JAX arrays is immutable and needs to use x = x.at[idx].set(y) for setitem. And the type ellipsis is not supported well.
> 
> > ```
> > def _unimplemented_setitem(self, i, x):
> >   msg = ("'{}' object does not support item assignment. JAX arrays are "
> >          "immutable. Instead of ``x[idx] = y``, use ``x = x.at[idx].set(y)`` "
> >          "or another .at[] method: "
> >          "https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html")
> >  raise TypeError(msg.format(type(self)))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > TypeError: '<class 'jaxlib.xla_extension.DeviceArray'>' object does not support item assignment. JAX arrays are immutable. Instead of `x[idx] = y`, use `x = x.at[idx].set(y)` or another .at[] method: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html
> 
> > ```
> > def dtype_to_str(dtype_in):
> >     if isinstance(dtype_in, str):
> >         return dtype_in
> >   return DTYPE_TO_STR[dtype_in]
> >   KeyError: Ellipsis
> > ```
> 
> > ```
> >   if not _any(isinstance(leaf, ndarray_types) for leaf in leaves):
> >     # TODO(jakevdp): falling back to numpy here fails to overflow for lists containing
> >     # large integers; see discussion in https://github.com/google/jax/pull/6047.
> >     # More correct would be to call coerce_to_array on each leaf, but this may have
> >     # performance implications.
> >    out = np.array(object, dtype=dtype, ndmin=ndmin, copy=False)
> >    TypeError: int() argument must be a string, a bytes-like object or a number, not 'ellipsis'
> > ```

**This issue in JAX still remain.**


>[3] Error message:
>hypothesis.errors.InvalidArgument: Expected tuple but got shape=TensorShape([]) (type=TensorShape)
>Looks like hypothesis can't extract shape info from TensorShape type?

**The issue in Tensorflow still remain.**

mattbarrett98 (changes requested):
Please could you remove the functions from their current files and add them to the correct files i.e move the `asarray` function from ivy/functional/backends/jax/old/general.py to ivy/functional/backends/torch/creation.py, and do the same for all backends 🙂

MrTooOldDriver:
Done. The issue with jax test and tensorflow test still exits. Torch test also reporting 
`TypeError: no implementation found for 'torch.Tensor.__setitem__'`
and
`E       KeyError: Ellipsis`

djl11:
Thanks!

PR #318, opened 2022-02-25
Added roll function to all the backends as well as Ivy

arxxv:
But all the tests fail. I get an AssertionError saying that "roll is not defined in ivy.functional.backends.framework"
for eg: "roll is not defined in ivy.functional.backends.numpy"
I have checked the Youtube tutorial, where you had the same issue, but that didn't work.  
What should I do?


djl11 (changes requested):
Please remove all modified files in .idea.

Also, I was able to run the tests locally without issue, please check your PyCharm configurations, otherwise please try and run via the command line. The tests are passing for NumPy, JAX and Torch but not TensorFlow.

arxxv:
Hey Dan, 
I have removed the changes in .idea directory. I managed to run the tests locally and as you said all of them pass except the Tensorflow one. It fails because the input to roll function has to be at-least 1D and cannot contain None inside it (throws cannot convert NoneType to Tensor). 
Is checking if the shape of the input is at least 1 and then checking if the input contains None in it, the best way to handle this?

arxxv:
Hey 1Doomdie1, thanks for the reply.
I would do that the next time onwards. 
It would be great, if you could answer those queries of mine. In case TF roll function doesn't support some input type, how should I handle it? What errors should I throw?

djl11 (changes requested):
Please remove all changes in the .idea folder :)

djl11:
> Hey Dan, I have removed the changes in .idea directory. I managed to run the tests locally and as you said all of them pass except the Tensorflow one. It fails because the input to roll function has to be at-least 1D and cannot contain None inside it (throws cannot convert NoneType to Tensor). Is checking if the shape of the input is at least 1 and then checking if the input contains None in it, the best way to handle this? Also, how should I handle the errors? What should I return if the input is a scalar or has NoneType?

You will need to handle this corner case explicitly, and ensure that the output is the same as for all other frameworks. Brute-force solutions to this are fine for now. We can always make it more efficient later.

1Doomdie1:
This PR is being closed because there was no activity on this PR in the last 7 days, and you have not requested a code review from us. Please check our policy on the closing of PRs which are linked to ToDo list issues [here](https://lets-unify.ai/ivy/contributing.html#todo-list-issues), please reach out on discord if you believe this issue has been closed mistakenly, and we’d be very happy to help 🙂

PR #323, opened 2022-02-25
eye

HotzingTone:


djl11 (changes requested):
The tests for TensorFlow and PyTorch are failing. Once these are passing I'll then merge :)

HotzingTone:
Thanks Dan!
Tests on my computer never work probably, but I'll try.

djl11 (changes requested):
Why has matrix power been added as well? Also could you please uncomment  the eye method from the .txt files so that it's included in the tests? :)

HotzingTone:
Sorry! matrix_power was accidentally added, now removed.
Also eye is included in the tests now.

djl11 (changes requested):
It seems as though the NumPy tests are the only ones passing. Could you make any updates so the other framework tests also pass?

djl11 (changes requested):
JAX test seems to be failing, could you try and get this passing? :)

HotzingTone:
> 

It's weird that JAX used to pass when I request the review, but now it doesn't.

HotzingTone:
@djl11 So now all tests passed. But what I did was just rerun the test, no code changed :)

PR #332, opened 2022-02-26
Implementing log2 to ivy and all backends

1Doomdie1:


1Doomdie1:
Regarding 523e18a773ca067a9916d8be7c20bacded803fdb, that was a mistake by me. I forgot to add the files first :smile: 
All the test are passing, special cases as well.

djl11 (changes requested):
Please remove all contents from the commit which is not related to log2 (for example everything in .idea)

djl11 (commented):
Thanks!

PR #348, opened 2022-02-26
log1p implementation

1Doomdie1:


djl11 (changes requested):
Could you uncommet log1p in this file please? 

https://github.com/1Doomdie1/ivy/blob/master/ivy_tests/array_api_methods_to_test/elementwise_functions.txt

This will add it to the tests :)

PR #352, opened 2022-02-27
some corrections and edits

Ishticode:
3 edits to some of the recent merges
- Remove type from docstring of 3 functions
- Remove `f=f` argument in `less` function
- Change calling `any` in `flip` function


djl11:
Thanks!

PR #353, opened 2022-02-27
floor_divide_incomplete

kevinsu2024:
implemented numpy & torch floor_div to adjust to dtype for 1 out of 5 test cases for each. I don't quite understand the error that comes from other test cases, for some with the function name __floordiv__, I am not sure why the types don't match, and for other function names e.g. __ifloordiv__, it gives TypeError: array() got an unexpected keyword argument 'copy'. May someone help me take a look? Thanks

### DIFF
--- a/ivy/functional/backends/jax/array_api/elementwise_functions.py
+++ b/ivy/functional/backends/jax/array_api/elementwise_functions.py
@@ -44,6 +44,9 @@ def less(x1: JaxArray,x2:JaxArray)\
         -> JaxArray:
     return jnp.less(x1,x2)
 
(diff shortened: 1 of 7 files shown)


kevinsu2024:
I think a few of these errors came from type casting, but I'm not exactly sure how to fix it. Can you give me some suggestions? Thank you for your time!

mattbarrett98 (changes requested):
Maybe you can look at other successful PRs for ideas. I'm afraid I'm not sure what's causing the errors and I can't do much more without doing the task for you 🙂

PR #357, opened 2022-02-27
linspace

pantelic-dusan:
ISSUE1: in numpy test there is error when start or stop argument of linspace are -0.0. linspace sometimes return -0 sometimes +0 but must returns -0.
ISSUE2: in jax,tensorflow,torch maximum floating point precision is 64(unlike numpy) so there is no way of passing tests when start or stop arguments are large numbers like 9007199254740993 because they are internally converted to floating point numbers.
ISSUE3: tests are not consistent, sometimes all test passes sometimes there is error even if there is no change in functions code. For example i only encountered ISSUE1 few times and in most cases -0 isn't covered by tests.
ISSUE4: sometimes in torch i get ivy.array(23.999996, dtype=float32), but should be 24.0 because start and stop should be exactly the same as first and last element of returned array.
ISSUE5: logspace function is dependent on nonstandard linspace implementation in torch and tensorflow, but logspace itself don't exists in standard api so I can't rewrite it to match standards and use standard linspace implementation if necessary.

For ISSUE1 and ISSUE4 maybe i can manually set first and last element of array to be same as start and stop, but this solution is not so elegant

For ISSUE4 maybe data standard api can add special cases for linspace for edge float numbers and +0.0, -0.0

Also for some reason torch don't have problem with large start and stop parameters even it supports maximum of 64bit floating point precision.

### DIFF
--- a/ivy/functional/backends/jax/array_api/creation_functions.py
+++ b/ivy/functional/backends/jax/array_api/creation_functions.py
@@ -22,3 +22,16 @@ def zeros(shape: Union[int, Tuple[int], List[int]],
           device: Optional[jaxlib.xla_extension.Device] = None) \
         -> JaxArray:
     return to_dev(jnp.zeros(shape, dtype_from_str(default_dtype(dtype))), default_device(device))
+
+def linspace(start: Union[int, float],
+             stop: Union[int, float],
+             num: int,
+             dtype: Optional[jnp.dtype] = None,
+             device: Optional[jaxlib.xla_extension.Device] = None,
+             endpoint: bool = True) \
+             -> JaxArray:
+    if dtype is None:
+        dtype = jnp.float32
+    return to_dev(jnp.linspace(jnp.double(start), jnp.double(stop), num, dtype=dtype, endpoint=endpoint),
+                  default_device(device))
+
--- a/ivy/functional/backends/jax/core/general.py
+++ b/ivy/functional/backends/jax/core/general.py
@@ -161,12 +161,6 @@ def arange(stop, start=0, step=1, dtype=None, dev=None):
     return to_dev(_jnp.arange(start, stop, step=step, dtype=dtype), default_device(dev))
 
 
-def linspace(start, stop, num, axis=None, dev=None):
-    if axis is None:
-        axis = -1
-    return to_dev(_jnp.linspace(start, stop, num, axis=axis), default_device(dev))
-
-
 def logspace(start, stop, num, base=10., axis=None, dev=None):
     if axis is None:
         axis = -1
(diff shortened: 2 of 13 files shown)


pantelic-dusan:
@1Doomdie1 As you can see I have similar problem here as with trip function

1Doomdie1:
@pantelic-dusan I will make a clone after your fork and see what's the problem!
Thanks for flagging this.

pantelic-dusan:
@1Doomdie1 For reproducing different test cases you can just add few spaces or enters to current  test case file and save. As I noticed when I make some unsignificant change in test function I get whole different set of test input and sometimes there are edge cases included sometimes there aren't

djl11:
Hi @pantelic-dusan, you are running into some very challenging issues here, and you are right, these are recurring problems. I'm digging into this now. Thanks for pointing these out so clearly in your comment.

djl11 (changes requested):
In order to get the backend method using a particular data type, it is often necessary to perform the conversion **before** calling the method. For example, consider the current tensorflow code:

`tf.cast(tf.linspace(start, stop, num+1), dtype)`

This will pass in the large integers or floats into linspace, which will then convert them to the default data type, which may have a lower precision and cause rounding errors. In contrast, the following solution works correctly:

`tf.linspace(tf.constant(start, dtype), tf.constant(stop, dtype), num+1)`

I suspect similar logic will apply to other backend methods also.

Please make these updates first in order to solve issue 2, and then we can address the +0, -0 issues afterwards. I'm currently trying to get to the bottom of this myself.

djl11:
Regarding your point on Issues 1 and 4, your suggestion is not elegant, but it's better than a failing test. We can always revisit and optimize the implementations once we have unified behavior. So by all means feel free to go ahead with this inelegant hack to get the tests passing for the time being.

pantelic-dusan:
> In order to get the backend method using a particular data type, it is often necessary to perform the conversion **before** calling the method. For example, consider the current tensorflow code:
> 
> `tf.cast(tf.linspace(start, stop, num+1), dtype)`
> 
> This will pass in the large integers or floats into linspace, which will then convert them to the default data type, which may have a lower precision and cause rounding errors. In contrast, the following solution works correctly:
> 
> `tf.linspace(tf.constant(start, dtype), tf.constant(stop, dtype), num+1)`
> 
> I suspect similar logic will apply to other backend methods also.
> 
> Please make these updates first in order to solve issue 2, and then we can address the +0, -0 issues afterwards. I'm currently trying to get to the bottom of this myself.

When I try your solution there is no problem with large numbers but new error occurs because I need return array of same type as dtype argument:
E       AssertionError: out.dtype=float32, but should be uint8 [linspace(dtype=uint8)]

I tried to wrap your solution in cast to dtype but then old error with big numbers occurs again.

PR #358, opened 2022-02-27
Device

hubmiles:
As there was no test_device function on the code you told me on Discord to create the PR anyways.

### DIFF
--- a/README.rst.save
+++ b/README.rst.save
@@ -1,0 +1,323 @@
+.. image:: https://github.com/unifyai/ivy/blob/master/docs/partial_source/logos/logo.png?raw=true
+   :width: 100%
+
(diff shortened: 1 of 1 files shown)


djl11:
My apologies, I presumed there was still something that needed implementing that we could then check through manually without a test. However, I've just noticed that the device attribute is already created for the ivy.Array class, as can be seen [here](https://github.com/unifyai/ivy/blob/13b3be98a82cd839a95641abef8274f3918f7228/ivy/array/__init__.py#L74). Therefore, I will close this issue and mark this attribute as completed. We will free up many more methods soon in the ToDo lists. Please choose another one to work on when we do this. Thanks!

PR #366, opened 2022-02-27
tril implementation

pantelic-dusan:
ISSUE: Fails when array have -0.0 in elements

### DIFF
--- a/ivy/functional/backends/jax/array_api/creation_functions.py
+++ b/ivy/functional/backends/jax/array_api/creation_functions.py
@@ -22,3 +22,10 @@ def zeros(shape: Union[int, Tuple[int], List[int]],
           device: Optional[jaxlib.xla_extension.Device] = None) \
         -> JaxArray:
     return to_dev(jnp.zeros(shape, dtype_from_str(default_dtype(dtype))), default_device(device))
+
+
+def tril(x: JaxArray,
+         k: int = 0) \
+         -> JaxArray:
+    return jnp.tril(x, k)
+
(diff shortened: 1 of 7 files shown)


1Doomdie1 (commented):
Hey thank you for contributing! Can you let me know which backend has this problem?

pantelic-dusan:
@1Doomdie1 I think in all, because all test passed on local(there was no - 0.0 case generated when I tested on my pc) ,  so I can't reproduce errors. But I had same issue on numpy and jax on linspace pull request.

1Doomdie1:
@pantelic-dusan if all the test are passing locally then it's fine. For the time being the test-array-api-jax are failing when you make a PR so don't mind them :D

pantelic-dusan:
> @pantelic-dusan if all the test are passing locally then it's fine. For the time being the test-array-api-jax are failing when you make a PR so don't mind them :D 

Yeah but as you can se on my linspace pull request there is case where my tests are failing locally because of - 0.0 are randomly generated in tests but on github check there are no errors because - 0.0 is included. I will tag you in my linspace pull request so you can see whole image of problem with non consistent and random generated test cases.

pantelic-dusan:
Here is example with comparison problem with - 0.0 and 0.0 in jax

1Doomdie1:
> 
> 
> Here is example with comparison problem with - 0.0 and 0.0 in jax

Yes, as I said the tests for jax for now are failing so don't mind them. If all the tests for tril are passing locally then it's fine. I will take a look at your linspace PR.

pantelic-dusan:
@1Doomdie1 locally all test except jax are failing. On local test fails on jax in 
'Falsifying example: test_diagonal(
    x=ivy.array([[-0., -0., -0.]], dtype=float32), kw={},
)'
So if it's expected behaviour for now I think we don't have any issue

djl11:
> > @pantelic-dusan if all the test are passing locally then it's fine. For the time being the test-array-api-jax are failing when you make a PR so don't mind them :D
> 
> Yeah but as you can se on my linspace pull request there is case where my tests are failing locally because of - 0.0 are randomly generated in tests but on github check there are no errors because - 0.0 is not included. I will tag you in my linspace pull request so you can see whole image of problem with non consistent and random generated test cases.

Yes this is an ongoing problem. I've not managed to find a solution. They fail non-deterministically. For now we can go ahead and merge if these are the only failures, and I'll keep digging to find a fix for this.

djl11:
Thanks!

PR #372, opened 2022-02-28
unique_counts adhering to Array API standard

RKuipers:
Closed previous PR and started new one on new branch due to issues merging upstream. 

All backends excep MXNet and PyTorch pass unit tests. MXNet is for unrelated issues, PyTorch is because of an issue in the unit test.

The related unit test has a line "assume(x.size <= 128)" where x is the input of the unit test, a PyTorch Tensor in this case. Since Tensor.size does not exist (it's Tensor.size() and that still does not return a comparable value but a Tensor.Size object) this fails. Since this fails completely outside the function I'm meant to implement, and we should not change unit tests, I don't know how to resolve this. 

I've attempted to conform to the new contribers guide; if I made errors in this, please point out where so I can update them.


HotzingTone (commented):
Looks good to me for now, waiting for djl11 to review.

RKuipers:
@djl11 I implemented the changes you requested to the best of my ability, but Torch and Jax now fail unit tests in ways that I'm unsure how to solve without adapting the unit tests.

Jax has issues dealing with nan's, but this is a part of Jax that does not comply with Array API standard, as discussed in the respective repos https://github.com/data-apis/array-api/issues/249 and https://github.com/google/jax/issues/9182 . I can add a lot of code on top of the Jax call to solve this but I expect that's not ideal. (This same issue will likely pop up for the other unique functions btw.)

Torch has a TypeError in the unit test, on the input. It calls x.size as an attribute, but with the Torch backend x is a Tensor which has a size method instead. Since we're not supposed to change this unit test there is no way to get this passing.

1Doomdie1:
@RKuipers please request a code review

HotzingTone (changes requested):
JAX & Torch haven't passed. Would you please keep working on it? Thanks!

RKuipers:
@HotzingTone I've explained above why it's impossible to get Jax & Torch passing within the restrictions.

mattbarrett98:
Thanks for contributing!

PR #378, opened 2022-03-01
Creation Function- "ones_like" conforming to array api standard

jkeane508:
Git Issue #270


mattbarrett98 (changes requested):
Hi, just a few changes required before we can merge. Could you remove any changes made to files unrelated to the "ones_like" function, in particular the .idea files.

Also the ones_like function should be moved to the ivy/array_api/creation_functions.py file and typing information can be removed from the docstring. 

Please could you go to the ivy_tests/array_api_methods_to_test/creation_functions.txt file and uncomment your function ones_like, so that the tests can run.

PR #379, opened 2022-03-02
Conforming var functions with the Array API

oliverxue20:
Conforming var functions with the Array API


djl11 (changes requested):
Please remove all .idea changes from the commit

PR #387, opened 2022-03-03
Brought ceil into array API standard

mattbarrett98:
Ceil functions have been moved to the correct places and all tests are now passing.


djl11:
Thanks!

PR #399, opened 2022-03-04
updated broadcast_to function to conform to array_api standard to solve #390

uchennayah:
Made all the changes required to conform to array API standard, however, Tensorflow didn't pass due to an issue with the tensor data type, it required a tuple data type but got Tensorshape.


mattbarrett98 (changes requested):
Hi, changes look good, could you just go to the ivy_tests/array_api_methods_to_test/data_type_functions.txt and uncomment your function broadcast_to, so that the tests will run. Thanks :slightly_smiling_face:

mattbarrett98:
Also all tests will need to be passing before we can merge your changes.

mattbarrett98:
The TensorFlow tests still aren't passing this will need to be fixed. Don't worry about the failures from Jax this is being caused by something else 🙂

mattbarrett98 (changes requested):
The TensorFlow tests still aren't passing this will need to be fixed. Don't worry about the failures from Jax this is being caused by something else 🙂

PR #416, opened 2022-03-07
Implementation of subtract conforming to Array API Standard #217 facing some errors

rupsgrover:
- Updated Elementwise functions in all backends.
- Jax, Tensorflow, and Torch are passing 3/5 tests. The test cases failing are for __isub__(x1,x2) and __isub__(x,s).
- Numpy is passing 2/5 tests. The tests failing are __sub__(x,s),  __isub__(x1,x2) and __isub__(x,s).

Please help me with these issues, how can I make these test cases pass?

### DIFF
--- a/ivy/functional/backends/jax/array_api/elementwise_functions.py
+++ b/ivy/functional/backends/jax/array_api/elementwise_functions.py
@@ -60,3 +60,13 @@ def cos(x: JaxArray)\
 def logical_not(x: JaxArray)\
         -> JaxArray:
     return jnp.logical_not(x)
+
+
+def subtract(x1: JaxArray,
+                x2: JaxArray)\
+        -> JaxArray:
+    if hasattr(x1, 'dtype') and hasattr(x2, 'dtype'):
+        promoted_type = jnp.promote_types(x1.dtype, x2.dtype)
+        x1 = x1.astype(promoted_type)
+        x2 = x2.astype(promoted_type)
+    return jnp.subtract(x1, x2)
--- a/ivy/functional/backends/mxnet/array_api/elementwise_functions.py
+++ b/ivy/functional/backends/mxnet/array_api/elementwise_functions.py
@@ -45,3 +45,9 @@ def less(x1: mx.ndarray.ndarray.NDArray,x2: mx.ndarray.ndarray.NDArray)\
 def logical_not(x: mx.ndarray.ndarray.NDArray)\
         -> mx.ndarray.ndarray.NDArray:
     return mx.nd.logical_not(x)
+
+
+@_handle_flat_arrays_in_out
+def subtract(x1: mx.ndarray.ndarray.NDArray, x2: mx.ndarray.ndarray.NDArray)\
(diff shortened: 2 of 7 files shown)


mattbarrett98 (changes requested):
Could you fix the conflicts please. Also can you uncomment your functions from the relevant .txt files in ivy_tests/array_api_methods_to_test, so that the tests run. As it is I can't see what's failing 🙂

mattbarrett98 (changes requested):
It looks like the __sub__ function in NumPy is failing because the dtype of the output is incorrect. You need to amend the function so that the output is of the correct dtype as specified in https://data-apis.org/array-api/latest/API_specification/generated/signatures.elementwise_functions.subtract.html#signatures.elementwise_functions.subtract 🙂

mattbarrett98 (changes requested):
Don't worry about `__isub__` for now, I think the failures here are related to another function that is yet to be implemented. 

Regarding the other NumPy failure, the datatype of the output isn't as expected. The output should be uint8 but it is int64.

rupsgrover:
I have figured out the errors I was facing, the only thing left now is the implementation of the ```__isub__``` function.

mattbarrett98 (changes requested):
Looks good, could you just remove the .idea files from your commit these shouldn't be here 🙂 As for the remaining failures, I think they are occurring because of another method `asarray` which is yet to be implemented, so don't worry about these.

mattbarrett98 (changes requested):
There is still another .idea file in the commit, please could you remove it 🙂

rupsgrover:
I deleted the vcs.xml file in .idea but it was still showing in the files changed section of PR. So, I added that file again from the original ivy repo. Is it ok or should I delete it?

mattbarrett98 (changes requested):
You need to ensure that in your commit, there are no changes to any files in the .idea folder, thanks 🙂

mattbarrett98 (changes requested):
Thanks 🙂 `asarray` has now been implemented so most of your tests are now passing. `__isub__` is still failing for Jax however if you could try to fix this.

rupsgrover:
Hi @mattbarrett98, I'm unable to understand the error `__isub__` is facing for Jax. I think it might be a Jax specific error and not just limited to subtract function.

mattbarrett98 (changes requested):
Hi, I think the jax errors are occurring because of another problem with `asarray` so we should be okay to merge your changes. Before we do that however could you please re-fork the repo, re-add your changes and make another PR (and request my review). 

This is because it looks like things haven't gone smoothly when resolving conflicts. Under 'files changed' it's showing a large number of deletions and insertions which makes it difficult for me to check you haven't edited anything you shouldn't have (I know you probably haven't but I just need to make sure). 

Thanks 🙂

rupsgrover:
Hi, I have created a new PR #658 but I don't have access to request your review. Can you please check?
Thanks 🙂

PR #449, opened 2022-03-09
implementing outer

Husain-Ayoob:


mattbarrett98 (changes requested):
Your calls to outer were slightly incorrect for some of the backends, which I fixed. Now all the tests are passing except for Tensorflow, so you will need to amend the function in the tensorflow backend to make this pass. 

Also, please could you add a docstring to the outer function in the ivy/functional/ivy/array_api/linear_algebra.py file 🙂

Husain-Ayoob:
There are two different chunks of code on my TensorFlow backend as I am not sure which would work best. 
Docstring added in  ivy/functional/ivy/array_api/linear_algebra.py

mattbarrett98 (changes requested):
You have duplicated a lot of the functions, can you ensure the function in each backend is only defined once. 

You have duplicated the same code for the TensorFlow backend which won't stop the tests failing. To get the tests to pass you will need to change the logic of the function, in order to fix whatever errors you are receiving. Thanks 🙂

mattbarrett98 (changes requested):
Thanks! Could you resolve the conflicts with your fork so that the tests can run. Also, can you try to fix whatever error you are getting from Tensorflow 🙂

Husain-Ayoob:
Fixed all the conflicts, i do believe it is because the function matamul had x1, x2 aswell as my code so when I removed that and switched it with a1 and a2 all of my conflicts were resolved so can you please check this issue.

mattbarrett98 (changes requested):
Please could you make sure the TensorFlow implementation is in the right file (ivy/functional/backends/tensorflow/linear_algebra.py) 🙂

Husain-Ayoob:
It is in the right file

mattbarrett98 (changes requested):
No it isn't. Yours is in the ivy/functional/backends/tensorflow/array_api/linear_algebra.py, so you need to move it to the correct file and then delete the array_api folder since this was removed in a recent refactoring and should no longer exist

Husain-Ayoob:
@mattbarrett98 please review my code

Husain-Ayoob:
Just fixed conflicting conflicts by updating fork.

mattbarrett98 (commented):
The tests are no longer running I think something has gone wrong when merging upstream changes. I think the easiest thing to do now is to re-fork ivy's repo, add the outer functions as you have done here and make another pull request. 

There are just too many changes on this PR now so it's best to start over. Thanks 🙂

Issue #460, opened 2022-03-10
max


ar0usel:

ar0usel:
I did local tests (`ivy_tests/test_array_api/array_api_tests/test_statistical_functions.py::test_max`) and it passed successfully for all 4 backends. In my work, I based on discussion in this PR (std #98).

Issue #98, opened 2022-02-13
std


DhruvSondhi:

DhruvSondhi:
Hello, currently I am not able to find a suitable file to add the statistical functionality to the module. Can you please clarify the file where this could be done or should I try to create a new file for this?

DhruvSondhi:
Another question, I saw some tests for the function, that have been commented out in the related test file. I don't know if the tests are working or not. I wish to know that for this particular function do want me to write some different tests for the function or should my implementation pass the existing tests?

djl11:
You should un-comment these tests, and then make any necessary changes to get them passing. As explained in this [series](https://www.youtube.com/watch?v=b_hcj0xAOjY&list=PLwNuX3xB_tv-VNWD8e4Hqt2gb8BIcZ7Ny), currently the helper methods in the Array API test suite don't integrate well with Ivy's tests, so these tests need to be modified before they will run.

DhruvSondhi:
Check comment on PR & please reopen this as I wish to continue working with this 🤔 @1Doomdie1

PR #469, opened 2022-03-10
max 

ar0usel:
updated func 'max' in backend(jax, numpy, tensorflow, torch) and in ivy core for Array API.
Ref. max #460


ar0usel:
As I see, after merging last changes, it passes all array API tests. However, when I renamed reduced_max to max (as stated here https://github.com/unifyai/ivy/issues/98#issuecomment-1037946288), it caused problems with `test-ivy/run-nightly-tests`. What should I do with this?

HotzingTone:
> 

Thanks for flagging this, we apologies for the confusion

> Hi, @HotzingTone. I noticed, that one more contributor took my function (#543, #542). He took it after me, but his pr was already merged. What should I do with this?:)



> 

Hi, thanks for flagging this. We've sent you an email explaining the solution.

PR #517, opened 2022-03-12
Array api compatible floor function

nebilarega:


Ishticode (changes requested):
Hi @nebilarega It seems you have implemented the function perfectly in ivy core which is great.
Although, you would also need to implement this for the backends, and then uncomment your function from `ivy_tests/array_api_methods_to_test`.
Any problems please let us know : )

PR #532, opened 2022-03-13
Prod

Eng-Abdelrahman-M:
Added prod function 

Have 3 errors in PyTorch 
2 because of dtype PyTorch does not support uint16 , uint32 and uint64
-  input  dtype = <undefined stub for 'uint16'> **riase** an invalid argument
- input dtype = torch.uint8 **requies** expected = <undefined stub for 'uint32'>

1 error because of not applicable test case 
- input x = tensor([0], dtype=torch.uint8), axis = 1  **raise**  Dimension out of range
this error is right because of the tesor is 1 dim (0) there is no axis 1

Should I handle these errors or the problem in test cases?


Eng-Abdelrahman-M:
Thank you, Dan. I hope they fix it soon.

Eng-Abdelrahman-M:
@djl11  Sorry Dan, The error you mentioned there can be fixed from our side be if the statement. I already made if statement [here](https://github.com/Eng-Abdelrahman-M/ivy/blob/1de41f12141a766d8a1a15256bd36deff19f137c/ivy/functional/backends/torch/statistical.py#L28).
The problem is if the function got an unsupported dtype. I think you fixed this one.  
The two remaining problems are the axis problem, and how I can return uint32 by fixing the condition I mentioned [here](https://github.com/Eng-Abdelrahman-M/ivy/blob/1de41f12141a766d8a1a15256bd36deff19f137c/ivy/functional/backends/torch/statistical.py#L28).

djl11:
I have suppressed the unit test for now, and added [a comment](https://github.com/unifyai/ivy/blob/f4b78d354c7eef469600ad3f9bb69e4400a86aa6/ivy_tests/array_api_methods_to_test/statistical_functions.txt#L4) that this test is failing due to a bug with the Array API test suite. Thanks for the contribution!

Eng-Abdelrahman-M:
Thank you Dan!

PR #540, opened 2022-03-14
trunc adhering to Array API standard

Oluwadurotimi10:
A pull request for trunc element wise method for issue #371


HotzingTone (changes requested):
Hi, thanks for your work!
In the TF backend, could you try not using numpy.trunc? It's supposed to be implemented entirely using TF functions.

PR #543, opened 2022-03-14
[array api] add statistical max

daamoolaa:
background: https://github.com/unifyai/ivy/issues/542

change: implement the statistical max function, and enabled tests


Ishticode (approved):
@daamoolaa Excellent work : )
I will pass it on to Daniel.
@djl11 function in core is named `maximum` so won't need to deal with that.

djl11:
Thanks!

PR #544, opened 2022-03-14
Implemented Array api compatible floor function in the backends and in ivy core

nebilarega:
Also uncommented the floor function from ivy_tests/array_api_methods_to_test

### DIFF
--- a/ivy/functional/backends/jax/array_api/elementwise_functions.py
+++ b/ivy/functional/backends/jax/array_api/elementwise_functions.py
@@ -18,6 +18,13 @@ def ceil(x: JaxArray)\
     return jnp.ceil(x)
 
 
+def floor(x: JaxArray)\
+        -> JaxArray:
+    if 'int' in str(x.dtype):
+        return x
+    return jnp.floor(x)
+
+
 def isfinite(x: JaxArray)\
         -> JaxArray:
     return jnp.isfinite(x)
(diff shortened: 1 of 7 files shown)


nebilarega:
In the test_floor function, while testing, the test seems to fail with assert dtype error. But when I explicitly test the function from the backends they seem to work. 

Detail:
          - Error: Assert same dtype error:
                   The returned dtype and x.dtype are different.
          - Tried: Explicitly import elementwise_function floor:
                   In this case returned dtype and x.dtype are the same.

Numpy example: x.dtype = int[8]
Screen Shot 2022-03-15 at 12 04 21 PM

mattbarrett98 (changes requested):
You can ignore the failure of the special case test, this is due to another function not yet being implemented. As for the dtype errors, it looks like the function isn't handling the case of integer input arrays properly 🙂

nebilarega:
Hey @mattbarrett98, the actual backend functions pass the case when I explicitly import them while debugging the test case. As you see in the above picture when I call "ivy.functional.backends.numpy.array_api.elementwise_function" the case passes( For integer input dtype it returns integer output dtype). But when I just run the test_floor it fails. Whatever code I change in the backends it is the same dtype error. I don't know if I am missing sth. As the function is pretty straightforward and exactly the same as the ceil function.

mattbarrett98:
Strange. I wonder if you need to delete the definitions of `floor` in the old files i.e. is `floor` still defined in ivy/functional/backends/jax/old/general.py etc? Maybe this is causing the problem.

PR #551, opened 2022-03-15
unique_inverse adhering to Αrray API standard

maraPislar:
Frameworks:
- jax
- mxnet
- numpy
- tensorflow
- torch


mattbarrett98 (changes requested):
Please could you resolve the conflicts. There has been some refactoring so for example the torch implementation will need to be in the ivy/functional/backends/torch/set.py file, rather than ivy/functional/backends/torch/array_api/set_functions.py. Thanks 🙂

mattbarrett98 (changes requested):
Tests are currently failing for all backends. Please could you update the functions such that all the tests pass 🙂

mattbarrett98:
The reason for the remaining failures is that the test depends on other methods that haven't been implemented yet, so we're good to merge your changes! Thanks for contributing 🙂

PR #555, opened 2022-03-15
Matrix Inverse #433

SayanJerr:


SayanJerr:
Did all I was asked in https://github.com/unifyai/ivy/pull/516#pullrequestreview-909987767

Ishticode (changes requested):
Hi SayanJerr. Thanks for this PR.
Could you kindly remove the .idea files and also only uncomment your own function?
Did you get your tests passing locally?
You're welcome to ask any questions : )

SayanJerr (commented):
I have deleted the .idea files and the numpy and torch are passing.
the tensor low is raising a shape error(from the revertible matrixes) and the jax is having return a 2X2 array with inf and -inf

Ishticode:
> @Ishticode I have deleted the .idea files and the numpy and torch are passing. the tensor low is raising a shape error(from the revertible matrixes) and the jax is having return a 2X2 array with inf and -inf

Hi @SayanJerr. Just to be clear, when you change you PR, you should ask for a review so me or anyone requested gets notified. I am sorry for the delay but I never got notified. 

Since your last commit changed 92 files. we have two solutions, either you revert ur commit and we try to work with the previous version or you could close this PR add you function and create PR again. You need to exclude .idea files from your push. Deleting them on PR will try to delete them on repo. Sorry If caused any confusion. Furthermore, in the actual function files, could you only modify your function. Perhaps thats what you already did but git detects more changes. So whether you open a PR or make changes and ask for a review, I will respond quickly. Is that ok?

PR #556, opened 2022-03-16
not_equal adhering to Array API Standard

Christoforos00:
issue #538


Christoforos00:
Hi, is there something extra I should do?

mattbarrett98:
Thanks for contributing!

PR #557, opened 2022-03-16
bitwise_invert adhering to Array API Standard

ebyau:
All tests passing locally except for Tensorflow one test fails

### DIFF
--- a/ivy/functional/backends/jax/elementwise.py
+++ b/ivy/functional/backends/jax/elementwise.py
@@ -5,6 +5,11 @@ import jax.numpy as jnp
 from ivy.functional.backends.jax import JaxArray
 
 
+def bitwise_invert(x: JaxArray)\
+        -> JaxArray:
+    return jnp.bitwise_not(x)
+
+
 def bitwise_and(x1: JaxArray,
                 x2: JaxArray)\
         -> JaxArray:
--- a/ivy/functional/backends/numpy/elementwise.py
+++ b/ivy/functional/backends/numpy/elementwise.py
@@ -3,6 +3,11 @@ import numpy as np
 import numpy.array_api as npa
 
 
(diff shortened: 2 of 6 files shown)


ebyau:
@mattbarrett98 this is new PR for  bitwise_invert

mattbarrett98 (changes requested):
Thanks! Tensorflow is failing because `tf.bitwise.invert` doesn't take allow bool input types. I think then if you add some logic to the tf function so that when the datatype of `x` is bool, then you first cast `x` to an integer type (using `tf.cast`) and then call `tf.bitwise.invert`. I think this will fix the failure 🙂

ebyau:
I have done the  casting but the tensorflow tests still fail; but i tested the logic with tensorflow code by creating boolean tensor trying `tf.bitwise.invert()` it fails then when i cast the tensor to an i`nt32 dtype` it works well

ebyau:
Why do think the Tensor flow tests could be still failing?

mattbarrett98:
> Why do you think the Tensor flow tests could be still failing?

Looking into it I've noticed that `bitwise.invert` is a different function in Tensorflow, and doesn't do what we want here. I've tried something using `logical_not` instead, I'll see if the tests pass.

mattbarrett98:
Got there in the end 🙂 The problem was `tf.bitwise.invert` doesn't allow bool inputs, so in this case we just use `tf.logical_not`. Thanks for contributing!

ebyau:
You welcome
Will the `tf.logical_not` be able to handle float tensors?

mattbarrett98:
bitwise_invert only needs to handle int and bool dtype https://data-apis.org/array-api/latest/API_specification/generated/signatures.elementwise_functions.bitwise_invert.html 🙂

PR #558, opened 2022-03-16
floor adhering to Array API Standard

nebilarega:


nebilarega:
Hey @mattbarrett98 you were right. The problems were caused by `ivy/functional/backend/numpy/old/general.py` implementations of `floor` function. After commenting them out, the tests run successfully, other than the special cases.

mattbarrett98:
Thanks for contributing!

PR #563, opened 2022-03-16
var adhering to Array API standard

kevinxue1126:
ivy/ivy_tests/test_array_api/array_api_tests/test_statistical_functions.py

when you test, you need to delete .filter(lambda x: x.size >=2) on line 270 of test_statistical_functions.py
Because this line cannot be verified in jax, tensorflow, torch temporarily

Depends on the size method of ivy.array


mattbarrett98 (changes requested):
Please could you resolve the conflicts, so the tests run 🙂

mattbarrett98:
Thanks for contributing!

PR #592, opened 2022-03-19
Size attribute adhering to Array API Standard

UsamaHasan:


saeedashrraf (changes requested):
Hello @UsamaHasan, thank you for your efforts. It seems like there are changes being made to the function __ge__ withing your commit which is not related to the implementation of the size attribute. Can you undo these changes?

Additionally, to implement the size attribute, instead of summing up the values inside the shape attribute, we need to multiple them. To be able to do that, we should use the method reduce. You can take a look at how to use the method to multiply a sequence of elements [here](https://stackoverflow.com/questions/2104782/returning-the-product-of-a-list). Feel free to reach out if you have any questions :)

UsamaHasan:
Sure, already done. There are tests that are failing in the master branch. So Can't help that. Rest you can review.

saeedashrraf:
This looks great. Thank you for your efforts :)

PR #605, opened 2022-03-21
Adding issue comment workflow

1Doomdie1:
This workflow will keep track of the comments made on the issues labeled "ToDo" and "Array API" and assign users to whatever function they chose to work on.
- [ ] If the comment doesn't have the format shown in your videos it will be ignored.
- [ ] If the comment contains a function that has been already allocated then the issue in question will be closed.
- [ ] If the comment contains a function that is not allocated then the issue will be updated with the correct labels, the author will be assigned to it, and the functions list in the main issue will be updated.
- [ ] ATM GitHub CLI does not support comment deletion on issues or PRs, this means that we will need to go through the comments manually and delete them, but we will know that they have been handled correctly.

### DIFF
--- a/.github/workflows/issue_comment.yml
+++ b/.github/workflows/issue_comment.yml
@@ -1,0 +1,25 @@
+on:
+  issue_comment:
+    types:
+      - created
+jobs:
+  issue_commented:
+    name: Issue comment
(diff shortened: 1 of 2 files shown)


djl11 (approved):
lgtm!

PR #639, opened 2022-03-22
logical_xor adhearing to Array API Standard

sreegandh804:


aayushgoyal443:
I made a pull request for the same function before you. Also the issue for same function was assigned to me before you.

djl11:
@aayushgoyal443, presumably the confusion with the other [issue](https://github.com/unifyai/ivy/issues/636) and this PR was caused because you closed the issue before the task was complete. This then freed up the method so that @sreegandh804 could work on it, which they did. When we get round to reviewing, one of the PRs will be closed and the other will be merged.

PR #653, opened 2022-03-23
remainder

mattbarrett98:


djl11:
Hi @saeedashrraf, could you take a look at these failing tests? @mattbarrett98 has an m1 Mac and so cannot run the JAX tests locally. Thanks!

PR #662, opened 2022-03-23
Eig linalg extention

arv0804:
#615


saeedashrraf:
@arv0804  Great thanks!
Can you also make sure that there is only one active pull request per issue?  There is a duplicate pull request for the eigh issue. Can you close one of them ?

PR #679, opened 2022-03-25
Update Matrix-Rank Extension to Conform to Array API Standard

bethelhall:


saeedashrraf (changes requested):
Hello @bethelhall, There seems to be an error arising from the parameter types you declare in the function definition. Can you fix that? Make sure to only apply changes to the relevant parts of the function and to not include any other files or changes in the commits :)

bethelhall:
@saeedashrraf Thanks for the review. I am on it.

bethelhall:
Hey @saeedashrraf, the local tests pass on my machine. Can you please check? I have corrected the type of the argument, 'rtol. I think that was the problem.

saeedashrraf (changes requested):
There seems to be two implementation of the function matrix_rank for each backend framework. Can you make sure the function is added only once. Also, can you not add any changes to the .idea folder. There are changes to the .idea/ivy.iml file right now. Lastly, can you double check on on the arguments of the function and make sure they are the same as the arguments of the function in the Array API Standard as the jax implementation seems to have only one parameter instead of two.

bethelhall:
@saeedashrraf, there is no "matrix_rank" function implementation on the master branch except on my branch; I double-checked. And you can also check on the Main Issue that the method, "matrix_rank" has not been assigned to anyone yet. I do not why the .idea/ivy.iml file gets changed every now and then. 
I will check the Jax implementation again. Thank you.

bethelhall:
@saeedashrraf I see the method is duplicated in each backend implementation. Weird. I am not sure if I caused that but I checked again if the implementation is also on the master branch, but it wasn't.

saeedashrraf:
On your branch, there are two copies of matrix_rank for every backend framework for example. For instance, for numpy [first copy](https://github.com/unifyai/ivy/pull/679/files#diff-149f09470841122d5ecc546338cb9327b37781ea74a4265256ab89919a75777fR130-R134) and [second copy](https://github.com/unifyai/ivy/pull/679/files#diff-149f09470841122d5ecc546338cb9327b37781ea74a4265256ab89919a75777fR157-R159). Make sure to remove one of the two copies.

bethelhall:
@saeedashrraf Got it.

bethelhall:
@saeedashrraf, How about now? I have made the changes and updated the argument type according to the Array API Standard. Jax will not compile on my machine as I am using Mac Air M1. So, I can't run the test on my local machine.

saeedashrraf:
Hello @bethelhall , the tests still fail for tensorflow and numpy . Can you try to resolve the errors causing that ?

bethelhall:
@saeedashrraf Thanks for checking. I am working on it.

bethelhall:
@saeedashrraf The tests pass locally, but they seem to fail the remote test. Can you please check the new commit?

bethelhall:
Hello @saeedashrraf, I think the tests are failing becuase of other methods. Can you please check this commit out?

bethelhall:
@saeedashrraf, for example, this one. The failure seems to be linked to the linalg.slodget function.

Screen Shot 2022-03-31 at 9 30 32 AM

saeedashrraf (approved):

saeedashrraf:
Hello @bethelhall , the issues were caused by the way jax and tensorflow handled the the shapes of the vector array. This is all fixed now. Thank you for your effort!

bethelhall:
@saeedashrraf, Thanks, buddy! Happy to contribute more!

PR #693, opened 2022-03-26
Add atan2

tensorush:
This one's pretty straightforward. I've only added the doc string as well as two-line functions to numpy, tensorflow and jax backends (torch had already been put in place).

### DIFF
--- a/ivy/functional/backends/jax/elementwise.py
+++ b/ivy/functional/backends/jax/elementwise.py
@@ -230,7 +230,10 @@ def atan(x: JaxArray)\
     return jnp.arctan(x)
 
 
-atan2 = jnp.arctan2
+def atan2(x1: JaxArray, x2: JaxArray) -> JaxArray:
+    return jnp.arctan2(x1, x2)
+
+
 cosh = jnp.cosh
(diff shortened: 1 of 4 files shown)


tensorush:
[] #674

tensorush:
It seems that jax's test_equal fails due to an integer overflow error. I'm having some troubles installing jaxlib on Windows to enable local testing. I'll try fixing element-wise equal, pushing and running the tests here.

tensorush:
Strange, now that I've reverted the changes to look exactly as at the time of my first commit, there is no integer overflow error anymore in the test_equal for the jax backend. I'll now uncomment atan2 inside the `ivy_tests/array_api_methods_to_test`.

tensorush:
The integer overflow for jax test_equal has returned. There is a hardcoded solution for this in the element-wise bitwise_or function, though. I'm not sure if that's the way I should be solving it, please let me know. In the meantime, I'll be fixing errors with other backends.

tensorush:
Looks like everything's passing now.

saeedashrraf:
@tensorush , this looks good. Thank you for your effort!

PR #694, opened 2022-03-26
squeeze adhering to Array API Standard.

HadIhab:


HadIhab:
The test for the PyTorch backend passed successfully on my local machine, but it fails here in Github actions is it a problem with the env ?

Ishticode (approved):
Good to go. Thnx for the contribution : )

Ishticode:
> The test for the PyTorch backend passed successfully on my local machine, but it fails here in Github actions is it a problem with the env ?

That was not a problem with your code.

HadIhab:
> Good to go. Thnx for the contribution : )

All right, you're welcome I'm very happy to contribute in such great projects.

PR #714, opened 2022-03-28
Matrix Inverse #443

SayanJerr:
Tensorflow And Jax have issues from @given(x=invertible_matrices())


SayanJerr:
Adding @settings(max_examples=20) makes the tensorflow run

Aarsh2001:
Hi, I'm getting the same error in TensorFlow for my method, let me know if you find a solution

PR #729, opened 2022-03-30
solve to conform to array API standard

Aarsh2001:


Aarsh2001:
although a partial solution, I had some doubts with the tests -:

Torch error -: RuntimeError: Boolean value of Tensor with no values is ambiguous https://github.com/unifyai/ivy/issues/628
TensorFlow error - : AttributeError: 'tensorflow.python.framework.ops.EagerTensor' object has no attribute 'setitem'
jax error : FailedHealthCheck

Requesting Review @ivy-seed @djl11 @saeedashrraf @mattbarrett98

saeedashrraf (changes requested):
Hello @Aarsh2001 , thank you for your effort! You should not be raising exceptions in the function. Instead, you should handle the input properly to avoid those exceptions getting triggered. Can you please remove those?

Aarsh2001:
> Hello @Aarsh2001 , thank you for your effort! You should not be raising exceptions in the function. Instead, you should handle the input properly to avoid those exceptions getting triggered. Can you please remove those?

Alright, will do that

Aarsh2001:
will make new pull request

PR #733, opened 2022-03-31
meshgrid adhering to Array API Standard

abdrahmandiab:
There is a slight problem with testing:
the test_meshgrid was passing for most backends without me even changing anything. Furthermore when I searched for all implementations of meshgrid and commented them, all tests were still passing. So I am unsure whether or not the functions I created were the ones being testing or if the test is defaulting to something else.


abdrahmandiab:
Update: the PR checks all fail at the "checkout 🛎Ivy" phase, with the error message "Error: The process '/usr/bin/git' failed with exit code 128". Unsure why this happens. 
Please help :)

mattbarrett98 (changes requested):
Looks good! Could you just remove the device argument from the NumPy function. Also, are the `to_dev` calls necessary in the NumPy and jax functions? Thanks 🙂

PR #736, opened 2022-03-31
roll #688

nkennebre:


saeedashrraf (changes requested):
Hello @nkennebre , thank you for your effort! Can you add  the implementation of roll to the rest of the backend modules? Can you also make sure to not make any changes to any files other than the ones relevant to your function? Lastly, can you uncomment the roll test from "ivy_tests/array_api_methods_to_test/manipulation_functions.txt" ?

nkennebre:
> Hello @nkennebre , thank you for your effort! Can you add the implementation of roll to the rest of the backend modules? Can you also make sure to not make any changes to any files other than the ones relevant to your function? Lastly, can you uncomment the roll test from "ivy_tests/array_api_methods_to_test/manipulation_functions.txt" ?

Do I make a new pull request when these things are done?

nkennebre:
I couldn't figure out the tensorflow backend. I think that it has something to do with the axis parameters, but I am not sure how to fix this.

saeedashrraf (changes requested):
@nkennebre The issue is that the tf.roll function does not have a behavior when the axis parameter is None. If you look at the function described in the [Array API Standard](https://data-apis.org/array-api/latest/API_specification/generated/signatures.manipulation_functions.roll.html). They mention that if axis is None, the array must be flattened, shifted, and then restored to its original shape. So, you need to do that manually for tensorflow if the parameter axis is None.

Feel free to ask any questions if you need to. Good luck!

nkennebre:
roll #688 I am not sure how to do this. This is what I came up with...

    if axis is None:
        new_axis = list(range(num_dims))
        roll(new_axis, shift = 1, axis = 0)
        new_axis.shape = x.shape

I think the shift is where I am lost. I am not sure the value of the shift in this function. Please advise.

saeedashrraf:
As mentioned, the array must be flattened , then shifted, and then the resulting array should be restored to the original shape. This can be done as follows:
   
    if axis == None:
            originalShape = x.shape
            axis = 0
            x = tf.reshape(x,[-1])
            roll = tf.roll(x,shift,axis)
            return tf.reshape(roll,originalShape)

saeedashrraf (approved):

saeedashrraf:
Thank you for your contribution @nkennebre

Issue #809, opened 2022-04-04
tanh


Mia-dev200:
- [ ] #754

Mia-dev200:
#754

Mia-dev200:
[ ] #754

Mia-dev200:
- [ ] #754

Mia-dev200:
#754

1Doomdie1:
Please leave a comment with your issue id here: #754 
This way the function will be allocated to you automatically!
Thank you!

Mia-dev200:
> Please leave a comment with your issue id here: #754 This way the function will be allocated to you automatically! Thank you!

Heyy , how can i ask for review of my PR ... I'm not understanding where I'm going wrong

PR #822, opened 2022-04-04
Reformat activation#754 Tanh #809  

Mia-dev200:
I have done the reformatting of tanh activation  #809 ... if someone can review it and let me know if there are any changes needed. Thank you.


saeedashrraf:
@Mia-dev200 Thank you for your effort!

Mia-dev200:
@saeedashrraf Thanks .. however the issue isn't closed yet so is that something that i need to do or is it from your side ?

PR #824, opened 2022-04-05
elementwise ceil (#765)

Sohob:
Added the docstring example for ceil along with its test in the test file. Local tests for the function were green, so there was no need to edit the implementation.


saeedashrraf:
@Sohob Thank you for your effort!

PR #846, opened 2022-04-08
Reformatted det

bethelhall:


mattbarrett98 (changes requested):
You need to update the test suite submodule to the latest commit, your fork has an older version currently. Also can you check the formatting for all backends, the type hints for Jax are incorrect, thanks

bethelhall:
@mattbarrett98 Okay, thanks. Will do.

PR #848, opened 2022-04-08
num_gpu

MihaiAnca13:
Reformatted num_gpu #847 

When I was trying to run the test the following line failed inside the `test_memory_on_dev` function:
```
ret = ivy.total_mem_on_dev(dev_to_check)
```

I think this has to do with how I set up docker. I tried the dockergpu, but got the same results. However, I've isolated the `assert_docstring_examples_run` function call and it works.


MihaiAnca13:
The logs of the failed check are only saying the problem is in "test_memory_on_dev". Any idea why `ret = ivy.total_mem_on_dev(dev_to_check)` is failing for me? Could this be the same issue?

ivy/functional/backends/torch/device.py
```diff
@@ -54,7 +54,10 @@ def clear_mem_on_dev(dev):
 
 _callable_dev = dev
 gpu_is_available = torch.cuda.is_available
-num_gpus = torch.cuda.device_count
+
+
+def num_gpus() -> int:
+    return torch.cuda.device_count
```
bicycleman15, line 60: I think this should be `torch.cuda.device_count()`. You missed the parenthesis it seems :)

bicycleman15:
LGTM !
Can you give a final look @mattbarrett98 ?

mattbarrett98 (commented):
I think the example should show some output and also the doc string test has been added to the test of a different function which doesn’t seem right. Although Dan said he wanted to give the final review on these PRs to start with so I think check with him 🙂

MihaiAnca13:
Thanks for having a look @mattbarrett98. I received a message on upwork saying Dan is quite busy at the moment. I'm happy to add the changes you've proposed Matt, but I am not sure I understand what to change for the example. `num_gpus` is quite simple and don't know what else to add to the doc string. In terms of having a different test function for it, I can give that a go.

saeedashrraf:
Hello @MihaiAnca13 , I apologize for the delay. You should add an example output of the method to the docstring. This essentially is just the returned integer you get after running ivy.num_gpus(). Similar to how the result of print(y) is added [here](https://github.com/unifyai/ivy/commit/57d97e94026dc81c66ba580112f2d72a9d405d59#diff-aa9310d758ed4fd979817c39c9df57480f958f3fc333d31ce99cdf750673a719R36). 

Feel free to reach out again if you have anymore questions :)

MihaiAnca13:
@saeedashrraf Sorry for my late response, I was away. I added the string as suggested, but it looks like there are new conflicts now. Should I try to solve these or wait for you to check them to make sure there are no issues?

saeedashrraf:
@MihaiAnca13 Thank you for your effort.

PR #854, opened 2022-04-09
add out argument to `to_dev`

bicycleman15:


sherry30 (commented):
I see that you've also reformatted the doc string, can you add a print statement to the example and also its output like its done on an interactive session.
example of to_numpy:
"""
Examples:
    >>> x = ivy.array([-1, 0, 1])
    >>> y = ivy.to_numpy(x)
    >>> print(y)
    array([-1, 0, 1])
"""
The rest looks fine to me :)

PR #878, opened 2022-04-12
reformatted random_uniform

Vishwaak:


mattbarrett98 (changes requested):
Can you double check your type hints, not sure how the shape argument could be a dtype. Also you need to add a docstring test as dan showed in the youtube tutorial🙂

PR #884, opened 2022-04-12
reformatted to_numpy

sreegandh804:


sreegandh804:
> NumPy arrays print to the terminal like `array([-1, 0, 1])` not `[-1, 0, 1]`.
> 
> Also, perhaps add the line `print(type(y))` and the output, to make it very clear that it is a NumPy array that is returned.
> 
> Thanks :)

I made the necessary changes Dan . :)

djl11 (changes requested):
last thing, the terminal outputs should be directly underneath each individual print statement ideally. As though the commands were being run interactively, one line at a time. Otherwise it's all good!

sreegandh804:
Apologies for the wrong format, I just edited that. Now that I got a hang of reformatting, I shouldn't have these issues with future PRs

PR #899, opened 2022-04-13
reformatted to_numpy, to_scalar & to_list

sreegandh804:


sherry30 (changes requested):
your changes to the to_numpy function are also being tracked in this PR since that one has not been accepted yet, i suggest to uncheck changes for to_numpy function while making a  commit for this PR

sherry30 (approved):

PR #905, opened 2022-04-14
Reformatting bitwise_invert

ebyau:


HotzingTone (changes requested):
Hi, there're a few things to change:
1, The existing 'out' is actually a newly added argument, just keep it as it is, don't put it under 'Returns'.
2, Under 'Returns', create another 'out', but this one actually means output. (sorry I know it's easy to get confused with the argument 'out')
3, The example is wrong, you could have a look at the [numpy document](https://numpy.org/doc/stable/reference/generated/numpy.invert.html) to get a sense what it would look like.
4, The example should also include the correct output.

PR #915, opened 2022-04-14
Reformat Conv1d

YsBsr:
Conv1d operation reformatted for
jax, mxnet, numpy, tensorflow, torch
You can check the notebook file below
to see every implementation fully detailed.
--> https://github.com/YsBsr/CONV1D_NUMPY_JAX_TENSORFLOW_TORCH/blob/main/CONV1D_NUMPY_JAX_TENSORFLOW_TORCH.ipynb


djl11 (changes requested):
Looks good. Could you please add type hints for every parameter in the arguments, and also remove the commented out code in the test entirely? Thanks!

YsBsr:
All changes have been made. Please let me know if there is anything else to do. Thanks!

YsBsr:
Hello @djl11. I am having some issues with Jax tests. 

Below is the error message of some of the Jax test results. I think this error message is related to the test code. Because if I change these lines below from the test code failing ones become successful and succesful ones become failing.

```diff
@pytest.mark.parametrize(
-  "dtype", ['float32'])
+  "dtype", ['float64'])
```

`
name = 'conv_general_dilated', ignore_fp_precision = False
ttypes = (dtype('float32'), dtype('float64'))
types = [dtype('float32'), dtype('float64')]
msg = 'lax.{} requires arguments to have the same dtypes, got {}.'

    def _check_same_dtypes(name, ignore_fp_precision, *ttypes):
      """Check that dtypes agree, possibly ignoring float precision."""
      # the `ignore_fp_precision` flag exists because the XLA shape inference logic
      # allows mixed floating point precision, but the HLO verifier often rejects it
      types = list(map(np.dtype, ttypes))  # canonicalize
      if ignore_fp_precision:
        types = [
            np.floating if dtypes.issubdtype(dtype, np.floating)
            else np.complexfloating if dtypes.issubdtype(dtype, np.complexfloating)
            else dtype for dtype in types]
      if len({dtypes.canonicalize_dtype(t) for t in types}) != 1:
        if ignore_fp_precision:
          msg = ("lax.{} requires arguments to have same dtypes up to floating point "
                 "precision, got {}.")
        else:
          msg = "lax.{} requires arguments to have the same dtypes, got {}."
        if name in _JNP_FUNCTION_EQUIVALENTS:
          equiv = _JNP_FUNCTION_EQUIVALENTS[name]
          msg += f" (Tip: jnp.{equiv} is a similar function that does automatic type promotion on inputs)."
>       raise TypeError(msg.format(name, ", ".join(map(str, types))))
E       TypeError: lax.conv_general_dilated requires arguments to have the same dtypes, got float32, float64.

/home/y/.local/lib/python3.9/site-packages/jax/_src/lax/lax.py:4533: TypeError

`

saeedashrraf:
Hello @YsBsr , it seems that all the tests are currently passing for the function. The only traces of Jax failures in your commits was related to the data generation being too slow for the function reshape which you should not worry about as it is not within the scope of your function. Are you still having any other issues ?

YsBsr:
@saeedashrraf, Thanks for explaining to me. I am not having any other issues.

saeedashrraf (approved):

PR #924, opened 2022-04-16
reformat device

LINBOTANG123:
Reformat device gpu_is_available


ivy/functional/backends/numpy/device.py
```diff
@@ -13,18 +13,20 @@
...
 dev_to_str = lambda dev: 'cpu'
 dev_from_str = lambda dev: 'cpu'
 clear_mem_on_dev = lambda dev: None
-gpu_is_available = lambda: False
 num_gpus = lambda: 0
 tpu_is_available = lambda: False
 
 
+
+def gpu_is_available() -> int:
```
bicycleman15, line 27: return type should be `bool`.

bicycleman15:
Rest LGTM ! I will merge once you make the above change 🙂

ivy/functional/ivy/device.py
```diff
@@ -264,7 +264,13 @@ def gpu_is_available() -> bool:
...
     Determine whether a GPU is available to use, with the backend framework.
 
-    :return: Boolean, as to whether a gpu is available.
+    Returns
+    -------
+    out:
+        Boolean, as to whether a gpu is available.
+
+    Examples:
+    >>> print(ivy.gpu_is_available())
```
bicycleman15, line 273: also add an example output as done [here](https://github.com/unifyai/ivy/commit/57d97e94026dc81c66ba580112f2d72a9d405d59#diff-aa9310d758ed4fd979817c39c9df57480f958f3fc333d31ce99cdf750673a719R36) :)
A simple `True` would work

LINBOTANG123 (commented):
Change return type to bool and add an example

LINBOTANG123:
The followings are two error messages but I don't think that's caused by my PR.
Could you please review my PR again? Thanks for your time :)
ivy/ivy_tests/test_array_api/array_api_tests/test_statistical_functions.py:168: FailedHealthCheck
ivy/ivy_tests/test_ivy/test_functional/test_core/test_gradients.py::test_gradient_descent_update[cpu-ivy.functional.backends.torch-True-False-torch_call-torch-asarray-float32-ws_n_grads_n_lr_n_wsnew0]

ivy/functional/ivy/device.py
bicycleman15: Just one small change. I think `Examples` should be written like below following Numpy Doc String style:
```
Examples
--------
```
Rest LGTM 🙂

bicycleman15:
Yes, I think these are not related. Make the above small change, I will merge the PR 👍

ivy/functional/ivy/device.py
LINBOTANG123: Change made. Thanks for your help :)

PR #938, opened 2022-04-18
Adding std example to functional.py methos

MALLI7622:


SayanJerr:
In the docstring example,
No correct answer was given for the test function to test against

MALLI7622:
Oops sorry forget it will add and create new request closing this now

PR #939, opened 2022-04-18
Reformatted cross_entropy

kolejnyy:
Pull request for issue #937


mattbarrett98 (changes requested):
Looks good! Just a few small changes required:
1. Could you ensure that there are no changes to any .idea files in your commit, these shouldn't be there.
2. Take a look at this docstring https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/elementwise.py#L40 and make sure yours if formatted similarly. For example one difference is you have a `:` after the parameter names.
3. Please can you remove the docstring test you have added, it is no longer necessary to add these manually (but it will still be tested). 

Thanks 🙂

PR #955, opened 2022-04-20
reformatted files in ivy &array folders to numpy style

sreegandh804:


mattbarrett98 (commented):
In statistical we were changing the name of the return from ‘out’ to ‘return’ so that it wouldn’t clash with the parameter ‘out’. Not sure if you changed this in other files as well but maybe check with Dan whether this clashing is okay

sreegandh804:
Yeah sure! I'll make that change.

PR #982, opened 2022-04-22
reformatted get_min_denominator

Mamdouh-irl:


Mamdouh-irl:
Does anyone have a clue how to solve the fail of the last check? It has something to do with the docstring tests.

PR #1012, opened 2022-04-29
Cholesky

Aarsh2001:
#1007 
The method had been implemented twice in the docstring file. So I deleted the one which had no additional arguments. Hope that is not a problem

### DIFF
--- a/ivy/functional/ivy/linear_algebra.py
+++ b/ivy/functional/ivy/linear_algebra.py
@@ -230,22 +230,6 @@ def diagonal(x: ivy.Array,
     return _cur_framework(x).diagonal(x, offset, axis1=axis1, axis2=axis2)
 
 
-def cholesky(x):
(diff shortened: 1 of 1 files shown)


djl11 (changes requested):
different examples with different print statements should be added sequentially. like so:

```
some text

>>> some code

some more text

>>> some more code

```

djl11 (commented):
Thanks for adding more changes, I've just requested a review from @mattbarrett98 :)

mattbarrett98 (changes requested):
Hi, the submodule test_array_api is out of date on your fork (you can see it appearing in files changed). You will need to update the submodule so that it is up to date with the one on the master branch.

mattbarrett98 (changes requested):
If you look under files changed you can see test_array_api still has changes, hence yours still isn’t up to date with the one in master

PR #1027, opened 2022-05-04
Reformat random normal function

modanesh:
#1026


modanesh:
@sherry30 Could you please take a look at it? I'm a bit puzzled on what's causing the test to fail. I know it's generated in the `docstring_examples_run()` because of `random_normal()`'s docstrings but can't figure out what's the exact reason.

sherry30:
It seems like test_random_normal under ivy_tests is failing and not the docstring test. I suspect removing ivy.seed(0) in the test function to be the issue, try adding it back and see if the tests are passing.

modanesh:
> It seems like test_random_normal under ivy_tests is failing and not the docstring test. I suspect removing ivy.seed(0) in the test function to be the issue, try adding it back and see if the tests are passing.

I put back the `ivy.seed(0)` in the `test_random_normal` but tests are failing agian.

PR #1054, opened 2022-05-07
reformat tensordoti

Aarsh2001:
#1030


VedPatwardhan (changes requested):
Hey @Aarsh2001, please pull the updated master code into your PR, so that we can observe if the cross-entropy test failure persists. Thank you!

PR #1088, opened 2022-05-12
add hypothesis test for `expand_dims`

bicycleman15:


mattbarrett98 (commented):
Hi mate, not sure the composite dtypes function you have added is necessary, can't you just sample a dtype with `dtype=st.sampled_from(ivy_np.valid_dtype_strs)`? Also, the map_sequences argument is missing from the container method and I believe num_positional_args should be able to take the value 0 🙂

mattbarrett98 (commented):
Just removed `ivy.NativeArray` from the `out` type hints as they shouldn't be there, but it looks good to me 🙂 Only thing i can see is that your changes have some formatting issues appearing in the lint CI check which you should fix.

PR #1093, opened 2022-05-13
Adding func_wrapper comments

saeedashrraf:


djl11 (changes requested):
lgtm! But could you also add comments for the ivy.Container wrapping?

ivy/container/wrapping.py
```diff
@@ -37,13 +37,16 @@ def new_fn(
...
+        #retrieve all the containers in args and kwargs
         arg_conts = ivy.multi_index_nest(args, arg_cont_idxs)
         num_arg_conts = len(arg_conts)
         kwarg_conts = ivy.multi_index_nest(kwargs, kwarg_cont_idxs)
+        # Combine the retrieved containers from args and kwargs into a single list
         conts = arg_conts + kwarg_conts
         if not conts:
             raise Exception("no containers found in arguments")
         cont0 = conts[0]
+        # Get the container version of the function with the name fn_name
```
djl11, line 49: this is not actually the container version of the function, it's still the ivy version of the function, but we enable container-specific ivy modules. This enables containers to specify their own backend irrespective of the global ivy.

ivy/container/wrapping.py
```diff
@@ -54,7 +57,8 @@ def map_fn(vals, _):
             kw = ivy.copy_nest(kwargs, to_mutable=True)
             ivy.set_nest_at_indices(kw, kwarg_cont_idxs, kwarg_vals)
             return fn(*a, **kw)
-
+        # Call the container version of the function on every container in args and 
+        # kwargs and store the result in ret
```
djl11, line 61: Not quite, we call the normal ivy version of the function `fn` on each array at the leaves of the containers. `map_fn` replaces each container arg and kwarg with the array at this specific leaf, before passing to the standard ivy method `fn`

djl11 (changes requested):
a few changes requested as inline comments :)

PR #1096, opened 2022-05-13
add hypothesis integration for`test_flip`, strategy for valid axis

bicycleman15:


bicycleman15:
this test is failing for `numpy` since `numpy` does not support inplace operations when the array has gone through a flip operation, we need to use `np.ascontigousarray()` to rectify this in `update_inplace` method.

bicycleman15:
Even if I try to make it contiguous, turns out the `inplace_update` (apparently, the line `x_native.data = val_native` is not working as it should) still does not work. PFA.
image

Issue #1102, opened 2022-05-14
sparse_cross_entropy


anirudhakulkarni:
Reformatting the 'sparse_cross_entropy' function.

PR #1103, opened 2022-05-14
Reformatted sparse_cross_entropy

anirudhakulkarni:
Pull request for issue #1102


anirudhakulkarni:
@bicycleman15 Can you put some light on the test case thats failing? I could not understand the exact test case from the output. I ran the tests locally and got error in docstrings but exact input or issue is still unknown.

bicycleman15:
It's a good start, but still needs some changes.
I would like you to go through the following contributor guide pages, namely:
https://lets-unify.ai/ivy/contributing/2_type_hints.html, 
https://lets-unify.ai/ivy/contributing/3_docstrings.html, 
https://lets-unify.ai/ivy/contributing/4_docstring_examples.html

Let me know when you are done making changes. Let's not worry about failing tests for now 🙂

bicycleman15:
It's a good start, but still needs some changes.
I would like you to go through the following contributor guide pages, namely:
https://lets-unify.ai/ivy/contributing/2_type_hints.html, 
https://lets-unify.ai/ivy/contributing/3_docstrings.html, 
https://lets-unify.ai/ivy/contributing/4_docstring_examples.html

Let me know when you are done making changes. Let's not worry about failing tests for now 🙂

anirudhakulkarni:
@bicycleman15 Thanks for the resources.
1. I updated the type hints to include ```ivy.Container``` as its taking at least one input and returning at least one output.
2. added examples as suggested in the format: functional and instance method with different examples for "with Container", "with array", "with native_array" and combination of them.

anirudhakulkarni:
@bicycleman15 is there any update on this?

bicycleman15:
It looks good now, just some more changes are required 🙂 
Could you please add examples that cover:
1. functional examples with `ivy.Container` as inputs
2. functional examples with both `ivy.Array` and `ivy.Container` as inputs
3. Instance examples of `ivy.Container`

anirudhakulkarni:
@bicycleman15 done!

bicycleman15:
LGTM!

ivy/functional/ivy/losses.py
```diff
@@ -81,24 +81,100 @@ def binary_cross_entropy(true, pred, epsilon=1e-7):
...
+        axis: Optional[int] = -1,
+        epsilon: Optional[float] = 1e-7,
+) -> ivy.Array:
     """Computes sparse cross entropy between logits and labels.
 
     Parameters
     ----------
     true
-        True labels as logits.
+        input array containing the true labels as logits.
```
bhattbhuwan13, line 95: I think the word **logits** makes it confusing. The description for `true` should instead be `input array containing true labels`

ivy/functional/ivy/losses.py
```diff
@@ -81,24 +81,100 @@ def binary_cross_entropy(true, pred, epsilon=1e-7):
...
     pred
-        predicted labels as logits.
+        input array containing the predicted labels as logits.
```
bhattbhuwan13, line 97: It is stated that the `pred` is the `input array containing the predicted labels as logits.`. However, if you look at the docstring examples, you will find that the sum of the array values(`[0.1, 0.1, 0.7, 0.1]`) is 1. This makes the input to `pred` something like an output from the softmax layer and not the logits.

anirudhakulkarni: The way predicted labels are used implies it was indeed output of softmax layer. Otherwise one needs to take softmax explicitly. Hence the correct labeling should have been ```pred```: ```input array containing predicted probabilities after taking softmax```.

ivy/functional/ivy/losses.py
anirudhakulkarni: Generally ground truth is passed as ith categorical number. There is no sense of logits in ground truth. I agree with the logic, but I didn't want to change the initial wording.

bhattbhuwan13: Not always. If you are using  categorical crossentropy instead of sparse categorical crossentropy, you need to pass labels as probabilities (or one hot encodings).

bhattbhuwan13: You are right. The word `logits` even though it doesn't make sense but creates confusion. I think it should be removed. What do you think?

ivy/functional/ivy/losses.py
bhattbhuwan13: Exactly. You can make the docstring even shorter. Just write `pred`: `array containing softmax probabilities.`

PR #1107, opened 2022-05-14
reformat argsort

Daniel4078:
Currently, all frameworks except mxnet pass the unit test. The mxnet one failed due to a hypothesis.errors.InvalidArgument: Could not create full array of dtype=<class 'bool'> with fill value False. Need guide on what to do about it. Thank!


Daniel4078:
I am curious, what does "LGTM" means? @1Doomdie1

djl11:
> I am curious, what does "LGTM" means? @1Doomdie1

It means "looks good to me" :)

PR #1117, opened 2022-05-16
Added docstrings tests. 

MarShaikh:
- Extended the tests to assert that the print outputs from docstrings match with the values from running the code within it. 
- Skipped certain functions based on a particular condition: For more explanation please find in test_docstrings.py
- docstring test fails for JAX for some runtime error: Needs to be addressed.


VedPatwardhan:
I'll look into the code in some time and get back to you.

MarShaikh:
> I'll look into the code in some time and get back to you.

Just a heads-up, the lint test failing for test_image.py, I did not have to work with. I had the same errors in my local branch but did not correct it in case someone is working on that at the moment.

VedPatwardhan (commented):
Just to be sure that the tests failing are totally related to the changes you've made, could you please revert the changes made to test_helpers.py, test_docstrings.py and test_image.py ?

If the tests still fail then it's not a problem with the code you've added but with something else like submodules not being up to date, etc. 

If the tests pass after reverting the changes for those files, then we can be sure that the new additions are causing the failure and would simplify your work.

MarShaikh:
> Just to be sure that the tests failing are totally related to the changes you've made, could you please revert the changes made to test_helpers.py, test_docstrings.py and test_image.py ?
> 
> If the tests still fail then it's not a problem with the code you've added but with something else like submodules not being up to date, etc.
> 
> If the tests pass after reverting the changes for those files, then we can be sure that the new additions are causing the failure and would simplify your work.

I just resolved some of the merge conflicts that were occurring. I also checked all the error logs, and my PR isn't causing any of them as I specifically only worked on test_docstrings.py and helpers.py. One of the errors that is being caused in test_array_api_numpy is being caused due to the docstrings in some function, which in itself is a good thing as my PR is detecting that. Given the nature of the task I think some of the interns are doing(reformatting docstrings if I'm not wrong) I think it is important to get my PR merged if there are no further issues directly with my code(if some of the tests fail here again, I'll check the logs again but I doubt it will be because of my code). Thanks again, Ved!

VedPatwardhan (approved):
Sure, go ahead @MarShaikh

PR #1126, opened 2022-05-17
Create 5_setting_up.rst

Ishticode:


djl11 (changes requested):
we should actually put this as `0_setting_up.rst` and bump all other numbers up by 1. We should also add this as point (a) in the file `contributing.rst`, and link to the title correctly as done with the other points. Once these changes are done, I'll get it merged, and I'll add the linux instructions myself. Let me know if you have any questions :) Thanks!

PR #1144, opened 2022-05-19
Reformatting dev_to_str

Hoof60:
#1142 

Reformatting the dev_to_str function in device.py .

### DIFF
--- a/ivy/functional/ivy/device.py
+++ b/ivy/functional/ivy/device.py
@@ -148,21 +148,29 @@ def dev(
 # Conversions
 
 # noinspection PyShadowingNames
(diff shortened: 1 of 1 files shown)


mattbarrett98 (changes requested):
Couple of mistakes here:
- you have incorrectly changed the input of this function from device to x, this function does not take arrays as input
- consequently if you try to run the example you have given it will give an error as you are wrongly passing an array to dev_to_str
- the first line of the docstring is incorrect, this function isn't to do with data types
- your example shouldn't be indented

mattbarrett98:
Just noticed someone else is already working on this function #1031 so I have closed yours. Please choose a function which isn't already taken 🙂

PR #1146, opened 2022-05-19
threading test

saeedashrraf:


djl11 (changes requested):
I think it looks good, could you also add a unit test? :)

djl11 (approved):
lgtm!

PR #1149, opened 2022-05-19
first commit, reformatting outer function in linear_algebra from nump…

whatcoloris:
Here's a working reformatting of the outer function in numpy's linear algebra methods. As far as I can tell it passes all the tests and certainly works when I try it in the docker instance. Thank you!


mattbarrett98 (changes requested):
You have misunderstood the reformatting task- all of these changes aren't a part of the task. Maybe have another read of the `reformat linear_algebra` issue- the task is to make sure all backends are formatted correctly and update the docstring

PR #1176, opened 2022-05-23
Added new FAQ and updated docstrings_examples. 

MarShaikh:


djl11 (changes requested):
please remove the Transpiler QandA, and also add an explicit explanation somewhere that the `\` are needed at the end of the wrapped lines, and  then it's ready to merge :)

PR #1201, opened 2022-05-25
det function reformatted 

Hab-eeb:
- [ ] #1198

### DIFF
--- a/ivy/functional/backends/jax/linear_algebra.py
+++ b/ivy/functional/backends/jax/linear_algebra.py
@@ -174,7 +174,10 @@ def trace(x: JaxArray, offset: int = 0, out: Optional[JaxArray] = None) -> JaxAr
     return jnp.trace(x, offset=offset, axis1=-2, axis2=-1, dtype=x.dtype, out=out)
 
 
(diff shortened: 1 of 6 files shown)


sherry30 (changes requested):
Hi,
can you write the outputs of the examples in the docstring as ivy.array, for example your first example output should be written as 'ivy.array(-10.)' instead of '-10.000000000000002'.

One more thing, make sure no line exceeds more than 88 characters or else lint test will fail. You can separate inputs by putting ' \ ' and moving the rest of the inputs on the next line.

Hab-eeb:
Okay thank you, I'll make the changes now.

Hab-eeb:
So i  saw the lint check failed, i then checked my code again everything is below 88 lines. I saw the issue is from other function docstrings.

sherry30 (changes requested):
I think you accidentally pushed changes to .idea files as well, can you kindly remove those.

One last thing, you missed fixing the output of "With :code:`ivy.Container` input:" example. The rest looks good, I'll merge your PR as soon you make these changes :)

Hab-eeb:
Okay Great Thanks, I'll do that now.

PR #1212, opened 2022-05-27
Reformat image\uint8_img_to_float_img and fix jax\random

JialWu:
- [ ] #1200

Modified the method using the latest formatting rules and added examples. The problem "unexcepted argument" in jax\random was also fixed.


VedPatwardhan (approved):
Hi @JialWu, I have fixed a couple of linter issues with the examples and the parameters, apart from that everything looks good. Thanks for contributing!

PR #1218, opened 2022-05-27
functional.nn.activiations hypothesis integration

kirodev:


kirodev (commented):
Gelu  gives error  E       IndexError: list index out of range for all frameworks backends
[log Gelu.pdf](https://github.com/unifyai/ivy/files/8874106/log.Gelu.pdf)

PR #1260, opened 2022-05-31
Reformatted, function "maximum"

SaadHasib:
Added docstring examples (functional, instance).
Added "Array API Standard" conform string.
Renamed variables in docstring to match argument names.


SaadHasib:
Errors in example, will fix and open new pull request.

PR #1279, opened 2022-06-01
Hypothesis integration: test_norms.py

chie2727:


djl11 (changes requested):
Thanks! Quick question, why are we testing for `ivy.dropout` instead of `norm_layer`? Is that a typo or am I missing something?

djl11 (changes requested):
we should still try to incorporate the test with the `with_v` flag, I think this can also be handled programmatically with varied shaped generated by hypothesis.

PR #1292, opened 2022-06-03
updating argmax function PR

### LINKED ISSUE #1291: argmax

ItzHaad:
resolves #1291 
argmax docstring example addition and slight front-end + back-end reformating


MarShaikh (changes requested):
Hi, you would just need to add an additional example for out used to in place update the input array. As shown [here](https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html#ivy-tan). Could you also change the header of examples from 'Examples' to 'Functional Examples' and also add 'Instance Methods' examples as well. 
Thanks.

PR #1329, opened 2022-06-05
initial commit for hypothesis integration cum stateful changes

RickSanchezStoic:
Initial commit for hypothesis integration of test_layers.py and changes made to the stateful.


ivy/functional/backends/numpy/elementwise.py
```diff
@@ -15,6 +15,9 @@
 def add(x1: np.ndarray, x2: np.ndarray) -> np.ndarray:
     if not isinstance(x2, np.ndarray):
         x2 = np.asarray(x2, dtype=x1.dtype)
+    # numpy.array_api doesn't support float16
+    if x1.dtype == "float16" or x2.dtype == "float16":
+        return np.asarray(np.add(np.asarray(x1), np.asarray(x2)))
```
djl11, line 20: we should just return this version always, also, no need for the `np.asarray` in this case. Also I think this function should be symmetric with regards to `x1` and `x2`

ivy/stateful/initializers.py
```diff
@@ -75,7 +85,13 @@ def create_variables(self, var_shape, device, fan_out=None, fan_in=None):
                 "fan_sum | fan_avg ] "
             )
         wlim = ((self._numerator / fan) ** self._power) * self._gain
-        return ivy.variable(ivy.random_uniform(-wlim, wlim, var_shape, device=device))
+        return ivy.variable(
+            ivy.array(
```
djl11, line 89: `ivy.array` is redundant here

ivy/stateful/module.py
```diff
@@ -358,7 +360,7 @@ def _call(self, *args, v=None, with_grads=None, **kwargs):
         """
         with_grads = ivy.with_grads(with_grads)
         if not self._built:
-            self.build(*args, **kwargs, from_call=True)
+            self.build(*args, **kwargs, from_call=True, dtype=args[0].dtype)
```
djl11, line 363: `args[0]` is very likely to be an array, but strictly speaking it doesn't need to be. We should use one of the nested functions to find the first array argument. Check out [this function](https://github.com/unifyai/ivy/blob/523d745f3ea2492ca7714e1ff654428f4ff640ed/ivy/func_wrapper.py#L163).

ivy/stateful/module.py
```diff
@@ -647,6 +650,8 @@ def __call__(
             track_submod_call_order,
             expected_submod_rets,
         )
+        # if not v:
```
djl11, line 653: if we don't need these lines, then we don't need this comment.

ivy/functional/backends/numpy/elementwise.py
RickSanchezStoic: So we don't use numpy.array_api, ok. Regarding the symmetricity, yes the types are symmetric for most of the cases but for a few test cases, it happened that the dtype of x2 and x1 was different. However, if we are not going to use the numpy.array_api space at all then no need for this line.

ivy/stateful/initializers.py
RickSanchezStoic: Fixed.

ivy/stateful/initializers.py
```diff
@@ -10,9 +10,17 @@ class Constant:
...
         self._constant = constant
 
-    def create_variables(self, var_shape, device, fan_out=None, fan_in=None):
+    def create_variables(
+        self, var_shape, device, fan_out=None, fan_in=None, dtype=None
+    ):
         """Create internal variables for the layer"""
-        return ivy.variable(ivy.ones(var_shape, device=device) * self._constant)
+        return ivy.variable(
+            ivy.array(
```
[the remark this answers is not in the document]
RickSanchezStoic: I get the not using ivy.array part due to redundancy, but the code uses ivy.ones to create variables, why would we use ivy.full?

ivy/stateful/module.py
RickSanchezStoic: done

ivy/stateful/module.py
```diff
@@ -662,20 +667,25 @@ def save_weights(self, weights_path):
...
-    def build(self, *args, from_call=False, device=None, **kwargs):
+    def build(self, *args, from_call=False, device=None, dtype=None, **kwargs):
         """Build the internal layers and variables for this module."""
         self._dev = ivy.default(device, self._dev)
-
         # return False if not from_call but build_mode is on_call
+
         if not from_call and self._build_mode == "on_call":
             return self.v
+        if self._dtype:
```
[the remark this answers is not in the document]
RickSanchezStoic: done

ivy/stateful/module.py
RickSanchezStoic: done

ivy/functional/backends/numpy/elementwise.py
RickSanchezStoic: little more clarification is needed, see my latest commit.

djl11: yeah, we shouldn't use `np.array_api` at all.

ivy/stateful/initializers.py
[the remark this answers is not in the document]
RickSanchezStoic: full instead of ones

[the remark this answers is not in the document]
djl11: we can pass `self._constant` as an argument in `ivy.full`

ivy/stateful/initializers.py
djl11: good point, actually we should change `ivy.random_uniform` so that it does support the `dtype` argument.

ivy/functional/backends/numpy/elementwise.py
```diff
@@ -15,7 +14,8 @@
 def add(x1: np.ndarray, x2: np.ndarray) -> np.ndarray:
     if not isinstance(x2, np.ndarray):
         x2 = np.asarray(x2, dtype=x1.dtype)
-    return np.asarray(npa.add(npa.asarray(x1), npa.asarray(x2)))
+    # numpy.array_api doesn't support float16
```
djl11, line 17: don't need this comment

djl11 (changes requested):
after these two small changes, looks good to merge :)

PR #1331, opened 2022-06-05
example added to sum function doc_string

abubakaralsadiq:


sherry30 (changes requested):
Hi,

Can you remove changes in the .idea files. Also you have to add Functional Examples and Instance Method while following all the points as stated in docstring_examples (https://github.com/unifyai/ivy/blob/master/docs/partial_source/deep_dive/11_docstring_examples.rst).  This includes adding examples using NativeArray, Container etc.
Let me know if you have any questions 😊

PR #1333, opened 2022-06-05
Make pull request, keep issue active

frdysf:
Making an empty pull request to keep the issue active on my end; just completed setup.


mattbarrett98 (commented):
You can't reserve an issue with an empty pull request. Instead you should make a PR with some initial attempt at the coding challenge 🙂

PR #1361, opened 2022-06-08
adding test_num_arrays_on_dev

Ishticode:
@djl11 
The test sets a default device and creates a certain number of arrays (0 - 5) on the device. The number of arrays is then tested against the value given by `ivy.num_ivy_arrays_on_dev`. **Currently passing**. I assume it only requires checking the number of arrays, really. Could you make sure if this is accurate and if I missed any corner cases? Thanks.


ivy_tests/test_ivy/test_functional/test_core/test_device.py
```diff
@@ -585,11 +585,25 @@ def test_profiler(device, call):
         time.sleep(1)  # required by MXNet for some reason
 
 
+@given(num=st.integers(0, 5))
+def test_num_arrays_on_dev(num):
+    arrays = {"x": []}
+    arr_device = ivy.default_device()
+    arr_device = ivy.as_ivy_dev(arr_device)
```
djl11, line 592: this can one combined with the line above

ivy_tests/test_ivy/test_functional/test_core/test_device.py
```diff
@@ -585,11 +585,25 @@ def test_profiler(device, call):
...
+    if num:
+        for _ in range(num):
+            # arr size fixed as its irrelevant.
+            arr = ivy.array(np.random.uniform(size=2))
+            arrays["x"].append(arr)
+        assert ivy.num_ivy_arrays_on_dev(arr_device) == num
+    elif num == 0:
```
djl11, line 599: no need for this `if` and `elif`, the `for` loop will not enter if `num == 0`

djl11 (changes requested):
Some small changes requested. After those it's good to go I think! Feel free to merge yourself.

ivy_tests/test_ivy/test_functional/test_core/test_device.py
```diff
@@ -585,11 +585,21 @@ def test_profiler(device, call):
         time.sleep(1)  # required by MXNet for some reason
 
 
+@given(num=st.integers(0, 5))
+def test_num_arrays_on_dev(num, device):
+    arrays = {"x": []}
+    for _ in range(num):
```
djl11, line 591: I'm being picky now, but you can actually do this in one line: `arrays = {"x": [ivy.array(np.random.uniform(size=2)) for _ in range(num)]}` :)

djl11 (changes requested):
one last point!

PR #1363, opened 2022-06-09
Reformatting DefaultDevice.__enter__

kimbring2:
For the issue #1362


kimbring2:
@mattbarrett98 Can you please take a look at this PR for me?

mattbarrett98 (changes requested):
`__enter__` is called when a `with` block is used, so perhaps an example using a `with` block and an explanation would be more appropriate here 🙂

kimbring2:
@mattbarrett98 Thank you for reviewing. I modify the code according to your advice +1:

mattbarrett98 (commented):
Thanks 🙂

kimbring2:
@mattbarrett98 Thank you for merging. I check we can use 'from typing import Self'. 

I just find it from 'https://peps.python.org/pep-0673/'. However, it seems like it is just a proposal.

I try to use the same class as the return value to the DefaultDtype class. 

`
class DefaultDtype:
    """"""

    # noinspection PyShadowingNames
    def __init__(
            self,
            dtype: Union[ivy.Dtype, ivy.NativeDtype],
    ) -> None:
        self._dtype = dtype

    def __enter__(self) -> DefaultDtype:
        set_default_dtype(self._dtype)
        return self
`

And it gives the error say 'E   NameError: name 'DefaultDtype' is not defined'. Is there any proper way to return the class itself in Python?

Thank you.

PR #1384, opened 2022-06-10
Solving pip install fails on windowsOS

vinesmsuic:
If you do `pip install .` in WindowsOS you will get error. 
```
UnicodeDecodeError: 'cp932' codec can't decode byte 0x93 in position 471: illegal multibyte sequence
```

This PR fix the problem so windows users could build the lastest fork through `pip install .`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,9 +63,12 @@ def _is_raw_block(line):
         return True
     return False
 
+def read_description(path):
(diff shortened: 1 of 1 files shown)


vinesmsuic:
#1399

djl11 (approved):
lgtm! Thanks!

PR #1387, opened 2022-06-10
hypothesis integration for test_norms

Aarsh2001:


VedPatwardhan (changes requested):
Hey Aarsh, could you please see if you can use the test_layer_norm with the numpy backend to get the target to compare the test with, because for testing we are using numpy output as the ground truth, so it would be better to use it instead of a new helper function. Thanks!

PR #1388, opened 2022-06-10
fix docker image build failed

jackhong12:
When I build docker image by running `docker build -t unifyai/ivy:latest --no-cache -f Dockerfile .`, it shows the following error message.
docker_build_failed

The last step in Dockerfile will execute `test_dependencies.py` to check all modules. But it cannot find torch-scatter and raise an exception, which causes docker build failure. The reason is that the module name of torch-scatter is `torch_scatter` instead of `torch-scatter`.
The simple fix is adding mod_name in optional.txt.

### DIFF
--- a/optional.txt
+++ b/optional.txt
@@ -10,7 +10,7 @@ tensorflow
 tensorflow-addons # mod_name=tensorflow_addons
 tensorflow-probability # mod_name=tensorflow_probability
 torch
-torch-scatter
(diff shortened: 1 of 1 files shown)


simonetgordon:
Hi Jack,

Thanks for raising this issue, will look into it!

djl11:
Thanks!

PR #1389, opened 2022-06-10
functional_losses fix up

juliagsy:
in progress


ivy/array/losses.py
```diff
@@ -1,8 +1,37 @@
...
 class ArrayWithLosses(abc.ABC):
+    def cross_entropy(
+        self: ivy.Array,
+        pred: Union[ivy.Array, ivy.NativeArray],
+        axis: Optional[int] = -1,
+        epsilon: Optional[float] = 1e-7,
+        *,
+        out: Optional[ivy.Array] = None
+    ) -> ivy.Array:
+        return ivy.cross_entropy(self, pred, axis=axis, epsilon=epsilon, out=out)
```
djl11, line 18: `self._data` should be passed in rather than `self` for all instance methods. This is a small detail, but is better for consistency with the other functions. This point is explained [here](https://github.com/unifyai/ivy/blob/e7712dffabe126fc795d52c2fd0d5b05a91b064f/docs/partial_source/deep_dive/4_arrays.rst#ivy-array).

ivy/container/losses.py
```diff
@@ -1,9 +1,106 @@
...
 # local
+import ivy
 from ivy.container.base import ContainerBase
 
-# ToDo: implement all methods here as public instance methods
-
 
-# noinspection PyMissingConstructor
 class ContainerWithLosses(ContainerBase):
+    def cross_entropy(
```
djl11, line 10: All `ivy.Container` instance methods should make use of the static method internally. Check out the latest [docs on the container](https://github.com/unifyai/ivy/blob/e7712dffabe126fc795d52c2fd0d5b05a91b064f/docs/partial_source/deep_dive/5_containers.rst) for details. The examples given in the container docs are for `add`, `tan` and `roll`.

djl11 (changes requested):
Once the comments are addressed, please request another review and I'll take another look :)

juliagsy:
Hey Dan! Thanks for the review! I've made the changes accordingly! Also, the hypothesis test was included previously so its not changed in this PR. Would you like to browse through and leave comments if there are any further improvements to make? Thank you!

ivy/functional/ivy/losses.py
```diff
@@ -156,7 +156,7 @@ def binary_cross_entropy(
     """
     pred = ivy.clip(pred, epsilon, 1 - epsilon)
     return ivy.negative(
-        ivy.add(ivy.log(pred) * true, ivy.log(1 - pred) * (1 - true), out=out), out=out
+        ivy.add(ivy.log(pred) * true, ivy.log(1 - pred) * (1 - true)), out=out
```
djl11, line 159: we should use the `out` argument for both the `add` function and also the `negative` function. These functions are called sequentially, and so the same memory can be used for storing the result of both operations.

djl11 (changes requested):
aside from the small change requested, it looks good to go!

juliagsy:
Noted and I'll keep that in mind! Thanks for the review, Dan!

PR #1396, opened 2022-06-11
Reformatted the is_native_array function (#1382)

nbalcarc:
#1382


MarShaikh (changes requested):
Hi, can you update the type hints in ivy as well as all the other back ends? Thanks.

nbalcarc:
I've fixed the failing check, all checks should be passing now.

MarShaikh (changes requested):
Hi, can you update the type hints for input x? From your example, x takes inputs both ivy arrays and native arrays. Also, could you add an example of using exclusive as one of the parameters? Thanks.

MarShaikh (approved):
Great, looks good!

PR #1408, opened 2022-06-11
resolves  sqrt #1391

Al-ameen007:


Al-ameen007:
issue #1391

ivy/functional/ivy/elementwise.py
```diff
@@ -1275,7 +1275,74 @@ def sqrt(
...
+
+    Using :code:`ivy.Array` instance method:
+    
+    >>> x = ivy.array([[1., 2.],  [3., 4.]])
+    >>> y = x.sqrt()
+    >>> print(y)
+    ivy.array([[1.  , 1.41],
+           [1.73, 2.  ]])
 
+    Using :code:`ivy.NativeArray` instance method:
```
chie2727, line 1327: Hi @Al-ameen007 ,
Sorry for the delay in having your PR reviewed.

Your edits are looking good so far, just a few changes you need to make.

1. Have a look at the description in the link below (from line 209 to 216) and add the same thing into your `sqrt` docstring.

https://github.com/unifyai/ivy/blob/7b741439dc15db949a27d6deb530c3f17dc49e41/ivy/functional/ivy/elementwise.py#L209
(Also remember to update the URL to the correct one for `sqrt`)

2. There's no need to include a `ivy.NativeArray`  instance method since the output will vary depending on which backend is being used, so you can remove the example at line 1327.

chie2727 (changes requested):

chie2727 (changes requested):
Hi @Al-ameen007 ,
There seem to be a number of commits from other users showing up in your PR. Could you tidy this up so that only your commits and edits appear?
(If you need to make a new PR to do that and leave a comment here with a link to the new PR, then that's also not a problem.)

Besides that, one last edit you need to make to the sqrt docstring is have an example that uses `out=x` in the `ivy.Array input` section.
I can see you have one in the `ivy.NativeArray input` section so you can move out and implement that example into the `ivy.Array input` section instead.

Al-ameen007:
Hello, I've added an example as you have requested, but for the first part where you have asked me to tidy the PR or create a new one, it seems that even when I create a new branch on my fork and try to create a new PR with that branch, all the changes from this PR show up there, meaning I'm unable to tidy it up there also.

chie2727 (changes requested):
Hi @Al-ameen007 , thanks for making the changes.
As for merging this PR I can only do that after the commits have been cleaned up because there is a high chance that another user's changes might be removed after the merge.

It seems that you are working from the master branch of your fork, hence why creating a new branch keeps the previous commits made.
What I would recommend is that you create a new fork from unifyai/ivy and then immediately create a new branch from there, and then make a PR.

As a side note, it's always good practice to carry out your development work on a branch that isn't the master branch mainly to avoid issues with commits and conflicts.

Al-ameen007:
this is the link for the new PR, if there are any problems , please let me know.
https://github.com/unifyai/ivy/pull/1714

Issue #1415, opened 2022-06-12
to_numpy


davidlee1102:
"""Converts an array into a numpy array.

    Parameters
    ----------
    x
        input array

    Returns
    -------
    ret
        a numpy array copying all the element of the array ``x``.

    Examples
    --------
    >>> x = ivy.array([-1, 0, 1])
    >>> y = ivy.to_numpy(x)
    >>> print(y)
    [-1 0 1]

    >>> print(type(y))
    <class 'numpy.ndarray'>

    """

davidlee1102:
https://github.com/unifyai/ivy/pull/1416

PR #1416, opened 2022-06-12
to_numpy #1415

davidlee1102:
"""Converts an array into a numpy array.

    Parameters
    ----------
    x
        input array

    Returns
    -------
    ret
        a numpy array copying all the element of the array ``x``.

    Examples
    --------
    >>> x = ivy.array([-1, 0, 1])
    >>> y = ivy.to_numpy(x)
    >>> print(y)
    [-1 0 1]

    >>> print(type(y))
    <class 'numpy.ndarray'>

    """


VedPatwardhan (changes requested):
Hi @davidlee1102, thanks for contributing!
There are few changes needed to be made to your PR before we get it merged:
1. You seem to have modified the functions `array_equal` and `to_numpy` to internally use the `np.array` and `np.array_equal` functions respectively, which shouldn't be done, because `_cur_backend` refers to the backend framework set by the user, so depending on what backend framework is set, the corresponding backend function for `to_numpy` would be called automatically, which enables writing framework-agnostic code. So, please revert those changes back to using `_cur_backend`.
2. The task you've been assigned seems to be for the `to_numpy` function, but you seem to have updated the docstring of the `array_equal` function too, please revert that change. Feel free to create another PR for the `array_equal` function if it is also assigned to you. This PR should be only modifying the `to_numpy` function.
3. The docstring for the `to_numpy` function needs to be modified by adding examples for different kinds of inputs as mentioned [here](https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html). Feel free to read through this page and let me know if any doubts you might have.
Thanks🙂

davidlee1102:
Hello, I have some questions about this PR:
1. Question 1:
A. Is that the ivy.array is 1D right, and 2D right? | If 2D, does each same length or different length (For example: [[1,2,3,4,5],[1,2,3,4,5,6]])?
B. If return _cur_backend(x).to_numpy(x), so I did check that function named to_numpy is another function; So could you please check my code again?
2. Got it! I have clean my code on array_equal def
3. Got it!

VedPatwardhan:
> Hello, I have some questions about this PR:
> 
> 1. Question 1:
>    A. Is that the ivy.array is 1D right, and 2D right? | If 2D, does each same length or different length (For example: [[1,2,3,4,5],[1,2,3,4,5,6]])?
>    B. If return _cur_backend(x).to_numpy(x), so I did check that function named to_numpy is another function; So could you please check my code again?
> 2. Got it! I have clean my code on array_equal def
> 3. Got it!

Yeah, so `ivy.array` can have any shape just like with a `numpy.array` or a `torch.tensor`. The arrays can be **multidimensional**, so in case of a 2D array, we should be having all arrays in it with the same length 🙂

VedPatwardhan (changes requested):
1. The `to_numpy` function doesn't require you to use `np.array` before calling the function, so you can just remove line 412 from the code
2. In the example to the docstring, you have passed `to_numpy` as a parameter to the `ivy.array` function, you don't need to do that either, please revert it to passing the array as a parameter as it was before.
Feel free to ask any further questions 🙂

davidlee1102:
So what is the problem I need to solve? because I did see 
"""
Parameters
    ----------
    x
        input array

    Returns
    -------
    ret
        a numpy array copying all the element of the array ``x``.
"""
I am a new for contributing so I hope that you can instruct me how to contribute to IVY.

VedPatwardhan:
> So what is the problem I need to solve? because I did see """ Parameters ---------- x input array
> 
> ```
> Returns
> -------
> ret
>     a numpy array copying all the element of the array ``x``.
> ```
> 
> """ I am a new for contributing so I hope that you can instruct me how to contribute to IVY.

Actually, we can't solve this for you. Please check the deep dive (https://lets-unify.ai/ivy/deep_dive.html) and the open tasks (https://lets-unify.ai/ivy/contributing/4_open_tasks.html) to understand the reformatting task, and let us know if you have any more specific questions that we can clarify for you 🙂

davidlee1102:
Thank you, let me see

davidlee1102:
Hi, could you please check the code again, thank you

davidlee1102:
Hello, could you please check on my code again 👍

davidlee1102:
1. I have converted back the searching.py 
2. I have removed the error
3. I did check the pseudocode on this to_numpy function, I see the input X and output is numpy arr X, so I have solved by "try catch", but you told me no need to "try catch" this function so I have turn to "if else"; Please check again

VedPatwardhan (changes requested):
Hi @davidlee1102, you don't need to add an if-else block to the code either, `return _cur_backend(x).to_numpy(x)` enables calling an appropriate backend function without it. Because of this, you only need to have `return _cur_backend(x).to_numpy(x)` in the body of the function. Secondly, now I see a lot more functions modified in general, please make sure that you only modify the `to_numpy` function and not anything else.
Thanks🙂

davidlee1102:
1. Thank for instructing me to contribute the code, I did not modify many functions in general, I just edited the flake8 to the general.py; So, I have recovered to the original general.py and add my pseudocode for to_numpy function
2. This is the first one to understand how contribute to the code, thank for your advising
3. Let see my code again, if it ok, please return comment lgtm 😄 , if not please let me know

VedPatwardhan:
No worries! There are just a couple of final updates before we merge your PR. Firstly, please remove the extra tab space at the start of the line from all lines in the docstring of the `to_numpy` function and secondly, please merge the current Ivy master into your fork as your fork is 16 commits behind master currently. Rest looks good to me 🙂

davidlee1102:
I have done, please check, thank you 😃 @VedPatwardhan

VedPatwardhan (approved):
Looks good to me, thanks for contributing!

PR #1418, opened 2022-06-12
implicit backend setting changed

Aarsh2001:
Do we need an additional global variable? if the task can be accomplished without it.


Aarsh2001:
This doesn't solve the problem, will update !

VedPatwardhan:
Sure, take your time 🙂

PR #1420, opened 2022-06-12
tolereance dict changes to test_array_function

RickSanchezStoic:
Changes made as requested. I have added "bfloat16" as a dtype, but haven't encountered it till now, so don't actually know what would be a good tolerance value for this dtype. As of now, I have set its tolerance value to be the same as that of "float16". However, if anyone faces any issues around this, then they can change its value.


djl11 (changes requested):
`input_dtype` should actually be a list of length equal to the number of arrays in the input. Therefore, we should not give any special attention to `input_dtype[0]`. We should instead check against all entries in the list, not just the first :)

djl11 (changes requested):
all of this `rtol` logic should go after the `test_values` check, thanks!

ivy_tests/test_ivy/helpers.py
```diff
@@ -559,11 +557,23 @@ def test_array_function(
...
+    if not rtol:
+        if ret.dtype in tolerance_dict:
+            rtol = tolerance_dict[ret.dtype]
+        else:
+            rtol = 1e-05
     # flatten the return
     if not isinstance(ret, tuple):
         ret = (ret,)
-    if input_dtype == "bfloat16":
+    if "bfloat16" in input_dtypes:
```
djl11, line 576: This should be called before we compute `ret_from_np`, which will currently throw an error for `bfloat16` data types.

PR #1421, opened 2022-06-12
removed private imports from ivy 

RitujaPawas:


djl11 (changes requested):
actually sorry, could you please fix the lint errors?

VedPatwardhan (changes requested):
I'd suggest a few changes:
1. In `ivy/functional/backends/jax/device.py`, the `_to_dev` function seems to be commented out, please uncomment it as it is required in cases where the `to_device` function cannot be used. Basically, the `_to_dev` function is an unwrapped function, which is used in place of `to_device` when we don't want the function to return an `ivy.Array`.
2. Likewise, please revert back all instances where the `_to_dev` function was replaced by `to_device` to use `_to_dev` again for the reason specified in 1
3. Please also resolve merge conflicts.
Once you're done with these changes, feel free to merge your PR. Thanks 🙂

djl11 (commented):
I've delegated to Ved for the review :)

VedPatwardhan (approved):
lgtm! feel free to merge the PR 🙂

PR #1424, opened 2022-06-12
Docs: venv and conda setups

Ishticode:


Ishticode:
@djl11 
I had the concern about using Ivy from GitHub source as a package. This is easy enough with all requirements automatically installed BUT leads to a static install which will be outdated right the next day kind of. If you think this fine then this pr is pretty much done. Else I think it's better for people to be using their clone as the ivy source. This way they can create a .py file outside the ivy folder. `import ivy` in that and run functions to generate examples and it can be updated everyday on the fly. This way they can still have PyPI version installed or using docker interpereter but it gets overridden in the import by the cloned one when in the same directory. Let me know what you think.

PR #1448, opened 2022-06-15
Numeric support in operator functions

juliagsy:
I've only added numeric support for the add() method for now, if the changes look good, I'll continue with the rest! Thanks!


ivy/functional/backends/numpy/elementwise.py
```diff
@@ -11,7 +12,7 @@
     _erf = None
 
 
-def add(x1: np.ndarray, x2: np.ndarray) -> np.ndarray:
+def add(x1: Union[float, np.ndarray], x2: Union[float, np.ndarray]) -> np.ndarray:
     if hasattr(x1, "dtype") and hasattr(x2, "dtype"):
         promoted_type = np.promote_types(x1.dtype, x2.dtype)
         x1, x2 = np.asarray(x1), np.asarray(x2)
```
djl11, line 18: the line `x2 = np.asarray(x2, dtype=x1.dtype)` below will throw an error if x1 is not an array.

ivy/functional/backends/tensorflow/elementwise.py
```diff
@@ -19,8 +19,8 @@ def bitwise_left_shift(
 
 
 def add(
-    x1: Union[tf.Tensor, tf.Variable],
-    x2: Union[tf.Tensor, tf.Variable],
+    x1: Union[float, tf.Tensor, tf.Variable], 
+    x2: Union[float, tf.Tensor, tf.Variable]
 ) -> Union[tf.Tensor, tf.Variable]:
     if hasattr(x1, "dtype") and hasattr(x2, "dtype"):
         promoted_type = tf.experimental.numpy.promote_types(x1.dtype, x2.dtype)
```
djl11, line 26: the line `x1 = tf.constant(x1, dtype=x2.dtype)` below will throw an error if `x2` is not an array. Please check for these kinds of issues for all backends.

djl11 (changes requested):
Please address the comments and then re-request a review. In general, the workflow for all functions should probably be something like:

1. if neither inputs are arrays, return the simple operation performed on the two numbers
2. otherwise, explicitly ensure both inputs are arrays, by converting non-arrays to arrays
3. continue with the function as normal

PR #1457, opened 2022-06-15
cumprod #1446

Dante-Mazzini:
#1446


ivy/functional/ivy/general.py
```diff
@@ -1497,6 +1497,17 @@ def cumprod(
     -------
     ret
         Input array with cumulatively multiplied elements along axis.
+        
+    Examples
+    --------
+    >>> x = ([a, b, c])
```
bhattbhuwan13, line 1503: I think the type of array should be made clear `ivy.Array()` or `ivy.NativeArray`

ivy/functional/ivy/general.py
```diff
@@ -1497,6 +1497,17 @@ def cumprod(
...
+    >>> ivy.cumprod(x)
+    [a, a * b, a * b * c]
+
+    >>> x = ([a, b, c])
```
bhattbhuwan13, line 1507: Maybe add an example that makes use of the `axis` parameter. Include a multi-dimensional array in the same.

Dante-Mazzini: I added the example, same array but with two axes selections. I chose a simple matrix, in order to be easily understood.

ivy/functional/ivy/general.py
bhattbhuwan13: Since the function is calling backend specific methods`_cur_backend(x).cumprod(x, axis, exclusive, out=out)`, I think it should take `ivy.NativeArray` as well. @saeedashrraf , what do you think?

Dante-Mazzini: I might be wrong, but my process of thought, was that if the cumprod function is native, even if it calls a backend method, it will return an ivy.Array(). This I saw in the descriptions of the two methods (ivy.Array() and ivy.NativeArray()).
But I share the opinion that the function can take both. 
What would be the proposition? do you want me to alter the method of arranging the array in the examples, or do you prefer if I clarify in the section parameters, that you can use both methods (this part was merged 2 months ago)
I thought that it was implicit in the definition of parameters, but there is no harm in clarifying the subject even more.
I am at your disposal.

saeedashrraf (changes requested):
Hello @Dante-Mazzini, for the examples, make sure to follow the guide [here](https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html#docstring-examples). This essentially means that your examples must cover the relevant points mentioned in the docs. This also means you should use numerical values for your examples instead of using arbitrary values such as a, b, or c. Let me know if you have any questions.

Dante-Mazzini:
Hi! thanks for the observation, I understood your request, and already changed it. Hope this are the changes that you requested. If you need me to change anything more, please send me an example, of a function that is already in perfect format. 
Best regards.

ivy/functional/ivy/general.py
Dante-Mazzini: all cleared reading the guide, chapter 11, you were right! added this example, and a container one, with the modifications requested in the review!

saeedashrraf (changes requested):
Hello @Dante-Mazzini , can you make sure to pass the right argument value in the correct position. If the argument exclusive is being passed as the second argument as in [here](https://github.com/unifyai/ivy/pull/1457/files#diff-f6556baf0d4607e70527506a4c173c78f8e3452e3b1c10d4d356210c527b41c3R1694), make sure to add the keyword exclusive before it, making it ivy.cumprod(x, exclusive=exclusive). 

Also, make sure that all keywords arguments are followed by keyword arguments and not by positional arguments as in [here](https://github.com/unifyai/ivy/pull/1457/files#diff-f6556baf0d4607e70527506a4c173c78f8e3452e3b1c10d4d356210c527b41c3R1714)

The N should not be capitalized [here](https://github.com/unifyai/ivy/pull/1457/files#diff-f6556baf0d4607e70527506a4c173c78f8e3452e3b1c10d4d356210c527b41c3R1723)

Thank you for your effort!

saeedashrraf (changes requested):

ivy/functional/ivy/general.py
```diff
@@ -1813,6 +1813,62 @@ def cumprod(
...
+    >>> print(y)
+    ivy.array([1, 2, 6])
+    
+    Example specifying axes
+    
+    >>> x = ivy.array([[2, 3], \
+                       [5, 7], \
+                       [11, 13]])
+    >>> exclusivity = True
+    >>> ivy.cumprod(x, axis=1, exclusive=exclusivity, out=y)
```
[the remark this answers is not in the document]
Dante-Mazzini: thank you so much for the clarification, i redeemed to last examples were y = zeros() wasn't necessary, were it was given as called in the previous example! Now it's clear to me!

saeedashrraf (commented):
There is no harm in having the variable named as exclusive and the keyword being named exclusive as well. For consistency with other examples, let's actually have them be named exclusive. Thanks!

Dante-Mazzini:
That would be all the changes in the last reviews! Hope this time we got it right! thanks for all the help!

saeedashrraf (approved):

saeedashrraf:
Yes, looks good to me. Thank you for your effort @Dante-Mazzini 🎉

PR #1462, opened 2022-06-16
flatten() function

Ishticode:
@djl11 
Couldn't find `flatten()`. If it is implemented with a different name then i will close this. 
If you think this is good addition then i will write up some docstring and typehints and fix the test.
Just felt it was needed at times...

### DIFF
--- a/ivy/functional/backends/tensorflow/manipulation.py
+++ b/ivy/functional/backends/tensorflow/manipulation.py
@@ -167,7 +167,8 @@ def split(x, num_or_size_splits=None, axis=0, with_remainder=False):
 
 def repeat(
     x: Union[tf.Tensor, tf.Variable],
-    repeats: Union[int, List[int]], axis: int = None,
+    repeats: Union[int, List[int]],
+    axis: int = None,
 ) -> Union[tf.Tensor, tf.Variable]:
     ret = tf.repeat(x, repeats, axis)
     return ret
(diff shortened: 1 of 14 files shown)


ivy/functional/backends/tensorflow/manipulation.py
```diff
@@ -231,3 +232,10 @@ def clip(
     else:
         ret = tf.clip_by_value(x, x_min, x_max)
     return ret
+
+
+def flatten(x):
+    shape = 1
+    for dim in x.shape:
+        shape *= dim
+    return tf.reshape(x, shape)
```
djl11, line 241: Does TensorFlow not support `tf.reshape(x, -1)`?

djl11: If it does not, you can do this in one line like so:

```
import math
tf.reshape(x, math.prod(x.shape))
```

djl11 (changes requested):
Please respond to the comments, and then re-request a review.

ivy/functional/backends/tensorflow/manipulation.py
Ishticode: yh I will change this.

Ishticode:
ivy.reshape(x, -1) works with all except pytorch which requires shape to be tuple.

I will just add this bit to pytorch in reshape:
```
if isinstance(shape, int)
    shape = (shape, )
```

Ishticode:
Will reopen and merge.

PR #1482, opened 2022-06-17
Unsupported dtype attribute

chie2727:
・Unsupported dtype attribute added to elementwise functions
・2 helper functions made and added to ivy/functional/ivy/data_type.py
・'return' statement for unsupported dtypes removed from elementwise tests


ivy/functional/backends/numpy/elementwise.py
```diff
@@ -20,6 +20,7 @@ def add(x1: np.ndarray, x2: np.ndarray) -> np.ndarray:
     if not isinstance(x2, np.ndarray):
         x2 = np.asarray(x2, dtype=x1.dtype)
     return np.add(np.asarray(x1), np.asarray(x2))
+add.unsupported_dtypes = tuple(['float16'])
```
djl11, line 23: I just chose this one at random, and it seems as though `numpy.add` does support `float16`. This code runs without error:

```
import numpy as np
x = np.array([1.], dtype='float16')
y = np.array([2.], dtype='float16')
print(np.add(x, y))
```

djl11 (changes requested):
Could you please go through all of these again and make sure they are correct. Some data types might currently be skipped in the tests because of edge cases, and so we shouldn't use the currently skipped data types in the tests as a ground truth for the unsupported data types. Thanks.

chie2727:
@djl11 
I see! I had assumed that tests which had been skipped primarily in PyTorch for float16 data types didn't support float16.
I'll go through these and check for all of them whether they do or not.

chie2727:
@djl11 
I've gone through every single test in the elementwise module and made sure of which functions do have unsupported data types and which are failing for other reasons and updated their respective attributes.

If everything looks okay I'll squash and merge in the changes

djl11 (changes requested):
no need to pass a list into the constructor of the tuples, you can just do `tuple('some_str')` :)

chie2727:
When passing a string into a tuple, each character is treated as an individual string and gets splits up,
hence why I put it into a list first to prevent this.

clip_1

djl11 (changes requested):
yes but if you do it like this `('some_str',)` then it works.

There is no need to use the keyword `tuple` at all. Tuples can be instantiated with brackets and commas alone, and this also avoids the problem without needing to cast to a list first.

It's a small detail really, but one which is just a little bit cleaner I think :)

chie2727:
@djl11 
I see! In my individual task description you specified the attributes should be added like so:
`acosh.unsupported_dtypes = tuple(ivy.float16,)`
hence why I thought it was necessary to use the tuple keyword.
I'll clean this up and commit the changes!

djl11 (approved):
lgtm!

PR #1502, opened 2022-06-17
handle_nestable decorator

Aarsh2001:
added the decorator to ``` optimizer_update ``` and ``` eigh ```  for testing the functionality. It works.

### DIFF
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -3,6 +3,7 @@ import inspect
 import functools
 from types import FunctionType
 from typing import Callable
+from .container import _wrap_function as _wrap_function_container
 
 NON_WRAPPED_FUNCTIONS = [
     "copy_nest",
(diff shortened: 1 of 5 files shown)


ivy/func_wrapper.py
```diff
@@ -369,6 +370,12 @@ def new_fn(*args, out=None, **kwargs):
 
 def handle_nestable(fn: Callable) -> Callable:
     fn_name = fn.__name__
+    try:
```
djl11, line 373: The static container functions must be defined at import time, the methods are public and it should be possible to call them in isolation of the nestable behaviour of the functions in the functional API. It should be possible to solve this at import time, by re-arranging some things, but it's not an easy feat. Let me know if you have any questions!

djl11 (changes requested):

ivy/func_wrapper.py
Aarsh2001: Understood, will do. Thanks.

PR #1503, opened 2022-06-17
Reformat randint

CheezBallzPi:
Solves #1494.

I'm not sure how to run the style tests locally, so I didn't run them. If you could let me know how, I'll do that when I deal with any changes.


saeedashrraf:
Thank you for your effort @CheezBallzPi

PR #1504, opened 2022-06-17
reformat shape_to_tuple

mcandemir:
Hi, I've got a question on testings. I was getting errors, so I checked the logs and I found out that some of those tests aren't ready yet (e.g.: shape_to_tuple). Not sure if I'm right, but it'll be awesome if you give some information about this. Thank you.


mcandemir:
Looks like rebasing messed the commit history a little, sorry about that! Resolves #1479

simonetgordon:
Hi, 

Thanks for making a contribution 😊

Yes indeed, the unit tests for backend functions is a work in progress. We are implementing tests using the hypothesis suite.

Currently, `shape_to_tuple` is throwing a warning in test-docstrings/run-docstring-tests. To view this yourself, click 'details' beside this check and expand 'Run Docstring Tests`, then search for `shape_to_tuple`. This means the docstring examples for this function are failing for some reason in the master branch. 

The formatting and use of function all look reasonable to me. Since the source code is constantly evolving, could you double check that the printed output in the docstrings match the printed output using the master branch of ivy? You can create a .py with these examples within your local clone of the master branch ivy repo and `import ivy` from there. This should import the master version of ivy. A good way to check if you are in fact using the master branch is to run `ivy.set_backend('numpy')` and see if it works. In the stable branch, the equivalent instruction is `ivy.set_framework('numpy')`.

Simone

mcandemir:
Hi,

Thank you for the detailed review. I have a question.. just to clearify: 

Q-) Don't we use a feature branch instead of master for contributing? I wrote this docstring in shape_to_tuple branch, so testing on my local master is meaningless. But still I edited the Ivy repository (from the packages) and I get correct results as given.

Screenshot_20220621_005526

Forgive me if I misunderstood, still learning.

PS: just realized I wrote 0s instead of 1s in examples. Fixing it.

simonetgordon:
Hi,

When you `import ivy` in this manner on a jupyter notebook, it's the 'stable version' of the code. The 'master version' of the code is the one that the ivy team are making changes to continuously. In order to run the most up-to-date version of this function, go your cloned ivy directory locally and create and run this code within a `.py` file. You can run this file on the command line/terminal with `python x.py`. You should be in the `ivy/` directory when you run it. 

Are you using an IDE like PyCharm or VSCode?

mcandemir:
Hi, I made some tests under the local directory just as you suggested. It looks good to me. (tested on PyCharm).

I prepared the docstring for only torch backend, but it was chosen randomly. Is there any specific backend that we must prepare our docstrings with?

Screenshot_20220621_113837

simonetgordon:
Hi,

Ok great, you have the py file importing master ivy. Now all you need to do is copy and paste your docstring examples into this file (removing the >>> prompt) and run the file via the terminal (bottom left tab) as I mentioned before.

You should see that the output that is produced differs from what you currently have in the docstring examples. I obtain the output `(1.0, 2.0, 3.0)` from the first example, for instance. I think this is the cause of the warnings.

And no, you don't need to run with a specific backend.

Kind Regards, 

Simone

mcandemir:
Hi,

Interestingly I'm getting different results when I use the different backends. I'm not sure what is causing this problem.

If I use `numpy` as a backend, I get this propriate result: [a_numpy_ss](https://user-images.githubusercontent.com/55599562/174768874-ac90d747-40bc-4ba0-96ce-f3b27f6b117f.png)

But when I use `torch` instead, I get a more detailed result: [a_torch_ss](https://user-images.githubusercontent.com/55599562/174768893-70cd2039-f51b-4406-8ecb-12d07f1e5246.png)

As a temporal solution, I changed the backend to `numpy` and the following outputs are printed: backend_numpy

simonetgordon:
Hi,

Hmm, that shouldn't be happening. Thanks for flagging this. I will look into it. I would recommend that, for now, just leave the docstrings examples as they appear in numpy and I will merge when all the tests have run. It might still fail the docstrings test, but that's OK.

Thanks again!

Simone

mcandemir:
Hi,

I have made the desired changes in the doc-strings, should be ready now. Thanks for clearing it up!

simonetgordon:
Hi,

Thanks but could you include examples using native_array.shape, ints and lists as the function signature indicates? We're aiming for the docstrings to me as complete and exhaustive as possible. You can still use the [contributor guide](https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html#docstring-examples) for some guidance on the key aims.

Also perhaps you need to pull from master, but there an extra line in device.py which is causing lint to fail under your 'files changed'

Thanks!

Simone

simonetgordon:
Hi,

Are you able to fix the extra line in the device.py?

Also, I will get back to you re the extra examples 😊

Simone

mcandemir:
Yeah! Just forgot to push. Pushing the repo now :blush:

simonetgordon:
Hi, 

Sorry for the delay. Lists could look like:
```
x = [ivy.array([[1., 2., 3.], [4., 5., 6.]]).shape, ivy.array([[1., 2., 3.], [4., 5., 6.]]).shape]
print(ivy.shape_to_tuple(x))
# ((2, 3), (2, 3))
```

and tuples:
```
x = (ivy.array([[1., 2., 3.], [4., 5., 6.]]).shape, ivy.array([[1., 2., 3.], [4., 5., 6.]]).shape])
print(ivy.shape_to_tuple(x))
# ((2, 3), (2, 3))
```

Please add examples with these input types using the correct format.

Kind Regards, 

Simone

mcandemir:
Hi, sorry for the delay, I must've missed the e-mail. I have added the examples that you asked and went through all 3 checks, fixed the errors.

Also there is a thing that might worth to check. I think all `test_docstring.py` tests' expectations are based on torch. When I run the file, it starts to test docstrings for all jax, numpy, tensorflow and torch libraries. Although their outputs differ, the `test_docstrings.py` expect outputs based on torch. 

Backends;
[numpy](https://user-images.githubusercontent.com/55599562/177004334-731a71a6-7874-401a-8c1f-a00c04ab3605.png)
[torch](https://user-images.githubusercontent.com/55599562/177004373-ab217a72-8356-4432-9ce6-ec49c003eb7c.png)
[jax](https://user-images.githubusercontent.com/55599562/177004395-a0790349-ccc9-4a39-b626-22a62065db68.png)
[tensorflow](https://user-images.githubusercontent.com/55599562/177004412-0653ff88-9463-421b-9c97-db759e26365c.png)

`test_docstring.py` expectations:
[expectation](https://user-images.githubusercontent.com/55599562/177004589-1754f885-1da7-47ae-a3c7-bfb97a55e2be.png)


I wasn't sure if this was intentional or not, so I wanted to flag this.

simonetgordon:
Hi,

Thank you for flagging this. Indeed, the backends should produce the same output for all backends, with no framework-specific objects like `torch.size` included. This particular function is just an ivy function currently, with no backend-specific implementations. They may need to be implemented in order to rectify this issue.

I will get back to you later today about whether to merge this in its current docstring-test passing form or whether we should include the framework-agnostic output and fix the torch and tensorflow backends later.

Thanks again!

Simone

simonetgordon:
Hi again,

Apologies, but the solution is that these examples should not be included at all. I misinterpreted the type hint for the function. `shape: Union[int, Tuple[int], List[int]]` implies that a tuple of ints _or_ a list of ints are accepted, not a tuple of shapes.

If you remove the 'tuple of shapes' examples and include an example with a tuple of integers and a list of integers as inputs respectively, then we'll be ready to merge.

Apologies again about the confusion, and thanks again for investigating the issue yourself.

Simone

mcandemir:
Hi,

No problem at all! I have added the list of ints (`List[int]`) and tuple of ints (`Tuple[int]`) as you asked.

I just had a little doubt adding the examples: since the function name is `shape_to_tuples`, I thought the inputs might be tuple (or list) of ints in a shape representation (such as: (2, 3) or [2, 3]). I added them as normal tuple of ints and list of ints (such as: (1, 2, 3) and [1, 2, 3]). So if there is a mistake I can change them anytime.

 Thanks for the feedbacks!

PR #1509, opened 2022-06-18
Reformat function `shape`

ericcherny:


ericcherny:
#1476

mattbarrett98 (changes requested):
The parts of the return type hints where you have put `List[int]` I think should be `Tuple[int]` as shape returns tuples. 

Also in the examples, maybe a clearer name than `y_tuple` would be `x_shape` 🙂

ericcherny:
Sure, totally missed that. Will update right now

ericcherny:
Just updated! Fixed some typos as well.

ericcherny:
I will fix that now! Also for clarification, I have two questions if that's alright:

1. Is there a way to use ivy as my framework? For instance, right now we have `as_array` parameter in ivy/general.py, but parameter `as_tensor` in the backend. So with whichever framework I set (e.g., `ivy.set_framework('torch')`), it will still use `as_tensor` when I call `ivy.shape(...)`. When would `as_array` be used?
2. You said all of my docstring examples must work, but the shape function doesn't seem to work with Containers (`ivy.shape(ivy.Container(a=[0], b=[0, 1]))` -> AttributeError: 'property' object has no attribute '__code__') and as an instance method (`ivy.array([1]).shape()` -> TypeError: 'tuple' object is not callable). Does that mean that Container and instance method functionality doesn't belong on this function's docstring (because they don't run)? Or are we assuming that some functionality will be implemented in the future? Or am I running the code wrong? 🤣

Thanks!

ericcherny:
I'm getting an error where the `as_array` argument isn't recognized:
After setting up `import ivy` -> `ivy.set_framework('torch')`, I get this:
1. `ivy.shape(ivy.array([1]), as_tensor=True)` -> runs perfectly
2. `ivy.shape(ivy.array([1]), as_array=True)` -> TypeError: shape() got an unexpected keyword argument 'as_array'

But when I do `from ivy.functional.ivy import general` -> `general.shape(ivy.array([1]), as_array=True)`, I get perfect output. What am I doing wrong?

mattbarrett98 (commented):
In the first bit of code you're using `set_framework` which no longer exists it's called `set_backend`, so maybe pulling the recent changes may help. For me changing this to `set_backend` and running your code, it works fine. 

In the next example, you should write `print(x.shape)` not `print(x.shape())` and it works fine.

Again the container example works fine for me, so there's something wrong on your end.

PR #1511, opened 2022-06-18
operator functions numeric support for numpy

juliagsy:
operator functions with numeric support added: 
- add, subtract, divide, floor_divide, multiply, remainder, pow
- positive, negative, abs
- less, less_equal, greater, greater_equal, equal, not_equal
- bitwise => and, or, xor, invert, left_shift, right_shift

Feel free to let me know if I left out any other operator functions! Thank you!!

### DIFF
--- a/ivy/functional/backends/numpy/elementwise.py
+++ b/ivy/functional/backends/numpy/elementwise.py
@@ -57,14 +57,19 @@ def pow(
 
 @_handle_0_dim_output
 def bitwise_xor(
-    x1: np.ndarray, x2: np.ndarray, *, out: Optional[np.ndarray] = None
-) -> np.ndarray:
+    x1: Union[float, np.ndarray],
+    x2: Union[float, np.ndarray],
+    *,
+    out: Optional[Union[float, np.ndarray]] = None
+) -> Union[float, np.ndarray]:
+    if not isinstance(x1, np.ndarray) and not isinstance(x2, np.ndarray):
+        return x1 ^ x2
+    if hasattr(x1, "dtype") and hasattr(x2, "dtype"):
+        return np.bitwise_xor(
+            x1, x2, dtype=np.promote_types(x1.dtype, x2.dtype), out=out
+        )
     if not isinstance(x2, np.ndarray):
         x2 = np.asarray(x2, dtype=x1.dtype)
-    else:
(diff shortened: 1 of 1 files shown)


juliagsy:
Changes done according to review in PR #1448 

The last PR is a bit messy, so I opened a new PR and update the support by backend/framework. Functions changed are stated in the description! Thanks!

mattbarrett98 (commented):
Hey, just thought I’d add that the out argument and the return type should just be np.ndarray. This is because numpy’s out argument only works with numpy arrays, and the decorator on these functions ensures the output is always a numpy array (never a float, bool etc)

juliagsy:
Hey @mattbarrett98 thanks for pointing that out! I added that because of some clarifications mentioned in this discussion (https://github.com/unifyai/ivy/discussions/1320)

If that's a issue, I'll make a fix! Thank you!

mattbarrett98:
The `out` argument definitely can't accept anything other than np.ndarrays so you can remove this. Maybe we can wait for confirmation from Dan about the returns being numbers, I think this discussion was before I added the decorator which calls `np.asarray` on the return if it isn't an `np.ndarray` already.

djl11 (changes requested):
Yes, given that we've recently decided to return arrays in cases where numbers are passed in the input, then perhaps parts of this PR are incorrect. Could you remove any changes which cause numbers to be returned, and keep any changes you think are still relevant even with this updated decision of desired behaviour, and then I'll review again. Otherwise, feel free to close if you deem that all of these changes are no longer relevant.

mattbarrett98 (changes requested):
Hey I just wrote a function `_cast_for_binary_op`, which implements the logic we want, could you add this to the relevant functions (I've done it for `add`). This way we aren't repeating the same logic in many functions. 

Also, I think the bitwise functions can take both `int` and `bool`.

juliagsy:
@mattbarrett98 Will update them now! And you're right, almost missed out bool, thanks a lot!

mattbarrett98 (commented):
Just noticed you forgot to edit `atan2` (it's near the bottom of the file). After this change though I think I'm happy and we'll just wait for Dan's confirmation.

djl11 (approved):
lgtm!

PR #1513, opened 2022-06-18
added nestable decorator to functions

Aarsh2001:
Added decorator to all  functions as instructed, I don't think it triggers any unwanted behaviour. Awaiting review  @mattbarrett98 , thanks 🙂


mattbarrett98 (changes requested):
Hi mate, looks good. Could you just remove the decorator from functions which don’t take any arguments, eg get_all_arrays_in_memory, num_cpu_cores etc. 

Basically just any function where it doesn’t make sense for a container to be an input shouldn’t have it (even if it was being mistakenly added previously). Thanks🙂

Aarsh2001:
Sure, will do.

PR #1543, opened 2022-06-20
Fixing submodule image

rush2406:


VedPatwardhan (changes requested):
Hi @rush2406, I'd suggest a few changes in the submodule:
1. Please add the `out` argument to the functions as we need to enable users to store the result of the function in some space they've allocated too.
2. Please update the docstrings of all functions with examples according to the format specified [here](https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html).
3. The `return` statements of few functions seem to be preceded by a blank line after the function signature, please remove it.
4. If possible, please check if you can get the tests passing for `torch` with `float16`, if not please add a comment specifying the reason for the same in the test function above that `if` block.
Feel free to ask any questions. Thanks 🙂

djl11 (changes requested):
Thanks @VedPatwardhan, very helpful feedback. Small point, it's not necessary to add all the examples to the docstrings in this PR, but you can do if you want @rush2406 . We are keeping this "docstring example" task open for others in the community to engage with, so no need to exhaustively add these examples in this PR. Thanks!

rush2406:
Hi @VedPatwardhan 
1. out argument may not be suitable for most of the methods here, because the output array shape is different from the input array shape
2. I have added a few examples for each function just for a sanity check
3. I wasn't able to resolve **test_random_crop** which was giving some errors related to the shape of array, would appreciate if you could provide any hints for it

VedPatwardhan (changes requested):
Hi @rush2406, thanks for making the changes!
1. About the `out` argument, a general use-case for this submodule would be when the user has pre-allocated space for storing the results of image stacking, resampling, etc. And you're right, the output shape could be different from the input shape, but with the help of the [`inplace_update`](https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/norms.py#L153) function, we don't need to worry about it.
2. Great, thanks!
3. I'll take a closer look at this function soon, in the meantime, please share any error message which you're getting which would be really helpful.

Thanks 🙂

rush2406:
@VedPatwardhan I tried appending as well. Initially, it gave the following error
```
>       flat_batch_size = functools.reduce(operator.mul, batch_shape, 1)
E       TypeError: reduce() arg 2 must support iteration

../ivy/functional/ivy/image.py:376: TypeError
```

I tried to fix it by converting batch_shape to a list. After this change, it started showing multiple hypothesis failures -
```
E   hypothesis.errors.MultipleFailures: Hypothesis found 2 distinct failures.

test_ivy/test_functional/test_core/test_image.py:231: MultipleFailures
---------------------------------- Hypothesis ----------------------------------
Falsifying example: test_random_crop(
    shape=[2, 2, 2],
    seed=1,
    input_dtype='bfloat16',
    as_variable=[False, False],
    num_positional_args=3,
    native_array=[False, False],
    container=[False, False],
    fw='jax',
)
```

VedPatwardhan:
Yeah, so firstly, I think this issue is also related to the `bfloat16` dtype which shouldn't be sampled, to prevent that, you could just replace `ivy.valid_float_dtypes` by `ivy_np.valid_float_dtypes`, which would make sure that 'bfloat16' is never sampled as dtype. I think you'll be able to see the error message more clearly after that rather than the `2 distinct failures` message. But I think that appending `[3]` instead of prepending it should solve the problem because the previous error message could be related to a change pushed recently w.r.t. private updates but If that still doesn't work, let me know. Thanks 🙂

rush2406:
@VedPatwardhan I changed to ivy_np as well, but the same error appears. I think it's due to some issue with the implementation itself rather than the unit test.

VedPatwardhan:
Oh okay, no worries, I'll look into it tomorrow. Apologies for the delay 🙂

VedPatwardhan:
I tried to test the function, then I narrowed down to specific examples to see what's failing, there's this specific failure I found `E     TypeError: Shapes must be 1D sequences of concrete values of integer type, got [(3,), 3, 6, 3].`. Following is the example that causes this failure:
```
x.shape (3, 3, 6, 3)
crop_size [2, 4]
as_variable [False, False]
native_array [False, False]
container [False, False]
seed 1
input_dtype float16
num_positional_args 0
```
I think you could try and resolve this failure first, which would then enable you to see other failures in a better description than `Hypothesis found 2 distinct failures`. The way to resolve this error should be to see how shape is processed inside the ivy function on this line `x_flat = ivy.reshape(x, [flat_batch_size] + image_dims + [num_channels])`.

I think there would be other issues you might run into as well, feel free to ask them as you come across them.
Thanks 🙂

rush2406:
I tried to fix that - 
`hypothesis.errors.Flaky: Hypothesis test_random_crop(shape=[2, 2, 5], seed=1, input_dtype='float16', as_variable=[False, False], num_positional_args=3, native_array=[False, False], container=[False, False], fw='jax') produces unreliable results: Falsified on the first call`. 

I think there was a comment about this flaky behaviour earlier

VedPatwardhan:
> I tried to fix that - `hypothesis.errors.Flaky: Hypothesis test_random_crop(shape=[2, 2, 5], seed=1, input_dtype='float16', as_variable=[False, False], num_positional_args=3, native_array=[False, False], container=[False, False], fw='jax') produces unreliable results: Falsified on the first call`.
> 
> I think there was a comment about this flaky behaviour earlier

Oh okay, let me check 👍

VedPatwardhan:
Even that example failure is related to the same line of code `x_flat = ivy.reshape(x, [flat_batch_size] + image_dims + [num_channels])`

VedPatwardhan:
@rush2406 you're right about that, but in case of `random_crop`, even seed is an optional argument, so the sequence need not be changed. 
* Further, to test if the function even works if the optional parameter values are `None`, I'd suggest you also sample `None` from hypothesis for those parameters. 
* The way to do that is basically instead of just `st.integers()` we'd have `st.none() | st.integers()` which would sample an integer sometimes and `None` at other times. 
* For doing this, you might have to add a parameter inside `given` as `data=st.data()` and then inside the function you could sample something like `seed=data.draw(st.none() | st.integers(min_value=1, max_value=8))`, similarly for the other optional parameters.

My apologies in advance if this takes some more effort but I think we also need to test if the function properly infers optional arguments like `batch_size` and `image_size` from the input if they aren't provided.

rush2406:
>     * The way to do that is basically instead of just `st.integers()` we'd have `st.none() | st.integers()` which would sample an integer sometimes and `None` at other times.
> 
>     * For doing this, you might have to add a parameter inside `given` as `data=st.data()` and then inside the function you could sample something like `seed=data.draw(st.none() | st.integers(min_value=1, max_value=8))`, similarly for the other optional parameters.

I tried to do this for seed, 
```
test_ivy/helpers.py:374: AssertionError
---------------------------------- Hypothesis ----------------------------------
Falsifying example: test_random_crop(
    shape=[2, 2, 5],
    data=data(...),
    input_dtype='float16',
    as_variable=[False, False],
    num_positional_args=0,
    native_array=[False, False],
    container=[False, True],
    fw='numpy',
)
Draw 1: None
_ test_random_crop[cpu-ivy.functional.backends.torch-True-False-torch_call-torch] _
```
It refers to ```          assert np.allclose(
                np.nan_to_num(x), np.nan_to_num(y), rtol=rtol, atol=atol
            ), "{} != {}".format(x, y)```

Any idea what this implies?

VedPatwardhan:
> > ```
> > * The way to do that is basically instead of just `st.integers()` we'd have `st.none() | st.integers()` which would sample an integer sometimes and `None` at other times.
> > 
> > * For doing this, you might have to add a parameter inside `given` as `data=st.data()` and then inside the function you could sample something like `seed=data.draw(st.none() | st.integers(min_value=1, max_value=8))`, similarly for the other optional parameters.
> > ```
> 
> I tried to do this for seed,
> 
> ```
> test_ivy/helpers.py:374: AssertionError
> ---------------------------------- Hypothesis ----------------------------------
> Falsifying example: test_random_crop(
>     shape=[2, 2, 5],
>     data=data(...),
>     input_dtype='float16',
>     as_variable=[False, False],
>     num_positional_args=0,
>     native_array=[False, False],
>     container=[False, True],
>     fw='numpy',
> )
> Draw 1: None
> _ test_random_crop[cpu-ivy.functional.backends.torch-True-False-torch_call-torch] _
> ```
> 
> It refers to ` assert np.allclose( np.nan_to_num(x), np.nan_to_num(y), rtol=rtol, atol=atol ), "{} != {}".format(x, y)`
> 
> Any idea what this implies?

Not sure, I'll take a look 🙂

VedPatwardhan:
I just saw your code, but it doesn't seem to sample `seed` currently. In case you haven't committed the code after making the change to `seed`, please do that so that I'd be able to have a look at that 🙂

djl11 (commented):
Looks great! However, we recently decided to remove this module from Ivy entirely, and instead move it to Ivy vision. Therefore I will close this issue for now. However, the changes were perfect based on the requests I made. Thanks for this!

PR #1561, opened 2022-06-20
numeric support for operator functions - jax

juliagsy:
operator functions with numeric support added: 
- add, subtract, divide, floor_divide, multiply, remainder, pow
- positive, negative, abs
- less, less_equal, greater, greater_equal, equal, not_equal
- bitwise => and, or, xor, invert, left_shift, right_shift

Feel free to let me know if I left out any other operator functions! Thank you!!


juliagsy:
bool is not supported in Jax by default so I did not include it for now (it is added in NumPy)! but if it is preferred for the related code to be included (or removed in NumPy 😂), feel free to let me know! Thanks!

djl11 (approved):
lgtm!

juliagsy:
@mattbarrett98 Got it! I think I misunderstood something from yesterday hahah, will remove that before merging! Thanks!

PR #1563, opened 2022-06-20
all optimizer steps nestable

Aarsh2001:
I've also added public instance method implementations in the ``` ContainerWithGradients ``` class, let me know if i should remove them? Thanks🙂


ivy/functional/ivy/gradients.py
```diff
@@ -240,20 +248,27 @@ def adam_step(dcdws, mw, vw, step, beta1=0.9, beta2=0.999, epsilon=1e-7):
...
 # Optimizer Updates #
 
 
 @to_native_arrays_and_back
-def optimizer_update(ws, effective_grads, lr, inplace=None, stop_gradients=True):
+@handle_nestable
+def optimizer_update(
+    ws: Union[ivy.Array, ivy.NativeArray],
+    effective_grads: Union[ivy.Array, ivy.NativeArray],
+    lr: Union[int, ivy.Array],
```
djl11, line 268: this should be `float` not `int`, and should also include `ivy.NativeArray`

ivy/functional/ivy/gradients.py
```diff
@@ -284,21 +299,26 @@ def optimizer_update(ws, effective_grads, lr, inplace=None, stop_gradients=True)
...
     return ws
 
 
 @to_native_arrays_and_back
-def gradient_descent_update(ws, dcdws, lr, inplace=None, stop_gradients=True):
+@handle_nestable
+def gradient_descent_update(
+    ws: Union[ivy.Array, ivy.NativeArray],
+    dcdws: Union[ivy.Array, ivy.NativeArray],
+    lr: Union[int, List[int]],
```
djl11, line 318: again, `float` not `int`, and are you sure about `List[int]`?

ivy/functional/ivy/gradients.py
```diff
@@ -434,21 +463,22 @@ def adam_update(
...
 @to_native_arrays_and_back
+@handle_nestable
 def lamb_update(
-    ws,
-    dcdws,
-    lr,
-    mw_tm1,
-    vw_tm1,
-    step,
+    ws: Union[ivy.Array, ivy.NativeArray],
```
djl11, line 468: `w` not `ws`, we assume singular and then handle multiple via nestable property.

ivy/functional/ivy/gradients.py
```diff
@@ -434,21 +463,22 @@ def adam_update(
...
+    dcdws: Union[ivy.Array, ivy.NativeArray],
```
djl11, line 469: `dcdw` not `dcdws`, we assume singular and then handle multiple via nestable property.

djl11 (changes requested):
Please see comments, I've generally added one comment per change, but the there are many instances and the comment could have been repeated multiple times. For each comment, please check each function for the same change requested. Thanks!

ivy/functional/ivy/gradients.py
Aarsh2001: No, my bad.

ivy/container/gradients.py
```diff
@@ -1,9 +1,320 @@
...
+            stop_gradients=stop_gradients,
+            key_chains=key_chains,
+            to_apply=to_apply,
+            prune_unapplied=prune_unapplied,
+            map_sequences=map_sequences,
+        )
+
+    def optimizer_update(
+        self: ivy.Container,
+        ws,
```
djl11, line 38: `ws` should be removed here, `self` is replacing `ws`, likewise for the other instance methods. These should actually be causing the tests to fail, but maybe they are not yet implemented. Could you please verify that the container instance methods are all working correctly before requesting the next review? :)

djl11 (changes requested):
see comments :)

ivy/container/gradients.py
Aarsh2001: totally missed this, yes will do. Thanks.

ivy/container/gradients.py
```diff
@@ -1,9 +1,310 @@
...
+            stop_gradients,
+            key_chains,
+            to_apply,
+            prune_unapplied,
+            map_sequences,
+        )
+
+    @staticmethod
+    def static_lamb_update(
+        ws,
```
djl11, line 234: these should not be plural either, `ws` should be `w` etc. These must match the names of the functional API in order for the nestable variant of the functions to work correctly. The arguments and keyword arguments are copied directly to this static method, and so the argument names must match exactly. The instance methods should also use the same non-plural argument names.

ivy/functional/ivy/gradients.py
```diff
@@ -210,15 +209,24 @@ def execute_with_gradients(func, xs, retain_grads=False):
...
+) -> ivy.Array:
     """Compute adam step delta, given the derivatives of some cost c with respect to ws,
     using ADAM update. `[reference]
 
     <https://en.wikipedia.org/wiki/Stochastic_gradient_descent#Adam>`_
 
     Parameters
     ----------
-    dcdws
+    dcdw(s)
```
djl11, line 229: we should just write `dcdw` not `dcdw(s)`, likewise for other args

ivy/functional/ivy/gradients.py
```diff
@@ -210,15 +209,24 @@ def execute_with_gradients(func, xs, retain_grads=False):
...
         Derivates of the cost c with respect to the weights ws, [dc/dw for w in ws].
```
djl11, line 230: this should be singular, for example "derivate of the cost c with respect to the weight w"

ivy/functional/ivy/gradients.py
```diff
@@ -240,26 +248,33 @@ def adam_step(dcdws, mw, vw, step, beta1=0.9, beta2=0.999, epsilon=1e-7):
...
 
 # Optimizer Updates #
 
 
 @to_native_arrays_and_back
-def optimizer_update(ws, effective_grads, lr, inplace=None, stop_gradients=True):
+@handle_nestable
+def optimizer_update(
+    w: Union[ivy.Array, ivy.NativeArray],
+    effective_grads: Union[ivy.Array, ivy.NativeArray],
```
djl11, line 267: singular not plural

ivy/functional/ivy/gradients.py
```diff
@@ -240,26 +248,33 @@ def adam_step(dcdws, mw, vw, step, beta1=0.9, beta2=0.999, epsilon=1e-7):
...
+    inplace=None,
+    stop_gradients=True,
+) -> ivy.Array:
     """Update weights ws of some function, given the true or effective derivatives of
     some cost c with respect to ws, [dc/dw for w in ws].
 
     Parameters
     ----------
-    ws
+    w(s)
```
djl11, line 277: `w` not `w(s)`

djl11 (changes requested):
see comments

djl11 (approved):
lgtm!

PR #1564, opened 2022-06-20
added docstrings to classes in layers

RickSanchezStoic:
added docstrings to classes layers


djl11 (approved):
These changes seem to all revolve around adding `dtype` arg to the module class, rather than adding docstrings to the layer classes. When squashing and merging, please use a commit message as appropriate. Thanks!

PR #1587, opened 2022-06-22
bump torch to `1.11.0` for functorch

bicycleman15:


yugaljain1999:
@bicycleman15 I just tried making PR on random_normal issue(#1657) but it is throwing error in DockerFileGPU because of incompatible `torch==1.11.0+cpu` not in versions list.
So can I update `torch==1.11.0` to `torch==1.11.0`  or you will do that to fix this docker issue?
Issue spot - https://github.com/unifyai/ivy/runs/7070821636?check_suite_focus=true

Thanks

PR #1592, opened 2022-06-22
fix functional_meta submodule

juliagsy:


juliagsy:
I'll work on the Hypothesis test once the type hints and some other criteria are verified as I'm a bit confused about some of the arguments!

Also, I have some questions about this:
1. if the batch type is added correctly and it's ivy.Container, from what I understand, it doesn't need the ivy.Array instance method, is that correct?
2. if so, does it need an ivy.Container instance method? I'm a bit confused for this and thanks in advance for clarifications!

Thank you!

Issue #1616, opened 2022-06-23
less_equal


jarplahmer:

jarplahmer:
#1616

1Doomdie1:
Hey, can you please provide the issue from where you took this function?
Thank you!

jarplahmer:
Reformat elementwise #761

PR #1617, opened 2022-06-23
Less_equal #1616

jarplahmer:
#1616

### DIFF
--- a/automation_tools/pr_automation/volunteer_go_to_intern.json
+++ b/automation_tools/pr_automation/volunteer_go_to_intern.json
@@ -1,42 +1,68 @@
 {
-  "HotzingTone": [
-    "ebyau",
-    "keting-ai"
-  ],
   "Ishticode": [
-    "Husain-Ayoob",
-    "sohil-hub"
-  ],
-  "bicycleman15": [
-    "fmilheir",
-    "bethelhall"
-  ],
-  "mattbarrett98": [
-    "DhruvaBhardwaj404",
-    "RickSanchezStoic"
-  ],
-  "saeedashrraf": [
-    "kirodev",
-    "uchennayah"
+    null
   ],
   "sherry30": [
-    "RitujaPawas",
-    "sreegandh804"
+    null
   ],
-  "vedpatwardhan": [
-    "zaeemansari70",
-    "Aarsh2001"
+  "bicycleman15": [
+    null
+  ],
+  "MarShaikh": [
+    null
   ],
   "RashulChutani": [
-    "arv0804"
+    null
   ],
   "Darshan-H-E": [
-    "AnimeshGurjar"
+    null
   ],
   "simonetgordon": [
     null
   ],
+  "RickSanchezStoic": [
+    null
+  ],
+  "RitujaPawas": [
+    null
+  ],
+  "Aarsh2001": [
+    null
+  ],
   "chie2727": [
+    null
+  ],
+  "sreegandh804z": [
+    null
+  ],
+  "abdrahmandiab": [
+    null
+  ],
+  "JIAHAN-XIE": [
+    null
+  ],
+  "juliagsy": [
+    null
+  ],
+  "rush2406": [
+    null
+  ],
+  "jkeane508": [
+    null
+  ],
+  "Auxia": [
+    null
+  ],
+  "ZoeCD": [
+    null
+  ],
+  "Oxbeefbabe": [
+    null
+  ],
+  "modanesh": [
+    null
+  ],
+  "MrZilinXiao": [
     null
   ]
 }


ivy/functional/ivy/elementwise.py
```diff
@@ -1089,6 +1089,146 @@ def less_equal(
...
+    >>> print(z)
+    ivy.array([ False, True, True])
+
+    With :code:`ivy.Container` instances:
+
+    >>> x = ivy.Container(a=ivy.array([4, 5, 6]),\
+                  b=ivy.array([2, 3, 4]))
+    >>> y = ivy.Container(a=ivy.array([1, 2, 3]),\
+                      b=ivy.array([5, 6, 7]))
+    >>> z = x <= y
```
Ishticode, line 1210: operator example should be under `__le__` in array class

ivy/functional/ivy/elementwise.py
```diff
@@ -1089,6 +1089,146 @@ def less_equal(
...
+        a: ivy.array([False, False, False]),
+        b: ivy.array([True, True, True])
+    }
+
+    With mix of :code:`ivy.Array` and :code:`ivy.Container` instances:
+
+    >>> x = ivy.array([[5.1, 2.3, -3.6]])
+    >>> y = ivy.Container(a=ivy.array([[4.], [5.], [6.]]),\
+                          b=ivy.array([[5.], [6.], [7.]]))
+    >>> z = x <= y
```
Ishticode, line 1222: again, operator example should be under __le__ in array class.

Ishticode (changes requested):
Hi @jarplahmer. Thank you for the PR. Your examples are well done. As pointed out by the individual comments the `<=` operator examples should be placed under  the special method `__le__`  under Array class [here](https://github.com/unifyai/ivy/blob/9625eb4687017909c3bcefc668ed8380957ec208/ivy/array/__init__.py#L448) just the way examples for Array operator `+` are added under `__add__` method [here](https://github.com/unifyai/ivy/blob/9625eb4687017909c3bcefc668ed8380957ec208/ivy/array/__init__.py#L448). To read more on this checkout the [docstring examples page](https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html). I Hope that makes sense and please let me know if something is unclear.

ivy/functional/ivy/elementwise.py
jarplahmer: So to be clear the operator examples need to be placed under the special method in the array class? Do I need to then delete the operator examples in the elementwise.py class? Or should these operator examples exist in moth places? Thanks for your help!

Ishticode: only in one place. Under the class 😊

jarplahmer: Just committed those changes, let me know if I did this correctly!

ivy/array/__init__.py
```diff
@@ -368,6 +368,47 @@ def __lt__(self, other):
     @_native_wrapper
     def __le__(self, other):
         return ivy.less_equal(self._data, other)
+    """
+    Operator Examples
```
Ishticode, line 372: The top part of the docstring should follow thae same structure as __add__ method which u will find in the same file if u scroll above. lmk if this is unclear.

Ishticode (changes requested):
Thats great. pls check two comments...
Besides that would u be able to resolve the single conflict...
let me know how it goes for u... thanks

ivy/array/__init__.py
```diff
@@ -368,6 +368,47 @@ def __lt__(self, other):
     @_native_wrapper
     def __le__(self, other):
         return ivy.less_equal(self._data, other)
```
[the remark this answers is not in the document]
jarplahmer: Pretty sure now fixed both these!

Ishticode (changes requested):
Hi. In case u missed I wrote above asking if you are able to resolve the conflict... and could you aslo make sure the function is reformatted in the backends. After this, your PR can be merged hopefully. Thnx. Let me know for any questions...😊

jarplahmer:
def less_equal(x1: Union[float, JaxArray], x2: Union[float, JaxArray]) -> JaxArray:
    return jnp.less_equal(x1, x2)

jarplahmer:
Kind of confused at where in Back_ends I should be looking to update the function, I found the function in Backends/elementwise.py however everything seemed to checkout with the current less_equal function. If you could just point me in the right direction whether I'm looking in the wrong place or how exactly I'm supposed to update it that would help a lot Thank you!

Ishticode:
> Kind of confused at where in Back_ends I should be looking to update the function, I found the function in Backends/elementwise.py however everything seemed to checkout with the current less_equal function. If you could just point me in the right direction whether I'm looking in the wrong place or how exactly I'm supposed to update it that would help a lot Thank you!

Hi the ivy backends can be see [this link](https://github.com/unifyai/ivy/tree/master/ivy/functional/backends). You just need to find your function and see if it is formatted according to [this descript](https://lets-unify.ai/ivy/deep_dive/9_function_arguments.html#function-arguments). If not, change it accordingly else 

> Kind of confused at where in Back_ends I should be looking to update the function, I found the function in Backends/elementwise.py however everything seemed to checkout with the current less_equal function. If you could just point me in the right direction whether I'm looking in the wrong place or how exactly I'm supposed to update it that would help a lot Thank you!

Dw about it. I checked myself and its fine. I will need to fix some in the __init__ file as after u fixed the conflict it has lead to some duplications. Will run tests and merge it after that.

ivy/functional/ivy/elementwise.py
```diff
@@ -1751,6 +1751,203 @@ def greater_equal(
...
+    Instance Method Examples
+    ------------------------
+
+    Using :code:`ivy.Array` instance method:
+
+    >>> x = ivy.array([1, 2, 3])
+    >>> y = ivy.array([4, 5, 6])
+    >>> z = z = x.less_equal(y)
+    >>> print(z)
+    ivy.array([ False, True, Ture])
```
Ishticode, line 1865: True is misspelled to Ture in this line.

Ishticode (changes requested):
Could you get the test_docstring passing. There are few slight issues in the docstring due to which the `test_docstring` is failing. I have commented one where `True` is misspelled as `Ture` and the test picks things out like these. You can take a look by clicking on the details next to the test_docstring in the tests box below.

jarplahmer:
> Could you get the test_docstring passing. There are few slight issues in the docstring due to which the `test_docstring` is failing. As an example, I have commented one where `True` is misspelled as `Ture` and the test picks things out like these. You can take a look by clicking on the details next to the test_docstring in the tests box below. Pls let me know for any questions and when the docstring is fixed, could you ask for review. Thank you...

the test_docstring is now passing, ready for the review :)

jarplahmer:
Still waiting to get this reviewed!!

PR #1623, opened 2022-06-24
fix functional losses hypothesis test

juliagsy:
in progress...


juliagsy:
- fixed arguments hypothesis
- added tests for optional arguments
- included array and container instance methods for the related composite functions (static methods if related)

djl11 (approved):
lgtm! in general we have been trying to use `st.composite` functions in the `@given` decorator for generating samples, rather than using `data=st.data()` in `@given` and then using `data.draw` in the unit test body, and so maybe later we can try and refactor this into an `st.composite` helper function instead, but for now this is totally fine.

PR #1648, opened 2022-06-26
Changed behavior in tests regarding randint()

donaldchan10:
Changes that I made:

- In `ivy/functional/ivy/random.py`, out is now passed into `current_backend().randint()` call.

- In `ivy_tests/test_ivy/test_functional/test_core/test_random.py::test_randint()`, I noticed that it handled the case where `dtype == "uint64"`, but not `dtype == "uint32"`. I changed the if conditional to check if `dtype == "uint32"`. I feel like `dtype == "uint32"` should be handled in the same way as `dtype == "uint64"` because uint32's max value cannot be represented as a float, like uint64.

- In `ivy_tests/test_ivy/test_functional/test_core/test_random.py::test_randint()`, out is not taken in as an argument, which needs to be specified even if it is an optional argument. I added it as an argument and passed it into all `randint()` calls.

Issue: #1644


VedPatwardhan (changes requested):
Hi @donaldchan10, thanks for contributing and thanks @chie2727 for requesting the review:
1. You don't need to update the `test_random.py` file as getting the tests passing is not a part of the Function Formatting task. In your case, the function `randint` is failing the tests on master already, so provided you make the changes to the functions correctly, we can merge your PR. You can understand more about the task [here](https://lets-unify.ai/ivy/contributing/4_open_tasks.html#open-tasks)
2. Please verify if the backend implementations of the `randint` function are added properly, you can refer the `ivy/functional/backends/<jax | numpy | tensorflow | torch>/random.py` file as a part of the function formatting task.
3. This function wouldn't have instance method implementations but you'll need to add a Container static method for this function in the `ivy/container/random.py` file. You can refer the `random_uniform` function implementation there for the same.
4. There are some updates to be made to the docstrings, we'll discuss about that after 1, 2 and 3 are done.
5. Great that you tried to make changes with the `out` argument. But the way we implement the `out` argument support in our functions is using the `ivy.inplace_update` function like [this](https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/norms.py#L153). 

Please feel free to ask any questions. Thanks 🙂

donaldchan10:
I tried to implement points 3 and 5. I'm not sure how to verify whether the backend implementations are added properly. From a first glance, they seem to be ok.

ivy/container/random.py
```diff
@@ -58,3 +58,54 @@ def random_uniform(
             map_sequences,
             out=out,
         )
+
+
+    # randint
+    @staticmethod
+    def static_random_int(
```
VedPatwardhan, line 65: The name of the function should be `static_randint`

ivy/container/random.py
```diff
@@ -58,3 +58,54 @@ def random_uniform(
...
+        device: Optional[Union[ivy.Device, ivy.NativeDevice, ivy.Container]] = None,
+        key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
+        to_apply: bool = True,
+        prune_unapplied: bool = False,
+        map_sequences: bool = False,
+        *,
+        out: Optional[ivy.Container] = None,
+    ) -> ivy.Container:
+        return ContainerBase.multi_map_in_static_method(
+            "random_uniform",
```
VedPatwardhan, line 78: The function called should be `"randint"` instead of `"random_uniform"`

ivy_tests/test_ivy/test_functional/test_core/test_random.py
```diff
@@ -121,7 +121,8 @@ def test_multinomial(data, num_samples, replace, dtype, tensor_fn, device, call)
     dtype=st.sampled_from(ivy_np.valid_int_dtypes),
     as_variable=st.booleans(),
 )
-def test_randint(data, shape, dtype, as_variable, device, call):
+def test_randint(data, shape, dtype, as_variable, device, out,
```
VedPatwardhan, line 124: Please revert this change

VedPatwardhan (changes requested):
Hi @donaldchan10,
1. Seems like `test_random.py` still has some edits in it as I can see in the Files Changed tab, please revert them so that the `test_random.py` files doesn't show up in the Files Changed tab.
2. Yeah, even I think the backend implementations are correct for this function 👍
3. I have made some suggestions about this near the lines of code concerned, please have a look.
4. We'll discuss after the above changes are done.
5. I have made suggestions near the lines of code concerned, please have a look.

ivy/container/random.py
```diff
@@ -58,3 +58,47 @@ def random_uniform(
...
+            "randint",
+            low,
+            high,
+            shape,
+            device=device,
+            key_chains=key_chains,
+            out=out,
+        )
+
+    def randint(
```
VedPatwardhan, line 83: Please remove this function as we don't need it.

ivy/container/random.py
```diff
@@ -58,3 +58,47 @@ def random_uniform(
...
+    @staticmethod
+    def static_randint(
+        low: Union[int, ivy.Container] = 0.0,
+        high: Union[int, ivy.Container] = 1.0,
+        shape: Optional[Union[int, Tuple[int, ...], ivy.Container]] = None,
+        device: Optional[Union[ivy.Device, ivy.NativeDevice, ivy.Container]] = None,
+        key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
+        out: Optional[ivy.Container] = None,
+    ) -> ivy.Container:
+        return ContainerBase.multi_map_in_static_method(
```
VedPatwardhan, line 73: Please update the docstring and add the static method examples here, you can refer [this](https://github.com/unifyai/ivy/blob/master/ivy/container/elementwise.py#L82) docstring for the same.

VedPatwardhan (changes requested):

ivy/container/random.py
```diff
@@ -58,3 +58,57 @@ def random_uniform(
...
+        """
+        ivy.Container static method variant of ivy.randint. This method simply wraps the
+        function, and so the docstring for ivy.randint also applies to this method
+        with minimal changes.
+
+        Examples
+        --------
+        With no device argument and no out argument specified and int arguemnt for shape:
+
+        >>> x = ivy.randint(low=1, high=10, shape=2)
```
VedPatwardhan, line 82: In examples for `static_randint`, the examples should be of the form `ivy.Container.static_randint`, where `low` and `high` passed to it are also `ivy.Container`, something of this sort,
`ivy.Container.static_randint(low=ivy.Container(a=2, b=4), high=ivy.Container(a=3, b=6), shape=(3,2))`
Please update the examples you've added to bring them to a similar form to this, basically pass `ivy.Container` to `low` and `high` parameters of the `ivy.Container.static_randint` method.

Rest looks good to me! Let me know if you have any questions. Thanks 🙂

ivy/container/random.py
```diff
@@ -58,3 +58,57 @@ def random_uniform(
...
+        """
+        ivy.Container static method variant of ivy.randint. This method simply wraps the
+        function, and so the docstring for ivy.randint also applies to this method
+        with minimal changes.
+
+        Examples
+        --------
+        With no device argument and no out argument specified and int arguemnt for shape:
+
+        >>> x = ivy.Container.randint(low=ivy.Container(a=1), high=ivy.Container(b=10), shape=2)
```
VedPatwardhan, line 82: Hi @donaldchan10, so the main advantage with a`ivy.Container` is that it enables us to process multiple values at once, meaning `ivy.Container.randint(low=ivy.Container(a=1), high=ivy.Container(b=10), shape=2)` should be replaced by something like `ivy.Container.randint(low=ivy.Container(a=1, b=3), high=ivy.Container(a=11, b=10), shape=2)` for the examples, because otherwise it would generate a `KeyError`. Please make that update and also update the result of the functions by running the examples locally.

ivy/container/random.py
```diff
@@ -58,3 +58,57 @@ def random_uniform(
...
+        >>> print(x)
+        ivy.array([5,7])
+
+        With no device argument and no out argument specified and sequence for shape:
```
VedPatwardhan, line 86: Please also update these headers with headers like `With one ivy.Container input`, `With multiple ivy.Container inputs` similar to the ones mentioned in the guide [here](https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html)

VedPatwardhan (approved):
lgtm! Thanks for contributing!

PR #1653, opened 2022-06-27
Gradients submodule

jiahanxie353:
- Added out arguments
- Integrated hypothesis into the Ivy unit test
- Ivy unit test are passing
- Added type hints
- Verified native array handling is correct
- Nestable version is working


ivy/container/gradients.py
```diff
@@ -7,6 +7,63 @@
 
 # noinspection PyMissingConstructor
 class ContainerWithGradients(ContainerBase):
+    @staticmethod
+    def static_adam_step(
```
djl11, line 11: type hints needed

ivy/functional/ivy/gradients.py
```diff
@@ -180,7 +193,11 @@ def stop_gradient(x, preserve_type=True):
 
 
 @to_native_arrays_and_back
-def execute_with_gradients(func, xs, retain_grads=False):
+def execute_with_gradients(
+    func: Callable,
+    xs: Union[ivy.Array, ivy.NativeArray, ivy.Container],
```
djl11, line 198: we should not include `ivy.Container` in the type hints of any nestable functions.

ivy/functional/ivy/gradients.py
```diff
@@ -257,18 +275,19 @@ def adam_step(
...
+
 # Optimizer Updates #
 
 
 @to_native_arrays_and_back
 @handle_nestable
 def optimizer_update(
-    w: Union[ivy.Array, ivy.NativeArray],
-    effective_grad: Union[ivy.Array, ivy.NativeArray],
+    w: Union[ivy.Array, ivy.NativeArray, ivy.Container],
```
djl11, line 285: we should not include `ivy.Container` in the type hints of any nestable functions.

djl11 (changes requested):
I have duplicated comments, please apply these changes for all functions and methods, and then request another review :)

ivy/container/base.py
```diff
@@ -149,6 +149,7 @@ def __init__(
         )
         self._config = dict()
         self.inplace_update(dict_in, **self._config_in)
+        self._is_variable = ivy.is_variable(dict_in)
```
djl11, line 152: `dict_in` will never be a variable, so this check doesn't make sense. If we want `self._is_variable` to be an `ivy.Container` with bools at the leaves, then we should cast `dict_in` to an `ivy.Container` before calling this function, otherwise `ivy.is_variable` will not detect any `ivy.Container` instances in the input, and therefore will not be called in a nestable manner.

ivy/container/base.py
```diff
@@ -5038,3 +5039,20 @@ def max_depth(self):
         if not kcs:
             return 0
         return max([len(kc.split("/")) for kc in kcs])
+    
+    @property
+    def is_variable(self):
```
djl11, line 5044: these docstrings don't make sense, a container will never be a variable, but it might contain variables at the leaves.

ivy/container/base.py
```diff
@@ -5038,3 +5039,20 @@ def max_depth(self):
...
+
+        Returns
+        -------
+        ret
+            Boolean, true if the container is a trainable variable, false otherwise.
+        """
+        return self._is_variable
+
+# noinspection PyRedeclaration
+class Variable(ContainerBase):
```
djl11, line 5055: can you explain the purpose of this new class in the constructor?

ivy/container/gradients.py
```diff
@@ -7,18 +7,184 @@
...
+            func,
+            xs,
+            retain_grads,
+            key_chains,
+            to_apply,
+            prune_unapplied,
+            map_sequences
+        )
+    
+    def execute_with_gradients(
```
djl11, line 100: we should not include an instance method for `execute_with_gradients`, we should only include instance methods for functions whose **first** argument can be an array. This is a recent change, that will be reflected in the docs shortly.

ivy/container/base.py
jiahanxie353: Sorry I meant to write `dict_in[key]`. 

For example, say we create `x = ivy.Container({"w": torch.tensor([3.0],requires_grad=True)})`, so `is_variable(x)` should evaluate to `True` because the leaf is a torch tensor with `requires_grad=True`. Therefore, we can have `self._is_variable = ivy.is_variable(dict_in[key])` to extract the values out of the `dict_in`.

But I'm wondering what if we have a container such as `Container({"w1": torch.tensor([3.0],requires_grad=True)}, {"w2":np.array([2.0])})`, where we have mixed structures ,and numpy does not support auto differentiation.

Or we just do not consider a container to be a `Variable` at all?

djl11: there is no `key` variable passed into the constructor, and so adding `dict_in[key]` wouldn't work. If we want `self._is_variable` to be a container (which makes more sense), then we can just change this to `self._is_variable = ivy.is_variable(self)` at the end of the constructor.

djl11 (changes requested):

ivy/container/base.py
jiahanxie353: sorry for being unclear. The "key" isn't a key literal, I meant the keys for the `dict_in`, like `w1` and `w2` as I gave in the second example.
If we use `self._is_variable = ivy.is_variable(self)`, is there any method to deal with internally conflicting inputs such as `cont = ivy.Container({"w1": torch.tensor([3.0],requires_grad=True)}, {"w2":np.array([2.0])})`, where `is_variable(cont["w1"])=True` but `is_variable(cont["w2"])=False`. Will we perform broadcasts in this case?
I guess it's more of a design pattern to decide if a container can be treated as a variable as a whole that can support gradient computation, or if we would rather consider leaves separately.
Please correct me if I'm wrong

djl11: we should definitely consider leaves separately. Also, Ivy does not support mixed backend frameworks, and so `cont = ivy.Container({"w1": torch.tensor([3.0],requires_grad=True)}, {"w2":np.array([2.0])})` is something that we should not expect to deal with, but we should indeed support variables and non-variables at the leaves, such as `cont = ivy.Container({"w1": torch.tensor([3.0],requires_grad=True)}, {"w2":torch.tensor([2.0])})`

ivy/functional/ivy/gradients.py
```diff
@@ -470,15 +489,16 @@ def lamb_update(
...
-    beta1=0.9,
-    beta2=0.999,
-    epsilon=1e-7,
-    max_trust_ratio=10,
-    decay_lambda=0,
-    inplace=None,
-    stop_gradients=True,
-) -> ivy.Array:
+    step: Union[int, float],
+    beta1: Optional[float]=0.9,
```
djl11, line 493: `Optional` type hints should only be used when the default value is `None`

ivy/functional/ivy/linear_algebra.py
```diff
@@ -612,7 +614,7 @@ def matmul(
 @handle_nestable
 def matrix_norm(
     x: Union[ivy.Array, ivy.NativeArray],
-    ord: Optional[Union[int, float, Literal[inf, -inf, "fro", "nuc"]]] = "fro",
+    ord: Optional[Union[int, float, Literal['inf', '-inf', "fro", "nuc"]]] = "fro",
```
djl11, line 617: lint requires `"` and not `'`

ivy/functional/backends/torch/gradients.py
```diff
@@ -31,16 +30,25 @@ def execute_with_gradients(func, xs, retain_grads=False):
         y = func_ret
         rest = tuple()
     y = ivy.to_native(y)
-    x_grads_flat = list(
-        torch.autograd.grad(
-            [y],
-            [v for k, v in xs.to_iterator()],
+    if isinstance(y,ivy.Container):
```
djl11, line 33: why is this check only added for PyTorch and not TensorFlow or JAX, do these `execute_with_gradients ` implementations already support `ivy.Container` instance for the return `y`?

djl11 (changes requested):
see comments :)

ivy/functional/backends/torch/gradients.py
jiahanxie353: The Hypothesis tests only falsified on PyTorch so I missed to consider TF and JAX, and you are definitely right. I will double check TF and JAX

ivy/functional/ivy/gradients.py
```diff
@@ -86,7 +88,11 @@ def unset_with_grads():
 @to_native_arrays_and_back
 @handle_out_argument
 @handle_nestable
-def variable(x: Union[ivy.Array, ivy.NativeArray]) -> ivy.Variable:
+def variable(
+    x: Union[ivy.Array, ivy.NativeArray],
+    *,
+    out: Optional[ivy.Variable] = None
+) -> ivy.Variable:
```
djl11, line 95: we actually are not using the `ivy.Variable` class anymore, this will eventually be removed. This should instead by `ivy.Array`. We then simply have a property `is_variable` on this array class.

ivy_tests/test_ivy/test_functional/test_core/test_gradients.py
```diff
@@ -4,164 +4,196 @@
...
+    with_out,
+    native_array,
+    num_positional_args,
+    container,
+    instance_method,
+    fw
+):
+    dtype, x = dtype_and_x
+    x = np.asarray(x, dtype=dtype)
+    if x.shape == ():
```
djl11, line 42: we shouldn't check for certain shapes in the test body itself, we should instead avoid generating these shapes in the first place, by extending the hypothesis helper functions, or passing in different arguments.

ivy_tests/test_ivy/test_functional/test_core/test_gradients.py
```diff
@@ -4,164 +4,196 @@
...
+    native_array,
+    num_positional_args,
+    container,
+    instance_method,
+    fw
+):
+    dtype, x = dtype_and_x
+    x = np.asarray(x, dtype=dtype)
+    if fw == "torch":
+        x = torch.as_tensor(x,dtype=torch.float32)
```
djl11, line 115: should we hardcode `torch.float32` here?

ivy_tests/test_ivy/test_functional/test_core/test_gradients.py
```diff
@@ -4,164 +4,196 @@
...
+    if fw == "tensorflow":
+        x = tf.Variable(tf.convert_to_tensor(x,dtype=dtype))
```
djl11, line 117: could you write a comment in this test, explaining why we need to pass these backend-specific variables as inputs to `helpers.test_function`?

ivy_tests/test_ivy/test_functional/test_core/test_gradients.py
```diff
@@ -4,164 +4,196 @@
...
+        x=x,
+        preserve_type=preserve_type
+    )
 
 # execute_with_gradients
-@pytest.mark.parametrize(
-    "func_n_xs_n_ty_n_te_n_tg",
-    [
-        (
+@given(
```
djl11, line 168: this is okay for now, but once `helpers.test_function` supports self-consistency testing against an arbitrary backend framework, then this test should look similar to the other tests, with `torch` used as our reference backend framework. Until this is implemented, this hard-coded ground truth approach suffices.

ivy_tests/test_ivy/test_functional/test_core/test_gradients.py
```diff
@@ -202,40 +234,155 @@ def test_execute_with_gradients(
...
+        dcdw=dcdw,
+        lr=lr,
+        inplace=inplace,
+        stop_gradients=stop_gradients,
+    )
 
 # layerwise_gradient_descent_update
-@pytest.mark.parametrize(
-    "ws_n_grads_n_lr_n_wsnew",
+@given(
```
djl11, line 384: When using use hard-coded ground truth values (which is necessary for compositional functions), then we should use a unique name for the test, such as `test_<func_name>_ground_truth`, which should be added alongside the self-consistency test, which does not use any ground truth values.

ivy_tests/test_ivy/test_functional/test_core/test_gradients.py
```diff
@@ -272,44 +420,60 @@ def test_layerwise_gradient_descent_update(
...
+    lr,
+    decay_lambda,
+    inplace,
+    stop_gradients
+):
+    dtype, w = dtype_and_w
+    w = np.asarray(w,dtype=dtype)
+    dcdw = np.asarray(dcdw,dtype=dtype)
+    if fw == "torch":
+        w = torch.as_tensor(w)
```
djl11, line 454: why are we converting to `torch.tensor` instances here? I don't think this is necessary.

ivy_tests/test_ivy/test_functional/test_core/test_gradients.py
```diff
@@ -272,44 +420,60 @@ def test_layerwise_gradient_descent_update(
...
+        lr=lr,
+        decay_lambda=decay_lambda,
+        inplace=inplace,
+        stop_gradients=stop_gradients,
+    )
 
 # adam_update
-@pytest.mark.parametrize(
-    "ws_n_grads_n_lr_n_wsnew",
+@given(
```
djl11, line 475: When using use hard-coded ground truth values (which is necessary for compositional functions), then we should use a unique name for the test, such as `test_<func_name>_ground_truth`, which should be added alongside the self-consistency test, which does not use any ground truth values.

ivy_tests/test_ivy/test_functional/test_core/test_gradients.py
```diff
@@ -348,11 +516,78 @@ def test_adam_update(ws_n_grads_n_lr_n_wsnew, dtype, tensor_fn, device, call):
...
+        epsilon=epsilon,
+        inplace=inplace,
+        stop_gradients=stop_gradients
+    )
 
 
 # layerwise_adam_update
-@pytest.mark.parametrize(
-    "ws_n_grads_n_lr_n_wsnew",
+@given(
```
djl11, line 589: When using use hard-coded ground truth values (which is necessary for compositional functions), then we should use a unique name for the test, such as `test_<func_name>_ground_truth`, which should be added alongside the self-consistency test, which does not use any ground truth values.

ivy_tests/test_ivy/test_functional/test_core/test_gradients.py
```diff
@@ -395,8 +631,8 @@ def test_layerwise_adam_update(ws_n_grads_n_lr_n_wsnew, dtype, tensor_fn, device
 
 
 # lamb_update
-@pytest.mark.parametrize(
-    "ws_n_grads_n_lr_n_wsnew",
+@given(
```
djl11, line 634: When using use hard-coded ground truth values (which is necessary for compositional functions), then we should use a unique name for the test, such as `test_<func_name>_ground_truth`, which should be added alongside the self-consistency test, which does not use any ground truth values.

ivy_tests/test_ivy/test_functional/test_core/test_gradients.py
```diff
@@ -437,17 +674,91 @@ def test_lamb_update(ws_n_grads_n_lr_n_wsnew, dtype, tensor_fn, device, call):
...
+    stop_gradients
+):
+    w = np.asarray(w,dtype=dtype)
+    dcdw = np.asarray(dcdw,dtype=dtype)
+    lr = np.asarray(lr,dtype=dtype)
+    mw_tm1 = np.asarray(mw_tm1,dtype=dtype)
+    vw_tm1 = np.asarray(vw_tm1,dtype=dtype)
+    step = np.asarray(step,dtype=dtype)
+    if fw == "torch":
+        w = torch.as_tensor(w)
```
djl11, line 727: I don't think this should be needed

djl11 (changes requested):
please see comments :)

A general point, all _compositional_ functions should have **two** tests:

- `test_<func_name>`, which uses self-consistency testing against a reference backend framework, as usual
- `test_<func_name>_ground_truth`, which uses `@pytest.mark.parametrize` with a few ground truth examples

We also shouldn't need to cast the inputs to torch tensors for any of the tests, which I noticed a few times.

jiahanxie353:
> Sorry for the delay, could you please fix the conflicts? then I'll properly dive in :)

No worries! Conflicts fixed :)

PR #1663, opened 2022-06-27
Functional Creation Hypothesis Tests - Draft

jkeane508:
Hi,

I've wrote some hypothesis tests for `ivy_tests/test_ivy/test_functional/test_core/test_creation.py` but I'm not 100% confident working with hypothesis. If someone could look over and give pointer/advice on how to improve them and correct them.

Thanks!
James.


mattbarrett98 (commented):
Looks like you've realised the `out` argument shouldn't be added explicitly for these functions, can you request another review when you've finished removing these. Also, the tests aren't running in this PR, are they passing locally for you?

mattbarrett98:
Great, just one point- you seem to be skipping torch float16 cases in all the tests, but i imagine all these functions support float16 in torch, so it isn't necessary 🙂

PR #1665, opened 2022-06-27
Squeeze #1655

suhy-jang:
- [ ] #1655

### DIFF
--- a/ivy/functional/ivy/manipulation.py
+++ b/ivy/functional/ivy/manipulation.py
@@ -552,6 +552,23 @@ def squeeze(
     >>> print(ivy.squeeze(x).shape)
     (2, 2)
 
+    >>> x = ivy.array([[[[1, 2, 3]], [[4, 5, 6]]]])
+    >>> print(x.shape)
+    (1, 2, 1, 3)
+    >>> print(ivy.squeeze(x, axis=2).shape)
+    (1, 2, 3)
+
+    >>> x = ivy.array([[[0], [1], [2]]])
+    >>> print(x.shape)
+    (1, 3, 1)
+    >>> print(ivy.squeeze(x))
+    [0, 1, 2]
+    >>> print(ivy.squeeze(x, axis=0))
+    [[0], [1], [2]]
+    >>> print(ivy.squeeze(x, axis=2))
+    [[0, 1, 2]]
+    >>> print(ivy.squeeze(x, axis=(0, 2)))
+    [0, 1, 2]
     """
     return current_backend(x).squeeze(x, axis, out=out)
 
--- a/ivy/functional/ivy/manipulation.py
+++ b/ivy/functional/ivy/manipulation.py
@@ -609,7 +609,7 @@ def squeeze(
 
     >>> x = ivy.Container(a=ivy.array([10., 11., 12.]), \
                           b=ivy.array([13., 14., 15.]))
-    >>> y = x.repeat(2)
+    >>> y = x.squeeze(2)
     >>> print(y)
     {
         a: ivy.array([10., 11., 12.]),
--- a/ivy/functional/ivy/manipulation.py
+++ b/ivy/functional/ivy/manipulation.py
@@ -518,7 +518,7 @@ def roll(
 @handle_nestable
 def squeeze(
     x: Union[ivy.Array, ivy.NativeArray],
-    axis: Union[int, Tuple[int, ...]],
+    axis: Optional[Union[int, Tuple[int, ...]]] = None,
     *,
     out: Optional[Union[ivy.Array, ivy.NativeArray]] = None,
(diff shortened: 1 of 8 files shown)


ivy/functional/ivy/manipulation.py
```diff
@@ -549,7 +549,26 @@ def squeeze(
...
+    >>> x = ivy.array([[[[1, 2, 3]], [[4, 5, 6]]]])
+    >>> print(x.shape)
+    (1, 2, 1, 3)
+    >>> print(ivy.squeeze(x, axis=2).shape)
+    (1, 2, 3)
+
+    >>> x = ivy.array([[[0], [1], [2]]])
+    >>> print(x.shape)
+    (1, 3, 1)
+    >>> print(ivy.squeeze(x))
```
MarShaikh, line 564: Sorry for being nitpicky, but can you please add a single newline after every print output in the docstrings? This is because the way `ivy_test/test_docstrings.py` works is that it assumes a print statement output ends at a newline.

MarShaikh (changes requested):
The examples look great, but could you add them according to the [documentation](https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html)? Thanks!

suhy-jang:
I updated it to fit the docstring test format. 
Please review again @MarShaikh. Thank you.

MarShaikh (changes requested):
Hi, thanks for the updates. Could you include the examples within the functional, static, instance and other methods as well? They're missing from the docstrings. Like for example, the squeeze function has a definition in the `ivy/container/manipulation.py` as well as in the array. 
Can you also make sure that headings of each example is also exactly how they are in the [docs](https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html)? 
Thanks!

suhy-jang:
Thank you for your review.
I updated the docstring of the Squeeze function and modified the backend function.
@MarShaikh, please let me know if I need more work.

ivy/functional/ivy/manipulation.py
```diff
@@ -541,15 +541,103 @@ def squeeze(
...
+
+    >>> x = ivy.Container(a=ivy.array([0., 1., 2.]), \
+                          b=ivy.array([3., 4., 5.]))
+    >>> y = ivy.squeeze(x)
+    >>> print(y)
+    {
+        a: ivy.array([0., 1., 2.]),
+        b: ivy.array([3., 4., 5.])
+    }
+
```
MarShaikh, line 611: Hi, these examples look great, just that can you copy and paste them to ivy.Array.squeeze and ivy.Container.squeeze according to [array instance method and container instance method example](https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html#docstring-examples). Should be ready to merge then. Thanks.

MarShaikh (changes requested):

suhy-jang:
I copied the docstring example to ivy.Array.squeeze and ivy.Container.squeeze.
Please review again @MarShaikh. Thank you.

ivy/functional/ivy/manipulation.py
suhy-jang: Hi @MarShaikh. As you said, I have updated the instance methods of array and container. Since I do not have permission to merge, so could you please review and merge? Thank you in advance.

ivy/functional/ivy/manipulation.py
```diff
@@ -541,15 +541,103 @@ def squeeze(
...
+    Instance Method Examples
```
MarShaikh, line 612: Hi, sorry to be so nitpicky, could you just remove the instance method examples from here, since you've already added them to container and arrays, these are redundant. Thanks.

MarShaikh (changes requested):

suhy-jang:
I deleted the redundant examples from the main function, and please review them again @MarShaikh. 
Thank you so much for your attention and help.

MarShaikh (changes requested):
Could you please make sure the `test_docstrings.py` passes in the CI? I also noticed that you added the `out` argument in the backends. The out argument is only added if supported by the backend, which in the case of JAX, it isn't. Please see this for other backends as well. Thanks!

MarShaikh:
> Thank you for taking the time to review @MarShaikh. '`out` argument' was initially on my code, and it would have been more meaningful if included in the initial review. I updated it considering what you said, and now there is no problem with `test_docstrings.py` in the CI test related to `squeeze`. I am not familiar with this system yet, so if the problem of the failed test is still a problem in the following review, I would appreciate it if you could point out the exact cause and where it is. Thank you in advance, and please review it again.

Apologies for not pointing out the `out` argument earlier, in the .diff it seemed like something you added. Sorry the confusion again! Everything else looks great! Thanks for the contribution!

PR #1681, opened 2022-06-29
tolerance dict fix for test_frontend_function

RickSanchezStoic:
added tolerance dict fix for test_frontend_function


RickSanchezStoic:
tolerance dict fix for test_frontend_function

PR #1684, opened 2022-06-29
refactor losses tests

juliagsy:
- added range limits in helpers.dtype_and_values()
- refactored tests in losses submodule


djl11 (approved):
as long as the new arguments for `dtype_and_values ` doesn't cause other tests to fail, then the changes look good to me!

Issue #776, opened 2022-04-02
Reformat gradients 


saeedashrraf:
Each method in the [gradients](https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/gradients.py) submodule should be modified to use the latest formatting rules, as explained in the [contributor guide](https://lets-unify.ai/ivy/contributing.html) for the [Ivy API](https://lets-unify.ai/ivy/contributing.html#ivy-api) and [Backend API](https://lets-unify.ai/ivy/contributing.html#backend-api). This includes removing all lambdas and direct bindings for methods in the backend submodules, with each method instead defined using `def method_name`. Full type hints should also be given for each argument in both the ivy module and also each backend module. The docstrings should be updated to use NumPy style rather than Sphinx style, and code examples should be added to **every** method docstring.

- [ ] with_grads
- [ ] set_with_grads
- [ ] unset_with_grads
- [ ] variable
- [ ] is_variable
- [ ] variable_data
- [ ] stop_gradient
- [ ] execute_with_gradients
- [ ] adam_step
- [ ] optimizer_update 
- [ ] gradient_descent_update
- [ ] lars_update
- [ ] adam_update
- [ ] lamb_update

Joanes-Iri:
- [ ] #1220

hirwa-nshuti:
#1640

Psychofun:
#1723

Psychofun:
- [ ] #1725

sampreet-arthi:
- [ ] #1775

yugaljain1999:
#1833

Accumulated:
#2143

someshfengde:
#2234

yugaljain1999:
#2241

VikramxD:
#2317

VikramxD:
- [ ] #2317

hirwa-nshuti:
variable #2323

vandit98:
lars_update #2334

DrashtySoni:
- [ ] #2354

peacemakafui:
#2536

karimHemina:
#6605

dhakalaashish:
- [ ] #9175

waheeduddin:
- [ ] #9283

CaloloCosta:
- [ ] #9559

DavidPav123:
https://github.com/unifyai/ivy/issues/10385#issue-1575162882

[...]

PR #1687, opened 2022-06-29
reformat adam_step

### LINKED ISSUE #1640: adam_step

hirwa-nshuti:
Added functional examples to the adam_step function resolves #1640


hirwa-nshuti:
> The container instance methods can be found here: https://github.com/unifyai/ivy/blob/master/ivy/container/gradients.py#L161. I don't believe the array instance method exists yet so could you please add this (you should be able to basically copy the way it's done in other array instance methods)

As you can see here I think the issue I was working on is the method of adam_step()  as shown here https://github.com/unifyai/ivy/issues/776 and in the container instance, it is not there. The one you referenced is adam_update and I think I can work on it after this PR is merged so I can open a new Issue of the adam_update() method.

Thanks.

mattbarrett98 (changes requested):
Hi, yes you're right `adam_update` is another function my bad.

If you look under test_docstrings action it shows the list of functions/methods whose docstrings examples aren't working properly. It seems most of yours are currently listed in here, please could you fix the examples 🙂

hirwa-nshuti:
Hi, I worked on the changes you asked for and all the tests are passing now.
Thanks

hirwa-nshuti:
Hi, I hardly finally finished my examples and now the warnings am receiving are common to all other PRs I came across, I don't know if there is something I can do about it.

Thanks.

mattbarrett98 (changes requested):
Hey, in the type hints we don't include `ivy.Container` as you have in the functional `adam_step` (this is just for brevity). Also in your container example you have explicitly set the backend to torch which you shouldn't do, were you having issues with the other backends?

hirwa-nshuti:
Hello @mattbarrett98,

Actually, I had a slight problem using the backends but I found it was due to my mistakes in passing arguments to the functions. Here are the changes, I removed explicitly set backends and also removed the `ivy.Container` type hint in the functional `adam_step` method. But for the container method, I included them in there. 

Thanks

mattbarrett98 (commented):
Looks good, thank you!

PR #1696, opened 2022-06-29
Added examples to set_default_device

BKR42:


ivy/functional/ivy/device.py
```diff
@@ -664,6 +664,24 @@ def set_default_device(device: Union[ivy.Device, ivy.NativeDevice]):
     device
         The device to set as the default device
 
+    Examples
+    --------
+    >>> ivy.set_default_device("cpu")
+    >>> ivy.default_device()
+    'cpu'
+    
+    >>> import torch
```
VedPatwardhan, line 673: Please move this line to above the `ivy.set_backend("torch")` line as we don't need to `import torch` in this example but need it in the next example.

ivy/functional/ivy/device.py
```diff
@@ -664,6 +664,24 @@ def set_default_device(device: Union[ivy.Device, ivy.NativeDevice]):
...
+    >>> ivy.set_backend("torch")
+    >>> ivy.set_default_device("gpu:0")
+    >>> ivy.default_device(as_native=True)
+    device(type='cuda', index=0)
+
+    >>> ivy.set_backend("torch")
+    >>> device = torch.device("cuda")
+    >>> ivy.set_default_device(device)
+    >>> ivy.default_device(as_native=True)
+    device(type='cuda', index=0)
```
VedPatwardhan, line 683: This output should be `device(type='cuda')`

VedPatwardhan (changes requested):
Hi @BKR42, thanks for contributing, I have suggested a few changes near the lines of code concerned, please have a look and let me know if you have any questions. Thanks 🙂

VedPatwardhan (approved):
lgtm! Thanks for contributing!

PR #1700, opened 2022-06-30
conv1d #1510

TanishGupta15:


TanishGupta15:
> This seems incomplete, are you still working on adding examples?

There is one example left, which I am not sure how to run. I don't have much experience in using conv1d, can you please help with an example for the remaining case? (Only one example is left).
Thanks!

mattbarrett98 (commented):
I'm afraid I can't rally do the task for you, maybe take a look at some of the other docstring examples to gain some understanding.

PR #1702, opened 2022-06-30
Functional Creation Unit Tests

iamjameskeane:
Hi,

I got the test to work with the array function but its just throwing assertion errors now. Likely due to tolerances because they are all numbers very close to 0. Will continue to work on unless you have any further input :)

Thanks,
James

### DIFF
--- a/ivy_tests/test_ivy/test_functional/test_core/test_creation.py
+++ b/ivy_tests/test_ivy/test_functional/test_core/test_creation.py
@@ -484,6 +484,7 @@ def test_full_like(
 
 # meshgrid
 
+
 # allows for arrays of all 1d and same dtype
 array_shape = st.shared(
     st.lists(st.integers(min_value=1, max_value=5), min_size=1, max_size=1)
@@ -496,14 +497,21 @@ dtype_shared = st.shared(st.sampled_from(ivy_np.valid_numeric_dtypes))
         hnp.arrays(dtype=dtype_shared, shape=array_shape), min_size=1, max_size=3
     ),
     dtype=dtype_shared,
-    num_positional_args=helpers.num_positional_args(fn_name="meshgrid"),
 )
 def test_meshgrid(
     arrays,
     dtype,
-    num_positional_args,
     fw,
 ):
+
+    kw = {}
+    i = 0
+    for x_ in arrays:
+        kw["x{}".format(i)] = np.asarray(x_, dtype=dtype)
+        i += 1
+
+    num_positional_args = len(arrays)
+
     helpers.test_function(
         dtype,
         False,
@@ -518,7 +526,7 @@ def test_meshgrid(
         1e-06,
         True,
         "numpy",
-        arrays=arrays,
+        **kw,
     )
 
 


ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -2,159 +2,358 @@
...
+        "asarray",
+        object_in = np.asarray(x),
+    )
+
+
 # empty()
+@given(
+    array_shape = st.integers(0,5) | st.lists(st.integers(min_value = 0, max_value=10), min_size = 1, max_size = 5),
+    dtype = st.sampled_from(ivy_np.valid_int_dtypes),
+    as_variable=st.booleans(),
```
djl11, line 227: `as_variable` is not relevant here as the input contains no arrays. We should remove from the decorator and hard-code it as `False` in the inputs to `helpers.test_array_function`

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -2,159 +2,358 @@
...
+    with_out=st.booleans(),
+    num_positional_args=helpers.num_positional_args(fn_name="empty"),
+    native_array=st.booleans(),
```
djl11, line 230: `native_array` is not relevant here as the input contains no arrays. We should remove from the decorator and hard-code it as `False` in the inputs to `helpers.test_array_function`

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -2,159 +2,358 @@
...
+    container=st.booleans(),
+    instance_method=st.booleans(),
```
djl11, line 232: `instance_method` is not relevant here as the input contains no arrays. We should remove from the decorator and hard-code it as `False` in the inputs to `helpers.test_array_function`

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -2,159 +2,358 @@
...
+        num_positional_args,
+        native_array,
+        container,
+        instance_method,
+        fw,
+        "empty",
+        rtol,
+        atol,
+        test_values,
+        shape=array_shape
```
djl11, line 269: you should also be providing the keyword arguments `dtype` and `device` here. `dtype` is generated inside the hypothesis `@given` call, and `device` is generated by the pytest wrapping code, and so you can simply add `device` as an argument to `test_empty` and the argument will be populated with a device string.

djl11 (changes requested):
I've only reviewed the test for `empty`, but I imagine these requested changes are reflected in other tests as well :)

jkeane508:
Hi,
quick update alot of the test are working, I was hoping you could give insight into the following,

ivy/functional/backends/torch/creation.py
```diff
@@ -66,14 +66,9 @@ def arange(
...
         dtype = as_native_dtype(default_dtype(dtype))
-        if dtype in [torch.int8, torch.uint8, torch.int16]:
-            return torch.arange(
-                start, stop, step=step, dtype=torch.int64, device=device, out=out
-            ).to(dtype)
-        else:
-            return torch.arange(
-                start, stop, step=step, dtype=dtype, device=device, out=out
-            )
+        return torch.arange(
```
djl11, line 69: are you sure this is okay? It doesn't cause Array API tests to fail? I presume it was there for a reason, but maybe it was a mistake or is no longer needed with the latest `torch` version.

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -1,157 +1,404 @@
...
 
 # local
 import ivy
+from ivy.functional.backends.numpy.data_type import dtype
 import ivy_tests.test_ivy.helpers as helpers
 import ivy.functional.backends.numpy as ivy_np
 
 
-# array
+# native_array - DONE
```
djl11, line 17: best to remove comments to yourself from the commit, such as `DONE` in this case.

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -1,157 +1,404 @@
...
+        max_value=None,
+        allow_inf=False,
+        exclude_min=False,
+        exclude_max=False,
+        min_num_dims=1,
+        max_num_dims=5,
+        min_dim_size=1,
+        max_dim_size=10,
+        shape=None,
+        shared_dtype=True,),
```
djl11, line 32: seems like you are just passing the default values for some of the arguments above. This is not necessary, just pass the arguments which are different from their default value.

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -1,157 +1,404 @@
...
+    dtype_and_x,
+    as_variable,
+    num_positional_args,
+    native_array,
+    container,
+    instance_method,
+    fw,
+    device
+):
+    dtype, x=dtype_and_x[0], dtype_and_x[1]
```
djl11, line 49: no need to index here, you can just write `dtype, x = dtype_and_x` and it will unpack automatically, but you should use spaces between `=`

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -1,157 +1,404 @@
...
+        max_value=None,
+        allow_inf=False,
+        exclude_min=False,
+        exclude_max=False,
+        min_num_dims=1,
+        max_num_dims=1,
+        min_dim_size=2,
+        max_dim_size=2,
+        shape=None,
+        shared_dtype=True,),
```
djl11, line 80: same comment as above

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -1,157 +1,404 @@
...
```
djl11, line 132: same comment as above

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -1,157 +1,404 @@
...
+    axis,
+    device,
+    with_out,
+    num_positional_args,
+    container,
+    fw,
+):  
+    test_rtol=1, #if its less then one it'll test for inf
+    test_atol=1e-06,
+    test_values = True
```
djl11, line 153: you can save on lines of code and add these directly as keyword arguments below

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -1,157 +1,404 @@
...
+
+
+# empty() - failing, asserts bizzare numbers
+@given(
+    shape=helpers.get_shape(
+        allow_none=False,
+        min_num_dims=1,
+        max_num_dims=5,
+        min_dim_size=1,
+        max_dim_size=10,),
```
djl11, line 263: same comment as above

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -1,157 +1,404 @@
...
+        max_value=None,
+        allow_inf=False,
+        exclude_min=False,
+        exclude_max=False,
+        min_num_dims=1,
+        max_num_dims=5,
+        min_dim_size=1,
+        max_dim_size=5,
+        shape=None,
+        shared_dtype=True,),
```
djl11, line 310: again

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -184,18 +431,129 @@ def test_from_dlpack(
...
-# full()
-# full_like()
+# full() - Not working with ivy.container
+@given(
+shape=helpers.get_shape(
+        allow_none=False,
+        min_num_dims=1,
+        max_num_dims=5,
+        min_dim_size=1,
+        max_dim_size=10,),
```
djl11, line 441: again

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -209,28 +567,76 @@ def test_meshgrid(
...
+        max_value=None,
+        allow_inf=False,
+        exclude_min=False,
+        exclude_max=False,
+        min_num_dims=1,
+        max_num_dims=5,
+        min_dim_size=1,
+        max_dim_size=5,
+        shape=None,
+        shared_dtype=True,),
```
djl11, line 639: again

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -260,24 +667,179 @@ def test_ones_like(
...
+        max_value=None,
+        allow_inf=False,
+        exclude_min=False,
+        exclude_max=False,
+        min_num_dims=2,
+        max_num_dims=5,
+        min_dim_size=1,
+        max_dim_size=10,
+        shape=None,
+        shared_dtype=True,),
```
djl11, line 690: again

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -260,24 +667,179 @@ def test_ones_like(
...
```
djl11, line 742: again

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -260,24 +667,179 @@ def test_ones_like(
...
+
+
+# zeros() - not working with container
+@given(
+shape=helpers.get_shape(
+        allow_none=False,
+        min_num_dims=1,
+        max_num_dims=5,
+        min_dim_size=1,
+        max_dim_size=10,),
```
djl11, line 786: again

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -260,24 +667,179 @@ def test_ones_like(
...
+        max_value=None,
+        allow_inf=False,
+        exclude_min=False,
+        exclude_max=False,
+        min_num_dims=2,
+        max_num_dims=5,
+        min_dim_size=1,
+        max_dim_size=10,
+        shape=None,
+        shared_dtype=True,),
```
djl11, line 832: again

djl11 (changes requested):
just made some simple formatting requests for now, will then dig deeper into the failing tests on the next review :)

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -184,53 +393,191 @@ def test_from_dlpack(
...
+    fill_value,
+    dtype,
+    with_out,
+    device,
+    num_positional_args,
+    container,
+    fw,
+):
+
+    if type(fill_value) == float and dtype in ivy_np.valid_int_dtypes:
```
djl11, line 422: we should omit this case during the data generation phase, rather than in the function body itself. You might need to create a short helper function for this.

djl11 (changes requested):
small change requested, once that is fixed, provided that the tests are passing, I think it's good to merge!

PR #1709, opened 2022-07-01
fix `utility` submodule

juliagsy:
- type hints and instance methods
- hypothesis tests 
- added a new helper function for generating axis according to array


ivy/array/utility.py
```diff
@@ -12,17 +12,17 @@ class ArrayWithUtility(abc.ABC):
     def all(
         self: ivy.Array,
         axis: Optional[Union[int, Tuple[int], List[int]]] = None,
-        keepdims: bool = False,
+        keepdims: Optional[bool] = False,
```
djl11, line 15: There might be some conflicting information in some areas in the deep dive section, and I'm sorry if that's the case, but actually I have recently learnt that `Optional` should only be used to flag that `None` is a permitted value for the argument, as explained in the answer by Tomasz Bartkowiak [here](https://stackoverflow.com/questions/38727520/how-do-i-add-default-parameters-to-functions-when-using-type-hinting). Therefore, we should only use `Optional` for arguments which have a default value of `None`.

djl11 (changes requested):
see comment :)

juliagsy:
fixed the optional type hints! If the changes are all good, I'll check out the master branch and fix the lint before merging, thanks!

ivy_tests/test_ivy/test_functional/test_core/test_utility.py
```diff
@@ -11,30 +11,31 @@
...
     instance_method,
     fw,
 ):
-    num_positional_args = data.draw(helpers.num_positional_args(fn_name="all"))
-    shape = data.draw(helpers.get_shape(min_num_dims=1))
-    x = data.draw(helpers.array_values(dtype=input_dtype, shape=shape))
-    axis = data.draw(helpers.get_axis(shape=shape))
+    input_dtype, x = dtype_and_x
+    if not isinstance(x, list):
+        x = [x]
```
djl11, line 38: does this function being tested not support 0-dimensional arrays? If it does, then we should not perform this conversion. If it does not, then we should try to handle this in the `helpers.dtype_and_values` helper instead, perhaps refactoring it a bit in order to support the ability to avoid generating numbers.

ivy_tests/test_ivy/test_functional/test_core/test_utility.py
```diff
@@ -53,30 +54,31 @@ def test_all(
...
     instance_method,
     fw,
 ):
-    num_positional_args = data.draw(helpers.num_positional_args(fn_name="any"))
-    shape = data.draw(helpers.get_shape(min_num_dims=1))
-    x = data.draw(helpers.array_values(dtype=input_dtype, shape=shape))
-    axis = data.draw(helpers.get_axis(shape=shape))
+    input_dtype, x = dtype_and_x
+    if not isinstance(x, list):
+        x = [x]
```
djl11, line 81: does this function being tested not support 0-dimensional arrays? If it does, then we should not perform this conversion. If it does not, then we should try to handle this in the `helpers.dtype_and_values` helper instead, perhaps refactoring it a bit in order to support the ability to avoid generating numbers.

djl11 (changes requested):
see comments, after these I think it's good to go!

juliagsy:
updated `axis` to match data generated so that error is not thrown!

ivy/container/utility.py
```diff
@@ -10,44 +10,92 @@
 
 # noinspection PyMissingConstructor
 class ContainerWithUtility(ContainerBase):
+    @staticmethod
+    def static_all(
+        x: Union[ivy.Array, ivy.NativeArray, ivy.Container],
+        axis: Optional[Union[int, Tuple[int], List[int], ivy.Container]] = None,
```
djl11, line 16: we can replace `Tuple[int], List[int]` with `Sequence[int]` to be more concise :)

ivy/container/utility.py
```diff
@@ -10,44 +10,92 @@
...
+        keepdims: Union[bool, ivy.Container] = False,
+        key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
```
djl11, line 18: `Sequence[str]` rather than `List[str]`

djl11 (changes requested):
A few small changes, but otherwise looks great! In general, we should always support both `tuple` and `list` instances in the inputs, in order to be more robust (we may then need to explicitly cast to one or the other in the function body). Therefore, we should use the type hint `Sequence[type]` in general (this is actually a bit more general than `Union[List, Tuple]`, and is therefore technically incorrect, but it makes our type hints more short and readable, and generally suffices.)

ivy/container/utility.py
juliagsy: I see! Will take note of this and make the changes! Thanks!

ivy/container/utility.py
```diff
@@ -10,44 +10,92 @@
...
-        axis: Optional[Union[int, Tuple[int], List[int]]] = None,
-        keepdims: bool = False,
-        key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
+        axis: Optional[Union[int, Sequence[int], ivy.Container]] = None,
+        keepdims: Union[bool, ivy.Container] = False,
+        key_chains: Optional[Union[Sequence[str], Dict[str, str]]] = None,
         to_apply: bool = True,
         prune_unapplied: bool = False,
         map_sequences: bool = False,
         out: Optional[ivy.Container] = None,
```
djl11, line 90: should have `*` above `out` here, because `out` should always be keyword-only, even for instance methods.

ivy/container/utility.py
```diff
@@ -10,44 +10,92 @@
...
-        axis: Optional[Union[int, Tuple[int], List[int]]] = None,
-        keepdims: bool = False,
-        key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
+        axis: Optional[Union[int, Sequence[int], ivy.Container]] = None,
+        keepdims: Union[bool, ivy.Container] = False,
+        key_chains: Optional[Union[Sequence[str], Dict[str, str]]] = None,
+        to_apply: bool = True,
+        prune_unapplied: bool = False,
+        map_sequences: bool = False,
+        out: Optional[ivy.Container] = None,
```
djl11, line 45: should have `*` above `out` here, because `out` should always be keyword-only, even for instance methods.

ivy_tests/test_ivy/test_functional/test_core/test_utility.py
```diff
@@ -11,30 +11,31 @@
...
     container,
     instance_method,
     fw,
 ):
-    num_positional_args = data.draw(helpers.num_positional_args(fn_name="all"))
-    shape = data.draw(helpers.get_shape(min_num_dims=1))
-    x = data.draw(helpers.array_values(dtype=input_dtype, shape=shape))
-    axis = data.draw(helpers.get_axis(shape=shape))
+    input_dtype, x = dtype_and_x
+    if not isinstance(x, list):
```
djl11, line 37: we should not need to do any checks in the function body such as this. It should be possible to make sure we always _generate_ `axis=None` whenever `isinstance(x, list)`. This kind of `st.composite` helper function will be widely useful and reusable, potentially in other parts of the project. This will also mean we avoid creating redundant variations in the parameter space. For example, if `isinstance(x, list)` in the current design, hypothesis will still sample the `axis` space, each with unique seeds, which will "waste" some of our parameter searching budget.

ivy_tests/test_ivy/test_functional/test_core/test_utility.py
```diff
@@ -53,30 +54,31 @@ def test_all(
...
     container,
     instance_method,
     fw,
 ):
-    num_positional_args = data.draw(helpers.num_positional_args(fn_name="any"))
-    shape = data.draw(helpers.get_shape(min_num_dims=1))
-    x = data.draw(helpers.array_values(dtype=input_dtype, shape=shape))
-    axis = data.draw(helpers.get_axis(shape=shape))
+    input_dtype, x = dtype_and_x
+    if not isinstance(x, list):
```
djl11, line 80: same comment as above.

djl11 (changes requested):

djl11 (approved):
lgtm!

PR #1719, opened 2022-07-01
fix `ivy.Container` type hints in `losses` container instances

juliagsy:
I've only updated the type hints slightly but I'm still not very familiar with Container so would be better for you to take a look! Thanks a lot!

### DIFF
--- a/ivy/container/losses.py
+++ b/ivy/container/losses.py
@@ -11,8 +11,8 @@ class ContainerWithLosses(ContainerBase):
     def static_cross_entropy(
         true: Union[ivy.Container, ivy.Array, ivy.NativeArray],
         pred: Union[ivy.Container, ivy.Array, ivy.NativeArray],
-        axis: Optional[int] = -1,
-        epsilon: Optional[float] = 1e-7,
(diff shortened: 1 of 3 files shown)


ivy/container/losses.py
```diff
@@ -11,8 +11,8 @@ class ContainerWithLosses(ContainerBase):
     def static_cross_entropy(
         true: Union[ivy.Container, ivy.Array, ivy.NativeArray],
         pred: Union[ivy.Container, ivy.Array, ivy.NativeArray],
-        axis: Optional[int] = -1,
-        epsilon: Optional[float] = 1e-7,
+        axis: Optional[Union[int, ivy.Container]] = -1,
```
djl11, line 14: these new changes are great, but again I've noticed as miss-use of the `Optional` type hint, which should only be used for arguments with default value of `None`.

djl11 (changes requested):

juliagsy:
fixed the optional type hints! If the changes are all good, I'll check out the master branch and fix the lint before merging, thanks!

djl11 (approved):
lgtm!

PR #1726, opened 2022-07-02
fix optional type hints in `meta`

juliagsy:
fix the type hints here as well! Most functions here are new to me so would be better for you to take a look before merging in case there are mistakes! Thank you!


djl11 (approved):
lgtm!

PR #1738, opened 2022-07-02
Searching submodule

jiahanxie353:
- All tests in test_array_api passing

- Tests with test_array_function passing

### DIFF
--- a/ivy/functional/ivy/general.py
+++ b/ivy/functional/ivy/general.py
@@ -1395,10 +1395,10 @@ def shape_to_tuple(shape: Union[int, Tuple[int], List[int]]):
     """
     if ivy.is_array(shape):
         raise Exception("shape_to_tuple does not accept arrays as input")
-    if isinstance(shape, int):
+    if isinstance(shape, tuple):
         return shape
     else:
-        return tuple(shape)
+        return (shape,)
 
 
 @handle_nestable
--- a/ivy_tests/test_ivy/test_functional/test_core/test_searching.py
+++ b/ivy_tests/test_ivy/test_functional/test_core/test_searching.py
@@ -1,0 +1,212 @@
+"""Collection of tests for searching functions."""
+
+# global
+from hypothesis import given, strategies as st
+import numpy as np
+
+# local
+import ivy_tests.test_ivy.helpers as helpers
+import ivy.functional.backends.numpy as ivy_np
+
+# argmax
+@given(
+    array_shape=helpers.lists(
+        st.integers(1, 10), min_size="num_dims", max_size="num_dims", size_bounds=[1, 10]
+    ),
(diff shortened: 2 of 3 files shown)


ivy/functional/ivy/general.py
```diff
@@ -1395,10 +1395,10 @@ def shape_to_tuple(shape: Union[int, Tuple[int], List[int]]):
     """
     if ivy.is_array(shape):
         raise Exception("shape_to_tuple does not accept arrays as input")
-    if isinstance(shape, int):
+    if isinstance(shape, tuple):
         return shape
     else:
-        return tuple(shape)
+        return (shape,)
```
djl11, line 1401: we actually **do** want to explicitly cast to `tuple` here, rather than place the existing `shape` instance into a `tuple`. Backend-specific shape classes will in general support casting to `tuple`. If the shape is an `int`, it doesn't _really_ matter whether we return a `tuple` or not, but we can in this case change to `return (shape,)` rather than `return shape` if an `int` is detected, and I don't think this will break anything, and is in fact probably better given the name of the function.

ivy_tests/test_ivy/test_functional/test_core/test_searching.py
```diff
@@ -0,0 +1,212 @@
...
+    data,
+    as_variable,
+    with_out,
+    num_positional_args,
+    native_array,
+    container,
+    instance_method,
+    fw,
+):
+    x = data.draw(
```
djl11, line 37: we should not use `data.draw` in the test body, we should instead refactor the helper functions, and find a solution which uses `st.composite`. If the helper is not very general, then feel free to add a private helper function into this `test_searching.py` file. However, if the helper is general and potentially useful for other tests, then it should be added to the `helpers.py` file, so that others can also make use of it.

ivy_tests/test_ivy/test_functional/test_core/test_searching.py
```diff
@@ -0,0 +1,212 @@
...
```
djl11, line 88: see comment above.

ivy_tests/test_ivy/test_functional/test_core/test_searching.py
```diff
@@ -0,0 +1,212 @@
...
```
djl11, line 139: see comment above.

ivy_tests/test_ivy/test_functional/test_core/test_searching.py
```diff
@@ -0,0 +1,212 @@
...
+    data,
+    as_variable,
+    with_out,
+    num_positional_args,
+    native_array,
+    container,
+    instance_method,
+    fw,
+):
+    condition = data.draw(
```
djl11, line 184: see comment above.

djl11 (changes requested):
see comments :)

PR #1742, opened 2022-07-03
added softplus instance and static methods to array and container

CerberusLatrans:


CerberusLatrans:
#1741

ivy/container/activations.py
```diff
@@ -1,9 +1,38 @@
...
+import ivy
 from ivy.container.base import ContainerBase
 
 # ToDo: implement all methods here as public instance methods
 
 
 # noinspection PyMissingConstructor
 class ContainerWithActivations(ContainerBase):
-    pass
+    def softplus(
```
ZoeCD, line 13: Hi! Great job! But before merging, could you please add the container arguments needed? You can see an example [here](https://github.com/unifyai/ivy/blob/8d1eef71522be7f98b601e5f97bb2c54142795b3/ivy/container/elementwise.py#L1240)

ZoeCD (changes requested):

ivy/container/activations.py
CerberusLatrans: Hey Zoe, thanks for the feedback! I added the container args-- let me know if anything else needs changing.

ZoeCD (approved):
Thank you! 😊

PR #1744, opened 2022-07-03
gather_nd function docstrings done.

zaeemansari70:


zaeemansari70:
#1637

ivy/functional/ivy/general.py
```diff
@@ -2308,8 +2308,79 @@ def gather_nd(
...
+
+    With :code:`ivy.Container` input:
+
+    >>> x = ivy.Container(a=ivy.array([0., 1., 2.]), \
+                          b=ivy.array([4., 5., 6.]))
+    >>> y = ivy.Container(a=ivy.array([0]), \
+                          b=ivy.array([2]))
+    >>> print(ivy.gather_nd(x, y))
+    {
+     	a: ivy.array(0.),
```
VedPatwardhan, line 2379: This line causes a lint failure as it seems to have a mix of spaces and tabs, please update that to have only spaces.

ivy/array/general.py
```diff
@@ -39,3 +39,24 @@ def all_equal(
...
+
+        Examples
+        --------
+        >>> x = ivy.array([1, 2, 3])
+        >>> y = ivy.array([4, 5, 6])
+        >>> z = x.gather_nd(y)
+        >>> print(z)
+        ivy.array(3)
+        """
+        return ivy.gather_nd(self._data, indices, out=out)
```
VedPatwardhan, line 62: Please update this line to replace `self._data` with `self`, as `self._data` refers to native array while we want to call the function with the ivy array.

VedPatwardhan (changes requested):
Hi @zaeemansari70, please make the changes where I've mentioned. Also please update the docstring so that it's only indented with spaces, because otherwise it causes the numpy style docstring test to fail. 

ivy/array/general.py
```diff
@@ -39,3 +39,24 @@ def all_equal(
...
+    ) -> Union[ivy.Array, ivy.NativeArray]:
+        """
+        ivy.Array instance method variant of ivy.gather_nd. This method simply wraps the
+        function, and so the docstring for ivy.gather_nd also applies to this method
+        with minimal changes.
+
+        Examples
+        --------
+        >>> x = ivy.array([1, 2, 3])
+        >>> y = ivy.array([4, 5, 6])
```
VedPatwardhan, line 57: Please update `y` which is being passed to the function as index cannot be larger than 2 for the current value of `x`

VedPatwardhan (changes requested):

VedPatwardhan (changes requested):
Hi @zaeemansari70, the docstring tests are failing because of passing a `device` argument to the function in some examples. The function doesn't accept a `device` argument currently, so please remove it from all examples which pass a `device` argument to the function. Thanks 🙂

ivy/container/general.py
```diff
@@ -177,3 +177,97 @@ def all_equal(
...
+    ) -> ivy.Container:
+        """
+        ivy.Container instance method variant of ivy.gather_nd.
+        This method simply wraps the function, and so the docstring
+        for ivy.gather_nd also applies to this method
+        with minimal changes.
+
+        Examples
+        --------
+        >>> x = ivy.Container(a=ivy.array([1, 2, 3]),\
```
VedPatwardhan, line 254: Hi @zaeemansari70, please update this example as the output of the example is incorrect.

VedPatwardhan (changes requested):

ivy/functional/ivy/general.py
```diff
@@ -2309,8 +2309,83 @@ def gather_nd(
...
+    Examples
+    --------
+    With :code:`ivy.Array` input:
+
+    >>> x = ivy.array([0., 1., 2., 3., 4., 5., 6.])
+    >>> y = ivy.array([1])
+    >>> print(ivy.gather_nd(x, y))
+    ivy.array(1.)
+
+    >>> x = ivy.array([[0., 1., 2.], \
```
VedPatwardhan, line 2321: Based on the logs of the test, the output of this example seems to be incorrect as well, please execute it on your machine and then update the result accordingly.

PR #1753, opened 2022-07-03
set_split_factor function reformatting

utkusaglm:
There is an information about split factors in the map func.
`  split_factors
            The updated split factors 0 < sf < 1 for each device. Default is None.`
 I did not find any other information about factor parameter. If there is an issue, please let me know.


MarShaikh:
Everything looks great! Thanks!

PR #1757, opened 2022-07-04
Creation Function Docstrings

jkeane508:
Updated doc-strings

Change in ivy_tests to accompany for rename of test_array_function


mattbarrett98 (changes requested):
It seems as though you've copy and pasted the same docstring for all of them, so you're linking to the elementwise tan function in all of them. Also you're changes have quite a few failures in the `lint` CI check if you could take a look at these.

jkeane508:
Oh yea, I forgot to go back and change the links, thanks!

mattbarrett98:
Yes there is! I presume you haven't set up the pre commit hook yet. Go to your terminal and run `pip install pre-commit`, then run `pre-commit install`. This should then set up the pre commit hook for you, which means every time you try and git commit, the lint checks will automatically run (and if there are failures the commit will fail, and will continue to block your commits until the issues are resolved)

mattbarrett98 (commented):
this looks good to me, apart from a very small detail in the meshgrid link, it says `mesgrid`

jkeane508:
Ok, the branch was a lot of commits behind but it should be up do date now. The change to the torch backend, I think is from the pre commit hook? Its only a formatting change.

ivy/functional/ivy/creation.py
```diff
@@ -685,14 +854,20 @@ def full(
 @to_native_arrays_and_back
 @handle_out_argument
 @handle_nestable
-def from_dlpack(x: Union[ivy.Array, ivy.NativeArray]) -> ivy.Array:
+def from_dlpack(
+    x: Union[ivy.Array, ivy.NativeArray],
+    out: Optional[ivy.Array] = None,
```
djl11, line 859: `out` argument should be keyword-only, with a `*` added above it. This is the case for all functions (I may have missed this for others).

djl11 (changes requested):
aside from one small change, looks good to me! Feel free to merge once this small thing is fixed.

PR #1759, opened 2022-07-04
ivy.array creation functions implementation

jkeane508:
Hi,

I implemented the creation functions in ivy.array

Thanks,
James


mattbarrett98 (changes requested):
Hi, you should only add the instance method if one of the arguments of the function is an array. For example, `zeros` doesn't have any array inputs and so can't have an instance method.

Also I don't think you should be turning `self._data` into a list like in `tril`, unless I'm missing something 🙂

mattbarrett98 (changes requested):
You're still calling `[self._data]` in some of the functions, it should just be `self._data`. And in some you changed it to just `self`, these should be `self._data` also. Again some of these instance methods can't be added as they don't have any array inputs. You can completely remove any of them which don't have any array inputs aside from the out argument. So for example `arange`, `zeros` and a few others can be removed.

mattbarrett98:
I think you will need to make use of the `*arrays` syntax like in the functional `meshgrid` in order to allow a variable number of arrays, in addition to the array the instance method is called on. I think maybe you would have to create a list with `self` and all the other arrays, and then pass this into `ivy.meshgrid`, i'm not sure though.

jkeane508:
For `meshgrid` I just appended the `*args` to `[self]`. You can then expand the `list` using `*` when you call a function with `*args`

mattbarrett98 (changes requested):
Only noticing one thing- in the `native_array` method you are repeating the code in the `ivy.native_array` function. Instead of doing this you can just call `ivy.native_array` directly. 

After this small change i think you can request a final review from Dan 🙂

mattbarrett98:
Sorry didn't see that message, yes I agree using `._data` here is better

ivy/array/creation.py
```diff
@@ -1,8 +1,92 @@
...
+        return ivy.empty_like(self._data, dtype=dtype, device=device, out=out)
+
+    def meshgrid(
+        self: ivy.Array,
+        *arrays: Union[ivy.Array, ivy.NativeArray],
+        indexing: Optional[str] = "xy",
+    ) -> List[ivy.Array]:
+        list = [self._data]
+        for array in arrays:
+            list.append(array._data)
```
djl11, line 74: this is wrong, the arrays might be `ivy.NativeArray` instances in which case this will fail. You can just do `list = [self._data] + list(arrays)`.

djl11 (changes requested):
Aside from some small changes, it looks good to me!

The only other thing missing is the docstrings, as explained [here](https://github.com/unifyai/ivy/blob/master/docs/partial_source/deep_dive/10_docstrings.rst#docstrings) under **classes**. You don't need to add any docstring examples though :)

ivy/array/creation.py
jkeane508: Should we be calling ._data on the arrays argument like with self?

ivy/array/creation.py
```diff
@@ -1,8 +1,92 @@
...
+        *,
+        copy: Optional[bool] = None,
+        dtype: Optional[Union[ivy.Dtype, ivy.NativeDtype]] = None,
+        device: Optional[Union[ivy.Device, ivy.NativeDevice]] = None,
+    ) -> ivy.Array:
+        return ivy.asarray(self._data, copy=copy, dtype=dtype, device=device)
+
+    def full_like(
+        self: ivy.Array,
+        fill_value: Union[int, float],
```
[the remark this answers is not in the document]
jkeane508: I've just saw this and yea lint didn't like having `list()` and `list` in the same function, `list_arrays' should suffice

PR #1765, opened 2022-07-04
reformat  DefaultDevice.__exit__

ahembal:
- [ ] #1764

### DIFF
--- a/ivy/functional/ivy/device.py
+++ b/ivy/functional/ivy/device.py
@@ -81,7 +81,26 @@ class DefaultDevice:
         set_default_device(self._dev)
         return self
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(self, exc_type, exc_val, exc_tb) -> Union[ivy.Device, str]:
+        """
+        Exit the runtime context related to the specified device.
+
+        Returns
+        -------
+        ret
+            Self, an instance of the same class.
+
+        Examples
+        --------
+        >>> ivy.set_default_device("gpu:0")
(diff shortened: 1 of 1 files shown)


Ishticode (changes requested):
Hi @ahembal.
I think you need an example of what would happen after a `with` statement is **exited** from which is when `__exit__` is called. It should print the device set before it. The examples added are merely using the `unset_default_device` function directly, unless I missed something. Please request my review when you make the changes.

Ishticode (changes requested):
The examples are currently inaccurate. As clarified above in previous review `__exit__` statement isn't called inside `with` but rather is the method executed when `with` statement is finished and **exited** from. This is in contrast to `__enter__` which is called when `with` is initiated for example. You can check the docstring for `__enter__` for more context.  This is one usecase.  Please search  `__exit__` on google to read more on if this does not make sense. Let me know for any clarifications 😊.

ahembal:
:) Let me check it again, this is the context manager usage, and I need to show 'with block' in the example, right? 
So, maybe I just need to move the print statement outside of the with block, 
at the end of the block __exit__ code is run in background. 
Then, print statement outside of the block, can show the default device.
Do I understand correctly?

ahembal:
Hi @Ishticode, could you review it again? 
I changed the comment and put print statement after with block, so 'print' will be run after __exit__() function.
Thanks.

Ishticode (approved):
lgtm! Thanks for your contribution here. I will get it merged.

PR #1773, opened 2022-07-05
fix tests in `losses`

juliagsy:
- added assertion for `min` < `max` in `clip` (previously failing tests are fixed through this)
- refactored `epsilon` value in tests
- moved checking to generation phase


ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
```diff
@@ -52,10 +52,10 @@ def test_cross_entropy(
     fw,
 ):
     pred_dtype, pred = dtype_and_pred
-    if fw == "torch" and pred_dtype == "float16":
-        return
     true_dtype, true = dtype_and_true
     length = min(len(true), len(pred))
+    true = true[:length]
```
djl11, line 57: we should actually ensure that both arrays are the same length at _generation_ time, by using [st.shared](https://hypothesis.readthedocs.io/en/latest/data.html?highlight=shared#hypothesis.strategies.shared). See the [test](https://github.com/unifyai/ivy/blob/bdc34708852f87e4a600808d8892d0ddd3edc053/ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py#L38) for `ivy.concat` for an example where there are shape consistency requirements for the arrays being tested, which we handle at data generation time.

ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
```diff
@@ -86,7 +86,7 @@ def test_cross_entropy(
         min_dim_size=2,
     ),
     dtype_and_pred=helpers.dtype_and_values(
-        ivy_np.valid_float_dtypes,
+        tuple(set(ivy_np.valid_float_dtypes).difference(set({"float16"}))),
```
djl11, line 89: same question as above.

ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
```diff
@@ -117,10 +117,10 @@ def test_binary_cross_entropy(
     fw,
 ):
     pred_dtype, pred = dtype_and_pred
-    if fw == "torch" and pred_dtype == "float16":
-        return
     true_dtype, true = dtype_and_true
     length = min(len(true), len(pred))
+    true = true[:length]
+    pred = pred[:length]
```
djl11, line 123: same as above.

ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
```diff
@@ -131,16 +131,18 @@ def test_binary_cross_entropy(
...
         epsilon=epsilon,
     )
 
 
 # sparse_cross_entropy
 @given(
     dtype_and_true=helpers.dtype_and_values(
-        ivy_np.valid_int_dtypes,
+        tuple(
+            set(ivy_np.valid_int_dtypes).intersection(set({"uint8", "int32", "int64"}))
```
djl11, line 144: same as above, this function supports all other integer types in some capacity, and none of these are supported by any of the backends?

djl11 (changes requested):

ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
```diff
@@ -66,11 +65,11 @@ def test_cross_entropy(
         instance_method,
         fw,
         "cross_entropy",
-        true=np.asarray(true[:length], dtype=true_dtype),
-        pred=np.asarray(pred[:length], dtype=pred_dtype),
+        true=np.asarray(true, dtype=true_dtype),
```
djl11, line 68: Just want to confirm, are you varying the shape of `pred` and `true` during data generation? This [line](https://github.com/unifyai/ivy/blob/29ed0b33ef412b1325d186bd9ef741d95c287892/ivy_tests/test_ivy/helpers.py#L1219) indicates to me that you are, but just wanted to be sure.

ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
```diff
@@ -131,8 +129,8 @@ def test_binary_cross_entropy(
         instance_method,
         fw,
         "binary_cross_entropy",
-        true=np.asarray(true[:length], dtype=true_dtype),
-        pred=np.asarray(pred[:length], dtype=pred_dtype),
+        true=np.asarray(true, dtype=true_dtype),
```
djl11, line 132: same comment as above.

ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
juliagsy: I'm generating arrays of same length but st.shared has been implemented in the helpers functions for that after checking in the module, so additional checking is removed!

PR #1776, opened 2022-07-05
Added docstrings to with_grad

### LINKED ISSUE #1775: with_grads

sampreet-arthi:
Resolves #1775


ivy/functional/ivy/gradients.py
```diff
@@ -37,17 +37,42 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        If with_grads is boolean, it returns the boolean value representing
+        if gradient computation is enabled or not.
+        If with_grads is None, it returns the last element in the with_grads_stack
+        representing the parent of the current nested code block. If with_grads_stack
+        is empty, it returns True by default.
+        If with_grads is neither None nor boolean, it will raise an AssertionError
+
+    Examples
+    --------
+    Can be used by calling with ivy.with_grads(with_grads=...)
```
chie2727, line 64: Hi @sampreet-arthi , the examples are looking good!
You can delete line 64 as this line isn't necessary.

chie2727 (changes requested):

ivy/functional/ivy/gradients.py
sampreet-arthi: Okay sure, is there anything I should do regarding the failing checks?

chie2727 (changes requested):
Hi @sampreet-arthi , I've just added some linespaces between your examples to see if that was causing the issue in the `test-docstrings` check. It seemed that was one of the problems, but another problem now is that the check is expecting your outputs to be:
`False False True False`
 whereas your outputs are:
`True False True False`

Could you look into why your first output is different from what the check is expecting?

sampreet-arthi:
The with_grads_stack variable is initially an empty stack, so when we call it for the very first time, it should ideally add True to the stack and return True. Not completely sure about this, but it's possible that while running multiple tests, the with_grad_stacks variable might not be an empty stack when we test the example here. I've edited the example to make it more clear.

chie2727 (approved):

chie2727:
Hi @sampreet-arthi , the docstring test is passing now so thank you for making the changes!
(As for the lint check this is failing because of another file in the repository so no worries about that!)
I've removed the `import ivy` line as this isn't needed in the examples section.
I'll merge in the rest of your changes now.

PR #1780, opened 2022-07-05
removed raise exception

RickSanchezStoic:
removed the raise e from helpers.py for docstrings as requested.


RickSanchezStoic:
Changes made as requested. Hopefully, this should resolve docstring error lines not being printed.

Issue #794, opened 2022-04-03
Reformat Set


HotzingTone:
Each method in the [set](https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/set.py) submodule should be modified to use the latest formatting rules, as explained in the [contributor guide](https://lets-unify.ai/ivy/contributing.html) for the [Ivy API](https://lets-unify.ai/ivy/contributing.html#ivy-api) and [Backend API](https://lets-unify.ai/ivy/contributing.html#backend-api). This includes removing all lambdas and direct bindings for methods in the backend submodules, with each method instead defined using . Full type hints should also be given for each argument in both the ivy module and also each backend module. The docstrings should be updated to use NumPy style rather than Sphinx style, and code examples should be added to every method docstring.

- [ ] unique_inverse
- [ ] unique_values
- [ ] unique_counts

LINBOTANG123:
#858

ysiqueira:
[] #994

Ya2001:
- [ ] #1232

HarishmaElanchery:
unique_values #994

AnimeshSinha1309:
- [ ] #1022

CSheets53:
- [] #1748

EngrE:
unique_inverse #858

vaneesa-writes:
unique_values #6356

chkoutsis:
- [ ] #9390

Ahmadmalj:
#9534

Albert-Osei:
Can I have unique_values #9390 assigned to me?

soma2000-lang:
unique_values #10708

saib7:
- [ ] #10847

shatabdi-pal:
-[ ]#unique_values

adhamebr:
#12411

John-Aradan:
- [ ] #12544

ManasVardhan:
#12949

plapacz6:
#12984

UlisesCasal:
#12987

John-Aradan:
> #12544 

@ivy-seed @ivy-leaves my pull-request on unique_values has been merged. Please refer to it as complete in the subtasks list.

arlou022:
- [ ] #13203

alex-gdv:
#23422

adlub:
#23539

Mr-Array22:
#28313

PR #1786, opened 2022-07-06
unique_counts #1748 Function Reformatting

CSheets53:
Reformatted set.unique_counts per #794 and #1748

### DIFF
--- a/ivy/array/set.py
+++ b/ivy/array/set.py
@@ -1,8 +1,31 @@
 # global
 import abc
+from typing import Optional, Tuple
+
+# local
+import ivy
 
 # ToDo: implement all methods here as public instance methods
 
 
 class ArrayWithSet(abc.ABC):
-    pass
+    def unique_counts(
+        self: ivy.Array,
+        *,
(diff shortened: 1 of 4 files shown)


sherry30 (changes requested):
Hi,

You don't have to add `out` argument to `unique_counts` since it returns multiple arrays.
So you can remove that from the container and array function as well.
One more thing, you should test static_method like, `y = ivy.static_unique_counts(x)`

Thanks

CSheets53:
> Hi,
> 
> You don't have to add `out` argument to `unique_counts` since it returns multiple arrays. So you can remove that from the container and array function as well. One more thing, you should test static_method like, `y = ivy.static_unique_counts(x)`
> 
> Thanks

Ok, thank you! I interpreted the docs to mean `out` goes on every function - would I only use `out` when a method returns a single array then?

sherry30:
Yes, that's correct. Since `unique_counts` returns 2 arrays, we wont be using `out` argument for it.

sherry30 (approved):
Looks perfect, 
Thanks for the contribution 😃

PR #1788, opened 2022-07-06
fix `numpy` frontend `add` tests

juliagsy:
in progress...

### DIFF
--- a/ivy/functional/backends/jax/data_type.py
+++ b/ivy/functional/backends/jax/data_type.py
@@ -1,5 +1,6 @@
 # global
 import numpy as np
+import torch
 import jax
 import jaxlib
 import jax.numpy as jnp
@@ -86,6 +87,8 @@ class Finfo:
 
 def astype(x: JaxArray, dtype: jnp.dtype, *, copy: bool = True) -> JaxArray:
     dtype = ivy.as_native_dtype(dtype)
+    if isinstance(x, torch.Tensor):
+        x = x.numpy()
     if copy:
         if x.dtype == dtype:
             new_tensor = jnp.array(x)
--- a/ivy/functional/backends/numpy/data_type.py
+++ b/ivy/functional/backends/numpy/data_type.py
@@ -1,5 +1,6 @@
 # global
 import numpy as np
+import torch
 from typing import Union, Tuple, List
 
 # local
@@ -82,6 +83,8 @@ class Finfo:
 
 def astype(x: np.ndarray, dtype: np.dtype, *, copy: bool = True) -> np.ndarray:
     dtype = ivy.as_native_dtype(dtype)
+    if isinstance(x, torch.Tensor):
+        x = x.numpy()
     if copy:
         if x.dtype == dtype:
             new_tensor = np.copy(x)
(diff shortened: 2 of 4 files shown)


ivy/functional/backends/jax/data_type.py
```diff
@@ -86,6 +87,8 @@ def smallest_normal(self):
 
 def astype(x: JaxArray, dtype: jnp.dtype, *, copy: bool = True) -> JaxArray:
     dtype = ivy.as_native_dtype(dtype)
+    if isinstance(x, torch.Tensor):
```
djl11, line 90: we need `torch` code inside the `jax` backend. If this is required, this suggests a deeper problem with the test.

ivy/functional/backends/numpy/data_type.py
```diff
@@ -82,6 +83,8 @@ def smallest_normal(self):
 
 def astype(x: np.ndarray, dtype: np.dtype, *, copy: bool = True) -> np.ndarray:
     dtype = ivy.as_native_dtype(dtype)
+    if isinstance(x, torch.Tensor):
```
djl11, line 86: same again.

ivy/functional/backends/torch/data_type.py
```diff
@@ -39,6 +42,13 @@ def astype(x: torch.Tensor, dtype: torch.dtype, *, copy: bool = True) -> torch.T
     dtype = ivy.as_native_dtype(dtype)
     if isinstance(dtype, str):
         dtype = ivy.as_native_dtype(dtype)
+    if isinstance(x, np.ndarray):
+        x = torch.from_numpy(x)
+    elif isinstance(
+        x, (jax.interpreters.xla._DeviceArray, jaxlib.xla_extension.DeviceArray)
```
djl11, line 48: same again.

djl11 (changes requested):
We will need to fix the test in a different way. We shouldn't need to mix backend frameworks in the backend implementations at all.

juliagsy:
I've fixed the `torch` frontend tests for now, I'm looking into one `numpy` frontend issue but I'll raise my question later on or through another PR after finding out more! Thanks!

ivy/functional/frontends/numpy/mathematical_functions.py
```diff
@@ -15,8 +15,8 @@ def add(
     subok=True,
 ):
     if dtype:
-        x1 = ivy.astype(x1, ivy.as_ivy_dtype(dtype))
-        x2 = ivy.astype(x2, ivy.as_ivy_dtype(dtype))
+        x1 = ivy.astype(ivy.native_array(x1), ivy.as_ivy_dtype(dtype))
```
djl11, line 18: cleaner to call `ivy.array` here instead.

ivy/functional/frontends/numpy/mathematical_functions.py
```diff
@@ -38,7 +38,7 @@ def tan(
     subok=True,
 ):
     if dtype:
-        x = ivy.astype(x, ivy.as_ivy_dtype(dtype))
+        x = ivy.astype(ivy.native_array(x), ivy.as_ivy_dtype(dtype))
```
djl11, line 41: same comment as above.

djl11 (changes requested):
small changes, after which it's good to merge!

djl11 (approved):
lgtm, provided that the tests are passing!

PR #1791, opened 2022-07-06
fix `meta` tests

juliagsy:
- fix `test_ivy` failures
- get `meta` tests to pass

### DIFF
--- a/ivy/container/container.py
+++ b/ivy/container/container.py
@@ -150,9 +150,10 @@ class Container(
                           [8.1, 9.3, 3.4]])
         }
         """
-        return ivy.Container.multi_map(
-            lambda xs, _: operator.add(xs[0], xs[1]), [self, other], map_nests=True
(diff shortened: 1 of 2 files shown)


ivy/container/container.py
```diff
@@ -150,9 +150,10 @@ def __add__(self, other):
                           [8.1, 9.3, 3.4]])
         }
         """
-        return ivy.Container.multi_map(
-            lambda xs, _: operator.add(xs[0], xs[1]), [self, other], map_nests=True
-        )
+        # return ivy.Container.multi_map(
```
djl11, line 153: actually, `__add__` should not call `static_add`, as `__add__` should also support adding non-array types, such as `int` etc.

djl11 (changes requested):

djl11 (approved):
lgtm!

PR #1798, opened 2022-07-06
adding helper functions to get test_broadcast_to working

simonetgordon:
I've implemented some helper functions very similar to ones in the array-api hypothesis helpers in order to implement `test_broadcast_to` and `test_broadcast_arrays` in `test_dtype`. Just checking that they way I've implemented them is OK. I've put all strategies and associated function in `test_dtype` as they all pertain to generating broadcastable arrays. I don't think this is needed for other submodules (could be wrong).

I have also updated the implementations of `test_broadcast_to` and `test_broadcast_arrays` to have data generated in the decorator only by making use of the strategies I implemented.


ivy_tests/test_ivy/test_functional/test_core/test_dtype.py
```diff
@@ -103,10 +157,62 @@ def test_broadcast_to(
         fw,
         "broadcast_to",
         x=x,
-        shape=array_shape,
+        shape=to_shape,
     )
 
 
+# # broadcast_to - attempt to have only one data.draw in function
```
djl11, line 164: We should remove this commented-out code when merging I guess? It's replaced by another implementation it seems.

ivy_tests/test_ivy/test_functional/test_core/test_dtype.py
```diff
@@ -91,9 +136,18 @@ def test_broadcast_to(
     instance_method,
     fw,
 ):
-    x = data.draw(helpers.nph.arrays(shape=array_shape, dtype=input_dtype))
+    # smoke for torch uints
+    if fw == "torch" and "u" in str(dtype):
+        return
+    x = data.draw(helpers.nph.arrays(shape=in_shape, dtype=dtype))
+    to_shape = data.draw(
```
djl11, line 143: we can pretty much use the same logic as you have written here, but refactor it into a private helper function (which we can store in the file `test_dtype.py` if you don't think it is general enough to add into `helpers.py`). The first argument of any `st.composite` function is `draw`, meaning we can wrap this exact same logic into a private helper function, which is then called inside the `@given` decorator. Feel free to commit a failing attempt if you can't get it working, and I can then take a look as well :)

ivy_tests/test_ivy/test_functional/test_core/test_dtype.py
```diff
@@ -67,12 +67,57 @@ def test_astype(
...
+    dtype,
+    data,
+    as_variable,
+    native_array,
+    container,
+    fw,
+):
+    arrays = []
+    for c, shape in enumerate(shapes, 1):
+        x = data.draw(helpers.nph.arrays(dtype=dtype, shape=shape), label=f"x{c}")
```
djl11, line 94: see comment below on, regarding use of `data.draw`

djl11 (changes requested):
just some small comments, happy to dive deeper if you try and refactor the `data.draw` logic using `st.composite` and are unable to get it working!

djl11 (approved):
lgtm!

PR #1807, opened 2022-07-07
fix `handle_nestable` decorator in `is_variable`

juliagsy:
- include container instance method for `is_variable`
- add `handle_nestable` decorator to `is_variable`
- fix tests in `meta` to be compatible with `is_variable` returning containers


djl11 (approved):
lgtm! Please just remove the commented-out lines of code before merging :) Thanks!

PR #1810, opened 2022-07-07
Cholesky Function Docstrings

zaeemansari70:


VedPatwardhan (changes requested):
Hi @zaeemansari70, please fix the import issue as we won't be able to see if the docstrings are failing until we fix that. Thanks 🙂

ivy/functional/ivy/linear_algebra.py
```diff
@@ -50,37 +50,86 @@ def cholesky(
...
+    --------
+    With :code:`ivy.Array` input:
+
+    >>> x = ivy.array([[1., -2.], [2., 5.]])
+    >>> l = ivy.cholesky(x, 'false')
+    >>> print(l)
+    ivy.array([[ 1., -2.],
+               [ 0.,  1.]])
+
+    >>> x = ivy.array([[2, 3],[4, 9]])
```
VedPatwardhan, line 63: Please test this example as it's showing a not_implemented error

ivy/functional/ivy/linear_algebra.py
```diff
@@ -50,37 +50,86 @@ def cholesky(
...
+    >>> y = ivy.zeros(3)
+    >>> ivy.cholesky(x, 'false', out=y)
+    >>> print(y)
+    ivy.array([3.01e+23, 1.93e+00, 0.00e+00])
+
+    >>> x = ivy.array([[7, 2], \
```
VedPatwardhan, line 69: Please test this example as it's showing a not_implemented error

ivy/functional/ivy/linear_algebra.py
```diff
@@ -50,37 +50,86 @@ def cholesky(
...
+
+    >>> x = ivy.array([[1., -2.], [2., 5.]])
+    >>> upper = ivy.Container(a=1, b=-1)
+    >>> y = ivy.cholesky(x, 'false')
+    >>> print(y)
+
+    ivy.array([[ 1., -2.],
+               [ 0.,  1.]])
+
+    Using :code:`ivy.Container` instance method:
```
VedPatwardhan, line 121: Instance method examples should be in the container function, please move them to the container/linear_algebra.py file

VedPatwardhan (changes requested):
Hi @zaeemansari70, there are a few examples which are throwing a not_implemented error, when a backend is set, please test them once at your end and update accordingly. 

Thanks 🙂

VedPatwardhan (changes requested):
Hi @zaeemansari70, as the integer datatype is not supported in all backends, we shouldn't be having an example in the docstrings where the datatype of the input arrays passed is an integer, so please update that.
Thanks 🙂

ivy/functional/ivy/linear_algebra.py
```diff
@@ -50,37 +50,94 @@ def cholesky(
...
+                        [0., 0.707]])
+    }
+
+    With a mix of :code:`ivy.Array` and :code:`ivy.Container` inputs:
+
+    >>> x = ivy.array([[1., -2.], [2., 5.]])
+    >>> upper = ivy.Container(a=1, b=-1)
+    >>> y = ivy.cholesky(x, 'false')
+    >>> print(y)
+
```
VedPatwardhan, line 138: Please remove this additional line as it seems to fail the docstring tests

VedPatwardhan (changes requested):

VedPatwardhan (approved):
lgtm!

PR #1812, opened 2022-07-07
Docstrings and Type Hints added - Adam and LAMB failing tests

Auxia:
Requesting review on the added docstrings and type hints (Specifically the nestable step function type hints).
Also noticed failing tests of Adam and LAMB optimizers when updating the weights.

### DIFF
--- a/ivy/stateful/optimizers.py
+++ b/ivy/stateful/optimizers.py
@@ -2,6 +2,7 @@
 
 # global
 import abc
+from typing import Union, Optional
 
 # local
 import ivy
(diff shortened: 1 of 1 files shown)


ivy/stateful/optimizers.py
```diff
@@ -180,13 +188,11 @@ def __init__(
         compile_on_next_step
             Whether to compile the optimizer on the next step. Default is False.
         """
-        Optimizer.__init__(
-            self, lr, inplace, stop_gradients, compile_on_next_step=compile_on_next_step
-        )
+        Optimizer.__init__(self, lr, inplace, stop_gradients, compile_on_next_step)
```
saeedashrraf, line 191: The usage of the kword argument (compile_on_next_step=compile_on_next_step) is necessary because compile_on_next is the 6th argument and it is being passed as the 5th argument here.

ivy/stateful/optimizers.py
```diff
@@ -359,9 +362,13 @@ def __init__(
         self._first_pass = True
         self._should_compile = False
 
+        Optimizer.__init__(
+            self, lr, inplace, stop_gradients, compile_on_next_step, device
+        )
+
```
saeedashrraf, line 368: The True argument is necessary to be passed as True to always set the init_on_first_step value as True. This also ensures that compile_on_next_step is being passed in the correct position and so, it is not necessary to pass it as a keyword argument in this case. Since device follows fallback_to_non_compiled, it should be passed as a keyword argument (device=device).

ivy/stateful/optimizers.py
```diff
@@ -476,7 +484,7 @@ def __init__(
```
saeedashrraf, line 484: Since device follows fallback_to_non_compiled, it should be passed as a keyword argument (device=device).

saeedashrraf (changes requested):

saeedashrraf (approved):
Looks good to me!

PR #1814, opened 2022-07-07
Implicit out

jiahanxie353:
Finish adding out arg to all the creation submodules of each backend.


mattbarrett98 (changes requested):
The `out` argument should only exist in the backend function if the backend directly supports the `out` argument. jax and TensorFlow don't have any support for an `out` argument and so none of our backend functions for jax/tensorflow should have an `out` argument. numpy and torch in some cases support the `out` argument, and so maybe could be added in some cases, but only if the NumPy/torch function in question directly supports the `out` argument. 

Have you read the deep dive: https://lets-unify.ai/ivy/deep_dive/8_inplace_updates.html ?

jiahanxie353:
> The `out` argument should only exist in the backend function if the backend directly supports the `out` argument. jax and TensorFlow don't have any support for an `out` argument and so none of our backend functions for jax/tensorflow should have an `out` argument. numpy and torch in some cases support the `out` argument, and so maybe could be added in some cases, but only if the NumPy/torch function in question directly supports the `out` argument.
> 
> Have you read the deep dive: https://lets-unify.ai/ivy/deep_dive/8_inplace_updates.html ?

Thanks for pointing out :)
Yes, it's true that some backend functions do not support `out`. 
My first step is to simply add `out` to every backend function;
Then,  I will need to add a flag here https://github.com/unifyai/ivy/blob/c0b7e9424af99b0faf495b950ad249bc2ad54401/ivy/func_wrapper.py#L248
to indicate whether or not to defer to the backend, or to handle the `out` argument explicitly using `inplace_update`.

I haven't added this flag yet, and will re-request after handling it.

mattbarrett98 (commented):
lgtm apart from the unused imports in jax general 🙂

djl11 (approved):
provided that:

a) you don't introduce any newly failing tests
b) you are able to fix the merge conflicts
c) you are confident that it's all been done correctly

Then feel free to merge! Thanks!

PR #1816, opened 2022-07-07
`ivy.container` creation functions

jkeane508:
ivy.container implementation of creation functions


jkeane508:
I have a question regarding function arguments

mattbarrett98 (changes requested):
For the instance methods i think you can just make use of the static method, for example like with `add` https://github.com/unifyai/ivy/blob/8d1eef71522be7f98b601e5f97bb2c54142795b3/ivy/container/elementwise.py#L92. Also the `*` in the method arguments should probably be after `map_sequences`, again similar to how it's done in `add`.

djl11 (approved):
lgtm!

PR #1823, opened 2022-07-08
update shape to `ivy.Shape` and `ivy.NativeShape` 

juliagsy:
- added `ivy.Shape` and `ivy.NativeShape` classes
- updated type hints and returns in ivy and backends
- added shape_array_mode methods (set, unset) with docstrings
- implemented array wrapping to shapes

todo: will update `shape_to_tuple` method to the equivalent shape and native shape methods in another PR!

thanks!


ivy/__init__.py
```diff
@@ -59,6 +63,22 @@ def __new__(cls, dtype_str):
         return str.__new__(cls, dtype_str)
 
 
+class Shape(tuple):
+    def __new__(cls, shape_tup):
+        assert isinstance(
+            shape_tup, (int, list, tuple, ivy.current_backend().NativeShape)
```
djl11, line 69: you can replace `ivy.current_backend().NativeShape` with `ivy.NativeShape`, if a backend is currently set, it will have overwritten `ivy.NativeShape`

ivy/__init__.py
```diff
@@ -59,6 +63,22 @@ def __new__(cls, dtype_str):
...
+        )
+        if isinstance(shape_tup, int):
+            shape_tup = tuple([shape_tup])
```
djl11, line 72: small detail, but there is no need to cast to `list` first. You can just do `shape_tup = (shape_tup,)`

ivy/__init__.py
```diff
@@ -59,6 +63,22 @@ def __new__(cls, dtype_str):
...
+        elif isinstance(shape_tup, list):
+            shape_tup = tuple(shape_tup)
+        for v in shape_tup:
+            assert isinstance(v, int)
```
djl11, line 76: `assert all([isinstance(v, int) for v in shape_tup])`

ivy/functional/backends/numpy/random.py
```diff
@@ -2,10 +2,10 @@
 
 # global
 import numpy as np
-from typing import Optional, Union, Tuple, Sequence
+from typing import Optional, Union, Sequence
 
 # localf
```
djl11, line 7: typo

ivy/functional/ivy/general.py
```diff
@@ -2560,6 +2562,21 @@ def shape(
     return current_backend(x).shape(x, as_array)
 
 
+def set_shape_array_mode() -> None:
+    global shape_array_mode
```
djl11, line 2566: this should use a stack rather than a single global value enabling us to use nested blocks of setting and unsetting. See other global variables for examples.

djl11 (changes requested):
Please see comments :)

Didn't go through in crazy detail. Once you've done, request another review, and provided that no new tests are failing, we can get this merged pretty quickly. Thanks!

ivy/__init__.py
```diff
@@ -59,6 +63,19 @@ def __new__(cls, dtype_str):
...
 
 
+class Shape(tuple):
+    def __new__(cls, shape_tup):
+        assert isinstance(shape_tup, (int, list, tuple, ivy.NativeShape))
+        if isinstance(shape_tup, int):
+            shape_tup = (shape_tup,)
+        elif isinstance(shape_tup, list):
+            shape_tup = tuple(shape_tup)
+        assert ivy.all(ivy.array([isinstance(v, int) for v in shape_tup]))
```
djl11, line 73: no need to use `ivy.all` and `ivy.array` here, just `all([isinstance(v, int) for v in shape_tup])`

djl11 (changes requested):
small, change requested, then feel free to merge this yourself provided the tests are passing!

ivy/__init__.py
juliagsy: some of the functions threw array type errors when I used them directly 😅

or let's say we merge it first and I'll find a fix in another pr since I'll be changing up the `shape_to_tuple` methods in a new PR after merging this?

PR #1824, opened 2022-07-08
Set submodule Fix up 

rush2406:
- added array instance methods
- added container instance methods
- removed `np.random_uniform` and `data.draw()`
- fixed ivy tests


rush2406:
Currently, `unique_counts` and `unique_inverse` are failing for tensorflow with the following error 
```
x = array([1, 0], dtype=int8), y = ivy.array([0, 1]), rtol = 0.001, atol = 1e-06

    def assert_all_close(x, y, rtol=1e-05, atol=1e-08):
        if ivy.is_ivy_container(x) and ivy.is_ivy_container(y):
            ivy.Container.multi_map(assert_all_close, [x, y])
        else:
>           assert np.allclose(
                np.nan_to_num(x), np.nan_to_num(y), rtol=rtol, atol=atol
            ), "{} != {}".format(x, y)
E           AssertionError: [1 0] != ivy.array([0, 1])
```

Is this because of the difference in order between [0 1] and [1 0]?

VedPatwardhan:
I think it's not about the order, but it seems like the assertion fails because a list and an ivy.Array are being compared

ivy/array/set.py
```diff
@@ -1,12 +1,9 @@
 # global
 import abc
-from typing import Tuple
-
-# local
-import ivy
+from typing import Optional, Tuple
```
VedPatwardhan, line 3: Please revert this to what it was before, basically the structure should be 
```# global
import abc
from typing import Optional, Tuple

# local
import ivy

```

ivy/array/set.py
```diff
@@ -1,12 +1,9 @@
...
 
 # ToDo: implement all methods here as public instance methods
-
+import ivy
```
VedPatwardhan, line 7: Please add another blank line here

ivy/container/set.py
```diff
@@ -4,12 +4,47 @@
 # local
 import ivy
 from ivy.container.base import ContainerBase
+import ivy
```
VedPatwardhan, line 7: ivy has been reimported, please remove that.

ivy/container/set.py
```diff
@@ -4,12 +4,47 @@
...
+from typing import Optional, List, Union, Dict
```
VedPatwardhan, line 8: This import has been already made in line 2, please update that line as redefining fails the lint tests

ivy/container/set.py
```diff
@@ -46,21 +81,19 @@ def static_unique_counts(
...
+            prune_unapplied: bool = False,
+            map_sequences: bool = False
     ) -> ivy.Container:
         """
-        ivy.Container instance method variant of ivy.unique_counts. This method 
-        simply wraps the function, and so the docstring for ivy.unique_counts 
+        ivy.Container instance method variant of ivy.unique_counts. This method
+        simply wraps the function, and so the docstring for ivy.unique_counts
         also applies to this method with minimal changes.
-
```
VedPatwardhan, line 59: This blank line shouldn't be removed, as it fails the numpy-style-docstring test

ivy_tests/test_ivy/test_functional/test_core/test_set.py
```diff
@@ -32,6 +32,10 @@ def test_unique_values(
     fw,
     device,
 ):
+    #uint not supported with torch
```
VedPatwardhan, line 35: The comment should have a space after `#` otherwise it fails the lint tests

VedPatwardhan (changes requested):
Hi @rush2406, so I remember Dan saying that `np.random.uniform` shouldn't be used for data generation while writing tests, rather the data should be generated through hypothesis, please update the data generation accordingly. You could use the [`dtype_and_values`](https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/helpers.py#L1180) helper to generate the data instead, because that'll generate data according to a strategy rather than being totally random.

Please also update the code where I've mentioned.

rush2406:
@VedPatwardhan can you review the changes? The tests are passing now

ivy/array/set.py
```diff
@@ -1,20 +1,17 @@
...
 
-# ToDo: implement all methods here as public instance methods
+import ivy
 
 
 class ArrayWithSet(abc.ABC):
-    def unique_counts(
-        self: ivy.Array
-    ) -> Tuple[ivy.Array, ivy.Array]:
+    def unique_counts(self: ivy.Array) -> Tuple[ivy.Array, ivy.Array]:
```
djl11, line 11: looking at the example, seems as though this is actually a `NamedTuple`, does this exist in the typing module?

ivy/array/set.py
```diff
@@ -24,6 +21,17 @@ def unique_counts(
...
+
+    def unique_values(self: ivy.Array, out: Optional[ivy.Array] = None) -> ivy.Array:
+        return ivy.unique_values(self._data, out=out)
+
+    def unique_all(
+        self: ivy.Array,
+    ) -> Tuple[ivy.Array, ivy.Array, ivy.Array, ivy.Array]:
+        return ivy.unique_all(self._data)
+
+    def unique_inverse(self: ivy.Array) -> Tuple[ivy.Array, ivy.Array]:
```
djl11, line 36: again, this might be `NamedTuple`?

ivy/array/set.py
```diff
@@ -24,6 +21,17 @@ def unique_counts(
         >>> x = ivy.array([0., 1., 2. , 1. , 0.])
```
djl11, line 21: not necessary to add `With :code:'ivy.Array' instance method:`

ivy/container/set.py
```diff
@@ -1,22 +1,53 @@
...
     def static_unique_counts(
         x: ivy.Container,
         key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
         to_apply: bool = True,
         prune_unapplied: bool = False,
-        map_sequences: bool = False
+        map_sequences: bool = False,
     ) -> ivy.Container:
         """
         ivy.Container static method variant of ivy.unique_counts. This method simply
```
djl11, line 53: No need to include `With :code:'ivy.Container' static instance method:`

ivy/functional/backends/jax/set.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: again, is there not a `NamedTuple` type hint?

ivy/functional/backends/mxnet/set.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: `NamedTuple` type hint

ivy/functional/backends/numpy/set.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: `NamedTuple` type hint

ivy/functional/backends/tensorflow/set.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: `NamedTuple` type hint

ivy/functional/backends/torch/set.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: `NamedTuple` type hint

djl11 (changes requested):
see comments :)

ivy/array/set.py
rush2406: yes `NamedTuple` exists and it seems to be more suitable here, have changed it

djl11 (approved):
lgtm! Provided the tests are passing, feel free to merge :)

PR #1831, opened 2022-07-08
Added hard_sigmoid, linear, and exponential to Tensorflow frontend  

Bobbyorr007:
Added hard_sigmoid, linear, and exponential to Tensorflow frontend


djl11:
We will not be accepting extensions to Ivy until the next release is out. Additionally, all new functions must be implemented in all backends, the ivy API, and as methods on the `ivy.Array` and `ivy.Container` class. Also, the frontend task is not open yet. Please re-open another PR when this task is open.

PR #1843, opened 2022-07-09
`compilation` fix up

juliagsy:
I've been making changes through PR and reviews, but just in case, here's a final one!


djl11 (approved):
lgtm! Feel free to simply close with PR, given that there are no changes requested. Thanks!

PR #1848, opened 2022-07-10
`losses` fix up

juliagsy:
I've been making changes through pr and reviews, but here's a final one just in case! Thanks!

### DIFF
--- a/ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
+++ b/ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
@@ -1,3 +1,4 @@
+# for review
 # global
 import numpy as np
 from hypothesis import given, strategies as st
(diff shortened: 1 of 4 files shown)


ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
```diff
@@ -1,3 +1,4 @@
+# for review
 # global
 import numpy as np
 from hypothesis import given, strategies as st
```
djl11, line 4: explicit test skipping based on `dtype` such as `if fw == "torch" and pred_dtype == "float16":` and `fw == "tensorflow" and true_dtype not in ["uint8", "int32", "int64"]:` in the tests should not be necessary. We should remove these and handle these using the `unsupported_dtype` attributes on the backend functions.

djl11 (changes requested):
see comments.

PR #1849, opened 2022-07-10
`utility` fix up

juliagsy:
I have been making changes through pr and reviews, but just in case, here's a final one! Thanks!


ivy/functional/backends/jax/utility.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: `return ret` creates a redundant new line.

ivy/functional/backends/tensorflow/utility.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: `return ret` creates a redundant new line

ivy/functional/ivy/utility.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: we should not import `current_backend` explicitly, we should instead simply call `ivy.current_backend`

ivy_tests/test_ivy/test_functional/test_core/test_utility.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: we should not make any use of `data.draw`, we should either refactor into private helper functions in this test file itself, or add the helper functions to the `helpers.py` file for other tests to make use of.

djl11 (changes requested):
see comments.

PR #1854, opened 2022-07-11
`elementwise` fix up

rush2406:
- modified container instance methods
- fixed tests


rush2406:
@VedPatwardhan are all of the elementwise functions required to be added into `array/__init__.py`?

ivy/array/elementwise.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: could you please add the `ivy.Array instance method variant of ivy.<>....` docstrings for all instance methods? Could you please also add all parameters into the docstrings, as explained [here](https://github.com/unifyai/ivy/blob/d8f8fe6986145051e50816fe7bea33e92d997a86/docs/partial_source/deep_dive/11_docstrings.rst#docstrings)?

ivy/container/elementwise.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: could you please add the `ivy.Array instance method variant of ivy.<>....` docstrings for all instance methods? Could you please also add all parameters into the docstrings, as explained [here](https://github.com/unifyai/ivy/blob/d8f8fe6986145051e50816fe7bea33e92d997a86/docs/partial_source/deep_dive/11_docstrings.rst#docstrings)?

djl11 (changes requested):
see comments :)

PR #1856, opened 2022-07-11
formatting section in deep dive

juliagsy:
Thought it might be better to have you read or skim through just in case I left out anything :)) Btw, is there any particular section or location you want it to be in? Currently it's in its own page but I haven't list it down in the main deep dive lists. Thanks!


djl11 (approved):
lgtm! Could you please add this as a new sub-section of the deep dive, between "Inplace Updates" and "Function Arguments"? You will need to:

a) rename some of [these](https://github.com/unifyai/ivy/tree/2cdca28425ada3ca7d388442041b435f1bd6cf2e/docs/partial_source/deep_dive) files
b) modify the list (a) through (n) and also the `.. toctree::` [here](https://github.com/unifyai/ivy/blob/2cdca28425ada3ca7d388442041b435f1bd6cf2e/docs/partial_source/deep_dive.rst)

I think:

"How the code is automatically formatted 📋"

will suffice for the short description in the `deep_dive.rst` file

Thanks!

PR #1857, opened 2022-07-11
vmap implementations

Ishticode:
`ivy.vmap` implementations inspired by `jax.vmap`.


Ishticode:
Hi.
The `test_vmap` is passing for the backends `[numpy, tensorflow, jax]`. Checked with a max of 500 hypothesis generated examples and made sure its hitting actual examples and not passing for the wrong reasons.
However, `torch` requires that the function being vmapped return torch Tensors and not Ivy.Array. Just putting the decorator `@inputs_to_native_arrays` on the _to be mapped_ function kind of makes it work but not practical for testing ivy functions which are mostly decorated with `@to_native_arrays_and_back`. Will check more on this soon. It should be a simple fix as the results from torch are identical to jax across examples.
The implementation will be improved using optimized vectorisation functions by the host frameworks and extended/adding more functionality/control but this is the first iteration and has the essential working components.

ivy/functional/backends/numpy/transformations.py
```diff
@@ -0,0 +1,78 @@
...
+            for (arg, axis) in zip(args, in_axes):
+                if axis is not None:
+                    axis_size.add(arg.shape[axis])
+
+        if len(axis_size) > 1:
+            raise ValueError('''Inconsistent sizes. All mapped axes should have the same size''')  # noqa
+
+        # Making sure not all in_axes are None
+        if isinstance(in_axes, (list, tuple)):
+            assert not all(ax is None for ax in in_axes), "At least one of the axes should be specified (not None)"  # noqa
```
djl11, line 43: we can add new lines using `\`, no need for `# noqa` here

ivy/functional/backends/numpy/transformations.py
```diff
@@ -0,0 +1,78 @@
...
+        # Handling None in in_axes by broadcasting the axis_size
+        if isinstance(in_axes, (tuple, list)) and None in in_axes:
+            none_axis_index = list()
+            for index, axis in enumerate(in_axes):
+                if axis is None:
+                    none_axis_index.append(index)
+
+            for none_mapped_axis in none_axis_index:
+                args[none_mapped_axis] = np.broadcast_to(args[none_mapped_axis],
+                                                         (tuple(axis_size) + args[none_mapped_axis].shape))  # noqa
```
djl11, line 56: we can add a new line, here no need for `# noqa`. Maybe format it with pre-commit.

ivy/functional/backends/torch/transformations.py
```diff
@@ -0,0 +1,17 @@
...
+# local
+import ivy
+
+
+def vmap(func: Callable,
+         in_axes: Union[int, Sequence[int], Sequence[None]] = 0,
+         out_axes: Optional[int] = 0) -> Callable:
+    def _vmap(*args):
+        new_fun = lambda *args: ivy.to_native(func(*args)) 
+        new_func = functorch.vmap(new_fun, in_axes, out_axes)
```
djl11, line 14: is this not natively supported in `torch` now? Or maybe this is only a partial version?

djl11 (changes requested):
looks mainly good, couple of requests:

could you move this all to `general.py` for now? I think adding a very light `transformations` module doesn't make sense right now. Also, there were a lot of un-necessary `# noqa` lines, please revisit those. Otherwise looks good :)

ivy/functional/backends/torch/transformations.py
Ishticode: At the moment, torch only has vmap in the nightly versions as they consider it a prototype. The team has made `functorch` as an extension of torch for functional programming. In [this Issue](https://github.com/pytorch/pytorch/issues/42368) on Pytorch repo, they referrerd to functorch. It seems `vmap` might be incorporated into `torch` stable versions when it is fully functional in `functorch`.

ivy/functional/backends/numpy/transformations.py
Ishticode: sure.

ivy/functional/backends/numpy/transformations.py
Ishticode: sure

Ishticode:
Currently waiting on ivy tests to be fixed.

Issue #790, opened 2022-04-03
Reformat Creation


HotzingTone:
Each method in the [creation](https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/creation.py) submodule should be modified to use the latest formatting rules, as explained in the [contributor guide](https://lets-unify.ai/ivy/contributing.html) for the [Ivy API](https://lets-unify.ai/ivy/contributing.html#ivy-api) and [Backend API](https://lets-unify.ai/ivy/contributing.html#backend-api). This includes removing all lambdas and direct bindings for methods in the backend submodules, with each method instead defined using . Full type hints should also be given for each argument in both the ivy module and also each backend module. The docstrings should be updated to use NumPy style rather than Sphinx style, and code examples should be added to every method docstring.

- [ ] asarray
- [ ] zeros
- [ ] ones
- [ ] full_like
- [ ] ones_like
- [ ] zeros_like
- [ ] tril
- [ ] triu
- [ ] empty
- [ ] empty_like
- [ ] eye
- [ ] linspace
- [ ] meshgrid
- [ ] zeros_like
- [ ] full
- [ ] ones
- [ ] from_dlpack
- [ ] array
- [ ] logspace
- [ ] arange

kirodev:
[ ] #842

RitujaPawas:
- [ ]  #896

RitujaPawas:
- [ ] #906

RitujaPawas:
- [ ] #933

RitujaPawas:
- [ ] #934

RitujaPawas:
- [ ] #935

SayanJerr:
: - [ ] #945

SayanJerr:
-[ ] #963

RitujaPawas:
- [ ] #977

RitujaPawas:
- [ ] #978

RitujaPawas:
- [ ] #979

RitujaPawas:
- [ ]  #980

kirodev:
[ ] : #998

TudorAndrei:
- [ ] #1020

RitujaPawas:
- [ ] #1046

RitujaPawas:
- [ ] #1047

RitujaPawas:
- [ ] #1048

RickSanchezStoic:
- [ ] #1112

RickSanchezStoic:
- [ ] #1113

[...]

PR #1863, opened 2022-07-11
adding more example on zeros_like function doc.

MHamidA:


MHamidA:
#1862

MHamidA:
Hii @Oxbeefbabe,
I haven't assigned to the #790 issues yet, but I already did this pull request (idk if this is allowed or not), and about this pull request, I don't think there is a lot of reformatting that needs to be done. I just added more examples (and am still not quite sure if I should specify the input type or not). It'd be great if there are any suggestions or corrections. TIA!

Oxbeefbabe:
Hello @MHamidA ,

This is a good start! You should include examples showing how each of the arguments affects the output; using the function with ivy.Containers; and using the function as an instance method.

Oxbeefbabe:
I see that you are using the ivy.set_framework(). This function is outdated and we now use ivy.set_backend() (e.g. ivy.set_backend('numpy') ) If you are working in a fork then it is likely that you need to merge with upstream to get the most up-to-date version of Ivy, so you can use this new function. Using zeros_like() with NumPy as an instance method works fine for me.

I can't find the example where Daniel mentioned not to add instance method examples. It is still shown in the Docstings Examples page in the wiki. I think to be safe its better to add it :)

MHamidA:
I see.. thanks for the confirmation. I'll try to start work on it by tonight.🙌😁

MHamidA:
Hello @Oxbeefbabe :))

I've added more examples with some variations. Please review it thankuu.

MHamidA:
Sorry, I forgot to replace the output and forgot to check some input formats; please see the most recent commit thanku! 🙌

Oxbeefbabe:
Good stuff! You still should add an example making use of the device and out parameters. The device parameter is a bit of a tricky one. I'd recommend using the string 'cpu' and then adding a comment under the example output explaining that the array is now stored on the string.

MHamidA:
Hii @Oxbeefbabe, please review it once again (esp. the example of the device argument is the way to use that argument and the comment given is appropriate) Thankuuu 😅

Oxbeefbabe (commented):
That looks great! I'll accept and merge this shortly :)

MHamidA:
Thankuu! One more thing regarding the issue that I opened. Am I the one that needs to close the issue or is the Ivy team the one that will close it?

PR #1878, opened 2022-07-11
Reformat depthwise_conv2d

vincent1bt:
#1804 issue


sherry30 (changes requested):
Hi, 

The changes you made look good, a few more things need to be added:
1. You have to implement the array instance method in `ivy/array/layers.py` and write 1 example of it in that newly created function.
2. You have to implement container instance method in `ivy/container/layers.py` and write 1 example of it in that function
3. Implement `static_depthwise_conv2d` in the same container file and write 1 example of it.

You can use reference of `tan` function to implement these, [here](https://github.com/unifyai/ivy/blob/056bebf21a0222ae3978d7075b1db94ddfdc6c3c/ivy/array/elementwise.py#L379) is array instance method, [here](https://github.com/unifyai/ivy/blob/056bebf21a0222ae3978d7075b1db94ddfdc6c3c/ivy/container/elementwise.py#L1616) is container instance method, and [here](https://github.com/unifyai/ivy/blob/056bebf21a0222ae3978d7075b1db94ddfdc6c3c/ivy/container/elementwise.py#L1580) is container static method.

Feel free to ask any question, I'd be happy to help 😃

sherry30 (changes requested):
Hi,

the docstring examples are failing since `depthwise_conv2d` is not yet implemented in jax and numpy backend. Can you add `depthwise_conv2d` in the `skip_list_temp` in `test_docstrings.py`

vincent1bt:
Hello, I added ```depthwise_conv2d``` to the ```skip_list_temp``` list but it didn't work so I checked for new changes and use the ```skip_arr_cont``` list, now the ```depthwise_conv2d``` tests pass, however, the ```unique_counts``` and ```nonzero``` methods contain some errors that are show in the tests

sherry30 (approved):
The failing docstring seems to be unrelated to yours so its all good.

Thanks for the contribution.

PR #1884, opened 2022-07-12
`sorting` fix up

rush2406:
- modified tests to use dtype_values_axis
- fixed failing test


ivy/functional/backends/torch/sorting.py
```diff
@@ -20,3 +20,6 @@ def sort(
 ) -> torch.Tensor:
     sorted_tensor, _ = torch.sort(x, dim=axis, descending=descending, out=out)
     return sorted_tensor
+
+
+sort.unsupported_dtypes = ("uint16", "uint32", "uint64")
```
djl11, line 25: `torch` does not support any of these data types at all, and so this line is unnecessary

ivy_tests/test_ivy/test_functional/test_core/test_sorting.py
```diff
@@ -112,7 +104,7 @@ def test_sort(
         instance_method,
         fw,
         "sort",
-        x=x,
+        x=np.asarray(x),
```
djl11, line 107: same as above

djl11 (changes requested):
some small changes, but most importantly, you should add `# for review` at the top of **every** file which has anything to do with sorting, including `ivy.Array` and `ivy.Container` instance methods etc., the Ivy folder, and **all** backend `sorting.py` files.

ivy/functional/ivy/sorting.py
```diff
@@ -1,3 +1,4 @@
+# for review
 # global
 from typing import Union, Optional
```
djl11, line 4: we should simply call `ivy.current_backend`, rather than importing it.

djl11 (changes requested):
small change requested. After this, if the tests are all passing, then feel free to merge :)

PR #1898, opened 2022-07-12
fix full torch backend

ZoeCD:


simonetgordon:
Nice spot! So many tests failing due to the one function 😅 Many of the assertion errors are expected as similar ones have appeared for the other backends. We may end up just skipping the ones with health check errors.

PR #1902, opened 2022-07-13
argmax passing tests

201843052:
Please ignore the ivy.iml and misc.xml files generated by pycharm

All I've done here is to force the tests to pass by changing shape into [shape], so that torch.full() in ivy/functional/backends/torch/creation.py can be created as a result of shape_to_tuple in ivy/functional/ivy/general.py correctly returning a shape tuple.

I remember seeing shape_to_tuple being something different, that when the shape is an integer, (shape,) will be returned... need help to clarify this :)

### DIFF
--- a/.idea/ivy.iml
+++ b/.idea/ivy.iml
@@ -1,8 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module type="PYTHON_MODULE" version="4">
   <component name="NewModuleRootManager">
-    <content url="file://$MODULE_DIR$" />
-    <orderEntry type="jdk" jdkName="Remote Python 3.8.10 Docker (unifyai/ivy:latest)" jdkType="Python SDK" />
+    <content url="file://$MODULE_DIR$">
+      <excludeFolder url="file://$MODULE_DIR$/venv" />
+    </content>
+    <orderEntry type="inheritedJdk" />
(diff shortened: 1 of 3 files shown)


201843052:
Also, Functional Examples are being added to the Docstrings, should I still follow the python API standard docstrings?

JLU-Neal:
Hi @Ishticode!
Could you double-check whether this task is completed? If so, you can mark it as completed.

Ishticode:
Hi @201843052. Thanks for the PR 😊. Thanks @JLU-Neal for the cross checking and comments above.

I don't think test_argmax was failing. `test_ivy` only has `test_argmax`  under `array_api_tests` which is passing for all backends when I just checked now. Some big changes were being made recently so It might be that the the test was failing at the time you checked but at [this point in the commit hash](https://github.com/unifyai/ivy/tree/48f718b116b6ffebea9e87380dda898105cfbdb2) it is passing.

I will close the PR now. If you think I am mistaken then please open the PR again and elaborate. You could also go for the function reformatting task as explained in the [Open Tasks](https://lets-unify.ai/ivy/contributing/4_open_tasks.html#open-tasks) section and make a PR to contribute.
Thanks a lot 😊.

201843052:
Thank you all for the help. Will hopefully help somewhere else :)

PR #1913, opened 2022-07-13
broadcast_to #1910

aditya-pola:
Refactoring broadcast_to mentioned in issue #1910 
Note: The ivy.Container example only works when all the arrays in the container share the same shape. Unsure how or if we need to account for cases where the arrays have different sizes.


modanesh:
Hi,

Regarding the container shape concern, that's fine. Actually, that's the expected behavior. 

The examples mostly look good, just a few things you need to do:

1- In `ivy.data_type`, docstring example for Container has a syntax error in `broadcast_to()`. You should pass the container as input as well as the shape.

2- You have to add docstring examples for `ivy.array` and `ivy.container` instance methods. The docstring examples for `ivy.array` will be added in `ivy/array/data_types.py` and for container it will be added in `ivy/container/data_types.py`. 

All of these requirements are mentioned [here](https://github.com/unifyai/ivy/blob/76d65119fef75a3592a6e2cd744c568ea73f7651/docs/partial_source/deep_dive/11_docstring_examples.rst) in details. Don't hesitate to ask any further questions if you have any. I'd be happy to help.

aditya-pola:
@modanesh I've made the suggested changes. Please review them.

PR #1924, opened 2022-07-13
Reformatting repeat #1879

palazski:
Hi,

This is the pr for #1879 

One question: Should I replace the lambda function inside of repeat function in ivy/container/manipulation.py? Swapaxes method in that file hasn't done that but you merged it anyway so I got confused there because one of the tasks of Function Reformatting is to replace lambda functions with functions declared using def.

Also I believe I saw a few missed points in the merged toDO's.
(I used them as examples hence I noticed the problems)
* in ivy/array/manipulation.py, "clip" method's docstring is copy-paste from "roll" function. problem is that now "clip" method's docstring says "this is the container instance method of ivy.roll function" while it should be saying "this is the container instance method of ivy.clip function"

* in ivy/container/manipulation.py, "swapaxes" method has no docstring

### DIFF
--- a/ivy/container/manipulation.py
+++ b/ivy/container/manipulation.py
@@ -287,6 +287,45 @@ class ContainerWithManipulation(ContainerBase):
             out=out,
         )
 
+    @staticmethod
+    def static_repeat(
+        x: Union[ivy.Array, ivy.NativeArray, ivy.Container],
+        repeats: Union[int, Iterable[int]],
+        axis: Optional[Union[int, Tuple[int, ...]]] = None,
+        key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
+        to_apply: bool = True,
+        prune_unapplied: bool = False,
+        map_sequences: bool = False,
+        *,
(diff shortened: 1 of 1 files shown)


abdullahpalaz:
Hi,

I've looked through the checks here and tried to fix them. Some should be fixed. Once the checks done again, I will be updating the docstrings if needed.

Also, I've replaced the lambda function and added a Container static function for repeat. I am not really sure about it's correctness so can you give me feedback on that?

Thanks

abdullahpalaz:
After check update:

I guess only the "test-docstrings" check is the one that concerns me since other unsuccessful checks are not related to repeat method. I've looked "test-docstrings" up to find what is the problem.

Problematic example in docstring of "array instance method" seemed fine to me. Your feedback will be the guide here.

Thanks

ZoeCD:
Hi! Thank you for your contribution! Your PR looks good to me! As you mentioned, some CI tests fail for unrelated issues, so don't worry about them. Also, thank you for the missed points you mentioned; we'll take a look.

PR #1927, opened 2022-07-14
update `shape_to_tuple` to `to_ivy_shape` and `to_native_shape`

juliagsy:
- added `to_ivy_shape` and `to_native_shape` methods
- removed `shape_to_tuple` method, which is equivalent to `to_ivy_shape` as `ivy.Shape` is of type `tuple`
- replaced `shape_to_tuple` with `to_ivy_shape` in the codebase


ivy/functional/ivy/general.py
```diff
@@ -1458,6 +1459,48 @@ def shape_to_tuple(shape: Union[ivy.Shape, ivy.NativeShape]):
...
+    shape
+        The input to be converted
+
+    Returns
+    -------
+     ret
+        the input in ivy.Shape form
+
+    """
+    return ivy.Shape(shape)
```
djl11, line 1476: we should only call this if `shape` is not an `ivy.Shape` already. I know the output will be the same regardless. But calls to `ivy.Shape` will be compiled into the graph (whereas `isinstance` checks will not), and therefore we should not add any more of these `ivy.Shape` calls into the computation stack than is necessary.

djl11 (changes requested):
one small change, otherwise looks good!

djl11 (approved):
lgtm!

PR #1935, opened 2022-07-14
Reformating `cos` method

hirwa-nshuti:
Hi there 🙌
This PR resolves issue #1932 
I added Docstring examples to the `cos` method, removed the lambda function from the container method, and added the static method for the same function.

Thanks


jkeane508:
Looks good to me!

PR #1952, opened 2022-07-15
Creation functions Fixup

jkeane508:
Edits to:
Type hints on all Creation Backends
Small fixes to Unit tests
Small fix to asarray() docstring


ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -494,14 +494,19 @@ def test_full_like(
...
     dtype=dtype_shared,
 )
 def test_meshgrid(
     arrays,
+    indexing,
     dtype,
     fw,
 ):
 
+    if fw == "torch" and dtype == "uint16" or "uint32":
```
djl11, line 507: we should not have any checks inside the unit test function body, this goes for others as well, but I will just add this comment.

djl11 (changes requested):
Please add `# for review` at the top of every file which is being reviewed, so that I can leave file-specific comments.

Aside from the comment I left, I also noticed that `linspace` and `logspace` are only testing for a fixed size array. Could you further diversify the shapes of the start and stop arrays in these tests?

Once done, please request another review, thanks!

jkeane508:
Issues currently with unit tests not passing for torch backend due to changes to device.

ivy/functional/backends/jax/creation.py
```diff
@@ -20,14 +22,14 @@
 
 
 def arange(
-    start,
-    stop=None,
-    step=1,
+    start: Number,
```
djl11, line 25: `float` encapsulates `int`, and so we can just use `float` for all such type hints, rather than `Number`

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -58,13 +59,14 @@ def test_native_array(
...
-        allow_inf=False,
+        num_arrays=2,
+        min_value=None,
+        max_value=None,
         min_num_dims=1,
         max_num_dims=1,
         min_dim_size=2,
         max_dim_size=2,
-        safety_factor=0.5,
+        shared_dtype=True,
```
djl11, line 69: I think you might need to keep this safey factor? I observed some overflow and underflow errors without it.

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -58,13 +59,14 @@ def test_native_array(
 @given(
     dtype_and_start_stop=helpers.dtype_and_values(
         available_dtypes=ivy_np.valid_numeric_dtypes,
-        num_arrays=1,
-        allow_inf=False,
+        num_arrays=2,
+        min_value=None,
+        max_value=None,
         min_num_dims=1,
```
djl11, line 65: we should diversify the shapes more, with more diversity in the dimensions and dimension sizes.

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -89,25 +91,27 @@ def test_linspace(
...
 # logspace
 @given(
     dtype_and_start_stop=helpers.dtype_and_values(
         available_dtypes=ivy_np.valid_numeric_dtypes,
-        num_arrays=1,
-        allow_inf=False,
+        num_arrays=2,
+        min_value=None,
+        max_value=None,
         min_num_dims=1,
```
djl11, line 110: more diversity for `num_dims`

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -125,7 +129,7 @@ def test_logspace(
 ):
     dtype, start_stop = dtype_and_start_stop
     helpers.test_function(
-        input_dtypes=dtype,
+        input_dtypes=[ivy_np.float32, ivy_np.float32],
```
djl11, line 132: we shouldn't hard-code for `float32` only

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -1,3 +1,4 @@
+# For Review
```
djl11, line 1: we actually shouldn't need to test for `as_array` instance method, this is a pointless ivy.Array instance method, as it will always simply return itself. This will be removed, and so this can be hard-coded as `False` in the test.

djl11 (changes requested):
aside from these small changes, I think it's good. I would say get this merged once these changes have been made, and then you can get the few tests passing afterwards :)

.gitignore
```diff
@@ -21,3 +21,4 @@ with_time_logs/
 *.json
 .hypothesis
 .array_api_tests_k_flag*
+internal_automation_tools/
```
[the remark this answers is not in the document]
jkeane508: I have no idea, will remove.

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
jkeane508: Yea should have been kept in

.gitignore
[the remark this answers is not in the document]
Auxia: This is for the pr dashboard, horian mentioned that this folder should be in the .gitignore so I had added so that I dont accidently commit it to master. It was mentioned in a github discussion.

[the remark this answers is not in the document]
jkeane508: Oh so are we leaving in or removing?

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
jkeane508: yes of course

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
jkeane508: This makes sense

PR #1953, opened 2022-07-15
small fixes on `device` for ivy gym

juliagsy:
when diving into ivy gym, I found out that certain returns from Jax and TensorFlow does not have a device by default, thus throwing errors when an ivy.Array is being created from its native version

are we allowed to add these small changes so that ivy.default_device is set when there isn't one from the backend?

Thanks!


djl11 (commented):
This will only work if the array is on the default device, which it might not be. Let's not merge this yet. Can you think of a solution which will generalize? Could you make a PR in Ivy gym and I will take a look at the problem there?

Issue #1954, opened 2022-07-15
set_nest_at_indices 


Bhavik-Ardeshna:
Reformat nest #793

Working on Reformat nest #793 list of issues.

Bhavik-Ardeshna:
@sreegandh804  Do, I have to highlight this issue ID in the [#793](https://github.com/unifyai/ivy/issues/793) list??

Issue #793, opened 2022-04-03
Reformat Nest


HotzingTone:
Each method in the [nest](https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/nest.py) submodule should be modified to use the latest formatting rules, as explained in the [contributor guide](https://lets-unify.ai/ivy/contributing.html) for the [Ivy API](https://lets-unify.ai/ivy/contributing.html#ivy-api) and [Backend API](https://lets-unify.ai/ivy/contributing.html#backend-api). This includes removing all lambdas and direct bindings for methods in the backend submodules, with each method instead defined using . Full type hints should also be given for each argument in both the ivy module and also each backend module. The docstrings should be updated to use NumPy style rather than Sphinx style, and code examples should be added to every method docstring.

- [ ] index_nest
- [ ] set_nest_at_index
- [ ] map_nest_at_index
- [ ] multi_index_nest
- [ ] set_nest_at_indices
- [ ] map_nest_at_indices
- [ ] nested_indices_where
- [ ] all_nested_indices
- [ ] map
- [ ] nested_map
- [ ] copy_nest

1Doomdie1:
#795

1Doomdie1:
#795

JerryGCDing:
- [ ] #1820

Bhavik-Ardeshna:
set_nest_at_index #1948

Psychofun:
- [ ] #2094

theiyobosa:
#2162

Mouafak-Khaled:
insert_into_nest_at_index #2322

hirwa-nshuti:
nested_indices_where #2356

Jeff-Owuor:
all_nested_indices #1241

Jeff-Owuor:
all_nested_indices

baturgezici:
copy_nest #2416

Lancelot03:
- [ ] all_nested_indices #2609

Jeff-Owuor:
nested_map #2564

a-rushi:
nested_map #2801

philippe753:
- [ ] #3719

roo-afk:
#5848

archbung:
- [ ] 7242

soma2000-lang:
all_nested_indices #10707

kalcho100:
- [] #10709

gretagh93:
all_nested_indices #12112

[...]

PR #1960, opened 2022-07-15
set_nest_at_indices  #1954 

Bhavik-Ardeshna:
**Task:** Change the code of ```set_nest_at_indices``` nest function. (Reformat nest #793)

**Changes:**
- Added generic input argument to functions of Ivy
- Change the docstring of the function
- Run all tests for the same
- Added examples covering all cases


Bhavik-Ardeshna:
I have only changed the function and also tested using the .py file with all test cases which I have mentioned in docstring. But it's giving an error here. Please, address this to me.

Thank you

PR #1961, opened 2022-07-16
multiple containers for nestable returns

Aarsh2001:


djl11 (approved):
lgtm!

docs/partial_source/deep_dive/5_containers.rst
```diff
@@ -197,6 +197,82 @@ of **any** of the arguments.
...
 
+**Explicitly Nestable Functions**
+
+The *nestable* behaviour is added to any function which is decorated with the
+`handle_nestable <https://github.com/unifyai/ivy/blob/5f58c087906a797b5cb5603714d5e5a532fc4cd4/ivy/func_wrapper.py#L407>`_
+wrapper. This wrapper causes the function to be applied at each leaf of any containers
+passed in the input. More information on this can be found in the `Function Wrapping <https://github.com/unifyai/ivy/commit/384963a6d41801e713ec3d203b42bf78d1d7aa0d>`_
+section of the Deep Dive.
+
+Additionally, any nestable function which returns multiple arrays, will return multiple containers for the it's container
```
djl11, line 208: `for the it's` -> `for it's`

djl11, line 208: `will return multiple containers` -> `will return the same number of containers`

docs/partial_source/deep_dive/5_containers.rst
```diff
@@ -197,6 +197,82 @@ of **any** of the arguments.
...
+counterpart. This property makes the function symmetric with regards to the input-output behavior. Any argument in the input,
```
djl11, line 209: This property makes the function symmetric with regards to the input-output behavior

->

This property makes the function symmetric with regards to the input-output behavior, irrespective of whether `ivy.Array` or `ivy.Container` instances are based used.

docs/partial_source/deep_dive/5_containers.rst
```diff
@@ -197,6 +197,82 @@ of **any** of the arguments.
...
+can be replaced with a container, without changing the number of inputs.
```
djl11, line 210: Any argument in the input can be replaced with a container without changing the number of inputs, and the presence or absence of `ivy.Container` instances in the input should not change the number of return values of the function. In other words, if containers are detected in the input, then we should return a separate container for each array that the function would otherwise return.

djl11 (changes requested):
some small changes to the text, then it looks good to go! Thanks!

PR #1976, opened 2022-07-17
edited docstring float add

imranlala99:
hello!

I have looked over the add function elementwise and seemed to find no errors except one added docstring example for clarification, this is the only change I made
I am also yet to implement this change, the only other error that may exist is that tensorflow seems to support broadcasting, however that was not implemented in this backend doc. Instead the method tf.experimental.numpy.add is used. Does it make sense to instead resolve for tf.add and include broadcasting in the wrapper?

**UPDATE**
The function was already reformatted, it was just not updated on the todo list

### DIFF
--- a/ivy/functional/ivy/elementwise.py
+++ b/ivy/functional/ivy/elementwise.py
@@ -271,6 +271,11 @@ def add(
     >>> print(z)
     ivy.array([5, 7, 9])
 
+    >>> x = ivy.array([1, 2, 3])
+    >>> y = ivy.add(x, 5.)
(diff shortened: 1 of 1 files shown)


chie2727:
Hi @imranlala99 , is this PR ready for review?

imranlala99:
Hi @chie2727 I listed this issue because I saw it was open, looked over it and made a PR. But then a couple days later an admin commented on my listing to say that the function was already reformatted. I since made another PR - reshape #2137.

PR #1999, opened 2022-07-17
Container operators

saeedashrraf:


ivy/container/base.py
```diff
@@ -684,8 +684,20 @@ def multi_map(
                     )
                     if ret:
                         return_dict[key] = ret
-                elif isinstance(value0, (list, tuple)) and map_nests:
-                    ret = ivy.nested_multi_map(lambda x, _: func(x, None), values)
+                elif any(isinstance(x, (list, tuple)) for x in values) and map_nests:
+                    to_ivy = False
+                    for value in values:
```
djl11, line 689: could you add comments explaining why this is needed? I would have thought the function being mapped itself converts things to `ivy`? Why do we need to hard code this here you think?

ivy/container/container.py
```diff
@@ -93,15 +93,17 @@ def __pos__(self):
         return self
 
     def __neg__(self):
-        return self.map(lambda x, kc: -x)
+        return self.map(lambda x, kc: -x, map_sequences=True)
```
djl11, line 96: this is fine, but I think we should generalize `self.map` to also handle `map_nests` argument right? Maybe let's do this is another PR though.

djl11 (changes requested):
small request for comments explaining the `to_ivy` logic. Otherwise looks good. I didn't dive too deep into all the nested implementations, but provided the tests are passing and you are happy that the behavior is as expected - then I'm happy!

ivy/container/base.py
saeedashrraf: multi_map handles leaves that contain lists or tuples by delegating the work to the nested_multi_map function which by default always returns ivy arrays regardless of the function being mapped. However, for operator functions, if the leaves only contain numbers and not ivy arrays, nested_multi_map should be instructed to return number values and not ivy arrays through the to_ivy variable.

ivy/container/container.py
saeedashrraf: Currently, `self.map` function can handle `map_nests` with a different variable name `map_sequences` . Do you mean to set `map_sequences` as True by default or what does generalizing `self.map` refer to?

PR #2011, opened 2022-07-18
Added some tensorflow and computer vision modules

Abrar-Mustakim:
I have used this in object detection task. I have set modules version wise so that it will not create any error. If you need to do any tensorflow object detection/normal Deep Learning tasks, you can use this modules.


chie2727:
Hi @MarShaikh , there's no urgency in updating the requirements list at the moment so I'll close this PR. Many thanks!

PR #2013, opened 2022-07-18
Fix submodule linear algerbra

JLU-Neal:


ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -31,7 +31,7 @@ def test_vector_to_skew_symmetric_matrix(
     fw,
     a,
 ):
-    if "float16" or "int8" in input_dtype:
+    if "float16" in input_dtype or "int8" in input_dtype:
```
djl11, line 34: we should actually remove this `if` statement entirely, it should be handled correctly by the helper.

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -810,24 +805,25 @@ def test_svd(
     fw,
     fm,
 ):
-    if "float16" in input_dtype:
+    dtype, x = dtype_x
+    try:
```
djl11, line 809: debug code? We should not be allowing a failing test to pass.

djl11 (changes requested):
mostly looks good - just a couple of changes requested.

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
JLU-Neal: Oh I cooperated with Darshan to fix this submodule, and I was responsible for the last 12 tests (From line 536 test_tensordot to the bottom of the file)

JLU-Neal: Btw, I modified line 34 since I noticed that many of the tests were actually not passed.
eg. In this if block, 
`   
 if "float16" in input_dtype or "int8" in input_dtype:
        return

`
it will go to the "return"

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
JLU-Neal: Same, Darshan will update this later.

JLU-Neal: Ah, there is a problem, when the dtype is float16, which is not supported,
there will not be any return values (line 810)
but as you can see there are some operations (line 829) later which are based on these return values

djl11: Got it, we can instead catch the return with a single parameter, and then unpack the tuple if it is not none. If it is none, we can return. Make sense?

djl11 (changes requested):
Please see comment, the try-catch can be removed. I also think the if statement can be removed without causing errors, please make thus change also, then it's good to merge :)

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
JLU-Neal: Cool, this sounds great.

JLU-Neal:
> Please see comment, the try-catch can be removed. I also think the if statement can be removed without causing errors, please make this change also, then it's good to merge :)

Yeah I will remove the try-catch! 

And I am sure Darshan will remove the if statement in his PR later! (He is working on this tests and my changes could lead to unnecessary conflicts)

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
Darshan-H-E: > we should actually remove this `if` statement entirely, it should be handled correctly by the helper.

Hi @djl11, I've removed this and fixed the vector_to_skew_symmetric_matrix test function on another branch where I've worked on the remaining 12 functions of linalg submodule. I'll request a PR shortly.

PR #2016, opened 2022-07-18
Added Dropout Layer

Abrar-Mustakim:
This layer is essential to remove overfitting from training


Abrar-Mustakim:
Please suggest to me how I can solve the problem and make it a perfect contribution. It is my first time contributing somewhere

djl11:
This is already added compositionally, so no need to add it in the tensorflow backend :)

https://github.com/unifyai/ivy/blob/31a54b397207110a3705789cd407aba07acaad28/ivy/functional/ivy/layers.py#L94

PR #2019, opened 2022-07-18
Uint Data Type

juliagsy:
in progress...

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -89,6 +89,12 @@ class FloatDtype(Dtype):
         return str.__new__(cls, dtype_str)
 
 
+class UintDtype(IntDtype):
+    def __new__(cls, dtype_str):
+        assert "uint" in dtype_str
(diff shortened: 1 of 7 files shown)


ivy/__init__.py
```diff
@@ -312,18 +318,26 @@ class Node(str):
     float32,
     float64,
 )
+all_uint_dtypes = (
```
djl11, line 321: now that we have these added, we should also add them to each of the backend `__init__.py` files as well (for example the torch [`__init__.py`](https://github.com/unifyai/ivy/blob/591ac37a664ebdf2ca50a5b0751a3a54ee9d5934/ivy/functional/backends/torch/__init__.py) file).

You can follow the same format as for the others, such as `valid_int_dtypes`, `valid_float_dtypes` etc.

djl11 (changes requested):
some other additions for each backend, but otherwise looks good!

PR #2024, opened 2022-07-18
Adding droput x in tensor and rate in float

Abrar-Mustakim:


chie2727:
Hi @Abrar-Mustakim , the dropout function you added here isn't required so I'll close this PR. Thanks :)

PR #2026, opened 2022-07-18
Adding dropout with floating point tensor

Abrar-Mustakim:


Auxia:
I see you have made multiple PRs regarding the same function. I suggest closing the ones that are no longer being worked on. Also please mention the Issue Number (like #(No of the Issue)) so that its clearer to us what you are working on.

PR #2032, opened 2022-07-18
added ground_truth_test_functions for compositional_functions

RitujaPawas:


djl11:
We actually changed the way that this should be done now. The approach is to add hypothesis `@example` decorators. Sorry for changing this. This work all looks really helpful. Could you liaise with @Aarsh2001 who is also working on this now? I'll close this for the time being. Thanks @RitujaPawas!

PR #2038, opened 2022-07-18
Addded dropout, If fails, pass

Abrar-Mustakim:


modanesh:
Thanks for your contribution but there is already a compositional dropout function.

PR #2043, opened 2022-07-18
added linear layer

Abrar-Mustakim:


Abrar-Mustakim:
what was the problem in this contribution?

djl11:
Will open again once we have our PR extension pipeline well defined, thanks a lot for the contribution!

PR #2044, opened 2022-07-18
`elementwise` fix up

rush2406:
- removed `np.random_int` in `bitwise_left_shift` and `bitwise_right_shift`
- for `divide`, used `assume` to exclude divide by 0


ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py
```diff
@@ -693,12 +689,10 @@ def test_divide(
...
     x1 = np.asarray(x[0], dtype=input_dtype[0])
     x2 = np.asarray(x[1], dtype=input_dtype[1])
-    # ToDo: remove the checks below, and instead handle this during the
-    #  hypothesis data generation
-    if np.any(x2 == 0):
-        return  # don't divide by 0
-    elif np.any(x1 > 9223372036854775807):
-        return  # np.divide converts to signed int so values can't be too large
+    # np.divide converts to signed int so values can't be too large
+    assume(np.all(x1 <= 9223372036854775807))
```
djl11, line 693: is this needed, with the new `0.95` safety factor in the [data generation helpers](https://github.com/unifyai/ivy/blob/591ac37a664ebdf2ca50a5b0751a3a54ee9d5934/ivy_tests/test_ivy/helpers.py#L1470)?

djl11 (changes requested):
small changes, otherwise good to merge. But if you want a full review of the elementwise module then you will need to add comments to each relevant file, before I can indicate whether it is ready to be marked as complete.

djl11 (changes requested):
seems as though the tests are failing on the PR, could you merge with master and see if this fixes things?

ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py
```diff
@@ -1661,14 +1654,14 @@ def test_pow(
...
     x2 = np.asarray(x[1], dtype=input_dtype[1])
-    if fw in ["jax", "tensorflow"]:
-        return
+
     if (
         np.any(x2 < 0)
         and ivy.is_int_dtype(input_dtype[1])
         and ivy.is_int_dtype(input_dtype[0])
     ):
         return  # ints to negative int powers not allowed
```
djl11, line 1663: we should use `assume` statement here, rather than simply returning.

djl11 (changes requested):
on `request_floor_divide`, the most elegant solution would be to add [this](https://github.com/unifyai/ivy/blob/423074c63bfc751c806f514705c89b10b1597bf8/ivy_tests/test_ivy/helpers.py#L1463) `safety_factor` argument to the helper `test_function`. With a value of `0.5`. This should achieve the same effect as this custom logic in the unit test body.

Also, some tests are checking for conditions and then returning, these should instead make use of the hypothesis `assume` statement.

Once these changes are made, please request a final review and then we can get it merged :)

rush2406:
> on `test_floor_divide`, the most elegant solution would be to add [this](https://github.com/unifyai/ivy/blob/423074c63bfc751c806f514705c89b10b1597bf8/ivy_tests/test_ivy/helpers.py#L1463) `safety_factor` argument to the helper `test_function`. You should then pass a value of `0.5` for this test. This should achieve the same effect as this custom logic in the unit test body.

Currently, `safety_factor` is not in the arguments of `dtype_and_values`, so I have added it with this commit.

djl11 (approved):
please check whether the tests are passing. If so, then it looks okay to me :)

PR #2057, opened 2022-07-19
reformat atan2

roboFiddle:
issue #2050


roboFiddle:
@Oxbeefbabe I fixed the lints. The failure on run-docstring-tests appears to be because of the `repeat` function - should I fix that too. Otherwise, I think this is ready.

Oxbeefbabe:
Nope, you were focusing on atan2. I've had a look at the master branch and this is an issue that is ongoing for the time being, so you don't need to worry about it. This all looks good to me so I'll go ahead and merge this.

Thanks a lot! :)

PR #2065, opened 2022-07-19
Docstring failing tests fix

hirwa-nshuti:
Removed examples from `total_mo_on_dev` and fixed output for examples in `repeat` method


hirwa-nshuti:
@simonetgordon This is the PR solving the failing Docstring tests.

PR #2066, opened 2022-07-19
add try excep in set_backend for importlib

Ishticode:
The `backend_stack` gets emptied in the `while` loop just before the `importlib` block. So if an error is hit during import the stack remains empty. This PR remedies that for `KeyValue` errors i.e. for incorrect framework names passed to `set_backend`.


djl11 (approved):
lgtm, thanks!

PR #2067, opened 2022-07-19
Reformat and extend Expand dims

danielmunioz:
#2005


danielmunioz:
#2005

danielmunioz:
HI! A couple of things here:

1. I removed the 'out' argument for the backend call because some of the backend functions (like expand_dims in the numpy backend) don't take 'out' as an argument, so the code would often break after simply calling `ivy.expand_dims(x)`, however it's worth noting this would only happen **before** setting the backend manually, after using `ivy.set_backend('numpy')` the code would run as expected and after removing the out argument the code would run regardless of whether the backend was previously set or not. If I understand correctly, this should not break the expected behavior for the out argument on the main expand_dims function due to the wrappers, right? Please correct me if I'm wrong
2. The only two backends that accept a List/Tuple for the axis argument are Numpy and JAX, I decided to cautiously add two examples showcasing this use, however since this does not work with all the backends I'm not sure if I should remove it or maybe keep it but with a flag or something to signify the fact that this only works with those two frameworks.

If there is anything I may need to add delete, or change, please don't hesitate on letting me know!!

danielmunioz:
Hi! Just a quick note, As mentioned before the only two backends that accept a List/Tuple for the axis argument are Numpy and JAX, so following numpy's logic for expand_dims I added support for multi-index insertion to the rest of the backends, I'm aware the middle part of the code is a bit repetitive and I thought about wrapping it in something like a "helper" function but I wasn't sure on which script should I out it in, so please let me know if the changes are good and if so where would be the ideal place for this function.

Ishticode (changes requested):
Hi. Thank you for the very comprehensive PR @danielmunioz 
During conflict, the return statement was perhaps removed by mistake. Could that be re added and apologies if that was my mistake.
Could you kindly remove out argument from functional/ivy as @handle_out decorator is already added.
And finally, as the function has @handle_nestable decorator, could you also implement the function as a static method in the container as per the [instruction](https://lets-unify.ai/ivy/deep_dive/5_containers.html#api-static-methods).
Please feel free to ask anything if the above is not clear.

danielmunioz:
Hey! Thanks for the review I just Added the static method, fixed the missing return and modified some docstring examples!
About the removing the out argument on  functional/ivy/manipulation.py, I wonder if is it is necessary? because the functions still receives the argument and that's reflected on the docstrings, so if I remove it I would technically need to remove it from the docstrings as well right? making it a bit confusing for the end user, please feel free to correct me if I'm wrong with this.

Also, about, the logic for the code that I added to support the multi-index insertion is basically the same for all the backends and as a consequence a bit repetitive, so I was wondering if maybe there is more efficient place to put this code in? maybe wrap it in a function and then just call it when it's needed, please let me know if there is a way to do this or if we should just let it the way it is right now.

Thanks for your time!

Ishticode (changes requested):
Hi. You made a good point about out argument. Had slipped my mind. Thanks for that.
Could you get the lint test passing on the CI. The docstring test is also failing for expand_dims alongside some others. Could you check why it is failing for `expand_dims` and get it passing please. Thank you very much. 😊 You're welcome to raise any concerns...

danielmunioz:
Hey! I just modified the lines that were too long and also improved some of the docstring examples, I believe the `expand_dims` functions should be able to pass the lint and docstrings tests now!

danielmunioz:
Hi @Ishticode, just a quick note about the `lambda` used on `expand_dims` over at `ivy/containers/manipulation.py`, I noticed that some of the other functions in that file are using the `static` implementation instead of a `lambda`, however If I recall correctly, on the last "ask us anything", I heard Dan say that it's ok to leave the lambda functions when working on `ivy/containers`, so that's why I left it how it was, but of course, I can change it if it's necessary here!

danielmunioz:
Hey! @Ishticode, just a last minute point, So I noticed you guys added guidelines for helper functions, and in order to improve readability, I decided to pack some of the repetitive logic here into one, Please let me know if there is anything else needed in this PR!

Ishticode (approved):
Hi @danielmunioz 
Thank you for the patience. 
Your PR seems in perfect form 😊
I have tested the additional logic with several examples and I does work. May I suggest that if you do contribute to the logic of functions, updating the tests might be the best way to check if your logic is sound. 

Your addition hasn't broken the previous test and will perhaps pass additional checks (for tuple/list inputs for `axis`) which I will add soon in the test. 
Thank you very much for your contribution @danielmunioz.

Ishticode:
> Hi @Ishticode, just a quick note about the `lambda` used on `expand_dims` over at `ivy/containers/manipulation.py`, I noticed that some of the other functions in that file are using the `static` implementation instead of a `lambda`, however If I recall correctly, on the last "ask us anything", I heard Dan say that it's ok to leave the lambda functions when working on `ivy/containers`, hence why I left it how it was, but of course, I can change it if it's necessary here!

It should use the static version of the function. i will fix that dw.

danielmunioz:
Sure!!, Thanks for the review and your time!! :)

PR #2071, opened 2022-07-19
Updating ivy.device functions and tests

Unevilicorn:
Changes include:
- Type hint to some of the functions in ivy.device
- Hypothesis tests for test_device
- Array, Container instance and static methods for ivy.device


ivy/array/device.py
```diff
@@ -1,8 +1,52 @@
...
+
 
 # ToDo: implement all methods here as public instance methods
 
 
 class ArrayWithDevice(abc.ABC):
-    pass
+    def dev(self: ivy.Array) -> Union[ivy.Device, ivy.NativeDevice]:
+        """
+        ivy.Array instance method variant of ivy.device.dev. This method simply wraps
```
djl11, line 14: we should not include the submodule: we should write `ivy.dev` rather than `ivy.device.dev`, the same is true for most other docstrings.

ivy/functional/ivy/device.py
```diff
@@ -396,7 +400,7 @@ def clear_mem_on_dev(device: Union[ivy.Device, ivy.NativeDevice]) -> None:
         The device string to convert to native device handle.
 
     """
-    return current_backend(None).clear_mem_on_dev(device)
+    current_backend(None).clear_mem_on_dev(device)
```
djl11, line 403: we should actually use `ivy.current_backend` for these calls, and remove the explicit import at the top of the file.

djl11 (changes requested):
some small changes. Feel free to merge once these are fixed. Some other things are not perfect, but can revisit them later as I'm unsure of how best to handle it (for example which functions should be `nestable` and how they should be nestable exactly).

PR #2077, opened 2022-07-19
reformatted result_type #2041

c1boo:


c1boo:
#2041

sherry30 (changes requested):
Hi,

A few changes are required
1. You have to write docstring example under the container instance method and also container static method.
2. Fix the lint issues caused by your changes.
3. Fix the docstring errors caused by `result_type`.

If you face any trouble or have any question, feel free to ask I'd be happy to help you out.

Thanks.

sherry30:
Hi,

So it seems like there was a problem in the backend, this function is supposed to return ivy.Dtype but was returning the native dtype.
I have pushed a fix for this, now they all should be expecting the same answer i.e.  `{a:float32, b:int32}` which was the one you were receiving for torch before.

ivy/container/data_type.py
```diff
@@ -647,6 +647,45 @@ def static_result_type(
...
+
+        Returns
+        -------
+        ret
+            the dtype resulting from an operation involving the input arrays and dtypes.
+
+        Examples
+        --------
+        >>> x = ivy.Container(a = ivy.array([0, 1, 2]), \
+        b = ivy.array([3., 4., 5.]))
```
sherry30, line 680: indent array `b` exactly under array `a`

ivy/container/data_type.py
```diff
@@ -664,6 +703,46 @@ def result_type(
...
+            the dtype resulting from an operation involving the input arrays and dtypes.
+
+        Examples
+        --------
+        >>> x = ivy.Container(a=ivy.array([0., 1., 2.]), \
+            b=ivy.array([3, 4, 5]))
+        >>> print(x.a.dtype, x.b.dtype)
+        float32 int32
+
+        >>> print(x.result_type(x))
```
sherry30, line 740: running result type on the same containers, maybe create a different container `y` and then run it like `x.result_type(y)`

sherry30 (changes requested):
Hi,

Can you add 1 functional example using container as an input in `ivy/functional/ivy/data_type.py`
I pointed out a few other changes on the particular lines 😃

Issue #1708, opened 2022-07-01
copy_array


11Nick26:

11Nick26:
- [copy_array] #1708

11Nick26:
> _No description provided._

- [] #1708

11Nick26:
- [] #1708

1Doomdie1:
Hey, please provide the issue from where you took this function.
Thank you!

11Nick26:
> Hey, please provide the issue from where you took this function. Thank you!

#791

1Doomdie1:
This function has been reserved for you.

Issue #791, opened 2022-04-03
Reformat General


HotzingTone:
Each method in the [general](https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/general.py) submodule should be modified to use the latest formatting rules, as explained in the [contributor guide](https://lets-unify.ai/ivy/contributing.html) for the [Ivy API](https://lets-unify.ai/ivy/contributing.html#ivy-api) and [Backend API](https://lets-unify.ai/ivy/contributing.html#backend-api). This includes removing all lambdas and direct bindings for methods in the backend submodules, with each method instead defined using . Full type hints should also be given for each argument in both the ivy module and also each backend module. The docstrings should be updated to use NumPy style rather than Sphinx style, and code examples should be added to every method docstring.

- [ ] get_referrers_recursive
- [ ] is_native_array
- [ ] is_ivy_array
- [ ] copy_array
- [ ] array_equal
- [ ] arrays_equal
- [ ] all_equal
- [ ] to_numpy
- [ ] to_scalar
- [ ] to_list
- [ ] clip_vector_norm
- [ ] clip_matrix_norm
- [ ] floormod
- [ ] unstack
- [ ] fourier_encode
- [ ] value_is_nan
- [ ] has_nans
- [ ] exists
- [ ] default
- [ ] shape_to_tuple
- [ ] try_else_none
- [ ] arg_names
- [ ] match_kwargs
- [ ] cache_fn
- [ ] current_framework_str
- [ ] einops_rearrange
- [ ] einops_reduce
- [ ] einops_repeat
- [ ] get_min_denominator
- [ ] set_min_denominator
- [ ] get_min_base
- [ ] set_min_base
- [ ] stable_divide
- [ ] stable_pow
- [ ] get_all_arrays_in_memory
- [ ] num_arrays_in_memory
- [ ] print_all_arrays_in_memory
- [ ] set_queue_timeout
- [ ] queue_timeout
- [ ] tmp_dir
- [ ] set_tmp_dir
- [ ] container_types
- [ ] inplace_arrays_supported
- [ ] inplace_variables_supported
- [ ] supports_inplace
- [ ] assert_supports_inplace
- [ ] inplace_update
- [ ] inplace_decrement
- [ ] inplace_increment
- [ ] cumsum
- [ ] cumprod
- [ ] scatter_flat
- [ ] scatter_nd
- [ ] gather
- [ ] gather_nd
- [ ] multiprocessing
- [ ] indices_where
[...]

PR #2083, opened 2022-07-20
Reformatted copy_array #1708

11Nick26:
Reformat general #791

Not sure if backend formatting was implemented correctly but i can revert them back if they are incorrect so please let me know.

Thank you!

### DIFF
--- a/ivy/functional/backends/jax/general.py
+++ b/ivy/functional/backends/jax/general.py
@@ -44,7 +44,7 @@ def is_native_array(x, exclusive=False):
     )
 
 
-def copy_array(x: JaxArray) -> ivy.Array:
+def copy_array(x: JaxArray) -> JaxArray:
     return jnp.array(x)
 
 
--- a/ivy/functional/backends/jax/general.py
+++ b/ivy/functional/backends/jax/general.py
@@ -44,7 +44,7 @@ def is_native_array(x, exclusive=False):
     )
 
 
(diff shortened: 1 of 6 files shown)


simonetgordon:
Hi Nick,

Thanks for the contribution!

This looks good for the ivy function. Did you intend to update the backends also? Only 1 file has been altered.

Simone

11Nick26:
I was going to update the backend functions but i realized i don't think i have to. But if that is necessary just let me know and I'll make the proper changes!

chie2727 (changes requested):
Hi @11Nick26 , please include the following additional examples:

- [x] 1. A `With :code:`ivy.Array` input:` example that makes use of the `out` parameter.

- [x] 2. A Container Static Method Example

Have a look at the docstring guide for reference:
https://lets-unify.ai/ivy/deep_dive/11_docstring_examples.html

chie2727 (changes requested):
Hi @11Nick26 ,

Thanks for making all the changes.
One last thing: please include an ivy.Container instance method example as there isn't an example for this yet.

chie2727:
Hi @11Nick26 , thanks for all your patience. I removed some duplicated examples and native_array examples which were added by a previous contributor as those aren't required anymore.
Everything is looking good so I'll merge in your changes now.

chie2727 (approved):

PR #2096, opened 2022-07-20
Expand dims jax inplace update

Oxbeefbabe:
Looking for help with test_expand_dims using the JAX backend. JAX doesn't support in-place updates, but using the ivy wrapper, it should still be able to do it, however, there are cases where it seems the JAX array itself has an in-place update, and this is causing an error. I can't figure out what is causing it. I've checked the inputs into helper.test_function that hypothesis is generating, but the same inputs that cause the error also seem to run successfully in other tests, unless there is something I am missing.


Oxbeefbabe:
Test Function in question is under Line 107 in ivy_tests/functional/test_mainpulation.py.

Oxbeefbabe:
Solved

PR #2101, opened 2022-07-20
Hypothesis for conv1d() and conv3d(), layers submodule.

zaeemansari70:


sherry30 (changes requested):
Hi Zaeem,

Can you also add unsupported dtype for these functions in torch backend as well.
torch does not support `float16` dtype for these functions.

Thanks

PR #2109, opened 2022-07-20
Device tests and refactor

Unevilicorn:
- More device tests
- Dev instance method missing argument
- Refactor old test to test on all available devices


Ishticode (approved):
lgtm :)

PR #2111, opened 2022-07-20
added activations under tensorflow

VikramxD:


jiahanxie353:
Frontend tasks are not open yet - but will be in a few days. I will review the PR once it's open. If you would rather get a faster response, then please create a PR for the reformatting task. Thanks! :)

VikramxD:
thanks for the quick reply on this issue i was watching the frontend guide @djl11 made but am still unable to figure out what to do to pass these tests like do i have to modify them or something according to the function we implement

PR #2113, opened 2022-07-20
adding dropout fixing lint

Abrar-Mustakim:


djl11:
Will open again once we have our PR extension pipeline well defined, thanks a lot for the contribution!

PR #2129, opened 2022-07-20
ci dashboard

jiahanxie353:
Finish all functional tests yaml files;
not sure about naming


djl11 (changes requested):
what's the rational behind the reformatting of the exist CI files, such as `"` -> `'`, indentation changes etc.?

I think we should use the name:

`test-<functional | stateful>-<submodule>-<backend>`

so for example:

`test-functional-elementwise-jax`
`test-stateful-activations-tf`
etc.

we will also remove all of the training tests, and so there is no need to add them into the CI

jiahanxie353:
For formatting rational, I believe `"` is the conventional way of industry; 
and parallel objects should have the same indentation but there is no restriction on how much exactly, it depends on the `tab` indentation in the yaml file of our IDEs I believe.

Sure, will do.

PR #2132, opened 2022-07-20
Adding max_pool2D solving some lint

Abrar-Mustakim:


Abrar-Mustakim:
Can you please tell me why I getting docstring test error? And why there is an assertion error

Issue #780, opened 2022-04-02
Reformat data_type


saeedashrraf:
Each method in the [data_type](https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/data_type.py) submodule should be modified to use the latest formatting rules, as explained in the [contributor guide](https://lets-unify.ai/ivy/contributing.html) for the [Ivy API](https://lets-unify.ai/ivy/contributing.html#ivy-api) and [Backend API](https://lets-unify.ai/ivy/contributing.html#backend-api). This includes removing all lambdas and direct bindings for methods in the backend submodules, with each method instead defined using `def method_name`. Full type hints should also be given for each argument in both the ivy module and also each backend module. The docstrings should be updated to use NumPy style rather than Sphinx style, and code examples should be added to **every** method docstring.

- [ ] astype
- [ ] broadcast_arrays
- [ ] broadcast_to
- [ ] can_cast 
- [ ] finfo
- [ ] iinfo
- [ ] result_type

Aarsh2001:
- [ ] #782

johnkagger:
#798

kimbring2:
- [ ] #1393

davidlee1102:
- [ ] https://github.com/unifyai/ivy/issues/1642

flych3r:
#1688

AlexandraDI:
- [ ] #1782

YangQiuEric:
- [ ] #1818

yjeong48:
- [ ] #1842

DawoodTahir:
set_float_default_dtype #1845

hirwa-nshuti:
valid_dtype #1850

hirwa-nshuti:
set_default_dtype #1900

davidlee1102:
- [ ] https://github.com/unifyai/ivy/issues/1934

rensortino:
- [] #2020

Neelaksh-Singh:
#2028

c1boo:
- [ ] #2041

PrachetShah:
- [  ] #2058 invalid_dtype

hirwa-nshuti:
broadcast_arrays #2168

hirwa-nshuti:
dtype_bits #2300

Amrabdelhamed611:
#2321

LateefAkinola:
astype #2020

Doregon499:
#2362

[...]

PR #2169, opened 2022-07-22
Added iinfo docstrings

Neelaksh-Singh:


Neelaksh-Singh:
Hi, do I need to implement class functionality for Iinfo in ivy.backends like the Finfo Class. For example in [numpy](https://github.com/unifyai/ivy/blob/e12a1c3bda728c24a24aab936a203fe1ab2c142e/ivy/functional/backends/numpy/data_type.py#L53).

Auxia:
Hi, you dont need to implement a class for iinfo. You only need to follow the task as it was outlined in #780. I see you have added thorough docstring examples and they are passing the tests. So this much should be fine. Thanks for Contributing! :)

PR #2170, opened 2022-07-22
Reformat invalid_dtype

PrachetShah:
Solves issue #2058 
Added Hints to the invalid_dtype function


rush2406 (changes requested):
Hi @PrachetShah
Thanks for contributing! Can you please remove `ivy_dev` folder and `.idea` folder, these should not be part of the PR.
Also can you please add a few examples for this function covering different cases. You can refer [here](https://github.com/unifyai/ivy/blob/3b520bf6618db5b0913499d0d9e29a63166ae5d9/ivy/functional/ivy/data_type.py#L1241) for how to write the examples. Make sure to cover examples for `ivy.Dtype`, str and None as these are the types for the arguments.

PrachetShah:
Yes thanks for review, i will work on it and re-request a review for it

PrachetShah:
I have added some examples, pls check if they are right or not. Also i removed ivy_dev and .idea folder which got added by mistake

rush2406 (changes requested):
@PrachetShah the `.idea` is still present, could you check and remove it?

PrachetShah:
`.idea` folder is there in repository, do i still have to remove it?

rush2406 (changes requested):
so the problem is that there's a difference between the `.idea` folder in your branch and the master branch of `ivy`. Can you try pulling the master branch changes? Eventually the `Files changed` tab in this PR should only show 1 file instead of 87 files as shown currently.

PrachetShah:
So i checked on github docs, i guess there is an bug here which does not change the files changed once a second commit is made or something like that, it is telling me to close the pr and re-open it. Can i do that?

PR #2177, opened 2022-07-22
Gradients

vedpatwardhan:


ivy/functional/backends/torch/gradients.py
```diff
@@ -22,7 +24,9 @@ def variable_data(x):
 
 
 # noinspection PyShadowingNames
+# @to_container_and_back
 def execute_with_gradients(func, xs, retain_grads=False):
+    # print('xs', xs)
```
djl11, line 29: debug code

ivy/functional/backends/torch/gradients.py
```diff
@@ -22,7 +24,9 @@ def variable_data(x):
 
 
 # noinspection PyShadowingNames
+# @to_container_and_back
```
djl11, line 27: debug code

ivy/functional/backends/torch/gradients.py
```diff
@@ -31,15 +35,25 @@ def execute_with_gradients(func, xs, retain_grads=False):
...
         rest = tuple()
     y = ivy.to_native(y)
-    x_grads_flat = list(
-        torch.autograd.grad(
-            [y],
-            [v for k, v in xs.to_iterator()],
+    if isinstance(xs, ivy.Container):
+        x_grads_flat = [
+            torch.autograd.grad(
+                y[key],
```
djl11, line 41: we should assume that `y` is always single array, and never a container, you can use the original code unmodified in cases where `xs` is a container for all backends.

djl11 (changes requested):

djl11 (approved):
lgtm! Provided that the unit tests are all passing, please feel free to merge.

PR #2218, opened 2022-07-22
Nest submodule Fix Up 

MarShaikh:
Had a question related to type hints in the nest submodule, since nests consume iterables, should we just keep the type hint as `Iterable` or be specific and mention all the Iterables(Lists, dicts, tuples, the like). I noticed someone made changes and added them, but I thought it was better for me to confirm. Thanks!


djl11 (approved):
lgtm!

PR #2226, opened 2022-07-23
updating eigvalsh in linear algebra

Osamaelgendy:


ivy/functional/ivy/linear_algebra.py
```diff
@@ -601,6 +601,80 @@ def eigvalsh(
     ret
         an array containing the computed eigenvalues. The returned array must have shape
         (..., M) and have the same data type as x.
+    
+    This function conforms to the `Array API Standard
+    Both the description and the type hints above assumes an array input for simplicity,
+    but this function is *nestable*, and therefore also accepts :code:`ivy.Container`
+    instances in place of any of the arguments.
```
jiahanxie353, line 608: Your explanation is correct but you don't need to add it here

> but this function is *nestable*, and therefore also accepts :code:`ivy.Container`
    instances in place of any of the arguments

because you are working under `ivy/functional/ivy`. We will have `ivy/container/linear_algebra` for your point.

jiahanxie353:
Please also fix the docstring tests failures for `eigvalsh` :)

ivy/functional/ivy/linear_algebra.py
Osamaelgendy: Thank you, do I have to modify something?

Osamaelgendy: can you clarify the error with docstrings more please?

Osamaelgendy: I updated modified the description, but the error I found in the docstrings file says it's unable to find the docker image even though I ran the command given in the instructions, it's the first line in the test-docstrings details

jiahanxie353: > I updated modified the description, but the error I found in the docstrings file says it's unable to find the docker image even though I ran the command given in the instructions, it's the first line in the test-docstrings details

GitHub unable to find the ivy docker image locally, so then it starts to pull from the latest ivy image. It's how docker works, you can check docker's documentation for more information.
The error related to `eigvalsh` are checked below the workflow logs, such as line 158 which has an `E` alongside `eigvalsh`. You need to fix every `E` caused by `eigvalsh`.

Osamaelgendy: I'm on it

Osamaelgendy: I'm sorry I can't figure it out, I use a format for docstrings looking like another function that was already merged. I can't spot the problem

jiahanxie353: > I'm sorry I can't figure it out, I use a format for docstrings looking like another function that was already merged. I can't spot the problem

No worries at all! I didn't check the failure message in detail previously and now I just did. The failure is caused by `"linalg_eigh_cpu" not implemented for 'Int'`. So when the docstring tests are testing your inputs in the torch backend setting, your integer inputs are failing as shown above.
I should have checked the failure message more carefully last time, apologize for the potential time lag.

Osamaelgendy: No worries, I will check that
thanks a lot

Osamaelgendy: No worries, so should I modify the integers I wrote or it will be merged?
according to the docs it's better to be easy examples I think

jiahanxie353: Sorry this message somehow slipped out of the radar.
I can't tell you what you should do exactly, but you can check the [Deep Dive Docstrings Example](https://lets-unify.ai/ivy/deep_dive/13_docstring_examples.html). 
Basically, in terms of numeric numbers, you can keep it easy but not uniformly so other users can understand more quickly. For example, if you use `[1, 2, 3]` in an instance method example, you could try `[4, 5, 6]` in a static method example. In the mean time, the docstrings examples shouldn't be easy in terms of variations such as shape, `out` argument, type, etc. More details are discussed in [Deep Dive Docstrings Example](https://lets-unify.ai/ivy/deep_dive/13_docstring_examples.html)

Osamaelgendy: Yes I understand, but they still "integers" and maybe it will give the same failure "linalg_eigh_cpu" not implemented for 'Int', that's why I'm asking for the next step

jiahanxie353: > Yes I understand, but they still "integers" and maybe it will give the same failure "linalg_eigh_cpu" not implemented for 'Int', that's why I'm asking for the next step

Yeah I think what you can do is in the `ivy.array` example, you can change the example to    
```
x = ivy.array([[1.0, 2.0, -1.0],
               [2.0, 1.0, 3.0],
               [-1.0, 3.0, 0.0]])
```
But in the `ivy.native_array` example, please do not use the same input. You could change the float values and vary the shape a little bit.

Osamaelgendy: I'm on it

Osamaelgendy: I solved some issues in docstrings but I can't figure out others

Osamaelgendy: I think the problem is related to the results itself, but I calculated it again and they are right

Osamaelgendy: Is there any updates?

Osamaelgendy: is there any updates regarding my commit because I fixed some docstrigns, but i can't figure out the rest @JLU-Neal

ivy-seed:

ivy-seed:

PR #2245, opened 2022-07-23
add configurable test arguments

Aarsh2001:
I have added the decorator, and it performs as expected. Wanted some clarification on the following points -:

1.  name of the ```decorator```
2. which file should it go to, ```helpers.py``` or ```func_wrapper.py```
3. As per the current design,  every ```given``` block will be required to have the ```st.data()``` object, is this fine?
4.  is there anything else expected inside


djl11 (approved):
looks good - the next step is to integrate this into the current tests, by adding a decorator like this to all existing unit tests!

djl11:
closing as this is a dummy PR with some debug code and test files, not the final PR ready for incorporating into master.

PR #2250, opened 2022-07-23
configurable argument testing 

Aarsh2001:


Aarsh2001:
added ```--no-extra-testing``` flag as well. Some of the submodule functions draw multiple values for ```as_variable``` , ```native_array```,  and ```container``` parameters. For these functions I have created a list which duplicates these boolean flags received from the ```handle_cmd_line_args``` decorator, at the moment. Let me know if there are any other changes required, thanks🙂.

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -23,12 +24,13 @@
         max_dim_size=10,
         shared_dtype=True,
     ),
-    as_variable=st.booleans(),
```
djl11, line 26: we don't want to remove the arguments which are already here, particularly because many of them use `helpers.array_bools` to generate independent flags for each input array. The current approach would replace all of these with `st.booleans` which we do not want. We should receive the existing values as input to `handle_cmd_line_args`, and return the original if nothing has been set at the command line.

djl11 (changes requested):

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
Aarsh2001: Should we keep only the one's which use array bools or should we keep everything?

Aarsh2001:
Some pointers -:

1. For functions which use multiple flags a list which duplicates these boolean flags is being used, so this needs to be kept in mind(maybe a new task).
2. Some submodules haven't bee fixed yet, so the guide for the future would be to include ```data = st.data()``` and the test function should look something like this -:
```
def test_(
     *,
     data,
     **kwargs)
```
3. I noticed some functions aren't drawing values for the parameters ```as_variable```, ```container```, ```with_out```, and ```instance_method``` in the ```given``` decorator. Rather they're being passed directly inside the ```helpers.test_function```. Should this be fixed as well?

ivy_tests/test_ivy/helpers.py
```diff
@@ -1901,3 +1901,53 @@ def num_positional_args(draw, *, fn_name: str = None):
...
+
+def handle_cmd_line_args(test_fn):
+    # first four arguments are all fixtures
+    def new_fn(get_command_line_flags, fw, device, call, *args, **kwargs):
+        # inspecting for keyword arguments in test function
+        for param in inspect.signature(test_fn).parameters.values():
+            if param.kind == param.KEYWORD_ONLY:
+                if param.name == "data":
+                    data = kwargs["data"]
+                if param.name == "as_variable":
```
djl11, line 1920: from here below should all be `elif`, this will speed things up considerably (currently, we keep checking everything even after we have found a match.)

ivy_tests/test_ivy/helpers.py
```diff
@@ -1901,3 +1901,53 @@ def num_positional_args(draw, *, fn_name: str = None):
...
+
+def bool_val_flags(cl_arg: Union[bool, None]):
+    if cl_arg is not None:
+        return st.booleans().filter(lambda x: x == cl_arg)
+    return array_bools(num_arrays=1)
+
+
+def handle_cmd_line_args(test_fn):
+    # first four arguments are all fixtures
+    def new_fn(get_command_line_flags, fw, device, call, *args, **kwargs):
```
djl11, line 1914: this design is wrong overall, we shouldn't overwrite anything unless the flag is set. We are currently overwriting with `array_bools(num_arrays=1)` if the flag is not set, but we should just leave the arguments unmodified if a flag has not been set.

djl11 (changes requested):

djl11:
> Some pointers -:
> 
> 1. As of now a quick fix for the `array_bools` causing the tests to fail after it's integration inside the `bool_val_flags` is,  to set `num_arrays = 1`. The decorator (`handle_cmd_line_args`) at present will only handle parameters which require 1 boolean flag. For functions which require more than one flag, refer to point 2.
> 2. For functions which use multiple flags a list which duplicates these boolean flags is being used, so this needs to be kept in mind(maybe a new task).
> 3. Some submodules haven't bee fixed yet, so the guide for the future would be to include `data = st.data()` and the test function should look something like this -:
> 
> ```
> def test_(
>      *,
>      data,
>      **kwargs)
> ```
> 
> 4. I noticed some functions aren't drawing values for the parameters `as_variable`, `container`, `with_out`, and `instance_method` in the `given` decorator. Rather they're being passed directly inside the `helpers.test_function`. Should this be fixed as well?
> 5. The following submodules haven't been integrated with the `handle_cmd_line_args` decorator-:
> 
> * test_dtype
> * test_general
> * test_gradients
> * test_meta
> * test_nest
> * stateful tests

1. See comments, we shouldn't change anything unless the command line flag is set
4. No for these we should keep them hard-coded, if the value is hard-coded it means the variation does not make sense for that test
5. What's the reason for this?

djl11 (approved):
lgtm!

Issue #1888, opened 2022-07-12
min


AliYagizDizdaroglu:

AliYagizDizdaroglu:
min #1473

AliYagizDizdaroglu:
https://github.com/unifyai/ivy/issues/812#issue-1191480912

AliYagizDizdaroglu:
asd

deanyang123:
- [min] #2776

PR #2257, opened 2022-07-24
update the min function type hints, docstrings and docstring example

Pirunga:


Pirunga:
#1888

sherry30 (changes requested):
Hi,

The changes look good, just a few more things need to be added

1. static function for container instance method and its example.
2. example for container instance method

If you have any questions then feel free to ask, I'd be happy to help you out with anything.

Thanks.

sherry30 (changes requested):
Hi,

A few changes are required:

1. You need to add the instance method examples in their appropriate files, so for array instance method example, you have to add it in `ivy/array/statistical.py` and for container instance method it should go in `ivy/container/statistical.py`.
2. You have to implement static function in container file as well and an example of that.

For all of these, you can checkout `tanh` in elementwise for reference.
you can read all the detailed requirements [here](https://github.com/unifyai/ivy/blob/48069d3a774e279eec2a4fb85b7ace69a8925291/docs/partial_source/deep_dive/13_docstring_examples.rst)

Let me know if you have any more questions.

Thanks.

ivy/container/statistical.py
```diff
@@ -10,17 +10,130 @@
...
+            a container containing the minimun value of array  ``x``. The
+            returned container must have the same data type as ``x``.
+
+        Examples
+        --------
+        With one :code:`ivy.Container` input:
+
+        >>> x = ivy.Container(a=ivy.array([0., 2.6, 10]),\
+                            b=ivy.array([4.5, -7, 0, 9]))
+        >>> y = ivy.Container.min(x)
```
sherry30, line 130: this should be called as a container instance method,
`y = x.min()`

ivy/functional/ivy/statistical.py
```diff
@@ -225,6 +225,48 @@ def min(
...
+    ivy.array([[0, 1, 2]])
+
+    >>> x = ivy.native_array([[0, 1, 2], [3, 7, 15]])
+    >>> y = ivy.min(x)
+    >>> print(y)
+    ivy.array(0)
+
+    With :code:`ivy.Container` input:
+
+    >>> x = ivy.Container(a=ivy.array([7., 77., 777.]), b=ivy.array([8., 88., 888.]))
```
sherry30, line 255: indent array `b` on the next line

sherry30 (changes requested):
i requested 2 changes on specific lines,

other than that you don't have to add 2 examples of container in the functional examples, please remove the one being used as instance method.

sherry30:
Hi,

everything looks good, I think the only thing left to do is add examples for the array instance method in `ivy/array/statistical.py`.
Also there seem to be some lint errors in `ivy/container/statistical.py`, can you try fixing those.

Thanks

PR #2262, opened 2022-07-24
Reformat logical_not

### LINKED ISSUE #2231: logical_not

vamsivallepu:
Reformatted logical_not function 
resolves #2231

### DIFF
--- a/ivy/functional/ivy/elementwise.py
+++ b/ivy/functional/ivy/elementwise.py
@@ -2906,6 +2906,15 @@ def logical_not(
        accept input arrays having numeric data types. If non-boolean data types are
        supported, zeros must be considered the equivalent of ``False``, while non-zeros
        must be considered the equivalent of ``True``.
+       
+       **Special cases**
+       
+       For this particular case,
+       
+       - If ``x_i`` is ``NaN``, the result is ``False``.
+       - If ``x_i`` is ``-0``, the result is ``True``.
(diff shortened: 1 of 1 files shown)


Unevilicorn:
Hi @vamsivallepu,
The example you have provided does not cover all the variation of the input, you can read more about it [here](https://lets-unify.ai/ivy/deep_dive/12_docstring_examples.html).

vamsivallepu:
Hi @Unevilicorn 
Thanks for reviewing. I have gone through the link you have provided, and I think the only variation I have missed is ivy.Container. But as per the function arguments, it only accepts either ivy.Array or ivy.NativeArray. So does that mean I need to alter the function arguments too?

Unevilicorn:
You don't need to alter the argument type, since this is all handled by the `@handle_nestable` decorator, you can read more about it [here](https://lets-unify.ai/ivy/deep_dive/5_containers.html#nestable-functions).

vamsivallepu:
While working on examples of instance methods, I noticed that logical_not() is not working with ivy.NativeArray type as an instance method. Is it a bug, or do I have to mention, "this function doesn't work with the NativeArray type as an instance method"?

vamsivallepu:
@Unevilicorn, I hope I have made the necessary changes. Can you please review my latest commit?

Unevilicorn:
Looks good and I've merged this PR :slightly_smiling_face:

PR #2274, opened 2022-07-25
Reformat element-wise function isnan()

abekek:
#2217


sreegandh804:
Hey @abekek , i would suggest you try and get the numpy docstrings passing , after which you could let me know.

abekek:
@sreegandh804 , thanks for the review. I think numpy-docstrings is passing for this function, but it fails on other functions like `reshape` and `array_and_indicies`. Is this an expected behavior?

sreegandh804:
Hi @abekek , 
As long as the tests failing are not caused due to the changes you made(which doesn't seem to be the case here) it's fine.
I'll get this PR merged right away

abekek:
@sreegandh804, great, thank you for your help!

PR #2275, opened 2022-07-25
Fixed issue with test_function helper

vedpatwardhan:


ivy_tests/test_ivy/helpers.py
```diff
@@ -980,6 +980,7 @@ def test_function(
         if not max(container_flags) and ivy.native_inplace_support:
             # these backends do not always support native inplace updates
             assert ret.data is out.data
+    ivy.unset_backend()
```
djl11, line 983: this doesn't look right, why are we unsetting the backend here, where has it been set? It looks like the backend at this point will be the backend we are testing for, which should not be unset.

djl11 (changes requested):

djl11 (changes requested):
looks good, apart from the removed docstrings. Maybe you need to pull from upstream first before merging?

djl11 (approved):
lgtm!

PR #2279, opened 2022-07-25
getting_most_out_of_IDE

RitujaPawas:


djl11 (changes requested):
Looks great! Could you please add all images [here](https://github.com/unifyai/unifyai.github.io/tree/master/img/externally_linked) in a separate `getting_most_out_of_ide` folder instead? You will then need to reference this external URL in the .rst file. This is how the other images are all added in the docs :)

djl11 (approved):
lgtm!

PR #2280, opened 2022-07-25
BitwiseOr to Tensorflow frontend as RawOps Operations

kimbring2:
I need to modify the helpers.py file a little bit because only Tensorflow needs to call the BitwiseOr from the bitwise scope such as **tf.bitwise.bitwise_or**.

https://github.com/unifyai/ivy/issues/2278

### DIFF
--- a/ivy/functional/frontends/tensorflow/functions.py
+++ b/ivy/functional/frontends/tensorflow/functions.py
@@ -16,5 +16,16 @@ def tan(x, name=None):
 tan.unsupported_dtypes = {"torch": ("float16", "bfloat16")}
 
 
+def cos(x, name=None):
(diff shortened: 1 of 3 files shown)


RashulChutani (commented):
Changing the Helpers Module this way is not Ideal. I will get back to you on this ASAP.

djl11 (changes requested):
please look at the other examples in the frontend APIs, and read through the deep dive sections in more detail, and then come back to this PR afterwards. We should not be modifying anything in the helpers. Thanks!

kimbring2:
@RashulChutani @djl11 Thank you for the comment. I will find another way!

PR #2289, opened 2022-07-25
random_uniform fix

juliagsy:
- enabled bounds of type numeric only, arrays only, or a mixture of both
- updated array and container instance methods
- updated docstrings and examples
- removed @infer_dtype 
- add dtype arg where related and handling code (ivy and backends)
- updated hypothesis tests


djl11 (changes requested):
all looks good, but I was wrong when I said we should not use the @infer_dtype decorator. I mistakenly thought that `low`, `high`, `mean`, `std` etc. could be relevant scalars as explained [here](https://github.com/unifyai/ivy/blob/ae6afbb673bd08da9c3472b4843c7c29bd3362bb/docs/partial_source/deep_dive/6_data_types.rst#arguments-in-other-functions). However, that's not the case. There are no random functions which can operate with both `int` and `float` types, and so we cannot use the type of these scalars to infer the data type we should use in the function. You should add the `@infer_dtype` back and add the non-optional `dtype` argument in the backend implementations :)

PR #2290, opened 2022-07-25
Pytest docstring changes

Oxbeefbabe:
Added a section to the Contributing guide in the docs on how to get more detailed Hypothesis logs while using Pycharm.


djl11 (changes requested):
a couple of things:

a) it looks like you have inadvertently added all of your local changes, you should only add the `docs` folder

b) you should add the images [here](https://github.com/unifyai/unifyai.github.io/tree/master/img/externally_linked), perhaps in a folder called `setting_up` and then reference this external URL in the `.rst` file, in order to prevent the memory of this repo growing large with the inclusion of images

Other than that, looks good!

Oxbeefbabe:
@djl11 My bad about the extra added changes, I thought I had unselected them in Pycharm but that doesn't seem to be the case. I've removed the changes, so now it's just regarding the docs. I've also moved the images to the other repo as requested. Before accepting do you mind quickly checking to see if it looks like I've linked them correctly? I don't have any way to render .rst files on my PC so I can't check if they've been incorrectly linked. Cheers :)!

PR #2291, opened 2022-07-25
Device profiler refactor + fix

Unevilicorn:
Extended the profiler test
Added missing type hint and doc-string for profiler


Ishticode (approved):
lgtm! But perhaps deleting the logs at the end of the **test** again (as was done after the with block) would be ideal as you wouldn't want logs directory leftover after every time after this test is run.

EDIT: Do you think the test so far only checks for start method really. It doesn't seem that the test is testing the stop method. Any code executed after stop should not be logging and we could perhaps test that the logs should be unchanged by it. Sorry for adding this in approve box I think because these are additions not changes in you code really.

djl11 (changes requested):
@Ishticode makes a good point, we should make sure the log files and folders are all removed at the end of the test (even if an exception was raised during the test), and we should also make sure that things do not continue to be logged after `stop` has been called :)

djl11 (approved):
lgtm!

PR #2306, opened 2022-07-26
random_normal fix

juliagsy:
- enabled bounds of type numeric only, arrays only, or a mixture of both
- updated array and container instance methods
- updated docstrings and examples
- add dtype arg and @infer_dtype 
- updated hypothesis tests


ivy/functional/ivy/random.py
```diff
@@ -344,41 +357,220 @@ def random_normal(
...
+
+    >>> ivy.random_normal(x, y, device='cpu', dtype='float64')
+    ivy.array([-4.1   , -0.0366])
+
+    >>> z = ivy.zeros((2,))
+    >>> ivy.random_normal(x, y, device='cpu', dtype='float64', out=z)
+    ivy.array([12.4, 11. ])
 
     Instance Method Examples
     ------------------------
```
djl11, line 416: don't need these examples in the functional API docstring, check [Docstring Examples](https://github.com/unifyai/ivy/blob/517e457514defdb36af1eef135c5fc8e20470c5f/docs/partial_source/deep_dive/13_docstring_examples.rst) section.

djl11 (changes requested):
small change on the docstring examples, otherwise looks good!

PR #2309, opened 2022-07-26
`meta` submodule fix up

juliagsy:
I've been making changes through PR but here's a final one just in case! Thanks!


djl11 (approved):
lgtm!

Issue #858, opened 2022-04-09
unique_inverse


LINBOTANG123:

LINBOTANG123:
No unit test for set in "\IVY\ivy\ivy_tests\test_ivy\test_functional\test_core". How can I test by using assert_docstring_examples_run? Thanks for your help

Archymade:
- [ ] #858

PR #2329, opened 2022-07-26
unique_inverse #858 #2271

EngrE:
Docstrings with examples


MarShaikh (changes requested):
Your examples look great. Could you also add array instance and container instance methods as well? Thanks!

EngrE:
Hi, I just added array instance and container instance methods docstrings.
Thanks

EngrE:
Please review the new changes thanks

MarShaikh (changes requested):
Please add correct examples for container and array instance methods. These are incorrect. Please refer to the documentation to see what the correct example looks like.

EngrE:
> 

ok Sure i will update it soon

EngrE:
Hi, 
I have followed the tan() function example mentioned in documentation as well as in the code and made the required changes. I hope it is now up to the mark. 
Thanks

ivy/array/set.py
```diff
@@ -50,4 +50,39 @@ def unique_all(
...
+              - second element must have the field name ``inverse_indices`` and must be an
+              array containing the indices of ``values`` that reconstruct ``x``. The array
+              must have the same shape as ``x`` and must have the default array index data
+              type.
+
+          Examples
+          --------
+
+        >>> x = ivy.array([0.3,0.4,0.7,0.4,0.2,0.8,0.5])
+        >>> y = ivy.unique_inverse(x)
```
MarShaikh, line 82: This way of calling the instance method is wrong! Please see your correct invocation of instance method in `ivy/container/set.py`!

MarShaikh (changes requested):
Thanks for the changes, just requested a small correction, and you should be good to go.

ivy/array/set.py
EngrE: Ok thanks a lot. changed as requested by you.. just for info, i was getting same result when i called the method wrong way :P

MarShaikh: Yes, you would get the correct result because this way of calling the function is a functional call, whereas an instance method references the class of the variable itself.

MarShaikh (changes requested):
I can see some issues with the formatting that is causing an error in the CI. You can setup pre-commit to fix these errors easily.

EngrE:
> I can see some issues with the formatting that is causing an error in the CI. You can setup pre-commit to fix these errors easily.

ok sure let me pre-commit

EngrE:
Thanks, I have setup pre commit and pushed the changes

MarShaikh (approved):
sorry for the delayed response. Everything looks great.

EngrE:
> 

thats great

PR #2337, opened 2022-07-27
Documentation

RickSanchezStoic:
Adding documentation for workflows in pr


djl11 (changes requested):
please add all of the images [here](https://github.com/unifyai/unifyai.github.io/tree/master/img/externally_linked), in a new folder called "the basics" and reference this external URL rather than adding images to the repo. This makes it much faster to clone. Otherwise looks good, thanks!

PR #2338, opened 2022-07-27
device fix to allow empty string (also for ivy gym)

juliagsy:
- added int check for array in shape (torch allows this)
- changed empty device in Jax from `None` to `''`
- assert `cpu`, `gpu` or `cuda` only when device is not empty


djl11 (approved):
lgtm!

PR #2340, opened 2022-07-27
Reformat cross 

Ajay6601:
Reformated linear algebra cross function.


Ajay6601:
I have few questions?

PR #2364, opened 2022-07-28
using `assume` in ivy_tests

rush2406:
replaced `return` for unsupported cases with `assume`


rush2406:
In some tests, there are checks for the backend call such as [here](https://github.com/unifyai/ivy/blob/fb0a54a7ddbe80cf85033366fac25f7a7bf2b360/ivy_tests/test_ivy/test_functional/test_core/test_general.py#L138). Are these also to be replaced with `assume`?

ivy_tests/test_ivy/test_functional/test_core/test_device.py
```diff
@@ -257,9 +255,8 @@ def test_to_device(
         assert ivy.dev(x_on_dev, as_native=True) == ivy.dev(out, as_native=True)
 
         # check if native arrays are the same
-        if ivy.current_backend_str() in ["tensorflow", "jax"]:
-            # these backends do not support native inplace updates
-            return
+        # these backends do not support native inplace updates
+        assume(not (ivy.current_backend_str() in ["tensorflow", "jax"]))
```
djl11, line 259: let's use `fw` rather than `ivy.current_backend_str`, the same goes for all other tests.

ivy_tests/test_ivy/test_functional/test_core/test_dtype.py
```diff
@@ -489,8 +488,8 @@ def test_as_native_dtype(
 ):
     res = ivy.as_native_dtype(input_dtype)
     if isinstance(input_dtype, ivy.NativeDtype):
-        assert isinstance(res, ivy.NativeDtype)
-        return
+        assume(isinstance(res, ivy.NativeDtype))
```
djl11, line 491: we should not be removing any `assert` statements, these are added for a reason, and the test should fail if these are not met.

djl11 (changes requested):
a couple of requested changes, I haven't gone through all of it. Could you please make these changes in all files and the re-request a review? Thanks!

djl11:
> * In some tests, there are checks for the backend call such as [here](https://github.com/unifyai/ivy/blob/fb0a54a7ddbe80cf85033366fac25f7a7bf2b360/ivy_tests/test_ivy/test_functional/test_core/test_general.py#L138). Are these also to be replaced with `assume`?
> * There are a few functions which still have pytest implementations so I haven't made changes in [these](https://github.com/unifyai/ivy/blob/fb0a54a7ddbe80cf85033366fac25f7a7bf2b360/ivy_tests/test_ivy/test_functional/test_core/test_meta.py#L492) currently . Should I change for them as well?

Yes these backend checks should also be replaced with `assume`, we are in the process of changing the backend selection to be governed by hypothesis, which will then make these `assume` statements work correctly.

Yes you are correct, there is no need to add `assume` into the `pytest` implementations

ivy_tests/test_ivy/test_functional/test_core/test_device.py
```diff
@@ -173,9 +173,8 @@ def test_as_native_dev(*, array_shape, dtype, as_variable, fw, call):
         else:
             assert ret == device
         # compilation test
-        if call is helpers.torch_call:
-            # pytorch scripting does not handle converting string to device
-            return
+        # pytorch scripting does not handle converting string to device
+        assume(not (call is helpers.torch_call))
```
djl11, line 177: we should replace all `call is helpers.<name>_call` logic with `fw == "backend_name"` logic, both do exactly the same thing.

djl11 (changes requested):

djl11 (approved):
lgtm!

PR #2375, opened 2022-07-28
Cmd line args for testing gradients

VedPatwardhan:


ivy_tests/test_ivy/helpers.py
```diff
@@ -2195,6 +2195,11 @@ def new_fn(data, get_command_line_flags, fw, device, call, *args, **kwargs):
                         bool_val_flags(get_command_line_flags["nestable"])
                     )
                     kwargs["container"] = container
+                elif param.name == "test_gradients":
+                    test_gradients = data.draw(
+                        bool_val_flags(get_command_line_flags["test_gradients"])
+                    )
+                    kwargs["test_gradients"] = test_gradients
```
Aarsh2001, line 2202: So this decorator will only handle test function parameters, and in this case once the decorator is applied to every test function it will return a new keyword argument `test_gradients` to every function

Aarsh2001 (changes requested):

Aarsh2001:
looks good !

PR #2383, opened 2022-07-28
Update statistical_operations.py in numpy front end

Doregon499:
creation of the dtype_to_str function


Doregon499:
isue  std #2465 and  dtype_to_str #2362

PR #2402, opened 2022-07-29
NumPy Gradients Docstring tests

johanjino:


ivy/functional/ivy/gradients.py
```diff
@@ -325,6 +325,7 @@ def execute_with_gradients(func, xs, retain_grads=False):
     """
     return current_backend(None).execute_with_gradients(func, xs, retain_grads)
 
+setattr(execute_with_gradients, "computes_gradients", True)
```
djl11, line 328: `execute_with_gradients.computes_gradients = True` is cleaner, same in other locations across the code.

ivy_tests/test_docstrings.py
```diff
@@ -71,7 +71,13 @@ def test_docstrings(backend):
         if k == "Array":
             for method_name in dir(v):
                 method = getattr(ivy.Array, method_name)
-                if method_name in skip_arr_cont or helpers.docstring_examples_run(
+                try:
```
djl11, line 74: you have only added this for `ivy.Array` instance method tests, we should also add it for `ivy.Container` instance method tests and functional tests below.

ivy_tests/test_docstrings.py
```diff
@@ -71,7 +71,13 @@ def test_docstrings(backend):
...
+                    gradient_incompatible = method.computes_gradients and not backend.supports_gradients
+                except AttributeError:
```
djl11, line 76: we shouldn't use `try` `except` for this, we can do something like `gradient_incompatible = not backend.supports_gradients and hasattr(method, "computes_gradients") and method.computes_gradients`

we always evaluate expressions in `and`, `or` etc. from left to right, and this stops as soon as an evaluation leads to `False`, meaning that `method.computes_gradients` will not be evaluated if `hasattr(method, "computes_gradients")` returns `False`.

But actually, we should implement a simple dedicated helper function for this called `gradient_incompatible_function`, in a similar vein to [function_unsupported_dtypes](https://github.com/unifyai/ivy/blob/3dda99b1ed295e0ac67de7c4170dd041c23b78bc/ivy/functional/ivy/data_type.py#L1134) etc.

This helper function implementation can be as simple as:

```
return not ivy.supports_gradients and hasattr(fn_in, "computes_gradients") and fn_in.computes_gradients
```

djl11 (changes requested):

johanjino:
Hey @djl11 

I have made the changes. Thank you so much for pointing these out, I can definitely write more cleaner codes now 😄

djl11 (approved):
lgtm!

PR #2409, opened 2022-07-29
Replace legacy ivy method clip

JLU-Neal:


ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py
```diff
@@ -797,7 +797,7 @@ def test_swapaxes(
 @given(
     x_min_n_max=helpers.dtype_and_values(
         available_dtypes=ivy_np.valid_numeric_dtypes, num_arrays=3
-    ),
+    ).filter(lambda x_min_n_max: x_min_n_max[0][0] == x_min_n_max[0][1] and x_min_n_max[0][1] == x_min_n_max[0][2]),
```
djl11, line 800: does this logic work correctly? you are performing the `==` operator on two lists, do we only care about checking for total equality between the lists? I would have thought elementwise equality is more important

djl11 (changes requested):

ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py
JLU-Neal: Actually here x_min_n_max is an object of class ivy.FloatDtype, and I want to ensure all the types are the same since I observed some overflow when dtypes are different.

ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py
```diff
@@ -795,9 +795,10 @@ def test_swapaxes(
 
 # clip
 @given(
+    # use filter to ensure dtypes are all the same
     x_min_n_max=helpers.dtype_and_values(
         available_dtypes=ivy_np.valid_numeric_dtypes, num_arrays=3
-    ),
+    ).filter(lambda x_min_n_max: all(x == x_min_n_max[0][0] for x in x_min_n_max[0])),
```
djl11, line 801: if we want the `dtype` to all be the same, then you should just pass `shared_dtype=True` into `dtype_and_values`.

djl11 (changes requested):

ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py
JLU-Neal: Yeah, you are right! 
The filter is superfluous :D

PR #2411, opened 2022-07-29
Docs Glossary

jkeane508:
Creating a Glossary for Ivy Docs using sphinx glossary directive


jkeane508:
Hi,

Currently cannot get the Glossary directive to work perhaps you could weigh in?

Thanks!

VedPatwardhan (changes requested):
Hii @jkeane508, you just need to add the path to this file at the end of the `index_append.rst` file, then it would be visible in the table of contents. Thanks 🙂

jkeane508:
Thanks! @VedPatwardhan 

I was wondering if you could figure out why the glossary directive wasn't being processed? It just ignores it when viewing it on GitHub.

VedPatwardhan (commented):
But the PR hasn't been merged yet so I don't think we can know if the glossary directive works with GitHub or not. Let's see if it works after merging, and if there's an issue we'd work together on fixing it if that's okay?
Thanks 🙂

jkeane508:
@VedPatwardhan,

Sounds good!
 
I'll get @djl11 to skim over before pushing to master.

djl11 (approved):
looks great! Thanks James :)

PR #2414, opened 2022-07-29
jacobian gradient submodule added

modanesh:


VedPatwardhan (changes requested):
Hi @modanesh, I tried to execute an example with the `ivy.jac` function, but it seems like the input to this function always needs to be a JAX Array, because the `jax.jaxfwd` function is being used internally which does not allow the function passed to it to return an `ivy.Array`, so I think the best way would be to also wrap the `jax.jacfwd` and the `jax.jacrev` functions inside `ivy`.

VedPatwardhan (approved):
Great! I think we can merge it 👍

VedPatwardhan (approved):
lgtm!

PR #2417, opened 2022-07-29
Fix: remove legacy function vector_norm in base.py, and add docstring…

JLU-Neal:
… in linear_algebra.py


ivy/container/linear_algebra.py
```diff
@@ -1084,6 +1119,39 @@ def vector_norm(
         *,
         out: Optional[ivy.Container] = None,
     ) -> ivy.Container:
+        """Compute vector p-norm for each array in the container.
```
djl11, line 1122: these docstrings are wrong, please re-read the **classes** section [here](https://github.com/unifyai/ivy/blob/e7b3211dbe1d24ff651155d7899546a7c4dc30f7/docs/partial_source/deep_dive/12_docstrings.rst)

djl11 (changes requested):

ivy/container/linear_algebra.py
JLU-Neal: Yeah, you are right! I copied the legacy docstring in base.py.
I have updated it now :D

djl11 (changes requested):
these docstrings are still wrong, again, please re-read the **classes** section [here](https://github.com/unifyai/ivy/blob/e7b3211dbe1d24ff651155d7899546a7c4dc30f7/docs/partial_source/deep_dive/12_docstrings.rst). We defer to the functional docstring, avoiding repeating ourselves in multiple identical docstrings.

PR #2436, opened 2022-07-29
Subtract - Mathematical Function - Numpy - Frontend

### LINKED ISSUE #2435: subtract

TheMonocledHamster:
Closes #2435 

Possibly needs to be refactored to make the code more elegant? 
For now, I've stuck to the same format as that of the add function.


ZoeCD:
Hi! Thank you for your contribution! We are waiting a little longer before refactoring, so we can see what would be the most general and useful solution. 😊

PR #2514, opened 2022-07-31
added vdot to numpy frontend and tested

ayush12gupta:
#2479


ayush12gupta:
@ivy-seed @1Doomdie1 @djl11 
Please someone review my PR! it's been a week since my first commit.

kimbring2:
@ayush12gupta Hello, the function using the multiply works well when a, and b is a real numbers. However, it gives different results when I put a complex number.

--------------------------------------------------
import numpy as np

a = np.array([1+2j,3+4j])
b = np.array([5+6j,7+8j])

result1 = np.vdot(a, b)
print("result1: ", result1)
# result1:  (70-8j)

result2 = np.multiply(a, b).sum()
# result2:  (-18+68j)
--------------------------------------------------

At this moment, it seems like we do not need to consider that much.

Anyway, thank you for your contribution!

PR #2517, opened 2022-08-01
rename legacy method `unstack` to `unstack_conts` and add `unstack` method in `ivy/container/general.py`

juliagsy:
- rename legacy method to `unstack_conts` to indicate containers unstacking
- added array and container instance method for arrays unstacking `unstack`


djl11 (approved):
lgtm!

PR #2520, opened 2022-08-01
rename legacy method `split` to `split_conts` and add normal `split` method in `ivy/container/manipulation.py`

juliagsy:
- rename to `split_conts` for the container version (includes unstacking containers if applicable)
- added normal `split` instance methods for array and container (split at leaves only)


djl11 (approved):
lgtm!

PR #2522, opened 2022-08-01
`random` submodule fix

juliagsy:
a side note:
- as randomised values are returned, arrays are tested either on their range or types

one quick question on `multinomial`:
- as `probs` is the only array typed argument, and it is optional, should this method have array/container instances? I didn't implement them yet

please let me know if anything should be changed or improved! thanks!

### DIFF
--- a/ivy/array/random.py
+++ b/ivy/array/random.py
@@ -1,3 +1,4 @@
+# for review
 # global
 import abc
 from typing import Optional, Union
--- a/ivy/array/random.py
+++ b/ivy/array/random.py
@@ -194,6 +194,55 @@ class ArrayWithRandom(abc.ABC):
             out=out,
         )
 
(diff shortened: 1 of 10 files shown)


djl11 (changes requested):
great question with `multinomial`, I hadn't thought about this case. But yes, I think we should add it, with `self` occupying the `probs` argument in both cases. Thanks for checking!

juliagsy:
Thanks for the review! I've added the instance methods accordingly! Requesting another review just in case there's other changes needed! Thank you!

djl11 (changes requested):
`test_multinomial` should make use of `test_function` helper, currently we have the old legacy unit test implementation still in place.

djl11 (approved):
lgtm!

PR #2528, opened 2022-08-01
Update statistical.py

Paulyn01:


Paulyn01:
its now failing all the test which I find confusing

simonetgordon (changes requested):
Undo changes in `.idea/ivy.iml` and `.idea/misc.xml`, commit + push to branch in order to merge with master.

Paulyn01:
how do I do that?

simonetgordon:
You need to make sure that these files are exactly the same as in master and that changes don't appear in this PR. The simplest method might be to just copy and paste if the difference are small.

.idea/ivy.iml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module type="PYTHON_MODULE" version="4">
   <component name="NewModuleRootManager">
-    <content url="file://$MODULE_DIR$" />
-    <orderEntry type="jdk" jdkName="Remote Python 3.8.10 Docker (unifyai/ivy:latest)" jdkType="Python SDK" />
+   <content url="file://$MODULE_DIR$" />
+     <orderEntry type="jdk" jdkName="Remote Python 3.8.10 Docker (unifyai/ivy:latest)" jdkType="Python SDK" />
     <orderEntry type="sourceFolder" forTests="false" />
```
simonetgordon, line 6: I think they're still appearing as different as the spacing is different.

simonetgordon (changes requested):

.idea/ivy.iml
Paulyn01: alright.

simonetgordon (changes requested):
The extra files are still appearing. There are just small things which need to be changed, like spacing and adding a blank line at the end of the file.

Issue #2443, opened 2022-07-30
sin


hirwa-nshuti:

Ajisco:
I'll like to work on this issue.
Can someone kindly assign it tot me?
It's 4 days now and it's not closed yet.
Thanks.

hirwa-nshuti:
Hello @Ajisco if you follow the contribution guide you can be able to know how to pick an issue to work on and reserve it for you. Because I opened this issue and linked it to the issue I wanted to implement the function for.
I hope this link can help you out https://lets-unify.ai/ivy/contributing/4_open_tasks.html#open-tasks
Thanks.

PR #2553, opened 2022-08-02
Added `sin` function for Numpy frontend

hirwa-nshuti:
Hello @jkeane508 
This PR solves issue #2443 

I added the frontend `sin` function for NumPy.
But I have the failing tests and I saw the same goes for tan implemented function before, maybe the problem is with the helper function. 

Thanks.

### DIFF
--- a/ivy/functional/frontends/numpy/mathematical_functions/trigonometric_functions.py
+++ b/ivy/functional/frontends/numpy/mathematical_functions/trigonometric_functions.py
@@ -2,6 +2,28 @@
 import ivy
 
 
+def sin(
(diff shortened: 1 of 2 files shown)


jkeane508:
Hi @hirwa-nshuti,

It looks to be an issue comparing the calculated value for the frontend function and the framework we compare to (numpy). I will take a look at `sin` and then get back to you with whether its an issue with the helper function!

jkeane508:
I'm just going to go ahead and merge this, it looks good to me. The issue with the helper function is work in progress

PR #2554, opened 2022-08-02
Added statistical-functions (mean) to the numpy in frontends.

Ajisco:
I created a folder "statistical-functions" in the numpy in frontends then created a file "mean" where I wrote the mean functions.
All contribution rules read carefully and followed.


sherry30 (changes requested):
Hi,

If you look at the numpy documentation [here](https://numpy.org/doc/stable/reference/routines.statistics.html#) you'll see that under statistics there are multiple subheadings, and `mean` function is under `Averages and variances` so i think it would be best to rename the folder to `statistics` and rename the file to `averages_and_variances.py` and the `mean` function can go under this file.

Ajisco:
Corrected !

Kindly review and merge:
https://github.com/unifyai/ivy/pull/2578

Thanks & Regards.

PR #2561, opened 2022-08-02
Elementwise fixups

Unevilicorn:
Fixes tests in elementwise


ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py
```diff
@@ -269,6 +283,14 @@ def test_atan2(
 ):
     input_dtype, x = dtype_and_x
 
+    x1 = np.asarray(x[0], dtype=input_dtype[0])
+    x2 = np.asarray(x[1], dtype=input_dtype[1])
+
+    x1 = _not_too_close_to_zero(x1)
+    x2 = _not_too_close_to_zero(x2)
+
+    assume(not (np.any(np.isclose(x1, 0)) or np.any(np.isclose(x2, 0))))
```
djl11, line 292: this will eventually be fixed by the updates to the `safety_factor` argument in `test_function`, which will prevent the generation of values close to zero, but for now this is okay.

djl11 (approved):
this all looks okay, but are you sure this doesn't result in flaky tests? I would have thought that quite often at least one value in the array is close to zero, and so many of the `assume` calls result in skipping, and therefore too many tests being skipped overall? Again, eventually this should be handled during data generation itself, once the `safety_factor` also prevents very small values, but I don't see any harm in adding these `assume` statements in the meantime, provided that they don't cause any issues of their own.

PR #2563, opened 2022-08-02
Formatting clip_matrix_norm. Issue #2324

daniilgaltsev:
This is function formatting of clip_matrix_norm. If I've missed anything, please, tell me.


ivy/functional/ivy/general.py
```diff
@@ -1302,9 +1302,75 @@ def clip_matrix_norm(
...
+    >>> y = ivy.clip_matrix_norm(x, 2.0)
+    >>> print(y)
+    {
+        a: ivy.array([[0., 0.894, 1.79]]),
+        b: ivy.array([[0.849, 1.13, 1.41]])
+    }
     """
     norms = ivy.matrix_norm(x, p, keepdims=True)
-    ratios = ivy.maximum(ivy.stable_divide(max_norm, norms), 1.0)
+    ratios = ivy.minimum(ivy.stable_divide(max_norm, norms), 1.0)
```
simonetgordon, line 1373: Why make the change from maximum to minimum here? Was there an issue with the function?

simonetgordon:
Hi Daniil,

Thanks for the contribution! 😊

These changes look good. test-docstrings and test-numpy-style-docstrings are both passing for this function. You seem to cover all necessary instances in the docstring examples and the static method looks good.

I just have a question about why you have made the change from `maximum` to `minimum` in the `clip_matrix_norm` ivy function?

simonetgordon:
#2324

daniilgaltsev:
There was a bug with its implementation. During clipping the values should only be downscaled. When using maximum of (max_norm / norms) and 1.0, if max_norm is larger, it would increase the values in the matrix.

For example, before the change `ivy.clip_matrix_norm(ivy.array([[1.]]), 0.5)` would return 1 and `ivy.clip_matrix_norm(ivy.array([[1.]], 2.0)` would return 2.0, which is incorrect clipping.

simonetgordon:
Ah I see. Unfortunately, the appropriate ivy tests for this function aren't running to make sure everything is running smoothly. I will try to get them working before merging. Thanks for flagging this!

PR #2569, opened 2022-08-02
Adds abs to jax frontend api with tests

darkshredder:
Changes:-

- Jax frontend API for abs
- Adds Jax  tests for abs

Signed-off-by: darkshredder <srivastavayash58@gmail.com>


modanesh:
Thanks for the contribution. It looks good.

PR #2588, opened 2022-08-03
Gradient functions

VedPatwardhan:
Added the `value_and_grad` function


djl11 (approved):
I haven't gone through each implementation in crazy detail, but if the input-output behavior is consistent between each of the backends, then that's good enough for me!

PR #2607, opened 2022-08-03
add logical_xor Tensorflow frontend

ntg2208:


ntg2208:
PR for #2603

ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py
```diff
@@ -207,6 +207,8 @@ def _permute_dims_helper(draw):
 
 
 # permute_dims
+
```
Unevilicorn, line 210: There shouldn't be spaces between the comment and the function, same with the other functions in this file

Unevilicorn (changes requested):
Looks good! Just a small change then you're good to go.

Unevilicorn:
All merged :+1:

Issue #1578, opened 2022-06-22
Add Operator functions for Jax Frontend


Aarsh2001:
Add [Operator functions](https://jax.readthedocs.io/en/latest/jax.lax.html#operators)for Jax frontend:

- [ ] abs
- [ ] add
- [ ] acos
- [ ] approx_max_k
- [ ] approx_min_k
- [ ] agrmax
- [ ] argmin
- [ ] asin
- [ ] atan
- [ ] atan2
- [ ] batch_matmul
- [ ] bessel_i0e
- [ ] bessel_ile
- [ ] betainc
- [ ] bitcast_convert_type
- [ ] bitwise_not
- [ ] bitwise_and
- [ ] bitwise_or
- [ ] bitwise_xor
- [ ] population_count
- [ ] broadcast
- [ ] broadcasted_iota
- [ ] broadcast_in_dim
- [ ] cbrt
- [ ] ceil
- [ ] clamp
- [ ] collapse
- [ ] complex
- [ ] concatenate
- [ ] conj
- [ ] conv
- [ ] convert_element_type
- [ ] conv_dimension_numbers
- [ ] conv_general_dilated
- [ ] conv_general_dilated_local
- [ ] conv_general_dilated_patches
- [ ] conv_with_general_padding
- [ ] conv_transpose
- [ ] cos
- [ ] cosh
- [ ] cummax
- [ ] cummin
- [ ] cumprod
- [ ] cumsum
- [ ] digamma
- [ ] div
- [ ] dot
- [ ] dot_general
- [ ] dynamic_iindex_in_dim
- [ ] dynamic_slice
- [ ] dynamic_slice_in_dim
- [ ] dynamic_update_slice
- [ ] dynamic_update_index_in_dim
- [ ] dynamic_update_slice_in_dim
- [ ] eq
- [ ] erf
- [ ] erfc
- [ ] erf_inv
- [ ] exp
- [ ] expand_dims
- [ ] expm1
- [ ] fft
- [ ] full
- [ ] full_like
- [ ] gather
- [ ] ge
- [ ] gt
- [ ] igamma
- [ ] igammac
- [ ] imag
- [ ] index_in_dim
- [ ] index_take
- [ ] iota
- [ ] is_dinite
- [ ] le
- [ ] lt
- [ ] lgamma
- [ ] log
- [ ] log1p
- [ ] max
- [ ] min
- [ ] mul
- [ ] ne
- [ ] neg
- [ ] nextafter
- [ ] pad
[...]

PR #2610, opened 2022-08-03
Create std function

Doregon499:
issue  std #2465


modanesh:
Thanks for you contribution. There are a few problems with the PR:

1- There are syntax errors in the changes, for example: `ivi` on line 22. Please fix these.
2- Add your file to this directory: `ivy/functional/frontends/numpy/statistics` and name the file as the function being implemented.
3- Please write a test for it in this directory: `ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics` while following  the naming convention of Ivy.

Doregon499:
number 1 and 2 are done but I'm trying to do the test and there are some errors with the num_positional_args function

modanesh:
Thanks for the update.

There are some tests that are failing: "lint / check-formatting" and "test-frontend-numpy / run-nightly-tests". Could you please check them and update your commit accordingly?
Thanks.

Doregon499:
hey, can i have a review, so I did all the changes but the test function is having an error, I tried to solve it and even ask to the ivy team in discord but they told me that you can help me with that

Doregon499:
https://discord.com/channels/799879767196958751/982738436383445073/1007293241122631720

here is the answer from them

modanesh:
So regarding the lint checks, the error messages are pretty much self explanatory. For example, at line 11 of `test_standard_deviation.py`, it throws the message `expected 2 blank lines, found 1`. You can simply add a new line to resolve this error. You should do so for other errors as well.

Regarding the test-frontend-numpy checks, there is a key error at line 2540 of `helpers.py`. Again, looking at the tests outputs could be helpful.

After all, you can also take a look at similar PRs and see how they've addressed these issues to give you some ideas.

Let me know if there are other questions.

Doregon499:
after making some changes I got the following error ``AttributeError: module 'ivy.functional.frontends.numpy' has no attribute 'std'`` , with the fn_name with just std
but if I change it for the path ivy.functional.frontends.numpy.std the error continues with the helpers functions, I think that the error come from the __init__ but I'm not sure, can you help me with that?

modanesh:
Exactly! You need to add it to the `__init__` here: `ivy/functional/frontends/numpy/statistics/__init__.py`. That should resolve the error you mentioned.

Doregon499:
Thank you!! I made the changes but now I have the following error 
       # if the passed arguments do not contain a container, the function using
        # the passed arguments, returning an ivy or a native array.
>       return fn(*args, **kwargs)
E       TypeError: std() takes 1 positional argument but 4 were given

In my mind, this happens for the pip ivy version but after the update, it didn't change, also I changed the arguments but the same error appears, what can it be?

modanesh:
So in the definition of the `std` function in `ivy/functional/frontends/numpy/statistics/standard_deviation.py`, you have added a bare * as the function argument. In python, bare * is used to force the caller to use named arguments, rather than positional arguments. Now, in the `func_wrapper`, when it gets to this line `return fn(*args, **kwargs)` to call the `std` function, it is given `args` with only positional arguments: `(array(-1., dtype=float16), None, 0.0, False)`. That the reason behind the error. Except for the first argument `x`, you should change other ones to positional arguments.

modanesh:
There are a few errors with your latest commit that makes tests failing. For example one of them happens in `test_standard_deviation.py` line 57 leading to this error: `TypeError: _std_dispatcher() got an unexpected keyword argument 'correction'`. Another example could be the lint tests at line 11 of `standard_deviation.py`: `unexpected spaces around keyword / parameter equals`. 
For a reference, you can take a look at the `averages_and_variances.py` function.

modanesh:
There are still failing lint tests. The errors are pretty self-explanatory so I suggest first run tests locally and see if they pass, then you may commit them.

I assume this problem is too difficult with your current level of understanding of Ivy. There are [easier tasks](https://github.com/unifyai/ivy/issues/1578) that you could contribute to.

If it is fine with you, I may close this PR soon. We cannot spend too much time on only one commit. Sorry about that :(

Doregon499:
> There are still failing lint tests. The errors are pretty self-explanatory so I suggest first run tests locally and see if they pass, then you may commit them.
> 
> I assume this problem is too difficult with your current level of understanding of Ivy. There are [easier tasks](https://github.com/unifyai/ivy/issues/1578) that you could contribute to.
> 
> If it is fine with you, I may close this PR soon. We cannot spend too much time on only one commit. Sorry about that :(

Ok, but before you close it, can I ask how was the solution for the error of  _std_dispatcher(), since I tried to change the argument to ddof but it seems that didn't worked

modanesh:
Sure. At line 19 of `standard_deviation.py`, you set a keyword argument `ddof` which passes values to `ivy.std`. However, in the `ivy.std` definition in `functional.ivy.statistical.py`, it doesn't have `ddof` as its input arguments, rather it has `correction`.

PR #2623, opened 2022-08-04
Add concat to frontends.torch.indexing_slicing_joining_mutating_ops

PatternFinder:


PatternFinder:
Todo: Add Indexing, Slicing, Joining, Mutating Ops to PyTorch Frontend
Task in issues: #2621

PR #2624, opened 2022-08-04
Add test_torch_concat to test_frontends.test_indexing_slicing_joining…

PatternFinder:
…_mutating_ops


PatternFinder:
Todo: Add Indexing, Slicing, Joining, Mutating Ops to PyTorch Frontend
Task in issues: #2621

PR #2625, opened 2022-08-04
Add concat to frontends.torch.indexing_slicing_joining_mutating_ops

PatternFinder:

link to issue: #2621


PatternFinder:
@JLU-Neal Hi, could you take a quick look and give me some feedbacks on how I did, so I can use the feedbacks to create more PRs on these frontend apis

PR #2630, opened 2022-08-04
backend function attribute setting

juliagsy:
- added logic for finding and setting attributes
- simplified decorator wrapping logic

### DIFF
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -365,30 +365,42 @@ def _wrap_function(key: str, to_wrap: Callable, original: Callable) -> Callable:
                 )
         return to_wrap
     if isinstance(to_wrap, FunctionType):
-        if hasattr(original, "array_spec"):
-            to_wrap.array_spec = original.array_spec
-        if hasattr(original, "infer_device") and not hasattr(to_wrap, "infer_device"):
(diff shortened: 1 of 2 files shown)


ivy/func_wrapper.py
```diff
@@ -365,30 +365,22 @@ def _wrap_function(key: str, to_wrap: Callable, original: Callable) -> Callable:
...
-        ):
-            to_wrap = handle_out_argument(to_wrap)
-        if hasattr(original, "handle_nestable") and not hasattr(
-            to_wrap, "handle_nestable"
-        ):
-            to_wrap = handle_nestable(to_wrap)
+        # set attributes
+        for attr in original.__dict__.keys():
+            # private attribute or decorator
+            if attr.startswith("__") or hasattr(ivy, attr):
```
djl11, line 371: we use a single "_" for private attributes and functions, and so we should change to `attr.startswith("_")`

djl11 (changes requested):
small change requested, and I also actually updated the task description a bit recently (sorry for doing this last-minute). Could we skip the attribute `handles_out_arg`? If this attribute is set on the Ivy function, it should **not** be set on the underlying backend function.

PR #2633, opened 2022-08-04
Add stack 

PatternFinder:
to frontends.torch.indexing_slicing_joining_mutating_ops

link to issue: #2631

PR #2637, opened 2022-08-04
Add reshape frontends.torch.indexing_slicing_joining_mutating_ops

PatternFinder:
link to issue: #2636


juliagsy (changes requested):
Hey there! Thanks for contributing! As you chose `reshape` as your frontend function to work on, please only make changes on that function and its tests without anything else, thanks!

PatternFinder:
> Hey there! Thanks for contributing! As you chose `reshape` as your frontend function to work on, please only make changes on that function and its tests without anything else, thanks!

Hi! thanks for the followup! The reason I had the others in there is because I've also claimed those tasks too. PRs: #2625, #2633; but I haven't got any feedbacks on those yet.. 

1. For PR formatting: Should I not include all the changes up to this point in each new branch?
2. Are what I had with `reshape` in this PR the correct way to do things? would love some feedbacks so I can improve future RPs

PatternFinder:
@juliagsy I removed the other tasks currently in PR #2625, #2633 and left only `reshape`; please provide feedback on `reshape`, thanks :)

juliagsy:
oh! I see, yeah, please do each function in different PRs! I'll take a look at reshape and provide a review soon! Thanks!

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
```diff
@@ -2,5 +2,11 @@
 import ivy
 
 
-def cat(tensors, dim=0, *, out=None):
-    return ivy.concat(tensors, dim, out=out)
+def reshape(input, shape):
+    return ivy.expand_dims(input, shape)
```
juliagsy, line 6: please use `ivy.reshape` instead of expand_dims

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
```diff
@@ -70,3 +70,38 @@ def test_torch_cat(
         dim=unique_idx,
         out=None,
     )
+
+    
+# reshape
+@given(
+    xs_n_input_dtypes_n_unique_idx=_arrays_idx_n_dtypes(),
```
juliagsy, line 77: please use `helpers.dtype_and_values` as unique idx is not needed, and set the argument `available_dtypes` to:

```
available_dtypes=tuple(
            set(ivy_np.valid_float_dtypes).intersection(
                set(ivy_torch.valid_float_dtypes)
            )
        )
```

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
```diff
@@ -70,3 +70,38 @@ def test_torch_cat(
...
+)
+def test_torch_reshape(
+    xs_n_input_dtypes_n_unique_idx,
+    as_variable,
+    num_positional_args,
+    native_array,
+    with_out,
+    fw,
+):
+    xs, input_dtypes, unique_idx = xs_n_input_dtypes_n_unique_idx
```
juliagsy, line 93: once you've changed to use `dtype_and_values`, there will only be two variables - `dtype, x = ...` needed

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
```diff
@@ -70,3 +70,38 @@ def test_torch_cat(
...
+    xs = [np.asarray(x, dtype=dt) for x, dt in zip(xs, input_dtypes)]
```
juliagsy, line 94: keep only `np.asarray(x, dtype=...)`, extra square brackets for list casting is not required too

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
```diff
@@ -70,3 +70,38 @@ def test_torch_cat(
...
+    helpers.test_frontend_function(
+        input_dtypes=input_dtypes,
+        as_variable_flags=as_variable,
+        with_out=with_out,
+        num_positional_args=num_positional_args,
+        native_array_flags=native_array,
+        fw=fw,
+        frontend="torch",
+        fn_name="reshape",
+        tensors=xs,
```
juliagsy, line 104: please update parameters here to match parameters in `torch.reshape`, with variables mapped to their respective arguments

juliagsy (commented):
Hey there! Great work so far, there are a few changes required before it's good to go! Please request my review again when you're done! Thank you!

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
PatternFinder: updated to return ivy.reshape, as that is the main alias

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
PatternFinder: question: 
the args of helpers.test_frontend_function() should be the expected args of the backend functions, in this case, since torch.reshape only asks for 2 args, input: Tensor, shape: Tuple(int), we would only include these two args for helpers.test_frontend_function(), correct?

PatternFinder:
@juliagsy I've updated per your instructions. I do have two questions if I may:

1. the `args` of helpers.test_frontend_function() should take the expected `args` of the backend functions, in this case, since torch.reshape only asks for 2 `args`, `input: Tensor`, `shape: Tuple(int)`, we would only input these two `args` for helpers.test_frontend_function(), correct?

2. the returns of helpers.dtype_and_values() are `dtype`, `values` and `shape`; The `values` as returned is not yet a tensor correct? helpers.test_frontend_function() cast the `values: List[int]` to torch.tensors?

PatternFinder:
finally caught all the lint :)

Also, what's the reason for test-frontend-torch / run-nightly-tests (pull_request) to fail?

```FAILED ivy/ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py::test_torch_reshape[cpu-ivy.functional.backends.torch-True-False-torch_call-torch]```

**Comment by PatternFinder** [comment]
@juliagsy Hi, I've moved this PR to #2708, could we proceed from there? :)

**Comment by juliagsy** [comment]
Hello! To answer your questions:
1. yes, the inputs will be the 2 args only
2. `dtype_and_values` return `dtype`, `values` only, you can use `np.asarray(values, dtype=dtype)`, then the helper function will process the array internally

there are error logs printed if you open up the workflow, the messages can be checked through there!

as for the PR, we go by assignee so your new assignee will review your PR soon!

thanks for contributing!

**Comment by PatternFinder** [comment]
> Hello! To answer your questions:
> 
> 1. yes, the inputs will be the 2 args only
> 2. `dtype_and_values` return `dtype`, `values` only, you can use `np.asarray(values, dtype=dtype)`, then the helper function will process the array internally
> 
> there are error logs printed if you open up the workflow, the messages can be checked through there!
> 
> as for the PR, we go by assignee so your new assignee will review your PR soon!
> 
> thanks for contributing!

@juliagsy Hi! thanks for the answer. As far as test functions go, I've noticed there are some boilerplates, could you tell me their purpose so I can properly use them for future PRs?

`as_variable_flags`, `with_out`, `native_array_flags`, `container_flags`, `instance_method`, `fw`

**Comment by juliagsy** [comment]
oh right, I should've phrase it better, so the args for `test_function` will be a set of common parameters + function input args

As we can have array instances or container instances for the inputs, these parameters are used to indicate whether to test the function with: (all are booleans)
`container_flags`/`instance_method` - array or container inputs 
`native_array_flags` - ivy array or native array inputs

Some other args:
`fw` - as we support different backends (aka conversions), we configured it internally to test for 4 backends

We have a decorator to generate data for these as they are common to all test functions, but don't worry about this as we are currently working on it! I'll share an example from our previous test so that you have a better idea on how they're used!

Example: https://github.com/unifyai/ivy/blob/94d07431e1beedd61a7140167600a93e96ca95eb/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_operators.py#L12

so the parameters up till the `frontend` argument will be the same or similar for all tests, and arguments following are differing according to the function

we also have documentations for further explanations!
https://lets-unify.ai/ivy/deep_dive/17_ivy_frontends_tests.html

hope these will help in clearing your doubts!

**Comment by PatternFinder** [comment]
@juliagsy thank you for the detailed reply, very helpful!

so for `container_flags` `instance_method`, when are we supposed to use them? the docs didn't mention them

**Comment by juliagsy** [comment]
They're for testing the backend functions with variables/instances of type ivy.Array or ivy.Container (eg. x.add(y) where x is of type ivy.Container or ivy.Array)

It doesn't apply to frontend functions as these functions' behaviours follow their native framework, that's probably the reason it's not mentioned in the docs! 

Hope this answers your question!

**Comment by PatternFinder** [comment]
@juliagsy 
I see! that makes sense. 

also, when we write the frontend functions, the arguments should match exactly what the original framework has correct? like so:
```
def stack(tensors, dim=0, *, out=None):
torch.stack(tensors, dim=0, *, out=None):
```
and the return should match exactly what the backends has correct?
```
return ivy.stack(tensors, dim, out=out)
def stack(
    arrays: Union[
        Tuple[ivy.Array], List[ivy.Array], Tuple[ivy.NativeArray], List[ivy.NativeArray]
    ],
    axis: int = 0,
    *,
) -> ivy.Array:
```

juliagsy:
yep, that's right!

PatternFinder:
> yep, that's right!

thanks for the clarification :)

Issue #2647, opened 2022-08-05
flip, fliplr, flipud


PatternFinder:

1Doomdie1:
Hey, thank you for contributing.
Can you please open an issue for each function and then mention them in your PR?
This way we can allocate these functions to you in the main issue list.

PatternFinder:
@1Doomdie1 got it. closing this issue to file each functions separately.
Additionally, could you take a look at my other PRs? most of them have not received any attention for a while; I would really love to get some feedbacks on what the correct way to do thing so I can make lots of contributions in the following days. :)

my PRs: #2706, #2707, #2709, #2710, #2711

PR #2648, opened 2022-08-05
Add frontends.torch flip, fliplr, flipud

PatternFinder:
* Add flip, fliplr, flipud to frontends.torch.other_operations

* Add fliplr, flipud to backends.torch.manipulation

* Add test_flip, test_fliplr, test_flipud to test_ivy.frontends.torch.test_other_operations

link to issue: #2647


PatternFinder:
@saeedashrraf Hi, could you take a look at this PR and let me know if this is the correct way going forward? Thanks.

saeedashrraf:
Hello @PatternFinder, thank you for your effort! you need to create a separate issue for each of the functions you are working on instead of one issue for all 3 functions to get assigned the functions correctly. So, instead of only one issue #2647, there should be 3 issues, one for each function.

In addition to that, can you implement fliplr and flipud in the rest of the backends just like you did for torch?

PatternFinder:
> Hello @PatternFinder, thank you for your effort! you need to create a separate issue for each of the functions you are working on instead of one issue for all 3 functions to get assigned the functions correctly. So, instead of only one issue #2647, there should be 3 issues, one for each function.
> 
> In addition to that, can you implement fliplr and flipud in the rest of the backends just like you did for torch?

got it, first I'll close these group by alias PRs, file them separately, then look at implementing the other backends

PatternFinder:
closing to file each function separately

PR #2658, opened 2022-08-05
Out arg handling in mixed functions- docs update

simonetgordon:
I'm not sure I'm correct about editing the `handle_out_argument` function wrapping to check the `handles_out_arg` decorator. If I am, should I implement this preemptively in the function wrapping?

Also, should I attempt to make up an example of a mixed function to help illustrate the description?


djl11:
I did a bad job of explaining this task actually, apologies for this! I just pushed the description to the docs myself, thanks for creating the initial template! :)

PR #2668, opened 2022-08-05
Creation fixes

Unevilicorn:
Fix linspace


ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -2,17 +2,16 @@
 """Collection of tests for creation functions."""
 
 # global
+
+import hypothesis.extra.numpy as hnp
```
jkeane508, line 6: This should be a local Import :)

jkeane508 (changes requested):
This is good to merge apart from the small change as below

PR #2680, opened 2022-08-06
Added `small_value_safety_factor` for ivy tests

rush2406:
- renamed `safety_factor` argument to `large_value_safety_factor`
- add a new argument `small_value_safety_factor` at log scale to prevent very small float values


ivy_tests/test_ivy/helpers.py
```diff
@@ -1803,37 +1807,46 @@ def array_values(
...
+                min_value, round(-9223372036854775808 * large_value_safety_factor)
             )
             max_value = ivy.default(
-                max_value, round(9223372036854775807 * safety_factor)
+                max_value, round(9223372036854775807 * large_value_safety_factor)
             )
         elif dtype == "uint8":
-            min_value = ivy.default(min_value, round(0 * safety_factor))
-            max_value = ivy.default(max_value, round(255 * safety_factor))
+            min_value = ivy.default(min_value, round(0 * large_value_safety_factor))
```
djl11, line 1832: just realized, the `large_value_safety_factor` is not relevant for `min_value` of unsigned integers, so we should remove this for all of them.

ivy_tests/test_ivy/helpers.py
```diff
@@ -1790,7 +1793,8 @@ def array_values(
     exclude_min=True,
     exclude_max=True,
     allow_negative=True,
-    safety_factor=0.95,
+    large_value_safety_factor=0.95,
+    small_value_safety_factor=1,
```
djl11, line 1797: this default value should also be `0.95`

ivy_tests/test_ivy/helpers.py
```diff
@@ -1803,37 +1807,46 @@ def array_values(
     values = None
     if "int" in dtype:
```
djl11, line 1808: we should actually include the `small_value_safety_factor` for integer types as well, with `1` then being the smallest magnitude value generated, for any case where `small_value_safety_factor < 1`. This will enable us to prevent zero divisions via the safety factor when generating integers, as an integer divided by an integer leads to a float.

ivy_tests/test_ivy/helpers.py
```diff
@@ -1845,11 +1858,13 @@ def array_values(
                     width=16,
                     exclude_min=exclude_min,
                     exclude_max=exclude_max,
-                ),
+                ).filter(lambda x: x <= -1 * limit or x >= limit),
```
djl11, line 1861: we should use the `min_val` and `max_val` [arguments](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.floats).

djl11 (changes requested):
some general changes request. Also, we should actually change things so that the safety factor is always `>=1`. This is actually how safety factors are defined. Once you have made these changes, then the default safety factors should both be `1.1` rather than `0.95`. Thanks!

ivy_tests/test_ivy/helpers.py
rush2406: @djl11 
I think if `min_val` and `max_val` are used, there could still be very small values generated because it generates from [`min_val` , `max_val`]. The way I understood is that we need values generated from [`min_val`, -limit] and [limit, `max_val`] where `limit` is calculated based on `small_value_safety_factor`.

Should `limit` be defined by `log(safety_factor)`?

ivy_tests/test_ivy/helpers.py
rush2406: @djl11 
Would it be right to do it for `int8`, `int16` etc. because this would result in negative values not being generated?

djl11: you should do something like:

```
min_value = ivy.default(min_value, round(-128 * large_value_safety_factor))
max_neg_value = -1 if small_value_safety_factor > 1 else 0
min_pos_value = 1 if small_value_safety_factor > 1 else 0
max_value = ivy.default(max_value, round(127 * large_value_safety_factor))
```

and then at the end:

```
values = draw(list_of_length(x=st.integers(min_value, max_neg_value) | st.integers(min_pos_value, max_value), length=size))
```

ivy_tests/test_ivy/helpers.py
```diff
@@ -1627,7 +1627,8 @@ def dtype_and_values(
     num_arrays=1,
     min_value=None,
     max_value=None,
-    safety_factor=0.95,
+    large_value_safety_factor=0.95,
```
djl11, line 1630: these should be `1.1` not `0.95`

djl11 (changes requested):

ivy_tests/test_ivy/helpers.py
rush2406: When I change it to 1.1, specifically for `uint64` and `int64`, it starts throwing an error: `Python int too large to convert to C long`. This is because of overflow and the number can no loger berepresented by `uint` or `int`. I have tried to handle this by including a check while generating `uint64` and `int64`

djl11 (approved):
lgtm!

PR #2701, opened 2022-08-06
Added type hints, default value corrections to ivy/stateful

9harshit:


9harshit:
HI @Auxia,

I am not sure why, but the checks are failing on files that I haven't edited. So I am not sure how this will affect my PR acceptation.

Let me know if there is something I can do to fix it.

Thanks,
Harshit

Auxia:
Hi @9harshit,

Sorry for the delayed response. Currently the reformatting task is contained to the functional API and not the stateful API. You can still contribute, however the todo list for the required reformatting is different from the one outlined in the function reformatting task.

I would suggest that you pick up a task from Functional Reformatting Issues or the new Frontend Tasks.

Thank you!

Issue #2672, opened 2022-08-05
try_else_none


raotnameh:

raotnameh:
@HotzingTone Can you confirm if this is what is required of the function? 

def try_else_none(fn,*args,**kwargs):
    """Try and return the function otherwise, return None if an exception was raised
    during function execution.
    """
    try: 
    fn(*args,**kwargs)
    return fn
    else: return None

PR #2728, opened 2022-08-07
Updated the try_else_none() function.

raotnameh:
I have added the try_else_none() function to the ivy general.py. 
I opened the issue for the same, and the link to which is:  https://github.com/unifyai/ivy/issues/2672

### DIFF
--- a/ivy/functional/ivy/general.py
+++ b/ivy/functional/ivy/general.py
@@ -1759,7 +1759,8 @@ def to_native_shape(shape: Union[ivy.Shape, ivy.NativeShape]) -> ivy.NativeShape
 
 @handle_nestable
 def try_else_none(fn: Callable, *args: Any, **kwargs: Any) -> Union[Callable, None]:
-    """Try and return the function, otherwise return None if an exception was raised during function execution.
+    """Try and return the function, otherwise return None
+        if an exception was raised during function execution.
 
     Parameters
     ----------
@@ -1781,21 +1782,22 @@ def try_else_none(fn: Callable, *args: Any, **kwargs: Any) -> Union[Callable, No
     >>> x = ivy.array([1, 2, 3])
     >>> y = ivy.array([4, 5, 6])
     >>> z = ivy.try_else_none(ivy.add,x, y)
-    >>> print(y)
+    >>> print(z)
     ivy.add
 
     with: if the function is executed with an exception
     >>> x = ivy.array([1, 2, 3])
     >>> y = 'hemant'
     >>> z = ivy.try_else_none(ivy.add,x, y)
-    >>> print(y)
+    >>> print(z)
     None
 
     """
     try: 
-        fn(*args, **kwargs)
+        _ = fn(*args, **kwargs)
         return fn 
-    except: return None
+    except Exception: 
+        return None
 
 
 def arg_names(receiver):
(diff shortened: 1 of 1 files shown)


ivy/functional/ivy/general.py
```diff
@@ -1758,17 +1758,44 @@ def to_native_shape(shape: Union[ivy.Shape, ivy.NativeShape]) -> ivy.NativeShape
...
+        Either the function itself or None if an exception was raised
+        during function execution.
+
+    Examples
+    --------
+    with: if the function is executed without any exception
+    >>> x = ivy.array([1, 2, 3])
+    >>> y = ivy.array([4, 5, 6])
+    >>> z = ivy.try_else_none(ivy.add,x, y)
+    >>> print(y)
```
rush2406, line 1784: this should be `print(z)`?

ivy/functional/ivy/general.py
```diff
@@ -1758,17 +1758,44 @@ def to_native_shape(shape: Union[ivy.Shape, ivy.NativeShape]) -> ivy.NativeShape
...
+    ivy.add
+
+    with: if the function is executed with an exception
+    >>> x = ivy.array([1, 2, 3])
+    >>> y = 'hemant'
+    >>> z = ivy.try_else_none(ivy.add,x, y)
+    >>> print(y)
```
rush2406, line 1791: this should be `print(z)`?

ivy/functional/ivy/general.py
```diff
@@ -1758,17 +1758,44 @@ def to_native_shape(shape: Union[ivy.Shape, ivy.NativeShape]) -> ivy.NativeShape
...
+    None
 
     """
-    return default(fn, None, True)
+    try: 
+        fn(*args, **kwargs)
+        return fn 
+    except: return None
```
rush2406, line 1798: please change to `except Exception`:

rush2406 (changes requested):
Hi @raotnameh
Thanks for contributing! There are a few minor changes to be done, please check the comments. Also please fix the `lint` issues which are just formatting-related and request a review when done

ivy/functional/ivy/general.py
raotnameh: Should it return something or just pass?

rush2406: should be `None` I think as the function definition suggests

ivy/functional/ivy/general.py
```diff
@@ -1758,17 +1758,46 @@ def to_native_shape(shape: Union[ivy.Shape, ivy.NativeShape]) -> ivy.NativeShape
...
+        during function execution.
+
+    Examples
+    --------
+    with: if the function is executed without any exception
+    >>> x = ivy.array([1, 2, 3])
+    >>> y = ivy.array([4, 5, 6])
+    >>> z = ivy.try_else_none(ivy.add,x, y)
+    >>> print(z)
+    ivy.add
```
rush2406, line 1786: On running this example, it shows `<functionaddat0x7f81905a9700>` rather than `ivy.add`. Can you please check and change accordingly because this is causing the `docstring` tests to fail?

rush2406 (changes requested):
please check the example output as mentioned below!

ivy/functional/ivy/general.py
raotnameh: Made the change to resolve it.

Issue #1565, opened 2022-06-20
Add RawOps Operations for TensorFlow frontend


sreegandh804:
Add [Metrics functions](https://www.tensorflow.org/api_docs/python/tf/keras/metrics#functions_2) for TensorFlow frontend:
- [ ] binary_accuracy

Bobbyorr007:
- [ ] #1565

Bobbyorr007:
- [ ] #1693

kimbring2:
- [ ] #2278

Lancelot03:
- [ ]  ArgMax #2392

Lancelot03:
- [ ]  argmax #2433

anapartu:
- [ ] #2866

AmiraOmar99:
- [ ] #3217

ZoeCD:
- [ ] #3220

rush2406:
- [ ] #3231

rush2406:
- [ ] #3232

ZoeCD:
- [ ] #3264

ZoeCD:
- [ ] #3265

PatricYan:
- [ ] #3350

PatricYan:
- [ ] #3352

rush2406:
- [ ] #3357

rush2406:
- [ ] #3358

rush2406:
- [ ] #3363

PatricYan:
- [ ] #3367

PatricYan:
- [ ] #3368

Lancelot03:
- [ ]  #3370

Lancelot03:
- [ ]  #3371

Lancelot03:
- [ ]  #3372

ZoeCD:
- [ ] #3387
- [ ] #3388
- [ ] #3391

Lancelot03:
- [ ]  #3392

johanjino:
#3485 
#3484

rush2406:
- [ ] #3519 
- [ ] #3520
- [ ] #3521

sakibkuet10:
- #3667

carlacampas:
- [Concat] #3700

carlacampas:
- [Cross] #3700

kurshakuz:
[ ] #3713

kurshakuz:
- [ ] #3733

ahmshalan:
- [ ] #3811

[...]

PR #2753, opened 2022-08-08
added skeleton files to tensorflow frontend.

zaeemansari70:


Oxbeefbabe:
Hey, was this an assigned task? If so can you link to me where it was described so I can make sure it is all alright? Cheers!

ivy/functional/frontends/tensorflow/math.py
```diff
@@ -0,0 +1,16 @@
+# global
+import ivy
+
+
+def add(x, y, name=None):
```
djl11, line 5: this is also in layers? Is this a mistake, or it appears in both categories?

djl11 (changes requested):

ivy/functional/frontends/tensorflow/math.py
zaeemansari70: Yes, It appears in both categories. I've also attached the link of TensorFlow API of layers and math below just for reference.
Layers: https://www.tensorflow.org/api_docs/python/tf/keras/layers#functions_2
Math: https://www.tensorflow.org/api_docs/python/tf/math#functions

djl11 (approved):
lgtm!

KareemMAX:
Hi @zaeemansari70, @djl11. This PR addressed some subtasks in #1565, but those subtasks are not marked as completed. I created #5457, and #5458 to mark them as completed, not sure if that is right though.

Close #5457
Close #5458

PR #2754, opened 2022-08-08
add fliplr frontends.torch

PatternFinder:
link to issue: #2714


ivy/functional/backends/torch/manipulation.py
```diff
@@ -77,6 +77,22 @@ def flip(
 )
 
 
+def fliplr(
+    x: torch.Tensor,
```
JLU-Neal, line 81: should not add function to the backend, pls read the docs carefully.

JLU-Neal (changes requested):
The test is not passed, and you should not add function to the backend.

PR #2755, opened 2022-08-08
add flipud frontends.torch

Creative-Ataraxia:
link to issue: #2715


VedPatwardhan:
Hi @PatternFinder, seems like this function is not a part of our functional API currently and it won't get added until the next release, so I'd like to suggest you to pick another frontend task in the meantime, we'll take a look at this PR as an Extension according to the workflow mentioned in the [docs](https://lets-unify.ai/ivy/contributing/1_the_basics.html#extensions). Thanks 🙂

PatternFinder:
> Hi @PatternFinder, seems like this function is not a part of our functional API currently and it won't get added until the next release, so I'd like to suggest you to pick another frontend task in the meantime, we'll take a look at this PR as an Extension according to the workflow mentioned in the [docs](https://lets-unify.ai/ivy/contributing/1_the_basics.html#extensions). Thanks 🙂

got it. I was just following the torch docs along and added what the torch docs had :)
anyway, I've corrected some args on this function according to the instruction you gave on the torch.flip function too.

PatternFinder:
@VedPatwardhan I've updated the functions, hope it is correct. Please let me know what else I can do for this torch.flipud function :) thanks.

VedPatwardhan:
Hi @PatternFinder, thanks for making the changes! I think it looks good, although we'll take a look again after the next release 🙂

PatternFinder:
> Hi @PatternFinder, thanks for making the changes! I think it looks good, although we'll take a look again after the next release 🙂

ok, understood. Will make more changes when needed next

PR #2767, opened 2022-08-08
adding frontend numpy function sqrt

### LINKED ISSUE #2766: sqrt

yugaljain1999:
Fixes #2766 
@ZoeCD @RashulChutani Can you pls review this PR too?


ZoeCD:
Hi, I solve the merge conflicts and a NumPy frontend error. But there are more, can you please try to fix them?

yugaljain1999:
Hi @ZoeCD , I just checked test-frontend-numpy and test-docstrings checks but there are assertion errors which may be due to test cases written for numpy frontend functions. So I think someone will fix this through different PR as it's general issue.
So I think this PR would be fine to merge. What do you think?
Thanks

ZoeCD:
Yes, there are some errors related to the test function. However, I see that only the sqrt function raises this error `TypeError: No loop matching the specified signature and casting was found for ufunc sqrt` I would appreciate it if you could look into it, but if you conclude it is not related to the function itself, I'll merge the PR 🙂

yugaljain1999:
@ZoeCD That error is not related to this sqrt function implementation. That's casting issue which can be solved by different PR as we can not modify dtype only for this function. We need to modify dtype values for all mathematical functions in one shot so that it won't interfere functionality of other function. I hope you are getting me.
So you can merge this PR.)
Thanks

Issue #2635, opened 2022-08-04
matrix_rank


YushaArif99:

YushaArif99:
#2635

1Doomdie1:
Can you please provide the issue ID from where you took this function, please?
Thank you.

YushaArif99:
@1Doomdie1 It was #1546

1Doomdie1:
Thank you!

PR #2774, opened 2022-08-08
implementing Tensorflow matrix_rank Frontend function

YushaArif99:
#2635


YushaArif99:
@rush2406 Could you kindly ask @djl11 to review this PR. There is a technical aspect regarding the implementation of the `matrix_rank `function that might need Dan's approval

rush2406 (commented):
> @rush2406 Could you kindly ask @djl11 to review this PR. There is a technical aspect regarding the implementation of the `matrix_rank `function that might need Dan's approval

Okay sure, but the PR shows 0 files changed as the status. Can you please check about this?

YushaArif99:
@rush2406 I have made the changes to the file. I don't know why but for some reason, my previous commits didn't show changes to the file. Could you kindly request @djl11 to also have a look. Thanks!

ivy/functional/frontends/tensorflow/__init__.py
```diff
@@ -1,5 +1,4 @@
 # flake8: noqa
-# flake8: noqa
```
rush2406, line 2: this file need not be changed

rush2406 (changes requested):
Hi 
I don't think the backend related files should be changed, however, I'll request Dan's review as well. The frontend related changes seem correct to me!

YushaArif99:
> Hi I don't think the backend related files should be changed, however, I'll request Dan's review as well. The frontend related changes seem correct to me!

Yeah that was basically the reason why I wanted Dan to take a look as well. The reason I changed the backend files was to provide consistent behavior across the backend frameworks. Let me elaborate on this point. 

**Problem**: The implementation of the `matrix_rank ` function across all frameworks involves 2 steps. The first step is to decompose the input matrix using `svd` and the next step is to count all elements in the resultant matrix whose values are greater than the tolerance value(provided by the user or default). This count is what represents the rank of the matrix. Unfortunately, since `svd `is an iterative algorithm, the resultant matrix you get can be different across the frameworks depending on the initialization you start with. The problem here is that this behavior causes assertion errors when value testing. I even tested this myself where I passed the same matrix to the native svd functions (i.e: _torch.linalg,svd_ or _numpy.linalg.svd_ etc..) and got a completely different resultant _S_ matrix.

**Solution**: There are 2 solutions that I came up with.

1.   The input matrix is of the dimension (*, m, n) where * is zero or more batch dimensions. Therefore, first flatten the matrix to be of (1, z) dimension (where z = * . m . n). Then pass this flattened array to the `matrix_rank `function to ensure consistent results.
2. Leave the input matrix as is. Instead, set the absolute tolerance value(atol=1.0). This is to encounter for those edge cases where there is some discrepancy in the output across the frameworks. This is especially the case when choosing extremely large float values for the input matrix.

Although technically it worked, I didnt like the idea behind 1. because it did not preserve the original behavior of the function itself. Flattening the input would limit the range of output to be [0,1]. Whereas originally the rank can be as high as M for a MxN matrix. This is the reason I decided to opt for the second option

Note: The reason I changed the backend implementations was because they were inconsistent. For some backends (eg TF and Jax), the input was flattened before passing to the native function(_tf.linalg.matrix_rank_) whereas for others this wasnt the case(eg torch and numpy). I'm not sure why this was the case. Maybe I might have missed something.

ivy/functional/backends/torch/linear_algebra.py
```diff
@@ -147,7 +147,15 @@ def matrix_rank(
     out: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
     # ToDo: add support for default rtol value here, for the case where None is provided
-    ret = torch.linalg.matrix_rank(x, rtol=rtol, out=out)
+    if rtol is None and out is None:
```
djl11, line 150: can you please add a comment explaining why each of these conditions are needed. Seems like `ret = torch.linalg.matrix_rank(x, rtol=rtol, out=out)` should work? Are you testing with the latest Dockerfile?

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tf_linalg_functions.py
```diff
@@ -0,0 +1,48 @@
...
+# local
+import ivy_tests.test_ivy.helpers as helpers
+import ivy.functional.backends.numpy as ivy_np
+
+
+@given(
+    dtype_x=helpers.dtype_and_values(
+        available_dtypes=ivy_np.valid_float_dtypes[1:],
+        min_num_dims=2,
+        min_value=-10000000000.0,
```
djl11, line 14: could you write these numbers in `1e<x>` format?

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -1101,15 +1102,14 @@ def test_matrix_norm(
...
 # matrix_rank
 @given(
     dtype_x=helpers.dtype_and_values(
         available_dtypes=ivy_np.valid_float_dtypes[1:],
-        min_num_dims=3,
-        max_num_dims=3,
-        min_dim_size=2,
-        max_dim_size=3,
+        min_num_dims=2,
+        min_value=-10000000000.0,
```
djl11, line 1111: could you write these numbers in `1e<x>` format?

djl11 (changes requested):
small changes requested, otherwise looks good. Can you confirm that the functional ivy tests for `ivy.matrix_rank` are definitely passing? I've noticed you've lessened the restrictions on `num_dims` and `dim_size` to the default values of `dtype_and_values `, which is an improvement in generalisation of the original ivy test, so if it's passing then that's great.

Can you also please resolve the conflicts? We can then run the CI and check the linalg tests here as well.

ivy/functional/backends/torch/linear_algebra.py
YushaArif99: I am not using Dockers since its quite a heavy setup which my laptop can't sustain. Nonetheless, after rechecking, I did realize that I was using an older version of torch which is why I had to handle the `NoneType `arguments explicitly. But yeah, now I have removed that unnecessary logic from the torch backend

YushaArif99:
> small changes requested, otherwise looks good. Can you confirm that the functional ivy tests for `ivy.matrix_rank` are definitely passing? I've noticed you've lessened the restrictions on `num_dims` and `dim_size` to the default values of `dtype_and_values `, which is an improvement in generalisation of the original ivy test, so if it's passing then that's great.
> 
> Can you also please resolve the conflicts? We can then run the CI and check the linalg tests here as well.



I have rerun the functional ivy tests for  `ivy.matrix_rank` and confirmed that they are passing. However, it is worth noting that there is a trade-off here. Apparently, after an emperical analyses, I have come to the conclusion that there is a correlation b/w the range of values produced by the `dtype_and_values` helper method and the `atol` parameter used in the `np.allclose() `method during `value_testing`. 
The following are the values upon which the tests are passing:

- `min_value `= **-1e-5** ; `max_value`= **1e+5** ; `atol`=**1.0**
- `min_value `= **-1.0000000200408773e+20** ;  `max_value `= **1.0000000200408773e+20**  ; `atol `= **2.0**

> The general trend is that the more you restrict the range of values for `dtype_and_values`, the lower you can set the threshold for `atol`.  
I personally think that the 2nd option is more generic so to speak. However, I leave it upto you to decide which option to go for. 

Note: These values aren't optimal. It is possible that the range could be increased a bit more while maintaining the same value for `atol`. The values are just to give a general idea

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
YushaArif99: Yeah no worries!

djl11 (changes requested):
lets go with the first actually:
`min_value = -1e-5`; `max_value= 1e+5`; `atol=1.0`

djl11:
> > small changes requested, otherwise looks good. Can you confirm that the functional ivy tests for `ivy.matrix_rank` are definitely passing? I've noticed you've lessened the restrictions on `num_dims` and `dim_size` to the default values of `dtype_and_values `, which is an improvement in generalisation of the original ivy test, so if it's passing then that's great.
> > Can you also please resolve the conflicts? We can then run the CI and check the linalg tests here as well.
> 
> I have rerun the functional ivy tests for `ivy.matrix_rank` and confirmed that they are passing. However, it is worth noting that there is a trade-off here. Apparently, after an emperical analyses, I have come to the conclusion that there is a correlation b/w the range of values produced by the `dtype_and_values` helper method and the `atol` parameter used in the `np.allclose() `method during `value_testing`. The following are the values upon which the tests are passing:
> 
> * `min_value `= **-1e-5** ; `max_value`= **1e+5** ; `atol`=**1.0**
> * `min_value `= **-1.0000000200408773e+20** ;  `max_value `= **1.0000000200408773e+20**  ; `atol `= **2.0**
> 
> > The general trend is that the more you restrict the range of values for `dtype_and_values`, the lower you can set the threshold for `atol`.
> > I personally think that the 2nd option is more generic so to speak. However, I leave it upto you to decide which option to go for.
> 
> Note: These values aren't optimal. It is possible that the range could be increased a bit more while maintaining the same value for `atol`. The values are just to give a general idea

Could you also please confirm that the Array API tests are also passing? Just want to make sure we are not causing any breakages with the removal of the `if` statements in some of the backend implementations.

YushaArif99:
> lets go with the first actually: `min_value = -1e-5`; `max_value= 1e+5`; `atol=1.0`

Alright then. The changes have been made

YushaArif99:
> > > small changes requested, otherwise looks good. Can you confirm that the functional ivy tests for `ivy.matrix_rank` are definitely passing? I've noticed you've lessened the restrictions on `num_dims` and `dim_size` to the default values of `dtype_and_values `, which is an improvement in generalisation of the original ivy test, so if it's passing then that's great.
> > > Can you also please resolve the conflicts? We can then run the CI and check the linalg tests here as well.
> > 
> > 
> > I have rerun the functional ivy tests for `ivy.matrix_rank` and confirmed that they are passing. However, it is worth noting that there is a trade-off here. Apparently, after an emperical analyses, I have come to the conclusion that there is a correlation b/w the range of values produced by the `dtype_and_values` helper method and the `atol` parameter used in the `np.allclose() `method during `value_testing`. The following are the values upon which the tests are passing:
> > 
> > * `min_value `= **-1e-5** ; `max_value`= **1e+5** ; `atol`=**1.0**
> > * `min_value `= **-1.0000000200408773e+20** ;  `max_value `= **1.0000000200408773e+20**  ; `atol `= **2.0**
> > 
> > > The general trend is that the more you restrict the range of values for `dtype_and_values`, the lower you can set the threshold for `atol`.
> > > I personally think that the 2nd option is more generic so to speak. However, I leave it upto you to decide which option to go for.
> > 
> > 
> > Note: These values aren't optimal. It is possible that the range could be increased a bit more while maintaining the same value for `atol`. The values are just to give a general idea
> 
> Could you also please confirm that the Array API tests are also passing? Just want to make sure we are not causing any breakages with the removal of the `if` statements in some of the backend implementations.

The Array API tests are failing for the `matrix_rank `function. However, I'm not sure if that's caused by the removal of the` if `statements. From what I see, implementation for the `test_matrix_rank `function in` array-api-tests/test-linalg.py` folder seems to be incomplete.

djl11:
Hey @sherry30, would you mind taking a look? Provided the the backend changes do not cause any new failures for the Ivy tests or Array API tests, then I think it's good to merge :)

YushaArif99:
The errors do seem related to the changes I made. And it now makes sense as to why **only** _Tensorflow_ and _Jax_ were flattening the input before calling the `matrix_rank `function whereas _Numpy_ and _Pytorch_ weren't. The errors are being caused due to the following:

- **Tensorflow**: when passing a multidimensional matrix(ndim>2), tf.linalg.matrix_rank( ) raises an` Incompatible shapes` error if the `tol `parameter is a list of floats (rather than only a singular value). The workaround is to flatten the input which apparently solves the issue!
Note: I don't think this is the best way to resolve the issue since it modifies the original behavior of the `tf.linalg.matrix_rank( ) `function. A better solution would be to not generate lists of floats when testing for the `matrix_rank `function in the tensorflow backend during the Array API tests.

- **Jax**: not so sure about this one, but after testing it out on a colab notebook, I found that the issue is probably related to an outdated version of Jax. An older version of Jax(0.3.14) only accepts a 2 dimensional input for the matrix_rank function. This is why the following exception occurs when passing a multidimensional matrix
`TypeError: array should have 2 or fewer dimensions\n\nThe stack trace below excludes JAX-internal frames.`
 After installing the current latest version(0.3.16) in my colab notebook, the error seemed to have been resolved. So it might be possible that the Array-API tests are running on an older version of Jax.

sherry30:
Hi Yusha,

array API test for tensorflow seems to be failing because it also expects it to work in case `rtol `is `tf.Tensor` which is not being handled.

As for jax, since we have settled on the 0.3.14 version for now i think you should leave the old implementation like that. 

Also you don't have to create a new file `test_tf_linalg_functions.py` for the frontend test. You can add your test in the file `ivy_tests/test_ivy/test_frontends/test_tensorflow/test_linalg.py`

YushaArif99:
**> array API test for tensorflow seems to be failing because it also expects it to work in case `rtol `is `tf.Tensor` which is not being handled.**

Yes, that was my point. Btw, I have added an `if `statement in the `tf backend ` to handle this particular case


**> As for jax, since we have settled on the 0.3.14 version for now i think you should leave the old implementation like that.**

I could leave the old implementation as is. But wouldn't that fail the backend tests? Since, for Jax,  I am flattening the input whereas for others, I am not.

**> Also you don't have to create a new file `test_tf_linalg_functions.py` for the frontend test. You can add your test in the file `ivy_tests/test_ivy/test_frontends/test_tensorflow/test_linalg.py`**

My bad. I will make the requested change

sherry30 (commented):
Let me know if i understand this right
for tensorflow, You're trying to pass just a single float value as `rtol` to `tf.linalg.matrix_rank( )` instead of list of floats
If that's the case then 
1. you also have to handle the case when `rtol` is a multidimensional list
2. when `rtol` is a tensor, because currently the `if` statement is only checking for tuple or list.

It seems like array API test for tensorflow is failing right now for these 2 reasons

As for jax, for now you can just leave the old implementation as is, all the tests seem to be passing currently, if it creates an issue in the future then Dan will probably assign someone to look into it and we probably would have to come up with a different solution, since we have pinned this version of jax and it doesn't accept ndim > 2 for `matrix_rank`.

Thanks

YushaArif99:
**Let me know if i understand this right for tensorflow, You're trying to pass just a single float value as `rtol` to** **`tf.linalg.matrix_rank( )` instead of list of floats If that's the case then**
 
**> 1. you also have to handle the case when `rtol` is a multidimensional list**
**> 2. when `rtol` is a tensor, because currently the `if` statement is only checking for tuple or list.**

**It seems like array API test for tensorflow is failing right now for these 2 reasons**

Yes, you are right. TensorFlow does not accept a list of floats/tensor for the `tol` paramater which is why it needs to be explicitly handled. Btw, I have made the changes to now check for both lists and tensor.
> 

**As for jax, for now you can just leave the old implementation as is, all the tests seem to be passing currently, if it creates an issue in the future then Dan will probably assign someone to look into it and we probably would have to come up with a different solution, since we have pinned this version of jax and it doesn't accept ndim > 2 for `matrix_rank`.**

Well, if that's the case, then i'll leave the old implementation as is.


> I have made the requested changes. Could you kindly look into it and let me know if there are any further changes required?

sherry30:
Hey @djl11, all the tests seem to be passing now, can you take a look one last time before merging
Thanks

djl11 (approved):
lgtm!

PR #2779, opened 2022-08-08
Replace legacy ivy method matrix norm

JLU-Neal:


JLU-Neal:
For changes in ivy/functional/backends/tensorflow/linear_algebra.py, the arg ord was incorrectly passed to the backend (when ord=='fro', ord=2 was passed)

JLU-Neal:
For changes in ivy/functional/backends/numpy/linear_algebra.py, I found that the return value ret for the NumPy backend is not guaranteed to be np.ndarray, and sometimes could be float. In this situation, when we call method get_ret_and_flattened_array, the flattened_array could be a null list since before flattening, it will first check whether the value is an instance of ivy array.

ivy/functional/backends/tensorflow/linear_algebra.py
```diff
@@ -186,8 +186,6 @@ def matrix_norm(
             ret = x
         else:
             ret = tf.reduce_sum(tf.linalg.svd(x, compute_uv=False), axis=-1)
-    elif ord == "fro":
```
djl11, line 189: we need to support "fro" as shown [here](https://data-apis.org/array-api/2021.12/extensions/generated/signatures.linalg.matrix_norm.html?highlight=matrix_norm).

ivy/functional/backends/numpy/linear_algebra.py
```diff
@@ -97,6 +97,8 @@ def matrix_norm(
     out: Optional[np.ndarray] = None,
 ) -> np.ndarray:
     ret = np.linalg.norm(x, ord=ord, axis=(-2, -1), keepdims=keepdims)
+    if not isinstance(ret, np.ndarray):
```
djl11, line 100: we should use [this decorator](https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/numpy/elementwise.py#L32) for this. Maybe you can move it into a new `helpers.py` instead, located at `ivy/functional/backends/numpy/helpers.py` rather than having it in `elementwise.py`, it's a common problem for numpy functions.

djl11 (changes requested):

ivy/functional/backends/tensorflow/linear_algebra.py
JLU-Neal: Actually, the "fro" is still supported here. Since when the variable ord equals "fro", it will still be passed to tf.linalg.norm in line 190 (latest commit).
In the previous commit, when ord equals "fro", see line 190 (previous commit):
`
ret = tf.linalg.norm(x, 2, axes, keepdims)
` 
the second positional arg is for ord, but it uses a constant value 2.

ivy/functional/backends/numpy/linear_algebra.py
JLU-Neal: Okay, that's great!

ivy/functional/backends/tensorflow/linear_algebra.py
djl11: ah yes, thanks for explaining!

djl11 (approved):
lgtm!

PR #2790, opened 2022-08-08
DefaultDevice.__init__

Jayani:
Formating function __init__

### DIFF
--- a/DefaultDevice.py
+++ b/DefaultDevice.py
@@ -1,0 +1,43 @@
+class DefaultDevice:
+    """"""
+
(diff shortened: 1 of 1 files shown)


Jayani:
#2790

MarShaikh:
Hi @Jayani, thanks for the commits, but it seems that the changes you made were already pushed before you even started this PR. I'll be closing this PR then. So sorry for the inconvenience caused. The example looks good though, so you can still submit this PR to the link you received earlier. Thanks again for the efforts! 🙂

PR #2800, opened 2022-08-09
Implemented divide function for tensorflow-frontend

kanishkaditya:
removed the .idea folder and added the test function for Divide.


kanishkaditya:
found an issue fixing and re making a request.

PR #2844, opened 2022-08-10
fix det NumPy backend

### LINKED ISSUE #2839: det() inconsistent behaviour for NumPy
```det()``` should always return an ivy array as mentioned in the docs. This works fine for all the backends except for NumPy, the following code will return a number instead of a ivy array:

```
>>> ivy.set_backend('numpy')
>>> x = ivy.array([[0.0]])
>>> ivy.det(x)
0.0
>>> ivy.array([[0.0, 0.0], [0.0, 0.0]])
>>> ivy.det(x)
ivy.array(0.)
```
in both cases it should return ``` ivy.array(0.) ``` but it fails for the first example.

**PR to fix the issue:** #2844

CatB1t:
fix #2839. 2 tests used to fail because of this issue:
1. core ```test.linalg```, the edge case that causes it to fail isn't handled probably though.
2. frontend ```test.lingalg```.


JLU-Neal:
Hi!
I have encountered this issue before, and Dan suggested I use the wrapper @_handle_0_dim_output instead of adding another if statement. See this [commit](https://github.com/unifyai/ivy/pull/2779/commits/1c189382b900ade0181f2e5523db10390719ba1d).

CatB1t:
Hello @JLU-Neal thanks for pointing this out! I've updated it to use the helper method instead :)

PR #2856, opened 2022-08-10
fixes for `small_value_safety_factor`

rush2406:
There's no problem with using `None` in hypothesis generations and it generates values within the bounds. The tests were failing because of specified `min_value` and `max_value` in some of the tests such as [example](https://github.com/unifyai/ivy/blob/6849501d26bf60988aee894c5f581b12f69cf2cc/ivy_tests/test_ivy/test_functional/test_core/test_random.py#L141). This has been changed to not generate `max_neg_val` and `min_pos_val` for such cases and instead directly draw based on the specified values.


djl11 (approved):
I've not had the chance to go through in lots of detail, but if you're confident that this fixes the issues then I'm happy for you to merge it :)

PR #2893, opened 2022-08-11
ivy Variable __repr__

Ishticode:
Distinct representation for Variable class. Feel free to close @djl11 if unnecessary but I found it can be helpful sometimes. Will lead to some easy-to-fix docstring test failures I suppose.


djl11 (approved):
lgtm

PR #2898, opened 2022-08-11
add roll frontends.torch

PatternFinder:
link to issue: #2897


Oxbeefbabe:
Your test is failing. You should check test-frontend-torch / run-nightly-tests (pull_request) Line 64

PatternFinder:
> Your test is failing. You should check test-frontend-torch / run-nightly-tests (pull_request) Line 64. See if you can get it fixed :)

Hi @Oxbeefbabe :) hmm, so I don't know which helpers function gives me the ability to draw both an int and a shared shape tuple(int), so I wrote a st.composite myself, donno if it will work. Could you please have a look? thanks :)

Oxbeefbabe:
Hypothesis can be quite tricky at times. We try to avoid making redundant composite functions wherever possible. I actually worked on this recently, here is a set of data generation for the backend version of roll that will provide the same functionality. (remember that you need to provide shift and roll!) Check the main body of the function as well as I need to unpack the values I generated. You will also need to check what other variables you need /don't need to generate, as the front end and back end functions have different requirements.

PatternFinder:
@Oxbeefbabe you're right, donno what was I think :) helpers.get_axis already returns both ints and tuple(ints); and backend.roll auto applies the situation where shift: int and axis:tuple(int); so helpers.get_axis should suffice?

Oxbeefbabe:
The thing to remember with Roll is that if the shift is a tuple, then the axis must be a tuple of the same length. I had to make some changes to the get_axis function (that haven't hit the master branch yet) that make using the function a bit easier. You are allowed to drop some of the generated values, so you can see the approach I took for the axis was to always generate a tuple the same length as shift. 

Because you are working on a frontend torch function you should check the test-frontend-torch / run-nightly-tests (pull_request) workflow. Errors regarding your code will likely appear here. There are also some other errors not related to your code that you don't need to worry about. Having a look at the output of the workflow I mentioned shows that you do have an error being raised, likely due to your data generation.

PatternFinder:
> The thing to remember with Roll is that if the shift is a tuple, then the axis must be a tuple of the same length. I had to make some changes to the get_axis function (that haven't hit the master branch yet) that make using the function a bit easier. You are allowed to drop some of the generated values, so you can see the approach I took for the axis was to always generate a tuple the same length as shift.
> 
> Because you are working on a frontend torch function you should check the test-frontend-torch / run-nightly-tests (pull_request) workflow. Errors regarding your code will likely appear here. There are also some other errors not related to your code that you don't need to worry about. Having a look at the output of the workflow I mentioned shows that you do have an error being raised, likely due to your data generation.

ok will check. btw, the fact that we have `shape=st.shared(helpers.get_shape(min_num_dims=1), key="shape"),` on both `shift=helpers.get_axis` and `axis=helpers.get_axis` should return the same shape no?

PatternFinder:
ok, question. I checked `test-frontend-torch / run-nightly-tests` as you suggested, and found the following fails:

```
Falsifying example: test_torch_roll(
    dtype_and_values=('float32', [[[-0.0, -0.0]]]),
    shift=(0,),
    axis=(0, 2),
```

but since we used st.shared() on all 3 `dtype_and_values.shape`, `shift` and `axis`, why did hypothesis still drew different values for `shape`?

Oxbeefbabe:
Currently, the issue is not with the shape being used, but that the length of shape and axis (generated as tuples) are not of equal length, which the roll function requires. You need to have them generate as an equal length to one another. You can do some post-processing of the generated values in the body of the test function before passing them to the test_frontend_function helper.

PatternFinder:
Hi @Oxbeefbabe as you suggested, I've added some post-processing to `roll`'s test code; `test-frontend-torch / run-nightly-tests (pull_request)` doesn't show any errors pertaining to `roll` now. Please have a look thanks :)

Oxbeefbabe:
Looks great! Closing and merging your changes now :)

PatternFinder:
great, thanks :) have a wonderful day!

PR #2915, opened 2022-08-11
Adding tanh activation function for torch framework, addresses the issue #2686

bsovdat:
I think the function works properly, but it does get some errors when testing, but the errors are the same as for other already approved activation functions!

### DIFF
--- a/ivy/functional/frontends/tensorflow/raw_ops.py
+++ b/ivy/functional/frontends/tensorflow/raw_ops.py
@@ -34,6 +34,10 @@ def atan(x, name="atan"):
     return ivy.atan(x)
 
 
+def BitwiseAnd(*, x, y, name="BitwiseAnd"):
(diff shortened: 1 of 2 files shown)


bsovdat:
@Ishticode hi i am not sure how to request a review, so i Will Just tag you.

Ishticode (changes requested):
Hi @bsovdat Thanks for the tag and the PR. Mostly complete. Some minor changes requested. Could you kindly exclude `.idea` folder and files which might have been accidently added. Also could you make the `out` argument keyword-only by adding `*` in the function definition as in the official pytorch documentation. lmk for any queries and request my review when done. Thank you very much. 😊

bsovdat:
@Ishticode how can i re-open this pull request and how can i exclude certain files from pull request?

PR #2927, opened 2022-08-12
absolute #2926

JannisEbling:


JannisEbling:
Besides the function absolute, I found a small bug in the NumPy arccos function that I resolved.

PR #2939, opened 2022-08-12
Safety factor implementation in `helpers.floats` strategy

simonetgordon:


ivy_tests/test_ivy/helpers.py
```diff
@@ -513,6 +515,254 @@ def var_fn(x, *, dtype=None, device=None):
...
+            and max_value < lim_float32 * safety_factor
+        ):
+            dtype = "float32"
+        else:
+            dtype = "float64"
+
+    # dtype and min and max value mismatch?
+    if dtype == "float16":
+        if min_value is not None and max_value is not None:
+            min_value = float_of(min_value, 16)
```
simonetgordon, line 627: Need to add this for other dtype cases

PR #2903, opened 2022-08-11
fn_name -> fn_tree

jkeane508:
- Update to frontend test helper
- Update to frontend tests
- Update to frontend test doc

Issue #1515, opened 2022-06-18
Add Non-linear activation functions to `jax.lax` Frontend


Aarsh2001:
Add [Non-linear activation](https://jax.readthedocs.io/en/latest/jax.nn.html) functions to JAX frontend:

_

>Please keep in mind that the proper way to link an issue to this list is to comment "- [ ] #issue_number" while the issue's title only includes the name of the function you've chosen.

_

- [x] #2718 
- [x] #3126
- [x] #2963
- [x] #3030
- [x] #3130
- [x] #3136
- [x] swish
- [x] #3148
- [x] #2823 
- [x] #4260
- [x] #4261
- [x] #3720
- [x] #3469
- [x] #3536
- [x] #3477
- [x] #4919
- [x] #2936
- [x] #3149
- [x] #3015
- [x] #3146
- [x] #3539
- [x] #3178
- [x] #2899

VikramxD:
- [ ] #2718

VikramxD:
- [ ]  #2821

VikramxD:
- [ ] #2823

PatricYan:
- [ ] #2899.

VikramxD:
- [ ] #2936

VikramxD:
- [ ] #2963

PatricYan:
> 



> * [ ]  [one_hot #2899](https://github.com/unifyai/ivy/issues/2899).

#2974

PatricYan:
- [ ] #3015

PatricYan:
- [ ] #3030

Unevilicorn:
- [ ] #3126

Unevilicorn:
- [ ] #3129

Unevilicorn:
- [ ] #3130

Unevilicorn:
- [ ] #3136

sherry30:
- [ ] #3146

sherry30:
- [ ] #3148

sherry30:
- [ ] #3149

NikoSokratous:
- [ ]  #3157

NikoSokratous:
- [ ] #3159

Unevilicorn:
- [ ] #3178

VikramxD:
- [ ] #3235

VikramxD:
@Aarsh2001 can you add elu function to this list ?

Unevilicorn:
- [ ] #3469

Unevilicorn:
[...]

PR #2942, opened 2022-08-12
gelu added to jax frontend 

VikramxD:


VikramxD:
pr for #2936 , belonging to the todolist issue #1515
1. implemented gelu activation function in ivy/functional/frontends/jax/nn/non_linear_activations.py 
2. Added tests for gelu activation in  ivy/functional/frontends/jax/nn/non_linear_activations.py
3.  corrected parameter fn_name to fn_tree in the function helpers.test_frontend_function  in ivy_tests/test_ivy/test_frontends/test_jax/test_activations.py

Daniel4078:
Hello, can you tell me why you use fn_tree as the parameter name instead of fn_name while there is no such parameter for helpers.test_frontend_function? Thank you.

VikramxD:
> Hello, can you tell me why you use fn_tree as the parameter name instead of fn_name while there is no such parameter for helpers.test_frontend_function? Thank you.

This was in references to changes made in the pr #2903 , fn_name has been changed to fn_tree , if we run the test frontend function without changing the name it throws the error of missing argument fn_tree

Daniel4078:
> > Hello, can you tell me why you use fn_tree as the parameter name instead of fn_name while there is no such parameter for helpers.test_frontend_function? Thank you.
> 
> This was in reference to changes made in the pr #2903 , fn_name has been changed to fn_tree , if we run the test frontend function without changing the name it throws the error of missing argument fn_tree , if you refer the failed frontend tests made in this pr , the main reason they are failing is because they are following the old fn_name parameter , hence the error thrown : TypeError: test_frontend_function() missing 1 required keyword-only argument: 'fn_tree'

Thank you for your explaination as I just found that the branch I mainly working on has an outdated version of helpers.py.

VikramxD:
> > > Hello, can you tell me why you use fn_tree as the parameter name instead of fn_name while there is no such parameter for helpers.test_frontend_function? Thank you.
> > 
> > 
> > This was in reference to changes made in the pr #2903 , fn_name has been changed to fn_tree , if we run the test frontend function without changing the name it throws the error of missing argument fn_tree , if you refer the failed frontend tests made in this pr , the main reason they are failing is because they are following the old fn_name parameter , hence the error thrown : TypeError: test_frontend_function() missing 1 required keyword-only argument: 'fn_tree'
> 
> Thank you for your explaination as I just found that the branch I mainly working on has an outdated version of helpers.py.

No issues it failed for me too i was able to work on a local working fix in helpers.py's nest.py function but found out that @JLU-Neal had already patched it so i just changed the tests accordingly.

Daniel4078 (changes requested):
I suppose that the as_variable_flags and native_array_flags should not be generated randomly but I am not sure now.
Also, ivy.gelu has a optional out argument that your front_end wrapper and test ignored.

VikramxD:
> I suppose that the as_variable_flags and native_array_flags should not be generated randomly but I am not sure now.
> Also, ivy.gelu has a optional out argument that your front_end wrapper and test ignored.

the frontend test and the implementation has been done according to the jax implementation of the gelu function (https://jax.readthedocs.io/en/latest/_autosummary/jax.nn.gelu.html) which is why the function is not following the optional out argument in frontend function and test  , as for the as_variable_flags and native_array_flags in the helpers.test_frontend_function if we fix those tests wouldn't that hurt the testing because the generated data would be limiting to the fixed cases we assert

Daniel4078 (approved):
Then I think there is nothing wrong here.

PR #2944, opened 2022-08-12
linear_algebra Fix up

MarShaikh:
Partial fix for the tests in the linear algebra submodule + minor fix for copy_nest by changing positional args to keyword args.


ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -376,7 +396,7 @@ def test_det(
 
 # eigh
 @given(
-    dtype_x=_get_dtype_and_matrix(symmetric=True),
+    dtype_x=_get_dtype_and_symm_matrix(),
```
djl11, line 399: please don't create a new helper, given that `_get_dtype_and_matrix` already supports symmetric matrices. You're welcome to modify the implementation of this if you want, but if you do the `(x + x.T)/2` approach then this should be done in NumPy rather than Ivy.

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -703,7 +726,7 @@ def test_solve(
         max_value=50,
         min_num_dims=2,
     ),
-    num_positional_args=helpers.ints(min_value=0, max_value=1),
```
djl11, line 706: we should be using our own helpers here

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -1066,7 +1089,10 @@ def test_svd(
     if results is None:
         return
 
-    ret_np_flat, ret_from_np_flat = results
+    ret, ret_from_np = results
+    # flattened array returns
+    ret_np_flat = helpers.flatten(ret=ret)
```
djl11, line 1094: this is not needed because of `return_flat_np_arrays=True`

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -412,18 +432,21 @@ def test_eigh(
     if results is None:
         return
 
-    ret_np_flat, ret_from_np_flat = results
+    ret, ret_from_np = results
+    # flatten results
+    ret_flatten = helpers.flatten(ret=ret)
```
djl11, line 437: this is not needed because of `return_flat_np_arrays=True`

djl11 (changes requested):

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
MarShaikh: Oh okay sure @djl11 it seems like this functionality was added after I had already implemented my function. I didn't want to clutter the first function so I created another one, but the new approach is great too! Will make these changes.

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
MarShaikh: Okay, will verify this again, as it seems like something that was added after I made my changes because earlier the function wasn't returning flattened arrays. Thanks!

djl11:
The only non-superficial change is `rtol_=1e-3`, therefore I would recommend just pushing this small change to master without a PR if it is needed. Otherwise, I think we can treat this submodule as being complete.

PR #2945, opened 2022-08-12
Backend and test fixes for statistical.py

zaeemansari70:


ivy/functional/backends/torch/statistical.py
```diff
@@ -90,13 +90,12 @@ def min(
 
 
 def prod(
-    x: torch.Tensor,
+    x: torch.tensor,
```
johanjino, line 93: I believe `torch.tensor`  is a function which creates a tensor while `torch.Tensor` is the main tensor class, we dont want to change that. Curious to know, did this cause in difference🤔?

Same here [line 98](https://github.com/unifyai/ivy/pull/2945/files#diff-47b53a6f2bcf0391135291ca6206d31e6563b5f194e41427c79b17cf26c720f8R98)

zaeemansari70: Nope, it should be Tensor, you're right, was trying something out, Forgot to revert the changes. 
Just fixed it.
Thanks!

johanjino:
I just joined the unsupported dtypes for `std()`, it was redundant

Issue #2600, opened 2022-08-03
tanh


kanakmi:
Hi team
I am willing to add tanh function to the Tensorflow frontend. I have checked that there are no open issues and pull requests for the same.

1Doomdie1:
Can you please provide the main issue ID from where you took this function?
Thank you!

kanakmi:
Hey @1Doomdie1 
I know this function is not listed in one of the main issue IDs but can I still add it? Because eventually, it has to be added right?

kanakmi:
Hey @1Doomdie1 
This function is actually a part of #1565 
I will be making a PR soon.
Meanwhile, please change the labels to Single Function and TensorFlow Frontend.

PR #2966, opened 2022-08-13
Added tanh to TensorFlow Frontend

kanakmi:
This PR addresses issue #2600 which is tracked in issue #1565


ankitdsi2010 (changes requested):
Add out parameter in the function and then make the necessary changes in the test

Issue #7, opened 2022-01-19
Add Tensor functions to PyTorch Frontend


djl11:
Add [Tensor functions](https://pytorch.org/docs/stable/torch.html#tensors) to PyTorch frontend:

_

>Please keep in mind that the proper way to link an issue to this list is to comment "- [ ] #issue_number" while the issue's title only includes the name of the function you've chosen.

_

- [x] #2865 
- [x] #9903
- [x] #6830
- [x] #4975
- [x] #4977
- [x] #4976

agniiyer:
#3298

AnnaTz:
- [ ] #4975

AnnaTz:
- [ ] #4976

AnnaTz:
- [ ] #4977

AnnaTz:
- [ ] #4980

artykov1511:
#6830

HadeerArafa108:
- [ ] #9903

rumbleFTW:
Hey can I work on this issue?
Additionally, I'm a newbie contributor here, so might be a dumb question to ask, but-
Where do I implement these functions? There's no directory as pytorch frontend

AnnaTz:
> Hey can I work on this issue? Additionally, I'm a newbie contributor here, so might be a dumb question to ask, but- Where do I implement these functions? There's no directory as pytorch frontend

Hi there! All the functions in this ToDo list are already implemented (they are linked to issues and checked already as you can see above). Feel free to choose a free function from another ToDo list. 
I think you'll find this [guide](https://lets-unify.ai/ivy/contributing/open_tasks.html#frontend-apis) useful. It also explains where to add the functions.

rumbleFTW:
Thank you!

PR #2968, opened 2022-08-13
Pytorch Frontend: is_tensor function

Avinashpathapati:
This PR adds the is_tensor function to the Pytorch Frontend, related to the issue https://github.com/unifyai/ivy/issues/2865.

This is one of sub-tasks https://github.com/unifyai/ivy/issues/7


MarShaikh (changes requested):
Took a look at the code, for the backends, it is best to not include `is_tensor` for all the backends as Ivy already supports `is_array` function that will check for all the functions. Would recommend removing all the function instances from the backends as well as Ivy and just replacing the call in func `is_tensor` with `ivy.is_array` instead of `ivy.is_tensor`. Thanks, request for a review and I'll take a look again!

MarShaikh (changes requested):
Sorry for the delayed review. Could you fix the error in the `test-frontends-torch`? As well as fix the lint error. Would recommend using `pre-commit`(you can find more info in our docs) to fix linting errors! Thanks.

Avinashpathapati:
@MarShaikh this is a gentle reminder for reviewing the PR as I made the requested changes.

MarShaikh (approved):
LGTM!

PR #3008, opened 2022-08-14
torch negative

ytg32:


ytg32:
@djl11 Can someone check the pull request? It seems @chie2727 is offline.

PR #3010, opened 2022-08-14
added hypothesis framework setting functionality

Aarsh2001:
This enforces a hard constraint on the tests, meaning that we should now be adding the `handle_cmd_line_args` decorator to each test. Maybe we can have another decorator to handle just this but i thought since we are doing every cmd line arg here, why not sample the `framework` in there as well. Let me know your thoughts on this, thanks.


djl11 (changes requested):
Looks great to me. As discussed, we just need to fix the issue where `num_examples` will exponentially scale due to the double use of hypothesis `@given`, and we will also need to add the decorator to all tests. Design-wise I think it's good though.

djl11 (changes requested):
as discussed offline, you should change so that the global default is `max_examples=1`, and then we can remove all of these extra decorators.

djl11 (approved):
lgtm!

PR #3027, opened 2022-08-15
Add transpose frontends.torch

PatternFinder:
issue: #2703


ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
```diff
@@ -26,3 +26,14 @@ def swapdims(input, dim0, dim1):
     "uint32",
     "uint64",
 )
+
+
+def transpose(input, dim0, dim1):
+    return ivy.swapaxes(input, dim0, dim1)
+
+
+transpose.unsupported_dtypes = (
```
jkeane508, line 35: Are you sure that uint dtypes cannot be used for this function?

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
```diff
@@ -26,3 +26,14 @@ def swapdims(input, dim0, dim1):
     "uint32",
     "uint64",
 )
+
+
+def transpose(input, dim0, dim1):
+    return ivy.swapaxes(input, dim0, dim1)
```
jkeane508, line 32: We would use `ivy.matrix_transpose` here not swap axes

jkeane508 (changes requested):

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
PatternFinder: according to ivy.swapaxes, the uint16, uint32, uint64 does seem to be unsupported

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
jkeane508: Having looked at the docs, I would agree.

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
jkeane508: Are you sure? I've just ran it locally and it is accepting arrays with uint dtypes.

PatternFinder: This is the definition from `(https://github.com/unifyai/ivy/tree/master/ivy/functional/backends/torch)/manipulation.py`
it says `swapaxes.unsupported_dtypes` doesn't support uint types; did I misunderstood this or

```
def swapaxes(
    x: torch.Tensor, axis0: int, axis1: int, *, out: Optional[torch.Tensor] = None
) -> torch.Tensor:
    ret = torch.transpose(x, axis0, axis1)
    return ret


swapaxes.unsupported_dtypes = (
    "uint16",
    "uint32",
    "uint64",
)
```

jkeane508: Yep that is the backend PyTorch implementation of `swapaxes`. However, the frontend implementation is independent of this and `ivy.swapaxes()` has been designed as to accept any relevant dtype regardless of the framework. However, I have raised this with the team and will get back with an answer!

jkeane508 (changes requested):
As mentioned above :)

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
PatternFinder: ok got it. so what's our policy here regarding the `frontend_function.unsupported_dtypes = ()`; should I keep including them in the frontend or not include them?

PatternFinder:
> As mentioned above :)

ok I removed the `unsupported_types`; what else should I do for this PR? :)

jkeane508 (approved):
You can just resolve conflicts and provided everything is passing, we can merge :)

PatternFinder:
ok resolved conflicts

Issue #1210, opened 2022-05-27
sum


abubakaralsadiq:

omafka:
@1Doomdie1 Hello, I see no activity on this issue for a long time. Can I take it?

aryans1204:
@1Doomdie1 Hello,
 I dont see any activity on this issue since past 13 days, could I take it?

omafka:
> @1Doomdie1 Hello, I dont see any activity on this issue since past 13 days, could I take it?

Hello, @aryans1204 I've added a pull request https://github.com/unifyai/ivy/pull/3031 for this issue and now I'm waiting for review from @LINBOTANG123

ivy-seed:
This issue has been labelled as stale. If you would like to continue working on this issue, then please add another comment or create a PR . If a PR has already been created which refers to this issue, then you should explicitly mention this issue in the relevant PR. For more information please refer to this page in our documentation: https://lets-unify.ai/ivy/contributing/1_the_basics.html#todo-list-issues

PR #3031, opened 2022-08-15
Reformat sum #1210

omafka:


kimbring2:
@omafka Thank you, you modify them correctly (:

Issue #3007, opened 2022-08-14
softplus


ashmitkx:
softplus function for the PyTorch frontend.

PR #3038, opened 2022-08-15
add softplus to PyTorch frontend

ashmitkx:
#3007

### DIFF
--- a/ivy/functional/frontends/torch/non_linear_activation_functions.py
+++ b/ivy/functional/frontends/torch/non_linear_activation_functions.py
@@ -27,3 +27,12 @@ def gelu(input, approximate="none"):
 
 
 gelu.unsupported_dtypes = {"torch": ("float16", "bfloat16")}
+
+
+def softplus(input, beta=1, threshold=20):
+    arg = ivy.multiply(input, beta)
+    ivy_out = ivy.divide(ivy.softplus(arg), beta)
+    return ivy.where(ivy.greater(threshold, arg), input, ivy_out)
+
+
+softplus.unsupported_dtypes = ("float16",)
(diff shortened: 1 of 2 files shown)


MarShaikh (changes requested):
I had a look at your code. It looks good to me, but the test_frontend check in the CI is failing. Please get it to pass and I'll have a look again. Thanks.

ashmitkx:
Hey. Thanks so much for reviewing my code, I had been waiting for long.

I'm actually not sure why [the error](https://github.com/unifyai/ivy/runs/7836851475?check_suite_focus=true#step:4:605) is happening. Do you have any clue as to why `AssertionError: -0.0 != inf` is being thrown?

MarShaikh:
> Hey. Thanks so much for reviewing my code, I had been waiting for long.
> 
> I'm actually not sure why [the error](https://github.com/unifyai/ivy/runs/7836851475?check_suite_focus=true#step:4:605) is happening. Do you have any clue as to why `AssertionError: -0.0 != inf` is being thrown?

Sorry for the delay! I'm assuming the error is being thrown due to `tf` returning `inf` for a specific value of `x` whereas `torch` returns `-0.0` for that value. I think a fix would be to limit the range of values being tested, but I recommend just diving deeper in both the `torch`'s `softplus` function and `tf` `softplus` and finding the cause.

ashmitkx:
Alright, I'll have a look and let you know.

ashmitkx:
Hey @MarShaikh, I'm currently having trouble debugging the issue.\
I tried stepping through the testing function, but couldn't exactly understand what is happening with the input.

Do you have any tips on how I can better trace the issue in the code?

MarShaikh:
> Hey @MarShaikh, I'm currently having trouble debugging the issue. I tried stepping through the testing function, but couldn't exactly understand what is happening with the input.
> 
> Do you have any tips on how I can better trace the issue in the code?

It seems like the outputs you get from the Tensorflow and the output you get from the function itself are different. I would suggest using the failure input as shown in the error stack, you try to test the `softplus` function with different backends! It is possible that you might need to change the limits of your inputs as one of the outputs was `-inf`, maybe also read tf documentation.

ivy-seed:

PR #3041, opened 2022-08-15
Median in numpy frontend

### LINKED ISSUE #2999: median
#1529

5hv5hvnk:
closes #2999


ivy/functional/frontends/numpy/statistics/median.py
```diff
@@ -0,0 +1,23 @@
...
+):
+    if dtype:
+        a = ivy.astype(ivy.array(a), ivy.as_ivy_dtype(dtype))
+    ret = ivy.median(a, out=out)
+    if ivy.is_array(where):
+        ret = ivy.where(where, ret, ivy.default(out, ivy.zeros_like(ret)), out=out)
+    return ret
+
+
+medain.unsupported_dtypes = {"torch": ("float16",)}
```
Daniel4078, line 23: what is "medain"? Please fix this, thank you.

ivy/functional/frontends/numpy/statistics/median.py
```diff
@@ -0,0 +1,23 @@
+import ivy
+
+
+def median(
```
Daniel4078, line 4: This is the input arguments that you are supposed to recreated I suppose: numpy.median(a, axis=None, out=None, overwrite_input=False, keepdims=False)
Also, ivy.median is not implemented yet. So you may convert your existing GitHub issue to request for the function to be added to Ivy. Meanwhile, you can select another frontend function to work on from the ToDo list!

Daniel4078 (changes requested):
Well, I have to say that I do not understand your implementation here and there are some problems to fix. Thank you.

ivy/functional/frontends/numpy/statistics/median.py
Daniel4078: Or you may try using the other existing ivy functions to implement this function instead.

Daniel4078:
by the way, please write a test for it in this directory: ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics while following the naming convention of Ivy.

ivy/functional/frontends/numpy/statistics/median.py
5hv5hvnk: @Daniel4078 where can I add the median function in ivy itself (which file)?

Daniel4078: Firstly, you do not need to implement ivy.median yourself as it is someone else's job. Secondly, if it is implemented, I suppose it should be in ivy\functional\ivy\statistical.py.
But as I have said, you can just change your issue into a request of implementing such function and let other professional developers deal with the problem.

5hv5hvnk: Alright, is there a function to flatten an ivy array?

5hv5hvnk:
> by the way, please write a test for it in this directory: ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics while following the naming convention of Ivy.

Sure give me some time, If you can go through latest commit can you tell If the function seems fine to you

ivy/functional/frontends/numpy/statistics/median.py
Daniel4078: I found a flatten function in the helper functions for tests. I suppose you should not call the function, but you can use its code directly instead.
def flatten(*, ret):
    """Returns a flattened numpy version of the arrays in ret."""
    if not isinstance(ret, tuple):
        ret = (ret,)
    ret_idxs = ivy.nested_indices_where(ret, ivy.is_ivy_array)
    return ivy.multi_index_nest(ret, ret_idxs)

ivy/functional/frontends/numpy/statistics/median.py
```diff
@@ -0,0 +1,51 @@
...
+    if overwrite_input:
+        med = median_overwrite(a, axis, keepdims)
+    else:
+        med = _median(a, axis)
+
+    if out is not None:
+        out = med
+        return
+
+    return med
```
Daniel4078, line 51: I found that you forget about the line that states the unsupported_dtypes. And you should have a blank line at the end of each file according to the formatting requirements.

Daniel4078:
just a small tip, you can request a fresh review from a reviewer by click the refresh icon in the Reviewers section of the sidebar of the Conversation tab.

Daniel4078 (changes requested):
I suppose this median function should be placed in the average_and_variances.py file, please change that.

Daniel4078:
You should only use ivy functions in your implementation instead of functions from other frameworks as the function is supposed to work without relying on other frameworks. Thank you.

Daniel4078:
I just got an update on the policy of how to deal with this pull request's situation.
In a nutshell, the contributor can either implement the function as a composition, with a “#ToDo” comment added in the implementation, or they can add the “Next Release” label to the “Sub-Task” issue, and then choose another frontend function to work on.
Thank you.

5hv5hvnk:
@Daniel4078 I have replaced `numpy.flatten` with `.reshape(1,-1)` using ivy removing bumpy dependency in code

Daniel4078:
I can only certainly decide if the current implementation has all the desired behaviors when you produce a test for it. But for now, I think it looks fine to me.

Daniel4078:
Speaking about testing, here is a change about it:
The use of the  decorator handle_cmd_line_args has been changed a little. So now we're wrapping the  given decorator, instead of the test function. This is done so as to set the backend before assigning any values in the given decorator. We should now be using ivy.current_backend_str().valid_dtypes inside our helper functions, instead of ivy_np.valid_dtypes(i'm currently in the process of updating, this). Also, please don't pass in anything for as_variable, native_array, container, with_out, instance_method inside given, since that is being handled by the decorator, and it will only overwrite the values for these flags, which we don't want.

5hv5hvnk:
Hey sorry I got busy with my final submission can I continue working in this ?

Daniel4078:
> Hey sorry I got busy with my final submission can I continue working in this ?

of course you can, so firstly can you fix the conflicting file so the automated checks can start? Thank you.

5hv5hvnk:
sure let me resolve conflicts and write the tests

PR #3052, opened 2022-08-15
numpy frontend reshape

gunesevitan:
#3051


MarShaikh (changes requested):
Looks good, as for the new helpers you created, you can take a look at the backend tests for `test_reshape` and use the already existing helpers for `dtype_values` and `reshape_shapes`. Thanks!

PR #3053, opened 2022-08-15
activations Fix Up

sherry30:


ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: `print("alpha: {} ".format(alpha))` debug code below.

djl11 (changes requested):
small change, otherwise looks good.

PR #3071, opened 2022-08-16
Numpy's equal logic operation frontend

teddywaweru:
I haven't carried out the tests, due to some issues on my current setup. I've basically replicated the add operations since it also deals with two arrays.


teddywaweru:
Docstring format for Numpy front-end not yet finalised. Failing runs in jax & pytorch were not meant to be addressed in this PR.

teddywaweru:
closing this PR because I launched a different one for the same purpose #3071

PR #3080, opened 2022-08-16
Fix for floats generation in ivy tests helper

rush2406:


VedPatwardhan (changes requested):
Hi @rush2406, thanks for making the changes, I have pushed another update to enable interval checking before we pass it to `st.floats`, please pull the latest code on master into your branch, so that we could check if this works with passing `min_value` and `max_value` to the `dtype_and_values` helper. Thanks 🙂

VedPatwardhan (approved):
lgtm!

PR #3102, opened 2022-08-17
add atan2 to jax

PatricYan:
- [ ] #3091 
- add atan2 to jax


rush2406:
changes seem correct, thanks for contributing @PatricYan !

PR #3155, opened 2022-08-17
added trace linear algebra function for tensorflow frontend

RAHUL-KAD:
Issue Link
- [ ] #3153

### DIFF
--- a/ivy/functional/frontends/tensorflow/linalg.py
+++ b/ivy/functional/frontends/tensorflow/linalg.py
@@ -6,3 +6,10 @@ def det(input, name=None):
 
 
 det.unsupported_dtypes = ("float16", "bfloat16")
+
+
+def trace(input, offset):
+    return ivy.trace(input, offset)
+
+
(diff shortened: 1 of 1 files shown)


saeedashrraf:
Hello @RAHUL-KAD , thank you for your effort! Can you make sure that the implemented frontend function's parameters are equivalent to the framework's function parameters. It should be `def trace(x , name=None)` as is the case in TensorFlow's function definition [here](https://www.tensorflow.org/api_docs/python/tf/linalg/trace).

 You should also implement the test for the function by following the guide [here](https://lets-unify.ai/ivy/deep_dive/17_ivy_frontends_tests.html).

MrZilinXiao:
Hi @RAHUL-KAD! I am responsible for checking stale PRs regularly. Since this PR has been inactive for 33 days, I am wondering if this PR is still ongoing. I will now close it. In the meanwhile, you could reopen it if you believe the PR is still useful for you.

PR #3162, opened 2022-08-17
Add Non-linear activation functions to JAX Frontend (Swish)

NikoSokratous:
1. Added a test function in test_jax_nn_activations.py
2. Added a swish function in the frontend part of Jax
3. Added a swish function in the Backend part of Jax
4. Added a swish function in the Backend part of Ivy (I didn't add any docstring and examples to the function)


kimbring2:
@NikoSokratous Hello, it seems like you need to remove the dim argument from the test_frontend_function function. It gives the error(E       TypeError: swish() got an unexpected keyword argument 'dim').

PR #3171, opened 2022-08-18
Numpy equal frontend & test

teddywaweru:
Tests passed.


teddywaweru:
The numpy_equal function passes successfully. the failed checks are from other modules that do not pass the checks.

kimbring2:
@teddywaweru Hello, the code works well. Thank you for your contributuion (:

PR #3177, opened 2022-08-18
Minor fix for test_qr in linalg submodule + fixed pytest discovery fail

MarShaikh:
* Added `__init__.py` file in `test_frontends/test_tensorflow/__init__.py` as it's absence was causing PyTest discovery error in VSCode. 
* Updated `test_qr`, minor fix to include _only_ square matrices.


MarShaikh:
Created a PR instead of pushing it directly because I wanted to confirm the addition of a new `__init__.py` file for the sole purpose of pytest discovery in VSCode. Not sure if the same error was being caused elsewhere.

djl11 (approved):
thanks!

PR #3189, opened 2022-08-18
attempt to fix numpy add frontend

teddywaweru:


teddywaweru:
I noticed that the issue with the test function was the inclusion of the variable 'out=None' in the test_frontend_function.
This same issue affects the subtract, multiply & divide test functions, which cause the checks above to fail.
I'm not quite sure why they fail, here's a screenshot of the resulting error.
image

Please check & advise if removing the variable is a viable fix.
After which, I could combine all the corrections into this PR.

teddywaweru:
Closing this, as it may not be an effective solution to the test fails. During debugging, I noticed that, when comparing the output of the front-end, & the output of the library being tested, one produces the return as an `ivy.array()` while the other is an `int`, which may be the cause of the failures. Will analyze further in a different pull request.

johanjino:
@teddywaweru Thank you for trying out. Great to see people eager to fix and contribute with us 😄.

PR #3196, opened 2022-08-18
add squeeze frontends.torch

PatternFinder:
link to issue: #3194


saeedashrraf:
Hello @PatternFinder , there was a small error with the docs of `torch.squeeze` as the function does not support the `out` argument . I have added a fix for that to get the test to run properly. But, there remains another error that is detected by the test. Can you try to fix that. Let me know if you have any questions. Thank you for your effort!

PatternFinder:
I think the error was caused by the default value of dim=None that caused a mismatch with the input tensor's shape; I removed it

saeedashrraf:
Hello @PatternFinder , this looks ready for merging . Thank you for your efforts!

PR #3209, opened 2022-08-18
test_method Improvements

VedPatwardhan:


ivy_tests/test_ivy/helpers.py
```diff
@@ -873,26 +874,27 @@ def value_test(
...
-    else:
-        for ret_np, ret_from_np in zip(ret_np_flat, ret_np_from_gt_flat):
-            assert_all_close(
-                ret_np,
-                ret_from_np,
-                rtol=rtol,
-                atol=atol,
-                ground_truth_backend=ground_truth_backend,
-            )
+    if test_values:
```
djl11, line 877: this parameter seems a bit pointless, why would we ever call the function `value_test` with argument `test_values=False`?

ivy_tests/test_ivy/helpers.py
```diff
@@ -1303,69 +1321,94 @@ def test_method(
...
+            True if "v" in kwargs_np_constructor or "v" in kwargs_np_method else False
+        )
     value_test(
         ret_np_flat=ret_np_flat,
         ret_np_from_gt_flat=ret_np_from_gt_flat,
-        rtol=rtol,
-        atol=atol,
+        rtol=rtol_,
+        atol=atol_,
+        test_values=test_values,
```
djl11, line 1411: as discussed above, we should simply add `value_test` inside an `if` statement here, rather than passing `test_values` as an argument.

djl11 (changes requested):
small changes, otherwise looks good. Could you please test this on some dummy data locally to make sure it seems to be working. Once you've made the change and test locally, feel free to merge. Thanks!

ivy_tests/test_ivy/helpers.py
```diff
@@ -1120,70 +1136,110 @@ def test_method(
...
+                ret_np_from_gt_flat=ret_np_from_gt_flat,
+                rtol=rtol_,
+                atol=atol_,
+            )
+        else:
+            if type(ret_np_flat) != list:
+                ret_np_flat = [ret_np_flat]
+            if type(ret_np_from_gt_flat) != list:
+                ret_np_from_gt_flat = [ret_np_from_gt_flat]
+            assert len(ret_np_flat) == len(ret_np_from_gt_flat), (
```
djl11, line 1230: can we also do a nested type and shape test here? something like:

```
def _assert_same_type_and_shape(x, y):
    assert type(x) is type(y)
    if isinstance(x, np.ndarray):
        assert x.shape == y.shape
        assert x.dtype = y.dtype

ivy.nested_multi_map(_assert_same_type_and_shape, [ret_np_flat, ret_np_from_gt_flat])
```

djl11 (changes requested):

djl11 (approved):
lgtm!

PR #3224, opened 2022-08-19
tensorflow frontend - `nest` 

juliagsy:
related function:
- `flatten` -  only flattens tuple, list, dict. Arrays are treated as an atom, therefore returned without changes

updates:
- combined `ragged` and `sparse` checking to `_is_composite_array` and `_flatten_composite_array`

notes:
- when `expand_composites` is True, special arrays like `ragged_array` or `sparse_array` is expanded according to components or attributes.
- `ragged_array` is present only in tensorflow
- `sparse_array` is present only in tensorflow and torch, for torch I tried to recover the tf behaviour when `expand_composites` is True

Thanks!

### DIFF
--- a/ivy/functional/frontends/tensorflow/nest.py
+++ b/ivy/functional/frontends/tensorflow/nest.py
@@ -1,4 +1,3 @@
-# for review
 # global
 import ivy
 import tensorflow as tf
--- a/ivy/functional/frontends/tensorflow/nest.py
+++ b/ivy/functional/frontends/tensorflow/nest.py
@@ -5,8 +5,8 @@ import tensorflow as tf
 import torch
 
 
-def _is_sparse_array(x):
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/tensorflow/nest.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: replace repeated `if` calls with `if` `elif`

djl11: could you also remove the `if isinstance(structure, tf.RaggedTensor):` section inside the `flatten` implementation? Can tensorflow not also be handled correctly inside `_flatten_sparse_array`?

djl11 (changes requested):

djl11 (approved):
lgtm!

PR #3236, opened 2022-08-19
jax frontend - implement `while_loop` function and test

juliagsy:


ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_control_flow_operators.py
```diff
@@ -140,3 +140,47 @@ def _test_branch_2(x):
...
+        return ivy.less(x, 0, out=None)
+
+    def _test_body_fn(x):
+        return ivy.add(x, 100, out=None)
+
+    input_dtype, x = dtype_and_x
+    helpers.test_frontend_function(
+        input_dtypes=[input_dtype],
+        as_variable_flags=as_variable,
+        with_out=False,
```
Daniel4078, line 177: with_out should be passed in through the arguments of test_jax_while_loop as the @handle_cmd_line_args will fill in the with_out argument automatically.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_control_flow_operators.py
```diff
@@ -140,3 +140,47 @@ def _test_branch_2(x):
         branches=[_test_branch_1, _test_branch_2],
         operand=np.array(x, dtype=input_dtype),
     )
+
+
```
Daniel4078, line 144: please add a comment here like: "# while_loop"

Daniel4078 (changes requested):
Firstly, there is an Type error when running your test (TypeError('must be real number, not ShapedArray')). This seemed to be caused by a x = ShapedArray(bool[1]) passed to the format_float_positional function. I do not know the exact cause of the error but I suspect that it is cause by the fact that ivy.less produce an array of booleans but then the next loop expect an array of numbers
Secondly, there are some small errors that you need to fix and please check my comments below for details.

juliagsy:
Hey! Thanks for the comments! I'm one of the team member actually and I'm aware of the tests as well! Putting this on hold as I'm still thinking about it's design 😊 

You may skip this PR while searching for one to review! 😁

PR #3244, opened 2022-08-19
Frontend version

RickSanchezStoic:
Initial changes made.


ivy/functional/frontends/__init__.py
```diff
@@ -3,3 +3,90 @@
...
+        f_version = latest_version[f]
+
+    for i in list(frontend.__dict__):
+        if hasattr(frontend.__dict__[i], "version_support"):
+            orig_name = version_extractor(i, f_version)
+            if orig_name:
+                frontend.__dict__[orig_name] = frontend.__dict__[i]
+
+
+version_handler(torch)
```
djl11, line 90: `numpy` should be added too?

ivy/functional/frontends/__init__.py
```diff
@@ -3,3 +3,90 @@
 from . import jax
 from . import torch
 from . import tensorflow
+import importlib
+
+latest_version = {"torch": "1.12"}
```
djl11, line 8: other frameworks?

ivy/functional/frontends/__init__.py
```diff
@@ -3,3 +3,90 @@
...
+    f = str(frontend.__name__)
+    f = f[f.index("frontends") + 10 :]
+    try:
+        f = importlib.import_module(f)
+        f_version = f.__version__
+    except Exception:
+        f_version = latest_version[f]
+
+    for i in list(frontend.__dict__):
+        if hasattr(frontend.__dict__[i], "version_support"):
```
djl11, line 84: we should solve this without needing a `version_support` attribute. We can instead detect when `i` is a version-specific function based on the string pattern, and then update the function in the `__dict__` if it overlaps correctly with the current version; no need for this to be manually flagged.

djl11 (changes requested):

ivy/functional/frontends/__init__.py
RickSanchezStoic: Yeah, had to add this decided to get your review first.

ivy/functional/frontends/__init__.py
RickSanchezStoic: Yep, slipped. Will add.

ivy/functional/frontends/__init__.py
RickSanchezStoic: Noted

ivy/functional/frontends/torch/creation_ops.py
```diff
@@ -18,3 +18,26 @@ def full(
...
+    requires_grad=False,
+    memory_format=None
+):
+    ret = ivy.ones_like(input, dtype=dtype, device=device)
+    if requires_grad:
+        return ivy.variable(ret)
+    return ret
+
+
+def ones_like_v_0p3p0_to_0p3p1(input, out=None):
```
djl11, line 41: we will need another function to cover `<>_0p3p0_and_below`. If `0p3p0` is the first version released, or if this function also covers all earlier releases, then we should change this function name to `<>_0p3p1_and_below`

ivy/functional/frontends/__init__.py
```diff
@@ -3,3 +3,87 @@
...
+    else:
+        i = name.index("_v_")
+        e = name.index("_and_")
+        version_start = name[i + 3 : e]
+        version_start = int(version_start.replace("p", ""))
+        if version <= version_start:
+            return name[0:i]
+
+
+def version_handler(frontend):
```
djl11, line 57: this function should be given a more informative name, I would suggest `set_frontend_to_specific_version`

ivy/functional/frontends/torch/creation_ops.py
```diff
@@ -18,3 +18,26 @@ def full(
...
+    ret = ivy.ones_like(input, out=None)
+    return ret
```
djl11, line 43: no need for this extra line, can all be done in 1 line.

ivy/functional/frontends/__init__.py
```diff
@@ -3,3 +3,87 @@
 from . import jax
 from . import torch
 from . import tensorflow
+import importlib
+
+latest_version = {"torch": "1.12", "tensorflow":"2.9.0", "numpy":"1.23.2", "jax":"0.3.16"}
+
+
+def version_extractor(name, version):
```
djl11, line 11: we should make this function name more informative, such as `fn_name_from_version_specific_fn_name`.

ivy/functional/frontends/__init__.py
```diff
@@ -3,3 +3,87 @@
 from . import jax
 from . import torch
 from . import tensorflow
+import importlib
+
+latest_version = {"torch": "1.12", "tensorflow":"2.9.0", "numpy":"1.23.2", "jax":"0.3.16"}
```
djl11, line 8: Do you think it is necessary for us to manually track the latest versions of each framework? Seems like a lot of manual maintenance effort.

djl11 (changes requested):

PR #3292, opened 2022-08-20
add prod function for numpy frontend (Issue #3154)

Cybersheralt:
I am not quite sure why I am getting errors, so can you please check and give me some advices?


ivy/functional/frontends/numpy/mathematical_functions/sums_products_differences.py
```diff
@@ -0,0 +1,25 @@
+# global
+import ivy
+
+# local
+from ivy.func_wrapper import from_zero_dim_arrays_to_float
+
+
+@from_zero_dim_arrays_to_float
+def prod(
```
Daniel4078, line 9: Here you are supposed to recreate the input arguments of numpy.prod, which is shown here: https://numpy.org/doc/stable/reference/generated/numpy.prod.html
So the argument name and initial value needs to be as similar as possible without relying on the content of numpy framework.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_sums_products_differences.py
```diff
@@ -0,0 +1,59 @@
...
+from ivy_tests.test_ivy.helpers import handle_cmd_line_args
+
+
+# prod
+@handle_cmd_line_args
+@given(
+    dtype_and_x=helpers.dtype_and_values(available_dtypes=ivy_np.valid_float_dtypes),
+    dtype=st.sampled_from(ivy_np.valid_float_dtypes + (None,)),
+    where=np_frontend_helpers.where(),
+    as_variable=helpers.array_bools(),
```
Daniel4078, line 18: Please don't pass in anything for as_variable, native_array, container, with_out, instance_method inside given, since that is being handled by the @handle_cmd_line_args decorator, and it will only overwrite the values for these flags, which we don't want.

Daniel4078 (changes requested):
I found something that you need to fix for your code to work. Please check the comments below. Thank you.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_sums_products_differences.py
Daniel4078: By the way, we should now be using ivy.current_backend().valid_dtypes inside our helper functions, instead of ivy_np.valid_dtypes.

Cybersheralt: Thank you, I have changed everything you mentioned here.

ivy/functional/frontends/numpy/mathematical_functions/sums_products_differences.py
Cybersheralt: In that case I understood task correctly, but I am still having troubles with this part, because I am not sure which initial values I need to use instead of `np._NoValue()`. My current option is 
```
keepdims=False,
initinal=None,
where=True
```
but it is not working.

Daniel4078:
just a small tip, you can request a fresh review from a reviewer by click the refresh icon in the Reviewers section of the sidebar of the Conversation tab.

ivy/functional/frontends/numpy/mathematical_functions/sums_products_differences.py
Daniel4078: Well, this is a good question. My personal guess is that they all should be initially None. But I am asking other professional people in our team about this. Please be patient, thank you.

Daniel4078: I just got a confimation, you should use None in this case.

ivy/functional/frontends/numpy/mathematical_functions/sums_products_differences.py
```diff
@@ -0,0 +1,26 @@
...
+        x,
+        axis=None,
+        dtype=None,
+        out=None,
+        keepdims=None,
+        initinal=None,
```
Daniel4078, line 15: it should be "initial" not "initinal" here.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_sums_products_differences.py
```diff
@@ -0,0 +1,58 @@
...
+        num_positional_args=num_positional_args,
+        native_array_flags=native_array,
+        fw=fw,
+        frontend="numpy",
+        fn_tree="prod",
+        x=np.asarray(x, dtype=input_dtype[0]),
+        axis=None,
+        dtype=None,
+        out=None,
+        keepdims=np._NoValue,
```
Daniel4078, line 54: you should generate all valid inputs using hypothesis for these optional arguments or the test will not cover all the cases

ivy/functional/frontends/numpy/mathematical_functions/sums_products_differences.py
```diff
@@ -0,0 +1,26 @@
...
+import ivy
+
+# local
+from ivy.func_wrapper import from_zero_dim_arrays_to_float
+
+
+@from_zero_dim_arrays_to_float
+def prod(
+        x,
+        axis=None,
```
Daniel4078, line 11: in order to seperate positional arguments from keyword arguments, you should add "/" and "*" like below:
The use case will determine which parameters to use in the function definition:
def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
As guidance:
Use positional-only if names do not matter or have no meaning, and there are only a few arguments which will always be passed in the same order.
Use keyword-only when names have meaning and the function definition is more understandable by being explicit with names.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_sums_products_differences.py
```diff
@@ -0,0 +1,58 @@
...
+    where,
+    as_variable,
+    with_out,
+    num_positional_args,
+    native_array,
+    fw,
+):
+    input_dtype, x = dtype_and_x
+    input_dtype = [input_dtype]
+    where = np_frontend_helpers.handle_where_and_array_bools(
```
Daniel4078, line 35: there is something wrong in your use of handle_where_and_array_bools. it causes this problem:
E               TypeError: unsupported operand type(s) for +=: 'bool' and 'list' 
for as_variable argument and native_array argument.

Daniel4078 (changes requested):
Please take a look at my comments, thank you.

Cybersheralt:
Thank you for your comments, I have changed everything that you mentioned in them, but I still have ValueError while running the test.

Daniel4078 (changes requested):
So according to the error log, it seems like when the where argument is not None, then the initial value has to be not None as well. This caused the ValueError. So you may have a if statement that converts where's content to None when initial is generated to be None.
There is another error with the test but I think it is not your problem but caused by backend implementation: 
E       AssertionError: the return with a TensorFlow backend produced data type of int8, while the return with a numpy backend returned a data type of int32.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_sums_products_differences.py
```diff
@@ -0,0 +1,61 @@
...
+        frontend="numpy",
+        fn_tree="prod",
+        x=np.asarray(x, dtype=input_dtype[0]),
+        axis=axis,
+        dtype=dtype,
+        out=None,
+        keepdims=keepdims,
+        initial=initial,
+        where=where,
+        test_values=False,
```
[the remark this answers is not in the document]
Cybersheralt: I set `False`, because in Deep Dive [here](https://lets-unify.ai/ivy/deep_dive/17_ivy_frontends_tests.html) for NumPy it was `False`

Daniel4078:
I found another error in the test:
E       ValueError: output parameter for reduction operation multiply has the wrong number of dimensions: Found 1 but expected 0

Daniel4078:
I can not figure out how your code can cause the assertion error currently. And I do not think that you need to fix those error cause not by your implementation right?

Cybersheralt:
My task is to implement the function, write tests for that function and to verify that the tests for that function are passing.
As far as I understand, I don't need to fix errors which are caused not by my implementation.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_sums_products_differences.py
```diff
@@ -0,0 +1,61 @@
...
+    where,
+    as_variable,
+    with_out,
+    num_positional_args,
+    native_array,
+    fw,
+    keepdims,
+    initial
+):
+    input_dtype, x, axis = dtype_and_x
```
Daniel4078, line 36: I just found out, the output format of helpers.dtype_values_axis when requesting multiple arrays will be (dtype, [arrays, arrays, ...]). So this line here should be:
 input_dtype, [x, axis] = dtype_and_x

Cybersheralt: `input_dtype, [x, axis] = dtype_and_x` results in `ValueError: too many values to unpack (expected 2)`

Daniel4078: Oh, never mind, I made a mistake. Sorry.

Daniel4078:
I just found out that dtype_value_axis, doesn't produce bounded axis, the axis is not linked to the shape of the array. So it cause this error: E       numpy.AxisError: axis 1 is out of bounds for array of dimension 1
You should probably use a helper method like this one here: (https://github.com/unifyai/ivy/blob/a9742032079090925921c5f479858a870db1214c/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_operators.py#L782) to generate proper axis for the data.

Cybersheralt:
I used the function which you suggested, but modified it, because I had an error if `len(shape)` equaled to 0 (in that case `max_value` was less than `min_value`). I have added that modified function to numpy helper file, because it can be used for testing other numpy functions.

ivy/functional/frontends/numpy/mathematical_functions/sums_products_differences.py
```diff
@@ -0,0 +1,28 @@
...
+        dtype=None,
+        out=None,
+        keepdims=None,
+        initial=None,
+        where=None
+):
+    if dtype:
+        x = ivy.astype(ivy.array(x), ivy.as_ivy_dtype(dtype))
+    ret = ivy.prod(x, axis=axis, keepdims=keepdims, out=out)
+    if ivy.is_array(where):
```
Daniel4078, line 23: After reading the document of numpy.prod, I think that maybe we should use where to filter the input x array before the prod reduction is applied on the resulting array, rather than use where to filter the result of doing prod reduction on the input.

Cybersheralt: Do you mean to change it like this?

```
    if dtype:
        x = ivy.astype(ivy.array(x), ivy.as_ivy_dtype(dtype))
    if ivy.is_array(where):
        x = ivy.where(where, x, ivy.default(out, ivy.zeros_like(x)), out=out)
    ret = ivy.prod(x, axis=axis, keepdims=keepdims, out=out)
```

Daniel4078: You can have a try, may be it will solve this error:
E       AssertionError: len(ret_np_flat) != len(ret_np_from_gt_flat):
E       
E       ret_np_flat:
E       
E       [array([0], dtype=int32)]
E       
E       ret_np_from_gt_flat:
E       
E       []

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_sums_products_differences.py
```diff
@@ -0,0 +1,61 @@
...
+import ivy
+import ivy_tests.test_ivy.helpers as helpers
+import ivy_tests.test_ivy.test_frontends.test_numpy.helpers as np_frontend_helpers
+from ivy_tests.test_ivy.helpers import handle_cmd_line_args
+
+
+# prod
+@handle_cmd_line_args
+@given(
+    dtype_and_x=np_frontend_helpers.dtype_x_bounded_axis(available_dtypes=ivy.current_backend().valid_dtypes),
```
Daniel4078, line 15: Well, this line is too long as describe by the error message below.
ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_sums_products_differences.py:15:89: E501 line too long (110 > 88 characters)

Daniel4078 (changes requested):
Congrats! The test of your function passes with no errors at all. There is just this one single formatting error that you need to fix before this pull request can be merged. Thank you.

Daniel4078 (approved):
I think there is nothing else to change now. Thank you!

Cybersheralt:
Thank you very much for all your guidance and help.

PR #3305, opened 2022-08-21
Add binary accuracy for tf frontend

JLU-Neal:


ivy/functional/backends/numpy/statistical.py
```diff
@@ -34,7 +34,9 @@ def mean(
     out: Optional[np.ndarray] = None,
 ) -> np.ndarray:
     axis = tuple(axis) if isinstance(axis, list) else axis
-    return np.asarray(np.mean(x, axis=axis, keepdims=keepdims, out=out))
+    return np.asarray(
+        np.mean(x, axis=axis, dtype="float32", keepdims=keepdims, out=out)
```
JLU-Neal, line 38: The default dtype of return for numpy backend is float64, while for tensorflow is float32 (see this [tf.reduce_mean](https://www.tensorflow.org/api_docs/python/tf/math/reduce_mean) ).
However, I don't think this is an elegant solution and might cause failures for other tests. 
Any better ideas?

djl11: we should explicitly ensure that the data type is the same as that of the input array

djl11 (changes requested):
aside from the change regarding mean with a numpy backend, the rest looks good. Feel free to merge once this has been fixed!

djl11 (changes requested):
please make the changes discussed regarding the default data type.

JLU-Neal:
> aside from the change regarding mean with a numpy backend, the rest looks good. Feel free to merge once this has been fixed!

Hey Dan! 
The code you refer to is actually outdated, and it remains unchanged in the newest commit!

PR #3316, opened 2022-08-21
add round to jax

PatricYan:
- [ ] #3113 
- add round to jax

### DIFF
--- a/ivy/functional/frontends/jax/lax/operators.py
+++ b/ivy/functional/frontends/jax/lax/operators.py
@@ -212,13 +212,4 @@ def round(x):
     return ivy.round(x)
 
 
-round.unsupported_dtypes = (
-    "uint8",
-    "int8",
-    "uint16",
-    "int16",
-    "uint32",
-    "int32",
-    "uint64",
-    "int64",
-)
+round.unsupported_dtypes = {"torch": ("float16",)}
--- a/ivy/functional/frontends/jax/lax/operators.py
+++ b/ivy/functional/frontends/jax/lax/operators.py
@@ -206,3 +206,15 @@ def le(x, y):
 # round
 def round(x):
     return ivy.round(x)
+
+
+round.unsupported_dtypes = (
+    "uint8",
+    "int8",
+    "uint16",
+    "int16",
+    "uint32",
+    "int32",
+    "uint64",
+    "int64",
+)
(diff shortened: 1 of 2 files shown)


PatricYan:
@ivy-seed what's the meaning of the **next release**?

PatricYan:
Hi @mattbarrett98 , would you like to make a review for the PR?

mattbarrett98 (changes requested):
I believe `round` can take any numeric data type, but you are only testing against floats currently. Also, torch's `round` doesn't support `float16` which you should indicate with an `unsupported_dtypes` attribute similar to other functions in the operators file 🙂

PatricYan:
> I believe `round` can take any numeric data type, but you are only testing against floats currently. Also, torch's `round` doesn't support `float16` which you should indicate with an `unsupported_dtypes` attribute similar to other functions in the operators file 🙂

Hi @mattbarrett98 ,  I have tested the round by type int8, it is not supported.
```
TypeError: round does not accept dtype int8. Accepted dtypes are subtypes of floating.
```

mattbarrett98 (commented):
Which backend are you using to obtain this error? torch?

PatricYan:
> Which backend are you using to obtain this error? torch? Could you share the code 🙂

command 
```
Launching pytest with arguments --backend jax test_jax_lax_operators.py::test_jax_lax_round --no-header --no-summary -q in ~/GitHub/ivy/ivy_tests/test_ivy/test_frontends/test_jax

```

PatricYan:
> import jax.numpy as jnp
> >>> jnp.round(jnp.array(1, dtype="int8"))
> DeviceArray(1, dtype=int8)

after debugging by steps, it seems the error comes from 
```
frontend_ret = frontend_fw.__dict__[fn_tree](*args_frontend, **kwargs_frontend)
```

mattbarrett98 (changes requested):
Just leave it as floats only for now then. You still need to make the changes related to an unsupported_dtype attribute however

PatricYan:
> Just leave it as floats only for now then. You still need to make the changes related to an unsupported_dtype attribute however

Hi @mattbarrett98 , you mean to add the **int8, int32 ...** to the unsupported type?

mattbarrett98 (commented):
No, as I said before its torch which doesn't support `float16` which you should indicate in the attribute.

PatricYan:
> float16

done

PatricYan:
> No, as I said before its torch which doesn't support `float16` which you should indicate in the attribute.

Hi @mattbarrett98 , I have resolved it by adding the unsupported type **float16**

mattbarrett98 (approved):
Thank you 🙂

PR #3330, opened 2022-08-21
bitwise_not for jax lax operators

Ookamice:
- [ ]  #3318


Ookamice:
It looks like the frontend-jax test failed cus of my bitwise_not. Any advice would be greatly appreciated 🙏

Ookamice:
Just fixed an obvious mistake. Let me know if I need to fix anything else

kimbring2:
@Ookamice Hello works well. Thank you for your contribution.

PR #3341, opened 2022-08-21
reformat DefaultDevice.__exit__

mustafa1adel:


mustafa1adel:
Hi, I checked the failing tests, they are NumPy style docstrings, frontend-NumPy, and frontend-torch. However, the  __exit__ function at functional/ivy/device.py only, it does not exist at any of the backend frameworks. the docstrings check failed at torch, TensorFlow, Jax, and NumPy, which as I mentioned do not contain the __exit__ function.

PR #3343, opened 2022-08-21
Pixel shuffle frontends torch

AnnaTz:
link to issue: #3342


AnnaTz:
Hi @LINBOTANG123! Could you have a look?

AnnaTz:
@LINBOTANG123 is there a way to run the workflow on my own? I find it hard to spot the bugs without re-running multiple times...

AnnaTz:
@LINBOTANG123 all tests pass when I run them locally. Please let me know if there is something else I have overlooked.

AnnaTz:
Hi @rush2406. I know this is not your assignment but several days have passed without any feedback on my PR. Could you please have a look to close this?

ivy/functional/ivy/manipulation.py
```diff
@@ -471,7 +471,7 @@ def reshape(
     }
 
     """
-    return current_backend(x).reshape(x, shape, copy, out=out)
+    return current_backend(x).reshape(x, shape, copy=copy)
```
abdrahmandiab, line 474: Hi @AnnaTz, the out argument should be left here. The @handle_out_argument decorator at the start of the function handles this functionality, and it's good to have it in ivy even if the backends do not support it in the same way! :)

abdrahmandiab: you can leave copy=copy though :)

AnnaTz: Great, it's been resolved then :)

AnnaTz: Alright, I have merged the master into the PR, so the reshape function is now as you suggested :)

PR #3347, opened 2022-08-21
Implement reciprocal in ivy core and add to jax.lax frontend

### LINKED ISSUE #2873: Reciprocal function missing in Ivy Functional API
It appears as though an element-wise function, 'reciprocal', is missing from the Ivy Functional API.
I believe that it should have been implemented in the [elementwise.py](https://github.com/unifyai/ivy/blob/f2d2e225e1df4fe6ca2110dcc2ef5b0f5a14a20d/ivy/functional/ivy/elementwise.py) sub-module, however that does not seem to be the case. 
It has not been implemented in another sub-module either, to the best of my knowledge. 
Is this an intentional omission? If not, may I go ahead and create a pull request fixing the same?

TheMonocledHamster:
Resolves #2873 and #2869

Hi! 
I hope my implementation of the function meets the requirements. 
I look forward to iteratively improving the solution if necessary.
Cheers!

### DIFF
--- a/ivy/functional/backends/jax/elementwise.py
+++ b/ivy/functional/backends/jax/elementwise.py
@@ -339,6 +339,15 @@ def pow(
     return jnp.power(x1, x2)
 
 
+def reciprocal(
+    x: Union[float, JaxArray], 
+    /, 
+    *, 
+    out: Optional[JaxArray] = None
+) -> JaxArray:
+    return jnp.reciprocal(x)
+
+
 def remainder(
     x1: Union[float, JaxArray],
     x2: Union[float, JaxArray],
--- a/ivy/functional/backends/mxnet/elementwise.py
+++ b/ivy/functional/backends/mxnet/elementwise.py
@@ -279,6 +279,11 @@ def atan2(x: mx.nd.NDArray, y: mx.nd.NDArray) -> mx.nd.NDArray:
     return ret
 
 
+@_handle_flat_arrays_in_out
+def reciprocal(x: mx.nd.NDArray) -> mx.nd.NDArray:
+    return mx.nd.reciprocal(x)
+
+
 cosh = lambda x: math.cosh(x) if isinstance(x, float) else mx.nd.cosh(x)
 asinh = lambda x: math.asinh(x) if isinstance(x, float) else mx.nd.arcsinh(x)
 log = lambda x: math.log(x) if isinstance(x, float) else mx.nd.log(x)
--- a/ivy/functional/backends/numpy/elementwise.py
+++ b/ivy/functional/backends/numpy/elementwise.py
@@ -530,6 +530,16 @@ pow.support_native_out = True
 
 
 @_handle_0_dim_output
+def reciprocal(
+    x: Union[float, np.ndarray], /, *, out: Optional[np.ndarray] = None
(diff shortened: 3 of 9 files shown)


Daniel4078 (changes requested):
I only manage to find one thing to change currently, that is this formatting error:
ivy/functional/backends/tensorflow/elementwise.py:529:72: E502 the backslash is redundant between brackets
It is amazing that all other relative tests passes. Thank you.

TheMonocledHamster:
Thanks for the kind words!
I've made the requested changes, kindly re-review and let me know if there's any other improvement required.
Cheers!

Daniel4078 (commented):
Well, now there is only one assertion error happening in the test that found the ground truth return of array(True) is not the same with current output array(False). Which confuses me as reciprocal suppose to take in array of numbers and calcuate x^-1 on each of them and how does True and False get in to the test?
By the way, other test that reguard jax as front end uses this:
        available_dtypes=list(
            set(ivy_jax.valid_float_dtypes).intersection(set(ivy_np.valid_float_dtypes))
        )
    ) 
to generate values while the value you generate only have dtypes from ivy.valid_float_dtypes.

TheMonocledHamster:
> Well, now there is only one assertion error happening in the test that found the ground truth return of array(True) is not the same with current output array(False). Which confuses me as reciprocal suppose to take in array of numbers and calcuate x^-1 on each of them and how does True and False get in to the test? By the way, other test that reguard jax as front end uses this: dtype_and_x=helpers.dtype_and_values( available_dtypes=list( set(ivy_jax.valid_float_dtypes).intersection(set(ivy_np.valid_float_dtypes)) ) ) to generate values while the value you generate only have dtypes from ivy.

The failing test (due to the assertion error) is from the test_jax_lax_ge function, which had been merged into the repo before I made my PR. I'll try to look into it, however I haven't made any changes to the function in question, and I'm not sure if I ought make any within this PR? What do you think would be the best way forward for now?

As for the dtypes, ivy.valid_float_dtypes is a superset of ivy_jax.valid_float_dtypes and ivy_np.valid_float_dtypes (to the best of my knowledge). Of course, considering that the test is specifically for a jax frontend, it might have been a better choice to go with an intersection of ivy_jax and ivy_np.  I will make this change in the immediate next commit.

I feel that it would be appropriate to hold off on making any changes to the function that is throwing the assertion error, for now, as that has to do with the ge function and not the reciprocal function, so it would perhaps be out of place in this PR. Perhaps a separate PR to fix that would be more appropriate? What do you think?

Daniel4078:
Oh, then I must made a mistake and got the name of that test wrong. Well, then you only need to try fix the conflit in these files:
ivy/functional/frontends/jax/lax/operators.py
ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_operators.py
Then I will have a final look at the status of the tests and if they passes, I think the pull request is good to be merged. Thank you.

TheMonocledHamster:
Thanks for your patience and for the extraordinarily quick responses!
It has been a pleasure collaborating on this PR with you!!

Daniel4078 (approved):
I think there is no errors with your code and this pull request can be merged. Thank you.

PR #3349, opened 2022-08-22
torch allclose frontend implementation

hirwa-nshuti:
Hi @jkeane508 

Here is the PR for issue #3257 

I added the `torch.allclose` function and its tests.

Thanks


ivy_tests/test_ivy/test_frontends/test_torch/test_comparison_ops.py
```diff
@@ -0,0 +1,57 @@
...
+@handle_cmd_line_args
+@given(
+    dtype_and_input=helpers.dtype_and_values(
+        available_dtypes=tuple(
+            set(ivy_np.valid_float_dtypes).intersection(
+                set(ivy_torch.valid_float_dtypes)
+            ),
+        ),
+        num_arrays=2,
+        min_value=1.0013580322265625e-05,
```
jkeane508, line 22: We should be able to handle negative numbers aswell

ivy_tests/test_ivy/test_frontends/test_torch/test_comparison_ops.py
```diff
@@ -0,0 +1,57 @@
...
+        max_value=1,
```
jkeane508, line 23: This is a low value for the maximum value

ivy/functional/frontends/torch/comparison_ops.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+    )
+
+
+def allclose(input, other, rtol=1e-05, atol=1e-08, equal_nan=False):
+    finite_input = ivy.isfinite(input)
+    finite_other = ivy.isfinite(other)
+    if ivy.all(finite_input) and ivy.all(finite_other):
+        return _compute_allclose_with_tol(input, other, rtol, atol)
+    else:
```
jkeane508, line 19: It might not be necessary to handle the case with infinites, does np.all_close() handle inf and -inf?

ivy_tests/test_ivy/test_frontends/test_torch/test_comparison_ops.py
jkeane508, line 22: This means that values are never negative which is unrealistic compared to real world testing

ivy_tests/test_ivy/test_frontends/test_torch/test_comparison_ops.py
jkeane508, line 23: This value is too low, we need to test with large values

jkeane508 (changes requested):
Looks good, however I am not sure about whether or not the function is required to handle the inf and -inf dataypes.

ivy/functional/frontends/torch/comparison_ops.py
hirwa-nshuti: I think if we have inf or -inf values the comparison will depend on equal_nan mainly that's why if we have some inf values in our array we will either test them with `equal_nan=True` or `equal_nan=False`  and  the `np.allclose()` handles inf values

ivy_tests/test_ivy/test_frontends/test_torch/test_comparison_ops.py
hirwa-nshuti: I put 1e04 I hope it's large

ivy_tests/test_ivy/test_frontends/test_torch/test_comparison_ops.py
hirwa-nshuti: Changed the minimum to negative value

ivy/functional/frontends/torch/comparison_ops.py
jkeane508: Ok, if np handles it then we should indeed manage inf and -inf :)

ivy_tests/test_ivy/test_frontends/test_torch/test_comparison_ops.py
```diff
@@ -0,0 +1,59 @@
...
+@handle_cmd_line_args
+@given(
+    dtype_and_input=helpers.dtype_and_values(
+        available_dtypes=tuple(
+            set(ivy_np.valid_float_dtypes).intersection(
+                set(ivy_torch.valid_float_dtypes)
+            ),
+        ),
+        num_arrays=2,
+        min_value=-1.0013580322265625e-05,
```
jkeane508, line 22: We can actually just remove min_value and max_value paramters in this case as there is no need to constrain them at all :)

ivy/functional/frontends/torch/comparison_ops.py
```diff
@@ -0,0 +1,36 @@
+# global
+import ivy
+
+
```
jkeane508, line 4: If you could add the following line at top of document

`ToDo: Add this function to functional API`

Reason being if this is a function that exists on all frameworks, we should at some stage add it to all the functional api

jkeane508 (changes requested):
Just some small changes, if tests are passing, will merge :)

PR #3377, opened 2022-08-22
add scan function to jax frontend

Abdulhaffiz-Umar:
#3376


kimbring2:
@Abdulhaffiz-Umar Hello, it seems like you need to generate the value for the init argument for the Jax scan from the test function.

ivy-seed:

Issue #3298, opened 2022-08-20
is_nonzero


agniiyer:

agniiyer:
Pull request Added is_nonzero for torch frontend with tests passing #3386

PR #3386, opened 2022-08-22
Added is_nonzero for torch frontend with tests passing

agniiyer:


agniiyer:
Issue is_nonzero #3298

Auxia:
HI @agniiyer, 
I apologize for the late reply on this. It somehow slipped my radar, given the volume of other PR notifications.

I want to talk about two things here - 
1) You should only commit the changes to relevvant files and not the changes to files such as the IDE configurations or helper functions - unless you are making a PR on an issue with them.

2) The `is_nonzero` function has not been implemented in the Ivy backend. Infact, it is not implemented in any backend. The way you implement frontend functions is that you call, either the ivy backend function or a composition of ivy functions. You then return the result of those functions. Frontend functions are just paths to the relative backends and not actual implementations of the function themselves.

What I will suggest you do now is to start working on some other frontend task which has a backend function (You can search for implemented functions in the Ivy backend here - `/ivy/my_ivy/ivy/functional/ivy`). You can also start up an issue for functions that have not been implemented in Ivy yet, but are widely used in multiple backends. We will then flag that issue as `Future Release` and work on implementing those functions in future versions of Ivy.

agniiyer:
Thanks! I've submitted a PR for another issue #3800.

PR #3395, opened 2022-08-22
added average function

### LINKED ISSUE #2741: average
#1529

dgoel200113:
closes #2741


Daniel4078:
Hello and do not forget to add a hypothesis test for the function you created so that we can make sure its behavior is correct under all circumstances. Thank you.

PR #3433, opened 2022-08-22
Reformat cumsum

### LINKED ISSUE #1638: cumsum

JJKid:
Resolves #1638


CheezBallzPi (approved):

CheezBallzPi (changes requested):
Please update the backend APIs as well.

JJKid:
> Please update the backend APIs as well.

Done. Cumsum implementation on Tensorflow and MXNet had already a correct format for their arguments

CheezBallzPi:
The formatting looks good, but some of the backend functions have parameter issues. For example, JAX is missing the dtype argument and has an extraneous out argument. Read over [this](https://lets-unify.ai/ivy/deep_dive/10_function_arguments.html) for more info. It helps to look over the docs, [here](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.cumsum.html)'s cumsum in JAX.

CheezBallzPi (changes requested):
The formatting looks good, but some of the backend functions have parameter issues. For example, JAX is missing the dtype argument and has an extraneous out argument. Read over [this](https://lets-unify.ai/ivy/deep_dive/10_function_arguments.html) for more info. It helps to look over the docs, [here](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.cumsum.html)'s cumsum in JAX.

JJKid:
> The formatting looks good, but some of the backend functions have parameter issues. For example, JAX is missing the dtype argument and has an extraneous out argument. Read over [this](https://lets-unify.ai/ivy/deep_dive/10_function_arguments.html) for more info. It helps to look over the docs, [here](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.cumsum.html)'s cumsum in JAX.

Ok. I 've started working on it. 
One question. Should I add the device argument to ivy/functional cumsum function definition? 
I've added exclusive and reverse boolean arguments as well as are valid arguments for some backend implementations of cumsum. But I'm not sure if I should do the same for the device as no other similar functions on general.py have it. 
Thanks

CheezBallzPi:
Since the function has an input array, the device will be inferred from there, so you probably don't need the device parameter.

JJKid:
Hi. Just some comments on what I did. Correct me if I'm wrong :)

I left the unused argument 'out' for example in jax cumsum. If I remove it from the function definition, 
then when I call the ivy functional cumsum (after setting the backend to jax) with out in the arguments,
there is an error, as the 'out' argument doesnt exist on jax backend cumsum definition

For jax, mxnet and tensorflow I rely on the @handle_out_argument to do the inplace update with the return value of
the backend function, so I dont perform any operation involving the 'out' argument inside cumsum backend implementation for this backends

Thanks. I'll appreciate your feedback

CheezBallzPi (changes requested):
You shouldn't need to worry about dealing with `dtype`, the `@infer_dtype` decorator should take care of that for you. ([docs](https://lets-unify.ai/ivy/deep_dive/6_data_types.html)) You similarly do not need to worry about dealing with the `out` parameter, as the `@handle_out_argument` decorator should take care of that. ([docs](https://lets-unify.ai/ivy/deep_dive/8_inplace_updates.html)) Due to this, you will have to change the `dtype` parameter into a non-optional parameter and also remove the `support_native_out` attribute for JAX as it does not have an `out` argument natively.

CheezBallzPi (approved):

Issue #3471, opened 2022-08-23
Wrong placement of tanh function


Darshan2104:
NumPy-tanh is implemented in trigonometry which should be in the hyperbolic file.

PR #3473, opened 2022-08-23
change location of numpy-tanh

Darshan2104:
Wrong placement of NumPy-tanh function #3471


kimbring2:
@Darshan2104 Thank you for moving it (:

PR #3486, opened 2022-08-23
torch ones like test added

RickSanchezStoic:
torch ones like test added


jkeane508:
Hi,

First thing I have to ask is what is going on [here](https://github.com/unifyai/ivy/blob/e457ae30462e8b836b3f3b8c5b12b525129f481e/ivy/functional/frontends/torch/creation_ops.py#L26) with the function names? I haven't seen an implementation like this before :)

jkeane508 (approved):
Looking at this function, it takes pretty much the exact inputs as `torch.full`. So it's safe to assume that it should look pretty much the same as the test for full above with a few changes to function arguments.

ajay-vishnu:
@RickSanchezStoic, could you provide the link to the docs that explains about the implementation of the function names you've used?

PR #3489, opened 2022-08-23
Add argmin numpy frontend

Kaustav-Sarkar:
Added argmin to numpy frontend along with docstring.
References issue: 
- [ ] #3483


kimbring2:
@Kaustav-Sarkar It seems like you also need to write the test function for that.

Kaustav-Sarkar:
Ok sure I will add the tests

Kaustav-Sarkar:
All right I am making the function for argmax numpy

PR #3496, opened 2022-08-23
Backend version support

RickSanchezStoic:
Backend version support

### DIFF
--- a/ivy/backend_handler.py
+++ b/ivy/backend_handler.py
@@ -170,7 +170,7 @@ def set_backend_to_specific_version(backend):
             orig_name = fn_name_from_version_specific_fn_name(key, f_version)
             if orig_name:
                 backend.__dict__[orig_name] = backend.__dict__[key]
-                backend.__dict__[orig_name].__name__=orig_name
+                backend.__dict__[orig_name].__name__ = orig_name
 
 
 def current_backend(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


ivy/backend_handler.py
```diff
@@ -98,6 +98,80 @@ def _determine_backend_from_args(args):
...
+    -------
+
+    """
+    f = str(backend.__name__)
+    f = f[f.index("backends") + 9 :]
+
+    f = importlib.import_module(f)
+    f_version = f.__version__
+
+    for i in list(backend.__dict__):
```
djl11, line 168: replace `i` with `key`, `i` makes it sound like a numeric counter.

ivy/backend_handler.py
```diff
@@ -98,6 +98,80 @@ def _determine_backend_from_args(args):
...
+        The function doesn't return anything and updates the backend __dict__
+        to make the original function name to point to the version specific one
+
+    -------
+
+    """
+    f = str(backend.__name__)
+    f = f[f.index("backends") + 9 :]
+
+    f = importlib.import_module(f)
```
djl11, line 165: is this not exactly what `backend` already is? why do we need to import it again?

djl11 (changes requested):
small changes and questions, otherwise looks good!

ivy/backend_handler.py
RickSanchezStoic: Actually, the variable backend here would be a module say `ivy.functional.backends.torch`. We want the `torch` part so that we can import the original `torch` from the user's system and get its version.

ivy/backend_handler.py
RickSanchezStoic: yep, will fix this.

djl11 (approved):
lgtm!

PR #3497, opened 2022-08-23
added numpy reshape and its test function

ojask12:
- [ ] #3036


ivy_tests/test_ivy/test_frontends/test_numpy/test_manipulation_routines/test_changing_array_shape.py
```diff
@@ -0,0 +1,65 @@
...
+
+    shape = []
+    shape_len = draw(st.integers(min_num_dims, max_num_dims))
+    for _ in range(shape_len):
+        s_ = draw(st.integers(min_dim_size, max_dim_size))
+        shape.append(s_)
+
+    dtypes, x = draw( helpers.dtype_and_values(shape=tuple(shape)) )
+    reshape = draw(st.permutations(shape))
+    return dtypes, x, tuple(reshape)
```
Aarsh2001, line 31: Hi, i don't think these functions are required. You can do the same thing inside the `given` decorator, with the helper functions already provided. Kindly go through these, for example `dtype_and_values`, `array_dtypes` etc.

Aarsh2001 (changes requested):

ivy_tests/test_ivy/test_frontends/test_numpy/test_manipulation_routines/test_changing_array_shape.py
ojask12: I agree with copy and with out but the second function is used to generate a valid reshape for example: 
a numpy array of shape 3,2 cannot be reshaped to 7,1 so the function _x_dtypes_n_shape would draw an arbitrary shape create an array with the same shape & then shuffles the shape to generate a valid reshape. Alternatively I can randomly draw integers and keep dividing with the product. Is there a default function for that as well.

Aarsh2001: Hi @ojask12, apologies for the late reply. There's a function called `reshape_shapes` in the helper file, please have a look at it, thanks🙂

Aarsh2001 (changes requested):

Aarsh2001:
Hi @ojask12, can you please install `pre-commit` and sort out the lint errors? Kindly check the setting up sections of the docs for further steps. Otherwise the PR looks fine, and we'll get it merged in no time, thanks.

ojask12:
image

Aarsh2001:
Hi @ojask12, are the tests passing for this function on your local machine?

Aarsh2001 (changes requested):
failing tests need to be fixed

Aarsh2001:
lgtm!

ojask12:
Thanks!

PR #3514, opened 2022-08-24
numpy frontend - `logic`: `logical_operations` and  `truth_value_testing` submodule

juliagsy:
1. logical_operations
- logical_and
- logical_or
- logical_not
- logical_xor

2. truth_value_testing
- all
- any

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_numpy/test_logic/test_logical_operations.py
+++ b/ivy_tests/test_ivy/test_frontends/test_numpy/test_logic/test_logical_operations.py
@@ -1,3 +1,4 @@
+# for review
 # global
 import numpy as np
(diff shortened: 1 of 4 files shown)


ivy_tests/test_ivy/test_frontends/test_numpy/test_logic/test_logical_operations.py
```diff
@@ -1,3 +1,4 @@
+# for review
```
djl11, line 1: small detail, but `(ivy_np.bool,) + (None,)` should be replaced with `(ivy_np.bool, None)`

djl11 (changes requested):
very small change, after that it's good to merge! Thanks!

PR #3543, opened 2022-08-24
Add Tensorflow Frontend categorical_crossentropy 

wammy19:
Hi,

Here's an attempt at Tensorflow categorical_crossentropy. I'm a little skeptical of the test at the moment. 

It's currently passing because I'm explicitly setting the `dtype` of `y_pred` to `ivy.float64` on line 40. I don't image this is desirable since hypothesis is generating a random `dtype` for us. 

But when I change line 40 to:

`y_pred = ivy.array(y_pred, dtype=dtype)`

The test is failing with this assertion error:

`AssertionError: the return with a TensorFlow backend produced data type of float16, while the return with a numpy backend returned a data type of float64.`

I'm not totally sure how to go about solving this, unless explicitly setting the `dtype` to match is solving it. 

I also noticed the `sparse_categorical_crossentropy` test is failing, so I made some changes to get that passing also. Let me know if this should have been done in another issue, and/or if the changes are valid.

Any advice is much appreciated ✌️.

Edit:

I can't seem to link the issue this PR is related to on the right hand side. Issue number: #3285

### DIFF
--- a/ivy/functional/frontends/tensorflow/raw_ops.py
+++ b/ivy/functional/frontends/tensorflow/raw_ops.py
@@ -2,11 +2,11 @@
 import ivy
 
 
-def acos(x, name="acos"):
+def Acos(*, x, name="Acos"):
     return ivy.acos(x)
 
 
-def acosh(x, name="acosh"):
+def Acosh(*, x, name="Acosh"):
     return ivy.acosh(x)
 
 
@@ -14,11 +14,11 @@ def concat(values, axis, name="concat"):
     return ivy.concat(values, axis=axis)
 
 
-def cos(x, name="cos"):
+def Cos(*, x, name="Cos"):
     return ivy.cos(x)
 
 
-def cosh(x, name="cosh"):
+def Cosh(*, x, name="cosh"):
     return ivy.cosh(x)
 
 
@@ -26,11 +26,11 @@ def fill(dims, value, name="full"):
     return ivy.full(dims, value)
 
 
-def asin(x, name="asin"):
+def Asin(*, x, name="asin"):
     return ivy.asin(x)
 
 
-def atan(x, name="atan"):
+def Atan(*, x, name="atan"):
     return ivy.atan(x)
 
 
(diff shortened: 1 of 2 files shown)


ivy_tests/test_ivy/test_frontends/test_tensorflow/test_metrics.py
```diff
@@ -8,6 +8,59 @@
...
+    fw
+):
+    dtype, y_pred = dtype_y_pred
+    y_pred = ivy.array(y_pred, dtype=ivy.float64)
+    y_true = ivy.array(y_true)
+    ivy.one_hot(y_true, y_pred.shape[-1], out=y_true)
+
+    # Perform softmax on prediction if it's not a probability distribution.
+    if not from_logits:
+        y_pred = ivy.exp(y_pred) / ivy.sum(ivy.exp(y_pred))
```
sherry30, line 46: `exp` function doesn't work for `float16` with torch backend and this is giving an error, you will have to fix this.
1 possible fix i can think of is using `assume` in the test function. Its not the most elegant solution.
You can also create a composite function for this since this same logic is being repeated in sparse_categorical_crossentropy
 if you can think of something else feel free to implement it.

sherry30 (changes requested):
Hi,

a few things which need to be addressed:

1. To fix the dtype problem, you can just type cast with `ypred.dtype` before returning in the frontend function. that should fix the different dtypes being returned in all the backends.
2. You should also test for `axis` parameter.
3. After solving the dtype problem, It seems like there is an assertion error, its producing slightly different results from tensorflow so you will have to look into that.
4. It would be better to create a new PR for the failing `sparse_categorical_crossentropy` function. It would make things cleaner.
5. One last thing which i mentioned on the specific line.

Let me know if you have any other question

Thanks.

wammy19:
I re-synced my fork and that seemed to have closed this PR (I wasn't aware that would happen). I forgot to create a branch which made things messy, I apologies for this. Should I create a new PR for this and continue with the changes you have requested?

PR #3544, opened 2022-08-24
Added barebones hypothesis tests for all optimizers

Auxia:
This should be a good starting point to implement robust hypothesis tests for this submodule


jkeane508:
Nice, I was looking at tests today and was struggling to make sense :)

Auxia:
Alright I can merge this for now and we can start working yeah.

PR #3552, opened 2022-08-24
Added custom linear solve function to JAX frontend

ukj02:


CheezBallzPi (changes requested):
The code is not passing your unit test.

ukj02:
Hi, I am kind of confused as what the arguments should be for the test function. The function I am implementing has a lot of arguments and I think that is where it is failing. 
These are the arguments from test https://github.com/ukj02/ivy/blob/master/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_custom_gradient_operators.py
`
matvec,
b,
solve,
transpose_solve=None,
symmetric=False,
has_aux=False
`
Can you help me figure out what the correct arguments should be?

bicycleman15:
Hey @ukj02 , I feel this function is more complicated to implement in ivy than I thought. Could you pick some other task in the mean time ? Really sorry for the late response.

Issue #3562, opened 2022-08-25
sum


PatricYan:

Daniel4078:
Why did you close this as completed?

PatricYan:
ok, I want to open a new issue for the PR

PatricYan:
Hi @Daniel4078  I think you can reopen it.

Daniel4078:
Well, I thought you want to leave this and try another one. You should repeat the progress of commenting on the TODO list issues about your new issue as I removed this closed issue from that list. Thank you.

PR #3570, opened 2022-08-25
add sum to numpy

PatricYan:
- [ ] #3562


ivy/functional/frontends/numpy/statistics/correlating.py
```diff
@@ -0,0 +1,8 @@
+import ivy
+
+
+def sum(x, /, *, axis=None, dtype=None, keepdims=False, out=None):
```
Daniel4078, line 4: According to the document of numpy.sum, here you need two extra optional keyward argument called initial and where. Also, the initial value of keepdims should be None.
Here is the original arguments of numpy.sum: numpy.sum(a, axis=None, dtype=None, out=None, keepdims=<no value>, initial=<no value>, where=<no value>)

ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics/test_correlating.py
```diff
@@ -0,0 +1,57 @@
...
+    num_positional_args,
+    native_array,
+    fw,
+):
+    input_dtype, x, axis = dtype_x_axis
+
+    helpers.test_frontend_function(
+        input_dtypes=input_dtype,
+        as_variable_flags=as_variable,
+        with_out=False,
```
Daniel4078, line 46: The with_out argument should be passed in through command line arguments of the test function, since that is being handled by the @handle_cmd_line_args decorator.

ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics/test_correlating.py
```diff
@@ -0,0 +1,57 @@
...
+        num_positional_args=num_positional_args,
+        native_array_flags=native_array,
+        fw=fw,
+        frontend="numpy",
+        fn_tree="sum",
+        x=np.asarray(x, dtype=input_dtype),
+        axis=axis,
+        dtype=None,
```
Daniel4078, line 54: I think you should generate some values for dtype, where and initial as well in the given part so that all the features can be tested.

Daniel4078 (changes requested):
I found some errors that you need to fix before the pull request can be approved. Check the comments below for details. Thank you.

ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics/test_correlating.py
PatricYan: done

ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics/test_correlating.py
PatricYan: done

ivy/functional/frontends/numpy/statistics/correlating.py
Daniel4078: Well, the frontend function we create should have the exact same behaviour as the origenal function in the targeted framework.

Daniel4078:
Please update your implementation to have the same input arguments as numpy.sum as we want the frontend function to behave the exact way it does in its target framework

PatricYan:
> Please update your implementation to have the same input arguments as numpy.sum as we want the frontend function to behave the exact way it does in its target framework
Hi in **ivy/functional/backends/numpy/statistical.py** the sum function definition is below.
How can I implement what you have said? the definition is not same as [numpy sum doc](https://numpy.org/doc/stable/reference/generated/numpy.sum.html), it losses the ** initial** and **where**
would you like to take a review for the sum fucntion in numpy backend **ivy/functional/backends/numpy/statistical.py** 
```
def sum(
    x: np.ndarray,
    /,
    *,
    axis: Union[int, Tuple[int]] = None,
    dtype: np.dtype = None,
    keepdims: bool = False,
) -> np.ndarray:
    if dtype is None and np.issubdtype(x.dtype, np.integer):
        if np.issubdtype(x.dtype, np.signedinteger) and x.dtype in [
            np.int8,
            np.int16,
            np.int32,
        ]:
            dtype = np.int32
        elif np.issubdtype(x.dtype, np.unsignedinteger) and x.dtype in [
            np.uint8,
            np.uint16,
            np.uint32,
        ]:
            dtype = np.uint32
        elif x.dtype == np.int64:
            dtype = np.int64
        else:
            dtype = np.uint64
    dtype = ivy.as_native_dtype(dtype)
    return np.asarray(np.sum(a=x, axis=axis, dtype=dtype, keepdims=keepdims, out=out))

```

ivy/functional/frontends/numpy/statistics/correlating.py
PatricYan: the backend sum function losses **initial** and **where**

PatricYan:
if need to add the **initial** and **where**, I think the backend numpy sum function needs to be updated to support the two more arguments.

Daniel4078:
> 

I am asking other members of the ivy team about this. By the way, I think you can check how the numpy framework implemented the function and try to replicate their approach. You may also try update the backend function to accept those optional arguments.

PatricYan:
> > 
> 
> I am asking other members of the ivy team about this. By the way, I think you can check how the numpy framework implemented the function and try to replicate their approach. You may also try update the backend function to accept those optional arguments.

Ok. However, there is just one update for every PR. I think after this PR, I can make a new PR to optimize the backend function sum of Numpy.

Daniel4078:
> > > 
> > 
> > 
> > I am asking other members of the ivy team about this. By the way, I think you can check how the numpy framework implemented the function and try to replicate their approach. You may also try update the backend function to accept those optional arguments.
> 
> Ok. However, there is just one update for every PR. I think after this PR, I can make a new PR to optimize the backend function sum of Numpy.

Well, I have see pull request that deals with multiple issues get merged here. So as long as no new errors are brount into the code base, you do not need to worry about that.

PatricYan:
> > > > 
> > > 
> > > 
> > > I am asking other members of the ivy team about this. By the way, I think you can check how the numpy framework implemented the function and try to replicate their approach. You may also try update the backend function to accept those optional arguments.
> > 
> > 
> > Ok. However, there is just one update for every PR. I think after this PR, I can make a new PR to optimize the backend function sum of Numpy.
> 
> Well, I have see pull request that deals with multiple issues get merged here. So as long as no new errors are brount into the code base, you do not need to worry about that.

Hi any question about the PR? would you like to make a review for it again?

Daniel4078 (changes requested):
Firstly, you need to implement the two optional keyword argument initial and where.
Secondly, you need to update your test to include these two new arguments.
Thank you.

Daniel4078:
?

PatricYan:
Hi @Daniel4078 , I have resolved it by adding **where** and **initial**.

ivy/functional/frontends/numpy/statistics/correlating.py
```diff
@@ -0,0 +1,27 @@
...
+
+def sum(
+    x,
+    /,
+    *,
+    axis=None,
+    dtype=None,
+    keepdims=False,
+    out=None,
+    initial=np._NoValue,
```
Daniel4078, line 13: You should use None here instead of np._NoValue as these frontend functions should not have dependencies to those other frameworks.

Daniel4078 (changes requested):
Well done, it seems like there is currently your test is passing. There is just some small changes reguarding replaceing np._NoValue to None in your implementation and test to remove unwanted dependency to numpy framwork as I suggests in the comment below. Thank you.

ivy/functional/frontends/numpy/statistics/correlating.py
PatricYan: Hi @Daniel4078 , **None** is not same as **np._NoValue**, I have tested in my jupyter notebook

```
import numpy as np


a = np.arange(8).reshape((2, 2, 2))
print(a.shape)
print("a:\n", a)

b = np.sum(a, keepdims=False, axis=0)
print("b:\n", b)

b = np.sum(a, keepdims=False, axis=0, initial=np._NoValue, where=np._NoValue)
print("b:\n", b)
```

- bad case
```
import numpy as np


a = np.arange(8).reshape((2, 2, 2))
print(a.shape)
print("a:\n", a)

b = np.sum(a, keepdims=False, axis=0)
print("b:\n", b)

b = np.sum(a, keepdims=False, axis=0, initial=None, where=None)
print("b:\n", b)

```
with errors
```
ValueError                                Traceback (most recent call last)
Input In [3], in <cell line: 11>()
      8 b = np.sum(a, keepdims=False, axis=0)
      9 print("b:\n", b)
---> 11 b = np.sum(a, keepdims=False, axis=0, initial=None, where=None)
     12 print("b:\n", b)

File <__array_function__ internals>:180, in sum(*args, **kwargs)

File ~/.conda/envs/ivy/lib/python3.9/site-packages/numpy/core/fromnumeric.py:2298, in sum(a, axis, dtype, out, keepdims, initial, where)
   2295         return out
   2296     return res
-> 2298 return _wrapreduction(a, np.add, 'sum', axis, dtype, out, keepdims=keepdims,
   2299                       initial=initial, where=where)

File ~/.conda/envs/ivy/lib/python3.9/site-packages/numpy/core/fromnumeric.py:86, in _wrapreduction(obj, ufunc, method, axis, dtype, out, **kwargs)
     83         else:
     84             return reduction(axis=axis, out=out, **passkwargs)
---> 86 return ufunc.reduce(obj, axis, dtype, out, **passkwargs)

ValueError: reduction operation 'add' does not have an identity, so to use a where mask one has to specify 'initial'
```

Daniel4078: The error is caused by a mistake in the data generation, your data generation allow initial to be None while where is not None. So by adding a check to set where to none when initial is generated to be none will solve the problem. Thank you.

PatricYan:
```
import numpy as np


a = np.arange(8).reshape((2, 2, 2))
b = np.sum(a, keepdims=False, axis=0)
b = np.sum(a, keepdims=False, axis=0, initial=np._NoValue, where=np._NoValue)
b = np.sum(a, keepdims=False, axis=0, initial=None, where=None)

```

I think you maybe need to have a look at this code and test it.

Daniel4078: Well then, May be you can update the back end implementation the function in ivy code base to allow it to convert None to np._NoValue. Because we are not supposed to include any dependency of numpy in the front end implementations as this should be able to run on any backend framwork settings. Thank you.

PatricYan: Done, convert **None**  to **np._NoValue**. However, in the test, helpers.py, it will call the ground function to make the value test, so in the test case, we need to use **np._NoValue**, or, it will cause error for the **None** when there is not **initial**, but exist **where**

ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics/test_correlating.py
```diff
@@ -0,0 +1,74 @@
...
+    keep_dims,
+    initial,
+    as_variable,
+    num_positional_args,
+    native_array,
+    with_out,
+    fw,
+):
+    (input_dtype, x, axis), where = dtype_x_axis
+
```
Daniel4078, line 57: I think that you need to check if initial is None, then set where to be None as well, because where can not have value when initial does not. Thank you.

PatricYan: I think when **initial** is **None**, **where** can not be **None**, you can take a look at below example 
```
import numpy as np


a = np.arange(8).reshape((2, 2, 2))
where = a>3
print(where.shape)
print(where)
b=np.sum(a, keepdims=False, axis=0, where=where)
print("aaa:", b)
```

Daniel4078:
Understand, I will ask other members of my team about how to deal with np._NoValue, now, please use _NoValue as you like as it is part of the behavior of the original function anyway. Thank you.

ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics/test_correlating.py
Daniel4078: Thank you for pointing that out.

Daniel4078:
Please get to know about this pull request: (https://github.com/unifyai/ivy/pull/3698)
As it will resolve the numpy._NoValue issues for you by giving you a substitute to generate in the test and only use None in the frontend implementation. Thank you for your patience.

PatricYan:
> Please get to know about this pull request: (#3698) As it will resolve the numpy._NoValue issues for you by giving you a substitute to generate in the test and allow you to use None instead in the frontend implementation. Thank you for your patience.

Hi, I just used **None** in the fronted and **_NoValue** in the test. And after your PR mergerd, I can update the test case.  would you like to make a review and merge it?

Daniel4078 (approved):
Currently, your test passes and I think it is good to be merged for now. But do not forget to update the implementation and test once the NativeClass update got merged. Thank you!

PatricYan:
> Currently, your test passes and I think it is good to be merged for now. But do not forget to update the implementation and test once the NativeClass update got merged. Thank you!

ok, When code is updated, I will update the case.

Daniel4078:
I just double checked, numpy.sum do accept None instead of np._NoValue, but the true issue is that the where argument you have  should default to True instead of None. There is an error in the numpy document of numpy.sum as it states the default value of where is novalue, but the actural ufunc.reduce it calls under the hood set where to True in default.

Daniel4078:
I will fix these problems in my own pull request, do not worry.

PatricYan:
> should default to True instead of None

ok, do you test this case? I have marked the right and error cases.
```
import numpy as np


a = np.arange(8).reshape((2, 2, 2))
where = a>3
print(where.shape)
print(where)

b=np.sum(a, keepdims=False, axis=0, where=where)
print("aaa:", b)

# right
b=np.sum(a, keepdims=False, axis=0, initial=np._NoValue,where=where)
print("aaa:", b)

# error
b=np.sum(a, keepdims=False, axis=0, initial=None, where=where)
print("aaa:", b)
```

Daniel4078:
The thing is, np._NoValue in this case, act the same as 0, Not using the argument for our current implementation will produce the same result.
np.sum(a, keepdims=False, axis=0, initial=0, where=where)=array([[4, 5], [6, 7]])

Daniel4078:
the numpy just requires us not to set initial to None when where is not True.

PatricYan:
> the numpy just requires us not to set initial to None when where is not True. Just like in this case: np.sum(a, keepdims=False, axis=0, initial=None, where=True) = array([[ 4, 6], [ 8, 10]])

ok, maybe numpy doc needs to be updated to declare the **initial** can not be **None** when where is **None**

Daniel4078:
In addition, check out the current implementation

@from_zero_dim_arrays_to_float
def sum(
    x,
    /,
    *,
    axis=None,
    dtype=None,
    keepdims=False,
    out=None,
    initial=None,
    where=True,
):
    if not where:
        return 0
    if initial:
        s = ivy.shape(x, as_array=True)
        s[axis] = 1
        header = ivy.full(ivy.Shape(s), initial)
        if where:
            x = ivy.where(where, x, ivy.default(out, ivy.zeros_like(x)))
        x = ivy.concat([x, header], axis=axis)
    return ivy.sum(x, axis=axis, dtype=dtype, keepdims=keepdims, out=out)

Daniel4078:
It solves all the issues while recreate the numpy behavior well.

PR #3578, opened 2022-08-25
numpy frontend `exponents_and_logarithms`

rush2406:
A common error appearing for many of the numpy frontend functions seems to be - 
`AssertionError: len(ret_np_flat) != len(ret_np_from_gt_flat):
  ret_np_flat:     [array(-0.597, dtype=float16)],
  ret_np_from_gt_flat: []`

I wasn't able to exactly figure out if this error is due to the backend implementations or the test_frontend_helper


VedPatwardhan (changes requested):
Hi @rush2406, thanks for requesting the review! There are a few changes I'd suggest in order to get the tests to pass. Firstly, please add the `unsupported_dtypes` dictionary to all functions added to avoid cases like `torch` with `float16`. The second issue seems to be related to  arrays with zero dimensions, because their result is a `float` instead of an `ivy.Array`. I'd suggest you to look into the `from_zero_dim_arrays_to_float` decorator regarding the same.

VedPatwardhan (changes requested):
Hi @rush2406, one final change, please update the `small_value_safety_factor` for `test_numpy_log2` to a larger value like `1.2` or something, because it seems to be failing for very small values when I tested it locally. Thanks 🙂

PR #3613, opened 2022-08-25
implemented jax lax cos in the jax frontend and tested.

vaatsalya123:


vaatsalya123:
These error don't really have anything to do with my code.I think you should not have any problem in merging this please do let me know your  review of the code thanks.

PR #3633, opened 2022-08-26
add ivy.functional.fronteds.torch.loss_functions.smooth_l1_loss and ivy.smooth_l1_loss

WOWspring:


Ishticode (changes requested):
Hi @WOWspring Thanks for the PR 🙂
Few requests. Could you exclude `.idea` files and add test for the frontend. Currently we are not adding any new functions to Ivy functional. If your function doesn't have a corresponding backend function then we either try compostion or leave this function for the next release and select another funtion to work on which could be written using the currently supported funtions. Besides, just keep in mind that when we do add a function in Ivy functional it must be accompanied by its corresponding test.
Please check out the [guide](https://lets-unify.ai/ivy/deep_dive.html) and feel free to ask any questions. Thank you.

Issue #1525, opened 2022-06-19
Add Mathematical functions to Numpy Frontend 


Aarsh2001:
Add [Mathematical functions](https://numpy.org/doc/stable/reference/routines.math.html#) to Numpy frontend:

- [ ] sin
- [ ] cos
- [ ] tan
- [ ] arcsin
- [ ] arccos
- [ ] arctan
- [ ] hypot
- [ ] arctan2
- [ ] degrees
- [ ] radians
- [ ] unwrap
- [ ] deg2rad
- [ ] rad2deg
- [ ] sinh
- [ ] cosh
- [ ] tanh
- [ ] arcsinh
- [ ] arccosh
- [ ] arctanh
- [ ] around
- [ ] round_
- [ ] rint
- [ ] fix
- [ ] floor
- [ ] ceil
- [ ] trunc
- [ ] prod
- [ ] sum
- [ ] nanprod
- [ ] nansum
- [ ] cumprod
- [ ] cumsum
- [ ] nancumprod
- [ ] nancumsum
- [ ] diff
- [ ] ediff1d
- [ ] gradient
- [ ] cross
- [ ] trapz
- [ ] exp
- [ ] expm1
- [ ] exp2
- [ ] log
- [ ] log10
- [ ] logaddexp
- [ ] logaddexp2
- [ ] i0
- [ ] sinc
- [ ] signbit
- [ ] copysign
- [ ] frexp
- [ ] Idexp
- [ ] nextafter
- [ ] spacing
- [ ] lcm
- [ ] gcd
- [ ] add
- [ ] reciprocal
- [ ] positive
- [ ] negative
- [ ] multiply
- [ ] divide
- [ ] power
- [ ] subtract
- [ ] true_divide
- [ ] floor_divide
- [ ] float_power
- [ ] fmod
- [ ] mod
- [ ] modf
- [ ] remainder
- [ ] divmod
- [ ] angle
- [ ] real
- [ ] imag
- [ ] conj
- [ ] conjugate
- [ ] maximum
- [ ] fmax
- [ ] amax
- [ ] nanmax
- [ ] minimum
- [ ] fmin
- [ ] amin
- [ ] nanmin
- [ ] convolve
- [ ] clip
- [ ] sqrt
- [ ] cbrt
- [ ] square
- [ ] absolute
- [ ] fabs
- [ ] sign
- [ ] heaviside
- [ ] nan_to_num
- [ ] real_if_close
- [ ] interp

[...]

PR #3634, opened 2022-08-26
Implement Numpy's sum function. One of the tasks present in #1525

AmeenUrRehman:


AmeenUrRehman:
could anyone guide me why this happen?

PR #3635, opened 2022-08-26
Add torch sort frontend 

Nightcrab:
- [ ] #3632


ivy/functional/frontends/torch/comparison_ops.py
```diff
@@ -49,3 +49,11 @@ def eq(input, other, *, out=None):
...
 def argsort(input, dim=-1, descending=False):
     return ivy.argsort(input, axis=dim, descending=descending)
+
+
+def sort(input, dim=-1, descending=False, stable=False, out=None):
+    values = ivy.sort(input, axis=dim, descending=descending, stable=stable, out=out)
+
+    indices = ivy.argsort(input, axis=dim, descending=descending)
+
+    return [values, indices]  # should be a namedtuple?
```
sherry30, line 59: According to the torch documentation

> A namedtuple of (values, indices) is returned, where the values are the sorted values and indices are the indices of the elements in the original input tensor.

So i guess it would be better to return a named tuple.

sherry30 (changes requested):
Hi,

pointed out the change on the line

Thanks

ivy/functional/frontends/torch/comparison_ops.py
Nightcrab: Agreed. Looking at https://pytorch.org/docs/stable/jit_language_reference.html, it looks like torch uses `collections.namedtuple`, so I've updated sort frontend to match.

sherry30:
Hi,
The changes look good.
Can you resolve the conflicts on this PR, I'll merge it as soon as you do 👍 

Thanks

Nightcrab:
Resolved conflicts.

PR #3637, opened 2022-08-26
Added `torch.ge` and `torch.greater_equal` to the frontend 

hirwa-nshuti:
Hi

This PR resolves issues #3620 and #3621 
These two functions `torch.ge`  and  `torch.greater_equal` are aliases.

Thanks

### DIFF
--- a/ivy/functional/frontends/torch/comparison_ops.py
+++ b/ivy/functional/frontends/torch/comparison_ops.py
@@ -54,3 +54,6 @@ def argsort(input, dim=-1, descending=False):
 def ge(input, other, *, out=None):
     ret = ivy.greater_equal(input, other, out=out)
     return ret
+
+
+greater_equal = ge
--- a/ivy/functional/frontends/torch/comparison_ops.py
+++ b/ivy/functional/frontends/torch/comparison_ops.py
@@ -51,9 +51,9 @@ def argsort(input, dim=-1, descending=False):
     return ivy.argsort(input, axis=dim, descending=descending)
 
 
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/torch/comparison_ops.py
```diff
@@ -49,3 +49,11 @@ def eq(input, other, *, out=None):
 
 def argsort(input, dim=-1, descending=False):
     return ivy.argsort(input, axis=dim, descending=descending)
+
+
+def ge(input, other, *, out=None):
```
jkeane508, line 54: This is me being really picky but could you set greater_equal to be the main function and ge to be the alias. I think its more readable :)

ivy_tests/test_ivy/test_frontends/test_torch/test_comparison_ops.py
```diff
@@ -175,3 +175,44 @@ def test_torch_argsort(
         dim=axis,
         descending=descending,
     )
+
+
+# ge
```
jkeane508, line 180: Same here

jkeane508 (changes requested):
Just me being super picky, but small change and should be good to go :)

ivy/functional/frontends/torch/comparison_ops.py
hirwa-nshuti: Yeah makes sense Let me do it I only implemented this way following torch implementation

jkeane508 (approved):
Once passing will merge :)

PR #3657, opened 2022-08-26
fix `array_values` `min_value` and `max_value` 

CatB1t:


djl11 (approved):
lgtm!

CatB1t:
`Tensorflow` uses `numpy`'s `finfo` which doesn't have any info about `bfloat16` and it fails, according to Google, `The dynamic range of bfloat16 and float32 are equivalent`, so I've decided to return the `Finfo` for `float32` instead. 

Not sure if this is proper way to do it, as I've tested using `torch` to check the range of `float32` and `bfloat16` and they're slightly different.

djl11 (approved):
Good question about the tensorflow `finfo` issue, yeah I think your current solution is the most elegant for now. Thanks!

CatB1t:
```
finfo(resolution=0.01, min=-3.38953e+38, max=3.38953e+38, eps=0.0078125, tiny=1.17549e-38, dtype=bfloat16)
finfo(resolution=1e-06, min=-3.40282e+38, max=3.40282e+38, eps=1.19209e-07, tiny=1.17549e-38, dtype=float32)
```
this difference is almost unnoticeable, I'm not sure if it's an issue with representation of `bfloat16` or the the difference is insignificant that they ignored it.

PR #3669, opened 2022-08-26
Optimizers submodule fixup

Auxia:
-> SGD and LARS unit tests have been completed.
-> Minor changes to ADAM and LAMB as well:
     -> `with_v` was removed as it was not required for these functions
     -> `compile_next_step` removed as it goes unused as graph compiler is not being used currently


jkeane508:
Hey @Auxia,

So what needs to be done for this submodule?

Auxia:
First of all we need to add all the parameters for the ADAM and LAMB optimizers and we need to fix the tests. One peculiar thing happening on my end is that I do not encounter the errors locally, these errors are something I only encounter on the CI. Maybe its an issue with the fact that I am using VSCode but I will try to figure it out.

So yeah first order of business is to get these tests passing.

djl11 (changes requested):
is there a reason for there being only one call of the new `helpers.get_dtypes` approach and the rest use the old `ivy_tf.valid_float_dtypes` approach?

Also, could you please add `# For Review` to the top of all relevant files, not just the test file?

Thanks!

Auxia:
The reason why I used `ivy_tf.valid_float_dtypes` for ADAM and LAMB is because during testing, tensorflow returns outputs in `float16` while NumPy defaults `float32`. This causes dtype errors even though the value test passes. I didn't know how to skip float 16 with the `get_dtypes` approach so I used the old method instead.

That said the tests are all passing now. Please let me know if any other changes are required.

PR #3698, opened 2022-08-27
Individual task: Framework specific classes in the frontend

Daniel4078:
I have done to step c. Please have a check on my ideas to see if it make sense. Thank you!


Daniel4078:
May I currently ignore the F401 ('xxx' imported but unused) errors? They will be used in the future but now I do know know good examples to fit in to the dictionaries.

ivy_tests/test_ivy/helpers.py
```diff
@@ -1804,7 +1848,32 @@ def test_frontend_function(
 
     # frontend function
     frontend_fn = ivy.functional.frontends.__dict__[frontend].__dict__[fn_tree]
-
+    
+    # check and replace NativeClass object in arguments with ivy counterparts
+    if frontend == "jax":
```
djl11, line 1853: this can all be simplified, something like:

```
convs = {"jax": convjax, "numpy": convjax, "tensorflow": convjax, "torch": convjax}
if frontend in convs:
    conv =  convs[frontend]
    args = ivy.nested_map(args, fn=conv, include_derived=True)
    kwargs = ivy.nested_map(kwargs, fn=conv, include_derived=True)
```

ivy_tests/test_ivy/helpers.py
```diff
@@ -53,6 +62,41 @@
 import ivy.functional.backends.numpy as ivy_np
 
 
+def convtrue(argument):
```
djl11, line 65: this function should stay here...

ivy_tests/test_ivy/helpers.py
```diff
@@ -53,6 +62,41 @@
...
+    """Convert NativeClass in argument to true framework counter part"""
+    if isinstance(argument, NativeClass):
+        return argument._native_class
+    return argument
+
+
+def convjax(argument):
```
djl11, line 72: ...but these and all functions below should be moved into the framework-specific `__init__.py` files

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_sums_products_differences.py
```diff
@@ -9,41 +9,91 @@
 from ivy_tests.test_ivy.helpers import handle_cmd_line_args
 
 
-# prod
+@st.composite
+def _dtype_x_axis(draw, **kwargs):
+    dtype, x, shape = draw(helpers.dtype_and_values(**kwargs, ret_shape=True))
+    axis = draw(helpers.ints(min_value=0, max_value=len(shape) - 1))
+    if random.randint(0, 9) % 2 != 0:
```
djl11, line 16: we should not use `random.randint` to generate randomness, all randonmess should come from hypothesis, something like `axis = draw(helpers.ints(min_value=0, max_value=len(shape) - 1) | st.just(None))` would be better. This might not run, but something along these lines will work.

djl11 (approved):
lgtm!

Daniel4078:
> lgtm!

Did you find anything wrong with my new implementation  and test of the sum and prod function?

Daniel4078:
I have some problem fixing the not passing tests:
First, AssertionError: the return with a TensorFlow backend produced data type of float16, while the return with a numpy backend returned a data type of float32.
Second,      
# check if out is None or out is not present in args and kwargs.
out_args = out_arg_pos < len(args) and args[out_arg_pos] is None
out_kwargs = "out" in kwargs and kwargs["out"] is None
if ret.shape == () and (out_args or out_kwargs):
    AttributeError: 'int' object has no attribute 'shape'
Finally, 
cls = <class 'ivy.Shape'>, shape_tup = ivy.array([1])

    def __new__(cls, shape_tup):
        valid_types = (int, list, tuple)
        if len(backend_stack) != 0:
            valid_types += (ivy.NativeShape,)
        assert isinstance(shape_tup, valid_types)
 AssertionError

Daniel4078:
To be more specific, I do not know how to convert a ivy.array to ivy.shape. And the test does not deal with int returns correctly. And the same type error that is commonly faced by many people when dealing with numpy.

djl11:
These failures now seem unrelated to the specific task, and are instead general failures with Ivy. I have therefore requested a review from @sherry30 for this. Thanks both!

sherry30 (changes requested):
Hi,

Sorryfor the late review. 
I pointed out 2 changes for prod function, let me know if it still fails.

Thanks

sherry30 (changes requested):
Hi,

It's still failing because of the `where` argument being passed in as None.
You're supposed to use `np_frontend_helpers.where()` and `np_frontend_helpers.handle_where_and_array_bools` to generate `where` argument for numpy.
You can checkout any other numpy frontend function which uses `where` asrgument for reference.

Thanks

Daniel4078:
> Hi,
> 
> It's still failing because of the `where` argument being passed in as None. You're supposed to use `np_frontend_helpers.where()` and `np_frontend_helpers.handle_where_and_array_bools` to generate `where` argument for numpy. You can checkout any other numpy frontend function which uses `where` asrgument for reference.
> 
> Thanks

But the function shold work when where is set to None.

Issue #1225, opened 2022-05-29
relu


b4igiveup:

b4igiveup:
Hello there,

Not sure if I can commit here, but the original codes I've run already passed the test, idk where I need to modify the codes.

Thanks

PR #3701, opened 2022-08-28
Relu activation

mnsgrosa:
ToDo https://github.com/unifyai/ivy/issues/1225

### DIFF
--- a/ivy/container/activations.py
+++ b/ivy/container/activations.py
@@ -119,7 +119,7 @@ class ContainerWithActivations(ContainerBase):
 
         """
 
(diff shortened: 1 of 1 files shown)


juliagsy (changes requested):
Hey there! As your task is reformatting related, you'll need to add examples in the docstrings accordingly (eg. container instance examples in its container file, array instance examples in its array file, functional examples (mix of array and container) in its functional file).

More information can be found in the docs - https://lets-unify.ai/ivy/deep_dive/13_docstring_examples.html 

Please request my review again when you're done! Thank you!

PR #3727, opened 2022-08-28
Adding `torch.isclose` for pytorch frontend

hirwa-nshuti:
Hi this PR resolves issue #3673 

Added `isclose` function to the torch frontend.

Thanks


jkeane508 (changes requested):
Hi Felix,

This is one of those functions that it would make sense to add as a ToDo, to add to the Ivy functional API directly. So if you could add a ToDo in here somewhere that would be great!

PR #3752, opened 2022-08-29
Adding `isfinite` and `isinf` torch frontend

hirwa-nshuti:
Hi James,

This PR resolves issues #3728 and #3729 
Added `torch.isfinite` and `torch.isinf` to the torch frontend.

Thanks.


jkeane508 (changes requested):
Hey,

By default the `dtype_and_values` helper does not include inf, I think it would make sense to include it in the testing.

jkeane508 (approved):

PR #3794, opened 2022-08-30
fix `help()` for backend functions

CatB1t:
***WIP***
Currently, calling `help(ivy.prod)` before setting the backend, works as expected, showing the correct function signature and the docstring, but this fails after `ivy.set_backend()` is called to set any `backend`.


djl11 (approved):
lgtm!

PR #3805, opened 2022-08-30
added arcsinh to numpy 

jvt3112:
#3799


ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_hyperbolic_functions.py
```diff
@@ -108,3 +108,54 @@ def test_numpy_tanh(
...
+        with_out=with_out,
+        num_positional_args=num_positional_args,
+        native_array_flags=native_array,
+        fw=fw,
+        frontend="numpy",
+        fn_tree="arcsinh",
+        x=np.asarray(x, dtype=input_dtype[0]),
+        out=None,
+        where=where,
+        casting="same_kind",
```
Daniel4078, line 156: I think you shoulduse hypothesis test to generate different values for those optional arguments like casting, order and subok.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_hyperbolic_functions.py
```diff
@@ -108,3 +108,54 @@ def test_numpy_tanh(
...
+
+
+# arcsinh
+@handle_cmd_line_args
+@given(
+    dtype_and_x=helpers.dtype_and_values(available_dtypes=ivy_np.valid_float_dtypes),
+    dtype=st.sampled_from(ivy_np.valid_float_dtypes + (None,)),
+    where=np_frontend_helpers.where(),
+    as_variable=helpers.array_bools(num_arrays=1),
+    with_out=st.booleans(),
```
Daniel4078, line 120: the as_variable, with_out and native_array should not be given any value in the given section as they will be handled by the handle_cmd_line_args automatically.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_hyperbolic_functions.py
```diff
@@ -108,3 +108,54 @@ def test_numpy_tanh(
...
         test_values=False,
     )
+
+
+# arcsinh
+@handle_cmd_line_args
+@given(
+    dtype_and_x=helpers.dtype_and_values(available_dtypes=ivy_np.valid_float_dtypes),
+    dtype=st.sampled_from(ivy_np.valid_float_dtypes + (None,)),
+    where=np_frontend_helpers.where(),
```
Daniel4078, line 118: This way of generating where will cause the error of where having different size with the x array, so you should use a function like this:
@st.composite
def _dtype_x_axis(draw, **kwargs):
    dtype, x, shape = draw(helpers.dtype_and_values(**kwargs, ret_shape=True))
    axis = draw(st.one_of(helpers.ints(min_value=0, max_value=max(len(shape) - 1, 0)), 
                          st.none()))
    where = draw(st.one_of(helpers.array_values(dtype=ivy.bool, shape=shape), 
                           st.none()))
    return (dtype, x, axis), where
to generate those arguments.

Daniel4078 (changes requested):
Hello and I have found several things that you need to change before this pull request can be approved. Please check the comments below for more details. Thank you.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_hyperbolic_functions.py
jvt3112: Isn't this function similar to the fronted Numpy tan function as implemented on this link: https://lets-unify.ai/ivy/deep_dive/17_ivy_frontends_tests.html ?

Daniel4078 (changes requested):
Firstly, I do not think you should include the __init__.pyc file in this pull request.
Secondly, TypeError: arcsinh() got an unexpected keyword argument 'axis' in your test, so you should not have axis passed into the test.

Daniel4078 (approved):
well done, I think there is nothing wrong with your code now and the pull request can be merged.

PR #3828, opened 2022-08-30
some fixes to linalg.py

zaeemansari70:


ivy/functional/ivy/linear_algebra.py
```diff
@@ -792,6 +797,7 @@ def matrix_power(
 def matrix_rank(
     x: Union[ivy.Array, ivy.NativeArray],
     rtol: Optional[Union[float, Tuple[float]]] = None,
```
sherry30, line 799: This should be a keyword only argument

sherry30 (changes requested):
Hi Zaeem,

Some of these seem wrong. some optional argument are positional only argumetns.
I've pointed out of them on the specific line.

Thanks

PR #3839, opened 2022-08-30
implemented tensorflow tensordot in the tensorflow linalg frontend.

vaatsalya123:
#3734


ivy/functional/frontends/tensorflow/linalg.py
```diff
@@ -39,3 +39,8 @@ def pinv(a, rcond=None, validate_args=False, name=None):
 
 
 pinv.unsupported_dtypes = ("float16", "bfloat16")
+
+def tensordot(x,y,axes,name=None):
+    return ivy.tensordot(x,y,axes)
+
+tensordot.unsupported_dtypes = ("float16", "bfloat16")
```
zaeemansari70, line 46: Tensordot() for tensorflow supports only `float32` and `float64` according to the docs (https://www.tensorflow.org/api_docs/python/tf/tensordot).
Kindly add only the supported_dtypes.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_linalg.py
```diff
@@ -246,3 +246,35 @@ def test_tensorflow_pinv(
         validate_args=False,
         name=None
     )
+
+
+@handle_cmd_line_args
+@given(
+    dtype_and_x=_solve_get_dtype_and_data(),
```
zaeemansari70, line 253: This `composite function` does not return `axis`, which causes the frontends tests to fail for `tensordot()`, causing an error: `tensordot() missing 1 required positional argument: 'axes'`.
You have to figure out a way to generate `axis` aswell.
You can use some other composite function or write your own if one doesn't exists, which generates the arguments that are needed.
Refer to the docs for more details. (https://lets-unify.ai/ivy/deep_dive/17_ivy_frontends_tests.html#examples)

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_linalg.py
```diff
@@ -246,3 +246,35 @@ def test_tensorflow_pinv(
...
+        as_variable_flags=as_variable,
+        with_out=False,
+        num_positional_args=num_positional_args,
+        native_array_flags=native_array,
+        fw=fw,
+        frontend="tensorflow",
+        fn_tree="linalg.tensordot",
+        x=np.asarray(x, dtype=input_dtype1),
+        y=np.asarray(y, dtype=input_dtype2),
+    )
```
zaeemansari70, line 280: No new line at the end of both files, which causes the `lint` to fail.
There are other `lint` errors that need to be fixed to. 
Thanks!

zaeemansari70 (changes requested):
Hey! First of all thank you soo much for contributing to ivy! 🙂
Your PR looks amazing ✨!
Moreover, you have to un-commit the irrelevant folder which seems to be `.idea` in this case. Kindly revert this folder. 
Thanks!

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_linalg.py
vaatsalya123: let me know if any other changes need to be made thanks!

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_linalg.py
```diff
@@ -246,3 +246,33 @@ def test_tensorflow_pinv(
         validate_args=False,
         name=None
     )
+
+
+)
```
zaeemansari70, line 251: You have to add the `@given` decorator here and have to figure out a way to generate the required values. The best way is to implement a `composite function` or to use an already implemented one. I've also attached the link to the docs, kindly refer to the docs for more detail.
https://lets-unify.ai/ivy/deep_dive/15_ivy_frontends.html
https://lets-unify.ai/ivy/deep_dive/16_ivy_frontends_tests.html

ivy/functional/frontends/tensorflow/linalg.py
```diff
@@ -39,3 +39,7 @@ def pinv(a, rcond=None, validate_args=False, name=None):
 
 
 pinv.unsupported_dtypes = ("float16", "bfloat16")
+
+def tensordot(x,y,axes,name=None):
+    return ivy.tensordot(x,y,axes)
+
```
zaeemansari70, line 45: You have to add the supported or unsupported dtypes here according to the documentation https://www.tensorflow.org/api_docs/python/tf/tensordot

zaeemansari70 (changes requested):
Hey vaatsalya!
You have deleted the `.idea` folder, these changes should be reverted and these files should be un-committed.
They shouldn't be showing in the files changed folder.

Issue #3841, opened 2022-08-30
positive


deepanshug4:
[Add Mathematical functions to Numpy Frontend #1525](https://github.com/unifyai/ivy/issues/1525)

deepanshug4:
This issue is solved and merged.

PR #3858, opened 2022-08-31
Positive function for Numpy Frontend

deepanshug4:


deepanshug4:
This PR is for issue #3841. I couldn't assign this in the Development section, so I thought it would be better to just comment the issue number.

MrZilinXiao:
Hi @deepanshug4. Thanks for your contribution. But would you mind checking the test failures relating to your commits here? https://github.com/unifyai/ivy/runs/8107959439?check_suite_focus=true#step:6:4853
It seems you gave the wrong type for `input_dtype `, which is causing `TypeError: can only concatenate str (not "list") to str`.

deepanshug4:
Hi @MrZilinXiao, Thanks for the review I have fixed that part. Can you please check it out and give reviews about the same.

MrZilinXiao (approved):
Great! No more failures!

PR #3865, opened 2022-08-31
Edit tan parameter in Numpy frontend function

JonaOmara:
tan #3821

### DIFF
--- a/ivy/functional/frontends/numpy/mathematical_functions/trigonometric_functions.py
+++ b/ivy/functional/frontends/numpy/mathematical_functions/trigonometric_functions.py
@@ -59,7 +59,7 @@ def tan(
     *,
     where=True,
     casting="same_kind",
-    order="k",
+    order="K",
     dtype=None,
     subok=True,
 ):
@@ -69,7 +69,6 @@ def tan(
     if ivy.is_array(where):
         ret = ivy.where(where, ret, ivy.default(out, ivy.zeros_like(ret)), out=out)
     return ret
-
 
 tan.unsupported_dtypes = {"torch": ("float16",)}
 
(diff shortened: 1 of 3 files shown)


ivy/functional/frontends/numpy/mathematical_functions/trigonometric_functions.py
```diff
@@ -70,7 +70,6 @@ def tan(
         ret = ivy.where(where, ret, ivy.default(out, ivy.zeros_like(ret)), out=out)
     return ret
 
-
```
Daniel4078, line 73: You should not remove this blank line as the formatting rule requires two blank line after function definitions.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_trigonometric_functions.py
```diff
@@ -73,6 +73,7 @@ def test_numpy_cos(
     ),
     native_array=helpers.array_bools(num_arrays=1),
 )
+
```
Daniel4078, line 76: There should not be a blankline here between the decorator and its corrisponding function.

ivy/functional/ivy/elementwise.py
```diff
@@ -5210,13 +5210,13 @@ def tan(
...
     >>> y = ivy.zeros(3)
-    >>> ivy.tan(x, out=y)
+    >>> ivy.tan(x,out=y)
     >>> print(y)
     ivy.array([0.546, -0.842, -0.916])
 
     >>> x = ivy.array([[1.1, 2.2, 3.3],\
                         [-4.4, -5.5, -6.6]])
-    >>> ivy.tan(x, out=x)
+    >>> ivy.tan(x,out=x)
```
Daniel4078, line 5219: Even though this does not cause a formatting error as it is in the examples, but in actural code, you should add a space between the two arguments right after the comma. So I guess you should not make these changes to the examples.

Daniel4078 (changes requested):
Other than the problems I pointed out in the comments below, there is no other errors with your code. Thank you.

Daniel4078 (approved):
I think this pull request is fine now. Thank you.

PR #3871, opened 2022-08-31
added numpy frontend linalg inv pinv with test

sharique1006:
#3860 
#3888


sherry30 (changes requested):
Hi Sharique,

You're not supposed to commit any changes to any of the files in .idea folder, can you remove those. (`ivy.iml` and `misc.xml`)

There are a bunch of lint errors, so you will have to fix those as well.
Let me know if you need help with anything.

Thanks

PR #3874, opened 2022-08-31
Added `less`, `lt`, `not_equal`, and `ne` to torch frontend

hirwa-nshuti:
Hi James

This PR resolves issues #3869 #3870 #3872 and #3873

I added `less` an alias of `lt` and `not_equal` an alias of `ne` to the torch frontend.

Thanks


jkeane508 (approved):
Looks good to me.

I think we have to wait for the issues to be sorted before merging

PR #3880, opened 2022-08-31
Auto dtypes

Unevilicorn:


djl11 (approved):
didn't have time to go through in crazy detail, but in general it looks good to me :)

Just make sure you are not introducing a load of new test failures. If not, then it's good to merge I think!

PR #3890, opened 2022-08-31
Triu indices

### LINKED ISSUE #3818: triu_indices

JamieLine:
Fixes #3818.

Local tests pass on my PC.


mattbarrett98 (changes requested):
Is there a simpler implementation similar to what you did for `trip_indices`? You should also sample the dtypes not fix to `int32`.

JamieLine:
There is a similar implementation to `tril_indices`, I'm going to pause work on this function until `tril_indices` is done because I think that'll be much more efficient.

JamieLine:
Now this is just a copy of `tril_indices` with very small changes and `tril_indices` has already been reviewed.

PR #3894, opened 2022-08-31
`vecdot()` fixes for linalg.py

zaeemansari70:
1) Tensorflow:
I've added all the dtypes in `unsupported dtypes`, below vecdot(), for which `tensorflow` backend seems to fail when it's dtype is type-casted back to the original dtype of input. The inputs have to be type-casted for tensorflow (https://www.tensorflow.org/api_docs/python/tf/tensordot).

2) PyTorch:
The implementation of torch backend does not require any input to be type-casted. (https://pytorch.org/docs/stable/generated/torch.tensordot.html)
The `ivy_tests` seem to pass even if I type-cast for the torch backend or not, but the `array_api_tests` fail when I don't type-cast the torch implementation which shouldn't be the case as `torch.tensordot` does not require any type-casting according to the docs. I've commented the line which is type-casting the inputs, this should be the original implementation, without type-casting the inputs. But, it causes the `array_api_tests` to fail. for pytorch


ivy/functional/backends/tensorflow/linear_algebra.py
```diff
@@ -445,12 +445,16 @@ def vecdot(
     *,
     out: Optional[Union[tf.Tensor, tf.Variable]] = None
 ) -> Union[tf.Tensor, tf.Variable]:
-    x1, x2 = tf.cast(x1, tf.float32), tf.cast(x2, tf.float32)
-    ret = tf.tensordot(x1, x2, (axis, axis))
+    dtype = tf.experimental.numpy.promote_types(x1.dtype, x2.dtype)
+    if dtype != 'float64':
+        x1, x2 = tf.cast(x1, tf.float32), tf.cast(x2, tf.float32)
+    if dtype == 'float64':
```
djl11, line 451: `else`

ivy/functional/backends/tensorflow/linear_algebra.py
```diff
@@ -445,12 +445,16 @@ def vecdot(
...
+        x1, x2 = tf.cast(x1, tf.float64), tf.cast(x2, tf.float64)
+    ret = tf.cast(tf.tensordot(x1, x2, (axis, axis)), dtype)
     return ret
 
 
-vecdot.unsupported_dtypes = ("int8",)
+vecdot.unsupported_dtypes = ("int8", "int16", "int32", "int64", "uint8", "uint16", "uint32", "uint64")
```
djl11, line 457: you should use `supported_dtypes` attribute here to reduce the length of the tuple

djl11 (changes requested):
small changes requested.

djl11:
@sherry30 @ZoeCD could you both try to give feedback on point 2? I know you have worked on the torch Array API Tests recently @ZoeCD, and I know you have been working closely with Zaeem @sherry30. Thanks!

ivy/functional/backends/tensorflow/linear_algebra.py
zaeemansari70: Done

ivy/functional/backends/tensorflow/linear_algebra.py
zaeemansari70: Done

sherry30:
Seems like `torch.tensordot` doesn't work when both inputs have different dtypes. A possible fix is that you cast both inputs to the promoted dtype and then pass it in `torch.tenosordot`. Also no need to type cast the return of `torch.tensordot`.

One strange behaviour i noticed with `torch.tensordor`, it doesn't work when there is only 1 element in the given axis for dtypes `int32` and lower. Also tried this in colab with torch version 1.12.1 and still the same problem. The rest of the backends are working as expected.

Here's an example that doesn't work:
```
dtype = torch.int32
x1 = torch.Tensor([[1, 2]]).type(dtype)
x2 = torch.Tensor([[1, 1]]).type(dtype)
axis = 0
y2 = torch.tensordot(x1, x2, dims=([axis], [axis]))
```

with the error `expected scalar type Long but found Int`
But works fine with float dtypes and int64

PR #3896, opened 2022-08-31
num_positional args and kwargs for linalg.py

zaeemansari70:


ivy/array/linear_algebra.py
```diff
@@ -124,46 +127,54 @@ def cross(
...
     ) -> ivy.Array:
-        return ivy.matrix_norm(self._data, ord, keepdims, out=out)
+        return ivy.matrix_norm(self._data, ord=ord, keepdims=keepdims, out=out)
 
     def matrix_rank(
-        self: ivy.Array,
-        rtol: Optional[Union[float, Tuple[float]]] = None,
-        *,
-        out: Optional[ivy.Array] = None,
+            self: ivy.Array,
```
sherry30, line 174: you don't have to over indent these arguments

sherry30 (changes requested):
Hi Zaeem,

seems like you've over indented the arguments of all the functions, this won't cause lint errors, but its not the standard practice, can you change those back.

The rest looks good, just make sure docstring tests are not failing because of these changes

sherry30 (approved):
lgtm

Can you just make sure that none of the docstring tests are failing in linalg because of the change
other than that you can go ahead and squash and merge 👍

PR #3919, opened 2022-09-01
Updating _build() and _create_variables() for torch and jax

modanesh:


ivy/functional/backends/torch/converters.py
```diff
@@ -28,13 +28,13 @@ def __init__(
         )
         ivy.Module.__init__(self, device=device, devices=devices)
 
-    def _create_variables(self, device):
+    def _create_variables(self, device, dtype):
         return self._native_params
 
-    def _build(self):
+    def _build(self, *a, **kw):
```
djl11, line 34: change to `*args` and `**kwargs`

djl11 (changes requested):
small change requested.

djl11:
@RickSanchezStoic could you take a look and make sure these changes are okay? They were needed in order to get some tests passing for another repo, and I know you've been working on the `ivy.Module` in general.

ivy/functional/backends/torch/converters.py
```diff
@@ -28,13 +28,13 @@ def __init__(
...
-    def _create_variables(self, device):
+    def _create_variables(self, device, dtype):
         return self._native_params
 
-    def _build(self):
+    def _build(self, *args, **kwargs):
         self._native_module = ivy.default(
             lambda: self._native_module,
-            lambda: self._native_module_class(*self._args, **self._kwargs),
+            lambda: self._native_module_class(*args, **kwargs),
```
RickSanchezStoic, line 37: For `torch's Module` class  , there is no `dtype` argument, so this line will raise an error. https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module. Also, I would suggest running the `test_converteres` to look for any more errors.

RickSanchezStoic: @djl11 As of now this is not causing issues with `ivy.Module` but it is causing issues with `ivy.converters` .

modanesh: @RickSanchezStoic Thanks. As I ran the `test_converters.py`, there were no failing tests (4 skipped ones). How may I check the potential errors?

PR #3927, opened 2022-09-01
Subset of Reduction Ops for Pytorch Frontend

ahmedo42:
#3922 
#3923 
#3924 
#3925 
#4012 
#4013


ahmedo42:
Pending support for tuples in the out argument in `test_frontend_function`

ivy/functional/frontends/torch/reduction_ops.py
```diff
@@ -0,0 +1,33 @@
+import ivy
+
+
+def argmax(input, dim=None, keepdim=False):
```
saeedashrraf, line 4: Hey @ahmedo42 , dim should not have a default value of `None` as mentioned in the docs [here](https://pytorch.org/docs/stable/generated/torch.argmax). Please review the arguments of the other functions as well to make sure they are consistent with torch's docs

saeedashrraf (dismissed):

ivy/functional/frontends/torch/reduction_ops.py
ahmedo42: I think that I should keep that `dim=None` because if we don't, the user has to explicitly add it which is not the same behavior  in torch

saeedashrraf: @ahmedo42 you seem to be right. Sry for the inconvenience but It seems to be a bug with torch's docs. Anyway, in that case, I agree we should add `dim=None` to all the functions. Thanks!

ivy/functional/frontends/torch/reduction_ops.py
```diff
@@ -0,0 +1,33 @@
...
+    return ivy.argmax(input, axis=dim, keepdims=keepdim)
+
+
+def argmin(input, dim=None, keepdim=False):
+    return ivy.argmin(input, axis=dim, keepdims=keepdim)
+
+
+def amax(input, dim, keepdim=False, *, out=None):
```
saeedashrraf, line 12: Please add `dim = None` to this too to capture the same behavior as torch's function. It should also be added to `amin`.

saeedashrraf (changes requested):

saeedashrraf (approved):

saeedashrraf:
Hello @ahmedo42, looks good to me!

PR #3944, opened 2022-09-01
ArgMax

dovaaa:
#3939


Daniel4078:
Please do not forget to add a hypothesis test for ArgMax function in the ivy\ivy_tests\test_ivy\test_frontends\test_tensorflow\test_raw_ops.py file, After that we can determin if your code behaves like what we expect. Thank you!

dovaaa:
@Daniel4078 Please Review again

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
```diff
@@ -289,6 +289,39 @@ def test_tensorflow_Asin(
...
+    ),
+)
+def test_tensorflow_ArgMax(
+    dtype_x_axis, as_variable, num_positional_args, fw, native_array, output_type
+):
+    dtype, x, axis = dtype_x_axis
+    helpers.test_frontend_function(
+        input_dtypes=dtype,
+        as_variable_flags=as_variable,
+        with_out=False,
```
Daniel4078, line 313: with_out should be passed in through the arguments of test_tensorflow_ArgMax as the @handle_cmd_line_args will fill in the with_out argument automatically.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
```diff
@@ -289,6 +289,39 @@ def test_tensorflow_Asin(
...
+        num_positional_args=num_positional_args,
+        native_array_flags=native_array,
+        fw=fw,
+        frontend="tensorflow",
+        fn_tree="raw_ops.ArgMax",
```
Daniel4078, line 318: the content of fn_tree argument should be the path to the origenal function in the target framework, in this case is math.argmax.

Daniel4078 (changes requested):
I just found two small errors with your test, please check my comments below and fix them. Thank you!

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
dovaaa: I actually Implemented the one in rawops, There is another one with the same name in math

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
dovaaa: I fixed and pushed

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
Daniel4078: Well, this is my mistake then, thank you.

Daniel4078 (commented):
Well done, it seems like there is no error with your code now and your test is passing. Just one small suggestion, you should try generate different input using hypothesis for the output_type argument for the test so that all the behaviors will be covered. Thank you.

dovaaa:
> Well done, it seems like there is no error with your code now and your test is passing. Just one small suggestion, you should try generate different input using hypothesis for the output_type argument for the test so that all the behaviors will be covered. Thank you.

it actually fails when i do so due to incompatible data types i think, I dont know how to fix it

```
tensors = pywrap_tfe.TFE_Py_Execute(ctx._handle, device_name, op_name,
                                            inputs, attrs, num_outputs)
E                                           tensorflow.python.framework.errors_impl.InvalidArgumentError: Value for attr 'output_type' of int8 is not in the list of allowed values: int16, uint16, int32, int64
E                                           	; NodeDef: {{node ArgMax}}; Op<name=ArgMax; signature=input:T, dimension:Tidx -> output:output_type; attr=T:type,allowed=[DT_FLOAT, DT_DOUBLE, DT_INT32, DT_UINT8, DT_INT16, 16534343205130372495, DT_COMPLEX128, DT_HALF, DT_UINT32, DT_UINT64, DT_BOOL]; attr=Tidx:type,default=DT_INT32,allowed=[DT_INT16, DT_INT32, DT_INT64]; attr=output_type:type,default=DT_INT64,allowed=[DT_INT16, DT_UINT16, DT_INT32, DT_INT64]> [Op:ArgMax]

/usr/local/lib/python3.8/dist-packages/tensorflow/python/eager/execute.py:54: InvalidArgumentError
```

Daniel4078:
You just need to update your implementation of the function to accept only "int16", "uint16", "int32", "int64" as the input for output_type and treat all other inputs as the default value like what you did when output_type = None.

dovaaa:
I added static Type checking in ArgMax

Daniel4078 (changes requested):
Well, you need to change your test so that the generated type is only one of the four "int16", "uint16", "int32", "int64" as the test also pass these value to the real tensorflow function which also only accept these types. Except this, there is no other problem with your code, Thank you.

dovaaa:
I Fixed The issue with Data types but it fails in Torch due to unsupported casting by them i think, i attached the error

```
dtype_in = 'uint16'

    def as_native_dtype(dtype_in: Union[torch.dtype, str]) -> torch.dtype:
        if not isinstance(dtype_in, str):
            return dtype_in
        if dtype_in in native_dtype_dict.keys():
            return native_dtype_dict[ivy.Dtype(dtype_in)]
        else:
>           raise TypeError(
                f"Cannot convert to PyTorch dtype. {dtype_in} is not supported by PyTorch."
            )
E           TypeError: Cannot convert to PyTorch dtype. uint16 is not supported by PyTorch.

ivy/functional/backends/torch/data_type.py:168: TypeError
```

Daniel4078 (commented):
Well, then please just stop generate 'uint16' as output_type. Thank you.

dovaaa:
> Well, then please just stop generate 'uint16' as output_type. Thank you.

I actually added ArgMax.unsupported_dtypes = {"torch": ("uint16",)} 

but it doesnt seem to work right now

@Daniel4078  Please Re-review

dovaaa:
> By the way, there is a big change to how we deal with unsupported datatype when implementing frontend functions, please have a look: https://lets-unify.ai/ivy/deep_dive/6_data_types.html#unsupported-data-types

I actually already use this in the front end implementation

Daniel4078:
> > By the way, there is a big change to how we deal with unsupported datatype when implementing frontend functions, please have a look: https://lets-unify.ai/ivy/deep_dive/6_data_types.html#unsupported-data-types
> 
> I actually already use this in the front end implementation

Sorry, it seems like the document is not updated yet. The change I mentioned will soon be up there. Thank you!

Daniel4078:
here is the update of the doc I was talking about https://github.com/unifyai/ivy/commit/95c7427f30406addec1e1aff9342365025e4808b

Daniel4078 (commented):
Well, except the small change to the format of how to state the unsupported dtype, everything else seems right. Thank you for your work!

ivy/functional/frontends/tensorflow/raw_ops.py
```diff
@@ -14,6 +14,16 @@ def concat(values, axis, name="concat"):
...
 
 
+def ArgMax(*, input, dimension, output_type=None, name=None):
+    if output_type in ["uint16", "int16", "int32", "int64"]:
+        return ivy.astype(ivy.argmax(input, axis=dimension), output_type)
+    else:
+        return ivy.astype(ivy.argmax(input, axis=dimension), "int64")
+
+
+ArgMax.unsupported_dtypes = (ivy.uint16,)
```
Daniel4078, line 24: I think this should be like:
 ArgMax.unsupported_dtypes = ("uint16",)

dovaaa:
> Well, except the small change to the format of how to state the unsupported dtype, everything else seems right. Thank you for your work!

The problem with ArgMax is that it supports all data types,
but one parameter "output_type" doesn't it only supports 4 dtypes

and in torch it doesnt support uint16 as output_type which i cant handle from the frontend

as such i think this line is wrong because it does accept uint16 as a data type for some parameters

```
ArgMax.unsupported_dtypes = (ivy.uint16,)
```

dovaaa:
Is There anything else i need to do? @Daniel4078

Daniel4078 (approved):
Currently, you test passes with no errors and I can not find other problems with your code. You may remove the unsupported_dtype line as a test to see if it changes anything. Then I think this pull request can be merged.

dovaaa:
> Currently, you test passes with no errors and I can not find other problems with your code. You may remove the unsupported_dtype line as a test to see if it changes anything. Then I think this pull request can be merged.

It also passes pushing now

Daniel4078:
I see, then I will merge this as the tests still passes normally. Thank you.

PR #3955, opened 2022-09-02
Adding `msort` to torch frontend

hirwa-nshuti:
Hi
This PR resolves issue #3936 

Added `msort` to the torch frontend.


ivy/functional/frontends/torch/comparison_ops.py
```diff
@@ -1,4 +1,9 @@
-# ToDo: Add allclose(), isclose(), isposinf(), isneginf(), fmax() to functional API
+"""
+ToDo
+----
+Add allclose(), isclose(), isposinf(), isneginf(), fmax(), msort()
```
jkeane508, line 4: I think if we are only using 1 ivy function then its not worth adding to the API

jkeane508 (changes requested):
Small Change :)

ivy/functional/frontends/torch/comparison_ops.py
hirwa-nshuti: Yeah it makes sense.

PR #3959, opened 2022-09-02
numpy frontend linalg - norms and numbers

sharique1006:
slogdet - #3912


ivy/functional/frontends/numpy/linear_algebra/linalg/norms_and_other_numbers.py
```diff
@@ -0,0 +1,13 @@
+# global
+import ivy
+
+# slogdet
+
+
+def slogdet(a):
+    sign, logabsdet = ivy.slogdet(a)
+    ret = ivy.concat((ivy.reshape(sign, -1), ivy.reshape(logabsdet, -1)))
```
sherry30, line 9: can you pass -1 as a tuple, because `ivy.reshape` takes a sequence of int for `shape`

sherry30 (changes requested):
one small change requested

Thanks

PR #3968, opened 2022-09-02
added softmin to torch frontend

irmakaras:
#3966


ivy_tests/test_ivy/test_frontends/test_torch/test_non_linear_activation_functions.py
```diff
@@ -175,3 +175,44 @@ def test_torch_leaky_relu(
...
+@given(
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=tuple(
+            set(ivy_np.valid_float_dtypes).intersection(
+                set(ivy_torch.valid_float_dtypes)
+            )
+        ),
+        min_num_dims=1,
+    ),
+    axis=st.integers(-1, 0),
```
Daniel4078, line 190: I think that the possible range of axis is not between -1 and 0 but determined by the dimensioned of the x array generated.

Daniel4078 (changes requested):
Firstly, try resolve the conflicts in the two files so the automated tests can run so we can see if your test passes.
Secondly, please check my comments below about some changes that you need to make.
Thank you.

ivy_tests/test_ivy/test_frontends/test_torch/test_non_linear_activation_functions.py
irmakaras: I am a little bit confused about how to proceed with the implementation of this. Should I implement a loop logic inside the test_torch_softmin function to account for all axes of the input? Or is there a way to do it in the hypothesis decorator?

(Sorry, I'm new to hypothesis and testing)

Daniel4078: You can consider adding a helper function to the test file and use it in the given section like this one:
@st.composite
def _dtype_x_bounded_axis(draw, **kwargs):
    dtype, x, shape = draw(helpers.dtype_and_values(**kwargs, ret_shape=True))
    axis = draw(helpers.ints(min_value=0, max_value=len(shape) - 1))
    return dtype, x, axis

Daniel4078: also, this example here need some small fix as when the shape generated by the helper becomes () representing an empty array x, then the max_value of the int should be set to 0 not -1.

Daniel4078:
Please try and resolve the conflicting files and retain changes from both sides, thank you.

irmakaras:
@Daniel4078 I have updated the branch. If you have any change requests I'd be glad to work on them.

Daniel4078 (approved):
Well done, your test passes without error and it seems like there is nothing wrong with your code. Thank you.

PR #3980, opened 2022-09-02
Implemented Tril Function in Numpy Frontend 

cgill7:
Tracked in #3946


MrZilinXiao (approved):
Hi @cgill7. Thanks for your first-time contribution. Those commits look great.

PR #3985, opened 2022-09-02
Add manipulation routine for numpy frontend: roll

jucamohedano:
Implementation of roll function and its test


ivy/functional/frontends/numpy/manipulation_routines/rearranging_elements.py
```diff
@@ -1,6 +1,10 @@
-# local
+# global
```
Unevilicorn, line 1: This should be local, since ivy is a local library

Unevilicorn (changes requested):
Just a small change, I'll get this merged once this is completed

ivy/functional/frontends/numpy/manipulation_routines/rearranging_elements.py
jucamohedano: Thank you for spotting that! I made the change :)

Unevilicorn:
Looks good, I'll get this merged!

PR #4017, opened 2022-09-03
Reformat print_all_arrays_on_dev in device.py 

WilliamHirst:
Fixes the issue #3833 

Changed structure of function arguments as well as altered the docstring.

### DIFF
--- a/ivy/functional/ivy/device.py
+++ b/ivy/functional/ivy/device.py
@@ -295,7 +295,10 @@ def num_ivy_arrays_on_dev(device: Union[ivy.Device, ivy.NativeDevice], /) -> int
 
 @handle_nestable
 def print_all_ivy_arrays_on_dev(
-    device: Union[ivy.Device, ivy.NativeDevice], /, *, attr_only: bool = True
+    device: Union[ivy.Device, ivy.NativeDevice], 
+    /, 
+    *, 
+    attr_only: bool = True
 ) -> None:
     """
     Prints the shape and dtype for all ivy arrays which are currently alive on the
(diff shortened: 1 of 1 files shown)


WilliamHirst:
add_reformatting_checklist_ device

WilliamHirst:
add_reformatting_checklist_device

ivy/functional/ivy/device.py
```diff
@@ -305,8 +308,30 @@ def print_all_ivy_arrays_on_dev(
...
+    
     attr_only
         Whether or not to only print the `shape` and `dtype` attributes of the array
+    
+    Examples
+    --------
+
+    With attr_only = True
+
+    >>> x = ivy.array([1,0,2], [3,2])
```
saeedashrraf, line 320: Hello @WilliamHirst , thank you for your effort! The array should have an identical number of elements per dimension. So , you can change x to be = `ivy.array([[1,0,2],[3,2,1]])` instead of `ivy.array([1,0,2], [3,2])`. You should then update the function call's results accordingly.

saeedashrraf (changes requested):

ivy/functional/ivy/device.py
WilliamHirst: Oh yes, I missed that! Thank you for the feedback. It is now fixed.

ivy/functional/ivy/device.py
```diff
@@ -305,8 +308,30 @@ def print_all_ivy_arrays_on_dev(
...
+    
     attr_only
         Whether or not to only print the `shape` and `dtype` attributes of the array
+    
+    Examples
+    --------
+
+    With attr_only = True
+
+    >>> x = ivy.array([1,0,2], [3,2,1])
```
saeedashrraf, line 320: The array constructor does not accept two arrays as input. You should change the input to be a single array with 2 dims instead as `x = ivy.array([[1,0,2], [3,2,1]])`. The same applies to the other example. Thanks!

ivy/functional/ivy/device.py
```diff
@@ -305,8 +308,30 @@ def print_all_ivy_arrays_on_dev(
...
     device
         The device on which to print the arrays
+    
     attr_only
         Whether or not to only print the `shape` and `dtype` attributes of the array
+    
+    Examples
+    --------
+
+    With attr_only = True
```
saeedashrraf, line 318: This line should be removed

ivy/functional/ivy/device.py
```diff
@@ -305,8 +308,30 @@ def print_all_ivy_arrays_on_dev(
...
+
+    >>> x = ivy.array([1,0,2], [3,2,1])
+    >>> y = ivy.dev(x)
+    >>> ivy.print_all_ivy_arrays_on_dev(y)
+    ((3,), 'int32')
+    ((3,), 'int32')
+    
+    
+
+    With attr_only = False:
```
saeedashrraf, line 328: This line should be removed

saeedashrraf (changes requested):

ivy/functional/ivy/device.py
WilliamHirst: Both comments should now be fixed.

saeedashrraf (approved):

saeedashrraf:
Hello @WilliamHirst, thank you for your effort!

Issue #3791, opened 2022-08-30
inplace_decrement


aryans1204:

aryans1204:
waiting for review from @zaeemansari70

PR #4021, opened 2022-09-04
inplace_decrement #3791

aryans1204:
…d all lambda bindings, updated type hints

### DIFF
--- a/ivy/functional/ivy/general.py
+++ b/ivy/functional/ivy/general.py
@@ -2732,7 +2732,7 @@ def assert_supports_inplace(x: Union[ivy.Array, ivy.NativeArray]) -> bool:
         True if support, raises exception otherwise
 
     """
-    if not ivy.supports_inplace(x):
+    if not ivy.supports_inplace_updates(x):
         raise Exception(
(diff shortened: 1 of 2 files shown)


aryans1204:
add_reformatting_checklist_general

aryans1204:
awaiting review from @rush2406

aryans1204:
to do list issue #3791

CatB1t (changes requested):
Hello @aryans1204, Please make changes that address only one function, `inplace_decrement`. 
Also we should be adding docstring examples to this function here: https://github.com/unifyai/ivy/blob/8c677baa152cf8f91534b8118e89b492dd6a01d8/ivy/functional/ivy/general.py#L2763

aryans1204:
Hi,

From the Deep Dive, I had believed that while working on function reformatting issues, we had to fix all lambda bindings and type hints for the five backends. So do you want me to remove the type hints and lambda bindings I fixed for other functions other than inplace_decrement?

CatB1t:
Hello @aryans1204, exactly, the task should only update 1 function at a time, you should only do changes related to the function `inplace_decrement`.

PR #3294, opened 2022-08-20
Added numpy shape function and tests

alvin-98:
- [] #3284

PR #4029, opened 2022-09-04
Added numpy shape function and tests

alvin-98:
Creating a pull request again based on comments by @roboFiddle (Previous PR: #3294)
Issue 
- [] #3284


ivy_tests/test_ivy/test_frontends/test_numpy/test_manipulation_routines/test_basic_operations.py
```diff
@@ -0,0 +1,40 @@
...
+import ivy_tests.test_ivy.helpers as helpers
+import ivy.functional.backends.numpy as ivy_np
+from ivy_tests.test_ivy.helpers import handle_cmd_line_args
+
+
+# shape
+@handle_cmd_line_args
+@given(
+    xs_n_input_dtypes_n_unique_idx=helpers.dtype_and_values(available_dtypes=ivy_np.valid_dtypes),
+    as_variable=helpers.array_bools(),
```
roboFiddle, line 15: add `num_arrays=1` as a parameter to `array_bools` for `as_variable` and `native_shape`

ivy/functional/frontends/numpy/manipulation_routines/basic_operations.py
```diff
@@ -0,0 +1,7 @@
+# local
+from typing import Tuple
+import ivy
+
+
+def shape(array, /):
+    return ivy.shape(array)
```
roboFiddle, line 7: add empty newline at end of file

roboFiddle (changes requested):

alvin-98:
I am getting the below attribute error on running the test function:

`pytest ivy_tests/test_ivy/test_frontends/test_numpy/test_manipulation_routines/test_basic_operations.py`

`E       AttributeError: 'Shape' object has no attribute 'dtype'
E       Falsifying example: test_numpy_shape(
E           xs_n_input_dtypes_n_unique_idx=('int8', -1),
E           num_positional_args=1,
E           native_array=[False],
E           fw='numpy',
E       )
E       Falsifying example: new_fn(
E           data=data(...),
E           get_command_line_flags={'as_variable': None,
E            'native_array': None,
E            'with_out': None,
E            'container': None,
E            'instance_method': None,
E            'test_gradients': None},
E           device='cpu',
E           f=None,
E           fw='',
E       )
E       Draw 1: 'numpy'

ivy_tests/test_ivy/helpers.py:623: AttributeError`

This is the stack trace for the error:
`ivy_tests/test_ivy/helpers.py:3250: in new_fn
    return test_fn(*args, **kwargs)
ivy_tests/test_ivy/test_frontends/test_numpy/test_manipulation_routines/test_basic_operations.py:13: in test_numpy_shape
    @given(
ivy_tests/test_ivy/test_frontends/test_numpy/test_manipulation_routines/test_basic_operations.py:31: in test_numpy_shape
ivy_tests/test_ivy/helpers.py:1944: in test_frontend_function
    value_test(
ivy_tests/test_ivy/helpers.py:769: in value_test
    assert_all_close(`

It seems to originate from this portion of the assert_all_close function:
`assert ret_np.dtype is ret_from_gt_np.dtype, (
            "the return with a {} backend produced data type of {}, while the return with"
            " a {} backend returned a data type of {}.".format(
                ground_truth_backend,
                ret_from_gt_np.dtype,
                ivy.current_backend_str(),
                ret_np.dtype,
            )
        )`

I am not clear on what rep_np and ret_from_gt_np are and both of these show up as empty tuples / arrays when printed. Please help me with resolving this error.

roboFiddle:
I think if you set `min_num_dims=1` in the parameters to `helpers.dtype_and_values` it should resolve that. The issue is coming from the fact that an `ivy.array` with just a scalar has shape `()` which has no `dtype`. Even if this doesn't resolve it, I'll accept this PR because the issue isn't being caused by your code.

alvin-98:
Thanks for the feedback. I tried out what you said.
I added the setting: `min_num_dims=1` . I still got the same error. The error seemed to be due to `ret_np` coming out to be tuples. So, in addition to setting `min_num_dims=1`, I also added `as_array=True` to the frontend shape function in numpy like so -

```
def shape(array, /):
    return ivy.shape(array, as_array=True)
```

This finally gave an assertion error due to a difference in dtypes at the same place. 

```
assert ret_np.dtype is ret_from_gt_np.dtype, (
            "the return with a {} backend produced data type of {}, while the return with"
            " a {} backend returned a data type of {}.".format(
                ground_truth_backend,
                ret_from_gt_np.dtype,
                ivy.current_backend_str(),
                ret_np.dtype,
            )
        )
```

I printed out the dtypes and values of `ret_np` and `ret_from_gt_np`, the former was int32 and the latter int64. This is the reason for the assertion error. 
Also, I am not sure if setting `as_array=True` as the default is a good idea because the default Numpy behaviour is to return the shape as a tuple of ints.

roboFiddle:
It seems like the testing framework we have doesn't cleanly convert the data-types of non-array outputs (which shape has). I just changed it to be a manual equality check and it passes, so I'll merge the PR.

Thanks for the contribution!

PR #4045, opened 2022-09-05
NumPy Frontend - linalg

JonathanYang-yzh:


ivy/functional/frontends/numpy/linear_algebra/linalg/matrix_eigenvalues.py
```diff
@@ -0,0 +1,10 @@
+# for review
+# global
+import ivy
+
+
+def eigh(a, UPLO='L'):
+    return ivy.eigh(a, UPLO=UPLO)
```
sherry30, line 7: `ivy.eigh` doesn't have a `UPLO` argument

sherry30 (changes requested):
Hi,

since `ivy.eigh` doesn't have a `UPLO` argument, you will have to implement it compositionally, if that's not possible, then we will have to add this argument in ivy API.

Thanks

PR #4063, opened 2022-09-05
WIP: accumulate #3897

ddzhulaii:
accumulate #3897

### DIFF
--- a/ivy/functional/frontends/numpy/ufunc/methods.py
+++ b/ivy/functional/frontends/numpy/ufunc/methods.py
@@ -2,4 +2,5 @@ import ivy
 
 
 def accumulate(method, array, axis=0, dtype=None, out=None):
-    return array.__array_ufunc__(method, array=array, axis=axis, dtype=dtype, out=out)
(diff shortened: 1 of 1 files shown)


ddzhulaii:
Hello 👋  I have a few questions about this issue.
I was thinking about three possible ways to solve it:

1. Use already available NumPy functionality (as it's all already implemented in C): use a native NumPy method as input for `method` and convert the input array `array` to a NumPy array -> I believe this does not fit the whole idea of Frontend API (using only Ivy functions)
2. Take as input an Ivy method, for instance, `ivy.add` or `ivy.multiply`, and then with the bunch of if...else use already implemented `ivy.cumsum` , `ivy.cumprod`, etc. But this does not look future proof for me?

What do you think? Thanks in advance :)

kimbring2:
@ddzhulaii Hello, thank you for your PR. I need some time to answer your question well. 

Additioanlly, you need to write a test code for the Frontend function (:

ivy-seed:

ivy-seed:

PR #4069, opened 2022-09-05
Fix for individual task regarding speeding up ivy.asarray

JamieLine:
Previously, the function took several times as long to run as the backend library function for simple Python lists, now that overhead has dropped to a couple of percent. Most of the speed-up comes from removing a `to_native_arrays_and_back` decorator in the case of a Python list, and replacing it with `outputs_to_ivy_arrays` in that case as it provides the same I/O behavior but skips the very expensive step of running `inputs_to_native_arrays` just for it to be an identity function. The rest of the speed-up comes from allowing the backend to do data type inference in the case where `dtype=None`. 

Please feel free to comment on anything / give corrections on anything regardless of how small / insignificant it is, if I know about a mistake it makes it much easier to stop making that mistake.

This code passes all tests discovered by PyCharm for `ivy.asarray`.


sherry30:
Hi,

The tests seem to pass but isn't `_is_raw_python_list_or_scalar` traversing through the whole list recursively anyway. According to the description of the task, there shouldn't be any recursion while checking for inputs and separate decorators should be created for that reason which will be added to the backend specific implementations of `asarray`.

JamieLine:
Hi, 

Thank you for having a look, the difference had come about because I had some misunderstandings about the requirements for `asarray`, I thought it needed to handle lists of `NativeArray` instances or similar, and the other patches had been done outside of decorators as the first time I attempted those I kept getting strange errors, but now that I have done it without decorators I'm working on putting those changes into decorators now.

JamieLine:
One decorator has had to stay on the `ivy.asarray` function to prevent crashes, but this has either no performance impact or extremely close to no impact. I've tested the performance of `ivy.asarray` under different backends and here are the results.

These results are averages over 10 runs, where Ivy(X) means calling Ivy with X set as the backend, X means calling the backend function from that framework directly, and the tuples are either (Time taken in nanoseconds, Time taken / Time taken by the fastest framework) or (Time taken in nanoseconds, Time taken / Time taken by the fastest framework, Time taken / Time taken by calling the framework directly).
```
Ivy (jax)       : (3833.7, 1.0349044379656624, 1.5426154299227328e-07)
jax             : (24851949005.8, 6708764.983749054)
Ivy (tensorflow): (4053.7, 1.0942932728647015, 1.946493713747076e-06)
tensorflow      : (2082565163.9, 562186.9031152143)
Ivy (torch)     : (4002, 1.0803368966634272, 8.120758052050341e-06)
torch           : (492811135.9, 133033.99630169527)
Ivy (numpy)     : (3736.7, 1.008719360760177, 5.733892472967069e-06)
numpy           : (651686444.7, 175922.26668286364)
Ivy (mxnet)     : (3704.4, 1.0, 8.436000333004759e-06)
mxnet           : (439118048.1, 118539.58754454163)
```

I can push the script used to generate these numbers if it would be useful to Ivy to have it, it's a basic script but it does automatically handle outputting the results in a readable way and dealing with the CPU cache causing some runs to go much faster than they should.

sherry30 (changes requested):
Just a few general changes I've requested.
Also i think you can do the same thing with `handle_nestable` decorator by just checking if the first arg is container or not and then calling the static cont method.
You can keep the `handle_out_argument` decorator on ivy `asarray` as that is unchanged.

I'm not sure why removing `asarray_infer_device` on ivy `asarray` would cause it to crash. Can you share the error you get when you run without this decorator.

Thanks

JamieLine:
GitHub decided that I closed this when I fixed a merge conflict, re-opened it. Nothing has been pushed into the main branch at this stage.

JamieLine:
I've tried removing `assary_infer_device` again and now it isn't crashing but I'm not sure why and I don't remember the error it was giving before, but it does appear to be stable. 

I've made the changes, when implementing `asarray_handle_nestable` I realised it could be rewritten to simply call `ContainerWithCreation.static_asarray`, I've decided not to do that for now as it a) doesn't save a measurable amount of time as the overhead is already so low that you can't measure it accurately, and b) making this code call that function directly without the `getattr` call feels like it will make the code inflexible if `ivy.Container` is ever refactored.

When making the change to `asarray_to_native_arrays_and_back` I changed the output to not use `output_to_ivy_array` to take out a small amount of additional overhead.

JamieLine:
I've made the requested changes, and I've added a fix for the container bug you mentioned, it fixes the bug as I have managed to recreate it, it seems like Python was simply misinterpreting arguments to the container wrapper as also being arguments to `asarray`, by changing them to keyword arguments Python doesn't get confused.

sherry30 (approved):
Looks good 👍

PR #4077, opened 2022-09-05
Added tril_indices to Torch frontend

### LINKED ISSUE #4070: tril_indices

JamieLine:
Fixes #4070

### DIFF
--- a/ivy/functional/frontends/torch/miscellaneous_ops.py
+++ b/ivy/functional/frontends/torch/miscellaneous_ops.py
@@ -38,3 +38,41 @@ cumsum.unsupported_dtypes = (
     "uint32",
     "uint64",
 )
+
+
+def tril_indices(row, col, offset=0, *, dtype="int64", device="cpu", layout=None):
+    if row == 0 and col == 0:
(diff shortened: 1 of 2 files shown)


mattbarrett98 (changes requested):
You should test against the arguments `dtype` and `layout` as well. Also, would there not be a shorter solution using `ivy.tril` and `ivy.nonzero`?

mattbarrett98 (changes requested):
You should use `helpers.get_dtypes` for all dtype sampling. Also `dtype_and_values` isn't relevant here because you don't need to generate an array- `row`, `col` and `offset` are just integers, which you can generate with hypothesis' `st.integers`

JamieLine:
This function is as done as it can be until Ivy has support for multiple memory layouts / specifying a memory layout, there is no way to specify to the test anything other than None for the `layout` argument that can be understood by both Ivy and Torch. However for the moment, that doesn't actually affect its usability as Torch only supports one value for this argument anyway, and the I/O behaviour (at some level of abstraction) is unaffected by this argument.

mattbarrett98 (commented):
okay thanks 🙂

PR #4079, opened 2022-09-05
Added trace to Torch frontend

### LINKED ISSUE #4078: trace

JamieLine:
Fixes #4078


mattbarrett98 (changes requested):
How come you aren't making use of `ivy.trace`?

JamieLine:
I think what happened was I looked at trace and thought "Oh that's simple enough I'll quickly do that" and stopped thinking so in the process I over complicated it, now that you've mentioned it using `ivy.trace` is clearly the correct way to do this.

mattbarrett98 (changes requested):
You should use `helpers.get_dtypes("numeric")` to draw dtypes

JamieLine:
I have genuinely no idea why this PR closed itself, I checked nothing was merged into the main repo, I simply fixed a merge conflict while updating my codebase as this was needed to make the code work, and now it has decided that by pushing this merge into my codebase I have closed this, which was not my intent.

JamieLine:
I've made the suggested change, and as `helpers.get_dtype("numeric")` won't read the `unsupported_dtypes` flag it has the `assume("float16" not in dtype)` line as suggested in the internal chat.

mattbarrett98 (commented):
Looks good and the tests are passing 🙂  Just wondering why is it necessary to convert everything to `int64`? and do you know why the `unsupported_dtypes` flag isn't working? This is something we probably shouldn't just ignore.

JamieLine:
The type conversion is done to match Torch's behaviour, and it's done slightly early as a safety measure against overflows, it appears that Torch does the conversion in the same place. Without the `int64` conversion Hypothesis will find cases where the `dtype`s disagree between Ivy and Torch.

The `unsupported_dtypes` flag is ignored because the function `_get_dtypes` is only checking in the case where the function is a backend function, should I make a PR to edit make it read the flag in all cases? This screenshot should help make things clearer.  In the case that `fn` is a frontend function, then `_get_dtypes` is invariant in what function `fn` actually is.

image

mattbarrett98 (changes requested):
Strange, maybe message Ved about why this is happening- I believe he has been involved with writing the test function.

JamieLine:
I've messaged Ved about it

JamieLine:
https://github.com/unifyai/ivy/pull/4518

JamieLine:
Pulled in change to testing framework from master

JamieLine:
The testing weirdness has been resolved, and the CI test passes on this function, it only fails on others.

mattbarrett98 (changes requested):
You can remove `bool` from unsupported dtypes, trace need only support numeric dtypes. Just add `get_dtypes("numeric")` to the test to only test numeric dtypes. If it's still working after this change feel free to merge to master 🙂

PR #4082, opened 2022-09-06
Better run_tests.sh

gjvnq:
I modified the `./run_tests.sh` script to include a help message as well as the possibility of running specific tests and of passing arguments to pytest.

The new interface is compatible with the previous one. That is, calling the script without passing any parameters runs all tests under `ivy_tests/`.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -51,6 +51,7 @@ fi
 
 # Process the arguments to prepend the paths so they will work inside the docker container
 FINAL_ARGS=()
+NUM_TESTS=0
 for ARG in ${RAW_ARGS[@]}; do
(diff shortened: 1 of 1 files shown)


bicycleman15:
Hey @gjvnq , thanks a lot for taking the time to improve our testing suite. 

This looks great !
We were actually planning to do this, however it got lost somewhere in the endless task list 😅 

I am looping in @djl11.
What do you think about this Dan ?

bicycleman15:
Also @gjvnq, could you give some example usage of your changes to the `run_tests.sh` if possible 🙂  ?

gjvnq:
Just run the script with `-h` and it will print a few usage examples.

djl11 (approved):
lgtm!

PR #4086, opened 2022-09-06
numpy frontend linalg matrix rank and det

sharique1006:
det - #3311 
matrix_rank - #3956

### DIFF
--- a/ivy/functional/frontends/numpy/linear_algebra/linalg/norms_and_other_numbers.py
+++ b/ivy/functional/frontends/numpy/linear_algebra/linalg/norms_and_other_numbers.py
@@ -1,6 +1,25 @@
 # global
 import ivy
 
+# matrix_rank
+
+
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/numpy/linear_algebra/linalg/norms_and_other_numbers.py
```diff
@@ -1,6 +1,25 @@
 # global
 import ivy
 
+# matrix_rank
+
+
+def matrix_rank(A, tol=None, hermitian=False):
+    ret = ivy.matrix_rank(x=A, rtol=tol)
```
sherry30, line 8: `x` is a positional only argument in `ivy.matrix_rank`

sherry30 (changes requested):
Hi,

Just a simple change requested on specific line, rest if good 👍 

Thanks

PR #4098, opened 2022-09-06
Bitwise operations for torch frontend

ahmedo42:
#4072 
#4073 
#4074 
#4075
#4076 

The shift operations unit tests are not passing because of a potential bug, see  https://github.com/unifyai/ivy/discussions/2190#discussioncomment-3566729

### DIFF
--- a/ivy/functional/frontends/torch/pointwise_ops.py
+++ b/ivy/functional/frontends/torch/pointwise_ops.py
@@ -165,6 +165,34 @@ def bitwise_and(input, other, *, out=None):
 bitwise_and.unsupported_dtypes = ("float16",)
 
 
+def bitwise_not(input, *, out=None):
+    return ivy.bitwise_invert(input, out=out)
+
+
+bitwise_not.unsupported_dtypes = ("float16",)
+
+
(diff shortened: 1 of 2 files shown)


saeedashrraf (commented):
Hello @ahmedo42 , we need to implement the frontend function in a way that captures this behavior. You can add a check to see if the shift number is larger than or equal to the number of bits in that dtype. If it is larger, you can just return 0. Otherwise, you can make a call to ivy's shift function.

ahmedo42:
> You can add a check to see if the shift number is larger than or equal to the number of bits in that dtype. If it is larger, you can just return 0. Otherwise, you can make a call to ivy's shift function like what is currently being done.

That's not enough to replicate the Pytorch behavior, I don't quite understand how Pytorch deals with this under the hood, for example:

```python3
import ivy
import torch
ivy.set_backend("torch")
t = torch.tensor(-1, dtype=torch.int8)
x = torch.tensor(100, dtype=torch.int8)
print(torch.bitwise_left_shift(t, x)) #tensor(-16, dtype=torch.uint8)
print(ivy.bitwise_left_shift(t,x)) #ivy.array(-128)

```

PR #4133, opened 2022-09-06
Added cartesian_prod to Torch frontend

### LINKED ISSUE #4128: cartesian_prod

JamieLine:
Fixes #4128


JamieLine:
Regarding the comment above the test for `cartesian_prod`, the function supports arbitrarily many inputs, but I cannot work out a way to have the test produce arbitrarily many inputs without the test throwing `TypeError: expected Tensor as element 0 in argument 0, but got list` when calling the real PyTorch function. I'm not sure if there is something I am missing or if this is a suitable workaround.

mattbarrett98 (changes requested):
Unfortunately you will need to figure out a way to test against multiple arrays, hopefully the test for `meshgrid` can help you out. Please also make the general changes I've requested on your other PRs as well 🙂

JamieLine:
Re-opening.

JamieLine:
I've added in the test setup to have variable numbers of tensors and improved the implementation of the frontend function itself to use Ivy functions rather than `itertools`

mattbarrett98 (commented):
Is the custom strategy necessary? I think you could just use `dtype_and_values` directly as in `test_meshgrid`

JamieLine:
I had it in there to prevent a case where simultaneously `num_arrays==1` and `dim_size==1` which was causing crashes, but now I can't replicate this crash and I could put some time in to work out why it isn't crashing now but I don't think it's a good use of time. I think if that bug had persisted then it would be worth the extra strategy (to allow for the valid cases where only one of those values were `1`) but since it is gone I've removed it.

mattbarrett98 (approved):
Looks good! Just to mention when I was checking through the torch frontend failures, a few of the previous functions you implemented are failing: `trace` `triu`, `triu_indices` I believe, could you take a look 🙂

PR #4136, opened 2022-09-06
vmap for all backends and the test

Ishticode:
`ivy.vmap` implementations inspired by `jax.vmap`. 
The `test_vmap` is passing for the backends `[numpy, tensorflow, jax, torch]`. Checked with a max of 500 hypothesis generated examples and made sure its hitting actual examples and not passing for the wrong reasons.

The implementation will be improved using optimized vectorisation functions by the host frameworks and extended/adding more functionality/control but this is the first iteration and has the essential working components.

Here is a list of additions.
 - vmap in ivy
 - vmap for all ivy backends
 - vmap test
 - helper function arrays_and_axes

### DIFF
--- a/ivy/functional/backends/jax/general.py
+++ b/ivy/functional/backends/jax/general.py
@@ -9,7 +9,7 @@ from numbers import Number
 from operator import mul
 from functools import reduce
 from jaxlib.xla_extension import Buffer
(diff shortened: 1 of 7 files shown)


djl11 (approved):
lgtm, thanks @Ishticode!

PR #4139, opened 2022-09-06
Diag

### LINKED ISSUE #4137: diag

JamieLine:
Fixes #4137 and #4138


mattbarrett98 (changes requested):
In `test_torch_diag` you are manually setting `with_out` to False but `diag` supports the out argument so should be tested. 

My comments on your other PR about `@handle_cmd_line_args` also apply here, and the comment about `available_dtypes`- don't these functions support more than just floats?

JamieLine:
I think I've made all of the requested changes including the ones mentioned in the other PRs, and made a few other slight improvements. I've checked through it multiple times and I can't see any errors but I may well have simply missed something.

mattbarrett98 (changes requested):
I don't think it should be necessary to add the uint's to unsupported dtypes, because they aren't supported by torch at all and so won't be tested against. Also, your test for diagflat doesn't have a `@given`?

JamieLine:
Thanks for having another look, not adding a given to the test for `diagflat` really was a bit dim-witted, I think that might have been me missing something from a merge or something getting lost, I've added that in now, and I've removed the `uint`s from the unsupported dtypes, but it appears that the tests are ignoring the unsupported dtype flag here anyway, as they keep drawing `float16`. I've checked through the test in the debugger and that does appear to be whats happening, I've had a quick chat with some people in the internal chat, they've said to either specify the unsupported dtype in the backend (which I cannot do because one or more of these functions doesn't exist in the backend), or add an assume line to the frontend test, which I feel is more elegant than the "set-theoretic" way of defining the types without a helper function, by specifying the available types as the set-theoretic difference between 2 sets.

mattbarrett98 (changes requested):
Both of these tests are failing if you take a look at the frontend tests for torch.

JamieLine:
Once again GitHub has interesting ideas about when the PR should be closed and what the word merged means.

JamieLine:
I've fixed the test in the typical case, turns out there were issues caused by Jax and TF not supporting inplace updates, but the test is currently failing in the real Torch function with the following error

``` 2022-09-14T02:37:54.3771751Z >           frontend_ret = frontend_fw.__dict__[fn_tree](*args_frontend, **kwargs_frontend)
2022-09-14T02:37:54.3772012Z E           RuntimeError: [enforce fail at alloc_cpu.cpp:55] ((ptrdiff_t)nbytes) >= 0. alloc_cpu() seems to have been called with negative number: 18446744073709551615``` which I'll have a look at tomorrow.

mattbarrett98 (commented):
Sure 🙂 `diagflat` looks like a pretty inefficient python heavy implementation atm, is there no way it can be written using a composition of `ivy` functions? If not maybe it'd be worth making an extension suggestion for some function which would simplify this.

JamieLine:
I'll have a think about if there's a way to do it using ivy functions, but I expect it'll end with either an extension suggestion or some extremely strange bit of linear algebra that's gonna take some thinking to get to. In the meantime I can try to change the implementation to do an inplace update where possible. I've pushed a fix for the Torch crashing, its due to this bug / quirk of PyTorch. https://github.com/pytorch/pytorch/issues/71204

image

mattbarrett98 (changes requested):
This is just an issue with the test since you were testing against infeasible `offset` values. Instead of using `assume` I think you should use some kind of shared strategy so that only valid values are generated to begin with.

ivy-seed:

ivy-seed:

ivy-seed:

PR #4143, opened 2022-09-07
Added triu to Torch frontend and patched ivy triu

JamieLine:


JamieLine:
The formatting changes are from the Pre-Commit hooks, I can't make a commit that doesn't have them without one of the linters "fixing it". The change to the ivy functional API is there because it appears to be that everywhere else the `k` parameter is keyword only and therefore the tests crash without that change. I changed it there instead of making it positional elsewhere because I feel that this will cause the least disruption to the rest of the codebase.

JamieLine:
I've made the requested changes (unless I have missed something), and fixed a merge conflict with `cumprod`, I've had to tweak the test for `cumprod` to make it pass after the merge, that test intentionally does not use `helpers.get_dtypes("valid")` because that was not picking up the `unsupported_dtypes` flag, and I couldn't make a `helpers.get_dtypes("valid")` based solution work correctly, and it looked like it would produce something much less elegant than the solution in the original `cumprod` merge. Somewhere that original got changed to a `helpers.get_dtypes("float")` based solution, but that has the same issue and fixing it from there would require converting that returned tuple to some mutable collection, then converting it back and ultimately you end up where the original was anyway.

mattbarrett98 (changes requested):
Can you remove the changes to cumprod. I don't think it should be necessary to add the uint's to unsupported dtypes- does the test fail when you remove this? Also in the test, it should read `np.asarray(values, dtype=dtype)` to ensure the inputs are the sampled dtype.

JamieLine:
Changes have been made and merge conflict is gone.

mattbarrett98 (commented):
Thanks 🙂

PR #4167, opened 2022-09-07
Average function

OluwatimilehinF:


JamieLine:
Hi, thank you for your PR! I just have a few questions I'd like to ask as I review the code.

1. Is this code for one of the open tasks? If so which one?
2. Does this function run correctly on your machine? I've tried to run it and depending on the inputs I pass in I always get some form of error.

Thanks!

PR #4169, opened 2022-09-07
Numpy Correlate frontend

nassimberrada:
Pull request for issue #4065

### DIFF
--- a/ivy/functional/frontends/numpy/statistics/correlating.py
+++ b/ivy/functional/frontends/numpy/statistics/correlating.py
@@ -20,7 +20,7 @@ def correlate(
                 a = ivy.concat((ivy.array(0),a),axis = None)
         elif mode == "same":
             r = m
-            for j in range(0,(n-1)//2):
+            for j in range(0,(n-1)%2):
                 a = ivy.concat((ivy.array(0),a),axis = None)
         elif mode == "valid":
             r = m-n+1
@@ -34,7 +34,7 @@ def correlate(
                 v = ivy.concat((ivy.array(0),v),axis = None)
         elif mode == "same":
             r = m
-            for j in range(0,(n-1)//2):
+            for j in range(0,(n-1)%2):
                 v = ivy.concat((ivy.array(0),v),axis = None)
         elif mode == "valid":
             r = m-n+1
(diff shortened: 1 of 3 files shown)


roboFiddle:
Hi, I just pushed some changes to get you on the right track. There is no need for a `test_correlate` method because there is no `ivy.correlate` function - you only need to test the frontend function. It seems like the function is returning a boolean instead of a number. You can test the function in a shell by running `import ivy` and then `ivy.functional.frontends.numpy.correlate(ivy.array([1, 2, 3]), ivy.array([0, 1, 0.5]))` which should return `ivy.array([3.5])`. You can use the examples from the [Numpy documentation](https://numpy.org/doc/stable/reference/generated/numpy.correlate.html) to help you.

nassimberrada:
Thanks for the code clean-up, especially on the testing side where i wasn't sure about what to do !
I looked into the function returning booleans, i believe the issue is cleared and the function returns the appropriate output now.

nassimberrada:
Thanks for the code clean-up, especially with the testing part where i wasn't sure what to do !
I looked into the function returning booleans, i believe the issue is cleared now and the function returns the appropriate output.

roboFiddle:
Thanks for the changes. It seems like your implementation is not correct for all values. For example:
```
>>> x = ivy.array([-0.9090909090909091, -0.9090909090909091], dtype='float16')
>>> y = ivy.array([-0.9090909090909091, -0.9090909090909091], dtype='float16')
>>> ivy.functional.frontends.numpy.correlate(x, y, mode="same")
ivy.array([1.65, 1.65])
>>> np.correlate(x, y, mode="same")
array([0.8267, 1.653 ], dtype=float16)
```
If you run the test and look for the text "Falsifying example", you can see the test cases that fail for your function. Once that is fixed, I can take another look at this.

nassimberrada:
It should be fixed now and running the tests locally gives me a pass. Could you check if there is still an issue ?

roboFiddle:
Thanks for fixing the code. Would you be able to fix the linter errors, found [here](https://github.com/unifyai/ivy/runs/8300792674?check_suite_focus=true). You can also run the linter using the `flake8` command on each file to confirm that it is fixed. Afterwards, I'll be able to merge this.

nassimberrada:
Done, it should be all clear now. Thanks for your assistance !

nassimberrada:
Done, it should be clear now. Thanks for your assistance !

roboFiddle:
Looks good, merging. Thanks for the contribution!

PR #4197, opened 2022-09-08
Auto dtype tuple

Unevilicorn:
most of this is just removing the existing set tuples in the frontend,
important changes happen in `data_type`, `general` and `device`


ivy_tests/test_ivy/test_frontends/test_torch/test_loss_functions.py
```diff
@@ -138,6 +138,10 @@ def test_torch_binary_cross_entropy(
     true_dtype, true = dtype_and_true
     weight_dtype, weight = dtype_and_weight
 
+    assume(not (pred_dtype == "bfloat16"))
```
djl11, line 141: these can be reduced into a single line:
```
assume "bfloat16" not in (pred_dtype, true_dtype, weight_dtype)
```

djl11 (changes requested):
small change, otherwise looks good!

PR #4207, opened 2022-09-08
implemented eye in tensorflow linalg

ahmedo42:
#4205


ivy/functional/frontends/tensorflow/linalg.py
```diff
@@ -33,3 +33,13 @@ def pinv(a, rcond=None, validate_args=False, name=None):
...
+
+
+tensordot.supported_dtypes = ("float32", "float64")
+
+
+def eye(num_rows, num_columns=None, batch_shape=None, dtype=ivy.float32, name=None):
+    return ivy.eye(num_rows, num_columns, batch_shape=batch_shape, dtype=dtype)
+
+
+eye.unsupported_dtypes = ("float16", "bfloat16")
```
saeedashrraf, line 45: It seems to me that the function works fine for these two dtypes for some of the backend frameworks. You should specify the framework that does not support the dtypes in this case as follows `eye.unsupported_dtypes = {"torch": ("float16", "bfloat16")}
` .

saeedashrraf (changes requested):

saeedashrraf (approved):

PR #4212, opened 2022-09-08
initial changes

RickSanchezStoic:


RickSanchezStoic:
Changes for version-specific unsupported dtypes and devices. Have added the documentation too.

ivy/functional/backends/jax/__init__.py
```diff
@@ -16,6 +16,8 @@
 # local
 import ivy
 
+version = {"version": jax.__version__}
```
djl11, line 19: I think this can be `backend_version = jax.__version__` instead right? Then it becomes an attribute that we can simply query on the ivy namespace like `ivy.backend_version` directly. Could this update be made to work with the rest of the changes? What do you think?

djl11 (changes requested):
I haven't gone through in crazy detail, but in general it all looks good. Just one small question, Other than this small request, I'm happy for you to go ahead and merge!

ivy/functional/backends/jax/__init__.py
RickSanchezStoic: yeah this can be done

PR #4242, opened 2022-09-09
exceptions and assertions

juliagsy:
exceptions:
- added exception classes in `exceptions.py`
- added `handle_exceptions` decorator in `exceptions.py`
- updated `utility` and `statistical` submodules to use the decorator 
- configured import as `ivy.exceptions.<exception_class>` due to its specific purpose

assertions:
- refactored commonly used assertion into function in `assertions.py`
- added `check_elem_in_list` in `assertions.py`
- updated `gradients.py` to use the assertion helper function
- configured import as `ivy.assertions.<assert_func>` due to its specific purpose

if the changes for both exception and assertion handling looks good, I'll directly make the remaining changes after the merge of this PR! 

feel free to let me know if anything needs to be improved or changed, or the decorator sequence needs rearrangement, thanks!


djl11 (approved):
lgtm!

PR #4244, opened 2022-09-09
add Miscellaneous Operations to PyTorch Frontend #4164

Hellenn07:


mattbarrett98 (changes requested):
Your merging seems to have gone a bit wrong, which has resulted in `tan_for_frontends` and `master` all over your code, which as you can see in the tests is causing errors. 

Also, which function are you working on- the title says torch but you've edited tensorflow and jax functions?

Hellenn07:
Hi, 
Thank you for review. I was working on adding searchsorted operation on Pytorch. I had a look at Contribution Guide and working to correct my code to include pytorch only. Allow me a few more hours to work on the issue. Thank you.

PR #4321, opened 2022-09-11
added empty function torch

elhusseinn:
- [ ] #4301

### DIFF
--- a/ivy/functional/frontends/torch/creation_ops.py
+++ b/ivy/functional/frontends/torch/creation_ops.py
@@ -51,3 +51,20 @@ def zeros(size, *, out=None, dtype=None, device=None, requires_grad=False):
     if requires_grad:
         return ivy.variable(ret)
     return ret
+
+
+def empty(
+    size,
+    *,
+    out=None,
+    dtype=None,
+    layout=None,
+    device=None,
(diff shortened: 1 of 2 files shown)


Oxbeefbabe:
Hi, apologies for the slow response, I am currently away and have minimal access to the internet! I'll comment when possible, and I should be back to my regular schedule from next week.

This is a good start. Can you please make sure to implement the functionality for all of the optional parameters in the torch.empty function? Also, can you please make sure that testing for these variables is included in your tests?

elhusseinn:
Thanks for you reply, which params haven't I tested for, I think i tested for all params, can I get more clarification please?
thanks.

Oxbeefbabe:
You should implement the parameters for all of the parameters described in the pytorch docs for empty: https://pytorch.org/docs/stable/generated/torch.empty.html. The only two you do not need to implement are pin_memory and memory_format, as these do not change the input/output of the function.

elhusseinn:
The only parameter I didn't implement is the layout parameter, which is related to the memory , every other parameter is implemented and included in the tests.

Oxbeefbabe:
Ok, I've had a closer look and this seems all good to me then. Good job, merging your PR now!

PR #4322, opened 2022-09-11
add numpy.ndarray.transpose frontend

ddzhulaii:
transpose #4172


Oxbeefbabe:
Hi, apologies for the slow response, I am currently away and have minimal access to the internet! I'll comment when possible, and I should be back to my regular schedule from next week.

For the frontend functions, you shouldn't need to implement this much functionality yourself :). Instead, we want you to make use of the existing functions in the Ivy API, and call them to do most of the heavy lifting. You may need to add some basic functionality regarding some optional parameters if Ivy does not inherently support them. 

Also, you need to implement the test for your function. More information about the Frontend API and the Frontend API Tests can be found in the corresponding deep dives in the Ivy Documentation.

ddzhulaii:
Thanks for your comment! Initially, I was hesitant about changing the parts of Ivy API but now will do so. As well as adding tests :)

Oxbeefbabe:
You shouldn't need to change the Ivy API. The frontend function should "fill-in" any functionality that the function in the Ivy API doesn't handle while calling on the Ivy API function

ddzhulaii:
Okay, now I got it

ddzhulaii:
@Oxbeefbabe hey, was AFK cause I had to deal with covid for a few days 🙈 

- added a missing `ndarray.transpose` method
- added a test for `numpy.transpose` frontend method
- but a bit stuck with a test for the first one: originally `numpy.ndarray.transpose` supports a variable number of positional args. I could not find anything in the docs about the support of such inputs during testing. 

Thanks!

Oxbeefbabe:
You would pass the tuple/list of axes to the test function with the keyword argument as `axes` and then pass the number of positional args generated in the given decorator. See test_ndarray.py/test_numpy_instance_add to see how the number of positional args is generated

ddzhulaii:
Ready for the review :) 

- both tests for `numpy.transpose` and `numpy.ndarray.transpose` are fixed
- I changed the test helper `get_axis`: before if the `unique=True` it was returning unique axes by their dimensionality. So for example, if you input shape (1, 1, 2, 3) and set `sorted=True` then you would get `(0, 2, 3)` (cause `shape[0] == shape[1]`) as an output but not `(0, 1, 2, 3)` which I believe is the expected output (correct me if it's wrong 🙈 ).

ivy_tests/test_ivy/test_frontends/test_numpy/test_manipulation_routines/test_transpose_like_operations.py
```diff
@@ -0,0 +1,41 @@
...
+def test_numpy_transpose(
+    array_and_axes,
+    as_variable,
+    num_positional_args,
+    native_array,
+    fw,
+):
+    array, axes = array_and_axes
+    helpers.test_frontend_function(
+        input_dtypes=["int8"],
```
Unevilicorn, line 31: It would be better if we could generate the dtypes dynamically along with the array too, but considering the function, this is most likely fine too

Unevilicorn (commented):
Looks good, the tests is all passing, I'll get this merged once you solve the merge conflicts.

ivy_tests/test_ivy/test_frontends/test_numpy/test_manipulation_routines/test_transpose_like_operations.py
ddzhulaii: Thanks! Added a dynamic one :)

ddzhulaii:
Thanks for a quick check! Resolved all the conflicts :)

Unevilicorn:
All merged :slightly_smiling_face:

PR #4358, opened 2022-09-12
fix `__getitem__` in array class.

CatB1t:


ivy/functional/backends/tensorflow/general.py
```diff
@@ -57,6 +57,21 @@ def current_backend_str():
     return "tensorflow"
 
 
+def get_item(x: tf.Tensor, query: tf.Tensor) -> tf.Tensor:
```
djl11, line 60: I think we discussed this before, but are you sure there is no elegant way to wrap `x.__getitem__(query)` in this case? I presume this might be more efficient than `tf.gather(x, query)` in cases where the former would work, but if the conditions to verify it works are complex then I think the approach you currently have is good.

djl11 (changes requested):
one small comment, otherwise it's good to merge!

Issue #4183, opened 2022-09-08
concatenate


optimusprime009:
#4183

optimusprime009:
hello, can anyone open the pull request? It is automatically closed by ivy seed.

optimusprime009:
#7828

optimusprime009:
https://github.com/unifyai/ivy/pull/8813

optimusprime009:
#8813

PR #4361, opened 2022-09-12
concatenate #4183

optimusprime009:


JamieLine:
Hi,

Thanks for your PR! I've just had a quick run through and I have a few initial concerns / questions.

1.) The function `layers_concatenate` is not featured in this PR, has it been deleted intentionally?
2.) Which Tensorflow function does this add to the frontend? I've only had a brief look so sorry if I'm just being a bit dim but I cannot find a function called `layers_concatenate` in the Tensorflow docs. The function signature of our frontend should match the signature of the original Tensorflow function, so if this function is a frontend implementation of `tf.keras.layers.concatenate` this function should be callable from `ivy.functional.frontends.tensorflow.keras.layers.concatenate`. This doesn't strictly mean that the file structure needs to be changed, I believe there is a trick to do it without up-heaving the file structure.
3.) The merge conflict that GitHub warns about is because the file structure of the Tensorflow frontend has changed recently, if you move your function into `layers.py` in the same directory that should help to resolve it.

If you have any questions or comments feel free to let me know!

optimusprime009:
Thank you for your feedback, 
1)No, I had not deleted the PR somehow when I committed the pull request it was there.
2)Yes layer_concatenate is tf.Keras.layers.concatenate and I will change that as the format of the function formatting in the file.
Thanks for your help.

Daniel4078:
This pull request is sort of abondanded and is continuing in #4648

JamieLine:
Being continued in #4648

PR #4364, opened 2022-09-12
Add rad2deg function to numpy frontend

### LINKED ISSUE #4226: rad2deg

JonaOmara:
Close #4226

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_trigonometric_functions.py
+++ b/ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_trigonometric_functions.py
@@ -469,3 +469,55 @@ def test_numpy_radians(
         subok=True,
         test_values=False,
     )
+
+    
+# rad2deg
+@handle_cmd_line_args
+@given(
+    dtype_and_x=helpers.dtype_and_values(available_dtypes=helpers.get_dtypes("float")),
+    dtype=helpers.get_dtypes("float", full=False, none=True),
+    where=np_frontend_helpers.where(),
+    as_variable=helpers.array_bools(num_arrays=1),
+    with_out=st.booleans(),
+    num_positional_args=helpers.num_positional_args(
+        fn_name="ivy.functional.frontends.numpy.rad2deg"
+    ),
+    native_array=helpers.array_bools(num_arrays=1),
+)
+def test_numpy_rad2deg(
+    dtype_and_x,
+    dtype,
+    where,
+    as_variable,
+    with_out,
+    num_positional_args,
+    native_array,
+    fw,
+):
+    input_dtype, x = dtype_and_x
+    input_dtype = [input_dtype]
+    where = np_frontend_helpers.handle_where_and_array_bools(
+        where=where,
+        input_dtype=input_dtype,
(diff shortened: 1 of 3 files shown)


ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_trigonometric_functions.py
```diff
@@ -469,3 +469,55 @@ def test_numpy_radians(
...
+        frontend="numpy",
+        fn_tree="rad2deg",
+        x=np.asarray(x, dtype=input_dtype[0]),
+        out=None,
+        where=where,
+        casting="same_kind",
+        order="K",
+        dtype=dtype,
+        subok=True,
+        test_values=False,
```
roboFiddle, line 521: You should be testing values, set this to true

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_trigonometric_functions.py
```diff
@@ -469,3 +469,55 @@ def test_numpy_radians(
...
+    fw,
+):
+    input_dtype, x = dtype_and_x
+    input_dtype = [input_dtype]
+    where = np_frontend_helpers.handle_where_and_array_bools(
+        where=where,
+        input_dtype=input_dtype,
+        as_variable=as_variable,
+        native_array=native_array,
+    )
```
roboFiddle, line 504: You don't need to call `handle_where_and_array_bools`. Take out lines 498-504 and just use `input_dtype` later.

roboFiddle (changes requested):
I added a few comments, otherwise it looks good.

roboFiddle:
It seems that your function does not match the NumPy function. For example, passing `-0.9090909090909091` returns `-52.08345413208008` from your function, while the NumPy rad2deg function returns `-52.09375`. As a guess, this probably is caused by the non-associativity of floating-point operations. Perhaps you can try first dividing by Pi and then multiplying by 180, or just multiplying by a factor of `180/pi` to do this in one operation. Secondly, you should make sure to set `out=out` in all Ivy function calls.

Lastly, you can check both these points by running the test using `python3 -m pytest --num-examples=500 -k test_numpy_rad2deg ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_trigonometric_functions.py`

roboFiddle (changes requested):
See above comment.

JonaOmara:
@roboFiddle , when I run the test `python3 -m pytest --num-examples=500 -k test_numpy_rad2deg ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_np_trigonometric_functions.py` I get this error `/usr/bin/python3: No module named pytest`, I think there is some setting that I didn't put correct, but I have only been running the tests with the Run button but I would also wish to run the tests in the terminal, what do I need to set right.

However I've divided `x/pi` and multiplied with `180 ` and I've also added `out=out` for all the Ivy functions, I hope this sorts the issue.

roboFiddle:
Don't worry about it, if you're able to run the tests it's fine. This looks good, merging. Thanks for the contribution!

PR #4370, opened 2022-09-12
cholesky

### LINKED ISSUE #4366: cholesky

karalleyna:
Closes #4366

### DIFF
--- a/ivy/functional/frontends/jax/lax/linalg.py
+++ b/ivy/functional/frontends/jax/lax/linalg.py
@@ -1,0 +1,5 @@
+import ivy
+
+
+def cholesky(a):
+    return ivy.cholesky(a)
--- a/ivy/functional/frontends/jax/lax/linalg.py
+++ b/ivy/functional/frontends/jax/lax/linalg.py
@@ -1,5 +1,12 @@
 import ivy
 
 
-def cholesky(a):
-    return ivy.cholesky(a)
+def cholesky(x, /, *, symmetrize_input=True):
+    def symmetrize(x):
+        # TODO : Take Hermitian transpose after complex numbers added
+        return (x + ivy.swapaxes(x, -1, -2)) / 2
+
+    if symmetrize_input:
+        x = symmetrize(x)
+
+    return ivy.cholesky(x)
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/jax/lax/linalg.py
```diff
@@ -0,0 +1,12 @@
+import ivy
+
+
+def cholesky(x, /, *, symmetrize_input=True):
```
ahmedo42, line 4: Since you're implementing a frontend function, the signature and behavior should be the same as JAX, I think the previous commit was fine and was ready to merge since it was passing the unit tests.

ahmedo42 (changes requested):

ahmedo42:
Hi @karalleyna, Thanks for the contribution!!

Ivy currently doesn't support complex data types, it would be better to handle that in a separate PR as it's considered an extension to the Ivy API

ivy/functional/frontends/jax/lax/linalg.py
karalleyna: Previous implementation was based on [jax.numpy.linalg.cholesky](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.cholesky.html) instead of [jax.lax.linalg.cholesky](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.linalg.cholesky.html#jax.lax.linalg.cholesky). So, if it is okay for you, I can revert to the old one.

ahmedo42: sorry for the confusion, yes we should be using the lax implementation! you're on the right track

karalleyna:
> Hi @karalleyna, Thanks for the contribution!!
> 
> Ivy currently doesn't support complex data types, it would be better to handle that in a separate PR as it's considered an extension to the Ivy API



> Hi @karalleyna, Thanks for the contribution!!
> 
> Ivy currently doesn't support complex data types, it would be better to handle that in a separate PR as it's considered an extension to the Ivy API

I have implemented complex dtypes in #4363 . However, it needs a bit of collaboration because the current functions in Ivy do not take complex dtypes into account for now. Hence, some of tests have failed. One temporary solution may be restricting dtypes for tests to float or integer but it is better to take your advice.

ahmedo42:
@karalleyna feel free to ignore the complex dtypes for now since we don't support them yet, restricting the tests to float is fine
once the functions' unit tests pass we can merge :)

ahmedo42:
LGTM @karalleyna!

PR #4373, opened 2022-09-12
Reformat linear_algebra trace function

SudhanvaMG:


sreegandh804:
Hey @SudhanvaMG  , I would suggest you look through the contributor's guide on how you name your issues, as you are also required to make a comment with just the issue number whose title should be just the function name, as this would automatically assign you that issue and update the main issue 🙂 https://lets-unify.ai/ivy/contributing.html

PR #4386, opened 2022-09-13
Root directory reorganisation

simonetgordon:
Issue with Dockerfile not recognising requirements.txt in the new `requirements` directory:
```

ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements/requirements.txt'
```


Dockerfile
```diff
@@ -18,21 +18,21 @@ RUN pip3 install --upgrade torch-scatter -f https://pytorch-geometric.com/whl/to
...
+    cat requirements/requirements.txt | grep -v "ivy-" | pip3 install --no-cache-dir -r /dev/stdin && \
+    cat requirements/optional.txt | grep -v "ivy-" | pip3 install --no-cache-dir -r /dev/stdin && \
     python3 -m pip install --user -e . && \
     cd ivy_tests/test_array_api && \
     pip3 install --no-cache-dir -r requirements.txt
 
 # Install local requirements
-COPY requirements.txt /
-RUN pip3 install --no-cache-dir -r requirements.txt
+COPY requirements/requirements.txt /
```
simonetgordon, line 28: Doesn't recognise requirements.txt here.

djl11 (changes requested):
Just ran it locally, for me the line `COPY requirements/requirements.txt /` runs without error. It's the line `RUN pip3 install --no-cache-dir -r requirements/requirements.txt` afterwards which is the problem. This is because `COPY requirements/requirements.txt /` only copies the file into `/`, not the directory structure. Therefore, it should be changed to `RUN pip3 install --no-cache-dir -r requirements.txt`. Likewise, the `git clone` command will also fail whilst the upstream repo does not contain the updated directory structure. Does this make sense?

PR #4409, opened 2022-09-13
add hardsigmoid to Pytorch frontend

### LINKED ISSUE #4263: hardsigmoid

y-h-chan:
Closes #4263


JamieLine:
The test looks mostly correct but it fails on my machine with `TypeError: hardsigmoid() got an unexpected keyword argument 'threshold'`, this is because the end of the call to `helpers.test_frontend_function` is wrong, every keyword argument after `fn_tree` in this call will be passed to the function that's being tested. Also you've specified `inplace` as always being false, this should be generated in the `@given` decorator, using `st.booleans()`

y-h-chan:
Should i merge after the workflow or i should keep it up to date?

y-h-chan:
May i get a pointer on proper implementation of the test method? The test in pytest failed but I believe the implementation is correct as the results are correct on manual testing on both the ivy frontend and the torch backend.

JamieLine:
Hi, what is the error you get in pytest? It looks alright at a first glance, I'll pull this version and see if it works for me

y-h-chan:
i guess its the inplace behaviour of the function changing the input. So im not sure if the helper functions are intended to test with inplace.

JamieLine:
Having got it running that makes sense, I had a test failure where the values disagreed and one of them was correctly `hardsigmoid(x)` and one was incorrectly `hardsigmoid(hardsigmoid(x))`, I'll quickly see how this is usually avoided.

y-h-chan:
Other tests simply ignore testing the inplace but this seems weird.

JamieLine:
It looks like the other functions just pass `False` in, so I'd say just do that and add a comment that says it was validated to work outside of the testing framework, but causes errors in the testing framework.

Also very small thing, but when you update to `inplace=False`, the spaces should be removed around the `=` character, it's a minor thing but its the standard style in the codebase to remove the spaces when assigning keyword arguments.

JamieLine:
Having had a quick once-over, try running `flake8` and `black` on the files you're working on, as is typically done by `pre-commit`. I think there's at least one line that goes over the typical maximum length (which I believe is 88 characters).

y-h-chan:
its the comment, sorry let me quickly fix it.

JamieLine:
Thanks for fixing it, I fixed a merge conflict against master and re-ran the formatting tools, so if you see my commit in your branch there's nothing to be alarmed about, and it doesn't change that this is your commit and you still receive full value out of it, I just felt it was unfair for me to introduce a formatting issue and ask you to fix it.

PR #4411, opened 2022-09-13
torch already handles type promotion as we want

mattbarrett98:


ivy-seed:

PR #4419, opened 2022-09-13
add `safety_factor` to `array_values`, refactor, some bugs fix.

CatB1t:


ivy_tests/test_ivy/helpers.py
```diff
@@ -2373,9 +2373,9 @@ def dtype_and_values(
     num_arrays=1,
     min_value=None,
     max_value=None,
-    large_value_safety_factor=1.1,
-    small_value_safety_factor=1.1,
-    max_op="divide",
+    large_abs_safety_factor=1,
```
djl11, line 2376: I would suggest we keep the default value above `1`, otherwise many tests start to fail I believe, with values being generated right at the representable limit. Did you observe different behavior when running the tests? Do they all seem to work okay with a value of `1`?

djl11 (changes requested):
one very small comment, otherwise it looks good. I haven't had the time to go through in crazy detail, but in general it all seems correct. Provided you are not introducing a load of new tests failures, and you think the implementations are all bug free, then feel free to merge after addressing the small comment. Thanks! :)

PR #4424, opened 2022-09-13
Torch arr api fixes: type promotion and test_signatures

ZoeCD:


ZoeCD:
Hi! I'm going to this PR because I think that it's better to make a PR for each test group. This will help me to quickly keep track of new failures that come up

Issue #3856, opened 2022-08-31
Extend Ivy Functional API


johanjino:
Extend Ivy functional API [ivy//ivy/functional/experimental](https://github.com/unifyai/ivy/blob/master/ivy/functional/experimental/). Each function will need implementation in each [backend API](https://lets-unify.ai/ivy/deep_dive/0_navigating_the_code.html#backend-api) and also the [Ivy API](https://lets-unify.ai/ivy/deep_dive/0_navigating_the_code.html#ivy-api). The [tests](https://lets-unify.ai/ivy/deep_dive/15_ivy_tests.html) will also need to be implemented. See the [Deep Dive](https://lets-unify.ai/ivy/deep_dive.html) for more general information on how the functions should be implemented. Particular attention should be placed on the [Superset Behaviour](https://lets-unify.ai/ivy/deep_dive/2_superset_behaviour.html) section when considering how best to design the function, and what features and arguments it should have. More information about this specific task is provided in the [Open Tasks](https://lets-unify.ai/ivy/contributing/4_open_tasks.html#ivy-experimental-api) section of the contributor guide. All the implementations have to be placed in `experimental` directories for both, functional API, backend APIs, Array, and Container instances.

`Linear algebra functions:` [ivy//ivy/functional/experimental/linalg.py](https://github.com/unifyai/ivy/blob/master/ivy/functional/experimental/linalg.py)

- [x] #5991
- [x] #6180
- [ ] #6269
- [x] #6228
- [x] #7748
- [x] #8990
- [ ] eigh_tridiagonal
- [x] #8806
- [ ] solve_triangular
- [ ] #9579
- [ ] #9325

`sorting functions:` [ivy//ivy/functional/experimental/sorting.py](https://github.com/unifyai/ivy/blob/master/ivy/functional/experimental/sorting.py)

- [x] #6138

`Manipulation functions:` [ivy//ivy/functional/experimental/manipulation.py](https://github.com/unifyai/ivy/blob/master/ivy/functional/experimental/manipulation.py)

- [x] #5542
- [ ] #5944
- [x] #5787
- [x] #5614
- [x] #5683
- [x] #5977
- [x] #6874
- [x] #5978
- [x] #6079
- [x] #7967
- [x] #6787
- [x] #6334
- [x] #6340
- [ ] frombuffer
- [x] #7222
- [x] #6873
- [x] #6222
- [x] #6734
- [x] #6728
- [x] #6761
- [x] #6883
- [x] #6886
[...]

Issue #3612, opened 2022-08-25
Add Instance Methods to PyTorch Frontend


jkeane508:
Add [Instance Methods](https://pytorch.org/docs/stable/tensors.html#tensor-class-reference) to PyTorch Frontend:
\_

>Please keep in mind that the proper way to link an issue to this list is to comment "- [ ] #issue\_number" while the issue's title only includes the name of the function you've chosen.

\_

- [x] new\_tensor
- [x] new\_full
- [x] new\_empty
- [x] #5265
- [x] new\_zeros
- [x] #6639
- [x] #16508
- [x] #19590
- [ ] device
- [x] grad
- [ ] ndim
- [x] #8882
- [x] #14411
- [x] #5804
- [x] #5818
- [x] #5819
- [x] #5820
- [x] #6011
- [x] acos\_
- [x] arccos
- [x] arccos\_
- [x] add
- [x] add\_
- [x] #14645
- [x] #14646
- [x] addcdiv
- [x] #17195
- [x] #14983
- [x] #14984
- [x] #14985
- [x] #14986
- [ ] sspaddmm
- [x] #17485
- [x] #20959
- [x] #15741
- [x] #15859
- [x] #19674
- [ ] close
- [x] #5789
- [x] #5889
- [x] #9946
- [x] #19673
- [x] #18185
- [x] #6556
- [x] #7574
- [x] #9267
- [x] #7859
- [x] #5812
- [x] asin\_
- [x] #5814
- [x] arcsin\_
- [x] #16267
- [x] atan
- [x] atan\_
- [x] #6020
- [x] #6553
- [x] #7898
- [x] atan2\_
- [x] #10227
- [x] #10493
- [x] #12996
- [x] #9438
- [x] #17001
- [x] #20970
- [x] #19438
- [ ] #22976
- [ ] bfloat16
- [x] #13643
- [x] bitwise\_not
- [x] #17484
- [x] #6693
- [x] bitwise\_and\_
- [x] #9848
- [x] #12333
- [x] #13007
- [x] #21659
- [x] bitwise\_left\_shift
- [x] #21770\_
- [x] #14845
- [x] #23768
- [x] #20116
- [x] bool
- [x] byte
- [ ] broadcast\_to
- [ ] #26790
- [x] #7081
- [x] ceil\_
- [x] #21621
- [x] #17922
- [ ] #26402
- [ ] cholesky\_solve
- [x] chunk
- [x] #7751
- [x] clamp\_
- [x] #13639
[...]

Issue #12, opened 2022-01-19
Add Pointwise ops to PyTorch Frontend


djl11:
Add [Pointwise ops](https://pytorch.org/docs/stable/torch.html#pointwise-ops) to PyTorch frontend:

- [ ] abs
- [ ] absolute
- [ ] acos
- [ ] arccos
- [ ] acosh
- [ ] arccosh
- [ ] add
- [ ] addcdiv
- [ ] addcmul
- [ ] angle
- [ ] asin
- [ ] arcsin
- [ ] asinh
- [ ] arcsinh
- [ ] atan
- [ ] arctan
- [ ] atanh
- [ ] arctanh
- [ ] atan2
- [ ] bitwise_not
- [ ] bitwise_and
- [ ] bitwise_or
- [ ] bitwise_xor
- [ ] bitwise_left_shift
- [ ] bitwise_right_shift
- [ ] ceil
- [ ] clamp
- [ ] clip
- [ ] conj_physical
- [ ] copysign
- [ ] cos
- [ ] cosh
- [ ] deg2rad
- [ ] div
- [ ] divide
- [ ] digamma
- [ ] erf
- [ ] erfc
- [ ] erfinv
- [ ] exp
- [ ] exp2
- [ ] expm1
- [ ] fake_quantize_per_channel_affine
- [ ] fake_quantize_per_tensor_affine
- [ ] fix
- [ ] float_power
- [ ] floor
- [ ] floor_divide
- [ ] fmod
- [ ] frac
- [ ] frexp
- [ ] gradient
- [ ] imag
- [ ] ldexp
- [ ] lerp
- [ ] lgamma
- [ ] log
- [ ] log10
- [ ] log1p
- [ ] log2
- [ ] logaddexp
- [ ] logaddexp2
- [ ] logical_and
- [ ] logical_not
- [ ] logical_or
- [ ] logical_xor
- [ ] logit
- [ ] hypot
- [ ] i0
- [ ] igamma
- [ ] igammac
- [ ] mul
- [ ] multiply
- [ ] mvlgamma
- [ ] nan_to_num
- [ ] neg
- [ ] negative
- [ ] nextafter
- [ ] polygamma
- [ ] positive
- [ ] pow
- [ ] quantized_batch_norm
- [ ] quantized_max_pool1d
- [ ] quantized_max_pool2d
- [ ] rad2deg
- [ ] real
- [ ] reciprocal
- [ ] remainder
- [ ] round
- [ ] rsqrt
[...]

PR #4436, opened 2022-09-14
Implemented sinc in the backend

pj-mathematician:
Can you review and please list the additions/deletions I should incorporate?
And I will also need help with the doc string for ivy/elementwise.py

Thanks


Unevilicorn:
Hi, 

Could you also link the issue this relates to. This is so that it's easier for us to track what's being done.
The test is failing for sinc, you can see by going through the `checks` near the bottom of this page, click on detail on the row for `test-core-ivy / run-nightly-tests ({backend}, elementwise)` then click on `Run Functional-Core Test`

Thanks

pj-mathematician:
@Unevilicorn thank you for the review, I will fix the tests by EOD. 

sinc is a part of the functional API and is referenced in the Frontend TODO lists #12 , #1525 and #3612 
But when I started the Frontend sinc I noticed that it isn't present in the backend at all. Hence I wrote the backend implementations. I believe it should be added to the list here #3856 . I have opened a separate issue to track this here #4447

Unevilicorn:
I would actually recommend doing this compositionally then, since this function isn't too complex to compose, you can see it's behaviour from [pytorch docs](https://pytorch.org/docs/stable/special.html#torch.special.sinc).

I'll take another look, and check if this should be added to the list :+1:

pj-mathematician:
I understand. So instead of adding any backend functions, I should implement sinc compositionally to only the frontend files of Numpy and Torch, as mentioned in the frontend todo tasks, right?

Unevilicorn:
Yeah, that's the simplest way to go about it, just implementing them compositionally in the frontend for your framework. However it does feel like it could be added as an extension, I'll let you know when I find out if this should be added or not.

pj-mathematician:
Alright then! Closing this PR for now, and will work on the frontend as you suggested. Let me know if there is a need for it in the backend, I'd love to get back on it 🙌🏻

pj-mathematician:
@Unevilicorn , I have opened #4462 to work on this. Thank you for the clarification

PR #4445, opened 2022-09-14
Hellenn07 searchsorted #4164

Hellenn07:


JamieLine:
Hi, 

Thanks for your PR! Just a few quick things,

1. Which function is this PR meant to be implementing? In the changed files there are multiple functions being changed, each PR should only be implementing one function.
2. The implementation for `search_sorted` has a lot more arguments than it passes into the Ivy version, does this version replicate the entirety of the behaviour of the Torch version? E.g. If I pass in `right=True` to the version in this PR, will it have the same effect as passing `right=True` into the Torch version of this function?
3. At some point you've done a merge that affected your version of `ivy/functional/frontends/torch/miscellaneous_ops.py` which has removed the implementation of a function (`triu`) and added the names of the merged branches to the contents of the file, those names will need to be removed and the function implementation will need to be restored from the `unifyai/ivy` repository.
4. The changes to `ivy_tests/test_ivy/helpers.py` are almost definitely too long, especially given the only change that I could see was the branch names being added as they were in `ivy/functional/frontends/torch/miscellaneous_ops.py`, these changes should probably be reverted.


Hellenn07:
Thank you for your comments. The task is to add searchsorted miscellaneous operation to Pytorch. I'm I in the right track? I'm new here and trying to follow the Contributing Guideline. 
I would appreciate your assistance.

JamieLine:
You're on roughly the right track, you've just got some extra bloat in this commit. To clear that up here's what is needed.

1. Remove the changes to everything in the other frameworks (look into using `git rm` for this, you can keep those changes on your local system / in another branch on your fork of the repository, but this PR needs to just include the changes specific to `searchsorted` in `torch`. You can always push somethig onto your fork and check the changes in the Files Changed tab on this page.

2. Add a test to the appropriate subfolder of `ivy_tests` to test your `searchsorted` implementation, it should look broadly similar to the tests for the other functions in that folder, just make sure that you do the following,

- Use `helpers.dtype_and_values(X)` where X is one of `"valid"`, `"numeric"`, `"float"`, or something similar, if you check the implementation for that function it will be clear with other types it accepts. You want the most general one that can pass the tests, so ideally `"valid"` then `"numeric"` then the others.
- Test against every argument of both your function and the Torch function.
- This test should eventually pass every time, make sure you run the test multiple times because the testing framework can sometimes take a few iterations to find edge-cases. But don't worry if it takes a while and a bit of effort to make it pass at all, sometimes it takes a little bit.

Once again if you have any questions feel free to let me know and I'll see how I can help.

Issue #4457, opened 2022-09-14
Positional/Keyword arguments errors in Ivy API (ivy/functional/ivy/layers.py)


ferrarodav:
When executing ivy `conv*d` and `conv*d_transpose` layers an "Too many positional arguments" exception is raised (tested with the latest docker image).

Example:
```python
from ivy.functional import *
i = ivy.zeros((2, 6, 6, 1)) # (batch_size, ..., channels)
w = ivy.zeros((3, 3, 1, 5)) # (..., in_channels, out_channels)
print(conv2d(i, w, 1, "VALID").shape)
```
Expected output: `(2, 4, 4, 5)`
Actual output: `TypeError: conv2d() takes 4 positional arguments but 6 positional arguments (and 1 keyword-only argument) were given`

PR #4458, opened 2022-09-14
Fixed Ivy API conv functions positional/keyword arguments error

ferrarodav:
Issue #4457


MarShaikh:
Sorry, this completely slipped out of my radar! Thanks for these pushes. However, we have since added these changes already, so it wouldn't make sense for me to accept your push. Great spot though! I'll be closing this PR, but have marked your issue as completed.

PR #4518, opened 2022-09-15
Fix for bug where frontend tests ignore unsupported_dtypes flag

JamieLine:
I've had a quick chat with Ved about this bug, this fixes the issue where it's ignored in the frontend tests. I've added in the `trace` test from another PR just to reproduce the error more easily, I'll remove that before anything gets merged. By running the test you can see (if you get good enough luck for `float16` to be drawn) that the value is nicely skipped.

I'm wondering if there's any reason to keep the `if "backend" not in fn.__module__ and "frontend" not in fn.__module__:` line and associated block at all, in my very quick testing nothing seems to actually meet this criteria. 

Also how should I go about testing that this doesn't break anything, I'm going to run all of the frontend tests and see if they agree with the master branch but I'm wondering if this is sufficient.


JamieLine:
Already fixed.

Issue #4531, opened 2022-09-16
Minor error in `Shared Nested Structure` example in Deep Dive - Containers


ash-ishh:
https://lets-unify.ai/ivy/deep_dive/5_containers.html

```
x = ivy.Container(a={'b': 2, 'c': 4}, d={'e': 6, 'f': 8})
y = ivy.Container(a=2, d=3)
```

```
print(x / y)
{
    a: {
      b: 1,
      c: 2
    },
    d: {
      e: 3,
      f: 2.67
    }
}
```
`d.e`  should be 2 instead of 3 (x.d.e /y.d  -> 6/3)

PR #4532, opened 2022-09-16
Container deep dive documenration's shared nested example fixed

ash-ishh:
PR for issue #4531

Along with `d.e ` value fix

1.  Output container print values converted to float instead of integer (which is inline with current ivy execution of example) 
2.  f parameter updated from 8 to 9 since 8/3 was returning 2.6666666666666665, now it will return 3.0 which is more clean.

### DIFF
--- a/docs/partial_source/deep_dive/6_containers.rst
+++ b/docs/partial_source/deep_dive/6_containers.rst
@@ -294,7 +294,7 @@ are shared but not identical.
 
 .. code-block:: python
 
-    x = ivy.Container(a={'b': 2, 'c': 4}, d={'e': 6, 'f': 8})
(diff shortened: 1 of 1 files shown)


hirwa-nshuti:
LGTM!
Hey, @ash-ishh I'll link this PR to its corresponding issue but next time you can use `Close #Issue_number` to link your PR to its issue.
Thanks for contributing 😊

hirwa-nshuti (approved):

ash-ishh:
Hi @hirwa-nshuti Thanks!

Quick question for future tracking, Does it strictly need to be separate comment with `Close #Issue Number`?
I mentioned the issue in Description and it seems to have successfully linked it correctly as shown in below screenshot.

image

hirwa-nshuti:
When opening a PR you need to add a comment `Close #issue_number` for the issue to be linked with that PR

PR #4544, opened 2022-09-16
frontend dtype supports

Unevilicorn:


Unevilicorn:
enables the frontend to add unsupported_dtypes

PR #4547, opened 2022-09-16
creating exception subclasses

simonetgordon:
To handle array-api tests which check if a particular exception is raised given certain 'incorrect' inputs.


simonetgordon:
@juliagsy Not sure if this is what you had in mind, but it has fixed the array-api test failures :)

ivy/exceptions.py
```diff
@@ -24,6 +24,28 @@ def __init__(self, message=""):
         super().__init__(message)
 
 
+class IvyIndex(IndexError, IvyBackendException):
```
juliagsy, line 27: can we rename this to `IvyError`, then make this inheriting from `IndexError`, `ValueError` and `IvyException` (`IvyBackendException` is not needed), so that it's more unifying and extendable if there is similar case for another class

ivy/exceptions.py
```diff
@@ -24,6 +24,28 @@ def __init__(self, message=""):
...
+        self._default = [
+            "numpy" if ivy.current_backend_str() == "" else ivy.current_backend_str()
+        ]
+        self._delimiter = ": "
+        for message in messages:
+            self._default.append(message)
+        super().__init__(self._delimiter.join(self._default))
+
+
+class IvyValue(ValueError, IvyBackendException):
```
juliagsy, line 38: this will not be needed if we update the above class

ivy/exceptions.py
```diff
@@ -44,6 +66,10 @@ def new_fn(*args, **kwargs):
         """
         try:
             return fn(*args, **kwargs)
+        except IndexError as i:
```
juliagsy, line 69: we can unify it here as `except (IndexError, ValueError) as e: ...`

ivy/exceptions.py
```diff
@@ -44,6 +66,10 @@ def new_fn(*args, **kwargs):
...
+            raise ivy.exceptions.IvyIndex(fn.__name__, str(i)) from None
```
juliagsy, line 70: then it would be `ivy.exceptions.IvyError` here

juliagsy (changes requested):
Hey! Thanks for the updates! They look great! I had a quick rethink and suggested a few changes to simplify the classes, let me know if there's any issue or question! Thanks a lot!

PR #4548, opened 2022-09-16
divide_superset

Ishticode:
Based on `rounding_mode` arg in [`torch.divide`](https://pytorch.org/docs/stable/generated/torch.div.html#torch.div)


djl11 (changes requested):
given that we already have `ivy.floor_divide`, and given that most frameworks separate these into different functions, I would suggest we do not add this extra argument, as it would cause overlap and redundancy with `ivy.floor_divide` However, I would suggest we add a new function `ivy.trunc_divide` which encapsulates the C-style integer division instead :)

Ishticode:
> given that both `divide` and `floor_divide` must exist for adherence to the Array API Standard, I would suggest we do not add this extra argument, as it would cause overlap and redundancy with our `ivy.floor_divide` function. However, I would suggest we add a new function `ivy.trunc_divide` which encapsulates the C-style integer division instead :).

Sure 🙂.

PR #4550, opened 2022-09-16
Frontends/torch/std

### LINKED ISSUE #4513: std

mexyusef:
Closes #4513


ivy/functional/frontends/torch/reduction_ops.py
```diff
@@ -35,3 +35,7 @@ def any(input, dim=None, keepdim=False, *, out=None):
 
 def mean(input, dim, keepdim=False, *, out=None):
     return ivy.mean(input, axis=dim, keepdims=keepdim, out=out)
+
+
+def std(input, dim, *, correction=0.0, keepdim=False, out=None):
```
juliagsy, line 40: I'm not sure if I missed something, but according to the docs I found, the native parameters are `unbiased` instead of `correction`

https://pytorch.org/docs/stable/generated/torch.std.html

juliagsy (changes requested):
Hey there! Great work so far! Just one small change and you're good to go! Please request my review again when you're done! Thank you!

ivy/functional/frontends/torch/reduction_ops.py
mexyusef: I initially used `unbiased ` according to the Pytorch docs. However the functional API `ivy.std` which that function calls has a different keyword parameter, `correction` instead of `unbiased` , as shown here:
https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/statistical.py#L453
If I use `unbiased` it will complain `ivy.exceptions.IvyBackendException: std() got an unexpected keyword argument 'unbiased'`, otherwise the test will pass.

juliagsy: I see, we should actually follow the parameters natively for the frontend functions (unless it's a mistake in the torch docs cuz there was once where the torch native function says `input` in the docs but `tensor` in the actual implementation)

and in the function body, you can do `ivy.std(..., correction=unbiased,..)`, we can match the argument ourselves when linking it to the backend

let me know if there are further issue or question!

juliagsy: or something like:

```
if unbiased:
    ivy.std(...,correction=1,...)
else:
    ivy.std(...,correction=0,...)
```

mexyusef: > I see, we should actually follow the parameters natively for the frontend functions

Ok. That makes sense. I was a bit confused which specification to follow.
Thank you.

juliagsy: You're welcome!

juliagsy:
Thanks for contributing!

Issue #4586, opened 2022-09-17
transpose


martinoywa:

elusenji:
Hi @martinoywa, if any of the issues in the task list accept contributions from the open source community, I'd love to jump in. Please advise.

martinoywa:
Hi @elusenji, they do. Please check the contributions docs, you'll be able to find out more. Here's the link.
https://lets-unify.ai/ivy/contributing.html

PR #4592, opened 2022-09-18
add transpose method to numpy frontend

martinoywa:
This PR addresses Issue #4586


Nightcrab (changes requested):
Please double-check the documentation for `numpy.transpose` at https://numpy.org/doc/stable/reference/generated/numpy.transpose.html

Since numpy doesn't normally accept an `out` keyword argument, I would omit it from the frontend function and pass `out=None` to the ivy function, to prevent inconsistent behaviour with different backends.

You also need to add hypothesis tests to https://github.com/martinoywa/ivy/blob/master/ivy_tests/test_ivy/test_frontends/test_numpy/test_matrix/test_methods.py

martinoywa:
> Please double-check the documentation for `numpy.transpose` at https://numpy.org/doc/stable/reference/generated/numpy.transpose.html
> 
> Since numpy doesn't normally accept an `out` keyword argument, I would omit it from the frontend function and pass `out=None` to the ivy function, to prevent inconsistent behaviour with different backends.
> 
> You also need to add hypothesis tests to https://github.com/martinoywa/ivy/blob/master/ivy_tests/test_ivy/test_frontends/test_numpy/test_matrix/test_methods.py

Hi, thanks for the hint on out.
Still working on the test.

Nightcrab (commented):
Do you need help with the hypothesis tests?

martinoywa:
> Do you need help with the hypothesis tests?

Hi, yes, please.
I noticed some changes in the implementation with the addition of classes.

Nightcrab (changes requested):
```
def transpose(self, axis=None):
        return ivy.matrix_transpose(self.A, out=None)
```

What `numpy.matrix.transpose` takes is `axes`, either as positional arguments or a tuple of integers. This is unlike `axis` which is a single integer keyword argument.

The exact section of the numpy docs for the method is https://numpy.org/doc/stable/reference/generated/numpy.matrix.transpose.html?highlight=transpose#numpy.matrix.transpose

It looks like I linked `numpy.transpose` earlier, but that takes `axes` as a keyword argument so matching it would be slightly incorrect. The behaviour is otherwise the same.

As described in docs, `transpose` should permute multiple axes. `ivy.matrix_transpose` is not what you're looking for as it only supports transposing the outer two axes. What I would suggest is using `ivy.permute_dims`:

https://lets-unify.ai/ivy/functional/ivy/manipulation.html#ivy.functional.ivy.manipulation.permute_dims

You would also have to adjust the hypothesis test accordingly. For instance, non-2D arrays would be supported and you would want to pass in axes tuples to the test.

Nightcrab:
@martinoywa  You might want to use the `_array_and_axes_permute_helper` helper function to generate the inputs for this function.

martinoywa:
That's actually what I was using to generate the test cases but they kept on crashing.
This is what I have for the test.

```
@handle_cmd_line_args
@given(
    x_axes=helpers.arrays_and_axes(
        min_num_dims=1,
        max_num_dims=2,
        min_dim_size=1,
        max_dim_size=2,
        num=1
    ),
)
def test_numpy_transpose(
    x_axes,
    as_variable,
    native_array,
    fw,
):
    x, axes = x_axes
    helpers.test_frontend_method(
        input_dtypes_init=ivy.int32,
        as_variable_flags_init=as_variable,
        num_positional_args_init=0,
        native_array_flags_init=native_array,
        all_as_kwargs_np_init={
            "data": x[0],
            },
        input_dtypes_method=[],
        as_variable_flags_method=as_variable,
        num_positional_args_method=0,
        native_array_flags_method=native_array,
        all_as_kwargs_np_method={
            "axes":axes,
        },
        fw=fw,
        frontend="numpy",
        class_name="matrix",
        method_name="transpose",
    )
```

Nightcrab:
Looks to me like a problem with the frontend function. Basically, `numpy.matrix.transpose` accepts `axes=None` while `ivy.permute_dims` doesn't, so you need to implement the default behaviour, which according to numpy docs is this

```None or no argument: reverses the order of the axes.```

I imagine you can do this by passing the integers `n` to `0` to `ivy.permute_dims` where `n` is the number of dimensions of the input array.

**Comment by martinoywa** [comment]
Hi, was wondering what you meant by `n` to `0` to `ivy.permute_dims`?

**Comment by Nightcrab** [comment]
If you call ``ivy.permute_dims(x,[5,4,3,2,1,0])`` where x has 6 dimensions for example, what you get is a reversal of the axes (a.k.a a transposition in 6 dimensions).

**Comment by martinoywa** [comment]
I get it now, thanks.

**Comment by martinoywa** [comment]
Hi, I coded up what you mentioned earlier but I'm getting this new error which seems to be coming from the helper function that implements `arrays_and_axes` where it generates an `axes` that's a mismatch with the arrays. Any ideas on how to solve this?

```
test_methods.py:118: 
../../../helpers/function_testing.py:1222: in test_frontend_method
    ret, ret_np_flat = get_ret_and_flattened_np_array(
../../../helpers/function_testing.py:1722: in get_ret_and_flattened_np_array
    ret = fn(*args, **kwargs)
../../../../../ivy/functional/frontends/numpy/matrix/methods.py:125: in transpose
    return ivy.permute_dims(self.A, axes=axes, out=None)

args = (ivy.array([[0]]),), kwargs = {'axes': [0], 'out': None}

    @functools.wraps(fn)
    def new_fn(*args, **kwargs):
        """
        Catch all exceptions and raise them in IvyException
    
        Parameters
        ----------
        args
            The arguments to be passed to the function.
    
        kwargs
            The keyword arguments to be passed to the function.
    
        Returns
        -------
            The return of the function, or raise IvyException if error is thrown.
        """
        try:
            return fn(*args, **kwargs)
        except (IndexError, ValueError) as e:
>           raise ivy.exceptions.IvyError(fn.__name__, str(e))
E           ivy.exceptions.IvyError: numpy: permute_dims: axes don't match array
```

**Comment by martinoywa** [comment]
The `transpose` function looks as such;

```
def transpose(self, axes=None):
        if None in axes:
            return ivy.permute_dims(self.A, axes=tuple(range(self.A.ndim)[::-1]))
        return ivy.permute_dims(self.A, axes=axes, out=None)
```

Nightcrab:
It seems like `arrays_and_axes` will not really do what we're looking for which is generating a permutation of size `ndims`.

I don't know of a helper that does this, so what I would suggest is using `dtype_and_values` to get the array then write a permutation generator in the test.

You could perhaps use `helpers.floats` to generate a float between 0 and 1 then use it to seed `random.shuffle(y)` where `y=[x for x in range(array.ndim)]`.

ivy-seed:

martinoywa:
WIP

martinoywa:
> It seems like `arrays_and_axes` will not really do what we're looking for which is generating a permutation of size `ndims`.
> 
> I don't know of a helper that does this, so what I would suggest is using `dtype_and_values` to get the array then write a permutation generator in the test.
> 
> You could perhaps use `helpers.floats` to generate a float between 0 and 1 then use it to seed `random.shuffle(y)` where `y=[x for x in range(array.ndim)]`.

I'm quite unfamiliar with hypothesis testing. How will this generated array help in testing the function?

Nightcrab:
In general there are helpers to randomly generate test cases, but here it would be best to write your own. With the randomly shuffled array, you can provide test inputs that match what the function is expecting (permutations).

Nightcrab:
Regular unit tests aren't randomly generated, and you're still using the hypothesis framework. You'd just be writing your own test strategy.

ivy-seed:

ivy-seed:

PR #4595, opened 2022-09-18
add max to numpy in matrix

### LINKED ISSUE #4589: max

PatricYan:
Close #4589


modanesh:
Thanks for you contribution. It looks good to me. Could you just please revert back the changes you made to other functions like `any()` in the `methods.py` or changes in the `test_methods.py`?

PatricYan:
> Thanks for you contribution. It looks good to me. Could you just please revert back the changes you made to other functions like `any()` in the `methods.py` or changes in the `test_methods.py`?

done

modanesh:
Thanks. Could you take a look at the logs of `test-frontend-numpy` tests? I see that the `test_numpy_max` is being run with `torch` as backend resulting in test failures.

PatricYan:
> Thanks. Could you take a look at the logs of `test-frontend-numpy` tests? I see that the `test_numpy_max` is being run with `torch` as backend resulting in test failures.

This is numpy's function, why run it in the torch backend?

modanesh:
So basically the outputs of the ground truth and the frontend functions are converted to numpy arrays (regardless of the original framework) and asserted to be the same. From the error logs, I see it throws `ivy.exceptions.IvyBackendException: torch: to_numpy: Expected a pytorch tensor`. So somewhere in the chain there might be a missing decorator to convert arrays between Ivy Arrays and Native Arrays. I would suggest you try debugging one of the examples that are failing (can be found in the logs) and see if that makes it any clearer. For that, what I'd do is to make a simple Python script with that particular case and go step by step to understand what's happening and what's going wrong. Just make sure that you set the backend to Torch.

PatricYan:
in my localhost, test case is ok.

PR #4602, opened 2022-09-18
Added unshuffle function and tests

Prabhu-V:
#4131


JamieLine:
Hi, 

Thanks for the PR! This is really good, I considered merging it as it is, I'm just going to request a few quick changes to keep it in line with the style of similar functions. 

1. The `assert` statements should be `ivy.assertions.check_equal` statements or similar `ivy.assertions.*` statements. This is just to ensure that `IvyException` is raised correctly. In the case where you're asserting two equalities, you can try either using two statements or checking that `[x, y] == [z, w]` which is `True` if and only if `x == z` and `y == w`. Whether or not you use the list method should depend on whether or not it produces "cleaner" code than two separate statements, which I think it will with a comment along the lines of `#Assert x == z and y == w` alongside it.
2. `"pixel_shuffle expects 4D input, but got input with sizes "` should be changed to match the function name. 
3. In the test, instead of setting `with_out=False`, create a new parameter to the test called `with_out` and in the call to `helpers.test_frontend_function` specify `with_out=with_out`. You don't need to add anything to the `@given` decorator, this parameter will be dealt with for you by `@handle_cmd_line_args`. This is just future proofing against this function gaining an `out` parameter in the future. I acknowledge this is nitpicky but it is good practice. 
4. Where you do string concatenations (e.g. `"downscale_factor, but got input with sizes " + str(input_shape)`) consider using f-strings, e.g. `f"downscale_factor, but got input with sizes {input_shape}"` which will do some formatting trickery to make it easier, I'm ultimately not sure if this produces cleaner code for a message that's so long but it's worth trying. Only do this if it actually produces nicer looking code, it might not.

Ultimately this is very good though, just a couple of quick changes here and there and I'll merge it in.

Prabhu-V:
> Hi,
> 
> Thanks for the PR! This is really good, I considered merging it as it is, I'm just going to request a few quick changes to keep it in line with the style of similar functions.
> 
> 1. The `assert` statements should be `ivy.assertions.check_equal` statements or similar `ivy.assertions.*` statements. This is just to ensure that `IvyException` is raised correctly. In the case where you're asserting two equalities, you can try either using two statements or checking that `[x, y] == [z, w]` which is `True` if and only if `x == z` and `y == w`. Whether or not you use the list method should depend on whether or not it produces "cleaner" code than two separate statements, which I think it will with a comment along the lines of `#Assert x == z and y == w` alongside it.
> 2. `"pixel_shuffle expects 4D input, but got input with sizes "` should be changed to match the function name.
> 3. In the test, instead of setting `with_out=False`, create a new parameter to the test called `with_out` and in the call to `helpers.test_frontend_function` specify `with_out=with_out`. You don't need to add anything to the `@given` decorator, this parameter will be dealt with for you by `@handle_cmd_line_args`. This is just future proofing against this function gaining an `out` parameter in the future. I acknowledge this is nitpicky but it is good practice.
> 4. Where you do string concatenations (e.g. `"downscale_factor, but got input with sizes " + str(input_shape)`) consider using f-strings, e.g. `f"downscale_factor, but got input with sizes {input_shape}"` which will do some formatting trickery to make it easier, I'm ultimately not sure if this produces cleaner code for a message that's so long but it's worth trying. Only do this if it actually produces nicer looking code, it might not.
> 
> Ultimately this is very good though, just a couple of quick changes here and there and I'll merge it in.
> 
> Any questions or comments please let me know!

@JamieLine I have made the suggested changes. Let me know if any other changes are required.

PR #4607, opened 2022-09-18
Addition of ivy.trunc_divide

Ishticode:
This could've just been implemented using `ivy.trunc(ivy.divide(x1,x2))` within functional/ivy rather than for individual backend as done here. 

**Let me know @djl11 whether I should update it to that...**

### DIFF
--- a/ivy/array/elementwise.py
+++ b/ivy/array/elementwise.py
@@ -2029,3 +2029,47 @@ class ArrayWithElementwise(abc.ABC):
 
         """
         return ivy.rad2deg(self._data, out=out)
+
+    def divide(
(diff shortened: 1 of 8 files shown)


djl11 (changes requested):
I agree this is a difficult call as to whether it should be implemented as a primary, compositional or mixed function. In this case, I would actually advise it be mixed, with a `torch` implementation and the rest all delegating to the compositional implementation. By the looks of it, all the other backend implementations are _pretty much_ just doing `ivy.trunc(ivy.divide(...))` anyway.

djl11 (approved):
provided the tests are passing, it lgtm!

PR #4611, opened 2022-09-19
Individual task: Adding `jax.numpy` frontend skeleton files

hirwa-nshuti:
Referring to how the documentation is set there is a group of `jax.numpy` general functions which are numpy functions supported in the `jax.numpy` namespace I made a .py file named `namespace_functions.py`


djl11 (approved):
lgtm!

PR #4643, opened 2022-09-19
changes and fixes

RickSanchezStoic:
fixes and changes for more dtype promotion


mattbarrett98:
Just curious, what makes these type promotions necessary- are any tests failing?

RickSanchezStoic:
Yeah, so this was part of the individual task that was assigned to me, while I had implemented the type promotions for the functions previously but it later turned out that some functions were missed or were not discovered yet. Also, it was proposed that many functions were not being tested for this either. So I looked for the functions mainly those that perform operations on two or more inputs and did the type promotion to stay on the safer side.

mattbarrett98 (commented):
Ah okay. Recently the tests for type promotion were fixed and are now all running. You can have a look in the `type_promotions` array API tests. I think the only ones still with errors are some of the bitwise functions. Hence we shouldn't bother with the casting in these functions, as it will only slow them down.

RickSanchezStoic:
alright then, will close this for now.

PR #4665, opened 2022-09-19
Add exists module

wmfidelis:


wfidelis:
Close #4665

wfidelis:
The failed checks are unrelated to my code. Please review when you can. Thanks!

johanjino:
@wmfidelis I think the exists module is already added. `return x is not None` does the work correctly. Not sure what you have tried to implement. Might be our mistake if the task was left opened. I apologize for that🙂.

PR #4680, opened 2022-09-20
exception handling docs

juliagsy:


docs/partial_source/deep_dive/18_exception_handling.rst
```diff
@@ -0,0 +1,209 @@
...
+            self._default = [
+                "numpy" if ivy.current_backend_str() == "" else ivy.current_backend_str()
+            ]
+            self._delimiter = ": "
+            for message in messages:
+                self._default.append(message)
+            super().__init__(self._delimiter.join(self._default))
+
+In some Array API tests, :code:`IndexError` and :code:`ValueError` are
+explicitly tested to ensure that the functions are behaving correctly. Thus, we
```
djl11, line 47: The :code:`IvyError` class unifies these special cases.

djl11 (changes requested):
very minor changes, and all just about the writing style. I think it's better to use language less tied to us as creators. For example avoiding phrases like "we created <class_name> because..." and instead having phrases like "<class_name> does this.", if that makes sense :) Once modified, feel free to merge without requesting another review!

juliagsy:
Noted! Will take note in future writing! Thanks for the review!

Issue #4614, opened 2022-09-19
Add Linear algebra functions to Jax frontend


hirwa-nshuti:
Add [Linalg functions](https://jax.readthedocs.io/en/latest/jax.numpy.html#module-jax.numpy.linalg) to Jax frontend

\_

>Please keep in mind that the proper way to link an issue to this list is to comment "- [ ] #issue\_number" while the issue's title only includes the name of the function you've chosen.

\_

- [x] #4716
- [x] cond
- [x] #4719
- [x] #6084
- [x] #4717
- [x] eigvals
- [x] #4724
- [x] #4718
- [x] #22867
- [x] #5972
- [x] matrix\_rank
- [x] #19403
- [x] #5758
- [x] #6207
- [x] #4720
- [x] slogdet
- [x] #5783
- [x] #7203
- [x] #7387
- [x] tensorsolve

> The main file paths where these functions are likely to be added are: 
> - `ivy/functional/frontends/jax/numpy/linalg.py`
> - `ivy\_tests/test\_ivy/test\_frontends/test\_jax/test\_jax\_numpy\_linalg.py`

PatricYan:
- [ ] #4702

karalleyna:
- [ ] #4716

karalleyna:
- [ ] eigh

karalleyna:
- [ ] #4718

karalleyna:
- [ ] #4719

karalleyna:
- [ ] #4720

karalleyna:
- [ ] #4724

ndurumo254:
- [ ] #4753

AbhishekJha22:
[] #4825

ukj02:
#5142

ndurumo254:
[]#4964

randomstate42:
- [ ] #5492

VikramxD:
#5668

mohame54:
#5758

JonathanGhaly:
- [ ] #5781

oscar1marte:
- [ ] #5783

oscar1marte:
- [ ] #5783

0xcro3dile:
- [ ] #5921

11Kclarke:
-[]#5972

sayantan1410:
#5999

rubythalib33:
- [] #6084

minivet:
- [ ]  #6207

[...]

PR #4466, opened 2022-09-14
svd

karalleyna:
Closes #4425

PR #4708, opened 2022-09-20
add svd to jax numpy linalg

### LINKED ISSUE #4702: svd

PatricYan:
Close #4702


karalleyna:
It has already been done. See #4466

PatricYan:
> It has already been done. See #4466

Hi, the subtask comes from https://github.com/unifyai/ivy/issues/4614, not same as what you have done.

PatricYan:
> Ah, sorry!
 you are welcome, we all contribute to the IVY.

JamieLine:
Hi!

Thanks for the PR! I've had a look and what's there seems quite good, but it is missing a few of the arguments from the JAX function it replicates. 

At https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.svd.html it says the function signature is `jax.numpy.linalg.svd(a, full_matrices=True, compute_uv=True, hermitian=False)` which means that the function added to the Ivy JAX frontend must have the same signature, that is, the same name and parameters.

 This may require some function composition in order to recreate the functionality when `compute_uv=True`, but you don't need to worry about using the input value of `hermitian` because it doesn't affect input-output behaviour. You'll still need to take in a value, so outside code can specify it without crashing, but the function doesn't need to be affected by its value. See `pinv` in `ivy/functional/frontends/numpy/linalg/solving_equations_and_inverting_matrices.py` for an example.

PatricYan:
> 
Hi, if you take a look at **ivy/functional/backends/jax/linear_algebra.py** you will find the difference between the  backend function **svd**  and the [official doc](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.svd.html)

JamieLine:
Hi, I see what you're saying with the backends. This difference occurs because the backends are there to implement Ivy API functions, e.g. `ivy.svd(x, /, *, full_matrices=True)`. It only has to worry about 2 parameters because the Ivy API only specifies 2 parameters for this function.

However the frontends are used for translating programs written for other frameworks, in this case JAX, which may make use of any parameters specified in the original framework's docs. To make what I'm about to say clearer I've added an example, the frontend implementation of `argmax` in the JAX frontend, found at `ivy/functional/frontends/jax/lax/operators.py`.

image

We see that the underlying Ivy function only takes in 2 parameters, `operand` and `axis`, but the corresponding JAX function takes in one additional argument, `index_dtype`. Therefore the frontend function `argmax` must therefore also support the `index_dtype` parameter to allow for this translation to Ivy code to be done more easily. 

To return to your implementation of `svd` for the JAX frontend, it must also take in the parameters `compute_uv=True` and `hermitian=False`, as those are legal parameters to be passed in by JAX programs. Additionally it must also be tested against both inputs. 

However, other than this detail the PR looks very good and most importantly the test looks good, once the 2 new parameters are added I'll run the test and if it passes it should be good to be merged into Ivy.

Once again, any questions or comments feel free to let me know!

PatricYan:
Hi @JamieLine , Done

JamieLine:
Hi, @PatricYan sorry this has taken a while.

I've gone through and had a look, here is what I spotted.

- I think the changes to the Jax backend will need to be undone unfortunately as I believe we need all of the backends to have the same function signature (to ensure the code is stable regardless of which backend has been selected), but I do understand that this provides a partial solution to the problem. 
- The frontend function should take the output from`ivy.svd` and change that if needed to match the output of the real Jax `svd` function. I think that shouldn't be too difficult as the Ivy function appears to compute everything needed all the time, and it looks like it will just be a matter of picking a subset of the output of the Ivy function to return, depending on the inputs.
- In the frontend, you don't need to pass `compute_uv` or `hermitian` into `ivy.svd` as `ivy.svd` doesn't accept these parameters, and therefore passing them causes a test failure.
- I think you can simplify `_svd_get_dtype_and_data` by removing the `ret_shape=True` flag as you don't use the shape parameter it returns later on.
- Try running the test locally to ensure the function doesn't crash, and then I can queue it to be tested on our CI to ensure it works on all backends.

The rest is good though, and again I'm very sorry I've taken so long to look through this.

ivy-seed:

ivy-seed:

ivy-seed:

PR #4730, opened 2022-09-21
feat: stack reformatting task

seppemarotta:
This task solves the - [stack] #4678. issue.

### DIFF
--- a/ivy/array/manipulation.py
+++ b/ivy/array/manipulation.py
@@ -251,6 +251,28 @@ class ArrayWithManipulation(abc.ABC):
         ivy.Array instance method variant of ivy.stack. This method simply
         wraps the function, and so the docstring for ivy.stack also applies
         to this method with minimal changes.
+
+        Parameters
+        ----------
+        self
+            input array.
+        axis
+            axis is the axiis(int) to stack along
+            Default ``0`` wich means the first dimension.
+        out
+            optional output array, for writing the result to. It must have a shape that
+            the inputs broadcast to.
+
+        Returns
+        -------
+        ret
+            An array with the elements of self.
+
+        Examples
+        --------
+        >>> x = ivy.stack([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
+        >>> x
+        ivy.array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
         """
         return ivy.stack([self._data] + arrays, axis=axis, out=out)
 
--- a/ivy/array/manipulation.py
+++ b/ivy/array/manipulation.py
@@ -251,28 +251,6 @@ class ArrayWithManipulation(abc.ABC):
         ivy.Array instance method variant of ivy.stack. This method simply
         wraps the function, and so the docstring for ivy.stack also applies
         to this method with minimal changes.
-
-        Parameters
-        ----------
-        self
-            input array.
-        axis
-            axis is the axiis(int) to stack along
-            Default ``0`` wich means the first dimension.
-        out
-            optional output array, for writing the result to. It must have a shape that
-            the inputs broadcast to.
-
-        Returns
-        -------
-        ret
-            An array with the elements of self.
-
-        Examples
-        --------
-        >>> x = ivy.stack([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
-        >>> x
-        ivy.array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
         """
         return ivy.stack([self._data] + arrays, axis=axis, out=out)
 
--- a/ivy/array/manipulation.py
+++ b/ivy/array/manipulation.py
@@ -252,22 +252,6 @@ class ArrayWithManipulation(abc.ABC):
         wraps the function, and so the docstring for ivy.stack also applies
         to this method with minimal changes.
 
-        Parameters
-        ----------
-        self
-            input array.
-        axis
-            axis is the axiis(int) to stack along
-            Default ``0`` wich means the first dimension.
-        out
-            optional output array, for writing the result to. It must have a shape that
-            the inputs broadcast to.
-
-        Returns
-        -------
-        ret
-            An array with the elements of self.
-
         Examples
         --------
         >>> x = ivy.stack([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
--- a/ivy/array/manipulation.py
+++ b/ivy/array/manipulation.py
@@ -254,9 +254,22 @@ class ArrayWithManipulation(abc.ABC):
 
         Examples
         --------
-        >>> x = ivy.stack([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
-        >>> x
-        ivy.array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
+        >>> x = ivy.array([[2,3],[1,4]])
+        >>> y = ivy.array([[4,5],[6,7]])
+        >>> x.stack(y)
+        ivy.array([[[ 6,  8],
+                [ 7, 11]]])
+        >>>
+
+        >>> x = ivy.array([[1,2],[3,4]])
+        >>> y = ivy.array([[5,6],[7,8]])
+        >>> x.stack(y,axis=1)
+        ivy.array([[[ 6,  8]],
+            [[10, 12]]])
+        >>> x.stack(y,axis=0)
+        ivy.array([[[ 6,  8],
+                [10, 12]]])
+        >>>
         """
         return ivy.stack([self._data] + arrays, axis=axis, out=out)
 
--- a/ivy/array/manipulation.py
+++ b/ivy/array/manipulation.py
@@ -254,22 +254,17 @@ class ArrayWithManipulation(abc.ABC):
 
         Examples
         --------
-        >>> x = ivy.array([[2,3],[1,4]])
-        >>> y = ivy.array([[4,5],[6,7]])
-        >>> x.stack([y])
-        ivy.array([[[ 6,  8],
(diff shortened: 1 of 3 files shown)


Nightcrab:
add_reformatting_checklist_manipulation

Nightcrab:
Please comment with `add_reformatting_checklist_manipulation` to generate the reformatting checklist for this PR.

seppemarotta:
add_reformatting_checklist_manipulation

seppemarotta:
@Nightcrab don

Nightcrab:
Regarding the way the checklist should be filled out, have a look at this section in the docs: https://lets-unify.ai/ivy/contributing/4_open_tasks.html#formatting-checklist

I would also advise checking the docs of the function being wrapped, ``ivy.stack`` at https://lets-unify.ai/ivy/functional/ivy/manipulation/stack/stack_functional.html 

Since your docstrings are missing information about the `arrays` parameter.

Regarding the examples, while ``ivy.stack`` acts as an identity function when passed just one array/container, since it accepts multiple there should be examples for that behaviour as well.

seppemarotta:
@Nightcrab thanks, for pointing those out, will review those docs in detail.

seppemarotta:
@Nightcrab I've updated my PR and I've also updated the checklist accordingly. I only have one question in relation to the docstring tests, I see under the logs that there is no error in relation to the stack function, but still, those docstring tests are failing. Is that something expected?

Nightcrab:
Yes, that's expected as some of the existing docstrings have issues. As long as the tests past for the docstrings of `stack` then they would be considered passing for this PR.

seppemarotta:
@Nightcrab Then this PR is ready for review. I've been double-checking those logs and I don't see any errors for the stack function in the run-docstring-tests Jenkins pipeline.

Nightcrab (changes requested):
**ivy/array/manipulation.py stack** method
Docstring missing examples using the `axis` keyword.
```
        >>> x = ivy.stack([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
        >>> x
        ivy.array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
```
The example uses the `ivy.stack` function instead of `ivy.Array.stack` which the docstring is for.

**ivy/container/manipulation.py static_stack** method
```
ivy.Container instance method variant of ivy.stack. This method simply
        wraps the function, and so the docstring for ivy.stack also applies
        to this method with minimal changes.
        Parameters
        ----------
        self
            input array.
```
The method is static not instance, so it takes positional argument `xs`, not a `self` instance.
There's no example showing the use of `axis`.

**ivy/container/manipulation.py stack** method
```
x = ivy.Container(a=ivy.array([0., 1., 2.]), \
            b=ivy.array([3., 4., 5.]), \
            c=ivy.array([6.,7.,8.]))
y = ivy.stack(x)
```
This is `ivy.stack` being used again when the docstring is for instance method  `ivy.Container.stack`.

**ivy/functional/ivy/manipulation.py stack** function
```
>>> x = ivy.Container(a=ivy.array([0., 1., 2.]), b=ivy.array([3., 4., 5.]))
    >>> y = ivy.Container.static_stack(x)
    >>> y
    {
        a: ivy.array([0., 1., 2.]),
        b: ivy.array([3., 4., 5.])
    }
```
This example is using `ivy.Container.static_stack` even though the docstring is for the functional method `ivy.stack`. 


```
>>> ivy.native_array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
    array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
```
This example has nothing to do with `ivy.stack`.

```
>>> x = ivy.Container(a=ivy.array([0., 1., 2.]), \
            b=ivy.array([3., 4., 5.]), \
            c=ivy.array([6.,7.,8.]))
    >>> y = ivy.stack(x)
    >>> y
    {
        a: ivy.array([0., 1., 2.]),
        b: ivy.array([3., 4., 5.]),
        c: ivy.array([6., 7., 8.])
    }
```

`ivy.Container.stack` is the only method which accepts an `ivy.Container` instance. Yet there are examples in this and the other docstrings where `ivy.Container` instances are passed despite absence from argument type-hints.

All examples also only show trivial use cases of `stack` where it does nothing to the input before returning. These seem redundant.

All are missing examples passing arguments that correspond to explicitly supported types, e.g.`Tuple[ivy.Array]`, `List[ivy.NativeArray]`.

seppemarotta:
@Nightcrab thanks for the details, working on it 👍

seppemarotta:
Still working on it, will have it updated by tomorrow.

seppemarotta:
@Nightcrab I've updated this PR with the required changes.

Nightcrab (changes requested):
`ivy/array/manipulation.py` `stack`

```
>>> x = ivy.array([[2,3],[1,4]])
        >>> y = ivy.array([[4,5],[6,7]])
        >>> x.stack(y)
        ivy.array([[[ 6,  8],
                [ 7, 11]]])
```

This should call `x.stack([y])` instead. 

Because the function takes a tuple/list of arrays, `x.stack(y)` tries to strip the outer dimension of `y` before calling `ivy.stack`, making `x.stack(y)` an illegal use case, but it works because of a bug with the implementation:

```
return ivy.stack([self._data] + arrays, axis=axis, out=out)
```

This should use an `ivy.Array.concat` instead of `+`, since when the function is passed arrays of different shapes, the `+` may be interpreted as addition rather than concatenation.

If you could make this change and then update the examples that would be good.

---

`ivy/container/manipulation.py` `static_stack`

```
>>> ivy.Container.static_tile(x,(2,3))
```

This is a typo.

```
>>> x = ivy.Container(a=ivy.array([[0, 1], [2,3]]), b=ivy.array([[4, 5]]))
        >>> ivy.Container.static_stack(x)
        {
            a: ivy.array([[0, 1],
                        [2, 3]]),
            b: ivy.array([[4, 5]])
        }
```

I would remove this example and keep the next one.

```
>>> ivy.Container.static_stack(x,axis = 1)
        {
            a: ivy.array([[0, 2],
                        [1, 3]]),
            b: ivy.array([[4],
                        [5]])
        }
```

I'd suggest also passing a tuple or list of Containers in an example.

---

`ivy/container/manipulation.py` `stack`

This function just calls `ivy.Container.static_stack` with `self` inserted into the list of Containers, so I'd recommend using the same examples in both but with different syntax.

---

`ivy/functional/ivy/manipulation.py` `stack`

```
With :code: `ivy.Array` & `ivy.Container` input:
    >>> x = ivy.Container(a=ivy.array([0., 1., 2.]), \
            b=ivy.array([3., 4., 5.]), \
            c=ivy.array([6.,7.,8.]))
```

This section seems to end after defining `x`.

```
 With :code: `ivy.Container` input and different `axis` :
    >>> x = ivy.Container(a=ivy.array([[0., 1.],[3.,4.]]), \
    ...             b=ivy.array([[5., 6.],[7.,8.]]), \
    ...             c=ivy.array([[9., 10.],[11.,12.]]))
    >>> ivy.stack(x,axis=0)
    {
        a: ivy.array([[0., 1.],
                    [3., 4.]]),
        b: ivy.array([[5., 6.],
                    [7., 8.]]),
        c: ivy.array([[9., 10.],
                    [11., 12.]])
    }
    >>> ivy.stack(x,axis=1)
    {
        a: ivy.array([[0., 3.],
                    [1., 4.]]),
        b: ivy.array([[5., 7.],
                    [6., 8.]]),
        c: ivy.array([[9., 11.],
                    [10., 12.]])
    }
```

I would leave out `Container` examples since the docstring and type-hints don't indicate that `ivy.stack` accepts Containers.

The other examples here are good.

seppemarotta:
@Nightcrab PR updated

ivy/array/manipulation.py
```diff
@@ -251,8 +251,27 @@ def stack(
...
+        >>> x.stack([y],axis=1)
+        ivy.array([[[ 6,  8]],
+            [[10, 12]]])
+        >>> x.stack([y],axis=0)
+        ivy.array([[[ 6,  8],
+                [10, 12]]])
+        >>>
         """
-        return ivy.stack([self._data] + arrays, axis=axis, out=out)
+        return ivy.stack(self._data.concat(arrays), axis=axis, out=out)
```
Nightcrab, line 274: should be 

```self.concat``` instead of ```self._data.concat```

**Review by Nightcrab — CHANGES_REQUESTED** [review]

File: ivy/array/manipulation.py (hunk 1 above)
Comment (seppemarotta) [reply ivy/array/manipulation.py]: Thanks for pointing that out. I've updated my pr.

Comment (Nightcrab) [reply ivy/array/manipulation.py]: You'll need to update the example now since the outputs are different

**Comment by seppemarotta** [comment]
@Nightcrab PR Updated

**Review by Nightcrab — CHANGES_REQUESTED** [review]
```
>>> x = ivy.array([[2,3],[1,4]])
        >>> y = ivy.array([[4,5],[6,7]])
        >>> x.stack([y])
        ivy.array([[[ 6,  8],
                [ 7, 11]]])
        >>>
        >>> x = ivy.array([[1,2],[3,4]])
        >>> y = ivy.array([[5,6],[7,8]])
        >>> x.stack([y],axis=1)
        ivy.array([[[ 6,  8]],
            [[10, 12]]])
        >>> x.stack([y],axis=0)
        ivy.array([[[ 6,  8],
                [10, 12]]])
```
Update these

seppemarotta:
@Nightcrab I've updated those with the right examples, let me know if it needs something else. 👍

Nightcrab (approved):
Looks good to me. I'll merge it.

seppemarotta:
Thanks! @Nightcrab

PR #4767, opened 2022-09-21
Ivy extension ifft function

AhmedMousa-ag:


JamieLine:
Hi, thanks for the PR!

I've had a quick look, and most of this is very good. I have a few quick questions / comments about some things in this PR.

1. In the Tensorflow backend implementation, why is there a `if not isinstance(ivy.backend(), ivy.get_backend("tensorflow")):` section where no equivalent is in the other frameworks? Is there some strange behaviour that occurs only in Tensorflow? (I ask mostly because I'm not an expert with Tensorflow so I might just be missing something obvious)
2. In the frontend test, you define a function called `_get_rand_str` which is a perfectly sensible thing to do to solve the problem at hand, but this can be done more easily using Hypothesis' `st.sampled_from` function, which also allows Hypothesis to more easily reproduce failures.

If you have any questions or comments feel free to let me know!

AhmedMousa-ag:
Regarding the TensorFlow function, TensorFlow doesn't allow dim or axis, only the name of the layer if it's a layer and the input.

And since Tensorflow is different than all other packages, thereby most likely other packages users will assume it's the same as other packages if they used Tensorflow.

Unfortnully it didn't occur to me until now that the frontend function will prohibit other parameters than inputs and the name of the layer. I will delete it. just wanted to be extra safe.

Regarding the test function, I checked ivy documentation on test functions and didn't find any related to this case. I will check your suggestion and modify it for better test results.

Thank you

AhmedMousa-ag:
I'm sorry I checked the logic again.

Regarding the TensorFlow function, In case the frontend function is the same as TensorFlow, then I don't want to calculate n and handling operations, as it's already will be handled by TensorFlow function.

If we didn't check this operation then we waste more speed and operation on calculations we don't need.

PR #4782, opened 2022-09-22
Modify parameter for the scatter_nd function of torch backend

kimbring2:
It seems like the result of the scatter_nd function of the Torch backend is not consistent with the outcome of the Tensorflow backend ground-truth.

I need to change the parameter to decide the initial_val when the reduction is max. It is the same value as the Tensorflow backend parameter. 

P.S. I tested it except for the 'bfloat type' which is not directly related to this issue.

### DIFF
--- a/ivy/functional/backends/jax/__init__.py
+++ b/ivy/functional/backends/jax/__init__.py
@@ -94,7 +94,6 @@ valid_numeric_dtypes = (
     ivy.uint16,
     ivy.uint32,
     ivy.uint64,
-    ivy.bfloat16,
     ivy.float16,
     ivy.float32,
     ivy.float64,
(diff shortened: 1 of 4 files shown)


VedPatwardhan (commented):
Hi @kimbring2, thanks for requesting the review. I think we shouldn't be removing `bfloat16` from the `valid_numeric_dtypes` for the backends, please revert that change. Apart from that, the change related to `initial_val` looks good to me, but I'll request Saeed's review on the PR given that he's been working on this function.

kimbring2:
@VedPatwardhan Thank you for checking my PR. I still do not know how to deal with the 'bfloat16' well. Most tests just give an error for that data type.

It is good to hear that Saeed is working on this function. Maybe, this PR can be helpful to save debugging time.

saeedashrraf:
Hello @kimbring2 , all the necessary changes covering this PR for scatter_nd are now implemented on the main repo. In lights of that, I will go ahead and close this PR now.

kimbring2:
@saeedashrraf Yes, I understand. Thank you for checking it!

PR #4788, opened 2022-09-22
test helpers update and tests refactor

CatB1t:
#### Functions that will always return NumPy arrays for values:
- `array_values` *Note: always returns a single NumPy array.
- `dtype_and_values`
- `dtypes_values_axis`

dtype is always returned as a list, values are always returned as a list of NumPy arrays, even if it's one np array.

#### Helper functions that will always return lists:
- `get_dtypes`

#### Flags that will always be generated as a list, they must also be always passed as a list:
- `native_array_flags`
- `as_variable_flags`
- `container_flags`

#### NumPy tests specific:
- `handle_where_and_array_bools` will return new lists for the flags, it doesn't update it in place anymore, this resolves a bug to owhere in the case of running the test multiple times (the decorator gets called once) and the list will always refer to the same one created when the test is ran.


ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -405,25 +404,23 @@ def test_from_dlpack(
     dtype, x = dtype_and_x
     helpers.test_function(
         input_dtypes=dtype,
-        as_variable_flags=False,  # can't convert variables
+        as_variable_flags=[False],  # can't convert variables
```
djl11, line 407: changing lines like this is okay (perhaps it runs more efficiently when passed as a list), but I think we should still be able to handle the case if this was passed as `False`. Are we still able to handle such cases?

djl11 (approved):
Just one small question, but otherwise great work as always @CatB1t!

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
CatB1t: We used to convert these into lists in the test functions, this adds some overhead which isn't necessary (we check if it's a list, if not we create one), also it's more intuitive to pass list to `flags`, I could add it back if that makes things simpler.

PR #4809, opened 2022-09-22
Searching Fixup

Unevilicorn:
superset added for nonzero


mattbarrett98 (commented):
You've caused a docstring failure (with one of your examples), but apart from this it looks good 🙂

Unevilicorn:
I have now fixed the docstring failure :smiley:

mattbarrett98 (commented):
Can `fill_value` only be an int? In jax's docs it says it's an array. Regardless, the type hints should be consistent, in `static_nonzero` the type hint is `Number` which is different from the rest which are `int`. Once these are consistent feel free to merge 🙂

Unevilicorn:
It's supposed to be `Number`, I must have somehow only changed one of the dtypes :sweat_smile:. Although the jax doc says it expects an array, it only expects a 0 dim array or a scalar, so for simplicity sake, I thought it might make more sense if we use Number instead.

PR #4836, opened 2022-09-23
added squeeze instance method to numpy frontend

NicAOB97:


JonathanYang-yzh (changes requested):
Hi! Thank you for the commit! Could you please also write the corresponding test for the squeeze function? Please let me know when you are finished.

NicAOB97 (commented):
Squeeze test added. Not sure whether it is okay.

PR #4838, opened 2022-09-23
Ivy extension ifft function

AhmedMousa-ag:
had to re-code due to big conflict with the main branch.
previous pull request link:
https://github.com/unifyai/ivy/pull/4767


jkeane508:
Hey @JamieLine, 

I believe this is a follow up to a PR that you were handling, I've assigned you and requested your review.

Thanks :)

JamieLine:
Hey @jkeane508, thank you for assigning me, I'm taking a look now and reading through the comments on the old PR

JamieLine:
Hi @AhmedMousa-ag 

I've had a quick look through the code, and when trying to run the test I got an error saying `KeyError: 'ifft'`. This is happening because right now the PR is missing an implementation for the Ivy version of this function, so `ivy/functional/ivy/extensions.py` needs to have a function added to call the functionality in the backends, and this Ivy version will be used by either other Ivy functions or the frontends to make it framework agnostic. 

This function will probably just be something along the lines of `return ivy.current_backend(x).ifft(.....)` and that should then allow the `num_positional_args` helper to find the function, and the test should then pass (if it doesn't then some work will need to be done to make it pass).

AhmedMousa-ag:
Hi @JamieLine,

It must have been lost while re-coding it. Thank you for the heads up.

JamieLine:
Hi,

No worries about losing the test I've done it before as well. I haven't quite been able to go through and review everything yet (I've been asked to work on something urgent, don't worry I'll still finish reviewing your PR but that might need to wait until Monday) but I have managed to find a few things in the code that I'm not too sure about.

1. The name parameter isn't being generated by the test, it always passes through as None, when it should have something passed in.
2. The name parameter might not be needed in the Ivy function. I'm not too sure about this so I've asked internally to see what the team thinks, but I'm not sure if it is needed as it doesn't affect I/O behaviour and only appears in one backend. I'll get back to you when I know what the team thinks.
3. `Axis` and `dim` are the same in the test, and I think that one of those arguments is redundant. From what I could see, `axis` is used for most backends, and `dim` is used in the others, for seemingly the same purpose. If this is the case, one of them can be removed, and the other one can be used in all backends. I think one of the backends named the argument `dim` in their documentation, we can just pass `axis=dim` to get around that as long as both our `axis` and their `dim` argument mean the same thing.
4. I found a test failure when I ran the new test, try running it locally and see what faults it gives you, I think the arguments are in a different order in the test, which might cause some strange behaviour. If the tests crash saying there's no module named `pytest` or `hypothesis`, install everything listed in the optional requirements file in the repository. 

JamieLine:
The team have decided that the `name` parameter should be removed from Ivy function, I understand why it was there and to be honest I'd have added it as well initially but as it doesn't affect I/O behaviour it should be removed.

AhmedMousa-ag:
@JamieLine No problem at all, take your time.
I'm actually doing this task to figure some things out, the Ivy test on my device behaves a little weird that it didn't pass one of the actual implemented function and was succefully merged. So I thought there must be something wrong.
Test results for me varies each time that I get an error, and it changes when I restart my device. I will push the current code now and will look for the issue until I find it.

JamieLine:
Hi, your tests sound like they might be working correctly then, there's a little bit of context here that might help.

The testing is done (partially) by a library called Hypothesis, which generates random (and continually changing) data to test the functions against, and stores any data that causes the test to fail. This means that if a function passes for most but not all values, it will usually pass but sometimes fail. Sometimes a function passes local tests, then gets pushed to the repository and the additional testing volume it goes through finds a failure. That's why your tests change so often, it's intentionally not quite repeating the same test. 

Which tests did you find that failed in functions which had already been committed? I can flag them up to whoever implemented them

JamieLine:
Also, there are some changes made to the `.idea` folder, these will need to be removed before the code can be merged to master

AhmedMousa-ag:
@JamieLine 
image

This is a test result in Ivy.losses. binary_cross_entropy, and actually my test gives the same result

JamieLine:
Thanks! I'll flag those up to the author of those functions!

PR #4840, opened 2022-09-23
add softmax function tensorflow frontend

zuruoke:


zuruoke:
https://github.com/unifyai/ivy/issues/4819

ivy/functional/frontends/tensorflow/raw_ops.py
```diff
@@ -131,6 +131,12 @@ def Square(*, x, name="Square"):
     return ivy.square(x)
 
 
+def Softmax(*, x, axis=None, name="Softmax"):
+    if x.dtype not in ["half", "float32", "float64"]:
+        x = ivy.astype(x, "float64")
```
jkeane508, line 136: I think that we should cast to float, but not necessarily float64 in all cases. We should maybe look at casting to the nearest reasonable float dtype instead

jkeane508 (changes requested):

jkeane508 (changes requested):
The test for Softmax is also failing [here](https://github.com/unifyai/ivy/actions/runs/3113328790/jobs/5047820850)

jkeane508:
Feel free to request a review form me whenever you have tests passing :)

jkeane508 (changes requested):
It seems to be from the testing that there are failures when the dims argument and the number of dims in the input for the function don't line up. I would suggest looking at the mathematics for softmax and creating a helper function that doesn't generate invalid data :)

ivy-seed:

ivy-seed:

PR #4842, opened 2022-09-23
Docstring style fixes and updated make_docs_without_docker.sh with proper directories

hirwa-nshuti:


docs/make_docs_without_docker.sh
```diff
@@ -21,12 +21,12 @@ rm -rf autogenerated_source
 rm -rf build
 
 # generate content
-python generate_src_rst_files.py  --root_dir ../ivy || exit 1
+python3 generate_src_rst_files.py  --root_dir ../ivy || exit 1
```
VedPatwardhan, line 24: This shouldn't be `python3` but `python` instead as the dockerless pipeline is aimed at working with a virtual environment where `python` is used. Same applies to the other updates from `python` to `python3` in this file

ivy/array/general.py
```diff
@@ -237,15 +237,15 @@ def scatter_nd(
         >>> indices = ivy.array([[4], [3], [1], [7]])
         >>> updates = ivy.array([9, 10, 11, 12])
         >>> scatter = indices.scatter_nd(updates, reduction='replace', out=arr)
-        >>> print(scatter)
+        >>> scatter
```
VedPatwardhan, line 240: I think we should be printing the result for every example, so I'd suggest you revert changes involving the removal of the use of the `print` function from the PR

ivy/array/general.py
```diff
@@ -973,13 +973,15 @@ def inplace_increment(
         With :code:`ivy.Array` instance methods:
 
         >>> x = ivy.array([5.7, 4.3, 2.5, 1.9])
-        >>> y = x.inplace_increment(1)
-        >>> print(y)
+        >>> val = ivy.array(1.)
```
VedPatwardhan, line 976: I've just pushed a fix to master that allows passing a constant directly to `inplace_increment` and `inplace_decrement`, so I'd suggest to revert that to it's previous state. Same applies to the array and container methods for `inplace_increment` and `inplace_decrement`

VedPatwardhan (changes requested):
Hi @hirwa-nshuti, I've suggested a few changes. Apart from that, provided the results of the examples are correct, feel free to merge your PR. Thanks 🙂

PR #4854, opened 2022-09-23
Triu fix

JamieLine:
When trying to fix the failing tests in the frontends, I found a failure in the Tensorflow backend implementation of `triu`.

It would crash if the input was a Boolean type, whereas the other frameworks would return a value for Boolean inputs. I've fixed it by casting the input to int8 and then back, which produces the same output as the other frameworks. This seemed like a more sensible way to fix the crashing test than removing Boolean types from the frontend test, given that the crash only happens in Tensorflow.

Edit: An analogous error happens with `tril`, whatever fix makes it through for this function I'll propagate over to `tril`.


mattbarrett98 (changes requested):
Actually our policy on this is to not just cast things back and forth every time something like this arises. This is because if we did, our code would quickly bloat up with lots of casting. Instead just make sure to only test against numeric data types.

PR #4890, opened 2022-09-25
jax.numpy.linalg.qr

### LINKED ISSUE #4720: qr

karalleyna:
Closes #4720


sharique1006:
`rtol` is not an argument in `qr`, `fw` is no longer needed. I have fixed that, rest lgtm. Cheers!

PR #4891, opened 2022-09-25
implemented jax lax shift_right_logical in the jax lax frontend  and …

vaatsalya123:
…tested.

#1578


vaatsalya123:
I don't think there is any bug so could you please merge this please thanks!

PR #4896, opened 2022-09-25
Add nanmin to numpy frontend

Dsantra92:
Closes: #4851


guillesanbri:
Hi! Thank you for the PR!

One small change: nanmin is located in the Extrema Finding section of the NumPy API (https://numpy.org/doc/stable/reference/routines.math.html#extrema-finding), therefore, these functions should be located in:

ivy/ivy/functional/frontends/numpy/mathematical_functions/extrema_finding.py
ivy/ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_extrema_finding.py

Dsantra92:
Thanks! Will update.

guillesanbri:
Don't worry if tests are not passing because something already in the repo, that's on us 😊​. While it's true that tests are not passing for the `np.amin` frontend function, in this last commit you are using `ivy.amin` which doesn't exist in the Ivy functional API. This should be reverted to the previous logic that used `ivy.min`.

Dsantra92:
np.nanmin relies on np.amin in the original implementation. So is there any way I can use np.amin defined in the same file?

guillesanbri:
If you want to reuse the logic in the `amin` function you can move that code to a private helper function and use that function in both `amin` and `nanmin` as a temporal patch.

What you can also do is open an issue to discuss if `ivy.amin` would be useful to have or if we should include the initial arg to the `ivy.min` definition (which seems sensible to me). There is more info about this kind of situations and the steps to follow [here](https://lets-unify.ai/ivy/contributing/4_open_tasks.html#frontend-apis).

Dsantra92:
I cannot reason as to why the current tests are failing for scalar returns, the ground truth seems empty, leading to Assert error. Another thing is, the initial must be provided for where, which causes it to fail. I am not sure, how to pass that test.

guillesanbri:
Regarding the initial and where arguments, that should be solveable with [hypothesis.assume](https://hypothesis.readthedocs.io/en/latest/details.html#hypothesis.assume) which is actually used in the `test_numpy_amin` function. I will take a look at the assertion you mentioned and update back.

Dsantra92:
The tests that are failing are due to the assertion error. I guess this issue will be solved by #5051 .
Sorry for the commit names, was actually working on another branch on the changes and forgot to squash the commits.

ivy/functional/frontends/numpy/mathematical_functions/extrema_finding.py
```diff
@@ -51,3 +50,41 @@ def amin(
...
+def nanmin(
+    a,
+    axis=None,
+    out=None,
+    keepdims=False,
+    initial=None,
+    where=True,
+):
+    nan_mask = ivy.isnan(a)
+    a = ivy.where(ivy.logical_not(nan_mask), a, ivy.default(out, a.full_like(+ivy.inf)))
```
guillesanbri, line 64: This introduces the out values before taking the min, should be `a = ivy.where(ivy.logical_not(nan_mask), a, a.full_like(+ivy.inf))`

ivy/functional/frontends/numpy/mathematical_functions/extrema_finding.py
```diff
@@ -51,3 +50,41 @@ def amin(
...
+    if initial is not None:
+        if ivy.is_array(where):
+            a = ivy.where(where, a, ivy.default(out, a.full_like(initial)))
```
guillesanbri, line 67: Same as above, out should be removed

ivy/functional/frontends/numpy/mathematical_functions/extrema_finding.py
```diff
@@ -51,3 +50,41 @@ def amin(
...
+            if isinstance(axis, (tuple, list)) or ivy.is_array(axis):
+                a = ivy.concat([a, header], axis=axis[0])
+            else:
+                a = ivy.concat([a, header], axis=axis)
+        else:
+            a = ivy.concat([a, header], axis=0)
+    res = ivy.min(a, axis=axis, keepdims=keepdims, out=out)
+    mask = ivy.all(nan_mask, axis=axis, keepdims=keepdims)
+    if ivy.any(mask):
+        res = ivy.where(ivy.logical_not(mask), res, ivy.nan, out=out)
```
guillesanbri, line 89: As I see it, you are reintroducing the nan values into the res array and thus these lines should be removed, let me know if I'm missing something!

Dsantra92: This is standard numpy operation, where if all the values in the axes is masked out by where, the axis should be nan.

Dsantra92:
Made the changes you requested.

ivy/functional/frontends/numpy/mathematical_functions/extrema_finding.py
guillesanbri: Yeah but you are overriding the nan_mask with both the nan values and the where array, so an input like this diverges from the behaviour in `np.nanmin`

```
a = ivy.array([[ivy.nan, ivy.nan, -3.], [1., ivy.nan, 3.]])
where = ivy.array([[True, False, False], [True, True, True]])
keepdims = True
initial = 3.
axis = 1
```

guillesanbri: I think it should be 

```
mask = ivy.all(ivy.logical_not(where), axis=axis, keepdims=keepdims)
if ivy.any(mask):
    res = ivy.where(ivy.logical_not(mask), res, ivy.nan, out=out)
```

Dsantra92: Oh, I masked the nan mask itself. Thanks!!

guillesanbri:
I think is good to go now, thank you very much for the extra effort! And sorry for the inconvenience 😅​

Dsantra92:
> I think is good to go now, thank you very much for the extra effort! And sorry for the inconvenience sweat_smile​

No worries 😸

PR #4910, opened 2022-09-26
Fix the generalization of inplace testing

hirwa-nshuti:
Summary
-------------
For functions that have `inplace` argument, we can now evaluate the reference of the returned value and also go to perform the value test.
Example of such functions `ivy.functional.frontends.torch.threshold`

For the functions that do `inplace_update` by default like `ivy.functional.frontends.torch.threshold_` which is `inplace` function for `threshold` we are currently checking whether the reference of the returned value is the same as that of the input value, yet the `test_values` is always set to `False` because whenever we call these functions they update `input` with a current return by default.

Thanks:)


djl11 (changes requested):
tiny change request, otherwise looks good and you're free to merge! Thanks!

PR #4911, opened 2022-09-26
refactor helpers structure

CatB1t:
New structure for the test helpers:
- `assertions`: includes all functions that is used by test functions to assert, includes check functions used by test functions too.
- `function_testing`: includes all test functions and its helpers, e.g. `test_function`, `_get_nested_np_arrays`, etc.
- `testing_helpers`: includes all functions that doesn't belong to the two categories above, and those are more general, e.g. `handle_cmd_line_args`, `num_positional_args`.

Hypothesis helpers (includes all hypothesis strategies):
- `array_helpers`: array and matrices strategies.
- `dtype_helpers`: `get_dtypes`, dtype generation related strategies
- `number_helpers`: numeric strategies.
- `general_helpers`: other strategies that are not specific to one category, `get_axis`, `get_bounds`.


djl11 (approved):
lgtm!

PR #4930, opened 2022-09-26
Elementwise frontend add

Unevilicorn:
changed the frontend add to use the alpha key instead of manual multiplication


Ishticode (approved):
looks good 🙂

PR #4945, opened 2022-09-27
Added Cumprod to raw_ops

### LINKED ISSUE #4944: Cumprod

KareemMAX:
Close #4944


KareemMAX:
- [ ] https://github.com/unifyai/ivy/issues/4944

KareemMAX:
Can you please review my PR when you have a chance? @abdrahmandiab

KareemMAX:
@abdrahmandiab I noticed you made some changes, can help me understand it? And I was wondering how the review is going?

abdrahmandiab:
> @abdrahmandiab I noticed you made some changes, can help me understand it? And I was wondering how the review is going?

Hi @KareemMAX , I believe I made 2 changes
first, to the tests, I removed the tolerance values `rtol` & `atol` since they are limiting failure detections and are unnecessary as tests pass without them. I also removed the with_out argument generation since for a Tensorflow backend, it is always set as `False`.

and second, I added aliases for `cumprod` and `cumsum` in tensorflow/math.py. This is  to match the Aliases used by Tensorflow since it has a [`tf.math.cumprod`](https://www.tensorflow.org/api_docs/python/tf/math/cumprod) and a [`tf.raw_ops.Cumprod`](https://www.tensorflow.org/api_docs/python/tf/raw_ops/Cumprod), and same goes for `cumsum` (which is not really related to your PR 😅 )

I will be adding some review comments soon to address a few tiny things that have to be updated so stay tuned for that!

ivy/functional/frontends/tensorflow/raw_ops.py
```diff
@@ -68,6 +68,15 @@ def Cosh(*, x, name="cosh"):
     return ivy.cosh(x)
 
 
+def Cumprod(*, x, axis, exclusive=False, reverse=False, name=None):
+    if reverse:
```
abdrahmandiab, line 72: The `ivy.cumprod` function has just been updated to handle the `reverse` argument, could you please use this here instead? 😄 

Thanks!

abdrahmandiab (changes requested):
Hi @KareemMAX , great work so far! We are very close to merging this PR. 🚀

Just a few changes outlined in this review that hopefully should not take too much of your time! 

It would also be nice if you handle the merge conflicts currently present with the branch, and let me know if you need any help with that!

Thanks! 😇

KareemMAX:
Hey @abdrahmandiab, thank you so much for the insights.
I have addressed your comment, while doing so I have found some bugs in `cumprod` backend which I fixed and tests now pass. Awaiting for your review.

Thanks a lot! ❤

abdrahmandiab:
Hi @KareemMAX,
Could you give me an example that was failing due to the bug you fixed? :)

Otherwise, everything looks great! 
Thanks so much for your work, will be merging soon! 🚀

KareemMAX:
When `exclusive=True` tests fail, this was due to backends multiplying tensors with `zeros_like(x)`, I changed them to `ones_like(x)` while keeping tensorflow backend as is. You can also refer to [tensorflow api reference](https://www.tensorflow.org/api_docs/python/tf/raw_ops/Cumprod) to recheck my changes.

PR #4989, opened 2022-09-27
Update document for pytorch inplace test

Daniel4078:


docs/partial_source/deep_dive/17_ivy_frontends_tests.rst
```diff
@@ -139,14 +137,13 @@ ivy.tan()
         np_frontend_helpers.test_frontend_function(
             input_dtypes=input_dtype,
             as_variable_flags=as_variable,
-            with_out=with_out,
+            with_out=True,
```
djl11, line 140: why are we hard-coding this to `True`? This should be randomized by hypothesis.

docs/partial_source/deep_dive/17_ivy_frontends_tests.rst
```diff
@@ -200,41 +197,45 @@ ivy.tan()
...
     ):
-        input_dtype, x = dtype_and_x
+        input_dtype, input = dtype_and_input
+        assume("float16" not in input_dtype)
         helpers.test_frontend_function(
             input_dtypes=input_dtype,
             as_variable_flags=as_variable,
-            with_out=with_out,
+            with_out=False,
+            with_inplace=True,
```
djl11, line 224: why are we hard-coding this to `True`? This should also be randomized by hypothesis.

docs/partial_source/deep_dive/17_ivy_frontends_tests.rst
```diff
@@ -200,41 +197,45 @@ ivy.tan()
...
             num_positional_args=num_positional_args,
             native_array_flags=native_array,
             fw=fw,
             frontend="torch",
-            fn_tree="tan",
-            input=np.asarray(x, dtype=input_dtype),
-            out=None,
+            fn_tree="nn.functional.threshold",
+            input=input[0],
+            threshold=0.5,
```
djl11, line 231: `threshold` and `value` should also be randomized by hypothesis.

djl11 (changes requested):

docs/partial_source/deep_dive/17_ivy_frontends_tests.rst
djl11: this has still not been resolved, it is still hard-coded to `True`?

djl11 (changes requested):

docs/partial_source/deep_dive/17_ivy_frontends_tests.rst
hirwa-nshuti: I have updated the tests I can update it

Daniel4078:
Maybe we should leave with_out to be still generated by the handle_cmd thing and fix the with_inplace argument by changing the test to not test for the with_out part when with_inplace is true.

docs/partial_source/deep_dive/17_ivy_frontends_tests.rst
djl11: yes that's what I meant, we should be randomly generating the value for `with_inplace` (which is the line of code this comment is pointing to), while the value for `with_out` should be hardcoded to `False`.

djl11 (approved):
lgtm!

djl11 (approved):

djl11 (approved):

PR #5047, opened 2022-09-28
Torch.cross Frontend  

nassimberrada:
PR for #4993


Ishticode (changes requested):
Mostly looks fine. But the test_torch_cross set with ivy's numpy backend as shown [here](https://github.com/unifyai/ivy/actions/runs/3146285000/jobs/5114558152#step:6:28242). Was it passing locally for you?

nassimberrada:
I made some changes to solve an issue with data consistency, i think the tests pass now

Ishticode (changes requested):
hmmm the `test_torch_cross` seems to fail for me when testing locally. Could you check that if possible 🙂

nassimberrada:
@Ishticode It's not passing either yes, and I'm facing an error that i don't know how to tackle, do you have an idea what might be the issue here ? 

self = <hypothesis.internal.conjecture.utils.many object at 0x7ff9100cf940>
data = ConjectureData(VALID, 13 bytes, frozen), min_size = 11424249584968187286
max_size = 11424249584968187286, average_size = 1.1424249584968188e+19

    def __init__(
        self,
        data: "ConjectureData",
        min_size: int,
        max_size: Union[int, float],
        average_size: Union[int, float],
    ) -> None:
>       assert 0 <= min_size <= average_size <= max_size
E       assert 1.1424249584968188e+19 <= 11424249584968187286

Also, there is one side of the testing process for which i am slightly uncomfortable, and it might be linked to the error maybe. I set the value of the dimension argument to be -1 (i.e the default value) because i don't know how to make the range of value vary depending on the array's shape, could you direct me to the part of the docs that address this kind of situation, or maybe give me a hint as to how i should proceed ?

Ishticode (changes requested):
Hi Nassim. From the face of it I can't tell what the cause of the error would be. Have you tried to debugger mode to track the problem....

If not sure then I would say to checkout just 2-3 vids of Dan fixing the tests in [this youtube playlist](https://www.youtube.com/playlist?list=PLwNuX3xB_tv-9cDyiDoInHgOpvQX4Tgq7)

This will give you some experience with tracking issues down. Let me know if still doesn't work out and then I will try to track it.

Also checkout the composite hypothesis strategies to see how to generate dependent values

PR #5048, opened 2022-09-28
Torch.ravel Frontend 

nassimberrada:


nassimberrada:
>           input=np.asarray(x, dtype=input_dtype),

ivy/functional/frontends/torch/miscellaneous_ops.py
```diff
@@ -154,3 +154,8 @@ def logcumsumexp(input, dim, *, out=None):
 
 def repeat_interleave(input, repeats, dim=None, *, output_size=None):
     return ivy.repeat(input, repeats, axis=dim)
+
+
+def ravel(input):
+    reshape_shape = (int(ivy.prod(ivy.array(tuple(input.shape)))), 1)
+    return ivy.array(input, dtype=input.dtype).reshape(reshape_shape)
```
Ishticode, line 161: I see what you're done here. But there is a quicker way:
`ivy.reshape(input, (-1,))` should just return the flattened array any way.

Ishticode (changes requested):

ivy/functional/frontends/torch/miscellaneous_ops.py
nassimberrada: Yes that does make it much cleaner ! 
I made that change, i think the tests pass now

nassimberrada:
**Status**: Tests Passed

Ishticode (approved):

nassimberrada:
There was a slight issue in the testing inputs indeed. I ran the tests locally several times and it passes now

Issue #4908, opened 2022-09-26
common functions


mattbarrett98:
I did a search of functions not in Ivy which are in at least 3 of jax, torch, numpy and tensorflow, and hence may be worth considering. I just searched by function name so some of these may exist in Ivy under a different name, but I tried to remove any such cases (sorry if i missed some). 

```{'vander', 'unravel_index', 'ldexp', 'nextafter', 'nanmean', 'trapz', 'fmod', 'atleast_3d', 'msort', 'histogram', 'nanmedian', 'triu_indices', 'broadcast_shapes', 'nanquantile', 'isclose', 'quantile', 'ravel', 'heaviside', 'imag', 'dstack', 'logaddexp2', 'float_power', 'corrcoef', 'tril_indices', 'fix', 'flipud', 'histogramdd', 'moveaxis', 'nansum', 'angle', 'diagflat', 'frexp', 'diff', 'fmax', 'fliplr', 'exp2', 'frombuffer', 'cov', 'vstack', 'fmin', 'nan_to_num', 'median', 'count_nonzero', 'isin', 'column_stack', 'diag', 'isneginf', 'hsplit', 'allclose', 'copysign', 'row_stack', 'i0', 'signbit', 'gradient', 'dsplit', 'hypot', 'atleast_1d', 'kron', 'vsplit', 'bincount', 'hstack', 'atleast_2d', 'rot90', 'isposinf', 'conj'}```

hirwa-nshuti:
Hi @mattbarrett98 
These functions should be added to Ivy's functional API. I added all of them to the  [Ivy Extension issue](https://github.com/unifyai/ivy/issues/3856) ToDo list. And I am going to close this issue. Thanks!

PR #5049, opened 2022-09-28
Torch.vander Frontend

nassimberrada:


nassimberrada:
>           x=np.asarray(x, dtype=input_dtype),

Ishticode (changes requested):
Depending on [this issue](https://github.com/unifyai/ivy/issues/4908), when the vander function is added to the Ivy API, we should be able to change the frontend to direct call to the `ivy.vander`.

Ishticode (changes requested):
I think this can move forward now and should be simple although there are some conflict. See if you can fix the easily. Let me know how it goes. Thnx.🙂

nassimberrada:
I pushed the implementation using a version that doesn't yet include ivy.vander because it is a bit lagging behind the master, would that make the tests on your side automatically fail ?

Ishticode:
> I pushed the implementation using a version that doesn't yet include ivy.vander because the branch i worked on is a bit lagging behind the master, would that make the tests on your side automatically fail ?

It would lag behind as I run your tests on your branch so its exactly the same as yours.    
I think if you sync your master fork it should able to update your branch by `git fetch and merge` then the new implementation should come in. Let me know if it doesn't work. If it causes any trouble, you can just create new branch from the synced in master and copy over the two updates and create a new pr.

nassimberrada:
Done, Should be ready to merge i think

Ishticode (changes requested):
Hi Nassim. Yh about ready to merge. 
But I tried test and it fails for numpy but when I change `N` min_value to 1 in the `@given` parametrisation of the test it passes.

Could you change the test OR see if/why N=0 might cause any issues. if it does we can handle N=0 as a special case. It's not a big deal.

Besides that just a reminder to fix the lints and then i will get this merged. Great work on this. Come a far way 🙂.

nassimberrada:
I tried to handle N=0 as a special case but somehow it started giving inconsistencies in data types, i figured i would just up the min value of N to 1 in the tests as you said, given that there's no much use in the function for a value of N=0 i guess.
I don't think there's any lint error left too, so unless you'd rather i try to crack down that special case then we could move forward with this.

Ishticode (changes requested):
Hmmm from the commit for handling N=0 case, I see that it has code

```python
def vander(x, N=None, increasing=False):
    if N == 0:
        return ivy.array([], dtype=input.dtype)
    else:
...
```

I guess you mistakenly put `input.dtype` instead of the `x.dtype`. Could you try that?

nassimberrada:
Commit name says it all, my bad

Ishticode (approved):
😂 it can happen. I will get this merged. Thank you very much @nassimberrada

PR #5051, opened 2022-09-28
fix test_frontend_method

juliagsy:


ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -1221,49 +1221,30 @@ def test_frontend_method(
...
-        input_dtypes=input_dtypes_method,
-        as_variable_flags=as_variable_flags_method,
-        native_array_flags=native_array_flags_method,
-    )
-    ins_gt = ins_class(*args_gt_constructor, **kwargs_gt_constructor)
-    ret_from_gt, ret_np_from_gt_flat = get_ret_and_flattened_np_array(
-        ins_gt.__getattribute__(method_name), *args_gt_method, **kwargs_gt_method
-    )
+    ivy.set_backend(frontend)
+    ins_gt = frontend_class(*args_constructor, **kwargs_constructor)
```
CatB1t, line 1235: One small note, otherwise lgtm! we should be converting `args` and `kwargs` to a native arrays, (ike here: https://github.com/unifyai/ivy/blob/4ee63edb384d2a37c87dbba5b46642f4d08f8b85/ivy_tests/test_ivy/helpers/function_testing.py#L600

I suppose NumPy doesn't throw an exception because all frameworks returns a np array because of the method `__array__`, but I'm assuming if we test a class method for a framework like torch, this will break. Can you do a brief testing locally?

CatB1t (changes requested):

PR #5054, opened 2022-09-29
Added eig() function for jax backend

AbhishekJha22:
Commit for issue #4825 (eig() function for jax backend).

### DIFF
--- a/ivy/functional/backends/jax/linear_algebra.py
+++ b/ivy/functional/backends/jax/linear_algebra.py
@@ -63,8 +63,10 @@ def diagonal(
         ret = jnp.diagonal(x, offset, axis1, axis2)
     return ret
 
+
 def eig(x: JaxArray, /, *, out: Optional[JaxArray] = None) -> JaxArray:
 	return jnp.linalg.eig(x)
+
 	
 eig.unsupported_dtypes = (
     "float16",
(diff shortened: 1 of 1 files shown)


ahmedo42:
HI @AbhishekJha22,  this function's correct path should `ivy/functional/frontends/jax/numpy/linalg.py` moreover you will have some difficulty implementing the function since `ivy.eig` is not implemented yet in the backends because Ivy doesn't yet support complex numbers. 

You are welcome to close this PR and move on to another function

AbhishekJha22:
Hi @ahmedo42 ! I apologize for the late response. I have been going through the documentation again to figure out the tasks better. I have one question about what you said in your comment -  what exactly is meant by ivy not supporting complex values yet? It is my understanding that ivy implements the backend functions using the implementation provided by the backend framework it is set to. So in this case, the jax support of complex numbers should be enough. Does ivy need to implement its own complex type that the jax complex numbers could be type cast into so that they could be used by the rest of the ivy program that might not use a jax backend?

ahmedo42:
Hi @AbhishekJha22, yes you are correct we need to implement our own complex data types and we have plans to implement this specific function in all backends [see this note](https://github.com/unifyai/ivy/blob/4b2007c71c80aaa3ac209e70626102e4afe7b754/ivy/functional/ivy/linear_algebra.py#L532)

AbhishekJha22:
Alright, thank you for your help! I will move on to a different function.

PR #5060, opened 2022-09-29
Jax.numpy.linalg.slogdet Frontend

nassimberrada:


nassimberrada:
>           a=np.asarray(x, dtype=input_dtype)

Ishticode (changes requested):
Mostly fine perhaps. According to [this](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.slogdet.html) There is an additional argument of method. Even if you don't make use of it in the call. This is just to ensure no arg mistmach between native jax and ivy's frontend jax. I Could explain this if it doesn't make sense. But essentially the frontend APIs need to have the same arguments as their corresponding frameworks.

Ishticode (changes requested):
The composite will be perhaps better to write it like so

```python
@st.composite
def square_shape(draw):
    shape = st.integers(min_value=2, max_value=20)
    return (shape, shape)
```

OR  maybe we don't even need a composite for this
we can just add shape like this below

```python
shape = st.tuples(st.shared(st.integers(1,20), key="sq"), st.shared(st.integers(1,20), key="sq"))
```
Feel free to let me know if you want me to explain this.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_linalg.py
```diff
@@ -298,3 +298,41 @@ def test_jax_numpy_cholesky(
...
+@handle_cmd_line_args
+@given(
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
+        max_value=1e4,
+        min_value=-1e4,
+        shape=square_shape(),
+        num_arrays=1,
+    ),
+    method=st.none(),
```
Ishticode, line 313: we dont really require this to be generated as it is passed as `None` manually.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_linalg.py
```diff
@@ -298,3 +298,41 @@ def test_jax_numpy_cholesky(
...
+    num_positional_args=helpers.num_positional_args(
+        fn_name="ivy.functional.frontends.jax.numpy.linalg.slogdet"
+    ),
+)
+def test_jax_slogdet(
+    dtype_and_x,
+    method,
```
Ishticode, line 320: can remove this

Ishticode (changes requested):
mthod in given and test parameters isn't required. but it should be passed to test_function as None as you already hv done. Let me know if this raises any issues.

nassimberrada:
Now it's my turn to have long testing times, even with tights input ranges that's weird. Could you try running this on your side ?

nassimberrada:
The tests pass under the condition that i remove the method argument from the testing altogether, even by just setting methods to None in the test_frontend_function is fails. I pushed a commit without the method argument in the tests

Ishticode (approved):
Yh I should have simply just said remove the method arg altogether. It is present in the frontend function just to match the api and avoid throwing errors (arg mismatch) when it was passed in the way it would in native `jnp`. It won't effect the output. 

Looks great. Will get this merged. Thanks Nassim.

PR #5064, opened 2022-09-29
PR for to_scalar function, issue #4812

MariaGamal:
I have a question.
I checked the function arguments for all current backends. All only had an array/tensor as input, except TensorFlow which had an array and a variable.
So, my question is should we add an `ivy.variable` as an argument in `to_scalar` function? 
I added a comment in the code for this.


MariaGamal:
There is no option for me to link my PR to the issue. Do you know why?

abdrahmandiab:
Hi @MariaGamal , could you comment the function # here?
Thanks!

ivy/container/general.py
```diff
@@ -2788,6 +2788,12 @@ def to_scalar(
             c: -1
         }
 
+        >>> print(ivy.is_int_dtype(y))
```
abdrahmandiab, line 2791: This is not necessary 😅 Just showing the output of `to_scalar()` itself should be sufficient!

Thanks!

abdrahmandiab:
> I have a question. I checked the function arguments for all current backends. All only had an array/tensor as input, except TensorFlow which had an array and a variable. So, my question is should we add an `ivy.variable` as an argument in `to_scalar` function? I added a comment in the code for this.

There are ongoing discussions about the handling of this problem within the team. But for now you may leave it as `ivy.array`, and we can change it in the future should we need to. 

Thanks for being so thorough!

abdrahmandiab:
Could you also please add the reformatting checklist and check it off as per the instructions [here](https://lets-unify.ai/ivy/contributing/4_open_tasks.html#formatting-checklist) once your PR is ready for review? Would appreciate that very much! 😃

MariaGamal:
Close #4812

MariaGamal:
How do I know which type of docstring examples is bringing the error? 
I get Assertion errors on the 4 frameworks, but how to know which example test fails? @abdrahmandiab 
I believe my function needs the first 4 types only mentioned [here,](https://lets-unify.ai/ivy/deep_dive/13_docstring_examples.html) which I made sure of.

MariaGamal:
add_reformatting_checklist_general

abdrahmandiab:
Hi @MariaGamal,

Great job! I ended up filling in most of the reformatting checklist myself just in the case someone looked at the PR in the future, but as you said there are only a few things that apply to this function anyway! 😄 

I ended up actually removing some of the examples and a few unnecessary lines, especially the examples with `NativeArray` since we decided they add clutter more than anything so we're removing them project wide. 😅

But all in all great job, will be merging soon so be on the lookout for that! 🚀 
Thanks!

MariaGamal:
Thank you a lot. @abdrahmandiab 
Is there anything else I should do now?

abdrahmandiab:
> Thank you a lot. @abdrahmandiab Is there anything else I should do now?

No no, I was just waiting for the CI checks to finish running to ensure we don't merge anything that causes failures, but otherwise everything is good! :)

PR #5073, opened 2022-09-29
Torch.combinations Frontend

nassimberrada:


nassimberrada:
**Error**: Operands could not be broadcast together with shapes (0,) (0,2)

PR #5076, opened 2022-09-29
Added examples for `instance` and `special` methods in ivy frontend docs

zaeemansari70:
Used the `add()` function as an example, as it's the most simple one and has all the methods implemented.


djl11 (approved):
looks great - thanks!

djl11 (approved):

PR #5079, opened 2022-09-29
refactor jax-nn-softmax frontend

ogbanugot:


ivy/functional/frontends/jax/nn/non_linear_activations.py
```diff
@@ -271,10 +271,13 @@ def soft_sign(x):
     return ret.astype(dtype)
 
 
-def softmax(x, /, *, axis=-1):
-    dtype = _type_conversion(x).dtype
+def softmax(x, axis=-1, where=None, initial=None):
     ret = ivy.softmax(x, axis=axis)
-    return ret.astype(dtype)
+    if initial:
```
juliagsy, line 276: `initial` can exists without `where` but `where` cannot exist without `initial`, so this logic should be reversed

if `where`:
    throw error if `initial` does not exists

ivy/functional/frontends/jax/nn/non_linear_activations.py
```diff
@@ -271,10 +271,13 @@ def soft_sign(x):
...
+        if ivy.is_array(where):
+            x = ivy.where(where, x, x.full_like(initial))
```
juliagsy, line 278: I've tried it out, for now the case when there is a combination of `where` and `initial` results in the error you pointed out, I'll have to dig in a bit more to know why, but you can fix the other first and maybe comment out the `where` argument and its related code before the issue is fixed :)

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_nn_activations.py
```diff
@@ -291,21 +292,35 @@ def test_jax_nn_one_hot(
         min_num_dims=1,
     ),
     axis=helpers.ints(min_value=-1, max_value=0),
```
juliagsy, line 294: there is a `dtype_values_axis` helper to generate dtype, x and its valid axis together (by toggling force_int_axis and valid_axis to True), might be an improvement to data generation

juliagsy (changes requested):
Hey! Thanks for the changes! I've left some comments below! Meanwhile, I'll look into the `where` argument related error and update you asap! Thank you!

ivy/functional/frontends/jax/nn/non_linear_activations.py
juliagsy: The native implementation reference is here - https://jax.readthedocs.io/en/latest/_modules/jax/_src/nn/functions.html#softmax

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_nn_activations.py
ogbanugot: Thanks!

juliagsy (commented):
Hey Ogban! Great work so far! There's nothing to change now! Would you like to attempt on the `initial` arg first? cuz this arg can exist without `where` but the opposite should not work. 

You can reference the implementation in the link I sent in the last review comment! 

You might want to merge this PR first and make new changes in a new PR to keep things tidy and avoid merge conflicts, but its up to you, anything you're more comfortable with! :)

Feel free to let me know if you're stuck somewhere when adding support for the `initial` arg, I'll try to help as far as possible :)

Thanks a lot!

ogbanugot:
Hey, thanks so much for the feedback! I'll merge this one and work on `initial` in a separate pr.

PR #5087, opened 2022-09-29
Added Superset functionality of seed to tensorflow random_uniform

ankitdsi2010:


saeedashrraf (approved):
Looks good!

saeedashrraf (changes requested):
It seems that the tests are failing for numpy because the seed value being generated is negative which is not accepted by the function

Issue #3938, opened 2022-09-01
top_k_categorical_accuracy


vandit98:
#1547

VedPatwardhan:
Hi @vandit98, please rename the issue according to the name of the function so that it gets linked to the ToDo list properly, i.e. the issue name shouldn't have spaces in it instead separated with underscore like `top_k_categorical_accuracy`. Thanks🙂

PR #5089, opened 2022-09-29
Create top_k_categorical_accuracy

hrak99:
- [ ] #3938

### DIFF
--- a/ivy/functional/frontends/jax/devicearray.py
+++ b/ivy/functional/frontends/jax/devicearray.py
@@ -62,3 +62,55 @@ class DeviceArray:
     @to_ivy_arrays_and_back
     def __abs__(self):
         return jax_frontend.abs(self.data)
+
+    @to_ivy_arrays_and_back
+    def __pow__(self, other):
+        return jax_frontend.pow(self.data, other)
+
+    @to_ivy_arrays_and_back
+    def __rpow__(self, other):
+        return jax_frontend.pow(other, self.data)
+
+    @to_ivy_arrays_and_back
+    def __and__(self, other):
+        return jax_frontend.bitwise_and(self.data, other)
+
+    @to_ivy_arrays_and_back
+    def __rand__(self, other):
+        return jax_frontend.bitwise_and(other, self.data)
+
+    @to_ivy_arrays_and_back
+    def __or__(self, other):
+        return jax_frontend.bitwise_or(self.data, other)
+
+    @to_ivy_arrays_and_back
+    def __ror__(self, other):
+        return jax_frontend.bitwise_or(other, self.data)
+
+    @to_ivy_arrays_and_back
+    def __xor__(self, other):
+        return jax_frontend.bitwise_xor(self.data, other)
+
+    @to_ivy_arrays_and_back
+    def __rxor__(self, other):
+        return jax_frontend.bitwise_xor(other, self.data)
+
+    @to_ivy_arrays_and_back
+    def __invert__(self):
+        return jax_frontend.bitwise_not(self.data)
+
+    @to_ivy_arrays_and_back
+    def __lshift__(self, other):
+        return jax_frontend.shift_left(self.data, other)
+
+    @to_ivy_arrays_and_back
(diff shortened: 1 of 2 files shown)


hrak99:
I'm applying the following code for top_k_categorical_accuracy in tensorflow,

def top_k_categorical_accuracy(y_true, y_pred, k=5):
    '''Calculates the top-k categorical accuracy rate, i.e. success when the
    target class is within the top-k predictions provided.
    '''
    return K.mean(K.in_top_k(y_pred, K.argmax(y_true, axis=-1), k))
but  there is no "_in_top_k" backend function is available. Kindly assist in this regard.

JamieLine:
Hi,

Thanks for the PR!

I'd love to help out, I just have a few quick questions before I can really help out properly.

1. Which file would this code go in? It makes it much easier for me to see this if you push your changes (even if they don't work yet) to your fork of the repository, then GitHub will let me see the changes in context and also run them locally.
2. What is `K` in the function definition? I can see that lowercase `k` is an integer argument, but I don't know what value uppercase `K` has.
3. There might just be no `in_top_k` in the Ivy API, in which case you'd need to find a way to replicate its functionality with a composition of Ivy functions, which could almost definitely be done by using `ivy.sort` and then indexing in from `0` to `k` or maybe there is a more efficient way to do it.

hrak99:
> Hi,
> 
> Thanks for the PR!
> 
> I'd love to help out, I just have a few quick questions before I can really help out properly.
> 
> 1. Which file would this code go in? It makes it much easier for me to see this if you push your changes (even if they don't work yet) to your fork of the repository, then GitHub will let me see the changes in context and also run them locally.
> 2. What is `K` in the function definition? I can see that lowercase `k` is an integer argument, but I don't know what value uppercase `K` has.
> 3. There might just be no `in_top_k` in the Ivy API, in which case you'd need to find a way to replicate its functionality with a composition of Ivy functions, which could almost definitely be done by using `ivy.sort` and then indexing in from `0` to `k` or maybe there is a more efficient way to do it.

Many thanks for your response.
I have pushed my partial work (just to share it) for your perusal. The files are used layers.py and test_layers.py.
K is used for tensorflow computation (can be just used as tf) while k is an integer in the query function (Number of top elements to look at for computing accuracy). 
And why the new commit (done through PyCharm) is not showing a verified tick?

JamieLine:
Thanks for pushing the code and explaining everything! To be honest I have no idea what leads to the verified notification, it's ultimately unimportant anyway. The tick however tells you whether or not the commit passed every test that GitHub ran on it. On the first commit there were no tests so it's declared to have passed all of them, the second however failed a linting check so it gets a cross. I've noticed a few things while looking through the commit and running it. 

1. You've included changes to a few files that are not part of the commit (i.e. the files in `.idea/`, the blank file `top_k_categorial_accuracy` and `ivy_tests/test_array_api`, these will need to be removed before the PR can be merged.
2. There are some formatting issues, admittedly they aren't particularly large issues for readability but we try to keep consistent code style across the repository, you can remedy these either by looking at the `lint/check-formatting` test that GitHub runs on PRs or you can install `pre-commit` as demonstrated in our documentation and it will run the checks locally, and try to fix the errors for you where it can.
3. The tests returned some errors, mostly either `ivy.exceptions.IvyError: numpy: argsort: axis 1 is out of bounds for array of dimension 1` or `tensorflow.python.framework.errors_impl.InvalidArgumentError: cannot compute InTopKV2 as input #0(zero-based) was expected to be a float tensor but is a int8 tensor [Op:InTopKV2]`. The second can be fixed by setting the test to only use floating point values, as that's all that Tensorflow supports for this function from the looks of things, and the first will require some re-writing to fix. You can run these tests locally using Pytest, which PyCharm has built in support for. There will be a green arrow next to the function definition, clicking it will run the test, but you might need to right click it first to set the working directory from `/ivy/` to wherever you have Ivy downloaded to.

hrak99:
Before push, the top_k_categorial_accuracy and ivy_tests/test_array_api files were disabled. Please let me know in case of existence of these files.

PR #5119, opened 2022-09-30
dirichlet extension

nassimberrada:


Ishticode (changes requested):
I guess this should be in  ivy's `random` submodule. What do you think?
`NativeArray` examples are to be omitted. `Container` examples to be moved to container methods and array instance examples to be moved to the equivalent array instance method. Please feel free to tag me if this does not make sense. Thanks 🙂

nassimberrada:
**Error** Same issue than torch.cross

nassimberrada:
@Ishticode I got a question for this function, which i guess is valid for all distribution functions. Given that the function is intended to randomly draw values given the underlying distribution, doesn't that mean that you will necessarily get different values when the test runs the same inputs on two back-ends, even if the function is correctly implemented in both ?

Ishticode (commented):
I don't think so that should be an issue... Maybe taking a look at the tests for random generating functions would be englightening.

If you get it fixed feel free to migrate and merge.

PR #5120, opened 2022-09-30
jax frontend special methods - pos, neg, abs, eq, ne, lt, le, gt, ge

juliagsy:
added special methods for ivy jax frontend DeviceArray, I've compared our ivy frontend returns with the native jax returns

just to confirm if this is the appropriate approach of testing the frontend special methods, feel free to let me know if anything should be changed or improved! 

thanks!


ivy/functional/frontends/jax/devicearray.py
```diff
@@ -1,18 +1,79 @@
...
+    @functools.wraps(f)
+    def decor(self, *args, **kwargs):
+        args = list(args)
+        if isinstance(self, jax_frontend.DeviceArray):
+            i = 0
+            for arg in args:
+                if isinstance(arg, jax_frontend.DeviceArray):
+                    args[i] = arg.data
+                i += 1
+            for u, v in kwargs:
```
djl11, line 20: iterating through a dict only gives its keys, you would need `kwargs.items()`

ivy/functional/frontends/jax/devicearray.py
```diff
@@ -1,18 +1,79 @@
+# global
+import functools
+
+
 # local
 import ivy
 import ivy.functional.frontends.jax as jax_frontend
 
 
+def _frontend_wrapper(f):
```
djl11, line 10: this should be handled in a fully recursive manner using something in the `ivy.nest` module. Could you also please add a comment explaining why this function is necessary at all?

djl11 (changes requested):
the tests look good, but I have requested some clarifications and one small bug fix.

ivy/functional/frontends/jax/devicearray.py
juliagsy: I think because it's a new class to resemble the native frontend behaviour (not ivy array nor native classes) it doesn't access the data automatically (since we have decorator or similar handling in the backend)

it seems to be trying to operate on the class object itself, therefore throwing various errors if the second onwards argument is the frontend type (first argument doesn't matter as we pass it as `self.data`)

so I've created a decorator to resolve the issue, but on second thought, we might need to add the frontend class type to the backend conversions as well so that they're recognised and handled properly (else, there may be certain class that we could inherit from..?)

ivy/functional/frontends/jax/devicearray.py
juliagsy: ah right, the very last commit for this fix was in my local, didn't know it's not pushed yet 😅 just pushed it, sorry!

PR #5124, opened 2022-09-30
Nll loss

### LINKED ISSUE #5083: nll_loss

mibo8:
Close #5083

### DIFF
--- a/ivy/functional/frontends/torch/Tensor.py
+++ b/ivy/functional/frontends/torch/Tensor.py
@@ -69,11 +69,11 @@ class Tensor:
     # Special Methoods #
     # -------------------#
 
-    def __add__(self, other):
-        return torch_frontend.add(self, other)
+    def __add__(self, other, *, alpha=1):
+        return torch_frontend.add(self, other, alpha=alpha)
 
-    def __radd__(self, other):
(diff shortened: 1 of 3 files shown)


ivy/functional/frontends/torch/loss_functions.py
```diff
@@ -100,4 +100,17 @@ def binary_cross_entropy(
...
-    return result
+    return 0
+
+
+def nll_loss(
+        input, target, weight=None, size_average=None, ignore_index=- 100, reduce=None, reduction='mean'
+):
+    input = ivy.exp(input)
+    depth = ivy.max(target) + 1
+    one_hot_target = ivy.one_hot(target, depth)
```
mibo8, line 111: I added one-hot encoding because torch.nll accepts indeces as targets. I think it should be done in torch frontend cross_entropy for the same reason

mibo8:
Hello @VedPatwardhan , I found some problems in computation of weighted nll and I'm working on it. No need to review my PR now. Thanks

mibo8:
Now it should be ok. Local tests return some type errors that seem unrelated to the function. Maybe it is related to my environment. I tested the function manually with all backends and its behaviour is identical to the torch version

PR #5133, opened 2022-10-01
#5130 Uint16

trihernandez:
- [ ] #5130


ivy/functional/frontends/jax/numpy/name_space_functions.py
```diff
@@ -115,4 +117,4 @@ def var(a, axis=None, dtype=None, out=None, ddof=0, keepdims=False, *, where=Non
     ret = ivy.var(a, axis=axis, correction=ddof, keepdims=keepdims, out=out)
     if ivy.is_array(where):
         ret = ivy.where(where, ret, ivy.default(out, ivy.zeros_like(ret)), out=out)
-    return ret.astype(dtype)
+    return ret.astype(dtype)
```
zaeemansari70, line 120: There should be a new line here.

ivy_tests/conftest.py
```diff
@@ -39,4 +39,4 @@ def pytest_collection_modifyitems(items):
         for id_ in skip_ids:
             if item.nodeid.startswith(id_):
                 item.add_marker(skip_ivy)
-                break
+                break
```
zaeemansari70, line 42: Here too, No change should be there in the `conftest.py` file.

zaeemansari70 (changes requested):
Hey! 
First of all thank you soo much for contributing to ivy! 🙂
Your PR looks amazing ✨!

1. It is giving an error for torch `Cannot convert to PyTorch dtype. uint16 is not supported by PyTorch.` 
2. There are a bunch of lint errors I've pointed out, they need to be fixed as well.
3. You've also committed a change in the conftest file please uncommit that change too.

Thanks!🙂

trihernandez:
I did the last two.  What would be the best way to interpret uint16 in pytorch (such as using an int32 datatype or two uint8 datatypes), or convert a value in pytorch to uint16 (even if this new data value is not using a pytorch datatype)?  Would I take the literal byte data of the uint16 and express that as the byte data for a signed 16-bit int?

trihernandez:
I did the last two. What would be the best way to interpret uint16 in pytorch (such as using an int32 datatype or two uint8 datatypes), or convert a value in pytorch to uint16 (even if this new data value is not using a pytorch datatype)? Would I take the literal byte data of the uint16 and express that as the byte data for a signed 16-bit int? How would I add a separate case for pytorch compared to the other data types?

When I ran the test cases on codespaces, my part of the code passed, so I do not know why my code is failing the test with this error now, especially since I worked on the jax frontend.

I read something that might be relevant to excluding pytorch here (https://lets-unify.ai/ivy/deep_dive/6_data_types.html#supported-and-unsupported-data-types), but it seems to be backend-related to it.

zaeemansari70 (changes requested):
Just one more small change before we can merge your PR 😊 

1. Please un-commit the `ivy_array.pickle` file.

Also, no need to worry about the dtype issue for Torch, it will surely be handled by the `Ivy` team in the future.
Thanks! 😊

trihernandez:
Deleted the pickle file.  You should be good to commit!

zaeemansari70 (approved):
Lgtm!

PR #5145, opened 2022-10-01
Adding conv_transpose1d and tests to torch frontends

CerberusLatrans:
#4826


CerberusLatrans:
note: trying to debug silently passing tests, can ignore

PR #5148, opened 2022-10-01
flatten extension

nassimberrada:


Ishticode (changes requested):
Hi @nassimberrada 
Lots of effort in here 🙂.
May I remind you that this flatten function would require a simpler implementation. 

For numpy:
`return x.flatten()`

for JAX
`return x.flatten()`

for torch
`return torch.flatten(x)`

for tensorflow
`return tf.reshape(x. -1)`

these should return the flattened array. So we might not need the function `_flatten`.

Let's see if you can simplify this

nassimberrada:
I have a question as far as this function is concerned @Ishticode . 
As far as i understand torch.flatten allows for the additional flexibility of flattening the array over a selected dimension range, as opposed to the other frameworks who just flatten the whole array. 
With that in mind, shouldn't ivy.flatten be constructed to allow for this behavior regardless of the backend framework chosen ? That's what motivated me to write the _flatten function, to mimic that functionality and be able to transfer it to other framework, is my reasoning incorrect here ?

Ishticode (commented):
@nassimberrada Ah I see. Sorry my bad. I will take another look shortly.

Ishticode (changes requested):
I guess this can be implemented as a container and array method too. Could you try that.

It will just be calls to the already implemented ivy.flatten... no new implementations really.
Checkout the [container](https://lets-unify.ai/ivy/deep_dive/5_containers.html) and [array](https://lets-unify.ai/ivy/deep_dive/4_arrays.html) pages if you like. Thanks.

nassimberrada:
@Ishticode Done, let me know if there is anything else that needs to be changed

Ishticode (changes requested):
Hi Nassim. The test fails for tensorflow and torch backends.... Is that something you've checked on

nassimberrada:
Hi Ishtiaq, i ran the tests locally several times and sure enough, after passing a couple times it started throwing the same error i keep getting in torch.cross. If you could take a look at it sometime that would really help, i'm starting to see this error more often than i'd like 😅.
Otherwise i added  static_flatten to the container.

Ishticode (changes requested):
somehow tensorflow backend has torch (I might have done that not sure).
Could you remove and run the test...
Then this can be merged.

nassimberrada:
Done. What about the testing inputs, do we keep them hard-encoded ?

nassimberrada:
I'm not sure but I don't think i'm seeing any torch in tensorflow's backend. 
Also what about the shape, start_dim and end_dim test inputs, do we keep them as is ?

Ishticode (changes requested):
I will get this merged asap. Could you just make sure to move the content to extensions.py folder and req review

Ishticode:
some conflicts but merged.

Ishticode:
Finally I will go ahead n merge. Thanks for bearing with me Nassim 🙂.

PR #5150, opened 2022-10-01
frontends.tensorflow.nn: add support for sigmoid_cross_entropy_with_logits

### LINKED ISSUE #5084: sigmoid_cross_entropy_with_logits

hmahmood24:
Close #5084

### DIFF
--- a/ivy/assertions.py
+++ b/ivy/assertions.py
@@ -120,13 +120,14 @@ def check_all_or_any_fn(
 def check_shape(x1, x2, message=""):
     x1_shape = ivy.shape(x1)
     x2_shape = ivy.shape(x2)
-    message = message if message != "" else "{} and {} must have the same shape ({} vs {})".format(
-        x1,
-        x2,
-        x1_shape,
-        x2_shape
+    message = (
+        message
+        if message != ""
+        else "{} and {} must have the same shape ({} vs {})".format(
+            x1, x2, x1_shape, x2_shape
+        )
     )
-    if (x1_shape != x2_shape):
+    if x1_shape != x2_shape:
         raise ivy.exceptions.IvyException(message)
 
 
--- a/ivy/assertions.py
+++ b/ivy/assertions.py
@@ -118,16 +118,14 @@ def check_all_or_any_fn(
 
 
 def check_shape(x1, x2, message=""):
-    x1_shape = ivy.shape(x1)
-    x2_shape = ivy.shape(x2)
     message = (
         message
         if message != ""
         else "{} and {} must have the same shape ({} vs {})".format(
-            x1, x2, x1_shape, x2_shape
+            x1, x2, ivy.shape(x1), ivy.shape(x2)
         )
     )
-    if x1_shape != x2_shape:
+    if ivy.shape(x1) != ivy.shape(x2):
         raise ivy.exceptions.IvyException(message)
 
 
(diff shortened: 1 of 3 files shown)


JamieLine:
Hi,

Thanks for the PR! I've noticed a few small things that will help to get this PR ready to be merged.

1. Right now, the PR fails GitHub's linting checks, you can see them on this page by checking the `lint/check-formatting` test, or by running `flake8` and `black` locally. The docs have a section on installing `pre-commit`, found here. https://lets-unify.ai/ivy/contributing/0_setting_up.html#pre-commit Installing `pre-commit` will ensure that these checks are run before code can be committed to the repository. It will need to pass both the `lint/check-formatting` test and the `test-frontend-tensorflow/run-nightly-tests` checks to be merged.
2. Your test crashes, with an error `TypeError: sigmoid_cross_entropy_with_logits() got an unexpected keyword argument 'fw'`. You can simply remove the `fw` parameter from the body of the test, then the test should run fine. Leave it as a parameter for the test itself, but ignore it. It should look like this. You can test that this works by running the test locally using `pytest`.

```
def test_tensorflow_sigmoid_cross_entropy_with_logits(..., fw, ...):
      some_code_which_does_not_reference_fw(...)
```
3. Your test specifies a strategy for generating `as_variable` and `native_array`. This is perfectly reasonable but isn't needed, it will be handled for you in `@handle_cmd_line_args` and should be removed from the test to ensure that as those "hidden" strategies are updated the changes take place everywhere.
4. In the test, you should also generate a value for the `name` parameter. It doesn't affect input/output behaviour but it is good practice and a requirement of Ivy to have all arguments tested where possible.

The rest of this PR however seems good, and once the test isn't crashing I'll be able to have a better look into how well the function runs and behaves.

If you have any questions or comments, feel free to let me know!

hmahmood24:
Hi,

Thanks a lot for the prompt and detailed response. I have looked into the points you raised and I have made the asked changes as per the requirements. The lint/formatting issues have resolved now. As for the nightly tests, they are still running. I'll wait a while and check back to look at the logs for those tests and see if there's any problem. Cheers.

hmahmood24:
Hi @JamieLine,
I have gone through the logs for the nightly tests and the tests seem to fail because of the following exception:

`hypothesis.errors.Flaky: Inconsistent data generation! Data generation behaved differently between different runs. Is your data generation depending on external state?`

But I can't seem to find the problematic logic inside my code that would be throwing this exception. Can you please help me out in debugging and fixing this? Thanks!

JamieLine:
Hi!

Flaky is the most annoying error Hypothesis can throw, I'll look into working out why that is happening because it can have some very strange causes. But in good news, it is usually caused by the test rather than the function being tested so there's a good chance that once the Flaky error goes away it'll all pass.

The only other thing I would suggest changing is in `check_shape`. Consider removing the `x1_shape` and `x2_shape` variables and replacing their usages with `x1.shape` or `x2.shape` as appropriate. This just to make the code a bit more concise without losing clarity. It's not required, once the tests pass if this isn't done I'll merge in anyway because it is fine as it is, it just appears to be a little bit unnecessary to me.

If you have any questions or comments, feel free to let me know!

hmahmood24:
Hi!

Thanks again for your help. The Flaky error was resolved after removing the strategy for name generation like you suggested, thankfully. I have also gone through the logs of the nightly tests and the test passes now without any exceptions. 😊 It is still a little difficult to understand why the name generation strategy was causing that exception though. Anyways, please take a look at the commits and let me know if there is anything else. Cheers!

JamieLine:
Hi! Sorry I've been delayed, I moved back up to Uni this week and I've been trying to quickly catch-up, I'll look through it all now and get back to you.

PR #5151, opened 2022-10-02
ivy.collapse_repeated

Change-m:


Change-m:
Close #5152

Change-m:
Collapse_repeated function is belong to Neural Network which is wildly used in Tensorflow, implemented compositionally that's why it is in extension.py file.

ivy/functional/ivy/extensions.py
```diff
@@ -453,3 +453,43 @@ def sinc(
...
     """
     return ivy.current_backend(x).sinc(x, out=out)
+
+
+@integer_arrays_to_float
+@to_native_arrays_and_back
+@handle_out_argument
+@handle_nestable
+@handle_exceptions
+def collapse_repeated(
```
zaeemansari70, line 463: Only tensorflow has this function, it has to be implemented compositionally using other Ivy functions.

zaeemansari70 (changes requested):
Your PR looks amazing ✨!

You have to come up with a logic to implement this function using ivy functions as implemented here, as only tensorflow has it and therefore it is to be implemented compositionally, https://lets-unify.ai/ivy/deep_dive/1_function_types.html#compositional-functions 🙂

Do let me know if you have any questions!
Thanks!

PR #5175, opened 2022-10-02
ivy.avg pool1d

### LINKED ISSUE #5174: ivy.avg_pool1d

Change-m:
Close #5174


zaeemansari70 (changes requested):
Your PR looks amazing ✨!
Either you can un-commit the unwanted files which is `test_layers` in this case or you can also make a new PR or also implement in all the backends and commit those changes too.
Also, if you are just adding the docstrings and type-hints, they need to be added in the array and container instance methods as well.
Thanks!

PR #5177, opened 2022-10-02
numpy frontend casting

juliagsy:


ivy_tests/test_ivy/test_frontends/test_numpy/helpers.py
```diff
@@ -147,3 +152,35 @@ def handle_where_and_array_bools(where, input_dtype, as_variable, native_array):
         input_dtype += ["bool"]
         return where, as_variable + [False], native_array + [False]
     return where, as_variable, native_array
+
+
+def handle_dtype_and_casting(
```
djl11, line 157: I think this should itself be a `st.composite` function right? The problem here is that hypothesis has consumed some of it's search space budget generating a variety of data types for us already, and then many of these generated data types end up not being used because of the independently generated casting mode.

It would be better if all of this logic was handled inside a hypothesis generation function, which would then correctly capture the inter-dependency between the data types and the casting mode, rather than generating them independently.

The current approach will generate both independently, and then remove generated data types based on the casting, which wastes hypothesis search space generating data types which end up being invalid for the casting mode.

Does this make sense? Perhaps I am missing something am my assessment is wrong.

ivy/functional/frontends/numpy/func_wrapper.py
```diff
@@ -0,0 +1,95 @@
...
+            """
+            ivy.assertions.check_elem_in_list(
+                casting,
+                ["no", "equiv", "safe", "same_kind", "unsafe"],
+                message="casting must be one of [no, equiv, safe, same_kind, unsafe]",
+            )
+            args = list(args)
+            if casting == "no" or casting == "equiv":
+                _assert_args_and_fn(
+                    args[:num_args],
```
djl11, line 56: this will not work for functions like `numpy.concatenate`, in which the first argument is a list of arrays. You will need a more general approach which uses the `ivy.nest` module to recursively find the indices of all arrays in the input `*args` and `**kwargs`. something like:
```
arg_arrays = ivy.multi_index_nest(ivy.nested_argwhere(args, ivy.is_ivy_array), args)
kwarg_arrays = ivy.multi_index_nest(ivy.nested_argwhere(kwargs, ivy.is_ivy_array), kwargs)
arrays = arg_arrays + kwarg_arrays
```

ivy/functional/frontends/numpy/func_wrapper.py
```diff
@@ -0,0 +1,95 @@
...
+def _assert_args_and_fn(args, dtype, fn):
+    ivy.assertions.check_all_or_any_fn(
+        *args,
+        fn=fn,
+        type="all",
+        message="type of input is incompatible with dtype: {}".format(dtype),
+    )
+
+
+def handle_numpy_casting(num_args: int = 1) -> Callable:
```
djl11, line 29: see comment below, passing in `num_args` should not be necessary once we enable dynamic inference of the number of arrays and their location in the input arguments and keyword arguments. See [here](https://github.com/unifyai/ivy/blob/e85a3f2b92b8768ad0c337937b4149f1385f4de6/ivy_tests/test_ivy/helpers/function_testing.py#L1645) for more examples of the nested functions in action.

djl11 (changes requested):

juliagsy (commented):
Hey Dan! Thanks for the review! 

Made me realised I haven't been using `ivy.nest` well so took a deeper dive in this! Hope I'm doing it right or in a more appropriate way now! 

While refactoring the decorator, I realised that there is a special case in `numpy` where the return array is type `bool`, therefore if the `dtype` parameter is specified, it must be a `bool` type and nothing else. Thus, I made a special decorator for this case. Feel free to let me know if there is a better or more elegant way to handle this!

For the hypothesis testing, I'll have to explore more in that part of hypothesis before refactoring and updating the function. I'll keep it like that for now and I'll make further updates or changes in the next PR (as I have more numpy frontend functions to add the casting decorator to)

Feel free to ask any questions or clarifications! Thank you!

djl11 (approved):
lgtm! Yes adding the `st.composite` hypothesis in the next PR sounds good :)

PR #5183, opened 2022-10-03
hann_window extension

nassimberrada:


ivy/functional/backends/torch/statistical.py
```diff
@@ -293,3 +293,24 @@ def einsum(
     out: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
     return torch.einsum(equation, *operands)
+
+
+def hann_window(
```
Ishticode, line 298: It seems you have implemented for torch already in statistical so can safely remove this

Ishticode (changes requested):
Very good PR.

Some comments were add in certain places. Additionally, as the window_length is a type hint, it maybe appropriate to generate using st.integers() or helper.ints functions and dtype can be separately generated by helper.get_dtype I think. Let me know if any of these doesn't make sense or If i am missing something. Thanks Nassim and sorry for the delay in this review. 

There should be no array method as you have correctly left out. 
For Container methods I need to check few things so leave that out for now.If test has any issues with the decorator @handle_nestable  you can remove it.

Ishticode (changes requested):
So yh. We should add container instance and static method. leaving @handle_nestable on.
Could you ensure lint is fixed and the tests are passing. Mostly seems good here.

ivy/functional/backends/torch/extensions.py
```diff
@@ -121,3 +121,24 @@ def lcm(
...
+        periodic=periodic, 
+        dtype=dtype, 
+        layout=torch.strided,
+        device=None,
+        requires_grad=None
+    )
+
+
+hann_window.support_native_out = False
+hann_window.unsupported_dtypes = ("int32", "int64")
```
Ishticode, line 144: It seems this part about unsupported dtypes is not required. Tests pass for me. Not sure if you had something on mind here.

ivy/container/extensions.py
```diff
@@ -333,3 +333,105 @@ def lcm(
...
+        -------
+        ret
+            The container containing the Hann windows.
+
+        Examples
+        --------
+        With one :class:`ivy.Container` input:
+
+        >>> x = ivy.Container(a=3, b=5)
+        >>> ivy.hann_window(x)
```
Ishticode, line 427: should be an instance method call here.   about the square brackets in the result printed below (and above), not sure where they came from... ususally container is printed with just the curly brackets

ivy/container/extensions.py
```diff
@@ -333,3 +333,105 @@ def lcm(
...
+        -------
+        ret
+            The container that contains the Hann windows.
+
+        Examples
+        --------
+        With one :class:`ivy.Container` input:
+
+        >>> x = ivy.Container(a=3, b=5)
+        >>> ivy.hann_window(x)
```
Ishticode, line 376: i think this should be `ivy.Container.static_hann`

ivy/functional/ivy/extensions.py
```diff
@@ -668,3 +668,48 @@ def lcm(
...
+        optional output array, for writing the result to.
+
+    Returns
+    -------
+    ret
+        The array containing the window.
+
+    Functional Examples
+    -------------------
+
```
Ishticode, line 706: no gap here. Sorry for being picky. The numpy style docstring test gets angry otherwise.

Ishticode (changes requested):
near ready to merge. Just minor few changes

nassimberrada:
I'm gonna have to thank you for your patience, i keep making the same typos, but i'll learn eventually...hopefully 😀

Ishticode (approved):
Thanks @nassimberrada. Should be merged now.✌️

nassimberrada:
close #5541

PR #5195, opened 2022-10-03
superset behaviour for argmin

sharique1006:
The tests are passing for jax, numpy and tensorflow. For torch, `ret.type(output_dtype)` and `ret.to(output_dtype)` are not working. How can I change the `dtype` of `ret` to `output_dtype`?

### DIFF
--- a/.github/workflows/test-array-api.yml
+++ b/.github/workflows/test-array-api.yml
@@ -26,6 +26,7 @@ jobs:
           submodules: 'recursive'
 
       - name: Download artifact
+        if: github.event_name == 'pull_request'
         uses: dawidd6/action-download-artifact@v2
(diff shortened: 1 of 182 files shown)


sherry30:
That's because ivy dtype is not being converted to native dtype before passing to the backed function, 
You can add `@infer_dtype` decorator on `argmin` and that will fix it.
You should also change the name of the argument to `dtype` otherwise it won't work with `infer_dtype` decorator

sherry30 (changes requested):
You should add `@infer_dtype` in functional ivy API not on a backend specific implementation.
No backend should receive ivy.Dtype.

sharique1006:
made the required changes

PR #5216, opened 2022-10-03
added argmax to jax frontend 

hassan-204:


ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_namespace_functions.py
```diff
@@ -68,6 +68,41 @@ def test_jax_numpy_absolute(
...
 
 
+# argmax
+@handle_cmd_line_args
+@given(
+    dtype_and_x=helpers.dtype_values_axis(
+        available_dtypes=helpers.get_dtypes("numeric"),
+        force_int_axis=True,
+        min_num_dims=1,
+        min_axis=-1,
```
CatB1t, line 78: Is there any specific reason we need to limit the axis? If not, use `valid_axis=True` instead.

CatB1t (changes requested):

CatB1t (changes requested):
we should be also testing for `keepdims`, generating a boolean value for it using `st.booleans()`.

ivy/functional/frontends/jax/numpy/name_space_functions.py
```diff
@@ -18,6 +18,10 @@ def all(a, axis=None, out=None, keepdims=False, *, where=False):
     return ivy.all(a, axis=axis, keepdims=keepdims, out=out)
 
 
+def argmax(a, axis=None, keepdims=None):
```
CatB1t, line 21: the function signature should match the one in JAX, [here](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.argmax.html#jax.numpy.argmax). an `out` parameter is missing.

CatB1t (changes requested):

CatB1t:
lgtm! thanks for contributing!

PR #5228, opened 2022-10-03
frontend testing `with_out` support for tuple

CatB1t:
### Few comments
* torch have a lot of namedtuple for return, should I refactor the logic inside min to be a decorator for torch frontend?
* torch will return `torch.return_types.min(tensor,tensor)`, but I've made it to return `min(tensor, tensor)`, should i update it to replicate the same name of the torch namedtuple?

### DIFF
--- a/ivy/functional/frontends/torch/reduction_ops.py
+++ b/ivy/functional/frontends/torch/reduction_ops.py
@@ -57,7 +57,7 @@ def var(input, dim, unbiased, keepdim=False, *, out=None):
 def min(input, dim=None, keepdim=False, *, out=None):
     if dim is None:
         return ivy.min(input, axis=dim, keepdims=keepdim, out=out)
-    elif out:
+    elif out is not None:
         ivy.min(input, axis=dim, keepdims=keepdim, out=out[0])
         ivy.argmin(input, axis=dim, keepdims=keepdim, out=out[1])
         return out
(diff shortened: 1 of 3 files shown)


ivy/functional/frontends/torch/reduction_ops.py
```diff
@@ -51,3 +52,20 @@ def prod(input, dim=None, keepdim=False, *, dtype=None, out=None):
 
 def var(input, dim, unbiased, keepdim=False, *, out=None):
     return ivy.var(input, axis=dim, correction=int(unbiased), keepdims=keepdim, out=out)
+
+
+def min(input, dim=None, keepdim=False, *, out=None):
+    if dim is None:
+        ret = ivy.min(input, axis=dim, keepdims=keepdim, out=out)
+    else:
```
djl11, line 60: this can be changed to
```
elif out is not None:
    ...
else:
    ...
No need for the double indentation. Also some arrays such as `array(False)` might boolean evaluate to `False` so it's best to make `is not None` explicit.

File: ivy/functional/frontends/torch/reduction_ops.py [hunk 2]
```diff
@@ -51,3 +52,20 @@ def prod(input, dim=None, keepdim=False, *, dtype=None, out=None):
 
 def var(input, dim, unbiased, keepdim=False, *, out=None):
     return ivy.var(input, axis=dim, correction=int(unbiased), keepdims=keepdim, out=out)
+
+
+def min(input, dim=None, keepdim=False, *, out=None):
+    if dim is None:
+        ret = ivy.min(input, axis=dim, keepdims=keepdim, out=out)
```
djl11, line 59: we can simply `return` here, and in fact we can do this inside each of the conditional blocks. No need to wait until the end of the function to return things. This in general makes it easier to read IMO, you know it's no longer necessary to continue following the logic further down the page once you see the `return` command in a certain conditional block.

djl11 (changes requested):
very small changes suggested, but otherwise looks great!

PR #5236, opened 2022-10-04
Tests for Array class magic methods

hirwa-nshuti:


ivy/array/array.py
```diff
@@ -689,7 +689,7 @@ def __rrshift__(self, other):
         --------
         >>> a = 32
         >>> b = ivy.array([0, 1, 2])
-        >>> y = a >> b
+        >>> y = b.__rrshift__(a)
```
djl11, line 692: Check the [doc](https://lets-unify.ai/ivy/deep_dive/13_docstring_examples.html), we don't need to add examples with the special method called explicitly. This will already trigger the reverse example, as the integer `__rshift__` special method will not have a rule for dealing with `ivy.Array` instances, and so the `__rrshift__` method will be called on `b` (the `ivy.Array`) anyway when calling `a >> b`

ivy/array/array.py
```diff
@@ -701,16 +701,7 @@ def __irshift__(self, other):
...
-            return to_ivy(self._data.__deepcopy__(memodict))
-        except AttributeError:
-            # ToDo: try and find more elegant solution to jax inability to
-            #  deepcopy device arrays
-            if ivy.current_backend_str() == "jax":
-                np_array = copy.deepcopy(self._data)
-                jax_array = ivy.array(np_array)
-                return to_ivy(jax_array)
-            return to_ivy(copy.deepcopy(self._data))
+        return to_ivy(self._data.__deepcopy__(memodict))
```
djl11, line 704: keep the old `try` `except` stuff here, this error might only arise when using Device arrays on the GPU for example. Are you sure the error never occurs for JAX?

djl11 (changes requested):
very small changes requested, otherwise the changes look great! Provided all the tests are passing, I'm happy for this to be merged once the small comments are addressed :)

ivy/array/array.py
hirwa-nshuti: Sure I can revert it

ivy/array/array.py
hirwa-nshuti: I thought there was a fix as I went through Jax Repository and found out that they solved the issue here [Deepcopy issue](https://github.com/google/jax/issues/2632), [Pull Request](https://github.com/google/jax/pull/10659). But if it is for GPU case I can revert the changes

hirwa-nshuti: I also tried it on google Colab running on GPU and seemed to work

hirwa-nshuti: As `bfloat16` is not still supported we remain with this exception handling.

PR #5239, opened 2022-10-04
Kaiser_window extension

nassimberrada:


Ishticode (changes requested):
@nassimberrada Great Addition
Are the tests passing for all backends?
I've checked this example `ivy.kaiser_window(5, beta=1.0)` and it gives different results for when ivy torch backend is set than when any other backend is set. Could you check why there is inconsistent result?

nassimberrada:
I had some issues with data types that i suspect to be the root cause of the issue. Introducing the dtype argument explicitly solved it and the tests seem to be passing on my side, could you check again ?

Ishticode (changes requested):
Hi @nassimberrada 
Looks mostly good. 
Currently throws an error because we have the `@handle_nestable` but we don't have the container methods (instance or static). Could you see to this issue. Let me know if you require more details from me

nassimberrada:
@Ishticode in this case should i just remove the handle_nestable property given that the function takes integers as inputs and not arrays per se, or do i go ahead and add the containers methods as well ?

Ishticode:
Take `@handle_nestable` for now. I think if its required it can be added later.

nassimberrada:
Done

nassimberrada:
I issued the PR for the vorbis_window yesterday and somehow i was assigned to someone else, they suggested i should move the implementation to extensions.py but didn't point out the handle_nestable argument, that's pretty much all i know about vorbis_window.
I made the necessary lint changes for both the kaiser_window and vorbis_window, hopefully it's clear.

nassimberrada:
No worries, i was expecting conflicts by creating multiple PRs. Should be merged now

Ishticode:
Are the tests running for you. For some reason they dont even start for when using docker.

nassimberrada:
They are on my side, maybe it's caused by the conflicts ? I'll try to solve those and you could try again

PR #5242, opened 2022-10-04
Kaiser_bessel_window extension

nassimberrada:


Ishticode (changes requested):
@nassimberrada Could the function naming match. in the backends the function is called `kaiser_bessel_window` while in ivy/functional its called `kaiser_bessel`. The naming should be the same to allow the updating in `ivy.__dict__` when we set a new backend. This repalces the functions from ivy functional with backend functionals.

nassimberrada:
Whoops my bad, that's what happens when you try to take shortcuts. Corrected

Ishticode (changes requested):
Good PR few things.

Firstly let me ask if this is the formula being used:
image

from wikipedia, linked from tensorflow docs?
If yes then it seems mostly fine. I had noticed a minor diff so just wanted to ensure we were looking at the same thing.

Secondly, shouldn't the function be named kaiser_bessel_derived_window? i know its a bit long. We can change it later if it feels appropriate to. But currently lets keep it as tensorflow has it and also lets follow the name suggested in API extension todo list.

Thirdly, can you think of a way that we have to write _KBDW only once? Let me know without implementing so u don't have to redo things. 

Much appreciated Nassim. Hope we can move forward with this soon. Thanks.

Ishticode:
> 1- That's exactly the function i took as basis for the implementation of _KBDW yes 2- I went for the name that was placed in the to-do list of function extensions tbh 3- The only meaningful difference between the different _KBDW functions is the initialization of the weights (w) which depends on the kaiser function and require calling a different back-end each time. I guess a possible fix would be to implement ivy.kaiser first then once it is merged, to write ivy.kaiser_bessel as a mixed function just like you did with ivy.flatten to avoid repeating the same code multiple times ?

1. Perfect. 2. i think the name in the list is ` kaiser_bessel_derived_window` not sure if it was edited later or something. 3. That sounds about right. Lets push through kaiser soon then... and see if we can make use of that.

PR #5247, opened 2022-10-04
Cupy backend

vaatsalya123:
Implemented cuda python backend and added a bit of testing.


JamieLine:
Hi!

Hope you're doing well, just wanted to point out that the `lint/check-formatting` test is failing on this page, you might want to have a look at these. I've had a glance, some of them I think will need you to put in a comment to disable the check for some files but some of them will need to be fixed. You can use `pre-commit` to make fixing it effectively automatic.

PR #5252, opened 2022-10-04
Update test_ndarray.py

varshithamanasa:


varshithamanasa:
could you tell me where I am going wrong

RashulChutani (changes requested):
Hi @varshithamanasa, Can you add the reference to the issue with respect to which the PR has been made? Thanks.

PR #5256, opened 2022-10-04
Fixes task #5245 to add ReLU activation frontend tensorflow

suvadityamuk:
#5245

### DIFF
--- a/.github/workflows/test-frontend-numpy-push.yml
+++ b/.github/workflows/test-frontend-numpy-push.yml
@@ -1,0 +1,63 @@
+name: test-frontend-numpy-push
+on: push
+permissions:
+  actions: read
+jobs:
+  run-nightly-tests:
+    strategy:
+      matrix:
+        backends: [numpy, torch, jax, tensorflow ]
+        submodules: [creation_routines, fft, indexing_routines, linear_algebra, logic, ma,
+                    manipulation_routines, mathematical_functions, matrix, ndarray, random,
+                    sorting_searching_counting, statistics, ufunc]
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️Ivy
+        uses: actions/checkout@v2
+        with:
+          path: ivy
+          persist-credentials: false
+          submodules: "recursive"
+          fetch-depth: 2
+
+      - name: Check Files Changed
+        shell: pwsh
+        id: check_file_changed
+        run: |
+          cd ivy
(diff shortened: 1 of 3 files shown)


ivy/functional/frontends/tensorflow/activations.py
```diff
@@ -12,3 +12,8 @@ def hard_sigmoid(x):
 
 def linear(x):
     return x
+
+def relu(x):
```
jkeane508, line 16: I think there might be a cleaner way to do this, could we just not set the max value in `ivy.clip` to `ivy.inf`?

jkeane508 (changes requested):
Just a quick thought about your implementation. Otherwise looks good :)

ivy/functional/frontends/tensorflow/activations.py
suvadityamuk: Also, it seems that `x_max` is a required parameter on `ivy.clip`

jkeane508: Yes but Relu essentially just keeps all positive values and set negative values to 0. So we could possibly try setting the x_max value to just ivy.inf?

jkeane508 (changes requested):

ivy/functional/frontends/tensorflow/activations.py
jkeane508: It wouldn't change the logic but it would just condense the implementation :)

jkeane508: Hi, so Relu basically turns any negative values to 0 otherwise it just keeps them. So we could just set the x_max parameter in ivy.clip to ivy.inf.

jkeane508: It would just reduce the implementation to one line of code:
`x = ivy.clip(x, 0.0, ivy.inf)`

jkeane508: Hopefully that makes sense :)

suvadityamuk: Yup, sounds good. I'll make the change by tomorrow, afk for the day sadly 😅

suvadityamuk: Done, please check now

suvadityamuk: Hi, the histories of this change was too divergent and showing merge conflicts, so I took the liberty of shifting to a new PR with merged changes at #5338

PR #5286, opened 2022-10-05
Add new_ones instance method to PyTorch frontend

### LINKED ISSUE #5265: new_ones

EricSchuMa:
Close #5265


ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py
```diff
@@ -54,6 +55,58 @@ def test_torch_instance_add(
...
+@given(
+    dtype_and_x=helpers.dtype_and_values(available_dtypes=helpers.get_dtypes("float")),
+    shape=helpers.get_shape(
+        allow_none=False,
+        min_num_dims=1,
+        max_num_dims=5,
+        min_dim_size=1,
+        max_dim_size=10,
+    ),
+    dtypes=_dtypes(),
```
EricSchuMa, line 72: I reused the st.composite functions _dtypes and _requires_grad from test_creation_ops.py as the test is very similar to test_creation_ops.test_torch_ones().

jeandamascyifuzo: Squashing your commits, it's too much

abdrahmandiab: Sure! We squash all PRs before merging anyway :)

ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py
```diff
@@ -5,6 +5,8 @@
 # local
 import ivy_tests.test_ivy.helpers as helpers
 from ivy_tests.test_ivy.helpers import handle_cmd_line_args
+from ivy_tests.test_ivy.test_frontends.test_torch.test_creation_ops \
+    import _dtypes, _requires_grad
```
abdrahmandiab, line 9: Seeing as these functions are quite small (basically 5 lines each), I think it would be easier if we just add to this file rather than importing them. 

I believe this would make the code more easy to understand if someone tries to do that! 😅

abdrahmandiab (changes requested):
Hi @EricSchuMa, 
Everything looks good to me with the exception of one little nit-pick that I've left a comment on below. 😅

Hope you don't mind making this tiny change!
Otherwise, great job so far, and thanks for your hard work! 🚀

abdrahmandiab:
Another update... 😅

I am unsure why yet but it seems the tests are failing when you run them locally because the function expects the arguments `dtype` to be `torch.dtype` and not `IntDtype` or `FloatDtype` which are the Ivy defined dtypes. I believe this is from our side and does not have to do with your implementation, so you don't need to do anything really, though you are welcome to help with the debugging if you feel like it, and let me know if you find something interesting in this case. 😄

I will try to solve the problem by tomorrow, and if not I will just merge the PR and handle things from there so as not to delay you! 😅

Thanks for your patience, and sorry for the inconvenience!

abdrahmandiab:
Hi @EricSchuMa , it seems there is a bug with our testing code that we'll handle from our side! 😅 
In the meantime, I've merged the PR so you can move on with your application process.

Thank you so much for your work and your time! 😄

PR #5289, opened 2022-10-05
Updated layer_norm behaviour to include elementwise scales and offset…

aditya-pola:
…s. Superset beahviour for torch is now satisfied (weight and bias arguments added).


VedPatwardhan (changes requested):
Hi @aditya-pola, please also update the array and container instance methods for the `layer_norm` function and also the `test_layer_norm` function in `ivy_tests/test_ivy/test_nn/test_norms.py` according to the new parameters. Thanks 🙂

aditya-pola:
oops, let me quickly do that

PR #5296, opened 2022-10-05
jax array-api tests 

simonetgordon:
Hi Matt, ignore the linspace bit, but the way I've implemented the methods `__dlpack__` and `__dlpack_device__` is a bit of a dummy implementation and I'm not sure how to implement them properly. They are passing the `test_signatures` and `test_names` array-api tests, but if you try to use these instance methods on Jax DeviceArrays, no attribute is found. Tensorflow has the same issue. These tests only look at superficial things so I guess it makes sense.

Both of these frameworks don't have this attribute defined themselves. I'm just not sure how to go about implementing them. Can see discussion of someone running into this issue [here](https://github.com/google/jax/discussions/11494) and the [python dlpack spec](https://dmlc.github.io/dlpack/latest/python_spec.html).

I think the goal is to have the following work:
```
x=jax.numpy.array([2,3,4])
print(numpy.from_dlpack(x))
```


simonetgordon:
Thanks Matt! However, with Jax, it doesn't seem to work for me locally:

```
ivy.set_backend("jax")
x = ivy.array([1.])
print(np.from_dlpack(x)) # 'DeviceArray' object has no attribute '__dlpack__'
```

But I'm sure it will work with the chunk of code you have given for tensorflow.

aditya-pola:
| Framework  | \_\_dlpack\_\_ implemented? | from_dlpack requires                            |
|------------|-------------------------|-------------------------------------------------|
| numpy      | Yes                     | Object that has \_\_dlpack\_\_                      |
| torch      | Yes                     | Object that has  \_\_dlpack\_\_ or Pycapsule object |
| tensorflow | No                      | Pycapsule object named 'dltensor'               |
| jax        | No                      | Pycapsule object named 'dltensor'               |

@mattbarrett98 your snippet for tf can be extended to other frameworks while respecting the above table. For e.g. when sending a tensor to tensorflow from numpy, we'd have to use tf.experimental.dlpack.from_dlpack(np_array.__dlpack__()) but torch can accept both torch.utils.dlpack(np_array) and torch.utils.dlpack(np_array.__dlapck__())

There is however, one critical issue, numpy.from_dlpack() can't be used for tensorflow or jax in any for as numpy only accepts objects with the __dlpack__ attribute and not capsules.

There is also a secondary trivial question if we disregard the issue with numpy. Where do we place the \_\_dlpack\_\_ implementation do we keep it in ivy/array.py or in the respective backends because we are doing it on a framework-by-framework basis?

mattbarrett98:
@simonetgordon It works for me, but I'm running it in a graph compiler codespace, so maybe the jax versions differ, but they have added it https://github.com/google/jax/pull/11513. 

@aditya-pola The implementation will need to be in `ivy.Array` since we are adding it so Ivy can adhere to the array API standard, and we don't even have an array object for each backend. Maybe some primary function will need to be added to handle the differences between backends however.

aditya-pola:
@mattbarrett98 which jax version are you using?

aditya-pola:
I checked jax release versions and the earliest one that supports \_\_dlpack\_\_ is v0.3.15 but ivy is pinned to v0.3.14. Can we get this changed?
Are you aware of a similar implementation for tf as well? Cause that would simplify things for us.

djl11:
Unpinning at the moment is likely to cause more problems than it solves I think. Also, we would like to support older versions as well anyway, so we will need a backwards compatible solution one way or another. We briefly discussed this in the call the other day. As @mattbarrett98 suggested, the best way will be to add both `from_dlpack` and `to_dlpack` as _primary_ Ivy functions, which use backend specific approaches (handling native arrays not ivy arrays), and then use these primary functions inside the `ivy.Array` special methods `__dlpack__` and `__dlpack_device__`, both operating on the `self._data` attribute. Does that make sense?

djl11 (changes requested):
see comment above

PR #5298, opened 2022-10-05
pytorch frontend type promotion

AnnaTz:


simonetgordon (changes requested):
Hi Anna, there is an import issue:
`E   ImportError: cannot import name 'promote_types_of_torch_inputs' from partially initialized module 'ivy.functional.frontends.torch' (most likely due to a circular import) `

PR #5352, opened 2022-10-06
Graph Compiler Documentation Update: more intuitive

aryans1204:
…he graph compiler

### DIFF
--- a/docs/partial_source/design/building_blocks.rst
+++ b/docs/partial_source/design/building_blocks.rst
@@ -376,42 +376,40 @@ First, let's set our desired backend as torch.
    :align: center
    :width: 75%
 
-In this example, we observe that both the Ivy code, and the native backend code, compile to the same desired backend code, which in our example was torch.
-Furthermore, the mix of Ivy and native backend code also compiles to the same graph in torch.
+In this example we observe how ivy code, raw torch code, and also a mixture of ivy and torch code all compile to the same torch computation graph.
 
-Now, let's take another example, this time setting the desired backend as tensorflow as tf.
+Now, let's take another example, this time observing how the same Ivy code will compile to different backends. 
 
-+----------------------------------------+-----------------------------------------+-----------------------------------------+
-|.. code-block:: python                  |.. code-block:: python                   |.. code-block:: python                   |
-|                                        |                                         |                                         |                                          
-| def pure_ivy(x):                       | def pure_tf(x):                         | def mix(x):                             | 
-|     y = ivy.mean(x)                    |     y = tf.math_reduce_mean(x)          |     y = ivy.mean(x)                     |             
-|     z = ivy.sum(x)                     |     z = tf.math.reduce_sum(x)           |     z = tf.reduce_sum(x)                    |     
-|     f = ivy.var(y)                     |     f = tf.Variable(y)                  |     f = ivy.var(y)                      |   
-|     k = ivy.cos(z)                     |     k = tf.math.cos(z)                  |     k = tf.math.cos(z)                    |     
-|     m = ivy.sin(f)                     |     m = tf.math.sin(f)                  |     m = ivy.sin(f)                      |     
-|     o = ivy.tan(y)                     |     o = tf.math.tan(y)                  |     o = tf.math.tan(y)                    |    
-|     return ivy.concatenate(            |     return tf.concat(                   |     return ivy.concatenate(             |     
-|         [k, m, o], -1)                 |         [k, m, o], -1)                  |         [k, m, o], -1)                  |        
-|                                        |                                         |                                         |                                        
-| # input                                | # input                                 | # input                                 | 
-| x = ivy.array([[1., 2., 3.]])          | x = tf.Tensor([[1., 2., 3.]])           | x = ivy.array([[1., 2., 3.]])           | 
-|                                        |                                         |                                         |                                   
-| # create graph                         | # create graph                          | # create graph                          | 
-| graph = ivy.compile_graph(             | graph = ivy.compile_graph(              | graph = ivy.compile_graph(              | 
-|     pure_ivy, x)                       |     pure_tf, x)                         |     mix, x)                             |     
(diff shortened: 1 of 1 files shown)


docs/partial_source/design/building_blocks.rst
```diff
@@ -344,36 +344,43 @@ The compiler takes in any Ivy function, backend function, or composition, and re
...
-|                                        |                                         |                                         |
-| # create graph                         | # create graph                          | # create graph                          |
-| graph = ivy.compile_graph(             | graph = ivy.compile_graph(              | graph = ivy.compile_graph(              |
-|     pure_ivy, x)                       |     pure_torch, x)                      |     mix, x)                             |
-|                                        |                                         |                                         |
-| # call graph                           | # call graph                            | # call graph                            |
-| ret = graph(x)                         | ret = graph(x)                          | ret = graph(x)                          |
-+----------------------------------------+-----------------------------------------+-----------------------------------------+
+The best part about the graph compiler is that it will convert code from ANY framework, including Ivy itself, into a computation graph with each node written in the
```
djl11, line 347: This is not correct, the graph compiler does not convert code between frameworks. The graph compiler cannot convert TensorFlow code into PyTorch code, this is the transpiler which is different. The task is to show how the same Ivy code compiled to a different graph depending on the backend we are compiling to.

djl11 (changes requested):

docs/partial_source/design/building_blocks.rst
```diff
@@ -374,6 +376,39 @@ As an example, the following 3 pieces of code all compile to the exact same comp
    :align: center
    :width: 75%
 
+Now, let's take another example, this time setting the desired backend as tensorflow as tf.
```
djl11, line 379: we can't compile `torch` code to `tensorflow` without using the transpiler, the rightmost example below does not make sense.

djl11 (changes requested):

docs/partial_source/design/building_blocks.rst
aryans1204: oops, that's my bad. I forgot to change those functions to tf. I'll do that right away

docs/partial_source/design/building_blocks.rst
```diff
@@ -344,27 +344,29 @@ The compiler takes in any Ivy function, backend function, or composition, and re
    :align: center
    :width: 75%
 
-As an example, the following 3 pieces of code all compile to the exact same computation graph as shown:
+Let's look at a few examples, and observe the compiled graph of the requisite Ivy code, based on the desired backend.
```
djl11, line 347: compiled graph "of the requisite Ivy code" doesn't really make sense. The middle example is not Ivy code at all for example. The point we are showing here is that the graph compiler can be applied either to "Ivy code" or to "native code".

djl11 (changes requested):

docs/partial_source/design/building_blocks.rst
```diff
@@ -376,6 +376,9 @@ First, let's set our desired backend as torch.
    :align: center
    :width: 75%
 
+In this example, we observe that both the Ivy code, and the native backend code, compile to the same desired backend code, which in our example was torch.
```
djl11, line 379: "Here we should an example where `ivy` code, raw `torch` code, and also a mixture of `ivy` and `torch` code all compile to the same `torch` computation graph.

docs/partial_source/design/building_blocks.rst
```diff
@@ -344,7 +344,7 @@ The compiler takes in any Ivy function, backend function, or composition, and re
    :align: center
    :width: 75%
 
-Let's look at a few examples, and observe the compiled graph of the requisite Ivy code, based on the desired backend.
+Let's look at a few examples, and observe the compiled graph of the Ivy code against the native backend code.
```
djl11, line 347: of both Ivy and native framework blocks of code.

docs/partial_source/design/building_blocks.rst
```diff
@@ -376,6 +376,9 @@ First, let's set our desired backend as torch.
    :align: center
    :width: 75%
 
+In this example, we observe that both the Ivy code, and the native backend code, compile to the same desired backend code, which in our example was torch.
+Furthermore, the mix of Ivy and native backend code also compiles to the same graph in torch.
```
djl11, line 380: remove this line.

docs/partial_source/design/building_blocks.rst
```diff
@@ -406,8 +409,9 @@ Now, let's take another example, this time setting the desired backend as tensor
    :align: center
    :width: 75%
 
-In both the examples, we observe that the same Ivy code, is converted into different computation graphs, based on the set framework. Similarly, we've also observed that using a mix of
-Ivy and the desired framework code, also creates the same computation graph as using pure Ivy code. Therefore, all Ivy code is converted into the set backend specific computation graph.
+Additionally, in the above example we also observe that Ivy code and native backend code is compiled to the desired backend code for any framework, not just torch.
```
djl11, line 412: replace all of this with:

The example above further emphasizes that the graph compiler creates a computation graph consisting of _backend_ functions, not _Ivy_ functions. Specifically, the same Ivy code compiles to different graphs depending on the selected backend. However, when compiling native framework code, we are only able to compile a graph for that same framework. For example, we cannot take `torch` code and _compile_ this into `tensorflow` code. However, we can _transpile_ `torch` code into `tensorflow` code (see :ref:`Ivy as a Transpiler` for more details).

djl11 (changes requested):
requested changes to the text, and also the new image should show an ivy function which entails very different computation graphs for each backend. For example, one backend is heavily compositional while the other wraps a simple function. The format should be:

one block of code -> three different computation graphs, one for each backend, and each graph being very different.

docs/partial_source/design/building_blocks.rst
aryans1204: This is definitely what I was trying to convey too, but much clearer and succinct

PR #5364, opened 2022-10-07
Adding tensorflow log_cosh metric

jordibardia:
Adds the log_cosh metric from the issue #5360.

### DIFF
--- a/ivy/functional/frontends/tensorflow/keras/metrics.py
+++ b/ivy/functional/frontends/tensorflow/keras/metrics.py
@@ -203,6 +203,13 @@ kld = kl_divergence
 
 kullback_leibler_divergence = kl_divergence
 
+def log_cosh(y_true, y_pred):
+    y_true = ivy.astype(y_true, y_pred.dtype, copy=False)
+    diff = y_pred - y_true
+    logval = ivy.astype(ivy.log(2.0), diff.dtype, copy=False)
+    return ivy.mean(diff + ivy.softplus(-2.0 * diff) - logval, axis=-1)
+
+log_cosh.unsupported_dtypes = ("float16",)
 
 @to_ivy_arrays_and_back
 def log_cosh(y_true, y_pred):
--- a/ivy/functional/frontends/tensorflow/keras/metrics.py
+++ b/ivy/functional/frontends/tensorflow/keras/metrics.py
@@ -202,6 +202,13 @@ kld = kl_divergence
 
 kullback_leibler_divergence = kl_divergence
 
+def log_cosh(y_true, y_pred):
+    y_true = ivy.astype(y_true, y_pred.dtype, copy=False)
+    diff = y_pred - y_true
+    logval = ivy.astype(ivy.log(2.0), diff.dtype, copy=False)
+    return ivy.mean(diff + ivy.softplus(-2.0 * diff) - logval, axis=-1)
+
+log_cosh.unsupported_dtypes = ("float16",)
 
 @to_ivy_arrays_and_back
 def mean_absolute_error(y_true, y_pred):
--- a/ivy/functional/frontends/tensorflow/keras/metrics.py
+++ b/ivy/functional/frontends/tensorflow/keras/metrics.py
@@ -1,4 +1,5 @@
 import ivy
+from ivy.functional.frontends.tensorflow import promote_types_of_tensorflow_inputs
 from ivy.functional.frontends.tensorflow.func_wrapper import to_ivy_arrays_and_back
 
 
@@ -204,13 +205,14 @@ kullback_leibler_divergence = kl_divergence
 
 
 def log_cosh(y_true, y_pred):
-    y_true = ivy.astype(y_true, y_pred.dtype, copy=False)
+    y_true, y_pred = promote_types_of_tensorflow_inputs(y_true, y_pred)
     diff = y_pred - y_true
-    logval = ivy.astype(ivy.log(2.0), diff.dtype, copy=False)
+    logval = ivy.log(2.0)
+    diff, logval = promote_types_of_tensorflow_inputs(diff, logval)
     return ivy.mean(diff + ivy.softplus(-2.0 * diff) - logval, axis=-1)
 
 
-log_cosh.unsupported_dtypes = ("float16",)
+log_cosh.unsupported_dtypes = ("float16", "float32",)
 
 
 @to_ivy_arrays_and_back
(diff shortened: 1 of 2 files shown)


jordibardia:
Review this pull request when you get a chance. Thanks.

AnnaTz:
Close #5360

ivy/functional/frontends/tensorflow/keras/metrics.py
```diff
@@ -162,6 +162,16 @@ def kl_divergence(y_true, y_pred):
 kullback_leibler_divergence = kl_divergence
 
 
+def log_cosh(y_true, y_pred):
+    y_true = ivy.astype(y_true, y_pred.dtype, copy=False)
+    diff = y_pred - y_true
+    logval = ivy.astype(ivy.log(2.0), diff.dtype, copy=False)
+    return ivy.mean(diff + ivy.softplus(-2.0 * diff) - logval, axis=-1)
```
AnnaTz, line 169: Actually we have a specific function for type promotion in each frontend. So, instead of using `ivy.astype`, it would be better to import the tensorflow frontend like so: `import ivy.functional.frontends.tensorflow as tf_frontend` and then use the function `promote_types_of_tensorflow_inputs` like so: `x1, x2= tf_frontend.promote_types_of_tensorflow_inputs(x1, x2)`.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_metrics.py
```diff
@@ -190,6 +190,38 @@ def test_sparse_categorical_crossentropy(
     )
 
 
+# log_cosh
+@handle_cmd_line_args
+@given(
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
+        num_arrays=2,
+        shared_dtype=True,
```
AnnaTz, line 199: You could let the `shared_dtype` be `False` by default, just to check that the type promotion is working correctly.

AnnaTz (approved):

AnnaTz (changes requested):
I think with the above changes you'll be good to go 🙂

ivy/functional/frontends/tensorflow/keras/metrics.py
jordibardia: I updated the function to use `promote_types_of_tensorflow_inputs` instead of `ivy.astype`. However, using the function with `ivy.log(2.0)` throws an assertion error in my test for dtypes float16 and float32, so I set them as unsupported. Is this okay, or should I revert it back to `ivy.astype`? Using `ivy.astype` passes the test for all dtypes.

AnnaTz: We should use `promote_types_of_tensorflow_inputs` to make sure the type promotion is accurate, but this seems very odd to me. I will investigate this locally and get back to you. There might be something wrong outside of your function.

AnnaTz: Hi @jordibardia. I see there has been a relevant fix in our main branch. Could you try and merge master in your PR?

ivy/functional/frontends/tensorflow/keras/metrics.py
```diff
@@ -162,6 +163,17 @@ def kl_divergence(y_true, y_pred):
 kullback_leibler_divergence = kl_divergence
 
 
+def log_cosh(y_true, y_pred):
+    y_true, y_pred = promote_types_of_tensorflow_inputs(y_true, y_pred)
+    diff = y_pred - y_true
+    logval = ivy.log(2.0)
```
AnnaTz, line 169: This line should be replaced by `logval = ivy.astype(ivy.log(2.0), diff.dtype)`. The problem now is that ivy's default float dtype (float64) is being used, and then typecasted to the other variables in the following line. This causes the test to fail due to type mismatch with the ground truth return.

ivy/functional/frontends/tensorflow/keras/metrics.py
```diff
@@ -162,6 +163,17 @@ def kl_divergence(y_true, y_pred):
...
+    diff, logval = promote_types_of_tensorflow_inputs(diff, logval)
+    return ivy.mean(diff + ivy.softplus(-2.0 * diff) - logval, axis=-1)
+
+
+log_cosh.unsupported_dtypes = ("float16", "float32",)
+
```
AnnaTz, line 175: This can now be removed. The function should work for all dtypes.

AnnaTz (changes requested):

ivy/functional/frontends/tensorflow/keras/metrics.py
jordibardia: Yeah, that's what I've noticed when troubleshooting this problem. I set `logval = ivy.astype(ivy.log(2.0), diff.dtype)`, which causes the test case to pass if `shared_dtype=True`, but still fails if set to `False`. If I switch `y_true, y_pred = promote_types_of_tensorflow_inputs(y_true, y_pred)` to be `y_true = ivy.astype(y_true, y_pred.dtype)`, the function passes the test case with `shared_dtype` set either to `True` or `False`.

AnnaTz: Yes, that's right. I had a look at the original source code here https://github.com/keras-team/keras/blob/v2.10.0/keras/losses.py#L1892-L1935 and it turns out this function doesn't use tensorflow's type promotion (probably because it's the keras one, 'cause `tfp.math.log_cosh` does promote the types the way I was suggesting).
Anyway, yes, the code should be like this
`def log_cosh(y_true, y_pred):
    y_true = ivy.astype(y_true, y_pred.dtype)
    diff = y_pred - y_true
    logval = ivy.astype(ivy.log(2.0), diff.dtype)
    return ivy.mean(diff + ivy.softplus(-2.0 * diff) - logval, axis=-1)`

jordibardia: Alright, switched the implementation back 👍

AnnaTz (changes requested):

ivy/functional/frontends/tensorflow/keras/metrics.py
```diff
@@ -1,4 +1,5 @@
 import ivy
+from ivy.functional.frontends.tensorflow import promote_types_of_tensorflow_inputs
```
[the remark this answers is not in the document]
jordibardia: Ah, I forgot to remove that haha. Sorry about that.

PR #5368, opened 2022-10-07
torch frontend `sin` and `sin_` instance methods

### LINKED ISSUE #5040: sin

### LINKED ISSUE #5041: sin_

hirwa-nshuti:
close #5041 
close #5040


hirwa-nshuti:
Hi, I went through the unit tests and found that the main issue was that for the `Tensor` class we were not calling it the same way as it is done for PyTorch `torch.tensor`  I added an alias for the class and renamed the file to `Tensor.py`
And it is well working now

PR #5393, opened 2022-10-07
Individual task: Frontend Import Docs Improvement

Daniel4078:


docs/partial_source/deep_dive/16_ivy_frontends.rst
```diff
@@ -453,6 +453,12 @@ which itself is implemented as follows:
     # ivy/functional/frontends/numpy/manipulation_routines/changing_array_shape.py
     def reshape(x, /, newshape, order="C"):
         return ivy.reshape(x, shape=newshape)
+        
+The :code:`np_frontend` above is a reference to the numpy frontend of ivy:
```
djl11, line 457: looks good, but could you please move the explanation to "The Basics" section instead, think it's best to mention this early on, rather than only in the "Classes and Instance Methods" section. Thanks!

djl11 (changes requested):

djl11 (approved):

PR #5420, opened 2022-10-07
fix matrix_norm superset behaviour

### LINKED ISSUE #5402: superset behaviour in matrix_norm (linear_algebra.py)
after adding superset behavoir to matrix_norm , the type hint in `axis: Optional[Union[int, Sequence[int]]] = None` gives the possibility to use an integer while the backend implemetations convert the axis to a tuple, in the case of an integer axis that would throw the error "int object is not iterable". Besides, in the original backend implementations, when the axis is an integer the vector norms are calculated instead while the ivy implementation separates the vector and matrix norms implementations, following the python array API standard . So I think there is no need for the int type hint. Also there is a mismatch between default values for the _axis_ argument in [Container, array, functional api] implementations (which is **None**) and the default values in backend implementations (which is **(-2, -1)**). This doesn't throw an error but I don't know if it's the convenient way.

randomstate42:
fixes #5402


randomstate42:
@sherry30 ready for review

ivy/array/linear_algebra.py
```diff
@@ -185,7 +185,7 @@ def matrix_norm(
         /,
         *,
         ord: Optional[Union[int, float, Literal[inf, -inf, "fro", "nuc"]]] = "fro",
-        axis: Optional[Union[int, Sequence[int]]] = None,
+        axis: Optional[Sequence[int]] = (-2, -1),
```
zaeemansari70, line 188: axis: Optional[Tuple[int,int]] = (-2, -1) 
else would remain the same.

ivy/container/linear_algebra.py
```diff
@@ -808,7 +808,7 @@ def static_matrix_norm(
         /,
         *,
         ord: Optional[Union[int, float, Literal[inf, -inf, "fro", "nuc"]]] = "fro",
-        axis: Optional[Union[int, Sequence[int]]] = None,
+        axis: Optional[Sequence[int]] = (-2, -1),
```
zaeemansari70, line 811: here

ivy/container/linear_algebra.py
```diff
@@ -870,7 +870,7 @@ def matrix_norm(
...
```
zaeemansari70, line 873: here

ivy/functional/backends/jax/linear_algebra.py
```diff
@@ -166,11 +167,12 @@ def matrix_norm(
     /,
     *,
     ord: Optional[Union[int, float, Literal[inf, -inf, "fro", "nuc"]]] = "fro",
-    axis: Optional[Union[int, Sequence[int]]] = (-2, -1),
+    axis: Optional[Sequence[int]] = (-2, -1),
```
zaeemansari70, line 170: here

ivy/functional/backends/numpy/linear_algebra.py
```diff
@@ -159,11 +160,12 @@ def matrix_norm(
     /,
     *,
     ord: Optional[Union[int, float, Literal[inf, -inf, "fro", "nuc"]]] = "fro",
-    axis: Optional[Union[int, Sequence[int]]] = (-2, -1),
+    axis: Optional[Sequence[int]] = (-2, -1),
```
zaeemansari70, line 163: here

ivy/functional/backends/tensorflow/linear_algebra.py
```diff
@@ -269,10 +270,11 @@ def matrix_norm(
     /,
     *,
     ord: Optional[Union[int, float, Literal[inf, -inf, "fro", "nuc"]]] = "fro",
-    axis: Optional[Union[int, Sequence[int]]] = (-2, -1),
+    axis: Optional[Sequence[int]] = (-2, -1),
```
zaeemansari70, line 273: here

ivy/functional/backends/torch/linear_algebra.py
```diff
@@ -197,10 +198,11 @@ def matrix_norm(
     /,
     *,
     ord: Optional[Union[int, float, Literal[inf, -inf, "fro", "nuc"]]] = "fro",
-    axis: Optional[Union[int, Sequence[int]]] = (-2, -1),
+    axis: Optional[Sequence[int]] = (-2, -1),
```
zaeemansari70, line 201: here

ivy/functional/ivy/linear_algebra.py
```diff
@@ -933,7 +933,7 @@ def matrix_norm(
     /,
     *,
     ord: Optional[Union[int, float, Literal[inf, -inf, "fro", "nuc"]]] = "fro",
-    axis: Optional[Union[int, Sequence[int]]] = None,
+    axis: Optional[Sequence[int]] = (-2, -1),
```
zaeemansari70, line 936: here

zaeemansari70 (changes requested):
Hey! 
First of all thank you soo much for contributing to ivy! 🙂
Your PR looks amazing ✨!

Mostly, the type-hints also have been fixed already what you can do is:

1. Add what is required to the `axis` argument in the type-hint only, I've also mentioned what to add and where to.

That's all, you can revert all other changes. 🙂
Thanks!

zaeemansari70 (approved):
Lgtm!

PR #5423, opened 2022-10-08
Tensorflow sets.difference #5327

Abheast12:


Abheast12:
I'm unsure if I'm doing this correctly -- is there anything I need to modify?

aditya-pola:
Hey @Abheast12 , could you please clarify what you are intending to do? You've removed a lot of the frontend functions, I'm unsure why

aditya-pola:
It seems like you are attempting to add the difference frontend function and have inadvertently changed other functions. Please revert the changes that aren't related to the implementation of difference

PR #5439, opened 2022-10-08
Lcm extension

nassimberrada:


Ishticode (changes requested):
looks mostly read to merge. I would ask for an addition in the container and array methods.

Also could you clarify `lcm.supported_dtypes....`? As far as I knew only **un**supported dtypes are required to be specified for a spcified funtion

Ishticode:
In addition, np and jax appliy lcm to absolute values anyway (indicated by |x1| notation) so there should be no need for the extra call on abs function

nassimberrada:
I don't know why i left those, used to think that's how you restrict the data types in the testing back then 😅

Ishticode (changes requested):
It turns out torch uses abs already as clear from this
image

and while tf is using the experimental numpy I am assuming that too uses absolute values as np does. So we can also remove abs from both.  For the container's static method, can you add the additional args of keychain and all. Check static_flatten to copy over. Then this should be ready I think.
also the numpy-style-docstring test requires that we have no gap between the heading `Examples` and its content...

nassimberrada:
Weird, some tests i ran locally gave me different signs for the same input, that's why i introduced the abs to begin with 🤔

Ishticode:
> Weird, some tests i ran locally gave me different signs for the same input, that's why i introduced the abs to begin with 🤔

Yes that is strange.

Ishticode:
> Weird, some tests i ran locally gave me different signs for the same input, that's why i introduced the abs to begin with 🤔

Yes that is strange. yet I suspect it might hv something to do with int8 overflowing. The native backends do return positive. But if two nums have lcm higher that 127, it perhaps becomes negative due to dtype limit not due the backend returning negative.

nassimberrada:
> > Weird, some tests i ran locally gave me different signs for the same input, that's why i introduced the abs to begin with thinking
> 
> Yes that is strange. yet I suspect it might hv something to do with int8 overflowing. The native backends do return positive. But if two nums have lcm higher that 127, it perhaps becomes negative due to dtype limit not due the backend returning negative. not sure. just guessing

I see that makes sense, i guess that's the origin because i also had some dtype correspondance issues between tensorflow and numpy especially with int8, i guess solving it also solved the negative sign issue and it had nothing to do with abs

Ishticode:
The following snippet does show the problem

``` python
>>> import numpy as np
>>> a = np.array([5], dtype=np.int8)
>>> b = np.array([26], dtype=np.int8)
>>> np.lcm(a,b)
array([-126], dtype=int8)
```

Mathematically, lcm(5,26) =130

Ishticode (commented):
If you run the

```python
import ivy

fw = ['numpy', 'torch', 'tensorflow', 'jax']

for f in fw:
    ivy.set_backend(f)
    x = ivy.array([5], dtype='int8')
    y = ivy.array([26], dtype='int8')
    print((ivy.lcm(x, y)))
```
you get negative results except for jax which gives positive. 

I will see what the best way is and if not we shall fall back to abs method. Meanwhile if you have any solution feel free to put it through

nassimberrada:
> The following snippet does show the problem
> 
> ```python
> >>> import numpy as np
> >>> a = np.array([5], dtype=np.int8)
> >>> b = np.array([26], dtype=np.int8)
> >>> np.lcm(a,b)
> array([-126], dtype=int8)
> ```
> 
> Mathematically, lcm(5,26) =130

Those are the exact input that kept throwing an error on my side, so it's not just a dtype issue after all

Ishticode (changes requested):
Could you try the tests by putting abs on torch, np and tf and if still fail, then jax too. should get it working but its not ideal.

The problem is different versions of backends act differently with this particular version. Even numpy had different behaviour between version 1.23 and 1.21. So not the ideal but to create the same behaviour for this particular case, we have to use abs. Then this is almost ready to merge imo. Thanks for the patience Nassim.

nassimberrada:
Seems to be passing without adding abs to jax. Also, i had to add unsupported uint dtypes for tensorflow because it was throwing errors for them.

Ishticode:
seems to have caused some conflicts within elementewise. best to move this to extensions. Sorry if thats painful 😐

Ishticode (changes requested):
see the latest comment above

nassimberrada:
Moved and merged

Ishticode (changes requested):
Hi @nassimberrada 
Two things
1. somehow the function `def` part for numpy backend has gone missing. I hope thats not just me seeing things 😂. Could you fix this and ensure the tests are passing and i will get this merged. 
2. The instance method examples should have calls like `x1.lcm(x2)` in the docstring. rest is fine.
and the static_lcm should be ivy.Container.static_lcm(x1,x2) in the example.

Request my review and Ig we can merge as long as the tests pass. Thank you for your patience @nassimberrada

nassimberrada:
See, that's also why i'm grateful to have a meticulous supervisor, i often try to go too fast and end up making this kind of blunders 😂
In any case, the tests do pass on my side, and i corrected the calls description in the docstrings, it should (hopefully) be good now

Issue #4612, opened 2022-09-19
Add `jax.numpy` namespace functions to Jax frontend


hirwa-nshuti:
Add [`jax.numpy` namespace functions](https://jax.readthedocs.io/en/latest/jax.numpy.html) to Jax frontend

- [ ] ndarray.at
- [x] #4884
- [x] #4915
- [x] #4916
- [x] #4917
- [x] #4897
- [x] #6902
- [ ] #7046
- [x] #7047
- [ ] angle
- [x] #6651
- [ ] append
- [ ] apply_along_axis
- [ ] apply_over_axes
- [x] #7322
- [x] #5248
- [x] #5249
- [x] #5250
- [x] #5251
- [x] #5369
- [x] #5371
- [x] #6895
- [x] #5214
- [x] #6091
- [x] #7423
- [x] #7425
- [ ] around
- [x] #7210
- [x] #5700
- [x] #5867
- [ ] array_repr
- [ ] array_split
- [ ] array_str
- [x] #7712
- [ ] atleast_1d
- [ ] atleast_2d
- [ ] atleast_3d
- [ ] average
- [ ] bartlett
- [x] #6163
- [x] #5857
- [x] #6211
- [x] #5925
- [x] #6235
- [ ] blackman
- [ ] block
- [ ] bool_
- [ ] broadcast_arrays
- [ ] broadcast_shapes
- [x] #4949
- [ ] c_
- [x] #7770
- [ ] cbrt
- [ ] cdouble
- [x] #6326
- [ ] character
- [ ] choose
- [x] #4957
- [ ] column_stack
- [ ] complex_
- [ ] complex128
- [ ] complex64
- [ ] complexfloating
- [ ] complexWarning
- [ ] compress
- [x] #4948
- [ ] conj
- [ ] conjugate
- [ ] convolve
- [ ] copy
- [ ] copysign
- [ ] corrcoef
- [ ] correlate
- [x] cos
- [ ] cosh
- [ ] count_nonzero
- [ ] cov
- [ ] cross
- [ ] csingle
- [x] #6258
- [x] #6465
- [x] #6466
- [x] #6467
- [ ] #7477
- [ ] delete
- [x] #6655
- [ ] diag_indices
- [ ] diag_indices_from
- [ ] #6893
- [x] #7429
- [ ] diff
- [ ] digitize
- [x] #7771
- [ ] divmod
- [x] #4946
- [ ] double
- [ ] dsplit
- [ ] dstack
[...]

PR #5452, opened 2022-10-08
Added jax.numpy.cos namespace function to jax frontend and tested.

vaatsalya123:
Nothing much a basic cos function implementation.

#4612


JamieLine:
Just had to fix a small linting error, it was just a few cases of only leaving one blank line between functions rather than the 2 that should be there.

PR #5456, opened 2022-10-08
Create top_k_categorical_accuracy

hrak99:
Tensorflow metric top_k_categorical_accuracy

### DIFF
--- a/ivy/functional/ivy/linear_algebra.py
+++ b/ivy/functional/ivy/linear_algebra.py
@@ -2320,15 +2320,19 @@ def vander(
 @handle_nestable
 @handle_exceptions
 def vector_to_skew_symmetric_matrix(
-    vector: Union[ivy.Array, ivy.NativeArray], /, *, out: Optional[ivy.Array] = None
+    vector: Union[ivy.Array, ivy.NativeArray], 
+    /, 
+    *, 
+    out: Optional[ivy.Array] = None
 ) -> ivy.Array:
-    """Given vector, return the associated skew-symmetric matrix
-    `[reference] <https://en.wikipedia.org/wiki/Skew-symmetric_matrix#Cross_product>`_
+    """Given vector, return the associated `Skew-symmetric matrix
+        <https://en.wikipedia.org/wiki/Skew-symmetric_matrix#Cross_product/>`_.
+
 
     Parameters
     ----------
     vector
-        Vector to convert *[batch_shape,3]*.
+        Vector to convert *(batch_shape,3)*.
     out
         optional output array, for writing the result to. It must have a shape that the
         inputs broadcast to.
@@ -2336,7 +2340,7 @@ def vector_to_skew_symmetric_matrix(
     Returns
     -------
     ret
-        Skew-symmetric matrix *[batch_shape,3,3]*.
+        Skew-symmetric matrix *(batch_shape,3,3)*.
 
 
     Both the description and the type hints above assumes an array input for simplicity,


hrak99:
I don't know why it's giving 

TypeError: one_hot() got some positional-only arguments passed as keyword arguments: 'depth'.

hrak99 (commented):
Don't know why this error is generated!
TypeError: one_hot() got some positional-only arguments passed as keyword arguments: 'depth'.

hrak99:
Waiting for your reply,:)

hrak99:
Error
Kindly assist, :).

hrak99 (commented):
Error
Kindly assist, :).

hrak99 (commented):
Error
with output, :)

Nightcrab:
```
y_true = ivy.one_hot(y_true,3)
```
You don't need to create a one-hot here because `y_true` should already be a one-hot according to tensorflow documentation.

```
accuracy = topk/len(y_true)
```

The return value should not be a scalar, but a Tensor of floats which are 0 or 1 depending on whether the ground truth was in the top K of predictions.

I would suggest reading the following:

TF docs: https://www.tensorflow.org/api_docs/python/tf/keras/metrics/top_k_categorical_accuracy
`top_k_categorical_accuracy` source: https://github.com/keras-team/keras/blob/v2.10.0/keras/metrics/metrics.py#L3678-L3706
`sparse_top_k_categorical_matches` source: https://github.com/keras-team/keras/blob/b80dd12da9c0bc3f569eca3455e77762cf2ee8ef/keras/utils/metrics_utils.py#L976

Then cross-check with them that your function at least in theory has the same behaviour.

hrak99 (commented):
Kindly assist, :)

hrak99 (commented):
Stuck

Nightcrab:
At the moment, you are neglecting to do an ```ivy.argmax``` on `y_true` to extract the target classes. You can just insert a

```
y_true = ivy.argmax(y_true,axis=-1)
```

It will behave like this:

```
x = ivy.array([[0, 0, 1], [0, 1, 0]])
print(ivy.argmax(x, axis=-1))
# > ivy.array([2,1])
```

Near the top of the function to correct this.

Regarding this subfunction:

```
def _top_k(input, topk):
            x = ivy.array(input)
            sort = ivy.argsort(x, descending=True)
            topk = min(x.shape[-1], topk)

            # Safety check for equal values
            result = []
            for ind, li in enumerate(sort):
                temp = [x[ind, _] for _ in li[:topk]]
                result.append(temp)

            return ivy.array(result)

        top_k = _top_k(predictions, topk)

        labels = ivy.shape(predictions)[1]
        # float comparison?
        return ivy.array(
            [
                (
                        0 <= res < labels
                        and ivy.min(top_k[ind] - predictions[ind, res]) <= 1e-9
                )
                for ind, res in enumerate(targets)
            ]
        )
```

This is excessively complicated for what it does, which may be why you are having difficulty debugging it. 

If you simply want to trim one axis of the array to a certain size, you can use `ivy.gather`:

```
x = ivy.array([[1, 2, 3, 4], [1, 2, 3, 4]])
y = ivy.array([[0,1]])
ivy.gather(x, y, axis=-1, out=x)
print(x)

# > ivy.array([[[1, 2]],
#             [[1, 2]]])
```

you can use this to extract the top K values in the predictions. After that, you just need to create the float array of `0` and `1` depending on if a target value is `in` the trimmed and sorted prediction.

As a side note, I would avoid using `enumerate` on ivy Arrays just because this isn't well documented.

hrak99:
> At the moment, you are neglecting to do an `ivy.argmax` on `y_true` to extract the target classes. You can just insert a
> 
> ```
> y_true = ivy.argmax(y_true,axis=-1)
> ```
> 
> near the top of the function to correct it.
> 
> It will behave like this:
> 
> ```
> x = ivy.array([[0, 0, 1], [0, 1, 0]])
> print(ivy.argmax(x, axis=-1))
> # > ivy.array([2,1])
> ```
> 
> Regarding this subfunction:
> 
> ```
> def _top_k(input, topk):
>             x = ivy.array(input)
>             sort = ivy.argsort(x, descending=True)
>             topk = min(x.shape[-1], topk)
> 
>             # Safety check for equal values
>             result = []
>             for ind, li in enumerate(sort):
>                 temp = [x[ind, _] for _ in li[:topk]]
>                 result.append(temp)
> 
>             return ivy.array(result)
> 
>         top_k = _top_k(predictions, topk)
> 
>         labels = ivy.shape(predictions)[1]
>         # float comparison?
>         return ivy.array(
>             [
>                 (
>                         0 <= res < labels
>                         and ivy.min(top_k[ind] - predictions[ind, res]) <= 1e-9
>                 )
>                 for ind, res in enumerate(targets)
>             ]
>         )
> ```
> 
> This is overly complicated for what it does, which may be why you are having difficulty debugging it.
> 
> If you simply want to trim one axis of the array to a certain size, you can use `ivy.gather`:
> 
> ```
> x = ivy.array([[1, 2, 3, 4], [1, 2, 3, 4]])
> y = ivy.array([[0,1]])
> ivy.gather(x, y, axis=-1, out=x)
> print(x)
> 
> # > ivy.array([[[1, 2]],
> #             [[1, 2]]])
> ```
> 
> you can use this to extract the top K values in the predictions, by passing an array like `[[0,1...k]]`. After that, you just need to create the float array of `0` and `1` depending on if a target value is `in` the trimmed and sorted prediction.
> 
> As a side note, I would avoid using `enumerate` on ivy Arrays just because this isn't well documented.

thanks for the comprehensive response.

hrak99:
The same code is working very well without using y_true = tf.argmax(y_true,axis=-1). I have just made changes in 
def sparse_top_k_categorical_accuracy(y_true, y_pred, k=5): function
original function:
 matches = tf.cast(
        tf.math.in_top_k(
            predictions=y_pred, targets=tf.cast(y_true, "int32"), k=k
        ),
        dtype=backend.floatx(),
    )
Revised:-
matches = tf.cast(
        tf.math.in_top_k(
            predictions=tf.cast(y_pred,"float32"), targets=tf.cast(y_true, "int32"), k=k
        ),
        dtype=tf.keras.backend.floatx(),
    )
[top_k_Cat_Acc.pdf](https://github.com/unifyai/ivy/files/9769653/top_k_Cat_Acc.pdf)

hrak99 (commented):
I have also left a comment in the 
[top_k_Cat_Acc.pdf](https://github.com/unifyai/ivy/files/9769691/top_k_Cat_Acc.pdf)
conversation. Please see it.

Nightcrab:
> The same code is working very well without using y_true = tf.argmax(y_true,axis=-1). I have just made changes in def sparse_top_k_categorical_accuracy(y_true, y_pred, k=5): function original function: matches = tf.cast( tf.math.in_top_k( predictions=y_pred, targets=tf.cast(y_true, "int32"), k=k ), dtype=backend.floatx(), ) Revised:- matches = tf.cast( tf.math.in_top_k( predictions=tf.cast(y_pred,"float32"), targets=tf.cast(y_true, "int32"), k=k ), dtype=tf.keras.backend.floatx(), ) [top_k_Cat_Acc.pdf](https://github.com/unifyai/ivy/files/9769653/top_k_Cat_Acc.pdf)

you can use three backticks \`\`\` to create a codeblock.


> I have also left a comment in the [top_k_Cat_Acc.pdf](https://github.com/unifyai/ivy/files/9769691/top_k_Cat_Acc.pdf) conversation. Please see it.

There are many comments in this pdf. If you want to tell me something you can send it in the PR conversation.

Regarding `sparse_top_k_categorical_matches`, this is originally a utility function in TF and shouldn't be exposed in the frontend.

If the code is working very well, then I can review it once the tests are passing.

hrak99 (commented):
Still facing errors but I think I have found out the reasons to solve them.

Looking forward to your assistance, :)

Nightcrab:
That input from the test looks valid to me. If you're getting a StridedSlice error it will be due to a use of ```[]``` array indexing for example if you take an index of a 0-dimensional array. `argmax` is one function that removes one dimension of an array and can result in a 0D array.

I noticed the current version of the code has an `argmax` but it is in the wrong place. It should be placed in front of the reshaping code.

```
result = []
    for ind, li in enumerate(sort):
        temp = [x[ind, _] for _ in li[:topk]]
        result.append(temp)
return ivy.array(result)
```
Converting a 2D list of ivy arrays to an ivy array will result in an exception with torch backend. 1D lists work however, so you can resolve this by making `temp` an ivy array. You should also add `dtype="float32"` to this line and the return as well to avoid unexpected int casts.

```
output =  ivy.array(
            [
                (
                    0 <= targets[i] < labels and ivy.min(top_k[i] - predictions[i,targets[i]]) <= 1e-9
                )
                for i in range(len(targets))
            ]
        )
```

Creating an array of booleans is a bad idea, but also unnecessary. Instead of comprehending the array again you should just do

```
return ivy.array(
            [
                1 if (
                    0 <= targets[i] < labels and ivy.min(top_k[i] - predictions[i,targets[i]]) <= 1e-9
                ) else 0
                for i in range(len(targets))
            ]
        )
```

There are some things that should be changed with the hypothesis test as well.

`helpers.dtype_and_values` produces a list of arrays, not an array itself, so you'll want to pass `x[0]` and `y[0]` instead of `x` and `y`, and set `min_num_dims` to 2.

`k`'s min value should be 1.

hrak99:
> That input from the test looks valid to me. If you're getting a StridedSlice error it will be due to a use of `[]` array indexing for example if you take an index of a 0-dimensional array. `argmax` is one function that removes one dimension of an array and can result in a 0D array.
> 
> I noticed the current version of the code has an `argmax` but it is in the wrong place. It should be placed before the reshaping code.
> 
> ```
> result = []
>     for ind, li in enumerate(sort):
>         temp = [x[ind, _] for _ in li[:topk]]
>         result.append(temp)
> return ivy.array(result)
> ```
> 
> Converting a 2D list of ivy arrays to an ivy array will result in an exception with torch backend. 1D lists work however, so you can resolve this by making `temp` an ivy array. You should also add `dtype="float32"` to this line and the return as well to avoid unexpected int casts.
> 
> ```
> output =  ivy.array(
>             [
>                 (
>                     0 <= targets[i] < labels and ivy.min(top_k[i] - predictions[i,targets[i]]) <= 1e-9
>                 )
>                 for i in range(len(targets))
>             ]
>         )
> ```
> 
> Creating an array of booleans is a bad idea, but also unnecessary. Instead of comprehending the array again you should just do
> 
> ```
> return ivy.array(
>             [
>                 1 if (
>                     0 <= targets[i] < labels and ivy.min(top_k[i] - predictions[i,targets[i]]) <= 1e-9
>                 ) else 0
>                 for i in range(len(targets))
>             ]
>         )
> ```
> 
> There are some things that should be changed with the hypothesis test as well.
> 
> `helpers.dtype_and_values` produces a list of arrays, not an array itself, so you'll want to pass `x[0]` and `y[0]` instead of `x` and `y`, and set `min_num_dims` to 2.
> 
> `k`'s min value should be 1.

thanks a lot. This will suffice to pass the tests, :). I have removed the enumerate function but i think due to a conflict with, ivy/functional/frontends/tensorflow/layers.py; is it showing the latest code?

Nightcrab:
For some reason the code in that quote was from an old version, probably a mistake on my end.

hrak99:
tensorflow.python.framework.errors_impl.InvalidArgumentError: cannot compute InTopKV2 as input #0(zero-based) was expected to be a float tensor but is a bfloat16 tensor [Op:InTopKV2]

That type of error I removed or made changes in the main code (already shared).

Now, why the y_pred is showing in bfloat16,
y_true = <tf.Tensor: shape=(1,), dtype=int64, numpy=array([0])>
y_pred = <tf.Tensor: shape=(1, 1), dtype=bfloat16, numpy=array([[0]], dtype=bfloat16)>

Nightcrab:
bfloat16 is one of the float types returned by the helper. You can do something like

```
assume("bfloat16" not in input_dtype)
```

to skip those test cases.

hrak99 (commented):
Now everything is working except that dtype error.

hrak99 (commented):
errors related to double or half tensors are gone due to skipping tests. but now facing a new errors related to  AssertionError
maybe i'm not doing the skip test properly as i declared two separate arrays or something else....

Nightcrab:
The Assertion errors will be related to the logic of the frontend function. If you run the function on the falsifying inputs, is the behaviour consistent (and incorrect)?

Nightcrab:
You took `ivy.min` over the wrong axis. You can't take it over a dimension of size 0. 

Also, it would be better to keep the commit messages short and write comments in the thread (even if the commits get squashed in the end).

hrak99:
> You took `ivy.min` over the wrong axis. You can't take it over a dimension of size 0. Probably, you wanted to take it over the -1 axis which has size 3.
> 
> Also, it would be better to keep the commit messages short and write comments in the thread (even if the commits get squashed in the end).

but if i use with np.min(same code), there is no problem. Why?

Nightcrab:
I have just tested the current commit relacing `ivy.min` with `numpy.min` and cannot reproduce what you mentioned (no errors). Are you sure that was the only local change you made?

As for why the error is occurring, it is because you have perfored `argmax` twice:

```
return sparse_top_k_categorical_matches(
        ivy.argmax(y_true, axis=-1), y_pred, k
    )
```

```
targets = ivy.argmax(targets,axis=-1)
```

it looks like perhaps you forgot to remove the second one during a refactor and this is creating a 0D array with some test cases.

hrak99:
> I have just tested the current commit relacing `ivy.min` with `numpy.min` and cannot reproduce what you mentioned (no errors). Actually, it returns the same error still. Are you sure that was the only local change you made?
> 
> As for why the error is occurring, it is on closer inspection because you have performed `argmax` twice:
> 
> ```
> return sparse_top_k_categorical_matches(
>         ivy.argmax(y_true, axis=-1), y_pred, k
>     )
> ```
> 
> ```
> targets = ivy.argmax(targets,axis=-1)
> ```
> 
> it looks like perhaps you forgot to remove the second one during a refactor and this is creating a 0D array with some test cases.

If I used only the last one (argmax), then the Error: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

hrak99:
I have rerun with or without using argmax in the beginning (using NumPy and TensorFlow operations), the code is working without a hitch. I don't know why I'm stuck. It's taking a lot of time even on small query, :)

hrak99:
There is sign regarding conflict related to the branch. I hope you are receiving the new commit code every time. I will try and commit & push the code again in either condition (solved or unsolved).

hrak99 (commented):
this error means dealing with empty array, which I think. But as i said implementing with NumPy and TensorFlow, there is no problem even with or without calling argmax twice too.

Nightcrab:
> But as i said implementing with NumPy and TensorFlow, there is no problem even with or without calling argmax twice too.

You said as such but I'm not sure what that code looks like. It also doesn't make much sense that you could call argmax between 0 and 2 times on the inputs without affecting the function output. Even so, ivy frontend functions shouldn't use hardcoded backend functions so you will have to make the ivy version of the code work either way.

> If I used only the last one (argmax), then the Error: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

You get that error because the other one is resolved (no more 0D array) so this is just a different problem. 

It arises from a non scalar comparison in this code:
```
0 <= targets[i] and targets[i] < labels and ivy.min((top_k[i] - predictions[i,targets[i]])) <= 1e-9)
```

For example, ``[1,5] < 3`` will return the array ``[True,False]`` and similarly, ``[True, False] and [False, False]`` resolves to ``[False,False]``. As I understood with your code, this is just meant to resolve to a single boolean, so there is a problem with some preceding logic which is not reducing the arrays correctly.

Regarding the merge conflict, I would recommend regularly merging master into your fork to stay up to date. In some cases there will be a bug with Ivy that gets resolved in the latest version of master. For example, you are still working with a version of master where there was a mistake in the ``ivy.gather`` code.

hrak99:
> > But as i said implementing with NumPy and TensorFlow, there is no problem even with or without calling argmax twice too.
> 
> You said as such but I'm not sure what that code looks like. It also doesn't make much sense to me that one could call argmax between 0 and 2 times on the inputs and not affect the function output. Even so, ivy frontend functions shouldn't use hardcoded backend functions so you will have to make the ivy version of the code work either way.
> 
> > If I used only the last one (argmax), then the Error: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
> 
> You get that error because the other one is resolved (no more 0D array) so this is just a different problem.
> 
> It arises from a non scalar comparison in this code:
> 
> ```
> 0 <= targets[i] and targets[i] < labels and ivy.min((top_k[i] - predictions[i,targets[i]])) <= 1e-9)
> ```
> 
> For example, `[1,5] < 3` will return the array `[True,False]` and similarly, `[True, False] and [False, False]` resolves to `[False,False]`. As I understood with your code, this is just meant to resolve to a single boolean, so there is a problem with some preceding logic which is not reducing the arrays correctly.
> 
> Regarding the merge conflict, I would recommend regularly merging master into your fork to stay up to date. In some cases there will be a bug with Ivy that gets resolved in the latest version of master. For example, you are still working with a version of master where there was a mistake in the `ivy.gather` code.

you understand correctly. so give me any advice to finish it.

hrak99 (commented):
I removed that part ivy.min((top_k[i] - predictions[i,targets[i]])) <= 1e-9 to see any further errors. Now it's giving different errors which I don't know yet. I don't know about the falsified inputs. kindly check both files, so at least could figure out something. If I used assume command to ignore some dtypes then, some errors are like half or double tensor is not required.

Nightcrab:
For reference, the previous comment I made was written based on [ec1bc5d] though I see you just pushed [db23d25].

hrak99:
> For reference, the previous comment I made was written based on [[ec1bc5d](https://github.com/unifyai/ivy/commit/ec1bc5d0ee3ec9afccf57215e363a4344c2c348e)] though I see you just pushed [[db23d25](https://github.com/unifyai/ivy/commit/db23d259454240ef2547b1cdfcbc9162908cf428)]. The suggestions may still be helpful.

Sorry for bothering you again and again, :). I think, it's not a difficult task at all but maybe things take time in the beginning, :)

hrak99:
Yes, these suggestions are very helpful and I will let you know.  thanks

hrak99 (commented):
if i change the y_true and y_pred to define x and y values, then receive  ValueError: zero-size array to reduction operation minimum which has no identity. I will try to send you another file regarding that change in the test layer.

hrak99:
I would like to ask about " Regarding the merge conflict, I would recommend regularly merging master into your fork to stay up to date. In some cases there will be a bug with Ivy that gets resolved in the latest version of master. For example, you are still working with a version of master where there was a mistake in the ivy.gather code. "
if i update syn fork option as it is telling discard 16 commits, then that conflict will be resolved. last time i updated or sync the fork, the previous work was gone and i have to do all the process from the start like first selecting the issue and then so on

Nightcrab:
You will need to perform a merge using `git merge` on local and then push to the fork. (syncing fork is not what you want, as that will destroy any unmerged changes).

The hypothesis test looks fine to me, though `layers.py` is essentially the same as it was last time I code reviewed, so my previous advice on how to deal with the error
```
ValueError: zero-size array to reduction operation minimum
``` 
stands.

ivy/functional/frontends/tensorflow/layers.py
```diff
@@ -0,0 +1,94 @@
...
+def top_k_categorical_accuracy(y_true, y_pred, k):
+    #y_pred = ivy.astype(y_pred,'float32')
+    #y_true = ivy.one_hot(y_true,3)
+    #assert(ivy.astype!="bfloat16")
+    #assert(ivy.astype!="float16")
+    #assert(ivy.astype!="float64")
+    #assert (ivy.astype != "int8")
+    #assert(ivy.astype=='float32')
+    def in_top_k(targets, predictions, topk):
+        targets = ivy.argmax(targets,axis=-1)
```
Nightcrab, line 16: Remove this line

ivy/functional/frontends/tensorflow/layers.py
```diff
@@ -0,0 +1,94 @@
...
+
+        top_k = _top_k(predictions, topk)
+
+        labels = ivy.shape(predictions)[1]
+        # float comparison?
+
+
+        values_store = []
+        #print ('targets',targets)
+        targets = ivy.array([targets])
```
Nightcrab, line 40: Replace with 
```
targets = ivy.array(targets)
```

ivy/functional/frontends/tensorflow/layers.py
```diff
@@ -0,0 +1,94 @@
...
+        y_true_org_shape = ivy.shape(y_true)
+        y_true_rank = y_true.ndim
+        y_pred_rank = y_pred.ndim
+
+        # y_pred shape to (batch_size, num_samples), y_true shape to (num_samples,)
+        if (y_true_rank is not None) and (y_pred_rank is not None):
+            if y_pred_rank > 2:
+                y_pred = ivy.reshape(y_pred, shape=[-1, y_pred.shape[-1]])
+            if y_true_rank > 1:
+                reshape = True
```
Nightcrab, line 79: change to
```
reshape_matches = True
```

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_layers.py
```diff
@@ -0,0 +1,93 @@
...
+        frontend="tensorflow",
+        fn_tree="keras.metrics.top_k_categorical_accuracy",
+        y_true=x,
+        y_pred=y,
+        k = k
+    )
+
+
+x = ivy.array([[0,0,1],[0, 1, 0]],dtype="float32")
+y = ivy.array([[0.1, 0.9, 0.8],[0.05, 0.95, 0]],dtype='float32')
```
Nightcrab, line 91: Once the other changes are made, try running this test.

Nightcrab (changes requested):

Nightcrab:
By the way, ``ivy.top_k`` was added to ivy recently, so you might find that useful.

hrak99:
> By the way, `ivy.top_k` was added to ivy recently, so you might find that useful.

not available to me, maybe didn't merge properly.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_layers.py
hrak99: > Once the other changes are made, try running this test case.

passed the test on this test case. otherwise running the main function test, raises ivy.exceptions.IvyBackendException: tensorflow: argmax: 'numpy.ndarray' object has no attribute 'numpy'

hrak99 (commented):
I tried to merge all the three files like ivy and two ivy_tests files. On ivy, it says it has already merged but the other two files were updated in that process.

Nightcrab:
I can't really help you with merging, but there's probably reasonable guides out there.

hrak99:
Did you see the code?

Nightcrab:
Yeah the changes look good, at least the function runs now. But the hypothesis tests are the next step.

hrak99 (commented):
During the commit, I selected only the relevant files. Though some were showing as modified but ignored them.

hrak99:
would Rebase help to resolve the branch conflict? merge option is giving unable to save uncommitted changes during update

Nightcrab:
You may be able to merge more easily by removing all changes, merging, then recommitting the changes post-merge. This would just be removing the functions and hypothesis tests, committing, merging (on local), adding them back then committing and publishing to fork.

hrak99 (commented):
Still facing issues when clicking to project update in the git options.

PR #4811, opened 2022-09-22
Add complex dtype and isreal function - Ivy functional API extension

Nirzu97:
closes #4789

PR #4806, opened 2022-09-22
log_softmax ivy functional API extension 

hirwa-nshuti:
Close #4529

PR #4239, opened 2022-09-09
deg2rad,rad2deg extensions 

ahmedo42:
#4236 
#4237 
if this is approved I will then add the rad2deg in the same PR

PR #5468, opened 2022-10-08
new manipulation function (flatten) has been added

Viditagarwal7479:
Close [5465](https://github.com/unifyai/ivy/issues/5465)

### DIFF
--- a/ivy/functional/backends/jax/activations.py
+++ b/ivy/functional/backends/jax/activations.py
@@ -9,9 +9,7 @@ import jax.numpy as jnp
 from typing import Optional, Union
 
 # local
-from ivy.func_wrapper import with_unsupported_dtypes
 from ivy.functional.backends.jax import JaxArray
-from . import backend_version
 
 
 def gelu(
--- a/ivy/functional/backends/torch/linear_algebra.py
+++ b/ivy/functional/backends/torch/linear_algebra.py
@@ -357,7 +357,8 @@ def svdvals(x: torch.Tensor, *, out: Optional[torch.Tensor] = None) -> torch.Ten
 svdvals.support_native_out = True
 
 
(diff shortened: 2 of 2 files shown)


JamieLine:
Hi!

Thanks for the PR! I've had a look, most of it looks pretty good, but I did find a few things that need to be fixed before it can be merged.

1. As this is an API extension, it should be added to the `extensions.py` files rather than `manipulation.py` files. This change will be easy to make it's the same content just moved.
2. When I ran the tests locally, they found an error where the function attempted to reshape an array from size `2` to size `(1, 1)` which doesn't work because those 2 shapes are for arrays with different numbers of elements. It's probably just an error in the calculation of the output shape.
3. Related to the previous point, I found an error where if `end_dim` is less than `start_dim`, the output shape will be wrong. `end_dim` is allowed to be less than `start_dim` if `end_dim` is negative. It'll probably be fixed by converting any negative dimension values to their positive equivalents.
4. The GitHub CI tests found docstring failures, check the logs to find out what is causing them to fail and hopefully they'll be easy enough to fix.

The rest looks very good though. If you have any questions or comments, please do let me know!

Thanks!

Viditagarwal7479:
#5490 
I have incorporated those changes in this pull request [5490](https://github.com/unifyai/ivy/pull/5490)

Due to some issues in pulling and pushing, I had to delete my previous fork.

Viditagarwal7479:
Hey Jamie,

I just went through the ```1st``` point carefully and noticed that ```4 out of the 7``` tasks done in the list are made in the wrong file.
- https://github.com/unifyai/ivy/pull/4806/files
- https://github.com/unifyai/ivy/pull/4239/files (consisting of 2 functions added)
- https://github.com/unifyai/ivy/pull/4811/files

Shall I move those to ```extensions.py``` and create a pull request?

Thanks :smiley:

PR #5481, opened 2022-10-09
Added jax.numpy trig namespace functions to jax frontend and tested.

vaatsalya123:
Added most of the trig functions into jax numpy backend.

#4612


JamieLine:
Hi!

I've had a look through it all, it's mostly good but there's a few things here that need tweaks. All of these tweaks are fairly minor, so don't worry about how many of them there are, and they're all completely reasonable things to do, it's just a case of sticking with best practices. These are all written specifically about the `cosh` test, but where they apply to the other tests they should be fixed there as well.

1. There's a linting error still, I think what happened was you fixed the linting errors before, and then removed the `settings` flag, which gave a new linting error because `settings` was imported but not used.
2. The `available_dtypes` can be set to `helpers.get_dtypes("numeric")` or `helpers.get_dtypes("valid")`. I'd recommend the second. This is an error I made when I first wrote tests because obviously a floating point variable can hold integer values, but it is better to test not just integer values as floats, but also integer values as integers. If the original Jax function crashes with particular dtypes then feel free to remove those, but ideally you want the "broadest" test possible.
3. The tests take in a `dtype` argument that is then not used. The `dtype_and_x` tuple will contain the `dtype` that was used for `x`, so you don't need to generate a second data type.
4. The frontend flag in `helpers.test_frontend_function` can be set to `fw`, I think it'll then always be equal to "jax" so I don't think it makes a real difference, it's just a best practice thing in case somehow in the future it needs to be updated.

I want to make it clear that while this list might look demoralising, these are all things I've either done while getting my head around Hypothesis and the tests, or completely understand the rationale behind. Hypothesis is just a bit awkward, and it takes a bit of time to get to grips with the best practices of everything. 

vaatsalya123:
No problem I will sit down and fix all of them thanks.

PR #5488, opened 2022-10-09
Created percentile statistical file

kariswanjiru:


kariswanjiru:
percentile #4998

CatB1t (changes requested):
We should add the function `percentile` to the `numpy/statistics` module, we should also implement the frontend function and write the test for it, take a look here for more info: https://lets-unify.ai/ivy/

PR #5511, opened 2022-10-10
adding nanmean function to numpy frontend

### LINKED ISSUE #5497: nanmean

Ahmed-Khaled22:
close #5497


Ahmed-Khaled22:
please review my test. the run is passed, but I'm not sure if the test covers all the possible inputes especially  that I need to test with nan

CheezBallzPi (changes requested):
Don't add the changes in the .idea folder.

I believe the statistical_dtype_values function does incorporate NaN values, so everything else looks good.

PR #5516, opened 2022-10-10
lcm torch frontend

nassimberrada:


nassimberrada:
The tests seem to be failing due to the overflow error you mentioned previously when the data type is set to int8, do you know how to remove that data type from the list of possible inputs @Ishticode ?

Ishticode (commented):
I think they recently added the decorator to handle this. For the frontend torch trace function, it is handled with the decorator like so 

```python
@with_unsupported_dtypes(
    {"1.11.0 and below": ("float16", "bfloat16")}, versions["torch"]
)
...
```

Try if you can make this work.

nassimberrada:
Not sure that changes it, the tests still take int8 inputs even with the decorator added to the frontend function

Ishticode (changes requested):
Actually we shouldn't class int8 as unsupported i think because torch.lcm does work with int8 natively.

What is the error you get in the tests?

nassimberrada:
The same error that popped up with ivy.lcm basically, here's an excerpt: 

ret_np = array([  0, 126], dtype=int8)
ret_from_gt_np = array([   0, -126], dtype=int8), rtol = 0.001, atol = 1e-06
ground_truth_backend = 'torch'

E           AssertionError: [  0 126] != [   0 -126]

I did introduce the abs function for the torch, numpy, and tensorflow backends when working on ivy.lcm so this shouldn't be the origin of the issue, that's why i'm thinking it might have to do with int8 inputs.

Ishticode (approved):

PR #5527, opened 2022-10-10
torch and jax frontend namespaces

hirwa-nshuti:
Hi, I updated the namespace for both `jax` and `torch` frontends.

Example:
-------------

As we were current calling `ivy.functional.frontends.torch.sigmoid` we will be calling `ivy.functional.frontends.torch.nn.functional.sigmoid` as done for `torch.nn.functional.sigmoid`
I updated the testing helper also to do the training calling the namespace `fn_tree` as some of the functions were overriding 

for example:
----------------

for jax there was a case where when testing the `ivy.functional.frontends.jax.numpy.linalg.eigh` was overriding the function `ivy.functional.frontends.jax.lx.linalg.eigh`


djl11 (approved):
all looks great, thanks a lot @hirwa-nshuti!

PR #5544, opened 2022-10-10
Fixed test_activations in TensorFlow frontend

KareemMAX:
Fixed tests in `test_activations.py`. Specifically `test_tensorflow_softmax` as is didn't support `float16` nor `bfloat16` in Jax and PyTorch


KareemMAX:
Found out that Jax and PyTorch actually support `bfloat16` so I think it is a casting issue, so I'm closing this PR.

PR #5547, opened 2022-10-10
Reformatting multiply

nilosreesengupta6:


AnnaTz:
Hi, there are errors in the docstring examples you have provided. You can see them if you click on Details next to test-docstrings and then search for "multiply".

AnnaTz:
Great! Only thing needed now is to format the array and container methods too. You can find those in ivy/array/elementwise.py and ivy/container/elementwise.py. You should make sure the parameters and returns are updated, and also that a couple of examples are added.

AnnaTz:
> > Great! Only thing needed now is to format the array and container methods too. You can find those in ivy/array/elementwise.py and ivy/container/elementwise.py. You should make sure the parameters and returns are updated, and also that a couple of examples are added.
> 
> Hello @AnnaTz ,
> 
> Thank you for your response. I have added the docstrings as you said to ivy/array/elementwise.py and ivy/container/elementwise.py. Kindly check once.

The array examples should showcase the array instance functionality. The whole point of having the ivy.Array instance method variants is to be able to make calls like `x1.multiply(x2)` instead of `ivy.multiply(x1, x2)`.

Also, you should update the Parameters and Returns docstrings in the array and the container files.

AnnaTz:
That's good but you should remove the previous examples in array.

AnnaTz (approved):
Great! Ready to be merged 🙂

PR #5564, opened 2022-10-11
Update activations.py

Raneem-Khafagy:
Changed args & docstring for softmax array method


JonathanYang-yzh:
Thank you for the commits!

PR #5584, opened 2022-10-11
numpy frontend docs - add casting handling subsection and update classes and instance methods subsection

juliagsy:


djl11 (approved):
looks great, thanks!

PR #5588, opened 2022-10-12
Added dropout3d to Ivy Functional API - PyTorch superset behaviour

popraf:
Issue dropout3d #5466 
Subtask Extend Ivy Functional API #3856 

The dropout3d function is only present in Pytorch as a primary function, the superset behaviour follows the PyTorch API.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ with_time_logs/
 internal_automation_tools/
 .vscode/*
 .idea/*
+.venv/*
--- a/.gitignore
+++ b/.gitignore
@@ -28,4 +28,3 @@ with_time_logs/
 internal_automation_tools/
 .vscode/*
 .idea/*
-.venv/*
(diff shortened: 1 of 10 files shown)


Daniel4078:
I found that tensorflow also have a similar dropout3d function: https://www.tensorflow.org/api_docs/python/tf/keras/layers/SpatialDropout3D

Daniel4078:
why not try using git reset to not commit or stage any change to the .gitignore file at all?

popraf:
Hello @Daniel4078 I will use git reset to not commit or stage any changes to the .gitignore, thanks for pointing it out. Regarding the dropout3d function; I see that every function in [functional/backends/tensorflow/layers.py] uses mostly tf.nn module, not a Keras one. I'm unsure if I should import keras layer or work around only in native tensorflow, as Keras is just built over tensorflow. I would be grateful for advice.

Issue #5596, opened 2022-10-12
rint


sagnik1511:
NumPy frontend

sreegandh804:
Hi @sagnik1511 , i would suggest you make a comment on the main issue that is linked to the function you selected with this issue number, which would assign you this function, Feel free to refer to our contributor's guide here : https://lets-unify.ai/ivy/contributing.html

PR #5598, opened 2022-10-12
add rint function for numpy frontend

sagnik1511:
#5596


sagnik1511:
Hi, I can't understand why the [E226 Error](https://github.com/unifyai/ivy/actions/runs/3232790379/jobs/5293839483) is coming on file `determine_test_coverage.py`. While running on its own machine it does not return any error. Can anybody help?

Nightcrab:
Since this PR doesn't affect `determine_test_coverage.py` that lint error is not relevant.

What I did notice is that you use `ivy.rint` in this frontend function but unless I'm mistaken ivy does not have a function called `rint`.

Nightcrab:
It would be good if you could also merge the master branch into this one because there is a bug with the frontend tests that was since fixed.

PR #5601, opened 2022-10-12
Adding eigh function issue# 5476

Ammar-Shah:


ivy/functional/frontends/tensorflow/linalg.py
```diff
@@ -16,7 +16,10 @@ def det(input, name=None):
 
 
 def eigh(tensor, name=None):
```
jkeane508, line 18: This was already implemented :)

jkeane508 (changes requested):
Hey can you let me know what issue this PR resolves?

ivy/functional/frontends/tensorflow/linalg.py
Ammar-Shah: Hi James! Thank you for your response. The change I made to the already implemented eigh function was that the tensorflow function returns two variables, i.e., the eigen values and the eigen vectors, while the ivy.eigh function returns one vairable that comprises of a tuple containing eigen values and eigen vectors of the input. So, to better mimic the tensorflow function, I separated the tuple into two variables.

jkeane508: Oh I see! My bad, that makes sense. Good find!

jkeane508: Did you implement the previous eigh() for the tensorflow frontend?

Ammar-Shah: Thank you for your feedback. I did not implement the previous eigh function, however, it was not present when I opened the issue.

jkeane508: It seems to be the test `test_tensoflow_eigh` is still failing, this would need to be fixed before I can merge this PR :)

jkeane508 (changes requested):

Ammar-Shah:
Close #5476

ivy-seed:

Issue #5616, opened 2022-10-12
add tanh activation function in tensorflow frontend


anaghazachariah:

MrZilinXiao:
Closing it since the corresponding PR was closed.

PR #5617, opened 2022-10-12
added tanh activation function in tensorflow frontend

anaghazachariah:
close  issue #5616


anaghazachariah:
@JamieLine Can you tell me whats going wrong?

JamieLine:
Hi!

The only errors I could reproduce locally were,

1. A `KeyError` caused by setting the `test_frontend_function` to call the Keras function, when regular `tf.tanh` will do, the two are the same function. This can be fixed by changing the `fn_tree` argument to `tanh`.
2. An error where `tanh` recieves an unexpected keyword argument `axis`. Your code generates a value, a dtype and an axis, but the axis isn't needed in the `tanh` function. The correct way to fix this is to change the call to `helpers.dtype_values_axis` to a call to `helpers.dtype_and_values` and edit the parameters of that call as needed, which should just be the removal of the axis related arguments.

The rest looks good to me though! If there are any other errors you've found, or if you have any questions or comments, please do let me know!

PR #5618, opened 2022-10-12
added numpy flatten functionality and tests

### LINKED ISSUE #5607: flatten

acrulopez:
Close #5607


ZoeCD:
Hello! Thank you so much for your contribution, and I'm sorry for the late reply. We were deliberating whether we should support the `order` parameter for numpy `flatten`. Here is a minimal example of how `order` works:
```
import numpy as np
a = np.array([[1,2], [3,4]])
print(a.flatten(order='F'))
[1 3 2 4]
print(a.flatten(order='K'))
[1 2 3 4]
```
The `order` parameter is significant as it dictates how the flattening is performed. We are working on integrating this functionality in `ivy.flatten`. So, you are free to replicate this functionality by composition in the numpy frontend function, or you could work in another function if you prefer. 

Again, thank you for your time. Please let me know if you have any questions!

PR #5627, opened 2022-10-12
heaviside extension

nassimberrada:


ivy/functional/backends/numpy/extensions.py
```diff
@@ -285,3 +285,25 @@ def moveaxis(
...
+    x2: np.ndarray,
+    /,
+    *,
+    out: Optional[np.ndarray] = None,
+) -> np.ndarray:
+    return np.heaviside(
+        x1,
+        x2,
+        out=out,
+        where=True,
```
Ishticode, line 301: We don't need to call on these args as they are set by default for numpy.
```python
        where=True,
        casting='same_kind',
        order='K',
        dtype=None,
        subok=True,
```

UNLESS there was some instructions updates that I missed or something then please feel free to notify me.

Ishticode (approved):
But these aren't really an issue. I guess I will just merge and will update these anyways later. I think there were some other functions which had these added if I remember correctly. I will fix them all or you can make PR removing these. Without delay I merge this now. Thanks @nassimberrada

ivy/functional/backends/numpy/extensions.py
Ishticode: There area already remved so no need to concern with. Thanks.

PR #5640, opened 2022-10-13
added tan to torch frontend and tested

PaoloCatti:


.idea/vcs.xml
```diff
@@ -2,5 +2,7 @@
 <project version="4">
   <component name="VcsDirectoryMappings">
     <mapping directory="$PROJECT_DIR$" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$/ivy_tests/test_array_api" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$/ivy_tests/test_array_api/array-api" vcs="Git" />
   </component>
-</project>
```
jkeane508, line 6: Can you please undo changes to this file, its not relevant to the PR

jkeane508 (changes requested):
Looks good to me, just need to undo changes to the vcs.xml file :)

.idea/vcs.xml
PaoloCatti: Yes of course! I believe undid the changes and it now has only two files changes :)

PaoloCatti:
Good morning!! Could you please help figure out why the check is always failing at “Check on failures” and it gives me the same output again and again: 
Run exit 1
  exit 1
  shell: /usr/bin/bash -e {0}

jkeane508:
> 

Hey,

So basically that result isn't the actual log from the tests, if you look through the tabs above that on the testing, one of them is the log for the errors. I believe its `run_frontend_functions` :)

jkeane508 (approved):
Tests seem to passing and it looks good to me!

PaoloCatti:
Oh this is amazing thank you for this I am so happy it all went well!! :)

PR #5654, opened 2022-10-13
add_instance_method_numpy_cumprod

Hellenn07:
#5653


Hellenn07:
#5654

Ishticode (changes requested):
Hi @Hellenn07 Thank you for the pr.
Could you kindly explain what is being added in this PR under which to-do list.
At the time of writing, it shows a single changed file which adds two tests for numpy frontend.  `test_numpy_add` and `test_numpy_tan` which are already found in ivy tests previously. Please let me know for any details and it will be appreciated. Thank you.

PR #5656, opened 2022-10-13
 ivy.Container namedtuple repr

Aarsh2001:
prints the same way as pointed out by this [pr](https://www.google.com/url?q=https://github.com/unifyai/ivy/pull/4507&sa=D&source=docs&ust=1665663923620266&usg=AOvVaw1wYH-o4Jcfpq_3w-xSPzad)


ivy/container/base.py
```diff
@@ -3685,6 +3685,7 @@ def _align_arrays(str_in):
                     and (self._ivy.is_native_array(v[0]) or isinstance(v[0], ivy.Array))
                 ):
                     rep = (
+                        v,
                         "list[{}]".format(len(v)),
```
djl11, line 3689: we always show `list` even if it is a `tuple`, we should show whichever is correct. Also, this just shows the full value as well as all of the stats about the `list` or `tuple`. We only want to print things differently if it's a named tuple I would say. This doesn't solve the problem. We don't want it to always show the entire list. If this contains 1000s of entries then we want to show the stats instead for example.

djl11 (changes requested):

djl11:
To explain this in a bit more detail, named tuples should also adhere to the `print_limit` parameter. If there are more than `print_limit` elements in the named tuple, then we should show something like this if the tuple contains arrays:

```
(NamedTuple[100], <insert_type>, shape=<insert_shape>)
```

If there are less than `print_limit` elements in the named tuple, then we should show the keys of the named tuple. With regards to the values of the named tuple, these should follow the same rules as usual. For example, a named tuple with two entries, each being short lists, might look like this, with the `print_limit` not kicking in at all:

```
(first_key=[1, 2, 3], second_key=[1, 2, 3])
```

However, a named tuple which contains large arrays might look like this:

```
(first_key=(<ivy.Array>, shape=[20]), second_key=(<ivy.Array>, shape=[30]))
```

And a mix would look like this:

```
(first_key=(<ivy.Array>, shape=[20]), second_key=[1, 2, 3])
```

Within the context of this explanation, whether something is deemed to be _long_ or _short_ depends on whether the size is greater than or small than `print_limit`. Does this make sense?

Aarsh2001:
Results with different cases for `print_limit`-:


Screenshot 2022-10-15 at 11 28 46 PM
Screenshot 2022-10-15 at 11 29 02 PM
Screenshot 2022-10-15 at 11 29 13 PM

djl11 (changes requested):
It's looking good in general, a few points though:

- `[(3, 3)]` is wrong, it should be `[3, 3]` or `(3, 3)`, can't remember which one off the top of my head.
- the brackets around the named tuple should be `(` and `)` instead of `[` and `]`, in keeping with the tuple representation format
- we shouldn't show `NamedTuple[<tuple_size>]` in cases where we expand the named tuple in the representation, such as the first example. We should only show `NamedTuple[<size>]` when `<tuple_size> > self._print_limit`, such as the last example.
- when`<tuple_size> > self._print_limit` we shouldn't show the list of named tuple keys, so the last example should not contain `(sign, logabsdet)`
- to be more specific regarding the last example, I actually think we should only collapse the named tuple when`<tuple_size> > self._print_limit`, but it looks like we are collapsing it when`<tuple_size> >= self._print_limit` based on the last example, can you double check this?

Once these points are resolved, I think it's good to merge. Please feel free to respond back with counter arguments or queries to any of the points made above! Thanks @Aarsh2001!

PR #5665, opened 2022-10-13
new extension function (rfft) has been added

### LINKED ISSUE #5501: rfft

Viditagarwal7479:
Close #5501


kimbring2:
@Viditagarwal7479 Thank you for your contribution. I just find one issue with your code. 

When the input is zero dimension such as the -1, the Numpy backend gives an error because of axis can not be the -1.

In my opinion, the dimension of input should be at least great than 0.

Viditagarwal7479:
Hey,
Thank You, sir, for your kind feedback.

Yes, this is covered in JAX, TensorFlow, Numpy, and PyTorch itself and they will raise error corresponding to such input.

> I have included this in the test so that the test cases are such that they have a dimension greater than 0.

**I deleted my previous fork thus, I have opened this new PR #5865**

Thanks 
Vidit

PR #5670, opened 2022-10-13
Implemented jax_numpy_zeros frontend and test

samuelbido:
#5630


Daniel4078 (changes requested):
Hello, currently there is a failure with your test for you to fix, everything else seems fine to me.
TypeError: zeros() takes 1 positional argument but 2 positional arguments (and 2 keyword-only arguments) were given

PR #5690, opened 2022-10-13
tf frontend namespace refactoring

hirwa-nshuti:
Some functions were left from the Tensorflow namespace like we were using `tf.add` but we are implementing `tf.math` functions.


jkeane508 (approved):
yea looks like a good change to me :)

PR #5693, opened 2022-10-13
Trapz extension

nassimberrada:


ivy/functional/backends/torch/extensions.py
```diff
@@ -251,3 +251,29 @@ def moveaxis(
...
+    y: torch.Tensor,
+    /,
+    *,
+    x: Optional[torch.Tensor] = None,
+    dx: Optional[float] = None,
+    axis: Optional[int] = -1,
+    out: Optional[torch.Tensor] = None,
+) -> torch.Tensor:
+    if x is None:
+        dx = dx if not None else 1
```
Ishticode: shouldn't this be `dx = dx if dx is not None else 1`

Ishticode (changes requested):
the little push fixes torch but i am not sure ... will require migrating to the new extension sub directory else the conflicts will be many

PR #5707, opened 2022-10-14
fmax extension

nassimberrada:


Ishticode (changes requested):
Hi. Mostly looking good. Just few queries before merging.

1. Could you explain slightly in the docstring how this is differnt to `ivy.maximum`(handling NaN) or anything else.
2. Are you sure `tf.math.maximum` works as expected.
I tried this code with native tf
```python
>>> x = tf.constant([0., 3, 4.])
>>> y = tf.constant([np.nan, 3., 0])
>>> tf.math.maximum(x,y)
<tf.Tensor: shape=(3,), dtype=float32, numpy=array([nan,  3.,  4.], dtype=float32)>
```
Not sure why `nan` was considered the maximum vs. 0. as opposed to the examples in ivy functional.
Please feel free to enlighten me.

nassimberrada:
I guess the test fails with integers input because tf.math.is_nan only takes floats, i restricted the input values to be only floats and the tests pass this way, is this fine ?

Ishticode (changes requested):
Looks fine now I think. Could you deal with the conflicts and perhaps maybe remove `where` to `subok` arguments in np call which I think are already set by default. Let me know when done and can hopefully be merged.

Ishticode:
I will get this merged after the conflicts are resolved. Please ensure the lints are checked in case..

PR #5718, opened 2022-10-14
Flatten_extension_corrections

nassimberrada:


Ishticode (changes requested):
The torch test failed for me with the error
`flatten() has invalid args: start_dim cannot come after end_dim`

I think the the assignment `start_dim, end_dim = axes[0], axes[1]` in the test is supposed to be more like `start_dim, end_dim = min(axes), max(axes)`. But Not that exactly bcz you also need to take account of the negatives. My point is that at the moment it seems like that there is no ensuring that `start_dim` is before the `end_dim` which leads to failure case.

Could you try that and let me know if any queries. Thanks Nassim.

nassimberrada:
I think it should be corrected now, could you check on your side ?

PR #5752, opened 2022-10-15
Fmod extension

nassimberrada:


ivy/array/extensions.py
```diff
@@ -341,3 +341,43 @@ def heaviside(
...
+        Returns
+        -------
+        ret
+            Array with element-wise remainder of divisions.
+
+        Examples
+        --------
+        >>> x1 = ivy.array([2, 3, 4])
+        >>> x2 = ivy.array([1, 5, 2])
+        >>> ivy.fmod(x1, x2)
```
Ishticode, line 375: should be instance method use here

ivy/functional/backends/numpy/extensions.py
```diff
@@ -297,3 +297,25 @@ def heaviside(
...
+) -> np.ndarray:
+    return np.fmod(
+        x1,
+        x2,
+        out=None,
+        where=True,
+        casting="same_kind",
+        order="K",
+        dtype=None,
+        subok=True,
```
Ishticode, line 317: No need for these args
```python
        where=True,
        casting="same_kind",
        order="K",
        dtype=None,
        subok=True,
```

Ishticode (changes requested):
Minor changes then I will link this with an issue and merge....

Ishticode (changes requested):
This was almost ready but due to recent merging of the other PR i think there is more conflicts. Also there is `median` and `ndenumerate` functions added/modfied? is that intentional?

nassimberrada:
I don't think i worked on those, probably someone else added those extensions ?
I solved the conflicts, here's the link to the corresponding issue fmod #5793

Ishticode (approved):

PR #5764, opened 2022-10-15
Added arccos #5248

KaishvaShah:


sharique1006:
Hi @KaishvaShah you need to write tests for your function as well.

Issue #5773, opened 2022-10-15
deserialize


oscar1marte:

modanesh:
`deserialize` test is failing in the TF frontned

PR #5782, opened 2022-10-16
added deserialize in tensorflow frontends

oscar1marte:
#### **Reference**
#5773

#### **What does this implement?**
deserialize is a function that returns an activation function given a string identifier name.

#### **Comments**
I was wondering...
- Is there a problem that I imported sys and inspect, to perform this function?
- Is there a better way to extract the available activation functions than using sys.modules? or any improvement that I should do?
- Also, I have some doubts regarding the test of this function, since this function receives a string as input value, and returns a function. Is there an example of a function with a similar test that can inspire me?


YushaArif99:
Hi @oscar1marte.

Regarding the points you've raised: 

**Q: Is there a problem that I imported sys and inspect, to perform this function?** 
A: No, this is not a problem at all. However, a point worth mentioning here is that the purpose of the frontend functions is to enable transpilation during ivy's automatic code conversion. Hence, the frontend functions, in general, resort to ivy's implementation of the function rather than implementing the function itself. Thus, there is no need to implement the `deserialize` function in the frontend. Rather, the frontend would simply resort to ivy.deserialize (for eg).

**Q2: Is there a better way to extract the available activation functions than using sys.modules? or any improvement that I should do?**
A: Not so sure about it. Different backends have different ways of performing serialization and deserialization. Jax has the Flax API(_flax.serialization.to_bytes() and flax.serialization.from_bytes()_ ) which can serialize/deserialize high-level classes, Pytorch has TorchScript (_torch.jit.load() and torch.jit.save()_ ) which serialized/deserializes models as a whole, and similarly TensorFlow has the keras API which also has its own class-specific serialize/deserialize functions (eg _tf.keras.layers.serialize, tf.keras.optimizer.serialize). In this particular case, I think `ivy.activations.deserialize()` could be implemented as a mixed  function. Read the following section of the guide for more details( https://lets-unify.ai/ivy/deep_dive/1_function_types.html#mixed-functions)

**Q3: Also, I have some doubts regarding the test of this function, since this function receives a string as an input value, and returns a function. Is there an example of a function with a similar test that can inspire me?**
A: I don't think we currently have a similar test that you could refer to. In such a case, it's best to create your own composite function to do the testing. An example would be to randomly draw a string from a dictionary that contains all activation functions, and then do some form of validation (value testing, etc) on the deserialized function as well as the original function to see if they both behave the same.

oscar1marte:
Hello @YushaArif99 

Great, now I understand much better what I have to do.
I have a few last questions.

- I'm going to need to move this implementation from _ivy/functional/frontends/tensorflow/keras/activations.py_ to _ivy/functional/ivy/activations.py_, which is where that mixed function should be, right? (To then use ivy.deserialize on the frontends)
- Currently I didn't find any implementation of the serialize function in the Tensorflow backend (_ivy/functional/backends/tensorflow/activations.py_). Should I implement that backend function too? If so, will I need to create another issue to implement this activation deserialize?

Thank you very much!

oscar1marte (commented):
Hello,
I asked some questions in the comments to better understand.

YushaArif99:
Hey @oscar1marte 
Apologies for the late reply. I was swamped with work this week and so reviewing this PR totally slipped under my radar.
I'm not sure what's going on right now because there are a ton of commits from other members on this PR for some reason 😂. 
But nonetheless, regarding the changes you made, since TensorFlow does have a deserialize activation function, rather than using Ivy's implementation, you should invoke the backend-specific implementation in this case. For other backends, Ivy's implementation would be used. Also, you haven't implemented the test for this function. Could you kindly write a simple test to ensure that Ivy's compositional implementation is indeed working as it's supposed to? Other than that, I think the changes seem good.

MahmoudAshraf97:
deserialize functions needed in #6062, let me know if there's any thing I could do to accelerate the merging of this PR

oscar1marte:
Hello @MahmoudAshraf97

Sorry for the delay. I'm almost done with the deserialize test, I'm going to do everything I can to get everything ready for today.

oscar1marte (commented):
Hi @YushaArif99

Now I finished implementing the deserialize and the test function. I made some helper functions to be able to perform the test correctly. Also I added an deserialize function on backends/tensorflow.
Should I add or modify something?
(A thousand apologies for the delay 😞)

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -1644,3 +1644,45 @@ def _is_frontend_array(x):
         or isinstance(x, tf_tensor)
         or isinstance(x, DeviceArray)
     )
+
```
YushaArif99, line 1647: I believe this is a helper function that is unique to the `test_tensorflow_deseralize()` function. In these cases, It's best to implement such functions in the test file itself rather than in the helpers. Refer to the following section of the guide for more details.
https://lets-unify.ai/ivy/deep_dive/ivy_frontends_tests.html#hypothesis-helpers

ivy_tests/test_ivy/helpers/testing_helpers.py
```diff
@@ -231,3 +232,15 @@ def new_fn(
 @st.composite
 def seed(draw):
     return draw(st.integers(min_value=0, max_value=2**8 - 1))
+
```
YushaArif99, line 235: The same is the case here as well.

YushaArif99 (changes requested):
Hi @oscar1marte. 
Apologies from my side as well 😔. I have really been busy with work this past week and just haven't had the time to review PRs. But expect an early reply from now on. This PR has been dragged on for way too long because of my inadequacy 😅. Btw, could you kindly make the small changes requested? Other than that, everything seems fine and the PR is ready to be merged!

oscar1marte (commented):
Hi @YushaArif99 
I made the small changes requested, and now I think everything looks good.

oscar1marte:
> Hi @oscar1marte. Apologies from my side as well pensive. I have really been busy with work this past week and just haven't had the time to review PRs. But expect an early reply from now on. This PR has been dragged on for way too long because of my inadequacy sweat_smile. Btw, could you kindly make the small changes requested? Other than that, everything seems fine and the PR is ready to be merged!

Hey @YushaArif99 

Don't worry :grin:, it is normal, and there is no problem at all. As my brother usually says, this is a marathon, not a race.
:blush:

YushaArif99 (approved):
Hey @oscar1marte, thanks for cooperating 😊
The changes seem good to me. I will now merge this PR. Thanks a lot for your contribution!

PR #5786, opened 2022-10-16
Add jax.numpy.linalg to Jax frontend

mohame54:
#4614

### DIFF
--- a/ivy/functional/frontends/jax/numpy/linalg.py
+++ b/ivy/functional/frontends/jax/numpy/linalg.py
@@ -55,9 +55,9 @@ def norm(x, ord=2, axis=None, keepdims=False):
 
     # checking if the axis has 2 dims
     if (type(axis) in [tuple, list]) and len(axis) == 2:
-        return ivy.matrix_norm(x, ord=ord, axis=axis, keepdims=keepdims)
+        return ivy.matrix_norm(x=x, ord=ord, axis=axis, keepdims=keepdims)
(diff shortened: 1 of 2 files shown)


guillesanbri:
Hi! First of all, thank you very much for contributing to Ivy! Some things to review: 
- You have made a comment with a reference to the general ToDo list, you should make one referencing the issue you opened following the format `Close #issue_number`.
- The `fw` keyword is not used and should be removed from the `test_fax_norm` fn.
- As far as I know, `norm` should support other dtypes, so I'm not sure why did you include the `norm.supported_dtypes` bit. If I am missing something we can discuss it  😁
- In the test, as the `ord` keyword of the `norm` function accepts multiple values, we should be testing for all of them (i.e. `-inf`, `fro`, `nuc`). you can check out the [JAX](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.norm.html) and (specially) the [NumPy](https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html#numpy.linalg.norm) docs to see how the function should behave in each case.

Feel free to ask any question and/or clarification!

mohame54:
Thanks for answering me, as for norm.supported dtypes if i removed it the dtype assertion error that  would occur as i tried before and i will remove fw param.

mohame54:
What about the vector norm it doesn't support any of 'fro' or 'nuc'

PR #5796, opened 2022-10-16
feat: add numpy.ndarray.min frontend

### LINKED ISSUE #5791: min

vohoanglong0107:
close #5791

### DIFF
--- a/ivy/functional/frontends/numpy/ndarray/ndarray.py
+++ b/ivy/functional/frontends/numpy/ndarray/ndarray.py
@@ -57,3 +57,17 @@ class ndarray:
 
     def argsort(self, *, axis=-1, kind=None, order=None):
         return np_frontend.argsort(self.data, axis, kind, order)
+
+    def min(self, *, axis=None, out=None, keepdims=False, initial=None, where=True):
+        return np_frontend.amin(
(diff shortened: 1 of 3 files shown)


kimbring2:
@vohoanglong0107 Hello, Vohoanglong0107. Thank you for your contribution. 

You need to set the value for the argument of the function for testing such as x in the test_numpy_instance_min function.

vohoanglong0107:
@kimbring2 Thank for the feedback.

By setting the value for x, you mean place some constraint on the value that the testing function get when using helpers.dtype_value_axis, such as the min/max dim, min/max value, am I correct ?

kimbring2:
@vohoanglong0107 I just meant the argument of helpers.test_frontend_method function. However, it seems like the test_numpy_instance_argsort test also can not pass the test. 

I am not sure which way is the correct way because every test of the same file failed. It seems like I need some time to make one example for testing it.

vohoanglong0107:
@kimbring2 Have you found which is the correct way ?

kimbring2:
@vohoanglong0107 Yes, I also check that we should use the data value of the ndarray class like your code. It looks like a good idea to merge your code.

PR #5805, opened 2022-10-16
added solve function in jax frontend

oscar1marte:
#### Reference:
#5783

#### What does this implement?
Solve a linear matrix equation, or system of linear scalar equations.

#### Comments
- There is something that I would like to improve in the test, but I still don't see how, and it is in relation to the **random_dim** variable that I made. Since I need the column dimension of the square matrix "**a**" to be equal to the row dimension of the matrix "**b**", I cannot use **helpers.ints** to generate its shapes separately, because they would generate different values. That is the reason why I made a variable called **random_dim**, which I pass to the **min_value** and **max_value** of the **helpers.ints**, so that the criteria I mentioned before are met. How could I get rid of that variable, and make the criterion be met?

Thank you very much!

### DIFF
--- a/ivy/functional/frontends/jax/lax/linalg.py
+++ b/ivy/functional/frontends/jax/lax/linalg.py
@@ -33,3 +33,8 @@ def eigh(x, /, *, lower=True, symmetrize_input=True, sort_eigenvalues=True):
         x = symmetrize(x)
 
     return ivy.eigh(x, UPLO=UPLO)
+
+
+@inputs_to_ivy_arrays
+def solve(x1, x2):
+    return ivy.solve(x1, x2)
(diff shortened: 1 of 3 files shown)


modanesh:
Thanks for you contribution :)
I see your point. For that matter you may write a helper function that solves the problem.

oscar1marte:
Great, where should I put that function? 
It would be in _ivy/ivy_tests/test_ivy/helpers/hypothesis_helpers/array_helpers.py_, correct?

modanesh:
> It would be in ivy/ivy_tests/test_ivy/helpers/hypothesis_helpers/array_helpers.py, correct?

Yes. Please ping me whenever your PR is ready to be reviewed :)

oscar1marte (commented):
Hello 😄  
I would like to merge this branch, the tests work fine.
I plan to do the upgrade in the future to use a helper function to receive two arrays that I mentioned before, but for now I would like to do the merge, and focus on another task that I have open.

oscar1marte:
Although now that I think about it, I have an idea that could work without much complication.
Don't do the merge yet haha
I plan to generate an array of (n, n+1), and then from that array I can output the (n, n) array and the (n, 1) array.

oscar1marte:
It worked @modanesh
Now the code is cleaner, and I like it :)

modanesh:
Very well done! Thanks for your contribution @oscar1marte :)

Issue #5480, opened 2022-10-09
__pow__


AbhishekJha22:

modanesh:
reopening the issue since the `pow` test is failing in the TF frontned

AbhishekJha22:
Is the issue still assigned to me? If that is the case, I will take a look at it ASAP. Thanks!

PR #5811, opened 2022-10-16
Added __pow__ method for tensorflow frontend 

AbhishekJha22:
Issue #5480


modanesh:
Thanks for your contribution :)
Could you please write a test for the implemented method as well here `test_frontends/test_tensorflow/test_tensor.py`? Thanks!

AbhishekJha22:
Sure! I will get on it.

AbhishekJha22:
Hi @modanesh, I have got a couple of questions:
1. Should I be worried about all the tests that fail on github upon a push from me? Some of them list the source of errors as files that I haven't even touched.

2. Would I be correct in assuming that the test functions that I write for my issue will be run on github upon my push? I have not worked with pytest/hypothesis before and was wondering if leaving it to github might be a better option?

AbhishekJha22:
Thank you! I have gone through the documentation for tests and barring a couple of points, I was able to understand it. I tried running my own tests and based on the falsifying examples that were returned, I have a question - the falsifying examples all had to do with raising an integer to a negative integer (for example [0] ** [-1]). Tensorflow does not allow it, neither does Numpy. If it had been an entire type, say int32, I could have removed it from supported_dtypes tuple. How do I prevent raising an integer to a negative number?

modanesh:
Removing it from the supported_dtypes tuple would be fine. As it appears, your changes are fine regarding the `test-frontend-tensorflow`. You just need to take care of the `lint` tests, then your commit will be merged.

AbhishekJha22:
Hi @modanesh! I made a minor change to my implementation to make sure the function worked for all combinations of scalar and tensor values. The errors that pytest returned were mostly about raising integers to negative integers. I worked with a few examples manually and the implementation seemed to be working fine. About the lint formatting errors, I took care of all that came from my implementation. There are two errors left - one error is about not using an import statement that I did not remove and another one is about a blank line at the end of frontends.tensorflow.math.py which did not go away even after modification. 
Thank you for your patience while working with me! This was my first experience contributing to an open-source project and hopefully, things would go faster in the future. I look forward to a long collaboration on this project!

modanesh:
Thanks! You did a good job as your first contribution.
I made a few small changes to resolve conflicts with master and also pass the lint tests. When the test-frontend-tensorflow is done (hopefully with no errors), then it should be good to merge :)

modanesh:
According to the test-frontend-tensorflow, `test_tensorflow_instance_pow` is failing. Could you take a look at it?

AbhishekJha22:
Did you mean this:
Falsifying example: test_tensorflow_instance_pow(
    dtype_and_x=(['int32', 'int32'],
     [array(0, dtype=int32), array(-1, dtype=int32)]),
    as_variable=[False],
    native_array=[False],
    fw='tensorflow',
)

This example has -1 in the second array and raising an integer to a negative integer isn't allowed in TensorFlow or NumPy. For PyTorch it is allowed and the same operation of raising [0] to [-1] works fine on my machine and produces [0] as an answer. I think this is the source of error for the failing example.

Apart from this falsifying example, I could not find any other reference to test_tensorflow_instance_pow  in the logs that was preceded by 'Falsifying example:' in the frontend tests. Should I also look into these other references?

AbhishekJha22:
Hi @modanesh, could you take a look at my previous comment about test_tensorflow_instance_pow? Thank you!

modanesh:
Yes, the only issue is with the falsifying example. If that's not permitted in TF and Numpy then I'd suggest to change to something that is permitted, so we could verify the results. Otherwise it looks good to me.

AbhishekJha22:
Thank you for your response! I want to clarify the modification that I have to make here - it is my understanding that the tests compare the result of my implementation with what tensorflow returns for the same method called on the same arguments. So even if I change my implementation to something that allows taking a negative power, say, by type casting the parameters to float in my code, the tensorflow backend would still raise an exception during tests because tensorflow would still recieve negative integers because that is what the frontend function is being called with during the tests. 

In fact, I tried converting the type of the second parameter to float on my machine but the errors for x = [0] and y = [-1] did not go away. 

And even if the errors did go away, would this kind of sneaky type conversion really be advisable? Wouldn't it change the expected behavior of the backend to returning numeric values when numpy and tensorflow would clearly have raised an exception?

AbhishekJha22:
Hey @modanesh! Could you please tell me if you recommend modifying my implementation to allow for raising an int to a negative int? I have the following arguments against it (you can also look at my previous comment )- 

1. It wouldn't resolve the error as the Ivy TensorFlow frontend result would still be compared with the result of the native TensorFlow framework which would still raise an exception for taking a negative power of an integer.
2. The result that Ivy TensorFlow frontend would return as a result of this modification would now be in contradiction with what the native TensorFlow or NumPy backend would return. As it stands, my code reproduces the expected backend behavior of different native frameworks more closely.

modanesh:
Hey @AbhishekJha22 , sorry for late reply.
For the `pow()` function, we're getting the same error when we run the following code meaning the frontend and backend seem to be consistent:
```python
import ivy

ivy.set_backend("tensorflow")
x1 = ivy.array([0], dtype="int32")
x2 = ivy.array([-1], dtype="int32")
print(ivy.pow(x1, x2))
```
So, I think we shouldn't be generating negative numbers if the datatype is an integer as it's something that's expected of tensorflow to not work in that case. So you should change the example generator for your test to avoid generating negative numbers as exponents.

AbhishekJha22:
Understood. I will modify the test functions.

AbhishekJha22:
Hi @modanesh! I have fixed the test functions to exclude negative integers and pushed the commit. There is a lint error "ivy/functional/frontends/jax/numpy/name_space_functions.py:311:1: W391 blank line at end of file" but it is not from any files that I have modified. There are other docstring and formatting errors none of which came from my files. I hope this commit will do!

Also, should I be modifying others' files to get rid of minor formatting errors going forward?

modanesh:
If errors are not coming from the files you changed, then you shouldn't be worried about them.

Just a quick question, why the `__rpow__` test is removed from the `test_tensorflow/test_tensor.py`?

AbhishekJha22:
I implemented the __pow__ method, not __rpow__.

modanesh:
I know. `rpow` test used to be in the `test_tensorflow/test_tensor.py` file (https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py#L1010). It's now replaced with the `pow` test (https://github.com/unifyai/ivy/blob/55c792a04e59d2592ecf72dae91be85778a5d90f/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py#L1010).

AbhishekJha22:
I only added  test_tensorflow_instance_pow test, and never deleted anything.

modanesh:
Ah I see. Here's what happened. The PR working on `rpow` opened 15 days ago and merged into master. This PR opened 25 days ago. So there was no `rpow` implementation at the time, and that still is the case. You implemented `pow` and GitHub's diff checker thinks this is to update `rpow`, but it's more of a conflict with master. 
I'll add `rpow`'s test, and then merge the PR.

AbhishekJha22:
The tests show test_tensorflow_instance_pow to be failing for the case when 0 is being raised to 0. However, I checked for this operation right now and my implementation returns 1 for all backends, which is the correct value. Also, I never encountered this error when I ran the tests and I ran multiple tests. How should I proceed?

modanesh:
Are you having the latest version of Ivy's test pipeline on your local machine?

AbhishekJha22:
I am not sure. I haven't synced my repository with Ivy's in a while. 

Are these failures a serious problem? The examples the test failed for are executing fine.

AbhishekJha22:
I did a git pull right now and there is one change in the `ivy.functional.frontends.tensorflow.tensor` - the class `Tensor` has now been replaced with the class `EagerTensor`. I think this is the source of  `AttributeError: module 'ivy.functional.frontends.tensorflow'  has no attribute 'Tensor'` that my tests have been generating.

There is one place where I use the reference `Tensor` in my tests - in the `class_name` value of my  `test_tensorflow_instance_pow` function. I did try changing it to `EagerTensor` but the error did not go away.

And I checked the Ivy repository just now and the `class_name` is still set to `Tensor` for the tests there.

modanesh:
As I cloned this PR and ran tests, all of them failed with the errro `AttributeError: module 'ivy.functional.frontends.tensorflow' has no attribute 'Tensor'`. Easiest workaround I believe would be to rebase your PR with the latest changes to the Ivy so we can verify it better.

AbhishekJha22:
I believe this to be the source of the error - in line 1360 of ivy/ivy_tests/test_ivy/helpers/function testing -  https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/helpers/function_testing.py#:~:text=frontend_class%20%3D%20importlib.import_module(frontend), the native module of tensorflow is being imported and being stored with the reference `frontend_class.` Then in line 1363 which is - ` frontend_class = getattr(frontend_class, c_n)`, we are searching for the EagerTensor module ( `c_n` stores the string `'EagerTensor'` ) in the native tensorflow namespace which results in an `AttributeError` as EagerTensor is present at `tensorflow.python.framework.ops.EagerTensor`.  I think this happened because native tensorflow Tensor could be accessed from the tensorflow namespace  with tensorflow.Tensor which is why the same code worked in previous versions. When the class was renamed to EagerTensor, the code broke.

AbhishekJha22:
To add another point - if `class_name` is  `Tensor`, then the source of error is line 1362 in the same file I linked to in the previous comment and the error is `AttributeError: module 'ivy.functional.frontends.tensorflow' has no attribute 'Tensor'` as in this case, the `Tensor` module can be accessed from the native tensorflow namespace but this name then does not match `EagerTensor` in the Ivy's frontend tensorflow implementation.

When `class_name` is `EagerTensor`, the tests fail on line 1363 with the error `AttributeError: module 'tensorflow' has no attribute 'EagerTensor'` for the reason I mentioned in the previous comment. However, in this case, line 1362 works fine because the name `EagerTensor` can be found in Ivy's frontend implementation for Tensorflow.

In the former case, the `AttributeError` comes from Ivy's tensorflow frontend, in the latter case the same error arises from the native tensorflow implementation.

AbhishekJha22:
The required modification turns out to be not very difficult - I think all that is required is to add the key-value pair `"EagerTensor": tf.constant` in the dictionary `ARR_INS_METHOD` defined on line 1213 in the same file - https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/helpers/function_testing.py#:~:text=(input_dtypes_method)-,ARR_INS_METHOD,-%3D%20%7B. 

It is working for me so far. I will keep testing in the meantime.

AbhishekJha22:
I modified the `class_name`  from `Tensor` to `EagerTensor` for all 36 test functions after modifying the `ARR_INS_METHOD` dictionary that I mentioned in the previous comment. On my machine `AttributeError` errors from earlier are no longer being generated. The only errors were `AssertionError` and  `hypothesis.errors.MultipleFailures` for different tests in different runs, none of which have been for the `__pow__` method so far.

modanesh:
Sounds great. You've put a lot of effort into this, much appreciated. 
Let's see how tests goes. 🤞

modanesh:
The tests are passing. Thanks for your great contribution to Ivy :)

AbhishekJha22:
Thank you for all your help! Loved working on the problem and hopefully, this leads to a long collaboration!

PR #5815, opened 2022-10-17
Docs frontends import unification

hirwa-nshuti:
updated the docs and files to import `<framework_name>_frontend`
Did some corrections in frontend tests docs on `fn_name` including the full namespace of the function


jkeane508 (approved):
Looks good to me :)

PR #5825, opened 2022-10-17
Dlpack

aditya-pola:
DON'T MERGE YET.
- [x] \_\_dlpack\_\_ implementation in ivy/array
- [x] to_dlpack implementation in backends
- [x] Changes to from_dlpack behaviour in backends
- [ ] \_\_dlpack\_device\_\_ implementation

About implementing \_\_dlpack\_device\_\_:
According to the array-api specification, the \_\_dlpack\_device\_\_ method returns the device type and device id and there are no alternatives to this method (\_\_dlpack\_\_ had to_dlpack as an alternative which was avaiable in all frameworks)

Currently dlpack device is implemented only for torch and numpy, I am unsure how to simulate that behaviour for jax and tensorflow

torch example:
>>> a = torch.Tensor([0,1])
>>> a.__dlpack_device__()
(<DLDeviceType.kDLCPU: 1>, 0)
I don't know how to access device type to return the above for jax and tf


mattbarrett98 (commented):
`to_dlpack` doesn't return a single array, it returns a pycapsule object, so it shouldn't contain the `out` argument. Also since `to_dlpack` doesn't return an array the `to_native_arrays_and_back` isn't necessary, it only needs to use the `inputs_to_native_arrays` decorator instead. 

Maybe it would be useful to write a test, to verify that the `from_dlpack` function from each backend works with ivy arrays containing data from each backend.

aditya-pola:
@mattbarrett98 that makes sense, I added the out argument for consistency cause I wasn't entirely sure how to handle that.
I'll write the test.

aditya-pola:
Please disregard the files with added test_gradients, I'll remove them from the PR before merging

ivy-seed:

mattbarrett98 (changes requested):
Looks good, but the tests for `to_dlpack` and `from_dlpack` are currently failing if you could fix them

ivy-seed:

PR #5831, opened 2022-10-17
Nll loss

### LINKED ISSUE #5083: nll_loss

mibo8:
Close #5083


Ishticode (commented):
Hi @mibo8 
Thanks for the PR. I see some good compositional effort here for nll_loss torch frontend.
But it would be more practical if we had implemented  ivy.nll_loss and then we could simply call ivy.nll_loss which will work for tensorflow, jax and numpy frontends also. Before we move forward with that, do you think you can possibly implement the backends for this? let me know and we shall see

ivy/functional/frontends/torch/nn/functional/loss_functions.py
```diff
@@ -134,3 +134,32 @@ def smooth_l1_loss(
...
+        input, target,
+        weight=None,
+        size_average=None,
+        ignore_index=- 100,
+        reduce=None,
+        reduction='mean'
+):
+    probs = ivy.exp(input)
+    depth = input.shape[-1]
+    one_hot_target = ivy.one_hot(target, depth)
```
mibo8, line 149: Convert target from indeces to one hot encoding to use `ivy.cross_entropy`

ivy/functional/frontends/torch/nn/functional/loss_functions.py
```diff
@@ -134,3 +134,32 @@ def smooth_l1_loss(
...
+    ret = ivy.cross_entropy(one_hot_target, probs, reduction='none', epsilon=0.0)
+    if weight is not None:
```
mibo8, line 151: handle the strange way in which pytorch applies mean reduction when we pass weights as argument, so we cannot apply directly the reduction in this case.  See [Pytorch docs](https://pytorch.org/docs/stable/generated/torch.nn.NLLLoss.html)

mibo8:
> Hi @mibo8
> Thanks for the PR. I see some good compositional effort here for nll_loss torch frontend.
> But it would be more practical if we had implemented ivy.nll_loss and then we could simply call ivy.nll_loss which will work for tensorflow, jax and numpy frontends also. Before we move forward with that, do you think you can possibly implement the backends for this? let me know and we shall see

Hi @Ishticode , IMHO there's no need for a backend function.  Most of the compositional effort is due to Pytorch differences with other frameworks (mainly in how the reduction is applied). I leaved you some code comments to explain what I mean.
Moreover, `nll` is not implemented in tensorflow and Keras due to its correlation with `cross_entropy`.
Of course, if you think that  `ivy.nll` can be useful I can try to implement it, but I think it will not reduce the compositional effort of nll torch frontend if I implement it similarly to `ivy.cross_entropy` (that is, with one-hot encodings targets, without "weighted mean" reduction). 
One more thing, I guess the weighted mean reduction should be implemented in the already merged pytorch loss as well.

Ishticode (changes requested):
Hi @mibo8 
Sure. I can move forward with this composition and I understand your well put reasoning above. But just something to keep in the back of one's mind anyway is to try for backends first except in some cases. 

Anyways, I tried running the test and it failed.... did you try the test? 
Let me know if you have a follow up. Thanks Mibo 🙂.

Ishticode:
Just a reminder @mibo8 
Not sure you waiting for me to review but there was no review requested so I didn't really touch anything....

mibo8:
HI @Ishticode , I was aware of the absence of the review request. It seems that in the last commits the frontend test functions have changed. Locally, I'm receiving some strange TypeErrors that seem unrelated to my function (the helper function is unable to convert from numpy to torch tensors). Testing it manually it works as expected for all frameworks.  Let me know if it can be ok or if I need to find a solution for the test.

Ishticode (changes requested):
Hi @mibo8 
Sorry for the delayed response. I will try to follow directly. Please feel free to message me on discord too. 

As for this PR, the test still fails on my end... Are you using the remote docker interpreter for your development?

PR #5892, opened 2022-10-19
Individual task: generalization of aliases testing

hirwa-nshuti:
Hello,
Generally, alias functions have the same behavior as the original functions.
To generalize their tests and to make sure that we are testing them also. I modified the `test_frontend_function` and when we are testing the original function we will test all its aliases in that frontend by providing the namespace to that alias. 

The tests are done the same way we test for frontend functions. I included some code lines to inner functions to avoid repetition of statements and added two main parameters to the helper function.
`has_aliases`: When the function has aliases we will set it to `True`
`all_aliases`: a list that holds the namespaces of all aliases to that function.

Whenever we make an alias we will go and update the unit test of that original function to include the namespace of the newly introduce alias.

Thanks🙂


ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -420,6 +422,12 @@ def test_frontend_function(
     with_inplace
         if True, the function is only tested with direct inplace update back to
         the inputted array and ignore the value of with_out.
+    has_aliases
```
djl11, line 425: This is not necessary, let's just use `all_aliases`. If it is `None` then we can assume that there are no aliases.

djl11 (changes requested):
looks good to me, just a small request to remove a redundant extra argument. Will request another review from @CatB1t given that he's working on this code quite a lot.

djl11:
Once @sherry30 and @mustafa1adel have given the green light, it's good to merge!

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -479,37 +483,58 @@ def test_frontend_function(
...
-    function_dict = ivy.functional.frontends.__dict__[frontend]
+    def _get_function(fn_tree):
+        # parse function name and frontend submodules (jax.lax, jax.numpy etc.)
+        *frontend_submods, fn_name = fn_tree.split(".")
+        function_dict = ivy.functional.frontends.__dict__[frontend]
+
+        # Getting function attributes when we have function tree such as
+        # nn.functional etc
+        len_frontend_submods = len(frontend_submods)
+        if len_frontend_submods > 0:
```
CatB1t, line 494: We can simplify this by using `importlib.import_module`, we don't have to manually trace it.

CatB1t (commented):
Just one question, why do we need to test the function again? when assigning an alias basically both names will bind to the same memory location, so for example `mean_absolute_error` and `mae` will bind to function `0x01` in memory for example, it doesn't do any copying or anything, I'd suggest just testing if alias is correctly binding to the original function and it's not misspelled or assigned to another function by mistake.

ivy_tests/test_ivy/helpers/function_testing.py
hirwa-nshuti: Yeah sure we can simplify it and do proper concatenations for frontend modules strings

hirwa-nshuti (commented):
Well from what I see we want to make sure that the aliases from ivy frontends are providing the same results as the corresponding one from the target framework, and you're right the memory binding for both `mean_absolute_error` and `mae` will be the same as they're one function technically. The reason, why I think we can do the testing again, is for example in the TensorFlow frontend we can have `tf.eig` and `tf.linalg.eig` which I think when we implement one as alias of the other we will need to test our function against both `tf.eig` and `tf.linalg.eig`
Thanks:)

CatB1t:
@hirwa-nshuti I get your point on making sure the alias actually `are` aliases, but in that case we can compute 1 more and do 3 arrays comparison, instead of running 1 more redundant test. Regarding the second example, I'm not sure how `tf.linalg.eig` can be an alias for `tf.eig`, but even though, we shouldn't run **our** function twice.

CatB1t:
Anyways, I think the optimization can be done later as only few functions are affected by that double-run instances, should keep in a TODO list though, once @sherry30 gives the greenlight feel free to merge 👍🏻

sherry30 (changes requested):
Hi,

Just a small change requested, the rest looks good to me.
Feel free to merge after this change.
Thanks

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -521,193 +546,226 @@ def test_frontend_function(
...
+        # value tests, iterating through each array in the flattened returns
+        value_test(
+            ret_np_flat=ret_np_flat,
+            ret_np_from_gt_flat=frontend_ret_np_flat,
+            rtol=rtol,
+            atol=atol,
+            ground_truth_backend=frontend,
+        )
+
+    def _is_frontend_array(x):
```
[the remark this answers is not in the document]
hirwa-nshuti: Sure let me remove it

PR #5908, opened 2022-10-19
jax_flipud

nassimberrada:


ivy/functional/frontends/jax/numpy/name_space_functions.py
```diff
@@ -1,5 +1,6 @@
 # local
 import ivy
+from ivy.functional.frontends.jax.func_wrapper import inputs_to_ivy_arrays
```
Ishticode, line 3: lint error, this is imported but never used..

Ishticode (changes requested):
Could you fix the lint error and then resolve conflicts... should be good to merge.

PR #5909, opened 2022-10-19
safety factor linalg

zaeemansari70:
We still have to use `assume()` in the tests, there is no other way around where we can generate a numerically stable matrix(well-conditioned). The `condition` number of the matrix is checked and then the matrix is re-generated if the `condition` number is high. 
There is one other way around, we can make that matrix a little well-conditioned from ill-conditioned, reducing it's condition number, but that would be by something known as diagonal scaling, it would be numerically expensive and the original matrix would also be changed as well, it still doesn't make sure that to what extent it would be reducing the numerical instabilities.
The best approach would be to re-generate the matrix if the condition number is high as it leads to singularity.
Do let me know if this solution doesn't seem suitable.

### DIFF
--- a/ivy_tests/test_ivy/helpers/hypothesis_helpers/general_helpers.py
+++ b/ivy_tests/test_ivy/helpers/hypothesis_helpers/general_helpers.py
@@ -8,65 +8,40 @@ import numpy as np
 from . import array_helpers, number_helpers, dtype_helpers
 
 
-def safety_factor_linalg(
-    matrix,
-    *,
-    condition_index="high",
-):
+def matrix_is_stable(x, cond_limit=30):
     """
-    Applies safety factor to the condition of a matrix to avoid numerical
-    instabilities in further calculations.
+    Used to avoid numerical instabilities in further computationally heavy
+    calculations.
 
     Parameters
     ----------
     matrix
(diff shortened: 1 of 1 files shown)


djl11 (changes requested):
I think the layout of the function can be made much simpler. The "safety factor" should be a configurable parameter in the function, and I don't think we need the `condition_index` argument. In all cases we simply determine the condition of the matrix, and see if this is above or below a pre-defined threshold. Rather than `safety_factor` we should simply have a function like this using a `cond_limit`:

```
def matrix_is_stable(x, cond_limit=10):
    return np.linalg.cond(x.astype('float64')) <= cond_limit
```

What do you think?

zaeemansari70:
Oh yeah, this would be a lot simpler.
I've refactored the description and the function too.
Just one more thing, I think that the `cond_limit=10` should be starting from "30", and gradually being decreased if the tests still fail just to result in more and more stable matrices.
Starting it from "10" will restrict the function from being tested on a smaller subset of matrices beforehand. Keeping it to "30" in the start we will be able to test it on a wider range of matrices.

What do you think?

ivy-seed:

djl11 (approved):
lgtm!

PR #5953, opened 2022-10-20
Added examples to docstring for vector_norm

11Kclarke:
for issue:
#5943 
I've added examples. I'm not super sure everything else is correct, if someone points out something ill take another look at the docs and then fix it.

Also, it seems justpqa created some examples and did most of the reformatting in issue #5499 but I don't think this was accepted, and the issue was still marked open. If this means I should find another issue please say.


ahmedo42:
Hi @11Kclarke, Thanks for your contribution, can you please work on another function?, maybe close this PR and we can re-open it once we make sure @justpqa PR is closed as stale.

PR #5955, opened 2022-10-20
Added __len__ instance method and test_tensorflow_instance_len test

valeriocost:
#5667 
Note: the method does not work as expected because there is the name property of Tensor which is missing.


AnnaTz:
Close #5667

ivy/functional/frontends/tensorflow/tensor.py
```diff
@@ -125,5 +125,12 @@ def __sub__(self, y, name="sub"):
     def __truediv__(self, y, name="truediv"):
         return y.__rtruediv__(self.data)
 
+    def __len__(self):
+        raise ivy.exceptions.IvyError(
+            f"len is not well defined for a symbolic Tensor "
+            f"(*tensor name*). Please call `x.shape` rather than "
+            f"`len(x)` for shape information."
+        )
```
AnnaTz, line 133: Since we haven't included `self.name`, you could print this message instead:
"len is not well defined for symbolic Tensors. "
"Please call `x.shape` rather than `len(x)` for "
"shape information."

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
```diff
@@ -771,6 +771,11 @@ def test_tensorflow_instance_neg(dtype_and_x, as_variable, native_array, fw):
     )
 
 
+# __len__
+@handle_cmd_line_args
+#todo
+
+
```
AnnaTz, line 778: Is this something you forgot to remove?

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
```diff
@@ -968,6 +973,33 @@ def test_tensorflow_instance_invert(dtype_and_x, as_variable, native_array, fw):
...
+        input_dtypes_method=[],
+        as_variable_flags_method=[],
+        num_positional_args_method=0,
+        native_array_flags_method=[],
+        all_as_kwargs_np_method={},
+        frontend="tensorflow",
+        class_name="Tensor",
+        method_name="__len__",
+    )
+
```
AnnaTz, line 1002: That's great, but since `__len__` just raises an error, I don't think there is any point in having a test function for it.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
valeriocost: Will delete then

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
valeriocost: Yep

ivy/functional/frontends/tensorflow/tensor.py
valeriocost: Okay sure, makes sense! I wanted to match exactly the tf behavior. Will change that

ivy/functional/frontends/tensorflow/tensor.py
```diff
@@ -130,5 +130,10 @@ def __sub__(self, y, name="sub"):
     def __truediv__(self, y, name="truediv"):
         return y.__rtruediv__(self.data)
 
+    def __len__(self):
+        raise ivy.exceptions.IvyError(
+            "len is not well defined for a symbolic Tensor. Please call `x.shape` rather than `len(x)` for shape information."
+        )
```
AnnaTz, line 136: This causes a lint error "E501 line too long (126 > 88 characters)". 
There is a pre-commit hook you can use to avoid having to worry about this type of errors. You can read about it here https://lets-unify.ai/ivy/contributing/setting_up.html#pre-commit.
In this case of course you can just fix it manually 'cause it's just one line 🙂

AnnaTz (changes requested):

AnnaTz (approved):
Great! Ready to be merged🙂

PR #5956, opened 2022-10-20
fix failing raw_ops tensforflow frontend tests

kurshakuz:
Fixes `Concat`, `Div`, `Diag`, `Sum` raw_ops Tensorflow frontend failing tests

+cumprod, cumsum, Add, Maximum, Mul, Neg


ivy/functional/frontends/tensorflow/raw_ops.py
```diff
@@ -22,7 +22,10 @@ def Acosh(*, x, name="Acosh"):
     return ivy.acosh(x)
 
 
-Add = tf_frontend.math.add
```
kurshakuz, line 25: This fix was requried to fix the issue with keyword args as those functions can only accept them. Other alternative is to change the functions inside tf_frontend.math, but I am not sure if it will affect other places

ZoeCD: I'm sorry, but what was the issue with keyword args? I checked TF docs and `add` arguments are not keyword args 🤔

kurshakuz: The tests were failing due to that. I am going to mark this one as a draft for now as there are some changes going rn and wait until it gets resolved. I am going to remove that changes later

kurshakuz:
Closed in favor of https://github.com/unifyai/ivy/pull/6086

PR #5960, opened 2022-10-20
add additional information to data promotion chapter

kurshakuz:
Adds explanations to the data promotion chapter to clarify that type upcasting is not to be used for improving accuracy and precision.

### DIFF
--- a/docs/partial_source/deep_dive/data_types.rst
+++ b/docs/partial_source/deep_dive/data_types.rst
@@ -134,7 +134,7 @@ return new type promoted values, respectively.
 For an example of how some of these functions are used,
 the implementations for :func:`ivy.add` in each backend framework are as follows:
 
-# JAX
+JAX:
 
 .. code-block:: python
 
@@ -148,7 +148,7 @@ the implementations for :func:`ivy.add` in each backend framework are as follows
         x1, x2 = ivy.promote_types_of_inputs(x1, x2)
         return jnp.add(x1, x2)
 
-# NumPy
+NumPy:
 
 .. code-block:: python
 
@@ -163,7 +163,7 @@ the implementations for :func:`ivy.add` in each backend framework are as follows
         x1, x2 = ivy.promote_types_of_inputs(x1, x2)
         return np.add(x1, x2, out=out)
 
-# TensorFlow
+TensorFlow:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


djl11 (changes requested):
Do the double dashes render correctly via Sphinx? I haven't tried this locally yet. I've always been using `:code:` inline. Maybe I've been missing a trick.

More importantly, I think our explanation of not upcasting for precision purposes can be improved a bit. It's perhaps obvious that we should not upcast when doing a simple division. However, someone might think it's justified to upcast an almost singular matrix from `float32` to `float64` in order to avoid nan or inf when computing the inverse of the matrix, before then converting back to float32 before returning from the function. This would preserve the input-output behaviour, and would prevent overflows and underflows. The reason we should still not upcast in such situations is because this removes control from the user regarding the memory footprint of the underlying algorithm. If upcasting is needed for numerical stability, they will be expected to do this manually before calling any troublesome functions. I think this is a better example to use for explanation. Does that make sense? :)

kurshakuz:
> Do the double dashes render correctly via Sphinx? I haven't tried this locally yet. I've always been using `:code:` inline. Maybe I've been missing a trick.

Yes, they defintely do. They are used on that same page for referencing words like ``dtype`` here https://github.com/unifyai/ivy/pull/5960/files#diff-e0622ba2578c9b0cf4a35667f61aa6a13cb052b8251b6226dfc4f5a6690bdd25L210 (line 210). Also as I understood, the :code: makes the code inside bold and highligted, while double dashes make them simply highlighted. 

Rendered page looks like this:
image


> More importantly, I think our explanation of not upcasting for precision purposes can be improved a bit. It's perhaps obvious that we should not upcast when doing a simple division. We especially should not be returning an upcasted array in such cases to the user. For a more nuanced and informative example: someone might think it's justified to upcast an almost singular matrix from `float32` to `float64` in order to avoid `nan` or `inf` when computing the inverse of the matrix, before then converting back to `float32` before returning from the inversion function. This would preserve the input-output behaviour from the user perspective, and would also succeed in preventing overflows and underflows. The reason we should still _not_ upcast in such situations is because this removes control from the user regarding the memory footprint of the underlying algorithm. If upcasting is needed for numerical stability, they will be expected to do this manually before calling any troublesome unstable functions. I think this is a better example to use for the explanation. Does that make sense? :)
> 
> Feel free to copy and paste parts/all of this text if helpful.

Understood! I slightly reworded your sentences, so please let me know if they make sense to you 😄

kurshakuz:
Here is how the page and double dash highlights render:
image

djl11 (approved):
lgtm!

PR #5968, opened 2022-10-21
added numpy.ndarray.argmin to frontend

### LINKED ISSUE #5612: argmin

Oluwadurotimi10:
Close #5612


vaatsalya123:
Hey @Oluwadurotimi10 hope your doing great.The pr looks OK though is failing some tests which you might need to look into and fix.I believe that the issue is you have passed the parameter value which isn't accepted by that instance method.Instead the parameters that should be passed to the test are axis and keepdims instead.If you have any other questions please let me know also note that the fixes are small and it should not take you long to get them fixed (:

vaatsalya123:
LGTM!

PR #5982, opened 2022-10-21
Developed the Complex function in tensor_functions

affan-sid:


ivy/functional/frontends/torch/tensor_functions.py
```diff
@@ -14,8 +15,8 @@ def is_tensor(obj):
 # def is_storage(obj):
 # 	return ivy.is_storage(obj)
 
-# def is_complex(obj):
-# 	return ivy.is_complex(obj)
+def is_complex(obj):
+ 	return Tensor(ivy.is_complex(obj))
```
jkeane508, line 19: Ivy doesn't have `is_complex` in its api. You would need to make an implementation here using some of the Ivy Functions currently implemented

jkeane508 (changes requested):
Hey, can you please reference which issue this tackles. You also would need to write a test for this function.

affan-sid:
Thanks for the response.
To implement the is_complex() function, can I use built in functions from
the PyTorch Library?

jkeane508:
> 

No unfortunately, in order to wrap the function in Ivy code, we need to use functions that belong to Ivy. You can read up about Ivy frontends [here](https://lets-unify.ai/ivy/deep_dive/ivy_frontends.html)

ivy/functional/frontends/torch/tensor_functions.py
```diff
@@ -1,4 +1,5 @@
 # local
+from numpy import complex128, complex192, complex64
```
jkeane508, line 2: I don't think using numpy dtypes makes sense for a tensorflow frontend?

ivy/functional/frontends/torch/tensor_functions.py
```diff
@@ -14,8 +15,9 @@ def is_tensor(obj):
 # def is_storage(obj):
 # 	return ivy.is_storage(obj)
 
-# def is_complex(obj):
-# 	return ivy.is_complex(obj)
+def is_complex(obj):
+    if ivy.dtype(obj) == complex64 or ivy.dtype(obj) == complex128 or ivy.dtype(obj) == complex192:
```
jkeane508, line 19: I would actually cast to an Ivy.array here and then compare using Ivy dtypes. That would make the most sense, then return boolean value if it is complex or not :)

ivy/functional/frontends/torch/tensor_functions.py
```diff
@@ -43,4 +45,4 @@ def is_nonzero(input):
         1,
         message="bool value of tensor with more than one or no values is ambiguous",
     )
-    return Tensor(ivy.nonzero(input)[0].size != 0)
+    return Tensor(ivy.nonzero(input)[0].size != 0)
```
jkeane508, line 48: Seems to be that you dont have a precommit hook installed, you can read up about that [here](https://lets-unify.ai/ivy/contributing/setting_up.html#pre-commit)

ivy_tests/test_ivy/test_frontends/test_torch/test_tensor_functions.py
```diff
@@ -32,6 +32,33 @@ def test_torch_is_tensor(
...
+    helpers.test_frontend_function(
+        input_dtypes=input_dtype,
+        as_variable_flags=as_variable,
+        with_out=False,
+        num_positional_args=num_positional_args,
+        native_array_flags=native_array,
+        frontend="torch",
+        fn_tree="is_complex",
+        obj=x[0],
+    )
```
jkeane508, line 58: Having three line breaks here would cause a lint error here

jkeane508 (changes requested):
Hey, some changes needed but its getting there!

affan-sid:
Thanks again for the feedback.
I have made the required changes, can you please check it now?

ivy/functional/frontends/torch/tensor_functions.py
```diff
@@ -14,8 +14,18 @@ def is_tensor(obj):
 # def is_storage(obj):
 # 	return ivy.is_storage(obj)
 
-# def is_complex(obj):
-# 	return ivy.is_complex(obj)
+
+def is_complex(obj):
```
jkeane508, line 18: I believe the test relevant test is failing for this PR, I think it is due to the fact that you have labelled the parameter as `obj` and not `input` which is consistent with the PyTorch documentation :)

jkeane508 (changes requested):
Hi, the relevant test for this PR needs to be passing before we merge :)

jkeane508:
Seems that tests didn't run, on your last commit. Once they are done running Ill drop a review :)

jkeane508 (changes requested):
Hey, seems to be that the relevant test is failing, this would need to be passing before we can merge :)

ivy/functional/frontends/torch/tensor_functions.py
```diff
@@ -14,8 +14,18 @@ def is_tensor(obj):
 # def is_storage(obj):
 # 	return ivy.is_storage(obj)
 
-# def is_complex(obj):
-# 	return ivy.is_complex(obj)
+
+def is_complex(input):
+    arr = ivy.asarray(input)
```
jkeane508, line 19: The failures suggest that it could be because you are using `ivy.asarray` and not using the ivy.Array class init.

jkeane508 (changes requested):

jkeane508 (changes requested):
I suspect that the reason your test is failing is not due to your implementation, but your test data generation. Some cases generate a list of arrays which will fail. For the test to pass you need to adjust the `dtype_and_values` helper to limit the generation to single arrays.

ivy-seed:

ivy-seed:

PR #5990, opened 2022-10-21
torch moveaxis

nassimberrada:


Ishticode (approved):
lg as long the tests pass 🙂

PR #6018, opened 2022-10-22
Redis Publically Read-Only

Aarsh2001:
The strategy we're using now is `Database Multiplexing` provided by hypothesis. Here we'll have a local `DirectoryBasedExample` or `InMemoryExample`(for PRs) database along with our `RedisExampleDatabase` deployed on cloud, which is enabled in ReadOnlyMode.

Since there's only a limited number of connections to the database possible, we will fallback to the default scenario(`DirectoryBasedExample`) when we encounter a `ConnectionError`. 

Further improvments -:

1. Work with @CatB1t to remove exposed endpoint in the test logs for `array_api`
2. hide the readonly endpoint as well
3. Authenticate for full access using an SSL certificate only(default DB user - push event on master !).


ivy_tests/conftest.py
```diff
@@ -1,27 +1,96 @@
...
+    DirectoryBasedExampleDatabase,
+)
 from hypothesis.extra.redis import RedisExampleDatabase
 from pytest import mark
 from pathlib import Path
 
-if os.getenv("REDIS_URL", default=False) and os.environ["REDIS_URL"]:
+
+cache_path = os.getcwd() + "/.hypothesis/examples/"
+r = None
```
CatB1t, line 16: just rename the variable to something more meaningful, otherwise lgtm!

CatB1t (approved):

PR #6033, opened 2022-10-22
added relu_ to torch frontend and tested

### LINKED ISSUE #5797: relu_

TAUIL-Abd-Elilah:
Close #5797


jkeane508 (changes requested):
Hey, seems to be the test is failing for `rely_`, this would need to be passing before we can merge :)

PR #6057, opened 2022-10-22
Inplace_update assertion

abdrahmandiab:


abdrahmandiab:
Hey Dan (@djl11),
This PR addresses the inplace update assertions task ([more info here](https://docs.google.com/spreadsheets/d/1WBvoQCoNxxG_7wVXs2LpYFA54zFCnSx7L87bkVHCXxA/edit#gid=0&range=97:97))

2 points that I would like a review on...
1. I have added the assertion for appropriate shapes to the `numpy` and `torch` backends only seeing as it does not truly matter for the other two backends and would just throw an error that can be ignored. Should I include the assertion for `jax` and `tf` as well or leave them omitted as I described? 
2. The line [here](https://github.com/unifyai/ivy/blob/31e9e4e5881de0c58ef182108b2485308fe00fe3/ivy_tests/test_ivy/test_functional/test_core/test_general.py#L1312) asks about what we should do regarding the `tensor_fn` parameter used to include `ivy.variable` in the test since we do not use `helpers.test_function` for this test... so... should we leave that in? 😅

Thanks!

PR #6075, opened 2022-10-23
Graph Compiler different framework example

CheezBallzPi:
Continuation of #5352. Here are the images that will eventually fill the spaces:

compiled_graph_tf
compiled_graph_jax
compiled_graph_numpy


CheezBallzPi:
It seems that there are some changes in master that didn't carry over to my branch for some reason. I'll try to get those fixed.

CheezBallzPi:
Alright should be ok now.

CheezBallzPi:
Reversed all the arrows. New photos attached:
compiled_graph_numpy
compiled_graph_tf
compiled_graph_jax

djl11 (approved):
looks good to me! I was wondering whether we should also put green boxes around the singular functions as well, to show that these are also the ivy functions but just with one function, but this might clutter things and so I think your approach is actually perhaps better. I'll leave this decision with you anyway, the graphs and the text are both great!

CheezBallzPi:
I also think it will look a bit cluttered. I'll leave it as is.

PR #6086, opened 2022-10-24
tf_frontend raw_ops alias decorator

hirwa-nshuti:
Hello😊,
Here is the implementation of a decorator to wrap the `tf_frontend` function that is way similar to the `raw_ops` function except that they differ in the way they handle arguments, `raw_ops` functions accept key-word only arguments.

I made changes such that we pass the `tf_frontend` function similar to the `raw_ops` being implemented. And then pass all the parameters of the defined `raw_ops` function to the wrapper as keyword arguments to the `tf_frontend` function. 

With this more focus can be put on implementing `tf_frontend` functions and then wrapping them later to get the `raw_ops` functions.

Thanks 🙂


ivy/functional/frontends/tensorflow/raw_ops.py
```diff
@@ -24,10 +27,25 @@ def Acosh(*, x, name="Acosh"):
     return ivy.acosh(x)
 
 
-Add = tf_frontend.math.add
+@to_ivy_arrays_and_back
+def Add(*, x, y, name="Add"):
```
djl11, line 31: In this implementation, we will wrap the function every time we call `Add` which is not good. We should apply the decorator only once at import time. It would be better if we did something like:

```
Add = map_raw_ops_alias(tf_frontend.math.add, x=x, y=y, name=name)
```

This will only be called at import time, and will make `Add` much more runtime efficient.

djl11 (changes requested):
requested some changes, but in general it's looking good!

ivy/functional/frontends/tensorflow/func_wrapper.py
```diff
@@ -92,3 +92,35 @@ def new_fn(*args, **kwargs):
 
 def to_ivy_arrays_and_back(fn: Callable) -> Callable:
     return outputs_to_tensorflow_array(inputs_to_ivy_arrays(fn))
+
+
+def map_raw_ops_alias(alias: callable, **function_kwargs):
```
VedPatwardhan, line 97: We shouldn't accept `**function_kwargs` here, we could instead accept a dictionary by some name say `kwargs_to_replace` where every pair in the dictionary is a mapping from the old name of the keyword argument and the new name of the keyword argument when you call the function, e.g. `{"dimension": "axis"}` for the argmax special case you've mentioned below. This way we could handle all edge cases, whilst also making sure that we only wrap once during the import.

ivy/functional/frontends/tensorflow/func_wrapper.py
```diff
@@ -92,3 +92,35 @@ def new_fn(*args, **kwargs):
...
+    functions_kwargs:
+        All inputs to the raw_ops as keyword only arguments.
+
+    Returns
+    -------
+    ret
+        The output of the wrapped function passed key-word only arguments as input.
+    """
+
+    def _wrap_raw_ops_alias(fn: callable) -> callable:
```
VedPatwardhan, line 117: This function should also accept the dictionary mentioned above and we should remove the old keys from the kwargs and add the new ones in the `_wrapped_fn`

ivy/functional/frontends/tensorflow/func_wrapper.py
```diff
@@ -92,3 +92,35 @@ def new_fn(*args, **kwargs):
...
+        @functools.wraps(fn)
+        def _wraped_fn(*args, **kwargs):
+            kwargs.update(zip(fn.__code__.co_varnames, args))
```
VedPatwardhan, line 120: We should be doing the updation of keys mentioned previously above this line.

ivy/functional/frontends/tensorflow/func_wrapper.py
```diff
@@ -92,3 +92,35 @@ def new_fn(*args, **kwargs):
...
+            return fn(**kwargs)
+
+        return _wraped_fn
+
+    wrapped_function = _wrap_raw_ops_alias(alias)
+    return wrapped_function(**function_kwargs)
```
VedPatwardhan, line 126: We shouldn't be calling the function, we should be returning it instead, so it should be `return _wrap_raw_ops_alias(alias)` instead

ivy/functional/frontends/tensorflow/raw_ops.py
```diff
@@ -24,10 +27,23 @@ def Acosh(*, x, name="Acosh"):
...
+    return map_raw_ops_alias(
+        tf_frontend.math.add,
+        x=x,
+        y=y,
+        name=name,
+    )
 
 
-ArgMax = tf_frontend.math.argmax
```
VedPatwardhan, line 30: This can then just be `ArgMax = map_raw_ops_alias(tf_frontend.math.argmax, kwargs_to_replace={'dimension': 'axis'})`. Again, feel free to decide the name of this argument, `kwargs_to_replace` is just an example.

VedPatwardhan (changes requested):
Hi @hirwa-nshuti, just requested a few changes based on your conversation with @djl11 on the previous review, feel free to let me know if it's a bit unclear. Thanks 🙂

hirwa-nshuti:
Hello @VedPatwardhan I fixed the decorator and all the wrapping is occurring during the import `kwarg` dictionary updates is being handled by the decorator now.

ivy/functional/frontends/tensorflow/raw_ops.py
```diff
@@ -24,10 +27,17 @@ def Acosh(*, x, name="Acosh"):
     return ivy.acosh(x)
 
 
-Add = tf_frontend.math.add
+Add = to_ivy_arrays_and_back(
+    map_raw_ops_alias(tf_frontend.math.add, kwargs_to_update={"Add": "name"})
```
VedPatwardhan, line 31: I think the `kwargs_to_update` should be empty for this function as we are not renaming any keyword argument when it gets passed to the actual frontend function from raw_ops, I think given that function checks for the keys being present in both the dictionaries, you may not be able to facing any issues even after adding `{"Add": "name"}`, but I think it should be empty unless the names of the arguments change and "Add" isn't an argument in the funcftion, it's the name of the function instead. Same applies to all the other functions in this file.

VedPatwardhan (changes requested):
Hi @hirwa-nshuti, just suggested a minor change 🙂

VedPatwardhan (approved):
lgtm, feel free to merge 🙂

PR #6102, opened 2022-10-24
Kaiser bessel derived window extension

nassimberrada:
PR for issue #5620
Tests finally run, was an issue with my cloned installation after all, tests also pass.


Ishticode (approved):
Looks perfect to me....  as long as the tests are passing for all backends.

I see that this is a mixed function and for jax, np, and torch the call will be made from ivy functional but for tensorflow it will be made from an actual implementation in the tesorflow backend. 

Thanks Nassim and thank you for your patience.🙂

Ishticode:
close #5620

PR #6117, opened 2022-10-24
dirichlet extension

nassimberrada:
Throws the following error for all back-ends except numpy: 

`>       ret, ret_gt = call()`
`E       TypeError: cannot unpack non-iterable NoneType object`

### DIFF
--- a/ivy/array/extensions/random.py
+++ b/ivy/array/extensions/random.py
@@ -45,11 +45,11 @@ class ArrayWithRandomExtensions(abc.ABC):
         Examples
         --------
         >>> alpha = ivy.array([1.0, 2.0, 3.0])
-        >>> ivy.dirichlet(alpha)
+        >>> alpha.dirichlet()
         ivy.array([0.10598304, 0.21537054, 0.67864642])
 
         >>> alpha = ivy.array([1.0, 2.0, 3.0])
-        >>> ivy.dirichlet(alpha, size = (2,3))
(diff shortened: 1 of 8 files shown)


ivy/array/extensions/random.py
```diff
@@ -1,6 +1,61 @@
...
+
+        Returns
+        -------
+        ret
+            The drawn samples, of shape (size, k).
+
+        Examples
+        --------
+        >>> alpha = ivy.array([1.0, 2.0, 3.0])
+        >>> ivy.dirichlet(alpha)
```
Ishticode, line 48: Should be an instance method... and below too

Ishticode (approved):
I have fixed tests for all backends....
Let me know if you have any objection to the updates.
There was several errors. Could explain if required.
But the one you mentioned above was caused because the `call()` upon `test_funnction` was hitting a `return` (equivalent to skip test) which return None and hence it was unable to unpack it `ret` and `ret_gt`.

Anyway feel free to merge after taking care of the below comment and the lint (try linking up the issue if possible) ...

nassimberrada:
So the issue was with the testing, i'm still wondering why the tests passed with numpy if that's the case 🤔. In any case, glad this one is finally done, been dragging it around since the beginning of my work at Ivy hahah, thanks !

nassimberrada:
Close #6156

PR #6119, opened 2022-10-24
hamming window extension

### LINKED ISSUE #6120: hamming_window

nassimberrada:
Close #6120

### DIFF
--- a/ivy/functional/extensions/layers.py
+++ b/ivy/functional/extensions/layers.py
@@ -743,10 +743,12 @@ def hamming_window(
     ivy.array([-1.8000,  0.2000,  2.2000,  0.2000, -1.8000])
     """
     if periodic is True:
-        window_length = window_length + 1
-        if window_length == 1:
+        if window_length == 0:
+            return ivy.array([])
+        elif window_length == 1:
             return ivy.array([1])
         else:
+            window_length = window_length + 1
             return ivy.array(
                 [alpha - beta * cos((2 * n * pi) / (window_length - 1))
                     for n in range(0, window_length)][:-1],
@@ -754,7 +756,9 @@ def hamming_window(
                 out=out
             )
     else:
-        if window_length == 1:
+        if window_length == 0:
+            return ivy.array([])
+        elif window_length == 1:
             return ivy.array([1])
         else:
             return ivy.array(
--- a/ivy/functional/extensions/layers.py
+++ b/ivy/functional/extensions/layers.py
@@ -742,12 +742,12 @@ def hamming_window(
     >>> ivy.hamming_window(5, periodic=False, alpha=0.2, beta=2)
     ivy.array([-1.8000,  0.2000,  2.2000,  0.2000, -1.8000])
     """
-    if periodic is True:
-        if window_length == 0:
-            return ivy.array([])
-        elif window_length == 1:
-            return ivy.array([1])
-        else:
+    if window_length == 0:
+        return ivy.array([])
+    elif window_length == 1:
+        return ivy.array([1])
+    else:
+        if periodic is True:
             window_length = window_length + 1
             return ivy.array(
                 [alpha - beta * cos((2 * n * pi) / (window_length - 1))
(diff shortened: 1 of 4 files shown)


nassimberrada:
Written as a compositional function because only torch allows specifying the alpha and beta values of the window function, other back-ends implicitly assume default values of alpha=0.54 and beta=0.46.

Ishticode (changes requested):
The PR is setup good.😊

But we need to take care of some special cases.
Firstly when `periodic=False` and `window_length=1`, we get division by zero error as clear from the `else` clause i.e. `window_length-1 = 0`
Likewise, zero division error would occur when `window_length=0` and `periodic=True`.

I would try to align with `torch.hamming_windows` in these special cases which returns 1. (mentioned in official documentation too) or an empty array in case of window_length=0.... Instead of throwing an error. 

Thanks Nassim. lmk when done....

nassimberrada:
Definitely didn't pay attention to that, thanks for pointing this out !
I included the special case to return an array with 1 like for torch as you suggested.

ivy/functional/extensions/layers.py
```diff
@@ -694,3 +694,72 @@ def sum_2N_1_n(n):
...
+    >>> ivy.hamming_window(5)
+    ivy.array([0.0800, 0.3979, 0.9121, 0.9121, 0.3979])
+    >>> ivy.hamming_window(5, periodic=False)
+    ivy.array([0.0800, 0.5400, 1.0000, 0.5400, 0.0800])
+    >>> ivy.hamming_window(5, periodic=False, alpha=0.2, beta=2)
+    ivy.array([-1.8000,  0.2000,  2.2000,  0.2000, -1.8000])
+    """
+    if periodic is True:
+        window_length = window_length + 1
+        if window_length == 1:
```
Ishticode, line 747: I guess this conditional clause should be before `window_length = window_length + 1` otherwise we are really handling input case of window_length = 0. 🙂

Also we need to add `if window_length == 0: return ivy.array([])` before everything.

lmk if unclear...

Ishticode (changes requested):
hmmm some more 😂

ivy/functional/extensions/layers.py
nassimberrada: No definitely clear, i guess i'm just tired enough to get confused over this 😅, hopefully this time's good

Ishticode (changes requested):
To be cleaner we can put the block
``` python
if window_length == 0:
    return ivy.array([])
elif window_length == 1:
    return ivy.array([1])
```
right in the beginning rather than duplicating it inside the coditionals twice because periodic or not doesn't make a difference for torch 

```python
>>> import torch
>>> torch.hamming_window(1, periodic=True)
tensor([1.])
>>> torch.hamming_window(1, periodic=False)
tensor([1.])
>>> torch.hamming_window(0, periodic=False)
tensor([])
>>> torch.hamming_window(0, periodic=True)
tensor([]).
```
Lmk when done. Thanks for the patience.

nassimberrada:
Yeah, curious how torch basically skips the case of window_length + 1 = 2 when periodic is true, or maybe it's a feature of the function. In any case, moved it out of the loop.

Ishticode (changes requested):
@nassimberrada 
I am not sure but there seems to be something wrong here. Not too big I hope. 

The test is taking `x[0]` as input which is still an array bcz x is a list of arrays (with only one entry in this case). You can see this with debugger. This means this test function is happily taking invalid inputs and passing.  An array is an invalid input bcz if you try this code

```python 
import ivy
x = ivy.array([2])
print(ivy.hamming_window(x))
```
You will get an error ragarding input not behaving like a scalar.
So in reality our function cannot take arrays. 

Currently you only need to change to test so that window length is int rather than rather than an array. 
Why the test was passing anyway seems to be a bug likely in the `test_function` so not your problem.  It seems to skip test in this case or pass.

Please let me know if this is unclear and feel free to correct me if I am mistakened.

nassimberrada:
The test_function has been throwing several bugs lately, either it's that or because of the inconsistency between the declared input type in the function and the input type used in tests. Nicely spotted, switched the testing input type to integer to stay consistent with other window functions.

Ishticode (approved):
Looks good to me@nassimberrada 
Thanks for the patience. Feel free to merge.

PR #6122, opened 2022-10-24
Partial Solution

MattPoyser:
Hello,

I have committed a partial solution that I have not tested yet but just wanted to clarify a few things beforehand. I am working on the unravel_index function and wanted to make sure I understood what needs to be done / confirm below observations are correct.

1. NumPy takes arguments: indices, shape, and optional order. JAX and TensorFlow take indices, shape arguments. Torch has no implementation. Order will not be part of the superset argument since it is under the hood non mathematical stuff?
2. Torch has no implementation and therefore needs to be implemented from scatch without use of any NumPy (or other framework) functions?
3. JAX allows indices inputs that can either be negative or out-of-bounds (and corresponding returned value should be clipped). Therefore, the NumPy, Torch and TF implementations need to be adjusted to also allow such inputs?

Thank you for your time.

### DIFF
--- a/ivy/functional/backends/jax/statistical.py
+++ b/ivy/functional/backends/jax/statistical.py
@@ -210,3 +210,12 @@ def einsum(
     equation: str, *operands: JaxArray, out: Optional[JaxArray] = None
 ) -> JaxArray:
     return jnp.einsum(equation, *operands)
+
+
+def unravel_index(
+        indices: jnp.dtype,
+        shape: jnp.dtype,
+        /,
+        *,
+) -> jnp.dtype:
+    return jnp.asarray(jnp.unravel_index(indices, shape))
--- a/ivy/functional/backends/numpy/statistical.py
+++ b/ivy/functional/backends/numpy/statistical.py
@@ -236,3 +236,18 @@ def einsum(
 
 
 einsum.support_native_out = True
+
+def unravel_index(
+    indices: np.ndarray,
+    shape: np.ndarray,
+    /,
+    *,
+) -> np.ndarray:
+    max_value = np.prod(shape)
+    for i in range(len(indices)):
+        if indices[i] > max_value:
+            indices[i] = max_value
+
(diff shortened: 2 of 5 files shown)


CerberusLatrans:
Hi Matt,

Sure! I have some corrections since I didn't realize that this is an extension function.

Here is an example of a compositional function in ivy/functional/ivy/layers.py:
https://github.com/unifyai/ivy/blob/25c5f599d08079b1ce6d7d1e7fe39d32ff7ba38b/ivy/functional/ivy/layers.py#L474

Firstly, you do not need an unravel_index in the torch backend.

In your implementation of unravel)index in functional/extensions/statistical.py (don't worry about ivy/functional/ivy/statistical.py -- I mispoke), make the implementation compositional like the above example, also kind of like how you did it in the torch backend (but using ivy functions). This will allow torch to run this function with the compositional implementation instead of being directed to the torch backend (since torch does not even have this function). Other backends which do support this function will still be able to use their respective backend implementation due to the behavior of ivy.__dict__ which you can read more about here if interested:
https://lets-unify.ai/ivy/deep_dive/backend_setting.html

You should also have your compositional implementation support superset behavior, meaning that it should deal with negative/out of bounds inputs just like Jax can. 


Regarding the JAX frontend, the way I explained it might have been a bit confusing. I meant that you should write some code to deal with negative/out of bounds args before passing them to ivy.unravel_index(). This way, none of the backends have to worry about receiving negative/out of bounds args since they were already taken care of beforehand. Here is an example of a frontend torch conv function I implemented that cleans up some of the args before passing them to the ivy version:
https://github.com/unifyai/ivy/blob/master/ivy/functional/frontends/torch/nn/functional/convolution_functions.py#L49

Let me know if you have anymore questions. Thanks!

ivy-seed:

ivy-seed:

ivy-seed:

ivy-seed:

PR #6127, opened 2022-10-24
Tensorflow Math

jkeane508:
Started implementing `lbeta()` to `tensorflow.math`
Thought I needed to implement `ivy.random.beta` so implemented, turns out I need `ivy.random.gamma` so implementing that too.
Other general fixes to `tensorflow.math` submodule


jkeane508:
Hey @VedPatwardhan 

Question for you on design. So currently all our probability stuff is mostly in `ivy.random` and this makes sense for sampling continuous distributions. But there are also probability functions, e.g gamma function. I was wondering if an ivy.math module is worthwhile adding to the functional API?

Thanks :)

ivy_tests/test_docstrings.py
```diff
@@ -228,6 +228,7 @@ def test_docstrings(backend):
         "random_uniform",
         "randint",
         "shuffle",
+        "beta",
```
VedPatwardhan, line 231: I think we should also skip the docstring tests for `gamma`

VedPatwardhan (changes requested):
Hi @jkeane508, I do agree on that, having another submodule would be a better choice, but I think we can add beta and gamma to random for now until the next release. Also, could you please move the 2 new functions and their backend implementations to extensions instead? Thanks 🙂

PR #6128, opened 2022-10-24
Rename linear_algebra modules to linalg to conform with Array API Standard

kurshakuz:
This PR removes the `ivy.linear_algebra` attribute and leaves only the `ivy.linalg` to conform with the Array API Standard.

Behavior after the updates:
```
>>> import ivy
>>> ivy.linalg
<module 'ivy.functional.ivy.linalg' from '/workspaces/ivy/ivy/functional/ivy/linalg.py'>
>>> ivy.lin
ivy.linalg     ivy.linear(    ivy.linspace(  
>>> ivy.linear_algebra
  File "<stdin>", line 1, in <module>
AttributeError: module 'ivy' has no attribute 'linear_algebra'
```


ivy/functional/extensions/__init__.py
```diff
@@ -19,8 +19,8 @@
 from .gradients import *
 from . import layers
```
djl11, line 20: Actually `from . import <module>` inside this file is the problem. We should not be doing this at all, because then these will clash with those same imports in `ivy/functional/__init__.py`. Specifically, inside `ivy/functional/__init__.py` we have both `from .extensions import *` and also `from . import <module>`. This will duplicate `<module>` and one of them will be overwritten which should not happen.

ivy/functional/extensions/__init__.py
```diff
@@ -19,8 +19,8 @@
...
 from .layers import *
-from . import linear_algebra
```
djl11, line 22: The cause of the existence  `ivy.linear_algebra` issue was this line. Simply changing this line to `from . import linear_algebra as linalg` will resolve the issue. No need to rename all of them.

djl11 (changes requested):
two simple changes requested, and then it's good to merge :)

ivy/functional/extensions/__init__.py
kurshakuz: I just tried to only change that line separately on a new branch from master, but the `ivy.linear_algebra` still persists. 

Adding that update will still show the following message:
```
>>> ivy.linear_algebra
<module 'ivy.functional.ivy.linear_algebra' from '/workspaces/ivy/ivy/functional/ivy/linear_algebra.py'>
```
However, if you change the `functional/ivy/linear_algebra.py` file, the link will now point to that module:
```
>>> ivy.linear_algebra
<module 'ivy.functional.extensions.linear_algebra' from '/workspaces/ivy/ivy/functional/extensions/linear_algebra.py'>
```

So it appears that we still need to make changes to both of the files.

The reason for that is that in the `ivy/functional/__init__.py` we are calling 
```
from .extensions import *
```
which obviously simply loads modules in that folder as they are named

However, in the `ivy/__init__.py` there is no such import, but even if you change the import format on line 533 to 
```
from ivy.functional.ivy import (
    ...
    linear_algebra as linalg,
    ...
)
```

It is still somehow getting the `linear_algebra` module to be exposed and the only way to resolve that was actually renaming the file.

So I still think we need to change both instances here, but please correct me if I am wrong. I tried different ways, but only that one worked for me, and probably I am missing something.

ivy/functional/extensions/__init__.py
kurshakuz: Makes sense! Solved in https://github.com/unifyai/ivy/pull/6128/commits/a6df22558acee1dc3d818521c90d07ba49b26ace

kurshakuz:
🚧  Sorry, resolving conflicts rn!

VedPatwardhan (changes requested):
Hi @kurshakuz, I think we should also be renaming the array and container files and the backend files from `linear_algebra.py` to `linalg.py` for consistency. Thanks 🙂

VedPatwardhan (changes requested):
Thanks for making the changes @kurshakuz, one final change, I think we could also name the array and container classes as `ArrayWithLinalg` and `ContainerWithLinalg` instead of `ArrayWithLinearAlgebra` and `ContainerWithLinearAlgebra`. Thanks 🙂

kurshakuz:
@VedPatwardhan added those requests. Also renamed references in the `ivy_array.rst` documentation page

VedPatwardhan (changes requested):
Hi @kurshakuz, could you please pull the latest master code once again and resolve the conflicts as there was another issue with `to_numpy` that was causing a large number of failures, just to be sure that this change won't cause any. Rest seems good to go 🙂

kurshakuz:
@VedPatwardhan all conflicts are resolved, but for some reason tests does not start. Do you know what may be the reason?

VedPatwardhan (approved):
lgtm! Feel free to resolve the conflicts and merge it 🙂

kurshakuz:
Thanks! Resolved conflicts and run the tests, merging!

PR #6157, opened 2022-10-25
Numpy reciprocal

Abheast12:


Abheast12:
Hi @abdrahmandiab, could you please let me know if there are any changes I need to make?

PR #6174, opened 2022-10-26
Diag test fix

JamieLine:


AnnaTz:
Hi! I was the one working on fixing the linear_algebra tests (as you can see here: https://docs.google.com/document/d/1cqtKRphgEaTAK-OoQq1aeUJZ5wuPXQk0uXWWeo2WR1g/edit), and I already fixed this test yesterday. 

However, I would like to explain some things about the diagonal-related functions because I noticed yesterday that there is confusion between backends. So, here it goes
1. There is `diagonal` that returns the specified diagonals of the input array. This was already implemented using `numpy.diagonal`, `torch.diagonal`, and `tf.experimental.numpy.diagonal`.
2. There is `diagflat` that returns an array with the flattened input as diagonal. I implemented this yesterday using `numpy.diagflat`, `torch.diagflat`, and `tf.experimental.numpy.diagflat`.
3. There is `diag` that combines 1 and 2 by either returning the specified diagonal or an array with the input as diagonal. This was the one that was causing failing tests and I fixed it yesterday using `numpy.diag`, `torch.diag`, and `tf.experimental.numpy.diag`.
All of the above receive three arguments: `input`, `k`, and `out`. This is common in numpy, torch, and tensorflow.experimental.numpy.
Also:
4. Tensorflow has another function `diag_part` that does the same thing as `diagonal` but with extra arguments: `padding_value` and `align` (in addition to `input`, and `k`). 
5. Tensorflow has a function `diag` that does the same thing as `diagflat` but with extra arguments:     `num_rows`, `num_cols`, `padding_value`, and `align` (again, in addition to `input`, and `k`). 

I can see you have worked out how to implement these extra arguments. So, I would suggest that you incorporate these as superset behavior to the already existing `diagonal` and `diagflat` accordingly (I think the naming should be consistent with numpy and torch to avoid extra confusion).

I hope this made sense!

JamieLine:
Hi Anna! 

Thank you for the detailed breakdown, this function had originally been created as part of the superset behaviour task where it had been written down with the name `diag` which had probably been inherited from Tensorflow. Once the tests were fixed, my next plan was to work out how to resolve this naming issue, so thank you for helping out with that! This PR does add all of the superset behaviour for what Tensorflow calls `diag`, so I'll move the functionality when I next get a chance to work on it. 

Thank you for the insight!

AnnaTz:
I'll just drop the doc links here too for easier reference:

1.
https://pytorch.org/docs/stable/generated/torch.diagonal.html
https://numpy.org/doc/stable/reference/generated/numpy.diagonal.html#numpy.diagonal
https://www.tensorflow.org/api_docs/python/tf/experimental/numpy/diagonal

2. 
https://pytorch.org/docs/stable/generated/torch.diagflat.html#torch.diagflat
https://numpy.org/doc/stable/reference/generated/numpy.diagflat.html#numpy.diagflat
https://www.tensorflow.org/api_docs/python/tf/experimental/numpy/diagflat

3.
https://pytorch.org/docs/stable/generated/torch.diag.html
https://numpy.org/doc/stable/reference/generated/numpy.diag.html
https://www.tensorflow.org/api_docs/python/tf/experimental/numpy/diag

4.
https://www.tensorflow.org/api_docs/python/tf/linalg/diag_part

5.
https://www.tensorflow.org/api_docs/python/tf/linalg/diag

JamieLine:
I've moved the changes to `diagflat` in line with your recommendations.

PR #6186, opened 2022-10-26
add alias tests to frontend test docs

hirwa-nshuti:


jkeane508 (changes requested):
Looks great!

Just for added clarity maybe we should add a functional example?

jkeane508 (approved):
Yea this is great! We just have to be mindful as we need to assume that we are writing docs for absolute beginners. So we can't assume they understand all relevant concepts :)

hirwa-nshuti:
Yea makes sense🙂

PR #6197, opened 2022-10-26
exp2 extension

### LINKED ISSUE #6196: exp2

nassimberrada:
Close #6196

### DIFF
--- a/ivy/array/extensions/elementwise.py
+++ b/ivy/array/extensions/elementwise.py
@@ -240,3 +240,36 @@ class ArrayWithElementWiseExtensions(abc.ABC):
         ivy.array([1.,   8.,  27.,  16.,   5.])
         """
         return ivy.float_power(self._data, x2, out=out)
+
+    def exp2(
+        self: Union[ivy.Array, float, list, tuple],
+        /,
+        *,
+        out: Optional[ivy.Array] = None,
+    ) -> ivy.Array:
+        """ivy.Array instance method variant of ivy.exp2. This method simply
+        wraps the function, and so the docstring for ivy.exp2 also applies to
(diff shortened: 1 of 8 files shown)


Ishticode (changes requested):
About the tensorflow backend
I am not fully sure of the consequences of `np_config.enable_numpy_behavior()`.... But enabling it file wide could lead to some unforeseen results in other functions. 

**I think its best** if we use 
```python
    return tf.cast(
        tf.math.pow(tf.constant(2, dtype=x.dtype), x), dtype=x.dtype
    )
```
 instead of involving numpy into this at all.

So could you delete the np config import and behaviour setting and updte the implementation in the tesorflow backend. 

edit: maybe just `2` instead of `tf.constant(2, dtype=x.dtype)` **might** work to as the first input for `pow` also I think.

nassimberrada:
I actually got an error that suggested using that when making use of numpy related functions, unless i used it the tests wouldn't pass. Switched for tf.math.pow as you suggested and it works as well, that's definitely better than relying on a file-wide change.

Ishticode (changes requested):
We actually dont need the `tf.cast`. It should be removed.
It should just be `return tf.math.pow(2,x)`. I think it should pass. Lmk when done and hopefully its ready

And the error for numpy related behaviour wast that tf's numpy experimental exp2 called x.astype internally with x being a tf tensor. This required np behaviour to be enabled. Bus fortunately we can steer clear from the whole thing.

nassimberrada:
I see, yeah the error was definitely related to the "astype" element, makes sense. 
Removed tf.cast.

Ishticode (approved):

PR #6198, opened 2022-10-26
numpy dirichlet

### LINKED ISSUE #4369: dirichlet

nassimberrada:
Close #4369

### DIFF
--- a/ivy/functional/frontends/numpy/random/functions.py
+++ b/ivy/functional/frontends/numpy/random/functions.py
@@ -6,3 +6,8 @@ from ivy.functional.frontends.numpy.func_wrapper import to_ivy_arrays_and_back
 @to_ivy_arrays_and_back
 def random(size=None):
     return ivy.random_uniform(low=0.0, high=1.0, shape=size, dtype="float64")
+
+
+@to_ivy_arrays_and_back
+def dirichlet(alpha, size=None):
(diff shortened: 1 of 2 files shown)


Ishticode (commented):
Looking good mostly. May I ask as to why is `test_values=False` in the `test_frontend_function` call?

nassimberrada:
I think that's a necessary argument to pass given the nature of the function itself which draws random samples at each draw. You can't test for the exactitude of values between two back-ends because they will necessarily output different value. Setting test_values = False seems to be a standard procedure when dealing with random functions apparently, that's what we did for ivy.dirichlet and it's the same for other front-end functions as well (see "random" function just above dirichlet in the numpy frontends).

Ishticode (approved):
Sorry my bad. This is pretty obvious. Thanks for explaining. I am absent minded sometimes 🙂. lgtm as long as the tests pass...

nassimberrada:
Took me a while to catch that one, but ivy.dirichlet threw enough errors at me to drill that inside my head for good 😃

Ishticode:
> Took me a while to catch that one, but ivy.dirichlet threw enough errors at me to drill that inside my head for good 😃

😂 Thats why debugging is perhaps the most active learning experience. Anyways good work @nassimberrada

PR #6218, opened 2022-10-27
torch cross

### LINKED ISSUE #4993: cross

nassimberrada:
Close #4993


nassimberrada:
Notable points:

- Followed the same testing process as ivy.cross implementations, using the same composite function and additional parameters
- Made use of ivy_promote_types_of_inputs in the implementation
- Set dim default value as None with a conditional block that inputs -1 if dim is None

Unless i included these points above, the tests would just keep throwing assertion errors with different results between the back-end and the ground-truth-back-end.

Ishticode (changes requested):
I am a bit unsure about whether setting `dim=-1` in case it is `None` is a good idea.
[PyTorch documentation](https://pytorch.org/docs/stable/generated/torch.cross.html) says that "if dim is not given, it defaults to the first dimension found with the size 3"    and if we force last dimension in that case, wouldn't it lead to some differences. For example if shape = (2,3,3). I didn't try out the computations anything so please correct me if wrong.

Two further suggestions though not absolutely necessary.
This helper function you wrote can be useful I guess so it might be worth placing it in the helpers... with tiny bit morecomments...... 

As I understood it
It can be a function where you want arrays to ensure certain dim size is found and that specific array is returned alongside arrays created.

Secondly, instead of hard cording two value draws... could perhaps generalise it by adding an arg like `num_values=2`  and then looping thru `for i in range(num_values)` and returning just the list instead of values[0], values[2]

These two are up for you to decide. Just a suggest from me. But the very first point needs adressing. Thanks Nassim.

nassimberrada:
@Ishticode I understand your first point and actually i wasn't sure if the implementation of ivy.cross was following the documentation for torch.cross as found in the link you sent, or torch.linalg.cross as found [here](https://pytorch.org/docs/stable/generated/torch.linalg.cross.html#torch.linalg.cross). But given that the description of the function in the docstrings was the same as the latter function, and that the axis argument was not set as optional, i figured ivy.cross actually followed torch.linalg.cross which, when you look at its arguments, takes dim=-1 as default value. That's why i figured i would stick to that as well.

As for your subsequent points, tbh i didn't come up with anything as far as the helper function is concerned, it is the exact same one used to test ivy.cross. I just followed the same testing process to replicate the testing conditions for ivy.cross. Unless i followed the same conventions, the tests would keep producing un-matched outputs between the back-end and the ground truth.

Ishticode (approved):

PR #6229, opened 2022-10-27
kron_extension

### LINKED ISSUE #6228: kron

nassimberrada:
Close #6228

### DIFF
--- a/ivy/array/extensions/linear_algebra.py
+++ b/ivy/array/extensions/linear_algebra.py
@@ -28,3 +28,24 @@ class ArrayWithLinalgExtensions(abc.ABC):
                    [0, 0, 0]])
         """
         return ivy.diagflat(self._data, k=k, out=out)
+
+    def kron(
+        self: ivy.Array,
+        b: ivy.Array,
+        /,
+        *,
+        out: Optional[ivy.Array] = None,
+    ) -> ivy.Array:
+        """
+        ivy.Array instance method variant of ivy.kron.
+        This method simply wraps the function, and so the docstring for
+        ivy.kron also applies to this method with minimal changes.
+
+        Examples
+        --------
+        >>> a = ivy.array([1,2])
+        >>> b = ivy.array([3,4])
+        >>> a.diagflat(b)
+        ivy.array([3, 4, 6, 8])
+        """
+        return ivy.kron(self._data, b, out=out)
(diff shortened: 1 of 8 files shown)


Ishticode (changes requested):
In ivy's functional. We should pass the arguments like `ivy.current_backend(a,b)`
This tries to infer the framework from the argument when no backend is set as you can see from this [line](https://github.com/unifyai/ivy/blob/c14cc21a242f40693863707e5edeeac31e0f50a4/ivy/backend_handler.py#L213) on ivy master.

SO here is a sequence.
When no backend is set, numpy is considered to be the implicit_backend. But if we find some native input like torch tensor then that backend is set for call.

so in a code like this

```python
import ivy
import torch
x = torch.tensor([1,2,3])
print(ivy.sum(x))
```
You can see that this goes to the torch backend sum function by first going to `ivy.current_backend(x).sum...` in ivy functional which sends it correctly to the torch backend. You can check this by setting breakpoints in pycharm/vscode I guess.

Sorry for the long winded explanation but I hope it helps. Let me know if unclear.

The summary is we should pass the input to the current_backend call in ivy functional.
Might need going through and adding this little piece to some of the previous implementations you have made with perhaps this missing. 

The only this that will happen if these were not included is the code snippet above wouldn't work. But it is not reallly the recommended use anyway. It is recommended for ivy to be used with a backend set. So in most cases this is insignificant.
Feel free to throw anything back at me if you like 🙂

nassimberrada:
@Ishticode So if my understanding is right, we only pass in current_backend() array inputs that vary in type between the different frameworks to allow ivy to infer which backend we're implicitly using, right ?
In which case i have a question, what happens when you pass two inputs with different frameworks ? For instance, suppose we have `current_backend(a, b).kron(a, b, out=out)` and that, following your code example, we write 
```
import ivy
import torch
import tensorflow
a = torch.tensor([1,2,3])
b = tensorflow.constant([4,5,6])
print(ivy.kron(a, b))
```
Which backend is going to be inferred ? Or will it go for the backend set as default ?
NB: I included the `current_backend(a, b).kron(a, b, out=out)` change in the new commit.

Ishticode (approved):
That's a very interesting question. atm it throws an error because it doesnt seem to be handled correctly... but I think we should be able to set an order. Like if multiple args are passed belonging to different fw then setting the fw of the first and taking the data from second and converting to ivy array. I will ask Daniel about it soon.

PR #6233, opened 2022-10-27
Added MatrixDiag to raw_ops

### LINKED ISSUE #5464: MatrixDiag

KareemMAX:
Close #5464


ivy/functional/frontends/tensorflow/raw_ops.py
```diff
@@ -228,13 +228,26 @@ def MatMul(*, a, b, transpose_a=False, transpose_b=False, name="MatMul"):
...
 
 MatrixDeterminant = tf_frontend.linalg.det
 
 
+@to_ivy_arrays_and_back
+def MatrixDiag(*, diagonal, name="MatrixDiag"):
+    last_dim = diagonal.shape[-1]
+
+    return ivy.astype(
+        ivy.expand_dims(diagonal, axis=-1)
```
CatB1t, line 244: Wouldn't it be more efficient if we make use of `ivy.zeros` and `ivy.scatter_nd` to update the array in place? this would make it more efficient for backends that supports `native_out`, we can generate the diag indicies using a list comprehension, e.g. `[(i,i) for i in diag]` and use it along with `scatter_nd`. What do you think?

PR #6236, opened 2022-10-27
torch frontend cos_ method

### LINKED ISSUE #6068: cos_

lovesahaj:
Close #6068

### DIFF
--- a/ivy/functional/frontends/torch/Tensor.py
+++ b/ivy/functional/frontends/torch/Tensor.py
@@ -41,6 +41,10 @@ class Tensor:
     def cos(self, *, out=None):
         return torch_frontend.cos(self.data, out=out)
 
+    def cos_(self):
+        self.data = self.cos()
+        return self.data
+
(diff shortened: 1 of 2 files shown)


lovesahaj:
@Daniel4078 torch frontend test only fails when testing on the 'bfloat16' type arguments, that's because numpy doesn't support that

Daniel4078:
> @Daniel4078 torch frontend test only fails when testing on the 'bfloat16' type arguments, that's because numpy doesn't support that

Then please check out this deep dive document (https://lets-unify.ai/ivy/deep_dive/16_ivy_frontends.html#supported-data-types-and-devices) about what to do in this case and maybe update your test to skip this unsupported datatype. Thank you!

lovesahaj:
I did that already, `cos_` doesn't fail any test cases.

Daniel4078 (approved):
Well done! I think it is all fine now. Thank you!

PR #6240, opened 2022-10-27
added matrix_power to jax frontend and its test

Infrared1029:


JamieLine:
Hi! Thanks for the PR!

This looks pretty good as it is, there's nothing that sticks out as an issue other than the fact that when I ran the test locally, it failed with a few errors. Have a look at trying to fix these, if you need any help then please let me know and I'll see what I can do. I noticed one of the errors mentioned overflows, so it might be worth investigating bounding the maximum and minimum values in the generated arrays to avoid that. 

Any questions or comments, please let me know!

Infrared1029:
So i think this should do it? the tests seem to be running fine here, wonder what's wrong with the linting tho

Infrared1029:
actually, my bad, it appears the person had their PR up 8 days ago, I should have noticed! ill close this PR since the method and its test are already merged! thanks for the review though:)

PR #6303, opened 2022-10-29
DCT

danielmunioz:


danielmunioz:
Close #5548

danielmunioz:
Hi! Sorry for the delay, I was quite busy, but I managed to get this one working!, one small thing, Tensorflow was chosen as the superset here due to all other backends not having a DCT function fully implemented (the Numpy workaround for this was to use Scipy in the backend), however type-I normalization is not supported for TF, but it does work with NP due to scipy being used in the backend, I believe I found a temporal solution for this but it may not be numerically stable, due to that I'm not sure if I should add it and maybe risk some potential inconsistencies, or if it's alright the way it is right now, which is basically having the same behavior as TF (just with the addition of being able to use the "axis" parameter).

Please LMK if any changes are necessary!!
Thanks for your time!

danielmunioz:
Sure! I can implement the algo using only numpy.

danielmunioz:
@RickSanchezStoic Done! LMK if there is anything else!

RickSanchezStoic:
Hi, @danielmunioz can you try resolving the conflicts, otherwise looks good to me

danielmunioz:
@RickSanchezStoic, Done!, Sorry about the force-push, I only did it because I was working alone and wanted to keep the branch clean, def will use `revert` next time tho :)

danielmunioz:
Hey @RickSanchezStoic!, I just fixed some new conflicts that just popped up today, please LMK if something else is needed! Thanks!.

PR #6316, opened 2022-10-29
add detach instance method

### LINKED ISSUE #6315: detach

ahmedo42:
closes #6315

### DIFF
--- a/ivy/functional/frontends/torch/Tensor.py
+++ b/ivy/functional/frontends/torch/Tensor.py
@@ -145,6 +145,9 @@ class Tensor:
     def view_as(self, other):
         return self.view(other.shape)
 
+    def detach(self):
+        return self.new_tensor(self.data, requires_grad=False)
+
     # Special Methods #
     # -------------------#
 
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/torch/Tensor.py
```diff
@@ -145,6 +145,9 @@ def new_tensor(
     def view_as(self, other):
         return self.view(other.shape)
 
+    def detach(self):
+        return self.new_tensor(self.data, requires_grad=False)
```
VedPatwardhan, line 149: I think the ivy equivalent of torch's `detach` is the `stop_gradient` function, please use it with a `preserve_type=False` in order to get a tensor with `requires_grad=False`.

VedPatwardhan (changes requested):
Hi @ahmedo42, just requested a few changes, rest looks good to me 🙂

ivy/functional/frontends/torch/Tensor.py
VedPatwardhan: Yeah but I think we could assume that the `torch.detach` function used inside the `stop_gradient` torch implementation should be more efficient with memory that our implementation.

VedPatwardhan (approved):
lgtm!

PR #6323, opened 2022-10-29
gcd_extension

### LINKED ISSUE #6280: gcd

nassimberrada:
Close #6280


nassimberrada:
If it ain't broken, don't fix it. Numpy functions work just fine, i guess they just didn't see a need to reinvent the wheel.

PR #6348, opened 2022-10-30
Instance Methods to PyTorch Frontend : expand

modanesh:


modanesh:
Hi @VedPatwardhan 
The `test_torch_instance_expand` is failing. Could you please take a look and help me resolve what's going wrong? 
Since Matthew is quite busy for the demo, I'm asking you. Hope that's fine.

ivy/functional/frontends/torch/Tensor.py
```diff
@@ -1,4 +1,6 @@
 # local
+import torch.onnx.symbolic_opset9
```
VedPatwardhan, line 2: I think we shouldn't be importing `torch` in the frontend file, but given that you're not using it anywhere, it may just be something that got automatically added via an IDE code suggestion.

VedPatwardhan (changes requested):
Just a minor change, rest looks good to me 🙂

ivy/functional/frontends/torch/Tensor.py
modanesh: Not sure how it added. Will remove.

PR #6357, opened 2022-10-31
changes

RickSanchezStoic:
Tried to implement the `__len__` instance method to reproduce the `len is not well defined for symbolic Tensors` error.


AnnaTz:
Minimal code to reproduce the original error:
`import tensorflow as tf

@tf.function
def test(ex_tensor):
    return ex_tensor.__len__()

example_tensor = tf.constant([2.0, 3.0, 4.0])
test(example_tensor)`

PR #6455, opened 2022-11-01
initial implementation and testing with decimal

CerberusLatrans:


CerberusLatrans:
still not too sure if this method is reliable enough, test just seems to keep running

PR #6470, opened 2022-11-01
cumsum and cumproduct to jax frontend

hirwa-nshuti:


hirwa-nshuti:
close #6466 
close #6465

PR #6480, opened 2022-11-01
unify top_k function

hirwa-nshuti:
Hi @jkeane508 

As per the conversation we had regarding this function here is the unified version of it.

Thanks:)


jkeane508 (changes requested):
Looks good, but now that they have a unified behaviour can we not value test them. I think its possible with the backend function tester?

hirwa-nshuti:
All backends are now producing the same results, the issue is with the torch frontend implementation of the function as we are testing against the actual torch function from the torch framework. I think having value test on backends is okay

jkeane508 (approved):

PR #6487, opened 2022-11-01
Fixed for bg, contributing

V-G-spec:
Part of my individual task. I could get docker up and running but not sure what `make_docs.sh` does and if I need it to update the website


RashulChutani (commented):
Great Work!

Issue #6492, opened 2022-11-01
eigvals


rubythalib33:

rubythalib33:
I'm gonna close this issue because there is implementations in the code

PR #6494, opened 2022-11-02
update eigvals

rubythalib33:
#6492

### DIFF
--- a/ivy/functional/frontends/jax/numpy/linalg.py
+++ b/ivy/functional/frontends/jax/numpy/linalg.py
@@ -19,6 +19,11 @@ def eig(a):
 
 
 @to_ivy_arrays_and_back
+def eigvals(a):
+    return ivy.eigvals(a)
+
+
+@to_ivy_arrays_and_back
 def eigh(a, UPLO="L", symmetrize_input=True):
     def symmetrize(x):
         # TODO : Take Hermitian transpose after complex numbers added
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/jax/numpy/linalg.py
```diff
@@ -18,6 +18,11 @@ def eig(a):
     return ivy.eig(a)
 
 
+@to_ivy_arrays_and_back
+def eigvals(a):
```
aryans1204, line 22: Hi, there is a redefinition of eigvals on Line 49. It seems that it was present before you amde that commit. Further, there is no eigvals function under the Ivy namespace, so your current implementation is incorrect. There is however, an `eigvalsh()` implementation under `ivy/ivy/functional/ivy/linear_algebra.py`. I'd recommend you take a look there and update your function adhering to the standard.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_linalg.py
```diff
@@ -62,6 +62,30 @@ def test_jax_numpy_eig(dtype_and_x, as_variable, native_array, num_positional_ar
     )
 
 
+# eigvals
+@handle_cmd_line_args
```
aryans1204, line 66: Similar issue here there is a redefinition of the test down in the file. This also needs to be fixed for taking into account, `eigvalsh` instead of eigvals. This would also cause the lint checks to pass, and hopefully the Jax frontend tests too if you've written the test correctly.

aryans1204 (changes requested):

PR #6354, opened 2022-10-31
Converter draft

VedPatwardhan:

PR #6503, opened 2022-11-02
Ivy Converters Implementation

YushaArif99:
The following implementations have been made:

- Fixes to `ivy.to_ivy_module()` [PyTorch and Jax]
- Implemented ivy.to_ivy_module() [TF]
- Implemented `to_keras_module()`
- Implemented `to_torch_module()`
- Implemented `to_haiku_module()`


djl11 (changes requested):
Looks good on first glance! Could you please move all of this logic to the `ivy.Module` instance methods instead though? Please see [this PR](https://github.com/unifyai/ivy/pull/6354) which was recently closed, which shows the kind of format we should now use for the method. Thanks!

ivy/stateful/module.py
```diff
@@ -1141,48 +1153,468 @@ def build_mode(self):
...
-    def built(self):
+    def built_(self):
         return self._built
 
-    # Methods #
-    # ------- #
+    # Module Converters #
+    # ------------------#
+    
+    @staticmethod
```
YushaArif99, line 1162: I have implemented `to_haiku_model`as a static method to preserve the natural behavior of the user-written code.
If I chose the design format as used in (`to_torch_module()` and `to_keras_module()`), where the `ivy_module` is stored as an attribute of the class( i.e Composition in OOP terminology), the user would then have to explicitly pass the ivy_module instance when constructing the forward_fn() function.
Consider the following example:

ivy_model = IvyModel()
haiku_model = ivy_model.to_haiku_module()

**_Using static_method_**
def forward_fn(*a, **kw):
    model = haiku_model()
    return model(*a, **kw)

**_Using Composition(having ivy_model stored as an attribute)_**
def forward_fn(*a, **kw):
    model = haiku_model(ivy_model)  #will have to pass ivy_model explicitly
    return model(*a, **kw)

Is the current design implementation fine or should I go with the composition approach?

ivy/stateful/module.py
```diff
@@ -1141,48 +1153,468 @@ def build_mode(self):
...
+    def from_haiku_module(
+            native_module,
+            args=None,
+            kwargs=None,
+            device=None,
+            devices=None,
+    ):
+        """
+        Converts a Haiku module instance to an Ivy module instance.
```
YushaArif99, line 1239: Because of the fact that both a native class as well as an instance of a native class can be passed to the `from_<backend>_module()` methods, I had to follow the convention that constructor parameters need to be keyword only whereas forward pass parameters need to be position only (when passing a class rather than an instance). This is necessary in order to avoid conflicts caused when propagating the `*args` and `**kwargs` through different methods of the Module class internally.

YushaArif99 (commented):
Hi @djl11 . I have refactored the changes to implement the converters as ivy Module instance methods and have written unit tests for them as well. Should I remove the converter files since they now seem obsolete?

ivy/stateful/module.py
```diff
@@ -1141,48 +1153,468 @@ def build_mode(self):
...
+    def built_(self):
         return self._built
 
-    # Methods #
-    # ------- #
+    # Module Converters #
+    # ------------------#
+    
+    @staticmethod
+    def static_to_haiku_module(ivy_module):
```
djl11, line 1163: no need to have a `static_to_haiku_module`, you can just have all of this logic implemented in the instance method `to_haiku_module`. The same goes for all other instance methods.

ivy/stateful/module.py
```diff
@@ -1017,7 +1028,7 @@ def save_weights(self, weights_path, /):
         os.makedirs("/".join(weights_path.split("/")[:-1]), exist_ok=True)
         self.v.to_disk_as_hdf5(weights_path)
 
-    def build(self, *args, from_call=False, device=None, dtype=None, **kwargs):
+    def build_(self, *args, from_call=False, device=None, dtype=None, **kwargs):
```
djl11, line 1031: why the new appending underscore `_`? Can you remove this?

djl11 (changes requested):
small changes requested, otherwise looks good to me! Yes could you please remove the converters module entirely from ivy and also the backends? This module is now redundant. Thanks!

djl11:
@VedPatwardhan could you please let us know if the changes to the gradient functions are okay? @YushaArif99 needed to make these changes in order to get the tests passing. Maybe you can elaborate on why this was needed @YushaArif99?

YushaArif99:
> @VedPatwardhan could you please let us know if the changes to the gradient functions are okay? @YushaArif99 needed to make these changes in order to get the tests passing. Maybe you can elaborate on why this was needed @YushaArif99?

Yeah, so I basically changed the order of operations so that the variables are first converted to native arrays before computing the forward pass. This is needed to keep track of the variables so that later their gradients can be calculated, otherwise, you get zero gradients returned.

VedPatwardhan (changes requested):
Hey @djl11 and @YushaArif99, I just had a few doubts regarding the change made to the `execute_with_gradients` function:
1. I was curious about the reason for calling the function after all required input arrays are converted to native arrays. Because ideally, the function `func` passed to the `execute_with_gradients` should be working with `ivy.Array`. I'm guessing that the callback function passed to `execute_with_gradients` has framework-specific code that wouldn't accept inputs of type `ivy.Array`.
2. And for the current `tensorflow` implementation this would raise an error as `xs` assigned on `line 58` contains the native arrays only for the ones that we're supposed to be calculating the gradients w.r.t. (ofc I could make another change to fix that).
Do let me know if there are any issues. Thanks!

YushaArif99:
Hey @VedPatwardhan 

I haven't really dived deep into the issue but from my understanding, and this is specific to the `from_<backend>_module()` converters case, since the forward pass of the model consists of framework-specific code, the variables need to first be converted to native arrays so that they can be tracked. 
The weird thing here is that not doing so does **not** generate an exception (at least for Pytorch and TF). Take the following code snippet for example:
`ivy.set_backend('torch')`
`print(torch.tanh(ivy.array([1.,2.,3.])))`
`>>> tensor([0.7616, 0.9640, 0.9951])`
Normally, you'd expect this to raise an exception of some form since `torch.tanh()` excepts a torch Tensor, not an ivy Array (which is what `ivy.array()` returns). However, somewhere internally perhaps, the ivy array is casted back to its native array form and the above code works just fine. 
This is perhaps the reason that not converting the variables to native arrays first, though does not generate any exception (i.e the forward pass is computed normally), it does cause the gradients to become 0 since the internal logic of tracking variables  (eg in Pytorch) assumes the variables to be pytorch tensors and hence is unable to track ivy Array variables.

The reason why TF works fine I believe has to do with the fact that the `execute_with_gradients()` function first explicitly calls the `tape.watch(xs_required)` where xs_required are the native array variables (not ivy arrays), before calling the forwards pass of the model.

Then again, these are just my thoughts on the possible reasons behind this strange behavior. Feel free to share any thoughts you might have. Thanks!

VedPatwardhan (commented):
> Hey @VedPatwardhan
> 
> I haven't really dived deep into the issue but from my understanding, and this is specific to the `from_<backend>_module()` converters case, since the forward pass of the model consists of framework-specific code, the variables need to first be converted to native arrays so that they can be tracked. The weird thing here is that not doing so does **not** generate an exception (at least for Pytorch and TF). Take the following code snippet for example: `ivy.set_backend('torch')` `print(torch.tanh(ivy.array([1.,2.,3.])))` `>>> tensor([0.7616, 0.9640, 0.9951])` Normally, you'd expect this to raise an exception of some form since `torch.tanh()` excepts a torch Tensor, not an ivy Array (which is what `ivy.array()` returns). However, somewhere internally perhaps, the ivy array is casted back to its native array form and the above code works just fine. This is perhaps the reason that not converting the variables to native arrays first, though does not generate any exception (i.e the forward pass is computed normally), it does cause the gradients to become 0 since the internal logic of tracking variables (eg in Pytorch) assumes the variables to be pytorch tensors and hence is unable to track ivy Array variables.
> 
> The reason why TF works fine I believe has to do with the fact that the `execute_with_gradients()` function first explicitly calls the `tape.watch(xs_required)` where xs_required are the native array variables (not ivy arrays), before calling the forwards pass of the model.
> 
> Then again, these are just my thoughts on the possible reasons behind this strange behavior. Feel free to share any thoughts you might have. Thanks!

I think in that case, we could be passing a function that accepts `ivy.Array` and then calls the actual function after converting them to native, something like 
```
func = lambda xs: actual_func(ivy.to_native(xs))
ivy.execute_with_gradients(func, xs)
``` 
instead of directly passing `actual_func`. I personally think the function passed to `execute_with_gradients` should work with `ivy.Array` as opposed to working with `ivy.NativeArray`. Maybe @djl11 could make the call here. Thanks 🙂

ivy/stateful/module.py
djl11: I realise that this is now out-dated, but `to_haiku_model` should not be implemented as a static method, it should only be added as an instance method.

ivy/stateful/module.py
djl11: because both the class and the class instance can receive arbitrary `args` and `kwargs`, a better approach would be to have something like `constructor_args`, `constructor_kwargs`, `instance_args`, `instance_kwargs` all as keyword arguements, without using `*` in the arguments to the outermost converter function. I haven't dived deeper and the argument names could be different, but in general if there are multiple targets for different `args` and `kwargs` then it's better to avoid `*` when catching them in the outer function arguments, if that makes sense?

ivy/stateful/module.py
djl11: why were there any conflicts? The `ivy.Module` does not inherit from `tf.keras.Model` and `tf.keras.Model` does not inherit from `ivy.Module`, so I don't understand why there would be any conflicts.

djl11:
> Hey @djl11 and @YushaArif99, I just had a few doubts regarding the change made to the `execute_with_gradients` function:
> 
> 1. I was curious about the reason for calling the function after all required input arrays are converted to native arrays. Because ideally, the function `func` passed to the `execute_with_gradients` should be working with `ivy.Array`. I'm guessing that the callback function passed to `execute_with_gradients` has framework-specific code that wouldn't accept inputs of type `ivy.Array`.
> 2. And for the current `tensorflow` implementation this would raise an error as `xs` assigned on `line 58` contains the native arrays only for the ones that we're supposed to be calculating the gradients w.r.t. (ofc I could make another change to fix that).
>    Do let me know if there are any issues. Thanks!

1. Yes this is correct, thanks for explaining @VedPatwardhan 

2. thanks for pointing this out!

djl11:
> 

Just to be clear, `execute_with_gradients` should be able to work with a function which consumes either `ivy.NativeArray` **OR** `ivy.Array` instances. This is already supported I believe, right @VedPatwardhan? The problem here is not that `execute_with_gradients` doesn't support one of these, the problem is that the function itself is fully native code, and we were trying to pass `ivy.Array` instances. Of course, `execute_with_gradients` should not be expected to detect and fix problems where the wrong input types are being passed to the function.

Ved's suggestion of:

```
func = lambda xs: actual_func(ivy.to_native(xs))
ivy.execute_with_gradients(func, xs)
```

is exactly how we should solve things in this case 👍 

Thanks both!

djl11 (changes requested):
see comments :)

VedPatwardhan:
> > 
> 
> Just to be clear, `execute_with_gradients` should be able to work with a function which consumes either `ivy.NativeArray` **OR** `ivy.Array` instances. This is already supported I believe, right @VedPatwardhan? The problem here is not that `execute_with_gradients` doesn't support one of these modalities, the problem is that the function itself is a fully native function, and we were trying to pass `ivy.Array` instances. Of course, `execute_with_gradients` should not be expected to detect and fix problems where the wrong input types are being passed to the function.
> 
> Ved's suggestion of:
> 
> ```
> func = lambda xs: actual_func(ivy.to_native(xs))
> ivy.execute_with_gradients(func, xs)
> ```
> 
> is exactly how we should solve things in this case I believe 👍
> 
> Please let me know if there are any other questions, or if you think I've missed something. Very happy to dive deeper!
> 
> Thanks both!

Yep, the current implementation of `execute_with_gradients` does support having `ivy.Array` or `ivy.NativeArray` in `xs` 🙂

ivy/stateful/module.py
YushaArif99: Now that you mention it, I just realized that its now unnecessary to change the function names. Actually my previous implementation of `to_<backend>_modules()` was based on multiple inheritence.for eg: `to_torch_module()` consisted of a module that was inherited from both `ivy.Module` as well as `torch.nn.Module.` 
But now I have changed the implementation to instead use composition(i.e store ivy_module as an attribute rather than inheriting from it). 
Anyways, I have reverted the changes. Thanks for pointing it out!

PR #6520, opened 2022-11-03
Add example docstring for argwhere function

iamlordaubrey:


iamlordaubrey:
Close 6351

iamlordaubrey:
Hey @Ishticode, thanks for taking a look.

With respect to moving the examples for Container, would this refer to both the `Container` input and the `Container instance` examples?

Ishticode (changes requested):
yes you are right about that.

iamlordaubrey:
Hey @Ishticode, I already moved examples for Container to the argwhere instance method in ivy/Container/searching.py. 

I see some failing tests, but I think it has to do with some other files. Should I attempt fixing them or is it okay for those to fail within this MR?

Ishticode (changes requested):
Hi the instance method example for the Array should be in the Array class mehtod argwhere.    The functional set of Container examples (not the instance one) can actually be placed in functional as they were (apologies for confusion) but you will need to remove the `ContainerWithSearching.argwhere(x)` part and replace it with just `ivy.argwhere(x)`

iamlordaubrey:
Hi @Ishticode, thanks for clarifying, and I am sorry if I am still getting it wrong. 
My understanding is to move the instance method example (and container instance method example) to the class method argwhere (in ivy/Container/searching.py); while the functional set of examples should remain in the functional argwhere (as they originally were).

I've implemented this. Is this correct/okay?

ivy/container/searching.py
```diff
@@ -424,6 +424,40 @@ def argwhere(
         -------
         ret
             Indices for where the boolean array is True.
+
+        Instance Method Examples
+        ------------------------
+
+        Using :class:`ivy.Array` instance method:
```
Ishticode, line 431: Just need to move this example to ivy/array/searching.py under the argwhere method there.

Ishticode (changes requested):
Nearly right. Just one change in the comment. Sorry for any confusion.

ivy/container/searching.py
iamlordaubrey: Moved this example to ivy/array/searching.py

Ishticode (changes requested):
Hi Great stuff. One minor change and this is ready to merge. Can you keep the title as just `Examples` for all three cases.

Ishticode (changes requested):

ivy/array/searching.py
```diff
@@ -191,5 +191,20 @@ def argwhere(self: ivy.Array, *, out: Optional[ivy.Array] = None) -> ivy.Array:
         ret
             Indices for where the boolean array is True.
 
+        Examples
+        ------------------------
```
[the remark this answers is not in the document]
iamlordaubrey: Oops... my bad. Sorry I missed that

PR #6564, opened 2022-11-04
Initial compiled Wrapping logic.

vaatsalya123:
Will add more generalized cache registers for the MRUcache method.Though for now this is the initial start will explore TTL based caching as well.I used eval because exec is not a Callable function!

docs for cachetools framework: https://cachetools.readthedocs.io/en/latest/
docs for compile function in python: https://appdividend.com/2019/07/25/python-compile-example-compile-function-in-python/


ivy/func_wrapper.py
```diff
@@ -40,13 +41,18 @@ def _get_first_array(*args, **kwargs):
             arr = ivy.index_nest(kwargs, arr_idxs[0])
     return arr
 
-
-# Array Handling #
+# Just a native python compile function with cache #
```
djl11, line 44: Just write "Compiling with Cache", and change the number of `-` so that the title aligns correctly please :)

djl11 (changes requested):

JamieLine:
Hi!

Because Daniel has already reviewed this, am I safe to assume I don't need to look through everything unless requested? Just want to make sure people aren't being held up.

vaatsalya123:
Hey @JamieLine yep you don't need to go through this pr Dan will go through all of it thanks.

ivy/func_wrapper.py
```diff
@@ -40,13 +41,18 @@ def _get_first_array(*args, **kwargs):
             arr = ivy.index_nest(kwargs, arr_idxs[0])
     return arr
 
-
-# Array Handling #
+# Compiling with Cache #
```
djl11, line 44: small detail but the number of `-` still don't align with the title length! aha

djl11 (changes requested):
tiny change requested aha, but do we still want to merge this. Maybe we should close until we have finalised the discussion around the best design. What do you think @vaatsalya123?

ivy-seed:

PR #6588, opened 2022-11-04
Changes required for the current version 1.1.10

omt66:
The example in the README is not working. In the Colab there is a working version that I updated the code here.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -111,6 +111,8 @@ You can immediately use Ivy to train a neural network, using your favorite frame
 
     import ivy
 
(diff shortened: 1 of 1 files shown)


CatB1t:
Hello @omt66, the README is updated for the next release, currently Ivy doesn't have `ivy.reduce_mean` or `ivy.set_framework` anymore.

omt66:
OK thanks, because when I first tried Ivy, I copied the code from Readme, and it didn't work with the current release, hence the reason for the PR :)

PR #6589, opened 2022-11-04
return import modules for numpy back

kurshakuz:
This PR resolves the issue with failing array api tests. However, a better approach is desirable


kurshakuz:
Closed in favor of https://github.com/unifyai/ivy/pull/6673

PR #6606, opened 2022-11-05
Update Accumulate

robertsn808:
- [ ] #6500

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_torch/test_blas_and_lapack_ops.py
+++ b/ivy_tests/test_ivy/test_frontends/test_torch/test_blas_and_lapack_ops.py
@@ -512,18 +512,6 @@ def test_torch_ger(
     )
 
 
-# inverse
-def test_torch_inverse():
-    # alias function
-    pass
-
-
-# det
-def test_torch_det():
-    # alias function
-    pass
-
-
 # logdet
 @handle_frontend_test(
     fn_tree="torch.logdet",
@@ -552,12 +540,6 @@ def test_torch_logdet(
         on_device=on_device,
         input=x,
     )
-
-
-# slogdet
-def test_torch_slogdet():
-    # alias function
-    pass
 
 
 # matmul
@@ -591,12 +573,6 @@ def test_torch_matmul(
         other=y,
         out=None,
     )
-
-
-# matrix_power
-def test_torch_matrix_power():
-    # alias function
-    pass
 
 
 # matrix_rank
--- a/ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
+++ b/ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
@@ -21,12 +21,6 @@ def _get_dtype_and_square_matrix(draw):
     return dtype, mat
 
 
-# diagonal
-def test_torch_diagonal():
-    # alias function
-    pass
-
-
 # inv
 @handle_frontend_test(
     fn_tree="torch.linalg.inv",
@@ -53,6 +47,7 @@ def test_torch_inv(
         input_dtypes=dtype,
         as_variable_flags=as_variable,
         with_out=with_out,
+        all_aliases=["inverse"],
         num_positional_args=num_positional_args,
         native_array_flags=native_array,
         frontend=frontend,
@@ -84,6 +79,7 @@ def test_torch_det(
         input_dtypes=dtype,
         as_variable_flags=as_variable,
         with_out=with_out,
+        all_aliases=["det"],
         num_positional_args=num_positional_args,
         native_array_flags=native_array,
         frontend=frontend,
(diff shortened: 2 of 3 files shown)


CerberusLatrans:
Hi @robertsn808, thanks for the PR! Your implementation currently just seems to be converting an array to an ivy array. You can see the functionality of numpy ufunc.accumulate here: https://numpy.org/doc/stable/reference/generated/numpy.ufunc.accumulate.html#numpy.ufunc.accumulate

What we have to do is to first create a ufunc class, then add accumulate as an instance method. You can see an example of this in the numpy matrix frontends submodule: https://github.com/unifyai/ivy/blob/master/ivy/functional/frontends/numpy/matrix/methods.py

Let me know if you have any problems or questions. Thanks!

CerberusLatrans:
Hi @robertsn808, nice start. The class unfunc should have a field to store the given operation  e.g. addition or multiplication (like how the matrix class stores fields for its shape, data, and dtype). 

The accumulate instance method should then refer to this field to perform its operations. One note is that accumulate should not be a special method, just an instance method. We also do not need out or axis methods.

Try looking at the numpy examples to understand the behavior. The operator starts as the identity, but changes as it progresses leftward, depending on the output of the previous element. e.g. the accumulating addition adds the previous output to the current element.

For now do not worry about the out argument-- this is used to modify an array (passed in as the out arg) in place instead of creating and returning a new one. 

Let me know if you have any other questions.

robertsn808:
Close #6500

CerberusLatrans:
I think this will make implementation a lot easier:
create an __init__ for the unfunc class and store the operation as a field
You then need to use the operation field in your implementation of accumulate (it is not an instance method, so do def accumulate instead of def __accumulate__)

robertsn808:
> I think this will make implementation a lot easier:
> create an __init__ for the unfunc class and store the operation as a field
> You then need to use the operation field in your implementation of accumulate (it is not an instance method, so do def accumulate instead of def __accumulate__)

My apologies. I don't quite understand what you mean.

CerberusLatrans:
If you are not familiar with python classes, please see 9.3.5 on https://docs.python.org/3/tutorial/classes.html#a-first-look-at-classes

The operation (e.g. addition) should be stored as a field aka instance variable (they use self.name as an example).
The accumulate method should be defined somewhat like how add_trick is.

Let me know if this is helpful.

robertsn808:
add_reformatting_checklist_ufunc

CerberusLatrans:
Hi @robertsn808, apologies for the late response. For some reason it does not directly notify me when you make the commit message @CerberusLatrans. Please just leave a comment instead, which I think will notify me.

I am a bit confused as to why 59 files have been changed. Are you still working on the numpy accumulate frontend? If not then that is fine and i will close this PR. However, if you are working on other to-do list tasks, please create a new PR for those, as they are not relevant to numpy update accumulate. Thanks.

robertsn808:
@CerberusLatrans 

I apologize for my inexperience. My intention is to complete Accumulate. I thought it needed to past tests? I'm not sure if my code is correct so I've been trying to do all that I can to learn. I'm sorry if my curiosity is troublesome.

CerberusLatrans:
Hi @robertsn808, that is no problem at all. Everyone has to start somewhere.

In order to get tests passing, that just means implementing the frontend method and implementing the frontend test (if the method is implemented correctly, the test should in theory pass, granted there is nothing else messing it up beyond our control). I know that the task checklist has a whole bunch of things on there, but these two are the most crucial ones, so lets just tackle these first.

Therefore, at least for now, you should only be making changes to `ivy/functional/frontends/numpy/ufunc/methods.py` and `ivy_tests/test_ivy/test_frontends/test_numpy/test_ufunc)/test_methods.py`

Did you manually add those lines to the dockerfiles, array.py,etc.? I suspect a lot of them were automatically generated somehow since the commit has almost 25,000 lines added. If you make edits to files that you don't want to be included in the commit, you can select only certain files to add, commit, and push.

Your implementation of accumulate is getting there. the __init__ does not need to take in an array as an arg.
The accumulate method needs to take in the same args as in the docs: `ufunc.accumulate(array, axis=0, dtype=None, out=None)` (you can see this is where the array gets passed in).

I would do `self.operator=op` instead.

Before implementing accumulate, make sure you understand the logic and examples in https://numpy.org/doc/stable/reference/generated/numpy.ufunc.accumulate.html#numpy.ufunc.accumulate
in their entirety. Think of the identity of the operation as a "base case."

Screen Shot 2022-11-22 at 9 14 03 PM
Here, the operation is addition, and we store the previous result in the local var `t`
`t` starts off as the identity of addition, which is 0 (since 0+x=x). If it were multiplication, then it would be 1.
We update `t` so that now it is the previous `t` plus A[0]: `t=0+2=2` (this is the 0th element of r)
We update `t` so that now it is the previous `t` plus A[1] : `t=2+3=5` (this is the 1st element of r)
We update `t` so that now it is the previous `t` plus A[2]: `t=5+5=10` (this is the 2nd element of r)
So we return `r=[2, 5, 10]`

Hopefully this makes sense. Try your best and let me know if you have any questions.

PR #6608, opened 2022-11-05
add logging warning for using sparse_grad

kurshakuz:


kurshakuz:
As discussed in the sync, gather will not through an exception for `sparse_grad=True` argument, as we currently don't support sparse matrices.

Issue #1537, opened 2022-06-19
Add Array Instance methods for Numpy Frontend


Aarsh2001:
Add [Array Instance methods](https://numpy.org/doc/stable/reference/arrays.ndarray.html#) for Numpy frontend:

- [ ] item
- [x] #4555
- [ ] itemset
- [ ] astype
- [x] copy
- [ ] view
- [ ] getfield
- [ ] fill
- [x] #6643
- [ ] resize
- [x] #4172 
- [x] #6711
- [ ] flatten
- [x] #6026
- [ ] squeeze
- [ ] take
- [ ] put
- [ ] repeat
- [ ] choose
- [x] #6485
- [ ] #6790
- [ ] partition
- [ ] argpartition
- [ ] searchsorted
- [ ] nonzero
- [ ] compress
- [ ] diagonal
- [ ] max
- [x] #2786
- [x] #5791
- [x] #5612
- [ ] ptp
- [ ] clip
- [ ] conj
- [ ] round
- [ ] trace
- [ ] #6620
- [ ] cumsum
- [x] #3945
- [ ] #6634
- [ ] std
- [ ] prod
- [x] #3781
- [ ] all
- [ ] any

sampreet-arthi:
I would like to take this up

Migabaj:
- [ ] #2786

gunesevitan:
- [ ] #3051

csowrab:
- [ ] #3304

fkhjoy:
#3545

sadra-barikbin:
- [] #3781

jucamohedano:
#3942

Harsimretsingh:
#3945

ddzhulaii:
transpose #4172

westrany:
tolist #4555

acrulopez:
flatten #5607

Oluwadurotimi10:
- [ ] #5612

vohoanglong0107:
- [ ] #5791

rotemgb:
- [ ] #5827

Oluwadurotimi10:
Hello, @JLU-Neal I have created a Pull Request for adding argmin to NumPy ndarray frontend. Kindly review and let me know if I should make any changes. (https://github.com/unifyai/ivy/pull/5636)

0xcro3dile:
- [ ] 5882

0xcro3dile:
- [ ] #5882

[...]

PR #6615, opened 2022-11-05
Create numpy reshape

hrak99:
I have selected a function 'reshape' from Add Array Instance methods for Numpy Frontend #1537. Kindly tell me is that function still available or if someone has already done it. As it is not cleared during creating new issue like purple or green circle on it. If that function is available to work, which relevant file has to be used for coding? Looking forward to hearing from you soon.


hrak99:
Also selected another issue 'diagonal' from Add Indexing routines for Numpy Frontend. So that anyone can be used to work on it. Kindly also tell me the relevant file too.

JoelPendleton:
Hi Muhammad, it now seems that numpy reshape is being attempted by somebody else. In order to claim a task that you wish to work, please comment your issue number as others have done in Add Array Instance methods for Numpy Frontend #1537. This will tell us you're working on this task and we can reference it and denote it as a green circle on the todo list on the top of the page.

PR #6691, opened 2022-11-07
cosh and cosh_ added to torch frontend and tested

rgrupesh:
close: #6685 
close: #6686 

Signed-off-by: Rupesh <rupeshgelal9@gmail.com>

### DIFF
--- a/ivy/functional/frontends/torch/tensor.py
+++ b/ivy/functional/frontends/torch/tensor.py
@@ -47,6 +47,13 @@ class Tensor:
         self.data = self.cos()
         return self.data
 
+    def cosh(self, *, out=None):
+        return torch_frontend.cosh(self.data, out=out)
+
+    def cosh_(self):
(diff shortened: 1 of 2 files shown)


rgrupesh:
@agniiyer

agniiyer:
Hi have you tried testing with all backends? There's still a bfloat16 data type error. Take a look and let me know when it's sorted out! :)

rgrupesh:
> Hi have you tried testing with all backends? There's still a bfloat16 data type error. Take a look and let me know when it's sorted out! :)

Can you please elaborate? Because the test passes on those implemented functions.
Screenshot from 2022-11-09 13-26-46

agniiyer:
Sure if you run the test with the flag `--backend all` you'll see the error. You can also see it in the test logs here on Github. However, I think this is an issue that needs to be fixed on our end, so I will go ahead and merge this! Nice work :)

PR #6738, opened 2022-11-08
ivy_atleast_1d

nassimberrada:
Hi @Ishticode , hope you're doing well 😄

Been managing on my own with extensions. While the tests pass locally, i'm still a little unsure about this type of functions who take *args kind of inputs as arguments. Could you maybe take a look at this one for e.g ? Particularly in terms of how the input is passed and how it's tested for.

Thanks in advance !

### DIFF
--- a/ivy/array/experimental/manipulation.py
+++ b/ivy/array/experimental/manipulation.py
@@ -571,3 +571,45 @@ class ArrayWithManipulationExperimental(abc.ABC):
             indices_or_sections=indices_or_sections,
             out=out
         )
+
+    def atleast_1d(
+        self: Union[ivy.Array, List[ivy.Array]],
+        /,
+        *,
+        out: Optional[ivy.Array] = None,
+    ) -> ivy.Array:
+        """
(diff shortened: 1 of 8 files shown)


Ishticode (changes requested):
Hi. I think most is fine but the `arys` arg should be passed **without** `*` operator in the `functional/ivy` (to atleast_1d). In case when a list of multiple entries is passed, it tries to unpack and pass each entry as an argument where there should only be one.

this can be checked with the following code
```python
import ivy
x = ivy.array([5])
y = ivy.array([2])
z = [x, y]
print(ivy.atleast_1d(z))
```
For the backends it makes sense to keep *. Feel free to experiment and let me know if I m mistakened or unclear. Thank Nassim.

nassimberrada:
Makes sense yeah, thanks for the clear explanation @Ishticode . I'll close the PR without making changes given that someone else opened an issue for this function, but i'll try to work on at_least_2d and/or at_least_3d instead. It follows the same logic so this is gonna be useful to take into consideration. Thanks !

PR #6748, opened 2022-11-08
Implement @handle_nans decorator

kurshakuz:
`@handle_nans` decorator:
Checks for the existence of nans in all arrays in the `args` and `kwargs`. The presence of nans is then handled depending on the enabled `nan_policy`. Following policies apply:
* raise_exception: raises an exception in case nans are present
* warns: warns a user in case nans are present 
* nothing: does nothing

Policy stack is present at `ivy.globals.nan_policy_stack`. Methods for modification of the curernt policy are: `ivy.get_nan_policy()`, `ivy.set_nan_policy()`, and `ivy.unset_nan_policy()`.

Decorator was tested on couple of numpy frontend functions and works as expected.


kurshakuz:
Hey @djl11! I tested the decorator with various methods, but have not yet attached it to any of them. Let me know if the current implementation is as you expected and I will then start attaching the decorator to required methods

ivy/func_wrapper.py
```diff
@@ -567,6 +569,68 @@ def _wrapped(func):
     return _wrapper_outer
 
 
+# nans Handling #
+# --------------#
+
+
+def _has_nans(x):
```
djl11, line 576: for maximal generalisation, `_has_nans` should make use `ivy.nested_any`, such as:

```
def _leaf_has_nans(x):
    if isinstance(x, ivy.Container):
        return x.has_nans()
    elif ivy.is_array(x):
        return ivy.isnan(x).any()
    elif x is float("nan"):
        return True
    return False

def _nest_has_nans(x):
    return ivy.nested_any(x, _leaf_has_nans)
```

ivy/func_wrapper.py
```diff
@@ -567,6 +569,68 @@ def _wrapped(func):
...
+        kwargs
+            The keyword arguments to be passed to the function.
+
+        Returns
+        -------
+            The return of the function, with handling of inputs based
+            on the selected `nan_policy`.
+        """
+        # skip the check if the current nan policy is `nothing``
+        if ivy.get_nan_policy() == "nothing":
```
djl11, line 613: we should cache this beforehand:
`nan_policy = ivy.get_nan_policy()`

ivy/functional/frontends/jax/func_wrapper.py
```diff
@@ -74,3 +87,50 @@ def new_fn(*args, **kwargs):
 
 def to_ivy_arrays_and_back(fn: Callable) -> Callable:
     return outputs_to_frontend_arrays(inputs_to_ivy_arrays(fn))
+
+
+def handle_nans(fn: Callable) -> Callable:
```
djl11, line 92: a lot of duplication, we should define this once in `ivy/func_wrapper.py` and then import this function into each frontend `func_wrapper.py` file, they are identical I believe right?

djl11 (changes requested):

ivy/func_wrapper.py
kurshakuz: Makes sense! Added in https://github.com/unifyai/ivy/pull/6748/commits/130b1ef83c5f6b13681370271c0b9cb767993003

ivy/func_wrapper.py
kurshakuz: Resolved in https://github.com/unifyai/ivy/pull/6748/commits/130b1ef83c5f6b13681370271c0b9cb767993003

ivy/functional/frontends/jax/func_wrapper.py
kurshakuz: Sure, that makes sense. I did it because I followed the same pattern as `inputs_to_ivy_arrays` that was also declared in multiple places.

Neverthess, fixed as requested and now can be directly imported from the `ivy.func_wrapper`. I did not added the `from ivy.func_wrapper import handle_nans` on top of the files as it will lead to unused import errors. Let 's just allow to directly import it from `ivy.func_wrapper` for all files.

djl11 (approved):
Makes sense, and yes lgtm! Thanks!

PR #6782, opened 2022-11-09
update experimental dashboards and workflows

ahmedo42:
Since we divided the experimental API into multiple submodules and it essentially became a mirror of the functional API, it would make sense to have two separate dashboards and workflows for `experimental-core` and `experimental-nn` , and for the dashboards to have the same appearance as the functional dashboards.

### DIFF
--- a/.github/workflows/test-ivy-experimental-core.yml
+++ b/.github/workflows/test-ivy-experimental-core.yml
@@ -1,4 +1,4 @@
-name: test-experimental-ivy
+name: test-ivy-experimental-core
 on:
   push:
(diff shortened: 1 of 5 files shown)


Aarsh2001 (approved):
Hey @ahmedo42, the workflow files look good. However, I'll have to update the update_db script and the JSON files on mongo. I'll pull this PR, make the necessary changes and merge it in an hour. Let me know if that sounds good, thanks🙂.

ahmedo42:
@Aarsh2001 Sounds good, Thanks!

PR #6789, opened 2022-11-09
Numpy reciprocal #6154

Abheast12:


Abheast12:
I'm unsure if what I am doing is correct, I would appreciate any feedback

Ishticode (changes requested):
Hi @Abheast12 Thanks a lot for your PR. 🙂.
There are some merge confilcts that need resolving. This happened for several reasons. One is that the updated file has seemingly replaced the older one instead of commiting the specific addtion. This appears to make a change in the whole submodule. Any change made on the repository will cause conflicts with the PR. You can either try to resolve the conflicts and request a review or open a new PR copying over just the changes you are making. Please feel free to let me know if anything is unclear. 🙂

PR #6805, opened 2022-11-09
pruned func_wrapper traces

juliagsy:
original:
```
...
During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/Users/julia/Documents/Ivy/ivy/ivy/func_wrapper.py", line 146, in new_fn
    ret = fn(*args, **kwargs)
  File "/Users/julia/Documents/Ivy/ivy/ivy/func_wrapper.py", line 84, in new_fn
  File "/Users/julia/Documents/Ivy/ivy/ivy/func_wrapper.py", line 360, in new_fn
    return fn(*args, **kwargs)
  File "/Users/julia/Documents/Ivy/ivy/ivy/func_wrapper.py", line 415, in new_fn
    return fn(*args, **kwargs)
  File "/Users/julia/Documents/Ivy/ivy/ivy/exceptions.py", line 66, in new_fn
    raise ivy.exceptions.IvyError(fn.__name__, str(e))
ivy.exceptions.IvyError: numpy: all: axis 1 is out of bounds for array of dimension 1
```

pruned:
```
...
During the handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/Users/julia/Documents/Ivy/ivy/ivy/exceptions.py", line 92, in new_fn
    raise ivy.exceptions.IvyError(fn.__name__, str(e))
```

### DIFF
--- a/ivy/exceptions.py
+++ b/ivy/exceptions.py
@@ -12,7 +12,7 @@ import traceback as tb
 def _print_new_stack_trace(old_stack_trace):
     print(
         "<func_wrapper.py stack trace is squashed,",
-        "call ivy.set_show_func_wrapper_traces(True) in order to view this>",
+        "call `ivy.set_show_func_wrapper_traces(True)` in order to view this>",
     )
     new_stack_trace = []
     for st in old_stack_trace:
--- a/ivy/exceptions.py
+++ b/ivy/exceptions.py
@@ -10,6 +10,10 @@ import traceback as tb
 
 
 def _print_new_stack_trace(old_stack_trace):
+    print(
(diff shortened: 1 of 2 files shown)


djl11 (changes requested):
This should be a configurable mode, and it should also be turned **off** by default.

Could you please also add just one line in the stack trace which explains that the func wrapper logic has been squashed? something like:

```
<squashed func_wrapper.py stack trace, call ivy.set_show_func_wrapper_traces() in order to view them in the stack trace>
```

or something similar, depending on what you deem to be the best approach for setting and unsetting this mode.

PR #6851, opened 2022-11-10
added an attribute named unsupported_gradients

RickSanchezStoic:
added an attribute named unsupported_gradients


ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -365,21 +370,46 @@ def test_function(
...
-            rtol_=rtol_,
-            atol_=atol_,
-            xs_grad_idxs=xs_grad_idxs,
-            ret_grad_idxs=ret_grad_idxs,
-            ground_truth_backend=ground_truth_backend,
-        )
+        if hasattr_unsupported_gradients and fw in fw_list:
+            if ivy.nested_argwhere(
+                all_as_kwargs_np,
+                lambda x: x.dtype in fw_list[fw] if isinstance(x, np.ndarray) else None,
```
VedPatwardhan, line 376: I think this line could be an issue, we should instead coompare the `input_dtypes` to the `fw_list[fw]` as they are guaranteed to be strings unlike `x.dtype`.

VedPatwardhan (commented):
Hi @RickSanchezStoic, just suggested a small change, rest looks good to merge 🙂

ivy_tests/test_ivy/helpers/function_testing.py
RickSanchezStoic: @VedPatwardhan if x is an instance of np.ndarray then x.dtype must be a string right?

PR #6871, opened 2022-11-10
Add frontend duplicate policy

ogbanugot:


docs/partial_source/deep_dive/ivy_frontends.rst
```diff
@@ -581,9 +583,37 @@ An example function using this is the :func:`numpy.isfinite` function.
...
+**Contributing duplicate frontend functions**
+
+Before working on a frontend function, contributors should check if the function is designated as an alias on the ToDo list.
+If the function is an alias, you should check if there is an existing implementation already.
+
+* If an implementation exist then you should use the existing implementation as an alias in your implementation.
+* If there is no existing implementation, then feel free to contribute the function with its full implementation and tests.
+* If two PRs are implementing a duplicate function (and no alias for that function exists), then reviewers should select the most advanced (comprehensive) implementation to serve as the alias while the second implementation should be refactored to use the alias. Both contributors should be rewarded equally if need be.
+
+Tests should be written for duplicate fronend functions using aliases. However, when defining the :code:`fn_name` in the test, be sure to use the full import path to the function.
```
djl11, line 608: could you please add a more concrete code example? Both aliases should be tested for in the same unit test. You should give the example of `np.asarray` as mentioned earlier.

ivy/functional/frontends/numpy/manipulation_routines/changing_kind_of_array.py
```diff
@@ -1,6 +1,6 @@
 # local
-import ivy
+from ivy.functional.frontends.numpy.creation_routines import asarray as as_array
 
 
 def asarray(a, dtype=None, order=None, *, like=None):
-    return ivy.asarray(a, dtype=dtype)
```
djl11, line 6: we can delete this function definition entirely, importing like so at the top of the file is enough to ensure that the function is included in this namespace:

```
from ivy.functional.frontends.numpy.creation_routines import asarray
```

ivy_tests/test_ivy/test_frontends/test_numpy/test_creation_routines/test_from_existing_data.py
```diff
@@ -18,7 +18,7 @@
         max_dim_size=5,
     ),
     num_positional_args=helpers.num_positional_args(
-        fn_name="ivy.functional.frontends.numpy.asarray"
+        fn_name="ivy.functional.frontends.numpy.creation_routines.asarray"
```
djl11, line 21: you should vary the `fn_tree` argument in `test_frontend_function`, which should be sampled from the combination of permitted namespaces. We should only have one unit test for this function, which should simply be the earliest of the two categories in the API. Currently we have two identical tests which is not what should happen.

djl11 (changes requested):

ivy/functional/frontends/numpy/manipulation_routines/changing_kind_of_array.py
```diff
@@ -1,7 +1,2 @@
 # local
-# import ivy
-
-
-# TODO: define this as an alias of asarray in from_existing_data.py
-# def asarray(a, dtype=None, order=None, *, like=None):
-#     return ivy.asarray(a, dtype=dtype)
+from ivy.functional.frontends.numpy.creation_routines import asarray
```
djl11, line 2: actually this might lead to a problem with flake8 linting. Is it saying unused import error? Maybe writing:

```
asarray = asarray
```

in the file will fix it, could you give this a try?

djl11 (changes requested):

ivy/functional/frontends/numpy/manipulation_routines/changing_kind_of_array.py
ogbanugot: yeah it's flagging it. Didn't catch it in my commit cause I merged from master (to fix the merge conflict) and pulled the change to my branch. Do we have to state the variable assignment explicitly in the docs as well?

ogbanugot: Sorry I'm not so clear, on the comment about handling it here  `ivy/functional/frontends/numpy/manipulation_routines/__init__.py`. Will we still import it and assign to a variable?

djl11: no we can just add this into the `__init__.py` file and it should work I believe:
```
from ivy.functional.frontends.numpy.creation_routines import asarray
```

djl11 (changes requested):
see comment

djl11 (approved):
lgtm!

PR #6904, opened 2022-11-11
Docs update - Backend Functional APIs codeblocks

Lrizika:
Updated codeblock examples of various `stack` backend API functions, which were quite out of date.
This was unfortunate, as there were several issues with the outdated versions which are now fixed.
While there are likely many other outdated codeblocks, this is one of the first pages a new user of Ivy will encounter when looking at the documentation.


abdrahmandiab (approved):
Hi @Lrizika , 
thank you so much for taking initiative with fixing the inconsistencies on this page!
I've added a few tiny changes to accommodate some other issues as well, but for the most part everything else is correct.

Thanks!

PR #6925, opened 2022-11-11
atleast_2d extension IvyAPI

### LINKED ISSUE #6883: atleast_2d

raghuveerbhat:
Close #6883

### DIFF
--- a/ivy/array/experimental/manipulation.py
+++ b/ivy/array/experimental/manipulation.py
@@ -589,3 +589,31 @@ class ArrayWithManipulationExperimental(abc.ABC):
                     [3, 4]]])
         """
         return ivy.dstack(self.concat(arrays), out=out)
+
+    def atleast_2d(self: ivy.Array, *arys: ivy.Array) -> List[ivy.Array]:
+        """
+        ivy.Array instance method variant of ivy.atleast_2d. This method simply
+        wraps the function, and so the docstring for ivy.atleast_2d also applies
+        to this method with minimal changes.
+
+        Parameters
+        ----------
+        self
+            Input array. Cannot be a scalar input.
+        arys
+            An arbitrary number of input arrays.
+
+        Returns
+        -------
+        ret
+            List of arrays, each with a.ndim >= 2. Copies are made
+            only if necessary.
+
+        Examples
+        --------
+        >>> a1 = ivy.array([[1,2,3]])
+        >>> a2 = ivy.array(4)
+        >>> a1.atleast_2d(a2,5,6)
+        [ivy.array([[1, 2, 3]]), ivy.array([[4]]), ivy.array([[5]]), ivy.array([[6]])]
+        """
+        return ivy.atleast_2d(self._data, *arys)
--- a/ivy/container/experimental/manipulation.py
+++ b/ivy/container/experimental/manipulation.py
@@ -1509,3 +1509,120 @@ class ContainerWithManipulationExperimental(ContainerBase):
             map_sequences=map_sequences,
             out=out,
         )
+
+    @staticmethod
+    def static_atleast_2d(
+        *arys: Union[ivy.Array, ivy.NativeArray, ivy.Container],
+        key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
+        to_apply: bool = True,
+        prune_unapplied: bool = False,
+        map_sequences: bool = False,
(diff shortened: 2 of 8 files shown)


JamieLine:
Hi! Thanks for the PR!

Most of this looks really good, there are only really 2 things I would change / have questions about:

1. The test only ever creates `int64` arrays. I'm not really sure why but it should draw from all of the `numeric` types. This fix shouldn't be too bad and shouldn't require any drastic work.
2. The torch backend converts the output of `torch.atleast_2d` to a `list`, and I'm not sure why that is done. If there is a reason for this conversion then please do let me know, but my intuition tells me that the conversion isn't needed.

The rest looks really good though, to only have 2 minor things to look at over that volume of code is seriously impressive.

If you have any questions or comments please let me know!

raghuveerbhat:
Hi @JamieLine , thank you so much for the quick review :)
My answers to your questions:
>The test only ever creates int64 arrays. I'm not really sure why but it should draw from all of the numeric types. This fix shouldn't be too bad and shouldn't require any drastic work.

Thank you for noticing this. I've made the required changes.

>The torch backend converts the output of torch.atleast_2d to a list, and I'm not sure why that is done. If there is a reason for this conversion then please do let me know, but my intuition tells me that the conversion isn't needed.

My reason for this was to unify the return behaviour across al backends (since torch backend return tuple and every other backends return list). Also I referred to the implementation of `broadcast_arrays()` where the behaviour is same (return list instead of tuple for torch backend) -  pull request\([#2261](https://github.com/unifyai/ivy/pull/2261)\). Please let me know your thoughts on this reasoning and if changes are required I'll do it :)

Also while working on you comments, I found one possible helper function which could be implemented.

1. Since there is already an helper function `get_dtypes()` which draws a list of dtypes and has support of `key` argument, I felt maybe a similar helper function `get_shapes()` which also draws list of shapes with `key` argument support could be implemented. The current `get_shape()` doesn't have this behaviour and we can sort of skip calling get_shape() inside the loop. The function can be implemented in `shape_helpers.py`. But I guess this can be part of another PR.

Please let me know your thoughts on this.

raghuveerbhat:
Hi @JamieLine , thank you so much for the quick review :)

>The test only ever creates int64 arrays. I'm not really sure why but it should draw from all of the numeric types. This fix shouldn't be too bad and shouldn't require any drastic work.

Thank you for noticing this. I've made the required changes.

>The torch backend converts the output of torch.atleast_2d to a list, and I'm not sure why that is done. If there is a reason for this conversion then please do let me know, but my intuition tells me that the conversion isn't needed.

My reason for this was to unify the return behaviour across al backends (since torch backend return tuple and every other backends return list). Also I referred to the implementation of `broadcast_arrays()` where the behaviour is same (return list instead of tuple for torch backend) - pull request\([#2261](https://github.com/unifyai/ivy/pull/2261)\). Please let me know your thoughts on this reasoning and if changes are required I'll do it :)

Please let me know your thoughts on this.

JamieLine:
Hi! 

Thanks for the quick response! Your changes sound good as does your reasoning for the list conversion. I assumed that each backend returned one of their native array types (like a `torch.Tensor` or similar) but if they aren't then it makes sense.

All looks good other than the `torch` backend producing a test failure. When I ran the tests with all backends (by adding `--backend=all` as an argument to PyTest, let me know if you'd like some help adding this parameter) Torch had an error, which gave the snippet

```You can reproduce this example by temporarily adding @reproduce_failure('6.55.0', b'AAAAAAEAAQEAAAAAAAAAAAA=') as a decorator on your test case```

Which will help to reproduce it.

If you need any help or have any comments, please let me know!

Thanks!

raghuveerbhat:
Hi @JamieLine ,

Thank you for your response. I'm glad the reasoning was okay. 

About the failing test case, this was indeed a bug in the torch backend. So I fixed it :)

But I wonder why it didn't get caught in my environment. If you know any way to increase the number of test examples, or if there are any other setting, I would be happy if you can share it with me:)

JamieLine:
Hi!

There is a way to increase the number of test examples, if you import `hypothesis.settings` and then add `@settings(max_examples=X)` to the test function it will run `X` examples.

You'd need to remove these from anything before it can be merged, but it is useful to use in development.

Here is an example:
image


Thanks!

raghuveerbhat:
Thank you so much for this information. This is super helpful :+1:

PR #6953, opened 2022-11-12
Add logos to frameworks and wrapper-frameworks pages

kurshakuz:
Sample from frameworks page:
image

Sample from wrapper-frameworks page:
image

Exchange formats:
image

Vendor specific APIs:
image


djl11 (approved):
lgtm!

PR #7000, opened 2022-11-14
Ivy zeta

nassimberrada:


Ishticode (changes requested):
Hi Nassim. This code shows some discrepancy between the frameworks. (numpy fails)
```python
import ivy
for fw in ['jax', 'torch', 'tensorflow', 'numpy']:
    ivy.set_backend(fw)
    print(fw)
    print(ivy.zeta(ivy.array([1.,2,3]), ivy.array([2.])))
```
The result we should be getting is `ivy.array([       inf, 0.64493406, 0.2020569 ])` but jax gives `ivy.array([8.7876997])`

Did you encounter any such examples?
Sorry for bringing in a problem rather than solution. I promise to take a better look at the code tomorrow. Meanwhile let me know if this example gives any insight.

Ishticode:
My last push was accidental.

Ishticode (changes requested):
For jax we already have an implementation... you can import using
```python
from jax.scipy.special import zeta
```
Explained [here](https://jax.readthedocs.io/en/latest/_autosummary/jax.scipy.special.zeta.html).

nassimberrada:
Hi @Ishticode ! Thanks for pointing out that example, made me realize there was a lot of holes that could be poked in that code, not sure how the CI overlooked all of those the first one 😀
I made some changes and tested with your example as well as other corner cases, should be working now hopefully.
As for jax's native implementation of zeta i had seen it but i think i had a couple issues that made me unsure about using it, for e.g it states that the q argument can be optional but when you don't pass it it tells you the Riemann function is not implemented yet...Also i wasn't sure if we could make calls to jax.scipy. In any case the implementation i wrote for jax is the same than the one for numpy. 
Let me know what you think and if there is any other issue or something that could be improved !

Ishticode (changes requested):
Ignore this. It was becaue I hadn't pulled the changes.

nassimberrada:
Close #7150

PR #7019, opened 2022-11-14
nanquantile

### LINKED ISSUE #7014: nanquantile

Change-m:
Close #7014


zaeemansari70 (changes requested):
Hi!
I've requested a few changes.
Thanks! 🙂

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -182,3 +182,53 @@ def test_unravel_index(
...
+    num_positional_args,
+    native_array,
+    container_flags,
+    instance_method,
+    backend_fw,
+    fn_name,
+    on_device,
+):
+    input_dtype, x, axis = dtype_x_axis
+    helpers.test_function(
```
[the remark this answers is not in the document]
zaeemansari70: the first argument should be
`ground_truth_backend=ground_truth_backend,`

zaeemansari70 (changes requested):
I've requested a change. :)

The tests are failing with the error `test_function() missing 1 required keyword-only argument: 'ground_truth_backend'`, I've mentioned where you should add it.

Thanks!

zaeemansari70 (changes requested):
Awesome, 
Just a few more more changes 🙂

The tests on the CLI are failing, just mentioning the error for your convenience,
`E   ivy.exceptions.IvyBackendException: torch: nanquantile: nanquantile() received an invalid combination of arguments - got (Tensor, Tensor, keepdims=bool, method=DataObject, overwrite_input=bool, out=NoneType, interpolation=DataObject, axis=int), but expected one of:
E    * (Tensor input, Tensor q, int dim, bool keepdim, *, str interpolation, Tensor out)
E    * (Tensor input, float q, int dim, bool keepdim, *, str interpolation, Tensor out)`

You can use the debugger and check what `type` of arguments are being passed through, this seem to be an error related to the typehints. 🙂
Thanks!

ivy/functional/backends/tensorflow/experimental/statistical.py
```diff
@@ -45,3 +45,38 @@ def unravel_index(
...
+    axis: Optional[Union[int, Tuple[int]]] = None,
+    keepdims: Optional[bool] = False,
+    interpolation: Optional[str] = None,
+    out: Optional[Union[tf.Tensor, tf.Variable]] = None,
+) -> Union[tf.Tensor, tf.Variable]:
+    count = 0
+    ret = []
+    res = 0
+    for i in a:
+        for j in i:
```
zaeemansari70, line 64: when you go into the nested loop, the value becomes a scalar and you cannot iterate through a scalar, also for the implementation of this I am referring you to a link from where you can understand what `nanquantile()` actually does and use the formula used there for the it's implementation.
https://www.statisticshowto.com/quantile-definition-find-easy-steps/

ivy/functional/backends/torch/experimental/statistical.py
```diff
@@ -47,3 +47,35 @@ def nanmean(
...
+    a: torch.Tensor,
+    q: Union[int, float],
+    /,
+    *,
+    axis: Optional[Union[int, Tuple[int]]] = None,
+    keepdims: Optional[bool] = False,
+    interpolation: Optional[str] = None,
+    out: Optional[torch.tensor] = None,
+) -> torch.Tensor:
+    if hasattr(axis, "__iter__"):
```
zaeemansari70, line 62: This shouldn't be used, the whole implementation should simply just be `return torch.nanquantile(
            a,
            q=q,
            dim=axis,
            keepdim=keepdims,
            interpolation=interpolation,
            out=out,
        )`

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -182,3 +182,50 @@ def test_unravel_index(
         indices=np.asarray(x[0], dtype=input_dtype[0]),
         shape=shape,
     )
+
+    
+# nanquantile
+
+@handle_test(
+    fn_tree="functional.experimental.nanquantile",
+    dtype_x_axis=statistical_dtype_values(function="nanquantile"),
```
zaeemansari70, line 191: you are using `nanquantile`, there is no condition in the function which handles `nanquantile()` in `statistical_dtype_values()`, there is handling for `quntile()`, you can use that, as the both the functions accepts the same inputs. You were also generating array of integers which is not accepted by torch and it was causing an error, this will also fix that.
This should be statistical_dtype_values(function="quantile")

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -182,3 +182,50 @@ def test_unravel_index(
...
+    as_variable,
+    with_out,
+    native_array,
+    container_flags,
+    instance_method,
+    backend_fw,
+    fn_name,
+    on_device,
+):
+    input_dtype, x, axis = dtype_x_axis
```
zaeemansari70, line 213: for unpacking the values, this should be:
`input_dtype, x, axis, interpolation, q  = dtype_x_axis`

zaeemansari70 (changes requested):
I've requested a few changes🙂
After these the implementation of `numpy`, `jax` and `torch` would be producing the correct output.
After these changes, you need to correct your implementation for `tensorflow` only
Thanks! 🙂

PR #7066, opened 2022-11-15
Add Std Instance Method to NumPy Frontend 

nkkrnkl:
It's not passing the tests yet. Any help would be much appreciated. Issue #6875


abdrahmandiab:
Hi @nkkrnkl, 
As we are currently undergoing some updates to our testing infrastructure, we are skipping all instance method tests due to an issue that is being solved 😅 

As such, we'll be merging this PR and you can go ahead with your application process, and we will fix any failures or errors with the tests ourselves if there happens to be any of those (hopefully not 😋), so you don't need to worry about fixing it in the future or anything.

Thanks for your hard work, and good luck with the remainder of your interview process! 🚀

PR #7074, opened 2022-11-15
Added interpolation function to torch front-end

artykov1511:


artykov1511:
interpolate #7062

RashulChutani (changes requested):
Hi @artykov1511, It seems you have also committed the .idea/ directory. Can you please remove the same? Thanks.

PR #7078, opened 2022-11-15
__mod__ #7073

StefanHUB:


guillesanbri:
Hi! Thanks a lot for contributing to Ivy :slightly_smiling_face: Looks like something happened when commiting changes and various files have been entirely modified, could you fix this? Not sure what happened but I would check the end-of-line characters and see if that's the problem. You can see the modified files in the `Files changed` tab. Thanks!

StefanHUB:
> 

Hi, thanks for the information!
It is my fault, as I probably did not manage to set up  the environment correctly. I will set up Codespaces and then try again!

PR #7082, opened 2022-11-15
Reformatting RELU function, added docstring examples

dimikave:


dimikave:
add_reformatting_checklist_relu

dimikave:
Close #6982

dimikave:
As far as I can see, the docstrings related tests that fail, seem to fail on other functions and not on relu. As for the other tests, I am not sure why the fail or if they fail in the original branch as well. 

Looking forward for your review,
Thanks in advance

ivy/functional/ivy/activations.py
```diff
@@ -95,8 +108,8 @@ def leaky_relu(
     ret
         The input array with leaky relu applied element-wise.
 
-    Functional Examples
-    -------------------
+    Examples
+    --------
```
Daniel4078, line 113: ivy/functional/ivy/activations.py:94 in public function `leaky_relu`:
        D412: No blank lines allowed between a section header and its content ('Examples')

Daniel4078 (changes requested):
Hello, I only found one small error for you to fix. Thank you.

ivy/functional/ivy/activations.py
dimikave: Hello @Daniel4078 , as far as I can say, leaky_relu is a different function tracked by someone alse on the specific issue: https://github.com/unifyai/ivy/issues/6626

As I understand I should not interfere with someone else work, is that correct? In any case, if you still want me to fix this error it is pretty easy and minor and of course I'll handle it. Can you please clarify?

Daniel4078: Well, I just saw in the change log and find this error is from a place that you changed. Of course you can fix existing errors by others as long as no new critical error are introduced. Thank you!

Daniel4078 (approved):
Well done!

PR #7083, opened 2022-11-15
Implements the convolve function in the numpy frontend

DarthEmpty:
- [ ] #7075


nathzi1505:
Hi @DarthEmpty, thanks for the PR. 
I see some tests especially the lint-check and the test-frontend-numpy not passing. 
Mind having a look at those? Also, if you haven't pull the new commits from the master branch as in those these two checks are OK. For related help you can also see here : https://lets-unify.ai/ivy/contributing.html.

Just have a look again and submit it for review. I will look at the changes again. 😊

Issue #779, opened 2022-04-02
Reformat sorting


saeedashrraf:
Each method in the [sorting](https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/sorting.py) submodule should be modified to use the latest formatting rules, as explained in the [contributor guide](https://lets-unify.ai/ivy/contributing.html) for the [Ivy API](https://lets-unify.ai/ivy/contributing.html#ivy-api) and [Backend API](https://lets-unify.ai/ivy/contributing.html#backend-api). This includes removing all lambdas and direct bindings for methods in the backend submodules, with each method instead defined using `def method_name`. Full type hints should also be given for each argument in both the ivy module and also each backend module. The docstrings should be updated to use NumPy style rather than Sphinx style, and code examples should be added to **every** method docstring.

- [ ] argsort
- [ ] sort

bideeen:
#783

bethelhall:
- [ ] #800

Daniel4078:
argsort #1104

Monty0252:
#1514

VidyashreeEshwarappa:
can I work on the issue sort ? or is it done?

VidyashreeEshwarappa:
Sort #1625

DankGunther:
Hello, can I do this task?

VidyashreeEshwarappa:
I have created PR for this task already. kindly pickup something else

ukj02:
#2319

DrHuaZhao:
- [ ] #2355

Adlai-1:
- [ ] #4671

trihernandez:
#5126

anurag629:
- [ ] #6558

chkoutsis:
- [ ] #9389

Abhinav-CHD:
Hey is anyone working on this issue? If not, then I would like to work on it.

vishwasjoshi2019:
hey @Abhinav-CHD I'm working on it

Abhinav-CHD:
> hey @Abhinav-CHD I'm working on it

Okay

anudeeps0306:
@Abhinav-CHD  can you check my pr request #23593

Bhargav-Dasari:
- [ ] #25680

GOW7527:
- [ ] #26044

2202-Aryan:
#26571

PR #7119, opened 2022-11-17
update sorting

AntoniosMalak:


AntoniosMalak (commented):
update

AntoniosMalak:
What should I do else?
@ivy-leaves @JamieLine

JamieLine:
Hi! Thanks for the PR!

I've had a quick look through, and I have some questions / concerns. If you disagree with / would like to question anything I say here, please feel free. If I mention that something happens in a specific function, and it needs to be changed, please perform that change in the other affected functions if needed. Remember that just because I am asking why a change is made doesn't necessarily mean it has to be reverted, it just means I am struggling to see what the rationale behind the change is.

1. The docstring for `argsort` links to `tan` rather than `argsort`. This should be reverted to link to `argsort`.
2. In the docstring for `argsort`, there is a `:code:` tag that is incomplete (it says `code:` and is missing the first `:`)
3. The header in the docstring for `argsort` says `Functional Examples` when it should say `Examples`.
4. Why were some examples removed from `argsort`?
5. Why were the calls to the backend `argsort`, `sort`, and `searchsorted` functions changed to call the backend with a positional argument rather than a keyword argument? In my tests, the functions don't run correctly with this set-up, instead they raise a `TypeError: argsort() takes 1 positional argument but 4 positional arguments (and 1 keyword-only argument) were given` instead.
6. Why have the `handle_array_like` and `handle_exceptions` decorators been removed?
7. Why has `sort` had almost its entire docstring removed?
8. Why has the `# Extra #` comment heading been removed? It was after `sort`, not in a docstring but in between functions.
9. In `searchsorted`, the argument `ret_dtype` has been given a type-hint of `int` but the default value is `ivy.int64` which is an `ivy.IntDtype` object. The type hint should probably be either an `ivy.IntDtype` or a parent class of that type depending on which is more appropriate.
10. In `sort`s docstring, it says "with the same dtype and shape as values" but `values` isn't defined, it should be `x`.
11. The blank lines between the examples for `sort` should be restored, removing those lines breaks the tests for the examples as the test will be unable to tell where each example starts and ends.

If you have any questions or comments please let me know!

AntoniosMalak:
Hi, @JamieLine thanks for your comments!
Can you check it now I changed the code as you mentioned?

AntoniosMalak:
Hi, @ivy-seed can you check it know?

JamieLine:
Hi! I cannot review it as I no longer work for Ivy, all on good terms I just need to focus on my studies. @CatB1t will review it in good time, I believe the standard response time is up to one week.

AntoniosMalak:
Okay, and Thanks for your time and previous comments and I hope you will do great in your study!

AntoniosMalak:
Hi, @CatB1t  may you review it.

CatB1t:
Hello @AntoniosMalak , can you please reference the issue number?

AntoniosMalak:
Yeah, sure. Here it is #7116

AntoniosMalak:
Hi, @CatB1t may you check my code to merge or close pr or make an update in code, it's been a week ago!

CatB1t:
Hello @AntoniosMalak , your issue doesn't seem to reference a task form the ToDo list, can you please make sure your issue is referenced in the ToDo list, so I can check which task you're trying to complete? Thanks!

AntoniosMalak:
This is Reformat sorting #779 in ToDo list.

AntoniosMalak:
Hi, @CatB1t Can you review it now?

AntoniosMalak:
Hi, @CatB1t Can you review it, pleases? it has been 10 days without any review so may you check it to close this pr or edit the code!

ivy/functional/ivy/sorting.py
```diff
@@ -254,11 +252,10 @@ def sort(
 def searchsorted(
     x: Union[ivy.Array, ivy.NativeArray],
     v: Union[ivy.Array, ivy.NativeArray],
-    /,
+    side: str = "left",
```
CatB1t, line 255: Why exactly are we moving these to be positional only? we should follow the Array API standard, but given that they don't have `searchsorted` in their standard, we follow the one that is most similar, in our case `argsort` and `sort`, therefore we don't need to do these changes.

CatB1t (changes requested):
Hello @AntoniosMalak, you seem to be making multiple changes which isn't related to the issue, can you revert those?

AntoniosMalak:
Hi @CatB1t, I did these changes as mentioned in this [comment](https://github.com/unifyai/ivy/pull/7119#issuecomment-1318855395), but I revert these things you mentioned can you check it now?

AntoniosMalak:
Hi, @CatB1 may u review it, please?

AntoniosMalak:
@CatB1t a week passed!

ivy/functional/ivy/sorting.py
```diff
@@ -321,4 +320,5 @@ def searchsorted(
         sorter=sorter,
         out=out,
         ret_dtype=ret_dtype,
+        x, v, side=side, sorter=sorter, out=out, ret_dtype=ret_dtype,
```
CatB1t, line 323: These are already passed in, could you revert that?

ivy/functional/ivy/sorting.py
```diff
@@ -138,12 +138,12 @@ def argsort(
 @handle_array_like
 def sort(
     x: Union[ivy.Array, ivy.NativeArray],
+    axis: int = -1,
```
CatB1t, line 141: According to the array API standard, this is not correct. look here: https://data-apis.org/array-api/latest/API_specification/generated/signatures.sorting_functions.sort.html

CatB1t (changes requested):
Also could you revert changes that are not related to your issue #7116 ? 
on a final note, seems like the function is already reformatted, but that wasn't marked properly on the ToDo list! we apologize for that, could you please pick a up another task to work on? 😅

AntoniosMalak:
@CatB1t So I should pick another one after 3 weeks ago :( 
Okay should I close this pr or just pick another one?

PR #7143, opened 2022-11-17
torch frontend ViewTensor class Update

JerryGCDing:


ivy/functional/frontends/torch/tensor.py
```diff
@@ -23,61 +23,61 @@ def add(self, other, *, alpha=1):
...
+        self.data = self.add(other, alpha=alpha).data
         return self.data
 
     def asin(self):
         return torch_frontend.asin(self.data)
 
     def asin_(self):
-        self.data = self.asin()
+        self.data = self.asin().data
         return self.data
```
djl11, line 34: Why exactly do these return `self.data`? Should they not return `self`? The current implementation will return an `ivy.Array` which is wrong I believe. Can you change these to all return `self` instead?

ivy/functional/frontends/torch/tensor.py
```diff
@@ -23,61 +23,61 @@ def add(self, other, *, alpha=1):
...
 
     def atan_(self):
-        self.data = self.atan()
+        self.data = self.atan().data
         return self.data
 
     def view(self, shape):
-        self.data = torch_frontend.reshape(self.data, shape)
-        return self.data
+        view = torch_frontend.ViewTensor(weakref.ref(self), shape=shape)
```
djl11, line 79: this can be done in a single line

djl11 (changes requested):
haven't dived too deep but looks good to me. Could you please add unit tests which verify that the view tensor behaves as expected? Feel free to implement this test using pytest.parameterize rather than fully generalized via hypothesis. Thanks!

JerryGCDing:
Added a `pytest.mark` keyword check in the fixture within `confest.py` file for frontend tests so that it doesn't prevent parameterized tests from running. The behavior test design for `ViewTensor` class is that for non-inplace instance methods, the test will check if the returned values are the same as its base tensor. For inplace methods, the test will check if the base tensor is being properly updated.

djl11 (approved):
looks good to me! I'll also just request a review from Mustafa to verify that the changes to `conftest.py` are good, thanks!

CatB1t (changes requested):
Hello @JerryGCDing, I've pushed a fix for `conftest.py` which should run all tests normally, you won't need to use a mark for it, feel free to merge after removing the markers.

PR #7152, opened 2022-11-18
Ivy Converters Implementation (Revised)

YushaArif99:


YushaArif99:
Hey @djl11,
I have made the changes you've requested. The previous PR became a total mess due to the lengthy conversations and edits so I thought to close that one and start a new one here. Could you kindly review the changes to see if everything is fine now. 
Oh and one more thing, the tests for converters are failing because of the `to_torch_module()`. The issue is a very small one but since its not related to the converters, i don't want to modify any other files. 
Basically the error is related to the PyTorch reshape function. It only accepts a tuple of ints as the argument for reshape. But instead, because of the decorators, the `shape` argument instead ends up becoming a native array(i.e torch Tensor) in the `reshape` function of torch backend which raises an exception  `ivy.exceptions.IvyBackendException: torch: linear: torch: reshape: reshape(): argument 'shape' (position 2) must be tuple of ints, not Tensor`
A small workaround is to explicitly convert the tensor to a list in the backend implementation (`ivy/functional/backends/torch/manipulations::reshape`). 
ie: `shape= shape.tolist()`
However, I'll leave it up to you to make the call 🙂

YushaArif99:
Hi, @djl11 
**_Quick update_**: I have made a small fix to the `@handle_array_like decorator` which was causing the converter tests to fail. Let me know if the change seems fine or not. Thanks!

djl11 (approved):
Looks okay to me on first glance, but I'll get @VedPatwardhan to take a deeper look as well. Thanks!

djl11:
@Ishticode could you also please weigh in on the `handle_array_like` changes? And @bicycleman15 could you please weigh in on the converters, given that you've recently dived a bit deeper into the `ivy.Module` class and will be continuing to work on the transpilations of the `ivy.Module` etc. Thanks!

Ishticode (approved):
Yes `inspect` is the right way to go in `array_handler_like` here instead of `typing` module functions. 

Reason being the latter mixes up the order of args in certain cases and includes `return` too both of which leads to unexpected conversions and failures.

VedPatwardhan (commented):
I think the changes would work well with gradients. 

Just wasn't sure if importing `haiku`, `jax`, `torch` and `tensorflow` inside `stateful/module.py` would be a good idea as I thought it would add a dependency on those modules rather than them being optional, which wouldn't scale well with more frameworks that could be added in the future. 

Please let me know what y'all think. Thanks 🙂

djl11:
I agree with @VedPatwardhan, we shouldn't import these in the module directly. We should use try-except style logic, so that it imports and runs fine even if frameworks are not installed. @JoelPendleton has been working on this, could you make any necessary changes here so that the attempted imports fail gracefully?

JoelPendleton:
Hi, I have made the changes you requested @djl11. I've tested it against [my larger set of changes](https://github.com/unifyai/ivy/pull/7352), to enable testing when only a subset of of frameworks of frameworks are installed, and it works fine.

JoelPendleton (approved):

bicycleman15 (approved):
Looks fine in my opinion, my only concern is that `ivy.Module` class is getting bloated up with a bunch of converter methods, but maybe that is un-avoidable here, converter functions are important to `ivy.Module`.

PR #7156, opened 2022-11-18
Numpy order implementation + ivy flatten

fspyridakos:
Everything works as expected, except ivy flatten tests fail for some reason and numpy ravel fails for unknown reason too.


fspyridakos:
Pushing flatten seperately for Ved to take a look messed up this PR so I'll make a new one with only the frontend order changes.

PR #7209, opened 2022-11-20
nll_loss

### LINKED ISSUE #7140: nll_loss

Gabcddrc:
Close #7140 
Add loss functions to PyTorch Frontend, added nll loss.


Gabcddrc:
Implemented nll loss for PyTorch frontend, referencing issue nll_loss #7140.

Gabcddrc:
Close #7140

WilliamHirst:
There are a couple of index errors. Please see the error log, specifically check out details of "test-frontend-torch" and then the "Run Frontend Test". Search for "Falsifying example: test_torch_nll_loss" to find the relevant erros.
Thanks 😊

Gabcddrc:
There appears to be a formatting issue, tho not from my files.
My test seems to be getting assertion errors, however, the same error appears to also exist within other loss function tests.
I manually tested my function by re-implement it with PyTorch functions, and the result seems to match torch's nll loss one-to-one. So the problem seems to be with the test itself, however, I am struggling to identify the exact issue.

WilliamHirst:
As for the formatting error I am experiencing the same, so this is most likely something wrong elsewhere. 

The error in test-frontend-torch I still belive is the fault of your PR, though I am not 100% sure. The error might be in the test not the frontend-func.  Please go over the log thoroughly and let me know if you still belive the error to be elsewhere..

WilliamHirst:
Hi, 
As for the formatting errors, these are not made from your PR and I have fixed them in main.

There seems to be a problem with the dimensionality of your input for the test. Try to look more into this.

Let me know when you have made changes you belive are ready for a review.
Thanks! 😊

Gabcddrc:
The dimension of the target and the weight seem to always match the dimension of the input, despite I have set different max and min dims. The dimension of the target should be restricted to [1] given the bounds, yet it is [2,1] matching the input dimension.

Gabcddrc:
After, fixing all the dim to 1, I am getting  " RuntimeError: "nll_loss_out_frame" not implemented for 'Half' ", do you know what this error means?

WilliamHirst:
I am not a 100% certain, but after digging into the code and a little on google I suspect it could be due to the other error you are getting. In other words the fact that it is expecting tf.float32 but is receiving tf.float16.

Gabcddrc:
Currently, no error is from my tests, instead, it is failing at a different file instead. I have set the expected type from float32 to float16 for my function. Do you know how to set the type in the test to float32, also it doesn't seem like I can just use the generic ivy.float type similar to torch, instead I must specify a precision level.

WilliamHirst:
Hi! Yes you are right, to call the ivy float you have to specify the precision (i.e 16 or 32). Is there a reason you want to specify float32? Because it seems as though the test generates data with float16.

Gabcddrc:
I just thought float32 might be more widely used.  Have you had a look at the current error? It is coming from a different file, and it appears that the tests did not continue to run after that failed test.

WilliamHirst:
Yes I agree. Have you tried not setting the dtype explicitly? This would force the function to retrieve the dtype implicitly from the argument array. 

As for the failing tests I will have a quick ask around. It seems weird that all other tests are now passing.

Gabcddrc:
looks like not setting the dtype works, assuming the tests ran correctly, I think this is ready for review.

WilliamHirst:
Looks great, good job! 👍

PR #7219, opened 2022-11-21
numpy logic function

ndurumo254:
adding isnat function


ndurumo254:
hello there, how can i prevent some of the test from being passed?

ndurumo254:
I would also appreciate if I know what this error really means

PR #7220, opened 2022-11-21
added tensorflow ragged tensor class, more methods to be added

RickSanchezStoic:


ivy/functional/frontends/tensorflow/ragged/ragged.py
```diff
@@ -0,0 +1,135 @@
...
+class RaggedTensor:
+    def __init__(self, values, row_partition, internal=False, data=None):
+        if not internal:
+            raise ivy.exceptions.IvyException(
+                "RaggedTensor constructor is private; please use one of the "
+                "factory methods instead "
+                "(e.g., RaggedTensor.from_row_lengths())"
+            )
+        self._values = values
+        self.data = data
```
fspyridakos, line 19: If `self.data` stores an ivy array/container, could we name it as such. Like how we renamed the data attribute in the frontend array/tensor classes to ivy_array

RickSanchezStoic: yes we can change this when we have NestedArray class implemented.

PR #7251, opened 2022-11-22
Numpy order implementation

### LINKED ISSUE #6859: flatten

fspyridakos:
closes #6859


fspyridakos:
Should pass all the tests, will need to change some parts of the testing to fully cover order testing but will do that seperately

ivy/functional/frontends/jax/func_wrapper.py
```diff
@@ -60,16 +103,45 @@ def new_fn(*args, **kwargs):
...
+            )
 
     return new_fn
 
 
 def to_ivy_arrays_and_back(fn: Callable) -> Callable:
     return outputs_to_frontend_arrays(inputs_to_ivy_arrays(fn))
+
+
+def outputs_to_frontend_arrays_with_order_manipulation(fn: Callable) -> Callable:
```
djl11, line 127: this function `outputs_to_frontend_arrays_with_order_manipulation` seems identical to `outputs_to_frontend_arrays`, with the only difference being the addition of `order=order` when calling the original function `fn`. I would suggest we handle all of this inside the original `outputs_to_frontend_arrays`.

We can use the `inspect` module to see if `order` is a keyword argument of `fn`, and we can determine this inside the scope of `outputs_to_frontend_arrays ` but outside the scope of `new_fn`, which means this will only be called once when the function is first decorated. When will then have a boolean variable such as `contains_order` defined in the outer scope, which can be evaluated inside `new_fn`. Then we can simply do the following:

```
ret = fn(*args, order=order, **kwargs) if contains_order else fn(*args, **kwargs)
```

If `order` does not have any effect on the output (it is mathematically irrelevant for this function), then it's not a problem to pass it as input to `fn`, as it will just be ignored in that case anyway.

I might have missed something, but I think each of these `_with_order_manipulation` decorators can be removed, and we can instead generalise the original versions. Let me know if I missed something in my reasoning :)

djl11 (changes requested):
See comment, which also applies to other decorators as well I believe, let me know if there's something I missed! :) Thanks!

ivy/functional/frontends/jax/func_wrapper.py
fspyridakos: Fixed it!

djl11 (approved):
It's possible there are other refactors that could be done, but this looks in good shape and we can refactor later if needed (there are also downsides to refactoring things too early). Feel free to merge! Thanks!

PR #7273, opened 2022-11-23
Added cupy to frontend and testing 

vaatsalya123:
This is an initial start to the cupy frontend.

https://docs.cupy.dev/en/stable/index.html


djl11:
Before deciding whether to officially include CuPy in the master branch, outside of the experimental folder, I have a couple of questions:

a) do you what percentage of tests are passing for Ivy's API in CuPy? Does it add a lot more work for us to get all of the unit tests fixed?

b) How much does this deviate from NumPy? Is it 99% the same, 80% the same etc.?

I'm wary of adding another officially supported frontend and backend before we get the current release out properly. Maybe we could flag that CuPy is an experimentally supported framework, perhaps even by adding a new sub-folder `experimental` in the frontend and backend folders to make it clear that this won't be officially supported in the next release. What do you think @vaatsalya123?

vaatsalya123:
The answer to (a) and (b) are very closely related to each other.The idea of cupy is to just unify cuda technology with Scipy and numpy for high performance computing and basically cupy just builds on top of numpy and scipy so basically cupy is just 100% numpy in interface with extensions involved inside it.Most tests are passing but a few are still failing because there is still stuff that is needed to be added.

Also cupy does not require much work the reason to that is because hypothetically let's say some test is failing in cupy but passing in numpy then we can just use cupy._numpy to enter into numpy's interface and fix those tests by using numpy interface.

A good example of the above statement is what I did here: ivy_tests/test_ivy/test_frontends/test_cupy/__init__.py

Also yes I agree currently cupy is not master ready but soon will be.The reason why I want to add cupy is because it will help in the future exploration we are doing for cuda graph's and compilations in C/C++ as also stated by Jēan as well.

Currently I think adding it to some experimental folder is for the best but yeah soon this would be master ready for the next release (:

djl11 (changes requested):
see comment :)

ivy-seed:

ivy-seed:

PR #7352, opened 2022-11-25
Changes to allow testing with only a subset of frameworks installed 

JoelPendleton:


ivy_tests/test_ivy/helpers/testing_helpers.py
```diff
@@ -160,7 +161,7 @@ def _get_supported_devices_dtypes(fn_name: str, fn_module: str):
 
 
 def handle_test(
-    *, fn_tree: str, ground_truth_backend: str = "tensorflow", **_given_kwargs
+    *, fn_tree: str, ground_truth_backend: str = "numpy", **_given_kwargs
```
Ishticode, line 164: I see that the default ground truth has been reset to numpy. This willl require a reconsideration for tests with `test_gradients=True` which may need to have to manually set tensorflow/jax/torch for their `test_function` call. This will lead to a dependency on tensorflow/jax etc in the [following line](https://github.com/unifyai/ivy/blob/8010806d72537c94b1b88eab5530739c80c4ecbc/ivy_tests/test_ivy/helpers/function_testing.py#L267) in test_function unless there is way to bypass such tests. Have you tried running any such tests. You can try `test_acosh` which will should fail on this branch for examples where gradient test is run.

Same comments can be made for the update in `handle_method` although it had some other updates that Mustafa was working on I suppose.

ivy_tests/test_ivy/test_functional/test_core/test_general.py
```diff
@@ -133,7 +158,6 @@ def test_array_equal(
     backend_fw,
     fn_name,
     on_device,
-    ground_truth_backend,
```
Ishticode, line 136: As `ground_truth_backend` is a required arg for `test_function` it has to be passed in here and also in the test_function call below. Running these tests throw an error.

Ishticode (changes requested):
Hi. Mostly looking good. Two Comments I made which require attention. Feel free to correct me if I misunderstand anything. 🙂Thanks Joel.

ivy_tests/test_ivy/helpers/testing_helpers.py
JoelPendleton: Ok I'll revert back to using to setting the `ground_truth_backend` to tensorflow and then leave it at the discretion of the developer to install tensorflow if they wish to use these tests, as it will be clear from the error received when running the test.

Ishticode: Can we instead not have a script which is run at import time to scan the available packages. If tensorflow/jax/torch is not found in the environment then gt should be set to numpy for all tests globally with a warning raised that gradient tests are unreliable with current environment?🙂

JoelPendleton: Yeah thank you, we can quite easily actually; I've implemented that in my latest commit. If the user has tensorflow installed, it first uses that. If they don't have tensorflow and they have torch, then it is set to torch. If they have only jax and numpy, it is set to jax. Otherwise it is set to numpy. Everything should work now 🙂 Let me know what you think.

JoelPendleton (commented):
Implemented changes.

Ishticode (approved):
This seems to working fine now. One point...

In the case where `available_frameworks` is just numpy, might be worth raising a warning with `logging.warning` that the environment only has numpy and and the tests run for for gradients are unreliable so as to avoid trying to debug when there is nothing be debug.

Feel free to do this and then merge or request review from Dan or merge.

PR #7365, opened 2022-11-25
Updated get_all_arrays_in_memory()

### LINKED ISSUE #6625: get_all_arrays_in_memory

92jpascal:
Close #6625


Nightcrab:
Is it necessary for the commit to do 5K line replacements? It looks like this change is only regarding ``get_all_arrays_in_memory`` and the commit should reflect that.

PR #7369, opened 2022-11-26
Made changes to handle wrappers for mixed functions

VedPatwardhan:
Hi, so the objective of this PR is to change the wrappings based on whether we're dealing with a compositional implementation of a backend-specific implementation. This is particularly useful for mixed functions where we have a compositional implementation for a few backends and backend-specific implementations for few others, where we want the function to defer to the backend-specific implementations when those backends are set and to the compositional implementation otherwise. Me and @fspyridakos have been working on this.
Now, the current way of wrapping functions on backend setting involves applying all the wrappings on the ivy function to the function that's wrapped, but there are some examples where this wouldn't work. 
For e.g. 
1. The compositional implementation may not require wrappers such as:`handle_out_argument`, `handle_nestable` while the backend-specific implementation of the same function would do so. 
2. The compositional implementation of the function would require wrappers such as `inputs_to_ivy_arrays`, but the backend-specific implementation would instead use `to_native_arrays_and_back`.

I have made some changes to `set_backend` and `_wrap_function` for doing this, but wasn't sure if this the cleanest and efficient way of doing so, which is why I wanted your suggestion @djl11 on what you think could be improved. Thanks 🙂


ivy/backend_handler.py
```diff
@@ -259,12 +259,16 @@ def set_backend(backend: str):
     backend_stack.append(backend)
     set_backend_to_specific_version(backend)
     for k, v in ivy_original_dict.items():
+        compositional = False
```
djl11, line 262: this would remove a LOC:
```
if compositional:
```

djl11 (changes requested):
Tiny change, otherwise looks good. Feel free to merge afterwards!

VedPatwardhan:
Done with the change, merging now 👍

PR #7375, opened 2022-11-26
Removing repeated words

rohini-ranjanR:
Removing repeated words from docs (https://lets-unify.ai/ivy/background/standardization.html)


RickSanchezStoic:
looks good, thanks!

PR #7399, opened 2022-11-27
added jax.numpy.linalg.tensorinv with tests

### LINKED ISSUE #7387: tensorinv

mmi333:
Close #7387

Added `jax.numpy.linalg.tensorinv` frontend function through copying [the original implementation](https://github.com/google/jax/blob/71edfc7422eeff1ec91cff6711a91685270bb192/jax/_src/third_party/numpy/linalg.py#L65) and replacing the appropriate functions with ivy functions.

Added `test_jax_numpy_tensorinv`  and  `_get_inv_square_matrices` in test_jax_numpy_linalg.py by closely following the test in [the original implementation](https://github.com/google/jax/blob/c1e1d64e66a7850281420bf029b025ec79de5681/tests/linalg_test.py#L765)

Notes:

- We needed to implement a helper function `_get_inv_square_matrices` because `tensorinv` requires invertible matrices that have a "square" shape as per [the documentation](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.tensorinv.html). Namely, 

> a (array_like) – Tensor to ‘invert’. Its shape must be ‘square’, i. e., prod(a.shape[:ind]) == prod(a.shape[ind:]).

Things I'm not sure about:

- Whether `input_dtype` needs to be filtered to exclude `float16` and `bfloat16` dtypes.
- Whether  `while not invertible` is the best way to ensure the generated matrix is non-invertible (this is how it's implemented in the original testing code)

### DIFF
--- a/ivy/functional/frontends/jax/numpy/linalg.py
+++ b/ivy/functional/frontends/jax/numpy/linalg.py
@@ -115,3 +115,19 @@ def tensorsolve(a, b, axes=None):
     res = ivy.solve(a, b)
     res = ivy.reshape(res, shape=ret_shape)
     return res
+
+
+@to_ivy_arrays_and_back
+def tensorinv(a, ind=2):
+    old_shape = ivy.shape(a)
+    prod = 1
+    if ind > 0:
+        invshape = old_shape[ind:] + old_shape[:ind]
+        for k in old_shape[ind:]:
+            prod *= k
+    else:
+        raise ValueError("Invalid ind argument.")
+    a = ivy.reshape(a, shape=(prod, -1))
+    ia = ivy.inv(a)
+    new_shape = tuple([*invshape])
+    return ivy.reshape(ia, shape=new_shape)
--- a/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_linalg.py
+++ b/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_linalg.py
@@ -745,3 +745,64 @@ def test_jax_numpy_pinv(
         fn_tree=fn_tree,
         a=x[0],
     )
+
+
+# tensorinv
+@st.composite
+def _get_inv_square_matrices(draw):
+
+    dim_size = draw(helpers.ints(min_value=1, max_value=10))
+
+    batch_shape = draw(st.sampled_from([2, 4, 6, 8, 10]))
+
+    shape = (dim_size,) * batch_shape
+    input_dtype = draw(helpers.get_dtypes("float", index=1, full=False))
+    invertible = False
+    while not invertible:
+        a = draw(
+            helpers.array_values(
+                dtype=input_dtype[0],
+                shape=shape,
+                min_value=-100,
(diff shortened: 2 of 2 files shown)


mmi333:
Hi @nassimberrada !

Apparently there was no problem with the square matrix generation part. The problem was rather with passing the matrix to `helpers.test_frontend_function` where I was passing `a=x[0]` because I was following the same structure as most of the other tests in the same file, but it turns out it should be `a=x` as the returned array from `_get_inv_square_matrices` is the matrix itself, and not a list containing the matrix as its first element.

I've also added a filter call to exclude float16 and bfloat16 data types as they're not currently supported by JAX.

I've taken a look at `square_matrix_shapes` as per your suggestion but it appears it only generates matrices with n_dims = 2 of shape (size, size). While `tensorinv`'s definition of square is more general like the samples in the test:

`  @jtu.sample_product(
    shape=[(1, 1), (4, 4), (200, 200), (7, 7, 7, 7)],
    dtype=float_types,
  )
`

While the current testing code takes into consideration cases like `(7, 7, 7, 7)`, it doesn't generate matrices with shape e.g `(8, 3, 4, 6)` which should also work because 8\*3 = 4\*6 = 24 which fits the definition of "square" in the function [docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.tensorinv.html).
I did not include it in the testing because I thought it would make the helper function more complicated as it would be hard to generate general cases like this and it's also not accounted for in the [original testing code.
](https://github.com/google/jax/blob/c1e1d64e66a7850281420bf029b025ec79de5681/tests/linalg_test.py#L765)

The CI tests still aren't passing, but by inspecting the logs it appears they're failing because of other functions unrelated to tensorinv. Would like to know your thoughts on this.

Thanks!

mmi333:
Thanks @nassimberrada !

I agree with your suggestion on adding handpicked examples to the tests to enable more general testing while avoiding more complex code. I have implemented that through generating examples like in the original library: `generated shape` and `generated ind` which always have the same dim_size repeated `batch_shape` times, with `ind` generated as in the library to be `int(np.floor(len(generated_shape) / 2))`. 

For the handpicked examples, I tried to come up with more general examples for the `shape` and `ind` parameters that follow the definition of 'square' in the [docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.tensorinv.html) e.g: `[(6, 7, 8, 16, 21), 3]` since 6\*7\*8 = 16\*21 demonstrates that `ind` doesn't have to be exactly in the middle of the array as long as the property `prod(a.shape[:ind]) == prod(a.shape[ind:])` is satisfied.

I have also added the `@with_unsupported_dtypes` decorator along with its import statement since no other frontend functions in the same file were using it.

I feel like this PR is complete now, although I'm open to any further suggestions.
Thanks!

nassimberrada:
Great work @mmi333 ! Your PR looks good to go i think we can wrap it up, i'll be merging at once, thank you for your contribution !

PR #7406, opened 2022-11-27
Small fix using De Morgan's law

vaatsalya123:
Using some functional theory and using De Morgan's law for data structure I think this format is better to use.


vaatsalya123:
This pr is sort of a start on the efficiency of the wrapping code (:

VedPatwardhan (approved):
lgtm!

PR #7412, opened 2022-11-28
Small fixes in func_wrapper to add more efficiency.

vaatsalya123:


ivy/func_wrapper.py
```diff
@@ -45,7 +45,6 @@ def _get_first_array(*args, **kwargs):
             arr = ivy.index_nest(kwargs, arr_idxs[0])
     return arr
 
-
```
VedPatwardhan, line 48: I think there needs to be a blank line here otherwise the lint tests may fail

VedPatwardhan (approved):
Just a small change, rest looks good to me 🙂

PR #7498, opened 2022-12-01
adding `inplace` and `mutable` for nest functions

sherry30:


ivy/functional/ivy/nest.py
```diff
@@ -163,12 +170,30 @@ def set_nest_at_index(
...
-        nest[index[0]] = ret
+        inner_tupled = False
+        if isinstance(_result[index[0]], tuple):
+            _result[index[0]] = list(_result[index[0]])
+            inner_tupled = True
+        set_nest_at_index(nest[index[0]], index[1:], value, shallow, _result[index[0]])
+        if inner_tupled:
+            _result[index[0]] = tuple(_result[index[0]])
+    if outer_tupled:
+        _result = tuple(_result)
```
guillesanbri, line 195: this would modify subclasses of tuple incorrectly, imo we should store `nest_type = type(nest)` instead of `outer_tupled` and simply do `_result = nest_type(_result)` without the `if`, given that if it's not a tuple subclass this won't affect the other classes, don't know if this makes sense

ivy/functional/ivy/nest.py
```diff
@@ -163,12 +170,30 @@ def set_nest_at_index(
...
-        ret = nest[index[0]]
-        ivy.set_nest_at_index(ret, index[1:], value)
-        nest[index[0]] = ret
+        inner_tupled = False
+        if isinstance(_result[index[0]], tuple):
+            _result[index[0]] = list(_result[index[0]])
+            inner_tupled = True
+        set_nest_at_index(nest[index[0]], index[1:], value, shallow, _result[index[0]])
+        if inner_tupled:
+            _result[index[0]] = tuple(_result[index[0]])
```
guillesanbri, line 193: can't these lines be replaced with something along the lines of `_result[index[0]] = set_nest_at_index(      nest[index[0]], index[1:], value, shallow, _result[index[0]])` ?

ivy/functional/ivy/nest.py
```diff
@@ -97,10 +97,12 @@ def prune_nest_at_index(nest: Iterable, index: Tuple, /):
...
 @handle_exceptions
 def set_nest_at_index(
-    nest: Union[ivy.Array, ivy.NativeArray, ivy.Container, Dict, List],
+    nest: Union[ivy.Array, ivy.NativeArray, ivy.Container, Dict, List, Tuple],
     index: Sequence[Union[str, int]],
     value: Any,
     /,
+    shallow: bool = True,
+    _result: Union[ivy.Array, ivy.NativeArray, ivy.Container, Dict, List, Tuple] = None,
 ):
```
guillesanbri, line 106: missing return annotation

ivy/functional/ivy/nest.py
```diff
@@ -189,6 +214,8 @@ def map_nest_at_index(
     index: Sequence[Union[str, int]],
     fn: Callable[[Any], Any],
     /,
+    shallow: bool = True,
+    _result: Union[ivy.Array, ivy.NativeArray, ivy.Container, Dict, List] = None,
 ) -> None:
```
guillesanbri, line 219: annotation and docstring of return

ivy/functional/ivy/nest.py
```diff
@@ -252,10 +284,31 @@ def map_nest_at_index(
...
+        inner_tuppled = False
+        if isinstance(_result[index[0]], tuple):
+            _result[index[0]] = list(_result[index[0]])
+            inner_tuppled = True
+        map_nest_at_index(nest[index[0]], index[1:], fn, shallow, _result[index[0]])
+        if inner_tuppled:
+            _result[index[0]] = tuple(_result[index[0]])
+    if outer_tupled:
+        _result = tuple(_result)
+    return _result
```
guillesanbri, line 311: I'd say a lot of the reviews for the first function make sense here as well and can enhance the code a little bit

ivy/functional/ivy/nest.py
```diff
@@ -345,9 +402,33 @@ def set_nest_at_indices(
     >>> print(nest)
     ivy.array([[1., 11., 3.], [4., 5., 22.]])
     """
+    if shallow:
+        result = type(nest)(nest)
+    else:
+        result = copy_nest(nest)
```
guillesanbri, line 408: same as above, we could replace this with just `result = copy_nest(nest)`

ivy/functional/ivy/nest.py
```diff
@@ -345,9 +402,33 @@ def set_nest_at_indices(
...
+    outer_tupled = False
```
guillesanbri, line 409: again, I think we should replace this with `nest_type=type(nest)`

ivy/functional/ivy/nest.py
```diff
@@ -345,9 +402,33 @@ def set_nest_at_indices(
...
+    if isinstance(nest, tuple):
+        outer_tupled = True
```
guillesanbri, line 411: no need for `outer_tuple` imo

ivy/functional/ivy/nest.py
```diff
@@ -345,9 +402,33 @@ def set_nest_at_indices(
...
+        if shallow:
+            result = set_nest_at_index(
+                nest, index, value, _result=result, shallow=shallow
+            )
+        else:
+            result = set_nest_at_index(
+                result, index, value, _result=result, shallow=shallow
+            )
+    if outer_tupled:
+        result = tuple(result)
```
guillesanbri, line 430: replace these two lines with `result = nest_type(result)`

ivy/functional/ivy/nest.py
```diff
@@ -345,9 +402,33 @@ def set_nest_at_indices(
...
+            )
+            continue
+        if shallow:
+            result = set_nest_at_index(
+                nest, index, value, _result=result, shallow=shallow
+            )
+        else:
+            result = set_nest_at_index(
+                result, index, value, _result=result, shallow=shallow
+            )
```
guillesanbri, line 428: maybe I'm missing something, but can't we replace this block with just `result = set_nest_at_index(nest, index, value, _result=result, shallow=shallow)`? with the try except block in the `set_nest_at_index` it should behave the same way I think

ivy/functional/ivy/nest.py
```diff
@@ -424,7 +514,27 @@ def map_nest_at_indices(nest: Iterable, indices: Tuple, fn: Callable, /):
...
+            continue
+        if shallow:
+            result = map_nest_at_index(nest, index, fn, _result=result, shallow=shallow)
+        else:
+            result = map_nest_at_index(
+                result, index, fn, _result=result, shallow=shallow
+            )
+    if outer_tupled:
+        result = tuple(result)
+    return result
```
guillesanbri, line 537: prob a lot of the changes above apply as well here

ivy_tests/test_ivy/test_functional/test_core/test_nest.py
```diff
@@ -402,12 +416,17 @@ def indices_nest(nest, indices):
 # nested_map
 @pytest.mark.parametrize("x", [{"a": [[0, 1], [2, 3]], "b": {"c": [[0], [1]]}}])
 @pytest.mark.parametrize("fn", [lambda x: x**2])
-def test_nested_map(x, fn):
+@pytest.mark.parametrize("inplace", [True, False])
```
guillesanbri, line 419: prob worth changing the name here as well :smile:

guillesanbri (changes requested):
I've reviewed `set_nest_at_index` and `set_nest_at_indices` more extensively, I'd say that most of the changes can be applied to the other functions. Docstring examples should be updated as well I guess. Let me know if I missed anything and obviously totally open to discuss and/or try to clarify anything :smile: thanks!

ivy/functional/ivy/nest.py
```diff
@@ -345,9 +397,26 @@ def set_nest_at_indices(
...
-    [set_nest_at_index(nest, index, value) for index, value in zip(indices, values)]
+    for index, value in zip(indices, values):
+        if shallow:
+            result = set_nest_at_index(
+                nest, index, value, _result=result, shallow=shallow
+            )
+        else:
+            result = set_nest_at_index(
+                result, index, value, _result=result, shallow=shallow
+            )
```
sherry30, line 416: If we do that then `set_nest_at_index` won't receive updated nest in every loop in case of `shallow=False`.
`result = set_nest_at_index(nest, index, value, _result=result, shallow=shallow)` will only end up updating on the last index provided. Thats why I'm passing `result` as input in case of  `shallow=False`

sherry30 (commented):
I made changes to `set_nest_at_index` and `set_nest_at_indices` and added a few comments.
I'll make similar changes to the other function once we resolve this one.

ivy/functional/ivy/nest.py
```diff
@@ -163,12 +170,30 @@ def set_nest_at_index(
         b: ivy.array([3., 4.])
     }
     """
+    outer_tupled = False
+    if _result is None:
+        if shallow:
+            _result = type(nest)(nest)
+        else:
+            _result = copy_nest(nest)
```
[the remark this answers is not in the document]
sherry30: Weren't we supposed to keep the iterables ids the same wherever we can for `shallow=True`?
If we we do `copy_nest` in both cases then the unexplored nests will have different ids.

ivy/functional/ivy/nest.py
sherry30: If we do that then `set_nest_at_index` won't receive updated nest in every loop in case of `shallow=False`.
`result = set_nest_at_index(nest, index, value, _result=result, shallow=shallow)` will only end up updating on the last index provided. Thats why I'm passing `result` as input in case of  `shallow=False`

ivy/functional/ivy/nest.py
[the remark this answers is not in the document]
guillesanbri: given that probably when `shallow=True` people are going to be looking at the input anyway I don't think this is a problem, if the result has different ids for the nests then I guess that's fine (as long as the leaves do keep their ids, that's why we couldn't do the deepcopy)

ivy/functional/ivy/nest.py
guillesanbri: resolving this as discussed in discord

ivy/functional/ivy/nest.py
```diff
@@ -163,12 +175,26 @@ def set_nest_at_index(
...
+    _result = list(_result) if isinstance(nest, tuple) else _result
     if len(index) == 1:
-        nest[index[0]] = value
+        if shallow:
+            try:
+                nest[index[0]] = value
+            except TypeError:
+                print("shallow cannot be set to True for immutable types")
```
guillesanbri, line 190: maybe this ends up being a little too verbose but we can change it afterwards anyway

guillesanbri (commented):
Everything looks good so far, if I've missed something let me know, thanks!

guillesanbri (approved):
Looks good to me! one thing left to do I'd say is to update the docstring examples but we can leave that as a separate task I guess, thanks a lot!

PR #7514, opened 2022-12-01
add tf searchsorted

### LINKED ISSUE #7493: searchsorted

Karthick47v2:
Close #7493

### DIFF
--- a/ivy/functional/frontends/tensorflow/general_functions.py
+++ b/ivy/functional/frontends/tensorflow/general_functions.py
@@ -111,3 +111,10 @@ def sort(values, axis=-1, direction="ASCENDING", name=None):
             message="Argument `direction` should be one of 'ASCENDING' or 'DESCENDING'",
         )
     return ivy.sort(values, axis=axis, descending=descending)
+
+
+@to_ivy_arrays_and_back
+def searchsorted(sorted_sequence, values, side="left", out_type="int32"):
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -111,3 +111,10 @@ def sort(values, axis=-1, direction="ASCENDING", name=None):
             message="Argument `direction` should be one of 'ASCENDING' or 'DESCENDING'",
         )
     return ivy.sort(values, axis=axis, descending=descending)
+
+
+@to_ivy_arrays_and_back
+def searchsorted(sorted_sequence, values, side="left", out_type="int32"):
+    if out_type not in ["int32", "int64"]:
```
DragosStoican, line 118: I am not sure if this if statement is required. Can you explain why you think it is needed?

Karthick47v2: 1. Tensorflow's [doc](https://www.tensorflow.org/api_docs/python/tf/searchsorted) says 'out_type' should be `int32` or `int64`.
2. Functions having return type as argument used the same convention, So, I sticked to it. (shape function on same file, functions on [math.py](https://github.com/unifyai/ivy/blob/master/ivy/functional/frontends/tensorflow/math.py))

PR #7518, opened 2022-12-01
Added properties for ivy Dtypes

fspyridakos:
I added properties relating to dtype methods we already have implemented in Ivy, I plan to handle any frontend specific dtype attributes in frontend Dtype classes.


djl11 (approved):
lgtm!

PR #7521, opened 2022-12-01
added mean method for jax frontend

SaifAshraf70:


SaifAshraf70:
3 tests passed but 1 failed 
I don't know why

PR #7533, opened 2022-12-02
optimise promote_types_of_inputs and handle int/float inputs accurately

Ishticode:
Currently the dtype function promotes `Number` (non-array) entries to their counterpart array. This can sometime leade to down promote like if the array is of `int` type and number of `float`. For example take `ivy.pow` function.
This code produces the wrong result.
```python
>>> x = ivy.array([2,3])
>>> ivy.pow(x, 2.3)
ivy.array([4, 9])
```
Whereas torch gives the mathematically accurate result.

```python
>>> x = torch.tensor([2,3])
>>> torch.pow(x, 2.3)
tensor([ 4.9246, 12.5135])
```

After this change the result will be
```python
>>> x = ivy.array([2,3])
>>> ivy.pow(x, 2.3)
ivy.array([ 4.92457771, 12.51350212])
```

Also the function had some extra needless calls to ivy.asarray and duplicated logic which was removed. Just needs approval before merge as its quite a funcdamental function.

### DIFF
--- a/ivy/functional/ivy/data_type.py
+++ b/ivy/functional/ivy/data_type.py
@@ -2007,25 +2007,19 @@ def promote_types_of_inputs(
     as inputs only for those functions that expect an array-like or tensor-like objects,
     otherwise it might give unexpected results.
     """
-    if hasattr(x1, "dtype") and hasattr(x2, "dtype"):
-        if x1.dtype != x2.dtype:
-            promoted = promote_types(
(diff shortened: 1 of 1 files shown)


VedPatwardhan (approved):
Hi @Ishticode, the changes look good to me and the CI doesn't seem to show large scale failures because of this, let's get the PR merged and then keep an eye on the dashboard. Thanks 🙂

PR #7539, opened 2022-12-02
added conv_transpose for nn tensorflow frontends

anurag629:
conv_transpose #7538


anurag629:
Hi @guillesanbri  Please review this PR.    Thanks!

guillesanbri:
Hi! Thank you for the contribution! seems like the function has been implemented calling `ivy.conv_transpose` which doesn't exist, you should use the existing functions or propose new ones as stated in the [docs](https://lets-unify.ai/ivy/contributing/open_tasks.html#frontend-apis) :slightly_smiling_face: happy to answer any questions!

anurag629:
> Hi! Thank you for the contribution! seems like the function has been implemented calling `ivy.conv_transpose` which doesn't exist, you should use the existing functions or propose new ones as stated in the [docs](https://lets-unify.ai/ivy/contributing/open_tasks.html#frontend-apis) :slightly_smiling_face: happy to answer any questions!

I have to add conv_transpose to backend  first then in frontend.

anurag629:
@guillesanbri  Should I write the backend function for all four i.e., numpy, jax, tensorflow, and torch or in tensorflow only.

Thanks!

guillesanbri:
I've been reading the discord chat and I guess that we are leaving this issue open right? if so, I can link the other issue :slightly_smiling_face:

anurag629:
@guillesanbri similar function is available in ivy <code>conv_genral_transpose</code> so, i used that. 
Review it again

guillesanbri:
Hi! Thank you very much for contributing to Ivy, I'll take a look at this now :slightly_smiling_face:

guillesanbri:
Looks like the test is not currently passing in the CI, is this something you have looked into?

anurag629:
> Looks like the test is not currently passing in the CI, is this something you have looked into?

@guillesanbri, I looked and tried to solve but I did not understand how to fix the error.

test_tensorflow_conv_transpose[cpu-ivy.functional.backends.tensorflow-False-False] _

    @handle_frontend_test(
>       fn_tree="tensorflow.nn.conv_transpose",
        x_f_d_df=_x_and_filters(
            dtypes=helpers.get_dtypes("float", full=False),
            data_format=None,
            padding=st.sampled_from(["VALID", "SAME"]),
            stride_min=3,
            stride_max=4,
            dilation_max=1,
            transpose=True,
        ),
    )

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_nn.py:926:

guillesanbri:
Hi, seems like there is a problem now with the `data_format` param of the test, don't know if you are already looking into this, sorry if that's the case! :slightly_smiling_face:

guillesanbri:
Wild guess, but I'd say it's probably related to the number of chars in the string being related to the number of spatial dimensions (as per the tf docs) and not fixed as in `data_format=st.sampled_from(["NWC", "NCW"]),`

anurag629:
Hi @guillesanbri Please review it again.

anurag629:
@guillesanbri

ivy/functional/frontends/tensorflow/nn.py
```diff
@@ -228,6 +228,27 @@ def local_response_normalization(
 
 
 @to_ivy_arrays_and_back
```
guillesanbri, line 230: seems like this line belongs to the function below :sweat_smile:

guillesanbri (changes requested):
I think that the problem is still there, you probably have to add some logic so that the `data_format` is sampled from the appropriate collection of possible values based on the number of dimensions

anurag629:
> I think that the problem is still there, you probably have to add some logic so that the `data_format` is sampled from the appropriate collection of possible values based on the number of dimensions

Can you provide some hints on how to write logic as I am not able to catch it.

guillesanbri:
I can't give an example off the top of my head :sweat_smile: , but you probably have to write a custom strategy (https://hypothesis.readthedocs.io/en/latest/data.html#composite-strategies) and use the output of `dtype_and_values` (the number of dimensions) to correctly get the set of options you are going to be sampling from, hope this makes sense :slightly_smiling_face:

anurag629:
hi @guillesanbri Now, it seems everything is fine. please merge it.

guillesanbri:
Hi, maybe I'm missing something but the code is the same as in the last review, you should add a composite strategy to deal with the test parameters. Also, as signalled in the code snippet above, you have removed the `@to_ivy_arrays_and_back` from the function below, which should also be added :slightly_smiling_face:

anurag629:
@guillesanbri my two tests are passing but I am not able to resolve the other. Can you help me with this?
Screenshot from 2023-01-14 03-36-09
Screenshot from 2023-01-14 03-35-52

guillesanbri:
can you try with `@to_ivy_arrays_and_back` added on top of the function? not sure if this is the problem but should be there anyway so worth trying :slightly_smiling_face:

anurag629:
> can you try with `@to_ivy_arrays_and_back` added on top of the function? not sure if this is the problem but should be there anyway so worth trying slightly_smiling_face

@guillesanbri  nothing changed after adding `@to_ivy_arrays_and_back`. Do you have any other suggestions?

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_nn.py
```diff
@@ -921,6 +921,57 @@ def test_tensorflow_max_pool1d(
...
+        fn_tree=fn_tree,
+        on_device=on_device,
+        input=x,
+        filters=filters,
+        output_shape=output_shape,
+        strides=stride,
+        padding=pad,
+        data_format=data_format,
+        dilations=dilations,
+    )
```
guillesanbri, line 972: Frontend tests have been refactored, the format for the test should be something along the lines of

```
def test_tensorflow_conv_transpose(
    *,
    x_f_d_df,
    frontend,
    test_flags,
    fn_tree,
    on_device,
):
    (
        input_dtype,
        x,
        filters,
        dilations,
        data_format,
        stride,
        pad,
        output_shape,
    ) = x_f_d_df
        input_dtypes=input_dtype,
        frontend=frontend,
        test_flags=test_flags,
        fn_tree=fn_tree,
        on_device=on_device,
        atol=1e-1,
        input=x,
        filters=filters,
        output_shape=output_shape,
        strides=stride,
        padding=pad,
        data_format=data_format,
        dilations=dilations,
    )
```

guillesanbri (changes requested):
I left a quick heads up about the test. Regarding the error on the `bfloat` dtype, I think it's weird, the way I understand it, the ground truth is the value that is wrong. 

Could you try running the function in tensorflow with the falsifying input and see if the output dtype is `float32`? If that not the case it may be an error in the tests, but if it is, the error is in the tensorflow side, as it's not keeping the input dtype in the output :thinking:

PR #7551, opened 2022-12-03
Implemented FE for pinv in torch.

### LINKED ISSUE #7550: pinv

WilliamHirst:
Close #7550  
Needed to implement special case for atol given that torch is the only backend that supports this argument.


modanesh:
The frontend tests are failing for `test_tensorflow_pinv`. Could you look into it?

PR #7554, opened 2022-12-03
feat: Add linspace examples

### LINKED ISSUE #7553: linspace

meetcshah19:
Close #7553

### DIFF
--- a/ivy/functional/ivy/creation.py
+++ b/ivy/functional/ivy/creation.py
@@ -1048,6 +1048,42 @@ def linspace(
     but this function is *nestable*, and therefore also accepts :class:`ivy.Container`
     instances in place of any of the arguments.
 
+    Functional Examples
+    -------------------
+
+    With float input:
+ 
+    >>> ivy.linspace(1, 2, 4)
+    ivy.array([1., 1.33333337, 1.66666663, 2.])
+
+    >>> ivy.linspace(1, 2, 4, endpoint=False)
+    ivy.array([1., 1.25, 1.5 , 1.75])
+
+    >>> ivy.linspace(1, 10, 4, dtype = int)
+    ivy.array([ 1,  4,  7, 10])
+
+    >>> ivy.linspace(1, 2, 4, device = "cuda:0")
+    ivy.array([1., 1.33333337, 1.66666663, 2.])
+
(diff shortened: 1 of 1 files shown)


nassimberrada:
Hi @meetcshah19, thank you for your contribution ! 
The functional examples look good to me but could you add container examples as well ? Thanks

meetcshah19:
Sure, on it!

meetcshah19:
@nassimberrada are you sure linspace() can actually be used with Containers is a way that would be logically required/ plausible? If yes, can you give me an example of such a case? To me applying space on a nested array structure doesn't make a lot of sense. Thanks!

nassimberrada:
Also, i just saw that a new PR overlaps with yours, i think this was caused by improper naming, could you rename your PR with just the function name ? Feel free to follow the same logic for future PRs as well to avoid any confusion.

meetcshah19:
Thanks for the explanation @nassimberrada, I did mean a real life example. Also, I get the mixup with the issue tracking and I believe the solution you suggested is ideal. I will write the array instance method example.

ivy/functional/ivy/creation.py
```diff
@@ -1076,14 +1076,24 @@ def linspace(
...
-    ivy.array([[1, 2, 3, 4],[2, 3, 4, 5]])
+    ivy.array([[1, 2, 3, 4], [2, 3, 4, 5]])
 
     >>> ivy.linspace(x, y, 4, axis = 0)
     ivy.array([[1, 2],
                [2, 3],
                [3, 4],
                [4, 5]])
 
+    Using array instance method
```
nassimberrada, line 1087: Could you add/change just the following line(s):
```
    Instance Method Examples
    -------------------
    With :class:`ivy.Array` input:
```

nassimberrada (changes requested):

ivy/functional/ivy/creation.py
meetcshah19: Done

nassimberrada:
There does seem to be an issue with the docstrings tests since your example output is correct. 
Fixing the testing process should make your examples pass but since that is not your responsibility here, i can merge the PR as it looks good to me. 
Thank you for your patience, and your contribution to Ivy !

PR #7591, opened 2022-12-05
a little fix to the get_axis

Nunyapa:
Sometimes hypothesis generates a particular value many times in a row and due to it in the while loop (inside the if statement) we are waiting for a new axis, but actually the axis array doesn't change, hypothesis raises FailedHealthyCheck exception.
In hypothesis documentation (https://hypothesis.readthedocs.io/en/latest/data.html#composite-strategies) they recommend to use the filter method to avoid that situation of getting identical values in a row.

### DIFF
--- a/ivy_tests/test_ivy/helpers/hypothesis_helpers/general_helpers.py
+++ b/ivy_tests/test_ivy/helpers/hypothesis_helpers/general_helpers.py
@@ -383,8 +383,12 @@ def get_axis(
 
     axis = draw(st.one_of(*valid_strategies))
     if unique and allow_neg and not isinstance(axis, int):
+        _prev_axis = []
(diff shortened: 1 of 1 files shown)


AnnaTz:
Actually I just pushed a fix https://github.com/unifyai/ivy/commit/6018bf2cc80d29b1d52b9949b1fb2715cc748a7d. I addressed both the infinite loop problem and another small bug I found. Of course @Ishticode feel free to check to verify it worked correctly now.

Nunyapa:
So, I can close the PR?

AnnaTz:
> So, I can close the PR?

Yeah I guess there is no point in keeping this open since the suggested change has already been merged via my commit.

PR #7594, opened 2022-12-05
support for tuple of tensors inside test_function

zaeemansari70:


sherry30 (changes requested):
1. You have to remove assertion on this line [here](https://github.com/unifyai/ivy/blob/bce2ca865a672fea24a5b5c61105028b1a7f1938/ivy_tests/test_ivy/helpers/function_testing.py#L294)
since `test_ret` will be a list in this case.
2. There is a problem with `handle_out_argument` decorator when `handle_out_in_backend` is False

ivy-seed:

PR #7596, opened 2022-12-05
Added support for python data classes

fspyridakos:
Implemented conversion of python data classes to the default ivy/native dtype


sherry30 (approved):
lgtm 👍

PR #7668, opened 2022-12-07
updated eye for function formatting changes.

ra9hur:
The following files have been checked:

ivy/ivy/functional/ivy/creation.py

ivy/ivy/container/creation.py

ivy/ivy/functional/backends/jax/creation.py
ivy/ivy/functional/backends/numpy/creation.py
ivy/ivy/functional/backends/tensorflow/creation.py
ivy/ivy/functional/backends/torch/creation.py

ivy/ivy_tests/test_ivy/test_functional/test_core/test_creation.py
ivy/ivy_tests/test_array_api/array-api/spec/API_specification/signatures/creation_functions.py

Here is the checklist:
All checks to be in √, >>, OK

1	OK
remove all lambda and direct bindings for the backend functions (in ivy.functional.backends), with each function instead defined using def.

2	OK
implement the following if they don’t exist but should do:
ivy.Array instance method,
ivy.Container static method,
ivy.Container instance method,
ivy.Array special method,
ivy.Array reverse special method,
ivy.Container special method,
Ivy.Container reverse special method.

Comment:
This check is mostly not applicable to eye function since it does not accept ivy.array as input.

Container static method is available at the location:
ivy/ivy/container/creation.py

3   >>
Make sure that the aforementioned methods are added into the correct category-specific parent class, such as ivy.ArrayWithElementwise, ivy.ContainerWithManipulation etc.

Comments:
Conforms to array API standards and correctly placed within "creation_functions" category.
eye(n_rows: int, n_cols: Optional[int] = None, /, *, k: int = 0, dtype: Optional[dtype] = None, device: Optional[device] = None) → array

Otherwise, this check is mostly not applicable to eye function since it does not accept ivy.array as input.

4	OK
correct all of the Function Arguments and the type hints for every function and its relevant methods, including those you did not implement yourself.

5   √
add the correct Docstrings to every function and its relevant methods, including those you did not implement yourself.

Comment:
Description for "batch_shape" has been added to the docstring.

6   √
add thorough Docstring Examples for every function and its relevant methods and ensure they pass the docstring tests.

Comment:
Docstring Examples were missing. This has been included in:
ivy/ivy/functional/ivy/creation.py


ra9hur:
eye #7616

ra9hur:
Close #7616

ra9hur:
Hi @Daniel4078 Please review this PR. Thanks!

Daniel4078 (changes requested):
I found something wrong with your docstring examples, check the comment before. Thank you.

ra9hur:
@Daniel4078, thanks for the reviews!!. It was very naive of me to not test examples on the python prompt. Now, it is done (in docker) and have included the ones that work.

ra9hur:
Have a question on 'out' parameter.
In the array API standards, 'out' is not a defined parameter.

https://data-apis.org/array-api/latest/API_specification/generated/signatures.creation_functions.eye.html

eye(n_rows: int, n_cols: Optional[int] = None, /, *, k: int = 0, dtype: Optional[dtype] = None, device: Optional[device] = None) → array

When I run on the python prompt with 'out' parameter, expect to get below result.

    Expected:
    >>> a1 = [1., 1., 1.]
    >>> x1 = ivy.eye(3, out=a1)
    >>> print(x1)
    ivy.array([1., 0., 0.])


However, get below results.
    
    >>> a1 = ivy.ones(3)
    >>> x1 = ivy.eye(3, out=a1)
    >>> print(x1)
    ivy.array([[1., 0., 0.],
        [0., 1., 0.],
        [0., 0., 1.]])

    >>> a1 = [1, 1, 1]
    >>> x1 = ivy.eye(3, out=a1)
    >>> x1
    ivy.array([[1., 0., 0.],
        [0., 1., 0.],
        [0., 0., 1.]])


The question is, 
- should this be treated as a bug ?
OR
- 'out' should not be supported and corresponding change is required ?

ra9hur:
Including method signatures of supported backends. 'out' parameter is supported only in torch.eye

tf.eye(
    num_rows,
    num_columns=None,
    batch_shape=None,
    dtype=tf.dtypes.float32,
    name=None
)

numpy.eye(N, M=None, k=0, dtype=<class 'float'>, order='C', *, like=None)

torch.eye(n, m=None, *, out=None, dtype=None, layout=torch.strided, device=None, requires_grad=False) → Tensor

jax.numpy.eye(N, M=None, k=0, dtype=None)

ivy/functional/ivy/creation.py
```diff
@@ -974,7 +977,82 @@ def eye(
 
     Both the description and the type hints above assumes an array input for simplicity,
     but this function is *nestable*, and therefore also accepts :class:`ivy.Container`
-    instances in place of any of the arguments.
+    instances in place instead of the arguments.
```
Daniel4078, line 980: This line should be left in its origenal form as it should mean that the datatype passed into any of the arguments can also be containers

ivy/functional/ivy/creation.py
```diff
@@ -974,7 +977,82 @@ def eye(
...
+    >>> x1.shape
+    (3, 2, 3)
+    # Suppose batch_shape = [a, b] then the returning identity array shape is [a, b, numRows, numColumns]
+
+
+    With :'out' input:
+
+    >>> a1 = ivy.ones(3)
+    >>> x1 = ivy.eye(3, out=a1)
+    >>> print(x1)
```
Daniel4078, line 1043: as I said before, this is not how out will be used.

Daniel4078 (changes requested):
Some other small errors for you to fix. Thank you!

ra9hur:
Thanks @Daniel4078, that clarified a lot of how 'out' is intended to be used.

This should be trivial, but I am still missing something.

I quickly ran a check based on your suggestions.

>>> a1 = ivy.ones(3)
>>> a1
ivy.array([1., 1., 1.])

>>> ivy.eye(3, out=a1)
ivy.array([[1., 0., 0.],
       [0., 1., 0.],
       [0., 0., 1.]])

>>> a1
ivy.array([[1., 0., 0.],
       [0., 1., 0.],
       [0., 0., 1.]])

- a1 is initialized with shape (1, 3).    ivy.array([1., 1., 1.])
- Once it is updated in place through eye function, I assume it should retain its shape as [1, 3] and the output should be  ivy.array([1., 0., 0.])
- However, a1 shape changes to (3, 3)
ivy.array([[1., 0., 0.],
       [0., 1., 0.],
       [0., 0., 1.]])
- We are saving memory  by not initializing a new variable. However, shouldn't  'a1'  retain its shape (1, 3) ?

Can you please clarify?

Daniel4078:
> Thanks @Daniel4078, that clarified a lot of how 'out' is intended to be used.
> 
> This should be trivial, but I am still missing something.
> 
> I quickly ran a check based on your suggestions.
> 
> > > > a1 = ivy.ones(3)
> > > > a1
> > > > ivy.array([1., 1., 1.])
> 
> > > > ivy.eye(3, out=a1)
> > > > ivy.array([[1., 0., 0.],
> > > > [0., 1., 0.],
> > > > [0., 0., 1.]])
> 
> > > > a1
> > > > ivy.array([[1., 0., 0.],
> > > > [0., 1., 0.],
> > > > [0., 0., 1.]])
> 
> * a1 is initialized with shape (1, 3).    ivy.array([1., 1., 1.])
> * Once it is updated in place through eye function, I assume it should retain its shape as [1, 3] and the output should be  ivy.array([1., 0., 0.])
> * However, a1 shape changes to (3, 3)
>   ivy.array([[1., 0., 0.],
>   [0., 1., 0.],
>   [0., 0., 1.]])
> * We are saving memory  by not initializing a new variable. However, shouldn't  'a1'  retain its shape (1, 3) ?
> 
> Can you please clarify?

You can look at the docstring of the function, it said that "out: optional output array, for writing the result to. It must have a shape that the inputs broadcast to." So even though we currently did not implement the ability to check if given shape is correct but instead we just overwrite the origenal shape.

ivy/functional/ivy/creation.py
```diff
@@ -952,11 +952,14 @@ def eye(
...
         value to a lower diagonal, and 0 to the main diagonal. Default: ``0``.
+    batch_shape
+        optional input that determines returning identity array shape.
+        Default: ``None``.
     dtype
         output array data type. If dtype is None, the output array data type must be the
         default floating-point data type. Default: ``None``.
     device
-         device on which to place the created array.
+        the device on which to place the created array.
```
Daniel4078, line 962: By the way, can you state why you add "the" at the start of this discription

ra9hur: PyCharm showed a grammatical error, hence included 'the'

Same with the below lines
    instances in place of any of the arguments.
    instances in place instead of the arguments.

Daniel4078: So please replace "instances in place instead of the arguments." with something that is correct but still convay the origenal meaning. And, if you did not make big changes, then please do not request a review as I will be notified of your comments and commits so there is no need.

Daniel4078 (commented):
please fix those small errors and update your example with out input.

ra9hur:
Thanks @Daniel4078 for the patience with my repeated review requests. Hope, I have incorporated all of your suggestions. Request you to please review.

ivy/functional/ivy/creation.py
```diff
@@ -974,7 +977,83 @@ def eye(
...
+    >>> x1.shape
+    (3, 2, 3)
+    # Suppose batch_shape = [a, b] then the returning identity array shape is [a, b, numRows, numColumns]
+
+
+    With :'out' input:
+
+    >>> a1 = ivy.ones(3)
+    >>> ivy.eye(3, out=a1)
+    ivy.array([[1., 0., 0.],
```
Daniel4078, line 1043: Where is the check on the content of a1?

Daniel4078 (changes requested):
just one more change needed I believe.

ivy/functional/ivy/creation.py
ra9hur: Thanks @Daniel4078 for being so meticulous with reviews!!
Have made changes to include contents of a1 before/after applying eye. Hope it looks OK now. Pl review.

Daniel4078 (approved):
LGTM

PR #7670, opened 2022-12-07
at #7499

norman135:
I might need help.
The method is currently taking in the ufunc as an argument and working as it should. It was the only way I could get the function to be dynamic that way I wouldn't have to define all ufuncs. The tests are printing nothing. I'd appreciate any form of help to optimize this like how it should be, because I think this way might not be optimal, or even considerable.
Thank you.🙏

### DIFF
--- a/ivy/functional/frontends/numpy/ufunc/methods.py
+++ b/ivy/functional/frontends/numpy/ufunc/methods.py
@@ -1,12 +1,14 @@
 # local
 from ivy.functional.frontends.numpy.func_wrapper import inputs_to_ivy_arrays
 
-
-@inputs_to_ivy_arrays
-def at(ufunc, array, indices, b=None, /):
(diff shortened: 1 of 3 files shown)


norman135:
#7499

RashulChutani (changes requested):
Hi @norman135 You will actually have to create a ufunc class (as given here, https://numpy.org/doc/stable/reference/generated/numpy.ufunc.html), and then add the "at" method for the same. Hope that helps!

norman135:
> Hi @norman135 You will actually have to create a ufunc class (as given here, https://numpy.org/doc/stable/reference/generated/numpy.ufunc.html), and then add the "at" method for the same. Hope that helps!

That should help. I'll get to it.
Appreciate.

norman135:
#8821

PR #7722, opened 2022-12-08
add numpy.random.exponential to numpy frontend #7719

ethanmclark1:


ethanmclark1:
Hello @sherry30, when you get the chance, could you explain to me what the link/check-formatting error is regarding?

PR #7742, opened 2022-12-09
Nested Array class

sherry30:


RickSanchezStoic:
Not sure if `from_row_splits` and `from_row_lengths` work for values being greater than 2-dimension. Check how these are done here:
https://github.com/unifyai/ivy/blob/master/ivy/functional/frontends/tensorflow/ragged/ragged.py

PR #7743, opened 2022-12-09
Fixed `ivy.Dtype` validation

fspyridakos:
Fixed `ivy.Dtype` and all `.as_ivy_dtype` backend functions. `ivy.Dtype` can return any valid ivy dtype, `.as_ivy_dtype` returns the respective ivy dtype as long as the backend supports it.

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -72,7 +72,7 @@ class Dtype(str):
             dtype_str = "bool"
         if not isinstance(dtype_str, str):
             raise ivy.exceptions.IvyException("dtype must be type str")
-        if dtype_str not in all_ivy_dtypes_str:
+        if dtype_str not in _all_ivy_dtypes_str:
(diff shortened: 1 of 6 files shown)


ivy/__init__.py
```diff
@@ -276,6 +277,25 @@ class Node(str):
 
 invalid_devices = ("gpu", "tpu")
 
+# supported ivy data types
+all_ivy_dtypes_str = (
```
VedPatwardhan, line 281: Instead of creating a new list/tuple of dtypes, we could just directly use `all_dtypes` present in this file. Given that all our `Dtype` classes extend `str`, `dtype_str not in ivy.all_dtypes` or  `dtype_str not in all_dtypes` in this case would also work.

VedPatwardhan (changes requested):
Hi @fspyridakos, just one small change, rest seem good 👍

PR #7764, opened 2022-12-09
Add tf-gather and tf-gather_nd

### LINKED ISSUE #7575: gather

Karthick47v2:
Close #7575


jkeane508 (changes requested):
Hi,

Thanks for submitting this PR. It seems as though the tests are failing for these functions and we would need to get them passing before we can merge. From looking at the testing logs it seems as though the tests are breaking some validation.

I would recommend looking at the parameters you are testing with and making sure that they don't cause any exception errors.

Thanks!

Karthick47v2:
Hi @jkeane508, 

Thanks for the review, I have fixed the issue with gather_nd() and tested locally. It should pass the test now.

jkeane508 (approved):
LGTM!

PR #7774, opened 2022-12-10
updates

RickSanchezStoic:
removed backend implementations

### DIFF
--- a/ivy/functional/backends/jax/data_type.py
+++ b/ivy/functional/backends/jax/data_type.py
@@ -1,5 +1,6 @@
 # global
 import numpy as np
+
 # import jax
 # import jaxlib
 import jax.numpy as jnp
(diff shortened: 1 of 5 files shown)


ivy/functional/ivy/data_type.py
```diff
@@ -510,7 +510,9 @@ def can_cast(
         b: true
     }
     """
-    return current_backend(from_).can_cast(from_, to)
+    if isinstance(from_, ivy.Dtype):
+        return (from_, to) in ivy.promotion_table
+    return (from_.dtype, to) in ivy.promotion_table
```
VedPatwardhan, line 515: I think this should be `ivy.as_ivy_dtype(from_.dtype)` instead or otherwise we could just use the `inputs_to_ivy_arrays` decorator for this function instead of the `inputs_to_native_arrays` one. I think the latter is a better option

VedPatwardhan (changes requested):
Just one small change 🙂

ivy/functional/ivy/data_type.py
RickSanchezStoic: done

ivy/functional/backends/jax/data_type.py
```diff
@@ -1,7 +1,8 @@
 # global
 import numpy as np
-import jax
-import jaxlib
+
+# import jax
```
VedPatwardhan, line 4: Please remove the comment if this import statement isn't required

VedPatwardhan (approved):
One last change, rest looks good to me 🙂

PR #7796, opened 2022-12-11
Remove compositional imp. from Ivy

karalleyna:


ivy/functional/backends/jax/linear_algebra.py
```diff
@@ -178,69 +178,18 @@ def matrix_rank(
     *,
     atol: Optional[Union[float, Tuple[float]]] = None,
     rtol: Optional[Union[float, Tuple[float]]] = None,
+    hermitian: bool = False,
     out: Optional[JaxArray] = None,
 ) -> JaxArray:
-    def dim_reduction(array):
```
simonetgordon, line 183: Hi Aleyna, sorry perhaps I wasn't clear enough in the Trello description. 

We don't want to remove the compositional implementations of matrix rank from the backends, but rather refactor and make more efficient/tidier. These functions were implemented compositionally in order to adhere to the Array API standard. I have added the `Array API` label to this PR so these tests will run. You can see that they will fail without the compositional implementations.

I recommend that you work on one backend first (let's say Jax), refactor the matrix_rank backend and make sure it's passing both the ivy and the array-api `test_matrix_rank`. 

Let me know if you have any further questions 😊

simonetgordon (changes requested):

simonetgordon:
@karalleyna Hey Aleyna, just going to close this issue following our updated discussion on the matrix_rank implementation!

PR #7800, opened 2022-12-11
Add torch.Tensor.sort

### LINKED ISSUE #7798: sort

jegork:
Closes #7798


DragosStoican:
Hi @jegork , thanks for your contribution! I see you are not testing for the `stable` argument. Is there any reason why you thought that might not be needed? If not you should test for all arguments and the signature of the function should be the same as the one in the torch documentation.

jegork:
Hey @DragosStoican 
Thanks for your reply! I am pretty sure there is no `stable` argument and the signature is correct.
`stable` is only present in `torch.sort`. Here is the related docs page (torch==1.11)
https://pytorch.org/docs/1.11/generated/torch.Tensor.sort.html#torch.Tensor.sort

DragosStoican:
Ah yes you are correct, well done. I reached the wrong link while looking it up :)

jegork:
@DragosStoican also, are there any problems with tests in github actions? the job fails but the output is empty

DragosStoican:
It doesn't matter for your PR as I've checked and your test passes just fine :) Thanks for flagging though I'll have a look at the logs. I will be merging your PR now

PR #7811, opened 2022-12-11
torch.where

### LINKED ISSUE #7806: where

LucaLumetti:
close #7806


LucaLumetti:
There are 2 details i would like to have a double check:
in _broadcast_trio, I can't set two different types, otherwise torch raise an error (not sure why)
even if in the documentation the signature of torch.where is `torch.where(condition, x, y)`, it looks like the real one is `torch.where(condition, self, other)` or `torch.where(condition, input, other)`. Quite confusing from torch and not sure why this difference from docs exists

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
```diff
@@ -63,6 +63,23 @@ def _array_idxes_n_dtype(draw, **kwargs):
     return x, idxes, dtype
 
 
+@st.composite
+def _broadcastable_trio(draw):
+    shape = draw(helpers.get_shape(min_num_dims=1, min_dim_size=1))
+    cond = draw(helpers.array_values(dtype="bool", shape=shape))
+    dtypes, xs = draw(
+        helpers.dtype_and_values(
+            available_dtypes=['float32'],
```
simonetgordon, line 72: Should be able to call a strategy like [this](https://github.com/unifyai/ivy/blob/435c4b11bbd260bd4fd0e0bc826d7b73a4bed842/ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py#L421).

simonetgordon (changes requested):

simonetgordon:
> 

Hi Daniel, apologies for the delay in reviewing this PR. See my comment above regarding the dtypes. As for the function signature, could you send me some links to where you saw these differing signatures?

LucaLumetti:
> > 
> 
> Hi Daniel, apologies for the delay in reviewing this PR. See my comment above regarding the dtypes. As for the function signature, could you send me some links to where you saw these differing signatures?

I did found a mismatch between the CPP implementation and the python documentation. Even if in [1] it is state that the signature is torch.where(condition, x, y), by using with with x and y keyword i get the following error:
```
TypeError: where() received an invalid combination of arguments - got (Tensor, x=int, y=int), but expected one of:
 * (Tensor condition)
 * (Tensor condition, Tensor input, Tensor other)
      didn't match because some of the keywords were incorrect: x, y
 * (Tensor condition, Number self, Tensor other)
      didn't match because some of the keywords were incorrect: x, y
 * (Tensor condition, Tensor input, Number other)
      didn't match because some of the keywords were incorrect: x, y
 * (Tensor condition, Number self, Number other)
      didn't match because some of the keywords were incorrect: x, y
```

I have also opened an issue [3] on PyTorch to delve more about it.
[1]: https://pytorch.org/docs/stable/generated/torch.where.html
[2]: https://github.com/pytorch/pytorch/blob/583d216c1a767fcb9a8fc33fc2f7a19ff20eba95/aten/src/ATen/native/TensorCompare.cpp#L511
[3]: https://github.com/pytorch/pytorch/issues/91003

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
```diff
@@ -148,3 +148,13 @@ def dstack(tensors, *, out=None):
...
 def take_along_dim(input, indices, dim, *, out=None):
     return ivy.take_along_axis(input, indices, dim, out=out)
+
+
+@to_ivy_arrays_and_back
+def where(condition, input=None, other=None, *, out=None):
+    if input is None and other is None:
+        return ivy.nonzero(condition)
+    elif input is not None and other is not None:
+        return ivy.where(condition, input, other, out=out)
```
simonetgordon, line 158: `ivy.where` is called the ivy torch backend implementation [here](https://github.com/unifyai/ivy/blob/abb765da1cb2b7b14de03220e06061c5efd7d14b/ivy/functional/backends/torch/searching.py#L86). Currently x1 and x2 only accept tensor types, but according to the torch.where documentation, it should also accept scalars. I will update this backend function now and you can pull the fix from master. Your frontend implementation should work then.

simonetgordon: @LucaLumetti I have update the `where` torch backend, please pull from master and see if your PR passes the checks 😊

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
LucaLumetti: I thought about it too, but when using such helper functions I get the following error:
```
RuntimeError: expected scalar type c10::Half but found float
E               Falsifying example: test_torch_where(
E                   broadcastables=(array([False]),
E                    [array([-1.], dtype=float16), array([-1.], dtype=float32)],
E                    ['float16', 'float32']),
E                   with_out=False,
E                   as_variable=[False],
E                   num_positional_args=0,
E                   native_array=[False],
E                   on_device='cpu',
E                   fn_tree='ivy.functional.frontends.torch.where',
E                   frontend='torch',
E               )
```
The error is due to a mixed dtype which is not allowed by `torch.where`. Not sure if I can set up a strategy to always have the same dtype but tests multiple dtypes at the same time.
By setting `available_dtypes=['float32']` when I run tests on my local machine everything works fine

LucaLumetti: @simonetgordon pinging you just in case you missed my comment's notification. I'll wait a response before moving forward

ivy-seed:

LucaLumetti:
I would like to continue to work on this PR @ivy-seed

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
simonetgordon: @LucaLumetti has this issue not been fixed by the changes to `torch.where`?

LucaLumetti: @simonetgordon The problem arises when calling `torch.where` with two tensors of different types, e.g. `float16` and `float32`. Using the strategy you suggested, the function is called with tensors of different types thus resulting in an error.

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
```diff
@@ -69,6 +69,24 @@ def _array_idxes_n_dtype(draw, **kwargs):
     return x, idxes, dtype
 
 
+@st.composite
+def _broadcastable_trio(draw):
+    shape = draw(helpers.get_shape(min_num_dims=1, min_dim_size=1))
```
simonetgordon, line 74: Each of the three input tensors need to be mutually broadcastable as you know, but setting the shape of each tensor to be the same is limiting the testing scope. There is a helper called `mutually_broadcastable_shapes` which you could use to generate three distinct but broadcastable shapes.

simonetgordon (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
simonetgordon: Hi @LucaLumetti, the `torch.where` promotes the dtypes of the input tensors so in principle this shouldn't be an issue:
```
def where(
    condition: torch.Tensor,
    x1: Union[float, int, torch.Tensor],
    x2: Union[float, int, torch.Tensor],
    /,
    *,
) -> torch.Tensor:
    x1, x2 = ivy.promote_types_of_inputs(x1, x2)
    return ivy.astype(torch.where(condition, x1, x2), x1.dtype, copy=False)
```

 There may be something going awry in the testing function when `input_dtypes` is specified with two different `x` and `y` tensor dtypes. Perhaps try set the final two `input_dtypes` as one dtype derived from something like `x1, x2 = ivy.promote_types_of_inputs(x1, x2)`.

ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py
simonetgordon: @LucaLumetti Hi there, are you still working on this PR?

LucaLumetti: @simonetgordon yes I am! Got a busy week but I'll go back on it in the next days

LucaLumetti: @simonetgordon I've added the mutually_broadcastable_shapes as suggested and tried to fix the dtype problem. Now the first two tests of four are good. About the last two, when the backend is tensorflow and torch, I get an exception that I cannot understand: `ivy.exceptions.IvyBackendException: torch: is_float_dtype: torch: as_ivy_dtype: dtype('bool')` (same is with tensorflow).
The stacktrace is:
```
ivy_tests/test_ivy/test_frontends/test_torch/test_indexing_slicing_joining_mutating_ops.py:866: in test_torch_where
ivy_tests/test_ivy/helpers/function_testing.py:476: in test_frontend_function
    as_variable_flags = [
ivy_tests/test_ivy/helpers/function_testing.py:477: in <listcomp>
    v if ivy.is_float_dtype(d) and not test_flags.with_out else False
ivy/func_wrapper.py:128: in new_fn
ivy/func_wrapper.py:459: in new_fn
    return fn(*args, **kwargs)
```
It looks like `ivy.is_float_dtype(d)` gets called on the type of the `condition` parameter, but I don't know when and why it works in half of the tests.

ivy-seed:

LucaLumetti:
I would like to continue to work on this PR @ivy-seed

ivy-seed:

ivy-seed:

PR #7917, opened 2022-12-13
reformatting bitwise_invert

### LINKED ISSUE #7916: bitwise_invert

AhmedEssam19:
Resolves #7916

### DIFF
--- a/ivy/container/elementwise.py
+++ b/ivy/container/elementwise.py
@@ -1485,7 +1485,7 @@ class ContainerWithElementwise(ContainerBase):
 
     @staticmethod
     def static_bitwise_invert(
-        x: ivy.Container,
+        x: Union[ivy.Container, ivy.Array, ivy.NativeArray],        
         /,
         *,
         key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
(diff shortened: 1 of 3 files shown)


zaeemansari70 (changes requested):
Hey! 
Your PR looks amazing ✨!
Just requested one change that is mentioned below which needs to be addressed before we can merge your PR! 🙂

1. You've added the container instance method docstrings kindly add the container static method docstrings as well.

Thanks!

AhmedEssam19:
@zaeemansari70 Thanks for your feedback. I have made the changes. kindly review them and let me know if there are other needed modifications.

ivy/container/elementwise.py
```diff
@@ -1485,7 +1485,7 @@ def bitwise_left_shift(
 
     @staticmethod
     def static_bitwise_invert(
-        x: Union[ivy.Container, ivy.Array, ivy.NativeArray],
+        x: ivy.Container,
```
zaeemansari70: This shouldn't have been changed in the process, only the docstrings were to be added as the typehints are correct, we take in `ivy.Array and ivy.NativeArray` as well in the static method . 🙂

zaeemansari70 (approved):
Lgtm!

ivy/container/elementwise.py
AhmedEssam19: @zaeemansari70 I have tried this method out previously and it didn't work with arrays or native arrays, only with containers. That's why I have changed the type hints

PR #8394, opened 2022-12-15
added __str__ method in numpy frontend

### LINKED ISSUE #8389: __str__

alizahidraja:
Close #8389


ivy/functional/frontends/numpy/ndarray/ndarray.py
```diff
@@ -336,6 +336,9 @@ def __float__(
     ):
         return ivy.array(ivy.reshape(self._ivy_array, -1), dtype=ivy.float64)[0]
 
+    def __str__(self):
+        return self._ivy_array.__str__()
```
jkeane508, line 340: I'm not sure this is correct, as `self` is a NumPy Array it would not contain an attribute called `_ivy_array.`

jkeane508 (changes requested):
Hi, thanks for submitting a PR! See below of requested changes :)

ivy/functional/frontends/numpy/ndarray/ndarray.py
alizahidraja: Hey! Thanks for the review

`self._ive_array` is the underlying array that is being used, the whole `self` object includes every detail, but for the `__str__` function we only need the array to be converted to string

Also, when I tried to call `str(self)` it got stuck inside an infinite recursion.

This is my opinion, let me know what the final version should be here and I will be happy to update accordingly

jkeane508: As this is the Ivy Frontend, what we expect to be input are NDarrays from Numpy. Therefore it would not contain an `_ivy_array` attribute. So this current implementation wouldn't work. As this simply returns a string I think you would only need to use Numpy to return the string object.

alizahidraja: If you have a look at other functions, e.g. `__int__` or `__float__` they are also using `_ivy_array` attribute inside them
I think this is why I am confused

And calling `str(self)` won't work as I already mentioned and got stuck in an infinite recursion 😅

alizahidraja: @jkeane508

jkeane508: Hi, I've actually looked into this (I wasn't super up to speed with how this module works).

But basically, the issue is this:

The ivy __str__() method return a different string than Numpy

`ivy.array([1]).__str__()`
`'ivy.array([1])'`

`np.array([1]).__str__()`
`'[1]'`

So in order to align behavior with NumPy we would need to format the string returned using the Ivy Method:

`ivy.array([1]).__str__()[10:-1]`

Hope this makes sense :)

jkeane508 (changes requested):

ivy/functional/frontends/numpy/ndarray/ndarray.py
alizahidraja: @jkeane508 I tried to make the function more readable and also called it in `__repr__` to avoid using the same line of codes and, instead use functions

Let me know what you think!

ivy-seed:

alizahidraja:
@jkeane508

ivy/functional/frontends/numpy/ndarray/ndarray.py
```diff
@@ -338,6 +336,9 @@ def __float__(
     ):
         return ivy.array(ivy.reshape(self._ivy_array, -1), dtype=ivy.float64)[0]
 
+    def __str__(self):
```
jkeane508, line 339: As I mentioned [here](https://github.com/unifyai/ivy/pull/8394#discussion_r1054456166), you need to recreate the function behaviour so that they match :)

jkeane508 (changes requested):
Still need to make the changes as discussed previously, also there seems to be lint (formatting) errors. You can setup a pre-commit hook as outlined [here](https://lets-unify.ai/ivy/contributing/setting_up.html#pre-commit) in the docs.

ivy/functional/frontends/numpy/ndarray/ndarray.py
alizahidraja: Hey @jkeane508 !
If you see the latest code, i have done the following
`return str(ivy.to_list(self._ivy_array))` for __str__ and it gives the correct format 

can you have another look at this and let me know

jkeane508 (changes requested):
The test [here](https://github.com/unifyai/ivy/actions/runs/3868388404/jobs/6593791389), is failing for this implementation. This would need to be passing before we can merge :)

alizahidraja:
Running into this error: `ivy.exceptions.IvyException: <U1 is not supported by ivy` need some assistance on it

jkeane508:
> Running into this error: `ivy.exceptions.IvyException: <U1 is not supported by ivy` need some assistance on it

U1 is the datatype unsigned 8-bit integer. Its being thrown as an exception, you must be passing unsigned int datatypes in testing, I would remove these test cases and see if that fixes things :)

jkeane508:
> @jkeane508 Thanks a bunch for bearing with me! I will work on the test cases and let you know!

No worries, request a review when you need a hand :)

ivy-seed:

ivy-seed:

PR #8521, opened 2022-12-16
Improved inputs_to_native_arrays.

vaatsalya123:
Added additional check in the beginning of the inputs_to_native_arrays function to see if any of the arguments are instances of ivy.Array.If none of the arguments are instances of ivy arrays, the function is called as usual and the conversion to ivy.NativeArray is skipped.Also changed del to pop both are linear complexity and added a cache register.

### DIFF
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -105,6 +105,8 @@ def inputs_to_native_arrays(fn: Callable) -> Callable:
         -------
             The return of the function, with native arrays passed in the arguments.
         """
+        if not ivy.get_array_mode():
+            return fn(*args, **kwargs)
         if not any(isinstance(arg, ivy.Array) for arg in args) and not any(isinstance(arg, ivy.Array) for arg in kwargs.values()):
             return fn(*args, **kwargs)
         # use pop on out from keywords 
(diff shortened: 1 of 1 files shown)


ivy/func_wrapper.py
```diff
@@ -105,23 +105,16 @@ def new_fn(*args, **kwargs):
         -------
             The return of the function, with native arrays passed in the arguments.
         """
-        if not ivy.get_array_mode():
+        if not any(isinstance(arg, ivy.Array) for arg in args) and not any(isinstance(arg, ivy.Array) for arg in kwargs.values()):
```
VedPatwardhan, line 108: I don't think we should be removing the `get_array_mode` usage as it's primary purpose is to avoid doing additional checks when Ivy isn't operated in array mode and removing this would prevent the users from having any control over this

VedPatwardhan (changes requested):
Just one small change, rest looks good to me 🙂

ivy/func_wrapper.py
VedPatwardhan: Just for reference, it was added here https://github.com/unifyai/ivy/pull/2276

VedPatwardhan (approved):
lgtm!

PR #7773, opened 2022-12-10
gelu #7679

Karim-Akmal:

PR #8807, opened 2022-12-17
gelu #7679

Karim-Akmal:


Karim-Akmal:
#7773
this was the last PR I did before the announcement. So I closed it and made a new PR like said in the announcement.

ivy/functional/backends/numpy/activations.py
```diff
@@ -25,10 +25,14 @@ def leaky_relu(
...
+    /,
+    *,
+    approximate: bool = False,
+    out: Optional[np.ndarray] = None
 ) -> np.ndarray:
     if approximate:
-        ret = 0.5 * x * (1 + np.tanh(x * 0.7978845608 * (1 + 0.044715 * x * x)))
+        ret = 0.5 * x * (1 + np.tanh(0.7978845608 * (x + 0.044715 * x * x * x)))
     else:
         ret = 0.5 * x * (1 + ivy.erf(x / np.sqrt(2)))
```
YushaArif99, line 37: When writing a backend implementation for a function as a composition, only the backend-specific functions should be used. Here you're using an `ivy` function (**ivy.erf()**) when implementing `gelu` for numpy backend. Also, you should use `np.astype()` instead of `ivy.astype()`

YushaArif99 (changes requested):
Hi @Karim-Akmal 

The overall implementation does look good to me. Could you kindly make the small changes requested 🙂 
Also, could you kindly mention the issue # for the PR. This is required so that we can automatically mark this function as completed once this PR gets merged. Thanks!

ivy/functional/backends/numpy/activations.py
Karim-Akmal: @YushaArif99 it is just , numpy has no erf method.

Karim-Akmal:
#7679

Karim-Akmal:
#8992
This is the issue for the PR
@YushaArif99

ivy-seed:

ivy/functional/backends/numpy/activations.py
YushaArif99: You could, however, this would create a dependency on the scipy module which is not something we want. Alternatively, I'd suggest creating your own NumPy-compatible version of the `erf` function. The scipy documentation defines the function as being a purely mathematical one ` 2/sqrt(pi)*integral(exp(-t**2), t=0..z)` and so implementing it should be straightforward. I think it would be better to implement this function for the NumPy backend itself.
One question that comes to mind though is that if there is an `ivy.erf()` function, isn't there also a NumPy backend implementation of this function as well? If so, you should directly use that. If it's not there for some reason, then it's best to implement it in the NumPy backend so that `ivy.erf()` works when using a NumPy backend. 
Hope that makes sense 🙂

YushaArif99 (changes requested):
Hi @Karim-Akmal 
Really sorry for the late reply. I was really busy wrapping up some due tasks before the holidays and so reviewing this PR totally slipped under my radar. 
Thanks for making the previous changes I requested. Could you kindly also resolve the issue with the NumPy backend that I have mentioned above? Thanks! 😃

ivy/functional/backends/numpy/activations.py
Karim-Akmal: @YushaArif99  Unfortunately for some reasons there is no erf in the NumPy. I understood you , but where exactly should I implement the ivy.erf() in the NumPy backend. What I mean should I just make it at the top of the activations.py file or in another file?

YushaArif99: Well I looked into the master branch of Ivy and noticed that the `erf` function has already been implemented for the NumPy backend. In this case, I think it's okay to use `ivy.erf()` since there is an existing NumPy implementation of it in Ivy

YushaArif99:
Hi @Karim-Akmal 

I looked at the Issue # you provided and realized that you're working on the `Reformat Activations` task. 
As part of the task, could you incorporate the following changes: 
- add the docstring examples for the `gelu` function for the ivy Array class. 
- revert the changes back to the original implementation for the NumPy backend.

Thanks!

ivy/array/activations.py
```diff
@@ -60,10 +60,10 @@ def gelu(
 
         Examples
         --------
-        >>> x = ivy.array([0.3, -0.1])
-        >>> y = x.gelu()
+        >>> x = ivy.array([-1.2, -0.6, 1.5])
+        >>> y = ivy.gelu(x)
```
YushaArif99, line 64: Shouldn't this be `x.gelu()`?

YushaArif99 (changes requested):
Hi @Karim-Akmal 
Thanks for making the requested changes. There are just a couple of small changes to make, otherwise, the PR seems ready to be merged. Can you kindly make those changes. Thanks!

Also, the tests for the NumPy backend are failing. I think the error is related to using `np.astype`. Could you replace that with `ivy.as_type()`

ivy/array/activations.py
Karim-Akmal: Yea, You are right, going to change it now.

YushaArif99 (changes requested):
Hi @Karim-Akmal 

Thanks for making the requested changes. I just need to verify one small thing. I have noticed that you've modified the `gelu` function in torch backend. The original version contains `@with_unsupported_dtypes({"1.11.0 and below": ("complex", "float16")}, backend_version)` whereas you've removed the "complex" dtype. Any particular reasons for doing so? I'm asking this because one of the intelligent tests is failing due to some unsupported dtype being passed to the `torch.nn.functional.gelu()` function. I just want to verify whether this was failing previously as well. Can you kindly add back the "complex" dtype and check if that fixes the failing test. 
Other than that everything looks good. Let me know if you have any other questions 😃

Karim-Akmal:
> Hi @Karim-Akmal
> 
> Thanks for making the requested changes. I just need to verify one small thing. I have noticed that you've modified the `gelu` function in torch backend. The original version contains `@with_unsupported_dtypes({"1.11.0 and below": ("complex", "float16")}, backend_version)` whereas you've removed the "complex" dtype. Any particular reasons for doing so? I'm asking this because one of the intelligent tests is failing due to some unsupported dtype being passed to the `torch.nn.functional.gelu()` function. I just want to verify whether this was failing previously as well. Can you kindly add back the "complex" dtype and check if that fixes the failing test. Other than that everything looks good. Let me know if you have any other questions 😃

@YushaArif99 FIRST, I didnt modify it at all, in the original version, there was no complex dtype, and now when I did add it , it raised a conflict, so should I commit and ask for a review while there is a conflict? or just discard these changes?

YushaArif99 (changes requested):
Hi @Karim-Akmal 

If you look at the master branch of Ivy, it did originally contain the `complex` dtype. (https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/torch/activations.py#L33)

This might not be the reason that the tests are failing. Though regardless of this fact, you shouldn't really modify the backend implementations since the task is to only reformat the functions. 
I think other than that everything seems fine and the PR should be ready to be merged. 

As always, let me know if you have any other questions ! 😃

Karim-Akmal:
> Hi @Karim-Akmal
> 
> If you look at the master branch of Ivy, it did originally contain the `complex` dtype. (https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/torch/activations.py#L33)
> 
> This might not be the reason that the tests are failing. Though regardless of this fact, you shouldn't really modify the backend implementations since the task is to only reformat the functions. I think other than that everything seems fine and the PR should be ready to be merged.
> 
> As always, let me know if you have any other questions ! 😃

@YushaArif99 So, Now what should I do? commit it with the complex dtype or no? I did not modify it from the original master branch.
If I should add the 'complex' I already did in the last commit, if not, should I delete the last commit and request a review again?

YushaArif99 (approved):
Hi @Karim-Akmal 

Thanks for making the changes. I think everything looks fine right now and the tests also seem to be passing. I'll now merge this PR. Thanks a lot for your contribution!  😄

Issue #8522, opened 2022-12-16
tensorflow.separable_conv2d


anurag629:

AnnaTz:
implemented in https://github.com/unifyai/ivy/commit/8000d02d3383dc73d08f7c41b7fc89d53c151084

PR #8812, opened 2022-12-17
ivy backend function for seperable_conv2d

anurag629:


anurag629:
#8522

agniiyer:
Thanks for taking a look! :) @Ishticode

Best,
Agni.

anurag629:
> Hi @anurag629. Thanks @agniiyer
> 
> Great PR. This requires some changes.
> 
> **Firstly**, we should not use `convolve` from scipy. Numpy is a dependency of Ivy but scipy should not be. Please kindly check [conv2d numpy backend implementation](https://github.com/unifyai/ivy/blob/38a87c31888b1960730bc9e8d3a1a5b34454785c/ivy/functional/backends/numpy/layers.py#L107) without the use of scipy's `convolve` and remove the `scipy` import slightly_smiling_face
> 
> **Secondly**, I see that you have used `torch.nn.functional.DepthwiseSeparableConv2d` for torch which I believe does not exist.
> 
> I couldn't find any docs on it and upon trying to call it using torch module I get the error `AttributeError: module 'torch.nn.functional' has no attribute 'DepthwiseSeparableConv2d'` (Besides, this call is a bit strange.sweat_smile The capitalisation suggests a class conventionally, while `torch.nn.functional` a function. We should only be calling functions in function implementation)
> 
> Lastly, there is a typo in the name for static method - `sepe` should be `sepa`. I know that is a bit picky sweat_smile but it would lead to test failures hence I am pointing it out.
> 
> I hope these points are understandable. Feel free to ask anything @anurag629
> 
> Thanks @agniiyer

Hi @Ishticode Now using `np.tensordot` in place of `convolve`. As you suggested `conv2d` but in `tensordot` there will be less code and also easy to understand.
 
And about `torch.nn.functional.DepthwiseSeparableConv2d` and  typo both are silly mistake by me slightly_smiling_face

One Question:
the test is showing the error of missing 1 required positional argument: 'padding' but for me everything is fine.

    E   TypeError: separable_conv2d() missing 1 required positional argument: 'padding'
    E   Falsifying example: test_separable_conv2d(
    E       x_k_s_p=(['float64'], [array([[[[-1.]]]])], (1, 1), (1,), 'VALID'),
    E       test_flags=<ivy_tests.test_ivy.helpers.test_parameter_flags.FunctionTestFlags at 0x7f7a04969370>,
    E       backend_fw=<module 'ivy.functional.backends.numpy' from '/ivy/ivy/functional/backends/numpy/__init__.py'>,
    E       fn_name='separable_conv2d',
    E       ground_truth_backend='jax',
    E   )
    E

PR #8834, opened 2022-12-18
add triu creation and test function

maciejbalawejder:
Everything seems to be working fine, all four tests passed.


AnnaTz (changes requested):
All is good but you need to remove changes in the .idea folder in order to merge your PR.

maciejbalawejder:
Cool, I revert those files to their original form but still it shows that I changed something.

AnnaTz:
> Cool, I revert those files to their original form but still it shows that I changed something.

Yes removing this can be tricky. You could try reverting back to the commit before those changes and then re-pasting the code of your function.

AnnaTz:
Alternatively you could create a new branch and then a new PR to be sure that you don't change the idea folder this time.

maciejbalawejder:
Yes exactly that's what I was thinking about haha

PR #8842, opened 2022-12-18
Std torch frontend

artykov1511:


kurshakuz:
Hey @artykov1511! Can you please link the original issue to that PR? It is also not clear for me what do you want to implement here. Are you adding std or sqrt functions? The title says std, but you have only added the test for the sqrt function.

artykov1511:
Hey @kurshakuz, I was implementing std, but I guess this task was already given to someone else  before I allocated it.

artykov1511:
Closed #8841

artykov1511:
Hey @kurshakuz, when I committed and pushed from my local branch, all other previous commits from other branches were pushed as well. I dont why this happened. Can you help delete these commits?

PR #8884, opened 2022-12-19
Update test_tensor.py

FrankShi9:
real #8882


jkeane508 (changes requested):
Hi, 

this test implementation needs to be in the same PR as the functional implementation. As they both need to be on the same branch in order for the test to work.

Thanks :)

PR #8902, opened 2022-12-19
nanquantile

### LINKED ISSUE #7014: nanquantile

Change-m:
Close #7014


zaeemansari70 (changes requested):
Hey! 
Your PR looks amazing ✨!
Just some small changes are required after then we can surely merge your PR 🙂

1. There are multiple lint errors which needs to be resolved, also can you please remove the comments from your code. 🙂

Thanks!

zaeemansari70 (changes requested):
There are still 3 lint errors that are being showed, can you kindly fix them? 🙂

ivy/functional/backends/tensorflow/experimental/statistical.py
```diff
@@ -70,3 +70,31 @@ def quantile(
...
+    interpolation: Optional[str] = None,
+    out: Optional[Union[tf.Tensor, tf.Variable]] = None,
+) -> Union[tf.Tensor, tf.Variable]:
+    count = 0
+    for i in a:
+        if tf.math.is_nan:
+            continue
+        count += 1
+    if count % 2 == 0:
+        print((count + 1) / 2)
```
zaeemansari70, line 91: Here

ivy/functional/backends/tensorflow/experimental/statistical.py
```diff
@@ -70,3 +70,31 @@ def quantile(
...
+    else:
+        print(a[((count + 1) // 2) - 1])
```
zaeemansari70, line 93: Here

zaeemansari70 (changes requested):
Thanks for the changes 🙂
There is just one more last change that needs to be done, I just noticed that you've used `print` statements inside the tensorflow backend implementation, no implementation should have those statements.
I've highlighted where they need to be removed from.
Thanks!

ivy/functional/backends/tensorflow/experimental/statistical.py
```diff
@@ -70,3 +70,31 @@ def quantile(
...
+    interpolation: Optional[str] = None,
+    out: Optional[Union[tf.Tensor, tf.Variable]] = None,
+) -> Union[tf.Tensor, tf.Variable]:
+    count = 0
+    for i in a:
+        if tf.math.is_nan:
+            continue
+        count += 1
+    if count % 2 == 0:
+        return ((count + 1) / 2)
```
zaeemansari70, line 91: here

ivy/functional/backends/tensorflow/experimental/statistical.py
```diff
@@ -70,3 +70,31 @@ def quantile(
...
+    else:
+        return (a[((count + 1) // 2) - 1])
```
zaeemansari70, line 93: here

zaeemansari70 (changes requested):
Thanks for the changes 🙂 
Actually you changed the `print` statements to `return`, which is causing the tests to fail, previously the tests were passing, was this by mistake or the changes were intended to be like that? 🙂 
Thanks!

zaeemansari70 (changes requested):
Thanks a lot for the changes! 🙂
All the backends are now passing except for torch, there is a problem in the torch backend implementation which is causing the tests for `torch` to fail.
There is an Assertion error being caused, which in this case states that your output is not matching the ground truth backend.

The values are:
 `AssertionError: [[[-1.0000000e+00 -1.0000000e+00][-1.0000000e+00 -1.9999624e+07]]] != [[[-1.0000000e+00 -2.0000000e+00][-1.0000000e+00 -1.9999624e+07]]]`

Which are not equal, there is a slight difference in the outputs, you have to fix the torch backend for it to pass.

Do let me know of you have any other questions🙂 
Thanks!

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -234,3 +234,40 @@ def test_quantile(
...
+
+# nanquantile
+
+
+@handle_test(
+    fn_tree="functional.experimental.nanquantile",
+    dtype_x_axis=statistical_dtype_values(function="quantile"),
+    q=st.floats(min_value=0, max_value=1),
+    keepdims=st.booleans(),
+    interpolation=st.data(),
```
zaeemansari70, line 247: Here.

zaeemansari70 (changes requested):
You can not use the `st.data()` inside the `@handle_test` decorator, you have to handle this in the composite strategy but I can see that it is being already handled where you are unpacking `dtype_x_axis`. 
I guess this `interpolation=st.data(),` can be removed if it doesn't cause the tests to fail, if it does cause them to fail then you have to handle it inside the composite strategy.

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -234,3 +234,36 @@ def test_quantile(
...
         keepdims=keep_dims,
     )
+
+
+# nanquantile
+
+
+@handle_test(
+    fn_tree="functional.experimental.nanquantile",
+    dtype_x_axis=statistical_dtype_values(function="quantile"),
```
zaeemansari70, line 244: You have to pass in the `safety factor` parameter here which will deal with the assertions, as the output values being generated are pretty close and I think for these statistical functions we are testing too extensively.
You can find the safety factor here, https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py#L16
It has the implementation in the function, you just have to pass the number, try not to pass a big value as it will restrict the dataset too much.

zaeemansari70 (changes requested):
Please use the safety_factor for these assertions, I've explained it below 🙂

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -234,3 +234,36 @@ def test_quantile(
...
+    fn_name,
+    ground_truth_backend,
+):
+    input_dtype, x, axis, interpolation, q = dtype_x_axis
+    helpers.test_function(
+        ground_truth_backend=ground_truth_backend,
+        input_dtypes=input_dtype,
+        test_flags=test_flags,
+        fw=backend_fw,
+        fn_name=fn_name,
```
zaeemansari70, line 262: You can pass `rtol=1e-03` and `atol=1e-03` as it will be handling the tolerance here for jax uptill 3 decimal spaces.
You can see here how to use these parameters https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_elementwise.py#L702

zaeemansari70 (changes requested):
Great effort, we are very near with it to be merged 🙂👍 
Just some more changes are needed and I've highlighted them below, as it is causing an assertion on a very close values with jax. 

Do ask if you have any other questions.🙂
Thanks!

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -13,8 +13,8 @@
 
 @st.composite
 def statistical_dtype_values(draw, *, function):
-    large_abs_safety_factor = 72
-    small_abs_safety_factor = 72
+    large_abs_safety_factor = 24
```
zaeemansari70, line 16: Please can you revert them back to their original value which was `2`?
If it makes the test to fail only then do the change below I've mentioned.

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -13,8 +13,8 @@
...
+    small_abs_safety_factor = 24
     if function in ["mean", "median", "std", "var"]:
```
zaeemansari70, line 18: This change is only to be done if it fails reverting the `safety factor = 2`:
 `if function in ["mean", "median", "std", "var", "nanquantile"]:`
 This condition will set the safety factor to 24.

zaeemansari70 (changes requested):
You have reverted them to 24 which was not the original value
I've requested the small changes below 🙂 
Thanks!

zaeemansari70 (changes requested):
There is no restriction for how many dimensions we are testing, testing very extensively also cause assertions, try to pass `max_dim_size=5,` after this line over here https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py#L29

Thanks!

ivy/functional/backends/torch/experimental/statistical.py
```diff
@@ -100,3 +100,33 @@ def quantile(
...
+        return torch.nanquantile(
+            a, q, keepdim=keepdims, interpolation=interpolation)
+    if isinstance(axis, list) or isinstance(axis, tuple):
+        desired_shape = []
+        current_shape = a.size()
+        for i in range(len(current_shape)):
+            if i not in axis:
+                desired_shape += [current_shape[i]]
+        a = a.reshape((-1,) + tuple(desired_shape))
+        a = torch.nanquantile(
```
zaeemansari70, line 128: torch's `nanquantile()` does not accept tuples or lists as `dim` but other frameworks do, you've passed `0` here which basically disturbs the output.
You have to come up with a logic to handle axis and tuples here except for passing `dim=0`

ivy/functional/backends/tensorflow/experimental/statistical.py
```diff
@@ -70,3 +70,30 @@ def quantile(
...
+    a: Union[tf.Tensor, tf.Variable],
+    q: Union[int, float],
+    /,
+    *,
+    axis: Optional[Union[int, Tuple[int]]] = None,
+    keepdims: Optional[bool] = False,
+    interpolation: Optional[str] = None,
+    out: Optional[Union[tf.Tensor, tf.Variable]] = None,
+) -> Union[tf.Tensor, tf.Variable]:
+    count = 0
```
zaeemansari70, line 85: Your code from line 85-93 is not usable in the implementation further, kindly remove this or do let me know if this is being used and I've missed a point.

zaeemansari70 (changes requested):
Requested some changes 🙂 
Do let me know if you have any questions, thanks! 🙂

PR #8948, opened 2022-12-20
Reformat fourier_encode: Docstring examples done

### LINKED ISSUE #8143: fourier_encode

akosua00:
close #8143


Ishticode (changes requested):
Hi @akosua00 
Thank you very much for the PR 🙂
This requires some changes. 
Could the instance calls on the array and container be used as an instance call instead of ivy.fourier_encode. You can check my commit for one example and apply this to array and container file examples.

Secondly, after the print statement the actual result should be printed but currently it is empty. Please ensure the result of `print(z)` is included in the docstring.

Please let me know if anything is unclear and request a review from me when you have made the changes. Thanks.

Ishticode (changes requested):
As you can see from my last commit on this PR. 
Can you simplify the second example as the output is unreasonably large and also the the third argument is specified as kw only and hence should be passed in as `num_bands=3`.

ALSO

The example in Containers submodule should be creating container instances and not just the arrays. Also the container's static method `static_fourier_encode` is still missing the examples.

Could you make these changes and request my review and hopefullly this can be merged. Lmk if anything is unclear. Thanks

akosua00:
Hi @Ishticode, I have changed the examples for the container instance and static methods, but the docstring test has to different output to what I expected for both methods.
e.g. 
 {
      a:(<classivy.array.array.Array>shape=[2,9]),
      b:(<classivy.array.array.Array>shape=[2,9])
} 
I will change the docstring example to be the same as this but I do not understand why this is the output

Ishticode (changes requested):
Sure. You can change it to this closed format output. The container displays a shorter format of arrays when they are too big. This is a builtin mechanism to allow display of container with many leaves easily.

So yh you can change it to this output and the test docstring should pass.

Ishticode (approved):
@akosua00 
Everything looks great. Thank you very much for your contribution 🙂.
I will get this merged now.

Issue #8952, opened 2022-12-20
Add percentile  - Statistical functions to Numpy Frontend


Tiago-B-C-Reis:
Add Statistical functions to Numpy Frontend #1529
https://github.com/unifyai/ivy/issues/1529

Tiago-B-C-Reis:
@ivy-leaves, can you please add this sub-task to the [main issue](https://github.com/unifyai/ivy/issues/1529)?

PR #8953, opened 2022-12-20
Created percentile function in the numpy frontend

Tiago-B-C-Reis:
Add percentile to function frontend and respective unitest + start of percentile def on functional Ivy.
percentile #8952


ivy/functional/frontends/numpy/statistics/percentile.py
```diff
@@ -0,0 +1,27 @@
+# global
```
jkeane508, line 1: Here we don't need a new file for percentile, we should implement this in order_statistics.py

ivy/functional/ivy/statistical.py
```diff
@@ -1260,3 +1260,17 @@ def einsum(
 
     """
     return current_backend(operands[0]).einsum(equation, *operands, out=out)
+
```
jkeane508, line 1263: This is not part of the task, and of currently we are not adding any more functions to the Ivy API

ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics/test_percentile.py
```diff
@@ -0,0 +1,60 @@
+# global
```
jkeane508, line 1: This should be in test_order_statistics.py not in its own file.

jkeane508 (commented):
Hi, thanks for submitting this PR some changes are needed :)

I would consider reading over the docs to better understand the structure of the project. Moreover, you seem to have lint errors in your implementations, you can add a pre-commit hook to prevent these errors - https://lets-unify.ai/ivy/contributing/setting_up.html#pre-commit

ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics/test_percentile.py
Tiago-B-C-Reis: Hello, thank you very much for the feedback, I will start working on this but I just have one question:
Should this 'percentile' task be attached to this issue #8953, like the median for example?

image

PR #8964, opened 2022-12-21
adding examples in docs where test_values=false

zaeemansari70:


docs/partial_source/deep_dive/ivy_frontends_tests.rst
```diff
@@ -490,6 +490,74 @@ This function requires us to create extra functions for generating :code:`shape`
 * We use :code:`helpers.get_dtypes` to generate :code:`dtype`, these are valid numeric data types specifically for Torch.
 * :func:`torch.full` supports :code:`out` so we generate :code:`with_out`.
 
+Testing Without Using Tests Values
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are some places where we cannot test using hypothesis, as some functions have multiple solutions and all those
```
CatB1t, line 496: We actually still use Hypothesis, it's just how our definition of `Equality` is different, for example we can still have Hypothesis to generate random values for `x`, if we have a function in our backend that is called `add_noise` and we call it on `x` and on we try to assert (we interally use `assert np.all_close`) that the result from `torch` backend matches `tensorflow` the test will always fail, because the function `add_noise` depends on a random seed internally that we have no control over, what we change is only how we test for equality, in which in that case we can not.

docs/partial_source/deep_dive/ivy_frontends_tests.rst
```diff
@@ -490,6 +490,74 @@ This function requires us to create extra functions for generating :code:`shape`
...
+            full_matrices=full_matrices,
+        )
+        ret = [ivy.to_numpy(x) for x in ret]
+        frontend_ret = [np.asarray(x) for x in frontend_ret]
+
+        u, s, vh = ret
+        frontend_u, frontend_s, frontend_vh = frontend_ret
+
+        assert_all_close(
+            ret_np=u @ np.diag(s) @ vh,
```
CatB1t, line 552: I'd suggest using a different example that is easier to understand, as this may be overwhelming a bit for new reader.

CatB1t (changes requested):

docs/partial_source/deep_dive/ivy_frontends_tests.rst
```diff
@@ -490,6 +490,61 @@ This function requires us to create extra functions for generating :code:`shape`
 * We use :code:`helpers.get_dtypes` to generate :code:`dtype`, these are valid numeric data types specifically for Torch.
 * :func:`torch.full` supports :code:`out` so we generate :code:`with_out`.
 
+Testing Without Using Tests Values
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+While even using hypothesis, there are some cases in which we do :code:`test_values=False` for example, we have a
```
CatB1t, line 496: First sentence is grammarly incorrect, try to rephrase that.

docs/partial_source/deep_dive/ivy_frontends_tests.rst
```diff
@@ -490,6 +490,61 @@ This function requires us to create extra functions for generating :code:`shape`
...
+function add_noise() and we call it on x and we try to assert (we interally use assert np.all_close) that the result
+from torch backend matches tensorflow the test will always fail, because the function add_noise() depends on a random
+seed internally that we have no control over, what we change is only how we test for equality, in which in that case
+we can not and we have to reconstruct the output as shown in the example below.
```
CatB1t, line 500: We should also explain the example of how we test these cases, the following example may be a bit **hard** to understand without explanation, as we still calling `assert_all_close`, which may be confusing for the reader. I'd suggest also adding a very simplified example, does not have to be in the codebase, just explain how we can check for size of the output, dtype, etc.

CatB1t (changes requested):

ivy-seed:

PR #8975, opened 2022-12-21
average #8853

hslau44:
#8853


ivy/functional/frontends/jax/numpy/statistical.py
```diff
@@ -145,3 +146,82 @@ def max(a, axis=None, out=None, keepdims=False, where=None):
...
+
+
+@to_ivy_arrays_and_back
+def average(a, axis=None, weights=None, returned=False, keepdims=False):
+
+    # canonicalize_axis to ensure axis or the values in axis > 0
+    if axis is None:
+        pass
+    elif isinstance(axis, int):
+        pass
```
Ishticode, line 158: Perhaps we don't need these two conditional branches? Could we remove these

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_statistical.py
```diff
@@ -486,3 +486,44 @@ def test_jax_numpy_max(
...
+    as_variable,
+    native_array,
+    on_device,
+    fn_tree,
+    frontend,
+):
+    x_dtype, x, axis = dtype_x_axis
+    
+    if isinstance(axis, tuple):
+        axis = axis[0]
```
Ishticode, line 512: Does this not fiter out axis data like (1,2). I believe we should test for this as jax.numpy's average function does take tuple of ints as inputs also.

Ishticode (changes requested):
Hi @hslau44 
Excellent PR.
Apologies for the delay and thank you for your patience.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_statistical.py
hslau44: Thank you for reviewing.  I am not able to pass the test for tuple cases. The message ver long and here is the last part of the message
```
def _average(a, axis: Optional[Union[int, Tuple[int, ...]]] = None, weights=None,
                 returned=False):
      if weights is None: # Treat all weights as 1
        _check_arraylike("average", a)
        a, = _promote_dtypes_inexact(a)
        avg = mean(a, axis=axis)
        if axis is None: 
          weights_sum = lax.full((), core.dimension_as_value(a.size), dtype=avg.dtype)
        else:
>         weights_sum = lax.full_like(avg, core.dimension_as_value(a.shape[axis]), dtype=avg.dtype)
E         jax._src.traceback_util.UnfilteredStackTrace: TypeError: tuple indices must be integers or slices, not tuple
E         
E         The stack trace below excludes JAX-internal frames.
E         The preceding is the original exception that occurred, unmodified.
```
In the latest document of Jax, there is an extra `elif` branch, it is to extract the shape of the array by arg `axis` as value and calculate the size by prod

```
def _average(a: ArrayLike, axis: Axis = None, weights: Optional[ArrayLike] = None,
             returned: bool = False, keepdims: bool = False) -> Union[Array, Tuple[Array, Array]]:
  if weights is None: # Treat all weights as 1
    _check_arraylike("average", a)
    a, = _promote_dtypes_inexact(a)
    avg = mean(a, axis=axis, keepdims=keepdims)
    if axis is None:
      weights_sum = lax.full((), core.dimension_as_value(a.size), dtype=avg.dtype)
    elif isinstance(axis, tuple):
      weights_sum = lax.full_like(avg, _prod(core.dimension_as_value(a.shape[d]) for d in axis))
    else:
      weights_sum = lax.full_like(avg, core.dimension_as_value(a.shape[axis]))  # type: ignore[index]
```

hslau44:
I want to test the argument `weights` but I am not able to navigate the test helper to find methods to produce an array for `weights` that depended on `array`

ivy/functional/frontends/jax/numpy/statistical.py
hslau44: I have remove the conditional branches and tested it based on the currently written test

Ishticode (changes requested):
Hey. 
Strangely enough the jax and numpys behaviour differ here. Which can be seen through this example `print(jnp.average(jnp.reshape(jnp.arange(0, 24), (2,3,4)), axis=(0,2)))` where you may exchange jnp with np etc. 
Numpy gives the result while jax throws an error. Your code seems to work in according with np which is not a bad thing.

To generate arrays based on weights you can either use the `st.shared` strategy from Hypothesis or you can [write you own composite strategy](https://lets-unify.ai/ivy/deep_dive/ivy_tests.html#writing-your-own-strategy) with the `@st.composite` decorator. The latter would be a helper function where  you can easily include the logic for dependency.

Your PR is nearly ready to be merged tbh. Could you just ensure to remove the conflicts.

Apologies for the late response and I will try to priotise your PR on next review. Thanks.

Ishticode (changes requested):
Although this works in most cases but hypothesis has found this example where the results are slightly different between the output given by the native jax and the one implemented here in the frontend. The example is below. The difference seems to be a round error and the test caught this and failed. Not sure if this should be tolerated or fixed. I will ask @VedPatwardhan to advice on this.

Here is the code:

```python
import jax.numpy as jnp
#import ivy.functional.frontends.jax.numpy as jnp

x = jnp.array([-1., -1., -1., -1., -1., -1., -1., -1., -1.], dtype=jnp.float16)
w = jnp.array([-1., -1., -1., -1., -1., -1., -1., -1.6777210e+07, 1.6777167e+07], dtype=jnp.float32)
print(jnp.average(x, axis=0, weights=w))
## output -0.98 with jax.numpy and -1.0 with jax frontend here
```

Ishticode:
Ive added the test tolerance on advice from VedPatwardhan given that the results are exact for most examples and any examples with minor divergence are understood to be the result of extensive compostion.

Ishticode (approved):
@hslau44 
Fabolous Work. Thank you very much for you patience and the contribution. I will get this merged now. 🙂

PR #8979, opened 2022-12-21
add tensorinv to torch frontend

### LINKED ISSUE #7786: tensorinv

sondertg:
Closes #7786


ivy/functional/frontends/torch/linalg.py
```diff
@@ -94,3 +94,33 @@ def svd(A, /, *, full_matrices=True, driver=None, out=None):
 def svdvals(A, *, driver=None, out=None):
     # TODO: add handling for driver
     return ivy.svdvals(A, out=out)
+
+
+@to_ivy_arrays_and_back
+@with_unsupported_dtypes({"1.11.0 and below": ("float16", "bfloat16")}, "torch")
+def tensorinv(input, ind=2):
```
AnnaTz, line 101: According to the documentation there is also an `out` argument to be included.

ivy/functional/frontends/torch/linalg.py
```diff
@@ -94,3 +94,33 @@ def svd(A, /, *, full_matrices=True, driver=None, out=None):
...
+                inverse_tensor = ivy.inv(input)
+            else:
+                return ivy.reshape(input, shape=inverse_shape_tuple)
+        else:
+            raise ValueError(
+                            "Tensor shape must satisfy prod"
+                            "(A.shape[:ind]) == prod(A.shape[ind:])")
+    else:
+        raise ValueError("Expected a strictly positive integer for 'ind'")
+    return ivy.reshape(inverse_tensor, shape=inverse_shape_tuple)
```
AnnaTz, line 126: There is a special case that I think needs handling. When `A` is a 2-dimensional tensor and `ind=1`, this function should return `torch.linalg.inv(A)`. This is a note in the [documentation](https://pytorch.org/docs/stable/generated/torch.linalg.tensorinv.html).

ivy/functional/frontends/torch/linalg.py
```diff
@@ -94,3 +94,33 @@ def svd(A, /, *, full_matrices=True, driver=None, out=None):
...
+            inverse_shape_tuple = tuple([*inverse_shape])
+            if len(ivy.shape(input)) > 1:
+                inverse_tensor = ivy.inv(input)
+            else:
+                return ivy.reshape(input, shape=inverse_shape_tuple)
+        else:
+            raise ValueError(
+                            "Tensor shape must satisfy prod"
+                            "(A.shape[:ind]) == prod(A.shape[ind:])")
+    else:
```
AnnaTz, line 124: You should also check if 'A' is invertible and raise an error if it's not.

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -529,3 +529,76 @@ def test_torch_svdvals(
...
+        native_array,
+        on_device,
+        fn_tree,
+        frontend,
+):
+    dtype, x, ind = dtype_input_ind
+    helpers.test_frontend_function(
+        input_dtypes=dtype,
+        as_variable_flags=as_variable,
+        with_out=False,
```
AnnaTz, line 596: As I said above you'll need to add an `out` argument, so then you should set `with_out=True`.

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -529,3 +529,76 @@ def test_torch_svdvals(
...
+            invertible = True
+        except np.linalg.LinAlgError:
+            pass
+
+    return input_dtype, input, ind
+
+
+@handle_frontend_test(
+    fn_tree="torch.linalg.tensorinv",
+    dtype_input_ind=_get_inv_square_matrices()
```
AnnaTz, line 580: You don't need all this code to produce `dtype_input_ind`. There is a simpler way to do this. 
Have a look at how `linalg_inv` does it:
```
        available_dtypes=helpers.get_dtypes("float"),
        min_value=-1e4,
        max_value=1e4,
        small_abs_safety_factor=3,
        shape=helpers.ints(min_value=2, max_value=10).map(lambda x: tuple([x, x])),
    ).filter(lambda x: np.linalg.cond(x[1]) < 1 / sys.float_info.epsilon),
```
In your case, you'll need to use `dtype_values_axis` instead of `dtype_and_values`. This will give you back an axis that will be your `ind`. 
The `.filter` part is basically a check that makes sure the array will be invertible. So you'll need it too.
And finally, `tensorinv` should work on square and non-square matrices, so you won't need the `.map` part at the `shape` argument. But you will need to `.filter` the shape so that the product of the first and last dimensions is equal.

ivy/functional/frontends/torch/linalg.py
```diff
@@ -94,3 +94,33 @@ def svd(A, /, *, full_matrices=True, driver=None, out=None):
...
+            prod_ind_start *= i
+        for j in shape_ind_end:
+            prod_ind_end *= j
+        if prod_ind_end == prod_ind_start:
+            inverse_shape = shape_ind_start + shape_ind_end
+            input = ivy.reshape(input, shape=(prod_ind_end, -1))
+            inverse_shape_tuple = tuple([*inverse_shape])
+            if len(ivy.shape(input)) > 1:
+                inverse_tensor = ivy.inv(input)
+            else:
```
AnnaTz, line 118: Does the original `tensorinv` function work for 1-dimensional input arrays? Maybe you should raise an error instead of returned a reshaped array in that case.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
sondertg: I tried this first but couldn't achieve to pass the tests, so I used helper implemented for jax tensorinv. I'll work on it.

ivy/functional/frontends/torch/linalg.py
sondertg: clion64_9f7jPFEAqJ
clion64_Tf8Q7y6hXd
I tried to follow logic implemented at origin of the function.  I also checked jax version of it. Here original function has no limit for case `ind=1` as far as i see.

AnnaTz: Then, you should not handle 1-d arguments either. How did you exclude them from the test and got an error?

sondertg: I set min dim size to 2:
```py
def _get_inv_square_matrices(draw):
    dim_size = draw(helpers.ints(min_value=2, max_value=10))
```
and got this error:
`hypothesis.errors.FailedHealthCheck: The smallest natural example for your test is extremely large.`

It is probably related to the wrong helper I used. I might solve this by removing that helper and using one you told me. I'll update when I am able to do.

ivy/functional/frontends/torch/linalg.py
sondertg: I think 2-dim tensor is basically a matrix and to satisfy product of shape dimensions equals till given axis(in our case it's 1, so the shape must be square) and product of rest, it should keep the shape same. Because `inverse_shape` is the same of `input_shape` in this case.
(e.g.: `shape = (n, n) and ind = 1`) 
In case it's not square shaped and `ind=1`, will throw an error. Imho, that's why the original c++ implementation has no handling code block for this case. 
If you still think that should be added, I can do that. However, I feel like it cause code redundancy.

ivy/functional/frontends/torch/linalg.py
sondertg: Could you please check if it is done properly?

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
sondertg: Related PR #7399
I found useful to check this PR and asked assignee of it about his opinion. He wasn't a hundred percent sure, but I got a confirmation that he said: ```
>i don't think there is a built-in helper function that allows you to generate such values, if there is such a function maybe you could ask your reviewer to point you towards it ?```

I tried `dtype_and_values_axis` helper, and It didn't work. I think it is not possible to control the shape of the tensor without drawing the axis and vice versa. If I filter the shape, the axis should be half of the length of the shape. If I filter arrays and their shape, again, can't control drew axis. 

```py
 np.prod(ivy.shape(x)[:int(len(ivy.shape(x))//2)]) == np.prod(ivy.shape(x)[int(len(ivy.shape(x))//2):]
 ```
As you may see, this has no relation with axis we need. I also checked arguments related to axis, none gives us to define it with respect to shape of drawn array.(`valid_axis` is just being sure that given ind, in the range of the shape dims.)

I also tried to solve it by using dtypes_values and set axis half of the shape inside test_linalg_tensorinv. Again, couldn't check product condition.

Sorry if it becomes a bit long.

File: ivy/functional/frontends/torch/linalg.py [hunk 7]
```diff
@@ -116,6 +116,38 @@ def svdvals(A, *, driver=None, out=None):
...
+            prod_ind_end *= j
+        if prod_ind_end == prod_ind_start:
+            inverse_shape = shape_ind_start + shape_ind_end
+            input = ivy.reshape(input, shape=(prod_ind_end, -1))
+            inverse_shape_tuple = tuple([*inverse_shape])
+            if len(ivy.shape(input)) > 1 and (ivy.inv(input)).any():
+            # TODO: replace ivy.inv(input) with ivy.linalg.cond or torch.linalg.cond
+                inverse_tensor = ivy.inv(input)
+            else:
+                ret = ivy.reshape(input, shape=inverse_shape_tuple, out=out)
```
AnnaTz, line 142: I'm a little confused as to why you reshape here. Is this part of the c++ implementation?

ivy/functional/frontends/torch/linalg.py
```diff
@@ -116,6 +116,38 @@ def svdvals(A, *, driver=None, out=None):
...
+        for i in shape_ind_start:
+            prod_ind_start *= i
+        for j in shape_ind_end:
+            prod_ind_end *= j
+        if prod_ind_end == prod_ind_start:
+            inverse_shape = shape_ind_start + shape_ind_end
+            input = ivy.reshape(input, shape=(prod_ind_end, -1))
+            inverse_shape_tuple = tuple([*inverse_shape])
+            if len(ivy.shape(input)) > 1 and (ivy.inv(input)).any():
+            # TODO: replace ivy.inv(input) with ivy.linalg.cond or torch.linalg.cond
```
AnnaTz, line 139: If I understand correctly you are using `(ivy.inv(input)).any()` to check if `input` is invertible. I think a simpler way to do this would be `ivy.det(input) > sys.float_info.epsilon`, i.e. checking if the determinant is close to zero.

ivy/functional/frontends/torch/linalg.py
```diff
@@ -116,6 +116,38 @@ def svdvals(A, *, driver=None, out=None):
...
+                inverse_tensor = ivy.inv(input)
+            else:
+                ret = ivy.reshape(input, shape=inverse_shape_tuple, out=out)
+                return ret if ivy.inv(ret) else not_invertible
+        else:
+            raise ValueError(f'{prod_cond} or {not_invertible}.')
```
AnnaTz, line 145: I think it would be more accurate to discriminate between these two errors.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
AnnaTz: It can be done like this 
'''
    dtype_input_ind=helpers.dtype_values_axis(
        available_dtypes=helpers.get_dtypes("float"),
        min_value=-1e4,
        max_value=1e4,
        small_abs_safety_factor=3,
        shape=helpers.get_shape(min_num_dims=1).filter(
            lambda x:  np.prod(x[:int(len(x)//2)]) == np.prod(x[int(len(x)//2):])
        ),
        valid_axis=True,
        force_int_axis=True,
    ).filter(lambda x: np.linalg.cond(x[1]) < 1 / sys.float_info.epsilon),
'''

sondertg: Hi again, as I mentioned earlier, I already tried that. In this case, our drawn axis should be equal to `len(ivy.shape(x))//2`. However, the axis will be a random integer between `len(ivy.shape(x))>axis>1` but we need the mid-point of the shape drawn that fit the product condition. 
We have three conditions here, two depends on each other (In this case third condition not satisfied.):
1. Invertibility 
> [x]
2. Product condition 
>[x]
3. Axis should be the one that satisfy the product condition. 
> [ ]

Other option I tried is that using `dtype_and_values` helper with this setup and setting axis to the midpoint of the shape inside the test function. This time, our test handler can't check ind values.

ivy/functional/frontends/torch/linalg.py
sondertg: Yes, I tried to follow the same logic while implementing the function.

> clion64_9f7jPFEAqJ clion64_Tf8Q7y6hXd 

```C++
shape_ind_end.insert()
```

As far as I see, this part does that job. At the end of the function, returned result reshaped with this variable.

ivy/functional/frontends/torch/linalg.py
sondertg: Original function uses `torch.linalg.inv_ex` with, `check_errors=True` parameter. I guess the correct way of checking invertibilty is using the original approach. That can be added later.

For now, I am not sure if non-zero determinant always imply invertibility. In case it is a two-dimensional tensor, that can work, but I couldn't find info about if it has more than 2 dimensions with a quick search. I can dig more, or I can use that as a condition with your advice If you want me to do.

sondertg: When I start, inv_ex wasn't defined. I added original condition.

ivy/functional/frontends/torch/linalg.py
```diff
@@ -131,6 +131,41 @@ def inv_ex(input, *, check_errors=False, out=None):
...
+            prod_ind_start *= i
+        for j in shape_ind_end:
+            prod_ind_end *= j
+        if prod_ind_end == prod_ind_start:
+            inverse_shape = shape_ind_start + shape_ind_end
+            input = ivy.reshape(input, shape=(prod_ind_end, -1))
+            inverse_shape_tuple = tuple([*inverse_shape])
+            try:
+                inv_ex(input, check_errors=True)
+            except RuntimeError:
```
sondertg, line 155: I am not sure if using try except clause or assert is better here, I can swap in case of request.

AnnaTz: I think for the product condition and for `ind > 0` it would be more comprehensive to use an assertion.

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
AnnaTz: Yeah sorry I haven't gotten into depth with this function so I had misunderstood the 3rd condition. 
What do you mean "This time, our test handler can't check ind values."? If I understand correctly, the function itself forces the axis to be a certain integer given the shape of the input array, otherwise it produces an error. So it's okay to have the test use only this certain axis value. Alternatively, you could first draw a random integer for the axis and then draw the array filtering the shape based on that integer.

AnnaTz: You wouldn't do that inside `handle_frontend_test`. You could create an `st.composite` function where you first draw `ind` and then the `dtype_input` using the drawn value.

sondertg: Sorry if I am wrong, but there is an already defined `st.composite` function to exactly fit this function, as discussed at PR #7399.  What will be different from removing the jax helper I use right now from the same with different name? In case the problem is repetition, adding jax helper to the `helpers` and importing may help? The one used in there is following original test function implementation in the c++ code. I'm just asking what we're trying to avoid to understand and do better?

AnnaTz: The issue I see with the helper in the PR you have referenced is that it handpicks some shapes and indexes. It doesn't really randomly generate them through the hypothesis module. This means that the testing is limited and will not cover a wide range of possible inputs.

AnnaTz: I'm proposing something like:
```
@st.composite
def _tensorinv_helper(draw):
ind = draw(helpers.ints(min_value=1, max_value=10))
 dtype, input = draw(helpers.dtype_values_axis(
        available_dtypes = helpers.get_dtypes("float"),
        min_value=-1e4,
        max_value=1e4,
        small_abs_safety_factor=3,
        shape=helpers.get_shape(min_num_dims=1).filter(
            lambda x:  np.prod(ivy.shape(x)[:ind]) == np.prod(ivy.shape(x)[ind:])
        ).filter(lambda x: np.linalg.cond(x[1]) < 1 / sys.float_info.epsilon))
```

AnnaTz: Our preferred approach is to exhaustively test each function using `hypothesis`, which is indeed different than what Pytorch is doing.

sondertg: > I'm proposing something like:

I just used this one, here's the result:
pycharm64_v9bD96fbgH
It draws arrays, but not tensorlike. 
And here is the original torch tensorinv test, written in python:
clion64_M7ZJiOP77v

I got what I need to focus. Furthermore, I'll try to increase test coverage If I can do. But as far as I see, this function requires a very specific set of inputs to test. 
Also, I have a question about invertibility condition defined with ` np.linalg.cond(x[1]) < 1 / sys.float_info.epsilon`: Is this non-zero determinant imply invertibility for multidimensional tensors too?

AnnaTz: You need to give `input=x[0]` not `input=x` to the `test_frontend_function`. I think this will solve the TypeError.

AnnaTz: Yes it checks invertibility for any multidimensional array.

sondertg: > You need to give `input=x[0]` not `input=x` to the `test_frontend_function`. I think this will solve the TypeError.

`x[0]` means we don't pass a tensor to the function. We got that error because the function got a one dimensional array, so asked for a tensor. I solved `TypeError` by setting `min dim=2`. This is the error we need to solve.

```py
E   hypothesis.errors.Unsatisfiable: Unable to satisfy assumptions of test_torch_tensorinv.
```

I also removed invertibility filter and just tested with product condition. Why did I do that? Because I think the condition is too complicated to filter. And the error above is still present.

Please correct me if I am wrong. The current implementation of helper draws random arrays with the random number of dimensions we set, then we try to filter the shape that satisfies our product condition. So from that many random arrays, it is very unlikely to find a few that fit our product condition. Because it sets, both dimensions and the array sizes randomly. 

Why not filtering array (also dimensions) sizes  ? -Because those helpers get random values, and filtering them to find few whose shapes obey the rule is really hard. (Even if it works, probably it will create very few tensors (maybe none) that we need. A lot of randomness for very particular outcome.)

How can we generate such rules that produce tensors with the given rule? - That is the question we need to find an answer. 

To create such a rule, the length of the shape (number of dimensions) should be defined first. Because our helper function should know how many numbers required to decide factorization. 

Then we need to set the result of product to a certain number.  That depends on previous step.
(Alternatively, we can try to start with this step.  We need a number that have at least `x`  distinct factors so that we can set half of the shape length smaller than x. Even that require further filtering.)

Finally, we can fill each half of the shape with factorization of the number with the required number of factors. 

At last, we can fill the elements that will create invertible sub-tensors and tensors.

I am still not sure if this is the appropriate way to test this function. Most of the `linalg` functions have very simple input sets, unlike `tensorinv`. Again, I am kinda feel confident about none of those helpers around can help us with this process without creating a nested logic. All we can do is just using a combination of those helpers to have maybe just a bit bigger input pool than the `_get_inv_square_matrices` helper already has.

sondertg: > That is the question we need to find an answer.

I wrote a shape generation block that satisfy product rule and appropriate ind. However, filtering invertibility is not working as expected.

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -715,3 +715,82 @@ def test_torch_svdvals(
...
+            break
+    if ind == 1:
+        shape += (product_half, product_half)
+    if ind == 2:
+        shape += (1, product_half, product_half, 1)
+    shape_cor = ()
+    for i in shape:
+        shape_cor += (int(i),)
+    dtype = draw(helpers.get_dtypes("float"))
+    input = draw(helpers.array_values(
```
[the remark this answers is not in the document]
AnnaTz: Why not use the usual `dtype_and_values`  that allows to control the dimensions? You can pass `shape=shape_cor` and `available_dtypes=dtype`.

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
AnnaTz: This is an interesting problem. I haven't come across such a case in the past so thanks for sharing your thought process.
Your approach seems to be going in the right direction. What is the problem with invertibility?

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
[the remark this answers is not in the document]
sondertg: I tried that one first, it didn't draw shapes I passed when I set `min_dim=2`. It added one extra dimension. Without the parameter It threw HealtyCheck error saying that size is too large. Working on it, will have time at weekend.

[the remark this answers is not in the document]
sondertg: update:
I wanted to test my shape generation part if it works well. I assumed `np.linalg.tensorinv` will work.  
```py
def test():
    def factors(x):
        result = [1, ]
        i = 2
        while i * i <= x:
            if x % i == 0:
                result.append(i)
                if x // i != i:
                    result.append(x // i)
            i += 1
        result.append(x)
        return np.array(result)

    ind = np.random.randint(2, 6)
    product_half = np.random.randint(2, 25)
    factors_list = factors(product_half)
    shape = ()
    while len(shape) < ind and ind > 2:
        while np.prod(shape) < product_half:
            a = factors_list[np.random.randint(len(factors_list))]
            shape += (a,)
        if np.prod(shape) > product_half or len(shape) > ind:
            shape = ()  
        while len(shape) < ind and shape != ():
            shape += (1,)
        if np.prod(shape) == product_half:
            shape += shape[::-1]
            break
    if ind == 1 and shape == ():
        shape += (product_half, product_half)
    if ind == 2 and shape == ():
        shape += (1, product_half, product_half, 1)
    shape_cor = ()
    for i in shape:
        shape_cor += (int(i),)
    return shape_cor, ind
x, y = test()
for i in range(100000):
    try:
        np.linalg.tensorinv(np.random.rand(*x), ind=y)

    except:
        print('failed')
```

I found that generated shape is not satisfying the product condition in some rare cases. So, I fixed that and tested if any of the generated `shape` and `ind` pairs are invalid for tensorinv function. By using `np.random.rand`, invertibility condition is satisfied without any extra code.

Then I used both helpers `dtype_and_values` and `array_values`. I went through step by step. As you can see from the code on above, invertibility the only condition left to satisfy. I used `ivy.det` bigger than epsilon as the filter and got healthy check error that saying generated data is so big. This make sense because hypothesis helpers draws array values and since we pass some relatively big shapes to the helpers.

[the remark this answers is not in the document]
sondertg: I forgot to mention checking determinant with `ivy.linalg.det` or `np.linalg.det` didn't work.
```py
shape_cor, ind = ((1, 2, 2, 1), 2)
x = np.random.rand(*shape_cor)
np.linalg.det(x)                             # first
np.linalg.tensorinv(x, ind=ind)     # second
```
First one throws an `LinAlgError: Last 2 dimensions of the array must be square`  second one works fine. Healthy check error comes when I use this as a filter `filter(lambda x: np.linalg.cond(x[1]) < 1 / sys.float_info.epsilon)`

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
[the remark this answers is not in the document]
AnnaTz: Thanks for the update! You're almost done I think 🙂

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -759,3 +759,75 @@ def test_torch_solve(
...
+            break
+    if ind == 1 and shape == ():
+        shape += (product_half, product_half)
+    if ind == 2 and shape == ():
+        shape += (1, product_half, product_half, 1)
+    shape_cor = ()
+    for i in shape:
+        shape_cor += (int(i),)
+    dtype = draw(helpers.get_dtypes("float"))
+    input = np.random.rand(*shape_cor)
```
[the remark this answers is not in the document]
AnnaTz: About the second case, I think there are a couple of possible ways around it. 
1. You could define a simple helper function like
```
def _check_invertibility(x, ind):
    try:
        np.linalg.tensorinv(x, ind=ind)
    except:
        return False
    return True
```
and then use the `array_values` helper with `.filter(lambda x: _check_invertibility(x[1], ind)`.
2. Does `np.linalg.tensorinv` ever cause an exception for 1-d inputs (or, equivalently, would we ever consider a 1-d array as non-invertible?). If not, then you could use the `array_values` helper with something like `.filter(lambda x: np.linalg.cond(x[1]) < 1 / sys.float_info.epsilon if len(x[1].shape) > 1 else True`.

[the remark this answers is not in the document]
AnnaTz: Fyi, the reason I insist on trying to find a way to use our hypothesis-based helpers is that hypothesis keeps an example database that helps a lot with debugging. 
Quoting from the documentation: "When Hypothesis finds a bug it stores enough information in its database to reproduce it. This enables you to have a classic testing workflow of find a bug, fix a bug, and be confident that this is actually doing the right thing because Hypothesis will start by retrying the examples that broke things last time.". 
Even if a function passes all tests when merged to master, it's possible it might fail in the future due to a change in another part of our code. In this case it would be best if we could debug it efficiently using the hypothesis test failure logs.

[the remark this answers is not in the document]
sondertg: I wonder why are we using `np.linalg.tensorinv` inside the test function that testing if four versions of this function gives same result. Isn't it against the purpose of the test. Still, I tried them.

1. Also, I didn't understand why we are using `x[1]` but here's the result. 

pycharm64_5DYdU9R5Id
result:
pycharm64_2CPie18owo
I changed `x[1]` to `x`:
pycharm64_N1TumRn14R
2.
When we try to send any shape doesn't satisfy product condition, it throws error. 
Ex:
```py
np.linalg.tensorinv(np.random.rand(1))
```
This is valid because `np.prod(shape[:ind] == np.prod(shape[ind:])` for any value of positive integer `ind`. But when we try:
```py
np.linalg.tensorinv(np.random.rand(2))
```
This throws an error that saying last 2-d is not square. My shape generator never creates a 1-d array. Cause only valid case is 1d array with 1 element and no coupling between `ind` values for this case. But since the first one didn't work, I didn't try this solution.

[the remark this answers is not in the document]
AnnaTz: Alright, I will experiment a bit with it locally and if nothing works for me either, we will merge it as it is.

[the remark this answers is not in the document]
AnnaTz: image
image
image
With the above changes the test passes locally. It's a bit hacky to use the `try-except` in `test_torch_tensorinv` but I think it's still preferable as it allows us to use the hypothesis helper.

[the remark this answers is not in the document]
sondertg: From my understanding, you bypass failed cases. In my humble opinion, this can cause passing the test for a set of values which shouldn't be.  Adding a keyword argument to the helper for this function might solve the issue? I can try to work on it after adding a `TODO` to this `_tensorinv_helper`. In case you want to proceed with your solution, I can push commits.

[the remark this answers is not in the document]
AnnaTz: Regardless of how we test it, you assert inversiblity inside `tensorinv`, which means that we expect `tensorinv` to work only for invertible inputs (if I understand correctly). So, I think there is no harm in skipping the cases where the inversibility assertion gets triggered. Is there another case where this would happen " this can cause passing the test for a set of values which shouldn't be"?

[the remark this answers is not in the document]
sondertg: But we are not checking invertibility of the input as it is. We reshape the input and then checking invertibility (which original function also does the same). I guess that's why we can't filter them in the first place. The equation to hold is the dot product of inverted tensor with itself should be equal to identity of the space. Let me try to explain with an example:
```py
# ind = 2, shape = (8, 3, 4, 6), a is input, b is the one we check if inverse exist
a = np.random.rand(8, 3, 4, 6)
prod_ind_end = np.prod(a.shape[:2])
print('prod_ind_end', prod_ind_end)
b = a.reshape(prod_ind_end, -1)
print('shape a', a.shape, '\n', 'shape b', b.shape)
# np.linalg.inv(a)
np.linalg.inv(b)
```
If you run this first, you'll see no error. Then remove `np.linalg.inv(b)` and  `#`, it will throw an error.  Any of determinant calculation also won't work because of the same reason. We need to check the determinant of the reshaped tensor. 

I just tried a filter by drawing a square shape and reshaping after that due to my generated shape, but it didn't work again. The main difficulty to test invertibility is that. Almost all the (probably all) input tensors are not invertible (correct way to say this, we can't check their invertibility with the functions we have. `det`, `inv` etc.) for dimensions higher than 2. We need a function that calculates determinant or inverse, w.r.t given axis. I also checked this but couldn't find.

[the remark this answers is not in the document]
AnnaTz: But if we do 
```
    except Exception as e:
        if hasattr(e, "message"):
            if 'Reshaped tensor is not invertible' in e.message:
                assume(False)
```
then we only skip the cases where `tensorinv` is not supposed to work. We assume False only for the exception thrown by your `tensorinv` code.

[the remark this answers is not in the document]
sondertg: > Is there another case where this would happen " this can cause passing the test for a set of values which shouldn't be"?

Yes, there might be. Explanation is in the previous comment. I didn't mean error message is wrong. We completely ignore this error, which should be there for a reason. 

For now, I just pushed commits with your correction. I'll think about this, and if I found a solution, I can create a PR and inform you.

PR #8995, opened 2022-12-22
Experimental `broadcast_shapes` function

### LINKED ISSUE #7967: broadcast_shapes

Ishticode:
Close #7967

### DIFF
--- a/ivy/functional/backends/jax/experimental/manipulation.py
+++ b/ivy/functional/backends/jax/experimental/manipulation.py
@@ -282,3 +282,7 @@ def hsplit(
     out: Optional[JaxArray] = None,
 ) -> JaxArray:
     return jnp.hsplit(ary, indices_or_sections)
+
+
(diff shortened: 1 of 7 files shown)


Ishticode:
This also extends the get_shape helper to allow for num of examples to be specified.

Currently the test fails when incompatible shapes are generated. But this shouldn't be failing the test but rather checking that all backends do indeed raise an exception correctly for the same incompatible shapes in a unified manner. The @handle_exceptions decorator is on.

Any ideas?

VedPatwardhan (commented):
Hi @Ishticode, thanks for requesting the review. Yeah, I think the `test_function` is just like any other code we run, so ig it should generate a unified exception as well, if it generates an exception for one backend and not for another with the same example, then I think it's an inconsistency between backends which we should fix. Also, another suggestion would be to look into `nph.mutually_broadcastable_shapes` or something named similar to this, which has been used in a number of tests to generate broadcastable shapes 🙂

PR #9065, opened 2022-12-24
jax frontend x64 behaviour and promotion fix

Ishticode:


ivy/functional/frontends/jax/func_wrapper.py
```diff
@@ -1,11 +1,12 @@
...
 from typing import Callable
+import inspect
 
 # local
 import ivy
 import ivy.functional.frontends.jax as jax_frontend
-
+import ivy.functional.frontends.numpy as np_frontend
 
 def _is_jax_frontend_array(x):
```
fspyridakos, line 11: small lint fix, add a line above this function

ivy/functional/frontends/jax/func_wrapper.py
```diff
@@ -88,3 +89,58 @@ def new_fn(*args, **kwargs):
...
 def to_ivy_arrays_and_back(fn: Callable) -> Callable:
     return outputs_to_frontend_arrays(inputs_to_ivy_arrays(fn))
+
+
+def to_ivy_dtype(dtype):
+    if not dtype or isinstance(dtype, str):
+        return dtype
+    if dtype in (int, float, bool) or ivy.is_native_dtype(dtype):
+        return ivy.as_ivy_dtype(dtype)
+    return np_frontend.to_ivy_dtype(dtype)._ivy_dtype
```
fspyridakos, line 99: Remove `_ivy_dtype` the function returns ivy dtypes

ivy/functional/frontends/jax/func_wrapper.py
```diff
@@ -88,3 +89,58 @@ def new_fn(*args, **kwargs):
...
+                "int64": "int32",
+            }
+            dtype = dtype_replacement_dict[dtype] \
+                if dtype in dtype_replacement_dict else dtype
+
+        if isinstance(dtype, str):
+            return fn(*args, dtype=dtype, **kwargs)
+        if isinstance(dtype, np_frontend.dtype):
+            return fn(*args, dtype=dtype._ivy_dtype, **kwargs)
+        return fn(*args, dtype=np_frontend.to_ivy_dtype(dtype), **kwargs)
```
fspyridakos, line 143: don't need these lines, conversion is already handled earlier, just `return fn(*args, dtype=dtype, **kwargs)`

ivy/functional/frontends/jax/func_wrapper.py
```diff
@@ -88,3 +89,58 @@ def new_fn(*args, **kwargs):
...
+
+        dtype = to_ivy_dtype(dtype)
+        if not jax_frontend.config.jax_enable_x64:
+            dtype_replacement_dict = {
+                ivy.int64: ivy.int32,
+                ivy.uint64: ivy.uint32,
+                ivy.float64: ivy.float32,
+                "float64": "float32",
+                "uint64": "uint32",
+                "int64": "int32",
```
fspyridakos, line 134: These are unecessary, the strings will reference the ivy dtype keys

ivy/functional/frontends/jax/numpy/__init__.py
```diff
@@ -366,6 +367,21 @@ def promote_types_jax(
...
 
 
+def _handle_x64_promotion(d):
+    dtype_replacement_dict = {
+        ivy.int64: ivy.int32,
+        ivy.uint64: ivy.uint32,
+        ivy.float64: ivy.float32,
+        "float64": "float32",
+        "uint64": "uint32",
+        "int64": "int32",
```
fspyridakos, line 377: same as above

fspyridakos (changes requested):
Mostly good, just a few comments

ivy/functional/frontends/jax/func_wrapper.py
Ishticode: Will do cleanup at the end including lints etc. 😊

Ishticode:
@fspyridakos 
I've updated the req changes. 

The example at the bottom throws and error. Yes when the dtype arg is passed as the string and `ivy.*` formats work.
However,  `jnp.float64` is interpreted as `ivy.frontends.numpy.dtype('float64')` and fails to be found in `ivy_dtype_dict`  [at this line numpy backend at this line](https://github.com/unifyai/ivy/blob/730156f69e26b654e659062be0ac0203e311ca9c/ivy/functional/backends/numpy/data_type.py#L162). Somehow, the `np.dtype("float64")` in the `ivy_dtype_dict` is considered different than `ivy.frontends.numpy.dtype('float64')`. 🤔

I will dig into this. Is there any soluton other than adding this to the replacement dictionary perhaps?


```python
import ivy.functional.frontends.jax.numpy as jnp
x = jnp.array([1., 2, 3], dtype=jnp.float64)
print(x.dtype)
```

ivy/functional/frontends/jax/func_wrapper.py
```diff
@@ -88,3 +89,51 @@ def new_fn(*args, **kwargs):
 
 def to_ivy_arrays_and_back(fn: Callable) -> Callable:
     return outputs_to_frontend_arrays(inputs_to_ivy_arrays(fn))
+
+
+def to_ivy_dtype(dtype):
+    if not dtype or isinstance(dtype, str):
+        return dtype
+    if dtype in (int, float, bool) or ivy.is_native_dtype(dtype):
+        return ivy.as_ivy_dtype(dtype)
```
fspyridakos, line 98: Add 
`if isinstance(dtype, np_frontend.dtype):
       return dtype.ivy_dtype`

Ishticode: Done. it was `_ivy_dtype` . Thanks for this and It works as expected now.

Ishticode:
I've added the exhaustive CI label to run all the tests. Feel free to review after that. Thank you so much for helping out @fspyridakos 🙂.

fspyridakos:
Hey @Ishticode, I made some changes, namely I added support for complex and I removed the jax `to_ivy_dtype` since I think the numpy one covered everything already. Anna added to conversion for `np_frontend.dtype` in the `np_frontend.to_ivy_dtype` a few days ago. Could you please check on your end if this works fine and if it does go ahead and merge. Thanks!

fspyridakos:
There also seems to be a lot of test failing because of this change, this is because we get the frontend function output by doing ivy.set_backend("jax") and ivy has `enable_x64` on by default, will need to talk with Mustafa about how we can fix the testing for jax

PR #9257, opened 2022-12-30
Ivy control flow ops Implementation

YushaArif99:
Implemented Basic Control Flow Ops for the Dynamic Control Flow task
- ivy.IfElse
- ivy.WhileLoop


Nightcrab:
There seems to be some pascal case functions in ``ivy/functional/ivy/control_flow_ops.py`` is there a reason they are named like this? (since this kind of capitalisation is used for classes)

The functions also need tests, but I'll open a separate issue for that.

YushaArif99:
No particular reason. I just did it because `ivy.IfElse` would call `if_else` ad so I thought to give it a different name. Should I rename it to `ivy.if_else()` and `ivy.while_loop()` instead?

PR #9261, opened 2022-12-30
Feature/ldexp

matthew-64:


jkeane508 (changes requested):
Hey,

Thanks for getting involved with Ivy. I see that this is a draft PR but I just thought I would let you know that we can't accept PRs that make changes to any files such as `.gitignore` or `.idea` files

Happy Coding :)

PR #9275, opened 2023-01-01
docs(setting_up) : fix gh cli incorrect command

y1450:
see https://github.com/cli/cli/issues/4616


y1450:
i feel bad for triggering the CI, just for a change in docs. Is there any I way I can avoid this in future.

rahulsnkr:
@y1450 Thanks for the fix! The CI always gets triggered for a PR, however we have intelligent tests in place, which run only when specific files are changed. Thanks for the concern though :)

PR #9297, opened 2023-01-02
Fixing np.std frontend function

hello-fri-end:
#8165

### DIFF
--- a/ivy/functional/frontends/numpy/statistics/averages_and_variances.py
+++ b/ivy/functional/frontends/numpy/statistics/averages_and_variances.py
@@ -74,7 +74,6 @@ def std(
     /,
     *,
     axis=None,
-    correction=0.0,
     keepdims=False,
     out=None,
     dtype=None,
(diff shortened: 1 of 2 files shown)


hello-fri-end:
@DragosStoican Hi there! Looks like the test case is still failing. I've figured out what's causing the error but I'm not sure how to fix it. Let me explain:


Consider the following input:
```python
a = np.array(-1.000e+00, -6.552e+04], dtype=np.float64)
dtype=np.float16
```

When we directly call `np.std` on this input, the output is `inf`
```python
>>> np.std(a, dtype = dtype)
inf
```
However, if we first convert the type of `a` to `np.float16` and then call the `np.std` function ( this is how it's implemented in the frontend), the output is `nan`

```python
>>> b = a.astype(dtype)
>>> b
array([ -1., -inf], dtype=dtype)
>>> np.std(b)
nan
```

And therefore, we get the assertion error that the ground truth `inf` is not equal to `nan`, which is what the frontend function returns.

How should this error be fixed?

DragosStoican:
Hi @hello-fri-end, for questions regarding the current implementation of the function it's best to ask this question on discord. Someone will be able to give you a more accurate answer there!

hello-fri-end:
@DragosStoican Hey! Looks like the issue has already been fixed in #9353

PR #9324, opened 2023-01-03
fixing the np.nanstd frontend test

### LINKED ISSUE #9045: Fix averages_and_variances.test_numpy_nanstd
|   |   |
|---|---|
|paddle|<a href="https://github.com/unifyai/ivy/actions/runs/6408417944"></a>
|numpy|<a href="https://github.com/unifyai/ivy/actions/runs/6408093971"></a>
|jax|<a href="https://github.com/unifyai/ivy/actions/runs/6408417944"></a>
|tensorflow|<a href="https://github.com/unifyai/ivy/actions/runs/6408417944"></a>
|torch|<a href="https://github.com/unifyai/ivy/actions/runs/6408417944"></a>

hello-fri-end:
Closes #9045


hello-fri-end:
The test for this frontend function will still fail and for the same reason as I pointed out in #9297. 

Copying and pasting the problem here for ref:

Consider the following input:
```python
a = np.array(-1.000e+00, -6.552e+04], dtype=np.float64)
dtype=np.float16
```

When we directly call `np.std` on this input, the output is `inf`
```python
>>> np.std(a, dtype = dtype)
inf
```
However, if we first convert the type of `a` to `np.float16` and then call the `np.std` function ( this is what the ivy front end implementation does), the output is `nan`

```python
>>> b = a.astype(dtype)
>>> b
array([ -1., -inf], dtype=dtype)
>>> np.std(b)
nan
```

And therefore, we get the assertion error that the ground truth `inf` is not equal to `nan`, which is what the frontend function returns.

How should this error be fixed?

kurshakuz:
What is the result produced from `ret_gt`? I assume we would want to fully replicate behavior of the original function, therefore the result produced from `np.std(a, dtype = dtype)` should be produced in our frontend implementation. You can change the frontend implementation to match the same behavior. Hope it makes sense! 🙂

hello-fri-end:
> What is the result produced from `ret_gt`?
```bash
>           assert np.allclose(
                np.nan_to_num(ret_np), np.nan_to_num(ret_from_gt_np), rtol=rtol, atol=atol
            ), "{} != {}".format(ret_np, ret_from_gt_np)
E           AssertionError: [nan] != [inf]
```

>I assume we would want to fully replicate behavior of the original function, therefore the result produced from `np.std(a, dtype = dtype)` should be produced in our frontend implementation. You can change the frontend implementation to match the same behavior. Hope it makes sense! slightly_smiling_face

The problem is that `ivy.std` ( which is what `nanstd` calls after converting all the nans to 0) does not have a `dtype` argument and hence we first convert the input using `ivy.astype` function.

Let me explain the test-case -- maybe it will help in coming up with a solution to the problem:

Input
```bash
Falsifying example: test_numpy_nanstd(
E               dtype_and_a=(['float64'], [array([-1.000e+00, -6.552e+04])], 0, 0),
E               dtype=['float16'],
E               where=True,
E               keep_dims=False,
E               num_positional_args=1,
E               native_array=[False],
E               with_out=False,
E               as_variable=[False],
E               fn_tree='ivy.functional.frontends.numpy.nanstd',
E               frontend='numpy',
E           )
```
So basically, when we cast the input from `float64` to `float16`,  `-6.552e+04` gets converted to `inf` and this is why `std` returns `nan`. My guess is that internally, NumPy, first checks if there are any `inf` in the input array and if there are, it will directly return `nan`. Otherwise, they first cast the type and then return the result.

kurshakuz:
Now the question is why are we converting float64 to float16? I believe that should not be the case as we still want to return the result in th same data type

hello-fri-end:
> Now the question is why are we converting float64 to float16? I believe that should not be the case as we still want to return the result in th same data type

Hey @kurshakuz ! I have added an explicit condition in the test case to prevent downcasting. This will ensure that we don't get any overflow errors. The tests are now passing. This is ready to be merged,

kurshakuz (changes requested):
Hey! I have just tested your implementation locally and it is currently failing due to the AssertionError. I have just handpicked the values and it seems that discrepancy is very small. So I would suggest if you can remove some rounding that occur in the implementation due to type casting. If that does not help, try tuning atol or rtol values.

hello-fri-end:
> Hey! I have just tested your implementation locally and it is currently failing due to the AssertionError. I have just handpicked the values and it seems that discrepancy is very small. So I would suggest if you can remove some rounding that occur in the implementation due to type casting. If that does not help, try tuning atol or rtol values.

Hey, that's strange. The tests are passing on my system. I've increased `atol` and `rtol`. Can you check again please?

hello-fri-end:
Hey @kurshakuz! Can you take a look here please?

kurshakuz (changes requested):
Hey @hello-fri-end ! I am sorry for a delayed response. I have checked the updated implementation and it is still producing the AssertionError. I have checked the values and the difference seems to be quite big and increasing rtol and atol further will not be a feasible solution. Can you please verify if the implmentation is correct and there is no rouding errors?

hello-fri-end:
> Hey @hello-fri-end ! I am sorry for a delayed response. I have checked the updated implementation and it is still producing the AssertionError. I have checked the values and the difference seems to be quite big and increasing rtol and atol further will not be a feasible solution. Can you please verify if the implmentation is correct and there is no rouding errors?

Hi @kurshakuz! I had accidentally decreased the tolerance values instead of increasing them. In my case, the difference is minimal even with `rtol=1e-4` and `atol=1e-4`. Can you share a minimal example where you're seeing a large difference?

PR #9326, opened 2023-01-03
Fixing the failing test np.power frontend function

hello-fri-end:
#8170

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_arithmetic_operations.py
+++ b/ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_arithmetic_operations.py
@@ -262,6 +262,7 @@ def test_numpy_multiply(
                 available_dtypes=helpers.get_dtypes("integer"),
                 num_arrays=2,
                 min_value=0,
+                max_value=7,
                 shared_dtype=True,
             )
         ],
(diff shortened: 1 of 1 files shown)


hello-fri-end:
The above commit fixes the failing test for the NumPy backend. For Tensorflow,  `Integers to negative integer powers are not allowed` error is not caught by the `ValueError` exception. Is there a way I can make sure that only positive integers are generated for the exponents?

hello-fri-end:
For JAX, the behavior is a little weird and I'm not sure how to fix it. Let me explain:

Basically, JAX handles overflow differently when an array is passed as input as opposed to a scalar. 
Simple reproducible code:
Using Jax arrays
```python
base = jnp.array([2])
exp = jnp.array([64])
```


```python
jnp.power(base, exp)
DeviceArray([1], dtype=int64)
```
 
However, if you pass a scalar, the output is 0
```python
jnp.power(2, 64)
DeviceArray(0, dtype=int64, weak_type=True)
```
Numpy, in both cases will return 0 -- hence the failing test case. What's the correct way to handle this?

hello-fri-end:
@vaatsalya123 Can you take a look please?

vaatsalya123:
Hey @hello-fri-end ,First of all really sorry for the late response to the PR!
So I had a look through the PR and I think the issue you are addressing for Jax could be resolved or understood with these following points:

1.)I think the problem could be solved if you set the ground_truth_backend to jax

2.)I think this test might need max limit for array values generated as very large exponents are leading to overflow which then leads to negative exponents.So the recommendation is to not use try except around the test function.I think if you only try and set the limit that should pass the test for jax.

Try and use some of these methods to see if they resolve the issue with jax and try to make the changes I requested in (2) thanks!

hello-fri-end:
> ght need max limit for array values generated as very large exponents are leading to overflow which then leads to negative exponents.So the recommendation is to not

Hey @vaatsalya123! Thanks for the review. I've added an upper limit for the array values as 7. All the tests are passing now :)

vaatsalya123:
lgtm!

PR #9331, opened 2023-01-03
Fixing the failing unit test of np.mod frontend function

hello-fri-end:
#8168


hello-fri-end:
> lgtm, but let's change the values to smaller ones

Updated, as suggested :+1:

PR #9333, opened 2023-01-03
Fixing np.floor_divide frontend test

hello-fri-end:


hello-fri-end:
@ahmedo42 Can you review this please?

hirwa-nshuti (changes requested):
Hey @hello-fri-end 
The reason why we didn't do `np.floor_divide` or `jnp.floor_divide` in first place is due to special cases from [array API](https://data-apis.org/array-api/latest/API_specification/generated/array_api.floor_divide.html). Doing so can lead to the failure of Array API tests.

hello-fri-end:
> Hey @hello-fri-end 

Hi @hirwa-nshuti ! Thanks for taking the time to review this PR. 

> The reason why we didn't do `np.floor_divide` or `jnp.floor_divide` in first place is due to special cases from [array API](https://data-apis.org/array-api/latest/API_specification/generated/array_api.floor_divide.html). Doing so can lead to the failure of Array API tests.

Oh, thanks for the information. I have reverted my changes and added a few `assume` blocks in the test to make sure that the function is tested on valid inputs.

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_arithmetic_operations.py
```diff
@@ -485,8 +485,13 @@ def test_numpy_floor_divide(
     fn_tree,
     on_device,
 ):
+
     input_dtypes, x, casting, dtype = dtypes_values_casting
     assume(not np.any(np.isclose(x[1], 0)))
+    assume(not np.any(np.isinf([x[1], x[0]])))
```
hirwa-nshuti, line 491: No need to filter out `inf` as this is handled inside the data generation

ivy_tests/test_ivy/test_frontends/test_numpy/test_mathematical_functions/test_arithmetic_operations.py
```diff
@@ -485,8 +485,13 @@ def test_numpy_floor_divide(
...
+    if dtype:
+        assume(not np.any(np.isclose(np.cast[dtype](x[1]), 0)))
```
hirwa-nshuti, line 493: This also is handled by the above assume block

hirwa-nshuti (changes requested):
Hey check a few comments below

hello-fri-end:
> Hey check a few comments below

Thanks for the suggestions. The tests are still failing and I'm not sure what the solution is here. It seems `np.floor_divide` does not simply return `np.floor` of `np.divide`.  Consider this minimal example:

```python
a= np.array([0.000977], dtype=float16)
b = np.array(-[32768.], dtype=float16)

>>> np.floor_divide(a,b)
-1.0

>>> np.divide(a, b)
-0.0
>>> np.floor(np.divide(a,b)).astype(a.dtype)
-0.0

```

Looks like `np.divide` already rounds the result. Maybe the difference is that `np.divide` will always return the result rounded to the nearest integer while `np.floor_divide` will always return floor of the division result. If that's the case, how can we make sure that we the same results from both?

hirwa-nshuti:
As the ivy implementation conforms to Array API standards these edge cases should not pass as Numpy fully follows the python standards and we can't go ahead and modify the backend implementation🙂

Here we can generate values that are naturally not nearly close to zero to make our tests pass.
That's the direct solution we can take now😊

hello-fri-end:
> As the ivy implementation conforms to Array API standards these edge cases should not pass as Numpy fully follows the python standards and we can't go ahead and modify the backend implementationslightly_smiling_face
> 
> Here we can generate values that are naturally not nearly close to zero to make our tests pass. That's the direct solution we can take nowblush

Thanks for the suggestions :) I've added bounds on the array values and the tests are passing for now :+1:

hirwa-nshuti:
What if we removed the limitations and use filtering for all values close to zero in inputs
`assume(not np.any(np.isclose(x[0], 0)))` and `assume(not np.any(np.isclose(x[1], 0)))`

hello-fri-end:
> assume(not np.any(np.isclose(x[1], 0)))

This seems to work for now, although, I've kept the tolerance values as 1e-1. I've pushed the changes, please review them again :)

hirwa-nshuti (approved):

PR #9399, opened 2023-01-04
Fix numpy.nanmin Frontend

mohamedehab0999:
Added a condition because it was causing an issue of not replacing inf with nan back after compute the max while putting initial with None value.


ivy-seed:

vaatsalya123:
lgtm!

PR #9426, opened 2023-01-05
Update device.py

Fayad-Alman:
Fixing direct bindings for default_device function
- Returning as_native_dev with parameter ret (ret=ret)


Fayad-Alman:
Issue:
#9319

Issue #9325, opened 2023-01-03
multi_dot


danielmunioz:

danielmunioz:
Hey @kunal266! Oh yeah It's almost done, I had forgotten to finish it to be honest, but it's just fixing the containers and arrays tests so hopefully it shouldn't take too long, I'll let you know when it's done! but if you wanna start working right away (and if it's not too complicated) you can try to implement the frontend function you're working on using a composition of functions! :)

PR #9436, opened 2023-01-06
Functional multi dot

danielmunioz:


danielmunioz:
close #9325

danielmunioz:
Hi! sorry, I've been having some issues running the tests for this function, the test is only able to run if you uncomment the `test_flags.instance_method = False` flag otherwise the test raises the error:

`ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_linalg.py:345: in test_multi_dot
    helpers.test_function(`
`ivy_tests/test_ivy/helpers/function_testing.py:249: in test_function
    ivy.prune_nest_at_index(args, instance_idx)`
`args = ([(ivy.array([[0, 0],
       [0, 0]]), ivy.array([[0, 0],
       [0, 0]]), ivy.array([[0, 0],
       [0, 0]]))], [0, 0]), kwargs = {}`

And the big one:
`ivy.exceptions.IvyBackendException: torch: prune_nest_at_index: torch: prune_nest_at_index: 'tuple' object doesn't support item deletion`

I'm not sure exactly why this happens, my guessing is that due to the nature of the inputs, the testing function somehow breaks when trying to test instance methods

danielmunioz:
It's also worth mentioning that due to the nature of the floating point arithmetic, there may be a bit of a difference in the results when running different backends, I managed to adjusts the rtol and atol test values for this, but the difference can sometimes be noticeable, as explained on [this github issue](https://github.com/numpy/numpy/issues/22953)

danielmunioz:
It's worth nothing that due to the nature of floating point arithmetic, there may be a bit of a difference in the results when using different backends, I managed to adjust the atol and rtol test values for this, but the difference can sometimes be a bit noticeable

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_linalg.py
```diff
@@ -285,3 +285,71 @@ def test_eig(
...
+)
+def test_multi_dot(
+    dtype_x,
+    test_flags,
+    backend_fw,
+    fn_name,
+    ground_truth_backend,
+):
+    dtype, x = dtype_x
+    # test_flags.instance_method = False
```
ZoeCD, line 344: If setting the instance method flag to False is a quick fix, the error may be in the array instance method or in the container one. Have you tried to use the debugger to find where the exception is exactly happening? 🤔

ZoeCD (changes requested):

Issue #9441, opened 2023-01-06
Add matplotlib as part of required dependencies in setup.py


dom-inic:
working with ivy-core==1.1.10 and getting missing dependency matplotlib. checked setup.py on ivy codebase which opens requirements.txt on the requirements folder, matplotlib is not part of the dependencies required by ivy. 
probably should be added for the next release?

dom-inic:
yeah i think it shouldn't be added for now. maybe future support for other frameworks might need it

PR #9444, opened 2023-01-06
 add matplotlib dependency

dom-inic:
In reference to this issue #9441 added matplotlib version 3.5.2 as part of the dependencies required during the installation of ivy.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -5,3 +5,4 @@ termcolor==1.1.0
 colorama==0.4.5
 packaging==21.3
 nvidia-ml-py<=11.495.46 # mod_name=pynvml
+matplotlib==3.5.2


jaketae:
I also ran into issue #9441 when installing `ivy==1.1.0`. However, I'm not sure if adding another dependency is the best fix. 

1. TF, torch, or JAX do not require it, so there's no reason Ivy should require it.
2. The file that's triggering the error has been deleted, and I can't seem to find references to `matplotlib` anymore in the codebase ([grep result](https://github.com/search?q=repo%3Aunifyai%2Fivy%20matplotlib&type=code)).

Curious to hear thoughts from the OP and the core devs!

dom-inic:
@jaketae maybe future support of a framework like scikit-learn would require it

ivy-seed:

hirwa-nshuti:
Hello @dom-inic  
Referring to the conversation we had on this issue https://github.com/unifyai/ivy/issues/9441
This dependency is unnecessary now as the development environment doesn't need it.
This means that this PR should be closed as it may not be merged.
Thanks🙂

dom-inic:
Yeah right

PR #9461, opened 2023-01-06
Added Pad to raw_ops

KareemMAX:
Close https://github.com/unifyai/ivy/issues/9393

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
+++ b/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
@@ -3087,3 +3087,52 @@ def test_tensorflow_Pack(
         values=x,
         axis=axis,
     )
+
+
+@st.composite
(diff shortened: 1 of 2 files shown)


ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
```diff
@@ -3087,3 +3087,52 @@ def test_tensorflow_Pack(
         values=x,
         axis=axis,
     )
+
+
+@st.composite
+def _pad_helper(draw, *, padding_types=None):
+    if padding_types is None:
```
CatB1t, line 3094: No need to have this implicitly set here, I'd suggest making making `padding_types` a required parameter or a fixed value, just to avoid the unnecessary overhead.

CatB1t (changes requested):
just a small change! then we're good to merge 😄

CatB1t (commented):
lgtm!

PR #9462, opened 2023-01-06
Added PadV2 to raw_ops

KareemMAX:
Closes https://github.com/unifyai/ivy/issues/9394
Please that this PR is based on https://github.com/unifyai/ivy/pull/9461 as they have common functionality


ivy/functional/backends/torch/manipulation.py
```diff
@@ -255,6 +255,7 @@ def tile(
     return x.repeat(repeats)
```
CatB1t, line 257: We shouldn't have this, as it's a false alarm, as explained in the other comment.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
```diff
@@ -3087,3 +3087,86 @@ def test_tensorflow_Pack(
...
+            available_dtypes=padding_types,
+            shape=(ndim, 2),
+            min_value=0,
+            max_value=10,
+        )
+    )
+
+    if return_constant_values:
+        _, constant_values = draw(
+            helpers.dtype_and_values(
```
CatB1t, line 3115: we simply can use the `int` strategy here, this is causing the test to fail in `PyTorch` as we're generating a `0-dim` array of type `bfloat16`, which isn't the expected input for this parameter.

CatB1t (changes requested):
Great work! @KareemMAX, just few updates and we're good to merge!

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
KareemMAX: I don't think that `int` strategy is suitable here, as we need a constant value with the same type of the original tensor, hence why I used `dtype_and_values` instead of a simple strategy while I need a simple scalar.

CatB1t: I see! completely missed this one, my bad! I'll have another look to see why it's exactly failing

CatB1t (changes requested):
Hello @KareemMAX, Could you please update the tests to use the new `test_flags` object? then we're good to merge!

CatB1t (approved):
lgtm!

PR #9487, opened 2023-01-08
Reformat creation-meshgrid

cosmicthermo:


cosmicthermo:
Close #9473

ivy/functional/ivy/creation.py
```diff
@@ -1258,7 +1258,7 @@ def meshgrid(
 
     This function conforms to the `Array API Standard
```
CatB1t, line 1259: We can actually delete this, it's not necessary to have them anymore.

ivy/functional/ivy/creation.py
```diff
@@ -1303,18 +1305,6 @@ def meshgrid(
     >>> print(yv)
     ivy.array([[4], [5], [6]])
 
-    With :class:`ivy.NativeArray` input:
```
CatB1t, line 1306: Why this has been removed? I think we can keep it 😄

ivy/container/creation.py
```diff
@@ -1,5 +1,5 @@
 # global
-from typing import Optional, Union, List, Tuple, Dict, Sequence
+from typing import Optional, Union, List, Tuple, Dict, Sequence, Any, Type
```
CatB1t, line 2: `Any, Type` doesn't seem to be used after importing here, we should undo this.

CatB1t (changes requested):
Hello @cosmicthermo! thanks for your great work, just a few points to be addressed, also could you please revert the changes to the `.idea` folder? thanks!

ivy/functional/ivy/creation.py
```diff
@@ -1283,6 +1280,8 @@ def meshgrid(
 
     >>> x = ivy.array([1, 2, 5])
     >>> y = ivy.array([4, 1])
+    >>> xv = ivy.zeros((3, 2))
```
CatB1t, line 1283: These are getting overwritten by the function return, we shouldn't have them.

CatB1t:
> Thank you so much for the review! This is actually my first PR! I am happy to finish this work and get everything right. One question about the out variable in the docstring of this function. They don't have this variable in the previous version. And in this case the output is more than one variables, which corresponds to the input variables. Do we need to add this out variable in this case? Thanks!

Hello @cosmicthermo, I'm not sure I understand, can you elaborate more?

ivy/container/creation.py
```diff
@@ -783,14 +880,90 @@ def static_meshgrid(
...
+        >>> x = ivy.Container(a=ivy.array([1, 2, 3]))
+        >>> y = ivy.Container(a=ivy.array([4, 5]))
+        >>> xv = x.meshgrid(y)
+        >>> print(xv)
+        {
+            a: ivy.array([[1, 2, 3],
+                        [1, 2, 3]]),
+            b: ivy.array([[4, 4, 4],
+                        [5, 5, 5]])
+        }
```
CatB1t, line 962: Could you add more examples including other parameters, simple ones will be sufficient 😄

CatB1t (changes requested):

PR #9513, opened 2023-01-08
ivy_zeta_fixes

nassimberrada:
Fixed corner cases resulting in either `inf` or `nan` output
@Ishticode I think this covers the discrepancy your pointed out and others, the intelligent tests seem to be passing as well, let me know if this looks okay

### DIFF
--- a/ivy/functional/backends/numpy/experimental/elementwise.py
+++ b/ivy/functional/backends/numpy/experimental/elementwise.py
@@ -337,17 +337,29 @@ def zeta(
     *,
     out: Optional[np.ndarray] = None,
 ) -> np.ndarray:
-    inf_indices = np.equal(x, 1)
-    temp = np.logical_and(np.not_equal(x, 1), np.less_equal(q, 0))
-    nan_indices = np.logical_or(temp, np.less(x,1))
+    arr_shape = q.shape
+    x, q = x.flatten(), q.flatten()
+
+    inf_indices1 = np.where(x == 1)
+    inf_indices2 = np.intersect1d(
+        np.array(np.where(q == 0)), np.array(np.where(x > 1))
+    )
+    inf_indices3 = np.intersect1d(
+        np.array(np.where(q <= -1)), np.array(np.where(x > 1))
+    )
+    nan_indices = np.where(x < 1.)
+
     n, res = 1, 1 / q**x
     while n < 10000:
         term = 1 / (q + n) ** x
         n, res = n + 1, res + term
     ret = np.round(res, decimals=4)
+    ret[inf_indices1] = np.inf
+    ret[inf_indices2] = np.inf
+    ret[inf_indices3] = np.inf
     ret[nan_indices] = np.nan
-    ret[inf_indices] = np.inf
-    return ret
+
+    return ret.reshape(arr_shape)
 
 
 zeta.support_native_out = False
--- a/ivy/functional/backends/jax/experimental/elementwise.py
+++ b/ivy/functional/backends/jax/experimental/elementwise.py
@@ -230,17 +230,30 @@ def zeta(
     *,
     out: Optional[JaxArray] = None,
 ) -> JaxArray:
-    inf_indices = jnp.equal(x, 1)
-    temp = jnp.logical_and(jnp.not_equal(x, 1), jnp.less_equal(q, 0))
-    nan_indices = jnp.logical_or(temp, jnp.less(x,1))
+    arr_shape = q.shape
+    x, q = x.flatten(), q.flatten()
+
+    inf_indices1 = jnp.where(x == 1)
+    inf_indices2 = jnp.intersect1d(
+        jnp.array(jnp.where(q == 0)), jnp.array(jnp.where(x > 1))
+    )
+    inf_indices3 = jnp.intersect1d(
+        jnp.array(jnp.where(q <= -1)), jnp.array(jnp.where(x > 1))
+    )
+    nan_indices = jnp.where(x < 1.)
+
     n, res = 1, 1 / q**x
     while n < 10000:
         term = 1 / (q + n) ** x
         n, res = n + 1, res + term
     ret = jnp.round(res, decimals=4)
+    
+    ret = ret.at[inf_indices1].set(jnp.inf)
+    ret = ret.at[inf_indices2].set(jnp.inf)
+    ret = ret.at[inf_indices3].set(jnp.inf)
     ret = ret.at[nan_indices].set(jnp.nan)
-    ret = ret.at[inf_indices].set(jnp.inf)
-    return ret
+    
+    return ret.reshape(arr_shape)
 
 
 # def gradient(


Ishticode (changes requested):
Now the examples do work but the tests seem to fail. Feel free to work on the test here or instead merge this and make a separate PR to fix the test. Maybe do let @Daniel4078 (Wang Jin) know as he was also perhaps pushing some changes to zeta.

Daniel4078:
> Now the examples do work but the tests seem to fail. Feel free to work on the test here or instead merge this and make a separate PR to fix the test. Maybe do let @Daniel4078 (Wang Jin) know as he was also perhaps pushing some changes to zeta.

I am currently fixing the zeta function. you can check the commits I made directly to the main repo about the function and see if you can find anything wrong with it. Otherwise, the tests will be passing with my fixes.

Daniel4078:
And by the way, your fix does not deal with the problem at all. Check this script I write to understand the actual behavior of tensorflow.math.zeta: import tensorflow as tf
import numpy as np

# min, max, increment
specx = [-5.0, 5.0, 0.5]
specy = [-5.0, 5.0, 0.5]

x = []
y = []
i = specx[0]
while i <= specx[1]:
    x.append(i)
    i += specx[2]
i = specy[0]
while i <= specy[1]:
    y.append(i)
    i += specy[2]

for a in x:
    for b in y:
        result = tf.math.zeta(tf.convert_to_tensor(a, dtype=tf.float32), tf.convert_to_tensor(b, dtype=tf.float32))
        result = result.numpy().tolist()
        # if np.isinf(result):
        #     print("result is inf when x="+str(a)+" and y="+str(b))
        if np.isnan(result):
            print("result is nan when x="+str(a)+" and y="+str(b))

Ishticode:
Hi @Daniel4078 Just ran the test now on latest MASTER and they have seemed to fail now with Hypotheiss quoting an example of -1, and 1.5.

here is sample code:

```python
import ivy
ivy.set_backend('torch')
print(ivy.current_backend_str())
x = ivy.array([1.5], dtype=ivy.float32)
y = ivy.array([-1.], dtype=ivy.float32)
print(ivy.zeta(x, y)) # gives ivy.array([inf])
ivy.unset_backend()

ivy.set_backend('numpy')
print(ivy.current_backend_str())
x = ivy.array([1.5], dtype=ivy.float32)
y = ivy.array([-1.], dtype=ivy.float32)
print(ivy.zeta(x, y)) # gives a ivy.array([nan])
ivy.unset_backend()
```
so torch give `inf` while np backend gives `nan`. 

However @nassimberrada 's code in this PR resolves this issue by giving `inf` for both backends.

Daniel4078:
> Hi @Daniel4078 Just ran the test now on latest MASTER and they have seemed to fail now with Hypotheiss quoting an example of -1, and 1.5.
> 
> here is sample code:
> 
> ```python
> import ivy
> ivy.set_backend('torch')
> print(ivy.current_backend_str())
> x = ivy.array([1.5], dtype=ivy.float32)
> y = ivy.array([-1.], dtype=ivy.float32)
> print(ivy.zeta(x, y)) # gives ivy.array([inf])
> ivy.unset_backend()
> 
> ivy.set_backend('numpy')
> print(ivy.current_backend_str())
> x = ivy.array([1.5], dtype=ivy.float32)
> y = ivy.array([-1.], dtype=ivy.float32)
> print(ivy.zeta(x, y)) # gives a ivy.array([nan])
> ivy.unset_backend()
> ```
> 
> so torch give `inf` while np backend gives `nan`.
> 
> However @nassimberrada 's code in this PR resolves this issue by giving `inf` for both backends.

But the ground truth is tensorflow and tensorflow.math.zeta for -1. and 1.5 returns nan

Daniel4078:
please run this `print(tf.math.zeta(tf.constant(1.5), tf.constant(-1.)))
print(torch.special.zeta(torch.tensor(1.5), torch.tensor(-1.)))`

Daniel4078:
Ok, maybe this pull request can solve this difference between torch and tensorflow, here I updated the script to help figure out the behavior of both torch and tensorflow:
```
import tensorflow as tf
import numpy as np
import torch

# min, max, increment
specx = [-3.0, 3.0, 0.5]
specy = [-3.0, 3.0, 0.5]

x = []
y = []
i = specx[0]
while i <= specx[1]:
    x.append(i)
    i += specx[2]
i = specy[0]
while i <= specy[1]:
    y.append(i)
    i += specy[2]

for a in x:
    for b in y:
        result1 = tf.math.zeta(tf.convert_to_tensor(a, dtype=tf.float32), tf.convert_to_tensor(b, dtype=tf.float32))
        result1 = result1.numpy().tolist()
        result2 = torch.special.zeta(torch.tensor(a), torch.tensor(b))
        result2 = result2.numpy().tolist()
        if np.isinf(result2):
            print("result is inf when x="+str(a)+" and y="+str(b))
        if np.isnan(result2):
            print("result is nan when x="+str(a)+" and y="+str(b))
        if result1 != result2:
            print("torch acts differently from tensorflow when x="+str(a)+" and y="+str(b)) 
```

Ishticode:
@Daniel4078 
I am not working on a fix. yes you are right in that tensorflow should be the ground truth and if that returns nan in the particular case above then we should be return nan too and not inf. What I mentioned above that Nassims solution return same result for both backends.. but if it is wrong then it is wrong 😂.
Could you and Nassim perhaps chat on this.

Thank you so much @Daniel4078 for investigating this in detail. I am not working on this and You and Nassim know better about it so I am sure you and Nassim will have this resolved soon.

ivy/functional/backends/numpy/experimental/elementwise.py
```diff
@@ -337,17 +337,29 @@ def zeta(
...
-    temp = np.logical_and(np.not_equal(x, 1), np.less_equal(q, 0))
-    nan_indices = np.logical_or(temp, np.less(x,1))
+    arr_shape = q.shape
+    x, q = x.flatten(), q.flatten()
+
+    inf_indices1 = np.where(x == 1)
+    inf_indices2 = np.intersect1d(
+        np.array(np.where(q == 0)), np.array(np.where(x > 1))
+    )
+    inf_indices3 = np.intersect1d(
```
Daniel4078, line 347: Also, this intersect1d does not work at all when the given inputs is more than 1d.

nassimberrada:
> @Daniel4078 I am not working on a fix. yes you are right in that tensorflow should be the ground truth and if that returns nan in the particular case above then we should be return nan too and not inf. What I mentioned above that Nassims solution return same result for both backends.. but if it is wrong then it is wrong joy. Could you and Nassim perhaps chat on this.
> 
> Thank you so much @Daniel4078 for investigating this in detail. I am not working on this and You and Nassim know better about it so I am sure you two will have this resolved soon.
> 
> You guys can perhaps decide if you want to work together or agree to just the one of you can take on this task. Whatever you prefer. I am just talking and doing nothing. Sorry joy

Thanks for connecting us on this, we'll try to figure it out !

ivy/functional/backends/numpy/experimental/elementwise.py
nassimberrada: Wouldn't flattening the input arrays solve that ?

Daniel4078: Then how would you preserve the structure of the input to the output?

Daniel4078: especial the case when the two inputs do not have the same shape as each other

Daniel4078: for tensorflow, the two inputs are allowed to be with same shape or having only one of them of length 1 in dimensions. like:
```
import tensorflow as tf
result1 = tf.math.zeta(tf.convert_to_tensor([[1,2],[3,4]], dtype=tf.float32), tf.convert_to_tensor([3,2], dtype=tf.float32))
print(result1)
```

nassimberrada:
@Daniel4078 i think this PR i made works for syncing the different backends in cases that should result in either `inf` or `nan` , i suspect my implementation is still failing because of the core approach i used to compute the values.

I relied on computing an approximation of the zeta function using an infinite series (not sure exactly how to transcribe that programatically so i relied on a simple for loop capped at an arbitrarily-set high value). That's the gist of my approach basically, maybe some of it can be incorporated into your script to arrive at a complete version.

ivy/functional/backends/numpy/experimental/elementwise.py
Daniel4078: which is another hard feature to implement

ivy-seed:

ivy-seed:

PR #9568, opened 2023-01-11
removing tensorflow type promotion table

sherry30:


sherry30:
I've added `check_same_dtype` in `ivy.assertions` and added this in the frontend functions.
Let me know if you prefer it some other way

PR #9591, opened 2023-01-11
Duplicate decorator in lamb_update function

dhakalaashish:
I came across a repeated decoraor in the function lamb_update
in the file
ivy/functional/ivy/gradients.py

Wouldn't using that decorator twice be a redundant? Please advise.


vaatsalya123:
Thank you so much for pointing that out your changes look lgtm!

Issue #9644, opened 2023-01-13
Softplus


KareemMAX:

KareemMAX:
This function already existed, although I'm fixing a small bug as it's not passing tests.

PR #9645, opened 2023-01-13
Fix Softplus test

KareemMAX:
Closes https://github.com/unifyai/ivy/issues/9644
This PR was intended to create the function which was already implemented but wasn't marked as done in the ToDo list, but I found a small bug and this is its fix.


CatB1t (approved):
lgtm!

PR #9646, opened 2023-01-13
Add Softmax to raw_ops

KareemMAX:
Closed https://github.com/unifyai/ivy/issues/9643


ivy/functional/frontends/tensorflow/raw_ops.py
```diff
@@ -482,6 +482,17 @@ def Relu6(features, name="Relu6"):
 )
 
 
+Softmax = to_ivy_arrays_and_back(
+    map_raw_ops_alias(
+        tf_frontend.keras.activations.softmax,
+        kwargs_to_update={
+            "logits": "x",
+            "": "axis",
```
KareemMAX, line 490: I'm not really sure if that is right, I wanted to invalidate the `axis` argument as it not needed as described in the [docs](https://www.tensorflow.org/api_docs/python/tf/raw_ops/Softmax)

KareemMAX:
Also, the corresponding tests doesn't pass in Jax and Torch when the dtype is `bfloat16` not sure if I could leave it as is or do further changes

CatB1t (approved):
LGTM! could you resolve the conflict and I'll merge right away? 😄 
regarding the `bfloat16` issue, seems to be an internal issue with testing, I'll have a look deeper!

PR #9721, opened 2023-01-16
add docstrings examples for dropout1d

xoiga123:
`black` auto-formatted `ifft` too

### DIFF
--- a/ivy/container/layers.py
+++ b/ivy/container/layers.py
@@ -350,7 +350,7 @@ class ContainerWithLayers(ContainerBase):
 
     @staticmethod
     def static_dropout1d(
-        x: ivy.Container,
+        x: Union[ivy.Array, ivy.NativeArray, ivy.Container],
         prob: float,
         /,
         *,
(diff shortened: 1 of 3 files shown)


ivy/container/layers.py
```diff
@@ -350,7 +350,7 @@ def dropout(
 
     @staticmethod
     def static_dropout1d(
-        x: ivy.Container,
+        x: Union[ivy.Array, ivy.NativeArray, ivy.Container],
```
hirwa-nshuti, line 353: This instance method only accepts `ivy.Container` as input

hirwa-nshuti (changes requested):
Hey @xoiga123 

All looks good😊
Just a few changes and the PR get merged.

Thanks

ivy/container/layers.py
xoiga123: Sorry for the late reply. I agree that your review is correct, but I've seen many conflicting examples (i.e. elementwise) where a `ivy.Container.static_` function does accept both `ivy.Array` and `ivy.Container` as a valid type hint. In `ivy.ContainerBase.cont_multi_map_in_function`, we even assert that a container must exist in the arguments. So are the type hints wrong in these examples?

hirwa-nshuti: Only the mentioned line should be changed as the function accepts one input as array and for the container. This means this static method can't accept the `ivy.Array` input and aswell `ivy.NativeArray` inputs.

hirwa-nshuti (changes requested):
Check the last comment

hirwa-nshuti (approved):
Looks good now😊

Thanks for contributing🙂

PR #9732, opened 2023-01-16
Adding ufunc class to the numpy frontend

hirwa-nshuti:
Hey @fspyridakos 

I finished the initial implementation of the `np.ufunc` class in the numpy frontend.
Waiting for your review after that we start calling the frontend functions that are instances of `np.ufunc`  using `ufunc(func_name)`. 

With the implementation below I was able to implement some of a few class methods:
- [x] __call__()
- [x] nin()
- [x] not()
- [x] nargs()

With their unit tests along, other methods associated with this class can take time to implement as they are complex 🙂. I think we can make these methods open for contribution or work on them progressively.

Thanks😊


fspyridakos:
Yup looks good, go ahead and start initialising the ufuncs in the main `__init__`, again we might need to rename the current function names to prevent name errors

PR #9762, opened 2023-01-17
Fixed array like handing by fixing problem with `handle_array_like` w…

fspyridakos:
…rapper, removed wrapper from functions with promotion as the promotion function handles array_likes separately to support weak_type promotion. Renamed wrapper to `handle_array_like_without_promotion` to improve clarity. Also added wrapper and promotion to experimental functions.

### DIFF
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -1,10 +1,11 @@
+import numbers
+
 import ivy
 import functools
 import logging
 from types import FunctionType
 from typing import Callable
 import inspect
-import typing
 
 
 # for wrapping (sequence matters)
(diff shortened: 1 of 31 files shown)


fspyridakos:
This wrapper will need to go before `integer_arrays_to_float` in order to not cause issues with tf, will fix that with another refactor in this PR

fspyridakos:
@VedPatwardhan I fixed most tests but `clip_matrix_norm` and `clip_vector_norm` are still failing. Not sure why  `clip_matrix_norm` fails but `clip_vector_norm` fails due to an overflow caused by `ivy.stable_divide` called in the function. Do you think the max value could be decreased for that test?

VedPatwardhan (commented):
Hey @fspyridakos, I think if that fixes the failure, we could change that. Apart from that whatever fails with the ivy tests, I'd be happy to fix because I suppose if a couple of ivy tests fails, generally there's a few more failures we may not be able to find right now 🙂

PR #9765, opened 2023-01-17
added dynamic_backend feature to the ivy.Array and ivy.Container class

YushaArif99:


YushaArif99:
Hi @djl11. 

Could you kindly review the changes? 
Also, I have currently made small changes to the `module.py` and `converter.py` file to handle the `dynamic_backend` feature. Let me know if this is the correct way to handle it or if I should change the logic. The reason I have set the default value as `False` is because there are other references to the `ivy_module.v` which cause unwanted side effects. That is, even if I set `ivy_module.v.dynamic_backend=False`, and change the backend, the variables also get converted because there are other Container variables that hold references to the `ivy_module.v` which have the default `True` value for the `dynamic_backend` attribute. This is why I have set the default value for those other references as `False`.
The unit tests still need to be added. Will add those shortly.

ivy/backend_handler.py
```diff
@@ -46,9 +48,16 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 _backend_reverse_dict["ivy.functional.backends.torch"] = "torch"
 
 
+_static_backend_dict = dict()
 # Backend Getting/Setting #
 # ----------------------- #
```
YushaArif99, line 54: Added this function to basically pre-load all ivy backends. This is later required in the `dynamic_backend` setter functions which basically use `determine_backend_from_args` to determine the backend of the input.

ivy/backend_handler.py
```diff
@@ -46,9 +48,16 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
 
+def _load_static_backends():
+    _static_backend_dict["ivy.functional.backends.numpy"] = importlib.import_module("ivy.functional.backends.numpy")
+    _static_backend_dict["ivy.functional.backends.torch"] = importlib.import_module("ivy.functional.backends.torch")
+    _static_backend_dict["ivy.functional.backends.jax"] = importlib.import_module("ivy.functional.backends.jax")
+    _static_backend_dict["ivy.functional.backends.tensorflow"] = importlib.import_module("ivy.functional.backends.tensorflow")
+
 
 def _determine_backend_from_args(args):
```
YushaArif99, line 62: made changes to `determine_backend_from_args` since it wasn't correctly handling `dict/ivy.Containers`.

ivy/backend_handler.py
```diff
@@ -241,6 +252,58 @@ def set_backend(backend: str):
...
+
+    variable_ids = set()  # create an empty set to store variable object ids
+    intermediate_objs = [] # create an empty list to store intermediate objects created during conversion
+    def _is_var(obj):
+
+        if isinstance(obj, ivy.Container):
+            def _map_fn(x):
+
+                x = x.data if isinstance(x, ivy.Array) else x
+                if x.__class__.__module__ in ("numpy", "jax.interpreters.xla","jaxlib.xla_extension"):
```
YushaArif99, line 270: Adding this check here because `is_variable()` returns `True` for numpy and JAX arrays. Basically, these 2 frameworks don't have the concept of `stateful` tensors and so they should technically not be considered as *variables*

This does mean that if you are converting variables from `TF->JAX/Numpy->Torch` for example, the variables would not get preserved as the middle backend would basically convert them to normal native arrays.

Is this the correct behavior?

ivy/container/base.py
```diff
@@ -3989,6 +3992,58 @@ def __setitem__(self, query, val):
...
+        if query == "_backend":
+            self._backend = val
+            return
+
+        if query == "dynamic_backend":
+            from ivy.functional.ivy.gradients import _variable
+            from ivy.backend_handler import _determine_backend_from_args
+
+            if val == False:
+                self._backend = _determine_backend_from_args(self)
```
YushaArif99, line 4009: Using `determine_backend_from_args` to determine the backend for the instance. This backend is then used to invoke the backend functions directly (i.e bypassing Ivy's API). 
Need to do this because I cant really use Ivy functions due to the `global dict` issue

ivy/array/array.py
```diff
@@ -135,10 +135,47 @@ def _init(self, data):
...
     # Properties #
     # ---------- #
 
+    @property
+    def dynamic_backend(self):
+        return self._dynamic_backend
+
+    @dynamic_backend.setter
+    def dynamic_backend(self, value):
+        from ivy.functional.ivy.gradients import _variable
```
VedPatwardhan, line 152: Through this logic, we'd be importing the `_variable` and `_determine_backend_from_args` functions every time we set a `dynamic_backend`, whereas it would be much better if we just import them once at the top, unless there's an issue with circular imports if you do that (which I think is a possibility)

ivy/array/array.py
```diff
@@ -135,10 +135,47 @@ def _init(self, data):
...
+            variable_data = self._backend.variable_data
+
+            if is_variable(self.data) and \
+                    not ( str(self._backend).__contains__("jax") or
+                          str(self._backend).__contains__("numpy")
+                    ):
+                native_data = variable_data(self.data)
+                np_data = to_numpy(native_data)
+                new_arr = ivy.array(np_data)
+                self._data = _variable(new_arr.data)
```
VedPatwardhan, line 170: I think `_variable` returns an `ivy.Array` even if we pass `new_arr.data` to it, so you'd have to convert it back to native before storing `self._data`.

ivy/backend_handler.py
VedPatwardhan: Hey @YushaArif99, I think this would add `jax`, `tensorflow` and `torch` as dependencies to the backend handler and in turn to the project, whereas users should be able to instead just install the libraries they wish to work with. I'd suggest looking into ways to check if a library is installed and only pre-loading the backends whose framework is installed in this function

ivy/backend_handler.py
VedPatwardhan: I suppose we refactor this implementation to make it more cleaner and use a single loop for a `container` or a `list ` or a `tuple` (something resembling the `nest` functions), but we can look into that once the previously mentioned issues are fixed given that the logic works well

ivy/backend_handler.py
```diff
@@ -217,8 +238,118 @@ def current_backend(*args, **kwargs):
     return importlib.import_module(_backend_dict[implicit_backend])
 
 
-def set_backend(backend: str):
+variable_ids = set()  # create an empty set to store variable object ids
```
VedPatwardhan, line 241: Is there a way to not use global variables here? Happy to discuss in detail!

ivy/container/base.py
VedPatwardhan: A similar suggestion to the local imports with the `ivy.Array` class

ivy_tests/test_ivy/test_misc/test_dynamic_backend.py
```diff
@@ -0,0 +1,151 @@
+import ivy
```
VedPatwardhan, line 1: Unless there are some issues with the globals, I think we could have these tests inside `test_backend_handler.py` instead

VedPatwardhan (changes requested):
Hey @YushaArif99, just suggested a few changes, happy to discuss in detail! Thanks 🙂

ivy/backend_handler.py
YushaArif99: Yeah, this logic is not needed anymore actually. I added it because there were some import issues I was facing when using `_determine_backend_from_args`. But they don't seem to be showing up now so I have removed this static_backend logic completely.

ivy/array/array.py
YushaArif99: Yup there's an issue with circular imports which is why I cant import it at the top

ivy/array/array.py
YushaArif99: In that case something like `_variable(new_arr).data` should work right?

ivy_tests/test_ivy/test_misc/test_dynamic_backend.py
YushaArif99: Yeah no worries. I have moved the tests in the `test_backend_handler.py`

ivy/__init__.py
```diff
@@ -1,6 +1,8 @@
 # global
 from types import SimpleNamespace
 import warnings
+
+import ivy
```
VedPatwardhan, line 5: I think this import may be getting used in one of the previous commits and is unused currently, so we could remove it

ivy/array/array.py
VedPatwardhan: Got it! Thanks for clearing it out!

ivy/backend_handler.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 from ivy import verbosity
 from typing import Optional
+import gc
```
VedPatwardhan, line 7: I think this is another unused import that can be removed

VedPatwardhan (changes requested):
Hey @YushaArif99, just a couple more changes, I'll add the `Exhaustive CI` label to the PR just to be sure that there aren't many failures after which I think we could merge the PR. Thanks 🙂

ivy/__init__.py
YushaArif99: Yeah not sure why this is there. Might have added it before for some reason 😅

ivy/backend_handler.py
YushaArif99: Are you referring to the import of `gc` module? It's basically being used [here](https://github.com/unifyai/ivy/blob/0c8d88b6d238509033b3298966e1450e506d4c3e/ivy/backend_handler.py#L273)

YushaArif99:
Hey @VedPatwardhan 

Thanks for taking such a meticulous look at the PR 😂.

With regards to the CI tests, the intelligent tests seem to be running the frontend tests for `view_tensor` for some reason which is why a lot of the tests are failing. The tests regarding `dynamic_backend` (`run_tests_1`) are all passing. 

I'm not sure if these other failing tests have something to do with the PR itself. Could you also kindly take a look and confirm that is the case. 

Thanks! 🙂

ivy/backend_handler.py
VedPatwardhan: Oh right, yeah thanks for clearing it out!

VedPatwardhan (approved):
I think the view tensor issue has been identified before and is being worked on by the team, so we could merge the PR provided the other tests don't fail 🙂

PR #9793, opened 2023-01-18
Close #9698

### LINKED ISSUE #9698: shuffle

Mmorad732:
Created Tensorflow shuffle function but I'm not sure why does it fail the testing


CatB1t (changes requested):
Hello @Mmorad732 , seems like you're overriding the current functions, could you please revert those, and add your function and the test, we shouldn't also update the backend module, unless it's implemented incorrectly, thanks!

PR #9887, opened 2023-01-20
tf.cond

### LINKED ISSUE #9862: cond

norman135:
close #9862


ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py
```diff
@@ -344,6 +344,45 @@ def test_tensorflow_convert_to_tensor(
...
+    *,
+    dtype_and_x,
+    pred_cond,
+    test_flags,
+    on_device,
+    fn_tree,
+    frontend,
+):
+    def _test_true_fn():
+        return list
```
kurshakuz, line 368: The implementation is passing tests, but can you tell me why are you returning `list` in both instances? For example if I change your implementation to `if not pred:`, result will be the same.

norman135: It's because the docs state that both true_fn and false_fn must return lists of output tensors: https://www.tensorflow.org/api_docs/python/tf/cond#:~:text=true_fn%20and%20false_fn%20both%20return%20lists%20of%20output%20tensors.%20true_fn%20and%20false_fn%20must%20have%20the%20same%20non%2Dzero%20number%20and%20type%20of%20outputs.So I thought the test fn functions could both return lists. However I could be wrong, so I appreciate any other insight on the best way to do it.

norman135:
@kurshakuz I have modified the test_true_fn and test_false_fn to return different results.✔️

kurshakuz:
Hey @norman135! Your implementation is still not correct as it always returns the same value that does not depend on the input. Please check this https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_control_flow_operators.py#L19 implementation of the cond operator on the jax frontend as your function should have similar logic. Thanks 🙂

kurshakuz (changes requested):
Hey @norman135! Your implementation is still not correct as it always returns the same value that does not depend on the input. Please check this https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_control_flow_operators.py#L19 implementation of the cond operator on the jax frontend as your function should have similar logic. Thanks 🙂

PR #9935, opened 2023-01-23
Updated expand to handle negative input dimensions and updated shape …

agniiyer:
…to not return a string

### DIFF
--- a/ivy/functional/frontends/torch/tensor.py
+++ b/ivy/functional/frontends/torch/tensor.py
@@ -35,7 +35,7 @@ class Tensor:
 
     @property
     def shape(self):
-        return "torch.Size(" + str(list(self._ivy_array.shape)) + ")"
+        return self._ivy_array.shape
 
     # Setters #
     # --------#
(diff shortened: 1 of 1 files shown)


ivy/functional/frontends/torch/tensor.py
```diff
@@ -350,6 +350,12 @@ def view_as(self, other):
         return self.view(other.shape)
 
     def expand(self, *sizes):
+
+        sizes = tuple(sizes)
```
fspyridakos, line 354: Make this `sizes = list(sizes)` as we had before

ivy/functional/frontends/torch/tensor.py
```diff
@@ -350,6 +350,12 @@ def view_as(self, other):
...
+        for i, dim in enumerate(sizes):
+            if dim < 0:
+                sizes[i] = self.shape[i]
+
         return torch_frontend.tensor(ivy.broadcast_to(self._ivy_array, shape=sizes))
```
fspyridakos, line 359: Here have `shape=tuple(sizes)`

fspyridakos (changes requested):
This should make it work properly

PR #10098, opened 2023-01-25
frontends.torch.adjoint

### LINKED ISSUE #10097: adjoint

xoiga123:
Close #10097 

I couldn't flatten `frontend_ret` in `test_frontend_function` due to the following error:
> Can't call numpy() on Tensor that has conjugate bit set. Use tensor.resolve_conj().numpy() instead.

So I had to return raw `frontend_ret` first.

### DIFF
--- a/ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
+++ b/ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
@@ -4,6 +4,11 @@ from ivy.functional.frontends.torch.func_wrapper import to_ivy_arrays_and_back
 
 
 @to_ivy_arrays_and_back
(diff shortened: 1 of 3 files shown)


xoiga123:
@hirwa-nshuti I think lots of tests are failing because I changed the `test_frontend_function`, which triggered all the failing tests to run. When I rerun those failing tests locally with the master branch, they are also failing with the same error.

hirwa-nshuti (changes requested):
Hey @xoiga123 
The failure was caused by a bug in the torch backend for `ivy.to_numpy`
I resolved it you can write your unit tests as normal no need to change the helper function.

hirwa-nshuti (approved):
Looks good now😊
Thanks for contributing

PR #10145, opened 2023-01-27
backends.torch.to_numpy: add resolve_neg

xoiga123:
We added `resolve_conj`, so let's add `resolve_neg` too.

And remove the conditional check because `torch.resolve_` already does.

### DIFF
--- a/ivy/functional/backends/torch/general.py
+++ b/ivy/functional/backends/torch/general.py
@@ -82,8 +82,7 @@ def to_numpy(
         else:
             return x
     elif torch.is_tensor(x):
-        if x.is_conj():
-            x = x.resolve_conj()
+        x = x.resolve_neg().resolve_conj()
         if copy:
             if x.dtype is torch.bfloat16:
                 default_dtype = ivy.default_float_dtype(as_native=True)


xoiga123:
@hirwa-nshuti do you mind checking this out? It's related to my previous `adjoint` PR

ivy/functional/backends/torch/general.py
```diff
@@ -82,8 +82,7 @@ def to_numpy(
         else:
             return x
     elif torch.is_tensor(x):
-        if x.is_conj():
-            x = x.resolve_conj()
+        x = x.resolve_neg().resolve_conj()
```
hirwa-nshuti, line 85: Yea we can use `resolve_neg` but the condition should stay as we are doing this only when the conjugate bit it set to `True`

hirwa-nshuti (changes requested):
Check the comment

ivy/functional/backends/torch/general.py
hirwa-nshuti: Anyway `resolve_conj` does all the job. I don't see any reason to stack it with `resolve_neg`

```
import torch

x = torch.tensor([-1 + 1j, -2 + 2j, 3 - 3j])
y = x.conj()

z = y.resolve_conj()
print(z.is_neg()) # False
print(y.imag.is_neg()) # True
print(z.imag.is_neg()) # False

```

xoiga123: > Yea we can use `resolve_neg` but the condition should stay as we are doing this only when the conjugate bit it set to `True`

Yes I think the condition helps readability, but torch already checks for the conjugate/negative bit inside their `resolve_`:
> Returns a new tensor with materialized conjugation/negative if input’s conjugate/negative bit is set to True, else returns input.

xoiga123:
```python
import torch

x = torch.tensor([-1 + 1j, -2 + 2j, 3 - 3j])
y = x.conj()

z = y.imag
print(z.is_neg()) # True
print(z.is_conj()) # False
# This implies that conj and neg are 2 different bits

z_np = z.numpy() # Error
z_np = z.resolve_conj().numpy() # Error
z_np = z.resolve_neg().numpy() # OK
z_np = z.resolve_neg().resolve_conj().numpy() # OK

y_np = y.resolve_neg().resolve_conj().numpy() # OK
```
Please check this example, I think we have to include `resolve_neg` for `.imag`

hirwa-nshuti: The reason why you're getting these results is that you're only operating on the `imaginary` part of the complex number try to work on the whole conjugated complex number you'll observe that it is all handled by `resolve_conj`
Also, the use of the conditional is because we are not only working with `complex` dtypes with that function so it can be better to specify the base condition that needs to be resolved.

hirwa-nshuti:
```python


import torch

x = torch.tensor([-1 + 1j, -2 + 2j, 3 - 3j])
y = x.conj()

z = y
print(z.is_neg()) # True
print(z.is_conj()) # False
# This implies that conj and neg are 2 different bits
# supposing w to be resolve
w = y.resolve_conj()
print(w.imag.is_neg())
print(w.imag.is_conj())

# if you operate on a whole complex Tensor
# resolve_conj() will also set the conj bit and neg bit to False

```

xoiga123: I thought someone may export the imaginary part to numpy and do smth, so I included it as a fail-safe. And I agree with keeping the conditional too, so you can close the merge if we don't need `resolve_neg`

hirwa-nshuti: In addition, the imaginary part itself is not even a conjugate that's why the resolve conjugate doesn't work.
As it is a newly introduced tensor. Would you add another condition to check if the negative bit is True?
I'll merge it😊

xoiga123: @hirwa-nshuti thanks for merging! Sorry I couldn't add the other condition before merging though, I didn't notice your comment :sob: Should I create another PR?

hirwa-nshuti: It's fine let's keep it this way

PR #10160, opened 2023-01-28
add backend implementation for lexsort

Daniel4078:
deal with this trello task: https://trello.com/c/b3z2X1wW/1174-complete-all-four-missing-backend-implementations-for-ivylexsort-function


Daniel4078:
currently need help implementing lexsort for tensorflow and torch as there is no origenal counterpart in those frameworks. I have done some research about that here are my findings, please take a look: https://discuss.pytorch.org/t/numpy-lexsort-equivalent-in-pytorch/47850/4

VedPatwardhan (commented):
Hey @Daniel4078, thanks for requesting the review. I think you'll have to implement `lexsort` for `tensorflow` and `torch` using their existing functions like `torch.unique`, `tf.unique` or their sort functions 🙂

Daniel4078:
@VedPatwardhan Thank you! If you have any idea about where can I start recreating the behavior of lexsort in tensorflow or torch, please give me a hint. I will continue research about it myself.

VedPatwardhan:
Hey @Daniel4078, the PyTorch discussion you've shared does mention an approach that involves using `torch.unique`, I suppose you could also implement the `tensorflow` function using `tf.unique`. 
Happy to answer any other questions 🙂

Daniel4078:
@VedPatwardhan Can you explain how we handle nestable inputs? Like do our function recursively unwraps high dimentional input and run the actural calculation on each of the parts before combining them back?

VedPatwardhan:
Hey @Daniel4078, so the `handle_nestable` decorator basically calls the function for every array at the leaves of the container passed as input. Feel free to refer the [docs](https://lets-unify.ai/ivy/deep_dive/function_wrapping.html#function-wrapping) 🙂

Daniel4078:
@VedPatwardhan I have encountered this strange error:
E   ivy.exceptions.IvyBackendException: torch: lexsort: argsort() received an invalid combination of arguments - got (Tensor, stable=bool, dim=int), but expected one of:
E    * (Tensor input, int dim, bool descending)
E         didn't match because some of the keywords were incorrect: stable
E    * (Tensor input, name dim, bool descending)
E         didn't match because some of the keywords were incorrect: stable

But according to the document https://pytorch.org/docs/stable/generated/torch.argsort.html, torch.argsort does have the stable argument. What happened here?

Daniel4078:
@VedPatwardhan I need help for this one!
__________ test_lexsort[cpu-ivy.functional.backends.jax-False-False] ___________
ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_sorting.py:45: in test_lexsort
    fn_tree="functional.ivy.experimental.lexsort",
ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_sorting.py:64: in test_lexsort
    helpers.test_function(
ivy_tests/test_ivy/helpers/function_testing.py:282: in test_function
    instance.__getattribute__(fn_name), *args, **kwargs
E   AttributeError: 'Array' object has no attribute 'lexsort'
E   Falsifying example: test_lexsort(
E       dtype_x_axis=(['int8'], [array([0], dtype=int8)], 0),
E       fn_name='lexsort',
E       test_flags=num_positional_args=1. with_out=False. instance_method=True. native_arrays=[False]. container=[False]. as_variable=[False]. test_gradients=False.,
E       backend_fw=<module 'ivy.functional.backends.jax' from '/ivy/ivy/functional/backends/jax/__init__.py'>,
E       on_device='cpu',
E   )
E   
E   You can reproduce this example by temporarily adding @reproduce_failure('6.55.0', b'AXicY2BkgAMQEwAAIwAD') as a decorator on your test case

I do not really get the cause of this error and what should I do in this case? Do I need to implement the instance counterpart of lexsort for both array and container?

VedPatwardhan:
> @VedPatwardhan I need help for this one! __________ test_lexsort[cpu-ivy.functional.backends.jax-False-False] ___________ ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_sorting.py:45: in test_lexsort fn_tree="functional.ivy.experimental.lexsort", ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_sorting.py:64: in test_lexsort helpers.test_function( ivy_tests/test_ivy/helpers/function_testing.py:282: in test_function instance.**getattribute**(fn_name), *args, **kwargs E AttributeError: 'Array' object has no attribute 'lexsort' E Falsifying example: test_lexsort( E dtype_x_axis=(['int8'], [array([0], dtype=int8)], 0), E fn_name='lexsort', E test_flags=num_positional_args=1. with_out=False. instance_method=True. native_arrays=[False]. container=[False]. as_variable=[False]. test_gradients=False., E backend_fw=<module 'ivy.functional.backends.jax' from '/ivy/ivy/functional/backends/jax/**init**.py'>, E on_device='cpu', E ) E E You can reproduce this example by temporarily adding @reproduce_failure('6.55.0', b'AXicY2BkgAMQEwAAIwAD') as a decorator on your test case
> 
> I do not really get the cause of this error and what should I do in this case? Do I need to implement the instance counterpart of lexsort for both array and container?

Hey @Daniel4078, yes exactly! You;ll have to implement the array and container instance and static methods for the `lexsort` function 🙂

Daniel4078:
@VedPatwardhan I have a question, is it a must to have an out argument even if none of your implementation can actually deal with it?

VedPatwardhan:
> @VedPatwardhan I have a question, is it a must to have an out argument even if none of your implementation can actually deal with it?

Yes @Daniel4078, that's because even if none of the backends natively support the `out` argument, we have a `handle_out_argument` decorator that would ensure identical references for the returned array and the `out` array. So you'll probably have to add that decorator to the main `lexsort` function 🙂

Daniel4078:
@VedPatwardhan Can you remind me about how to get unsupported datatypes for experimental functions? Thanks.

VedPatwardhan:
> @VedPatwardhan Can you remind me about how to get unsupported datatypes for experimental functions? Thanks.

Hey @Daniel4078, if you want to flag some datatype as unsupported for a backend-implementation of function, you can add the [`with_unsupported_dtypes`](https://lets-unify.ai/ivy/deep_dive/data_types.html#supported-and-unsupported-data-types) decorator. If you want to find the currently flagged unsupported dtypes you could use the `ivy.function_unsupported_dtypes` or `ivy.function_unsupported_devices_and_dtypes`(to get device-wise unsupported dtypes) 🙂

Daniel4078:
@VedPatwardhan I met a strange error reguarding datatypes, when the input is int8, the numpy would return int64 while the tensorflow would return int32. What should I do to resolve this different?

VedPatwardhan:
> @VedPatwardhan I met a strange error reguarding datatypes, when the input is int8, the numpy would return int64 while the tensorflow would return int32. What should I do to resolve this different? Should I manually cast result to be the same as input for all frameworks?

I'd suggest going through the implementation of the function and checking where the types are getting promoted, ideally the result should have the same dtype as the input, so both of those backends are returning incorrect results 🙂

Daniel4078:
@VedPatwardhan what is happening here? The test(https://github.com/unifyai/ivy/actions/runs/4128285191/jobs/7132549124) just stucks to be failing. What does this(Error: The operation was canceled.) mean?

VedPatwardhan:
> @VedPatwardhan what is happening here(https://github.com/unifyai/ivy/actions/runs/4128285191/jobs/7132549124)? What does this(Error: The operation was canceled.) mean?

I think that probably means there were some changes to the CI while this test was triggered on your PR 🙂

ivy/functional/backends/tensorflow/experimental/sorting.py
```diff
@@ -11,3 +11,29 @@ def msort(
...
+        return tf.convert_to_tensor([0])
+    result = tf.argsort(keys[0], axis=axis, stable=True)
+    if shape[0] == 1:
+        return result
+    for i in range(1, shape[0]):
+        key = keys[i]
+        ind = tf.gather(key, result)
+        temp = tf.argsort(ind, axis=axis, stable=True)
+        result = tf.gather(result, temp)
+    return tf.cast(result, tf.int64)
```
Daniel4078, line 39: here I cast all outputs to be int64 just like what numpy will do, but according to the test(https://github.com/unifyai/ivy/actions/runs/4137025627/jobs/7151674282), it still somehow returns int32. How is this possible? @VedPatwardhan

VedPatwardhan: Actually, there's a number of return statements before this one, and the casting is not applied to those (e.g. line 26, 30, 33). Also, I'd suggest having the result with the same dtypes as the input, rather than flat out type casting to `int64` 🙂

Daniel4078: Well, this function takes keys as input and always return integer indexes that can be used to sort the given keys, because of this, the return type will always be int64 as it represents position and is not influenced by the origenal type of the inputs but only their relative differences.

VedPatwardhan (approved):
Feel free to merge if the tests pass. The changes look good to me. Thanks 🙂

PR #10166, opened 2023-01-29
frontends.torch.dsplit

### LINKED ISSUE #10146: dsplit

xoiga123:
Close #10146 

1. Add `frontends.torch.dsplit` and `test_torch_dsplit`.
2. Fixed type hints for `dsplit` in backends.
3. Removed `out` argument from `dsplit` in backends (it returns a list of arrays).
4. Fixed `dsplit` docstring examples.


xoiga123:
@fspyridakos somehow because I removed `out` from `test_dsplit`, all the failing tests in `experimental/core/manipulation` are triggered. Before I committed 762153d067e78cac78cbe9097cc5732d5393490e, only `test_dsplit` is failing because of `out`, I can rewind the commit for u to check more easily. I don't think we can support out though, I manually tried and it didn't work.

ivy/functional/backends/torch/experimental/manipulation.py
```diff
@@ -156,11 +156,9 @@ def vsplit(
...
 def dsplit(
     ary: torch.Tensor,
-    indices_or_sections: Union[int, Tuple[int]],
+    indices_or_sections: Union[int, Tuple[int, ...]],
     /,
-    *,
-    out: Optional[torch.Tensor] = None,
-) -> torch.Tensor:
+) -> Tuple[torch.Tensor, ...]:
     return torch.dsplit(ary, indices_or_sections)
```
fspyridakos, line 162: Output should be converted to list to align with the rest of the backends

fspyridakos (changes requested):
Many thanks for your contribution! Looks pretty good apart from my one comment. Probably should rename the output typestrings to `List` instead of `Sequence`

ivy/functional/backends/torch/experimental/manipulation.py
xoiga123: Thanks, updated it

PR #10175, opened 2023-01-29
implement tf.where in tensorflow frontend

ZZYSonny:
This PR implements `tf.where` in the tensorflow frontend.

Workaround note:
`helpers.get_dtypes("bool")` seems to include complex types, instead of just boolean types. Once this is fixed, line 1292-1293 can be removed.

Usage note:
The function `tf.where` has two modes. When x,y is not supplied, condition can be some numeric type (subject to support in the framework). When x,y is supplied, `condition` can only be a boolean tensor.

Looking forward to hearing any feedback.


WilliamHirst:
Hi,
The contribution looks good. As for the workarounds, this might be changed later, but as is now it passes all test!

PR #10197, opened 2023-01-30
Added a percentile and quantile function for the numpy frontend order…

arananDev:
…statisitics folder. I am not sure of two things: 1. Would I need to implement the overwrite input function? would that be redundant as we need to output everything as an ivy array? 2. For the method arguement there are 8 different methods in the numpy documentation but only 3 in the Ivy, meaning there probable needs to be more methods added to the ivy function. Does that need to be implemented now?


nassimberrada:
Hi @arananDev, thank you for your contribution ! I see that your PR tries to implement two frontend functions but it would be good to work on a single function for further reference. I suggest we start with the `quantile` function, you are totally free and welcomed to work on the other once you're done !

Once you've narrowed down your PR, could you create an issue in the corresponding To-Do list so that you can reserve it, then link your PR to the issue ? Feel free to take a look at [this section](https://lets-unify.ai/ivy/contributing/the_basics.html) of the docs for more details.

Regarding the content of your PR, we would like you to implement the tests for the function along with the front-end implementation. Again you can find more details in [this section](https://lets-unify.ai/ivy/deep_dive/ivy_frontends_tests.html) of the docs. You can also take a look at other tests in the codebase (especially for statistical functions) to get some idea as well.

Concerning your questions:
- As you say there is no need to introduce the `overwrite_input` behavior since we want to output ivy arrays
- I would suggest you stick to what the functional implementation offers, it would be easier then to add the remaining methods to the functional version directly if needed.

Finally, a small comment regarding your code. I believe the `interpolation` argument in the returned `ivy.quantile` should be equal to `method` to reflect the argument specified in the front-end.

Hope this helps clarify some things, let me know if you have other questions. Otherwise feel free to ping me when you've made these changes and made sure all the tests pass for your implementation. Thanks !

arananDev:
Hey Nassim,

I was wondering if you checked on the pull request for the testing?

Thanks
Aranan Wijayasooria

PR #10243, opened 2023-02-01
Fix raw_ops.test_tensorflow_AddN

mohamedehab0999:
Closes https://github.com/unifyai/ivy/issues/9079


vaatsalya123:
Hey @mohamedehab0999 could you explain to me more on what this change is doing and how it is fixing the failing test just a little curious thanks!

mohamedehab0999:
Hey @vaatsalya123, This function returns the sum of `axis=0` of giving `inputs as one array`
As mentioned here:
https://www.tensorflow.org/api_docs/python/tf/math/add_n
https://www.tensorflow.org/api_docs/python/tf/raw_ops/AddN

So, We can't use `ivy.sum` in the function without specifying the axis parameter because its default `axis=None`.
https://github.com/unifyai/ivy/blob/71231829be9dc91a8fb5f8cac13f7ff744999f14/ivy/functional/frontends/tensorflow/raw_ops.py#L16
Also,We can't call function with multiple arrays in `inputs` parameter but we should spasify `inputs` parameter as an array of arrays with same shape to be valid.
https://github.com/unifyai/ivy/blob/71231829be9dc91a8fb5f8cac13f7ff744999f14/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py#L1509
`AddN(a, b) -> AddN([ a, b ])`

vaatsalya123:
Yeah OK thanks for the information and provided what you explained the changes look sensible will merge this!

PR #10259, opened 2023-02-01
M1 support for doc-builder

ZiadAmerr:
- Making docs file should now run on m1 mac users without errors.
- Fixed TensorFlow not found error


ZiadAmerr:
.

Issue #3607, opened 2022-08-25
Add Instance Methods to NumPy Frontend


jkeane508:
Add [Instance Methods](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html#) to NumPy Frontend:

- [ ] all
- [ ] any
- [ ] argmax
- [ ] argmin
- [ ] argpartition
- [ ] argsort
- [ ] astype
- [ ] byteswap
- [ ] choose
- [ ] clip
- [ ] compress
- [ ] conj
- [ ] conjugate
- [ ] copy
- [ ] cumprod
- [ ] cumsum
- [ ] diagonal
- [ ] fill
- [ ] flatten
- [ ] getfield
- [ ] item
- [ ] itemset
- [ ] max
- [ ] mean
- [ ] min
- [ ] newbyteorder
- [ ] nonzero
- [ ] partition
- [ ] prod
- [ ] ptp
- [ ] put
- [ ] ravel
- [ ] repeat
- [ ] reshape
- [ ] resize
- [ ] round
- [ ] searchsorted
- [ ] setfield
- [ ] sort
- [ ] squeeze
- [ ] std
- [ ] sum
- [ ] swapaxes
- [ ] take
- [ ] tobytes
- [ ] tofile
- [ ] tolist
- [ ] tostring
- [ ] trace
- [ ] transpose
- [ ] var
- [ ] view

NicAOB97:
- [ ] #4715

jkeane508:
> Hi! So each instance method needs to be added to: ivy/functional/frontends/numpy/ndarray.py

Yep, as documented [here](https://lets-unify.ai/ivy/deep_dive/16_ivy_frontends.html#classes-and-instance-methods) :)

hmahmood24:
- [] #4886

hmahmood24:
- [ ] #4887

westrany:
argmax #4924

varshithamanasa:
#4940

morleythomas:
- [ ] #5038

alizahidraja:
- [ ] #5304

geeythree:
- [ ] #5641

Shajeeran:
#6002

EvanMarie:
argmin #6239

codeamic:
#6242

EvanMarie:
[...]

PR #10277, opened 2023-02-03
Close #10179

### LINKED ISSUE #10179: choose

sergio-verduzco:
Hello guys. I'm submitting an almost complete solution, but I feel we should discuss some things before going forward.  

* My subtask, chosen form https://github.com/unifyai/ivy/issues/3607 has still not been assigned by the Ivy team, but it's been 4 days since I reserved it (https://github.com/unifyai/ivy/issues/10179), so I'm hoping it's OK to send the PR.
* I made a relatively "heavy" compositional implementation. This was because numpy.choose relies on Numpy's broadcasting rules, which seem different from the ones currently used in Ivy (e.g. `ivy.broadcast_to`). I thus added two auxiliary functions to perform array broadcasting as in Numpy. I think it's best if you guys take a look before I continue further.
* Numpy's choose does not support type promotion; it turns everything into integers. I removed this limitation. Type promotion is not performed when the argument `out` is not None
* When mode='wrap' my tests fail on torch and tensorflow. The error happens in `ivy.fmod`, apparently because the backends do not support the `int64` data type.
* I have not yet added the tests to the codebase. I have not managed to figure out how to use `@handle_frontend_tests` so: 1) the shape of the index and choices arrays make sense, and 2) the tests are performed for the 3 modes ('raise', 'wrap', and 'clip'). A little guidance would be appreciated.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -16,7 +16,8 @@
 	// "forwardPorts": [],
 
 	// Uncomment the next line to run commands after the container is created - for example installing curl.
-	"postCreateCommand": "bash .devcontainer/post_create_commands.sh"
+	"postCreateCommand": "bash .devcontainer/post_create_commands.sh",
+	"postStartCommand": "git config --global --add safe.directory ${containerWorkspaceFolder}"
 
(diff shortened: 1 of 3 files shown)


ivy-seed:

PR #10319, opened 2023-02-05
Add blackman_window

RicMat:


RicMat:
Close #10318

RicMat:
I've asked this on the 'contributing' group by I received no answer. Blackman window is not yet available for tensorflow. Shall I implement it? Shall I skip the tests? If so, how?

YushaArif99 (changes requested):
Hi @RicMat 

Extremely sorry for the late response. I was busy with exams for the past few weeks and so didn't get the time to review PRs.  Have a whole pile of pending PRs that I have to review 😅 Will try blasting through them by today hopefully 🤞🏼 

With regards to your PR, I think the overall changes do look good to me. However, I have noticed that you haven't added the implementation of `blackman_window` for the TF backend. 

If the function does not exist in the framework itself, you might need to create a compositional implementation (composed of other Ivy and/or TF functions).  The CI tests are failing currently because of this. 

Can you kindly add this implementation and re-request a review 🙂

YushaArif99:
> 

I just read your comment and realized you've already mentioned this 😅. Well, in this case, I'd suggest trying to come up with a compositional implementation. If that becomes too complicated, then maybe we can skip the tests for TF for the time being (though that would only be a temporary solution)

Issue #7479, opened 2022-11-30
Add Linear algebra operations to PyTorchFrontend


AnnaTz:
Add [Linear algebra operations](https://pytorch.org/docs/stable/linalg.html) to PyTorch frontend:

- norm
- vector_norm
- matrix_norm
- diagonal
- [x] #3594 
- [x] #3790
- [x] #3595 
- cond
- [x] #3754 
- [x] #3590
- [x] #3686
- lu
- lu_factor
- eig
- eigvals
- eigh
- eigvalsh
- [x] #3688
- svdvals
- solve
- solve_triangular
- lu_solve
- lstsq
- inv
- pinv
- matrix_exp
- [x] #3755
- cross
- [x] #3596
- vecdot
- multi_dot
- householder_product
- tensorinv
- tensorsolve
- vander
- cholesky_ex
- inv_ex
- solve_ex
- lu_factor_ex
- ldl_factor
- ldl_factor_ex
- ldl_solve
- [ ] chain_matmul
- [x] #3592 
- [ ] cholesky_inverse
- [ ] cholesky_solve
- [ ] dot
- [ ] eig
- [ ] geqrf

aymuos15:
- [ ] #7504

jiisa-k:
[ ] #7513

satani99:
[]#7517

nixx14:
#7519

MrShifu01:
#7526

MrShifu01:
#7527

Nunyapa:
- [ ] #7534

WilliamHirst:
- [ ] #7550

sagewhocodes:
#7559

sagewhocodes:
- [ ] solve_triangular #7559

WilliamHirst:
- [ ] #7571

TakshPanchal:
#7580

theRealBird:
- Taking a shot at this #7595

AnnaTz:
- [ ] #7595

hmahmood24:
- [ ] #7597

sondertg:
qr #7604

SIMRA99:
#7667

Ashvith10:
- [ ] #7720

ReneFabricius:
- [] #7740

sondertg:
> [ ] tensorinv #7786

satani99:
- [ ] #7815

[...]

PR #10326, opened 2023-02-06
Added matrix_exp linear algebra operation to Pytorch frontend

danny-1k:
Add matrix_exp linear algebra operation [#7479](https://github.com/unifyai/ivy/issues/7479)

### DIFF
--- a/ivy/functional/backends/torch/linear_algebra.py
+++ b/ivy/functional/backends/torch/linear_algebra.py
@@ -217,6 +217,15 @@ def matrix_power(
 
 matrix_power.support_native_out = True
 
+@with_unsupported_dtypes({"1.11.0 and below": ("float16", "bfloat16")}, backend_version)
+def matrix_exp(
+    x: torch.Tensor, /, *, out: Optional[torch.Tensor] = None
+) -> torch.Tensor:
+    return torch.linalg.matrix_exp(x, out=out)
+
+
+matrix_exp.support_native_out = True
+
 
 @with_unsupported_dtypes({"1.11.0 and below": ("float16", "bfloat16")}, backend_version)
 def matrix_rank(
--- a/ivy/functional/frontends/torch/linalg.py
+++ b/ivy/functional/frontends/torch/linalg.py
@@ -80,6 +80,11 @@ def matrix_power(input, n, *, out=None):
     return ivy.matrix_power(input, n, out=out)
 
 
+@to_ivy_arrays_and_back
+def matrix_exp(input, *, out=None):
+    return torch_frontend.matrix_power(input, out=out)
+
+
 @with_supported_dtypes(
     {"1.11.0 and below": ("float32", "float64", "complex64", "complex128")}, "torch"
 )


danny-1k:
Hey @CatB1t, please could you help me review my PR?

ivy/functional/backends/torch/linear_algebra.py
```diff
@@ -218,6 +218,16 @@ def matrix_power(
 matrix_power.support_native_out = True
 
 
+@with_unsupported_dtypes({"1.11.0 and below": ("float16", "bfloat16")}, backend_version)
```
CatB1t, line 221: If a function isn't defined in the backend, we should define it as an Ivy API, and implement it for all backends.

ivy/functional/frontends/torch/linalg.py
```diff
@@ -80,6 +80,11 @@ def matrix_power(input, n, *, out=None):
     return ivy.matrix_power(input, n, out=out)
 
 
+@to_ivy_arrays_and_back
+def matrix_exp(input, *, out=None):
+    return torch_frontend.matrix_power(input, out=out)
```
CatB1t, line 85: We should also implement the test for the frontend function `matrix_exp`.

CatB1t (changes requested):

ivy/functional/frontends/torch/linalg.py
danny-1k: sorry for the late response, I'll be working on this today :)

danny-1k:
Hey, @CatB1t , I've implemented the test for the `matrix_exp` frontend and the `matrix_exp` function for all backends

CatB1t (changes requested):
Hello @danny-1k, great work! just few things to fix then we're good to merge!

- We should create an issue addressing adding the new function `experimental` to Ivy and link it to the PR.
- We should add this new function in the `experimental` subpackage of the backends and Ivy.
- implementation for Ivy and `JAX` is missing, we should add these too.

danny-1k:
> 

Hi @CatB1t, I'm a bit unclear about what you mean by "create an issue addressing adding the new function experimental to Ivy and link it to the PR" and "add this new function in the experimental subpackage of the backends and Ivy".

I saw that for all the backends, the function `matrix_exp` has already been implemented in the `experimental` subpackage.
I'm so sorry, I forgot to implement for JAX, I'll work on that soon.

danny-1k:
@CatB1t, I've defined the `matrix_exp` function in the JAX backend

CatB1t (changes requested):
Hello @danny-1k , the backend implementation is actually referring `ivy.exp`, as in torch it is defined to be different: https://pytorch.org/docs/1.11/generated/torch.linalg.matrix_exp.html?highlight=matrix_exp#torch.linalg.matrix_exp . The results are incorrect.

If you were to implement `ivy.matrix_exp` we should create a new issue to expand Ivy functional API, those functions should be implemented in `ivy/functional/experimental/`, similarly for all backends too.

Thanks!

danny-1k:
I think I understand you now. Thanks.
Initially, when I wrote the code for `matrix_exp` in the torch frontend, I did use `torch.linalg.matrix_exp`, I got confused when I saw that the tests were failing because the ground truth was taking the element-wise exp.
I'll create the issue. Now I have to change back to the original matrix_exp implementation 😅

PR #10329, opened 2023-02-06
Fix backends._.diff

xoiga123:
1. Fix backend `diff` type hints.
2. Fix backend `diff` docstrings and examples.
3. Fix `array` and `container` method arguments.
4. Fix `backends.jax.diff` implementation (didn't work before).
5. Fix `backends.tensorflow.diff` implementation (had a minor bug).
6. Fix `test_diff` (didn't test all of the arguments)


kurshakuz (approved):
LGTM!

PR #10371, opened 2023-02-07
Fixing the randint function when using the torch backend

### LINKED ISSUE #10369: [Bug]: ivy.randint with torch backend produces undesired values.
### Bug Explanation

When using ivy.randint with the torch backend, the values that are returned are wrong. 

It should return a random uniform distribution between low and high, but it returns just the lower value. 

### Steps to Reproduce Bug

E.g.: `ivy.randint(2, 10, shape=(10, 2))`

### Environment

Arch Linux, Ipython

### Ivy Version

v1.1.9 

### Backend

- [ ] NumPy
- [ ] TensorFlow
- [X] PyTorch
- [ ] JAX

### Device

CPU

ju2ez:
Closes #10369

### DIFF
--- a/ivy/functional/backends/torch/random.py
+++ b/ivy/functional/backends/torch/random.py
@@ -109,7 +109,7 @@ def randint(
     shape = _check_bounds_and_get_shape(low, high, shape)
     if seed:
         torch.manual_seed(seed)
-    return torch.rand(low=low, high=high, size=shape, device=device).to(dtype)
+    return torch.randint(low=low, high=high, size=shape, device=device).to(dtype)
 
 
 def seed(*, seed_value: int = 0) -> None:
--- a/ivy/functional/backends/torch/random.py
+++ b/ivy/functional/backends/torch/random.py
@@ -96,14 +96,12 @@ def randint(
     high: Union[int, torch.Tensor],
     /,
     *,
-    shape: Optional[Union[ivy.NativeShape, Sequence[int]]] = None,
+    shape: Optional[Union[ivy.NativeShape, Sequence[int]]] = (1,),
     device: torch.device,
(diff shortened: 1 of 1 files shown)


hirwa-nshuti (changes requested):
Hey These changes are made to random_uniform yet in the issue you specified that it is randint🙂

ju2ez:
> Hey These changes are made to random_uniform yet in the issue you specified that it is randintslightly_smiling_face

whoooops , thanks for the pointer :) I now fixed the correct function.

ivy/functional/backends/torch/random.py
```diff
@@ -96,21 +96,18 @@ def randint(
     high: Union[int, torch.Tensor],
     /,
     *,
-    shape: Optional[Union[ivy.NativeShape, Sequence[int]]] = None,
+    shape: Optional[Union[ivy.NativeShape, Sequence[int]]] = (1,),
```
hirwa-nshuti, line 99: This one also should be kept as `None`

hirwa-nshuti (changes requested):
Hey @ju2ez 

Check the comments

ivy/functional/backends/torch/random.py
ju2ez: @hirwa-nshuti , thanks it's done. 

Why do you think `None` makes more sense in this case? 
Is it the convention for backend functions?

hirwa-nshuti: It's because we have explicitly stated it in the functional API🙂
https://lets-unify.ai/ivy/functional/ivy/random/randint.html

hirwa-nshuti (approved):
LGTM!

ivy/functional/backends/torch/random.py
ju2ez: Thank you :)

PR #10381, opened 2023-02-07
Add aminmax torch frontend instance method

### LINKED ISSUE #9946: aminmax

dagleaves:
Close #9946


dagleaves:
All test_tensor tests are currently failing with "backend must be one from ['numpy', 'jax', 'tensorflow', 'torch']", but it passed prior to the frontend test refactor that broke them

ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py
```diff
@@ -1220,6 +1220,37 @@ def test_torch_instance_amin(
...
+    frontend,
+):
+    input_dtype, x = dtype_x
+    helpers.test_frontend_method(
+        init_input_dtypes=input_dtype,
+        init_all_as_kwargs_np={
+            "data": x[0],
+        },
+        method_input_dtypes=input_dtype,
+        method_all_as_kwargs_np={},
```
CatB1t, line 1246: We should also generate a flag `keepdim` using `st.booleans()` strategy and pass it as a method parameter

CatB1t (changes requested):
Hello @dagleavesm, great work 💯! just a small point to be updated and then we're good to merge!

ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py
dagleaves: Got it! I had thought that flag had been removed in the refactor, thanks! Added it back :)

dagleaves:
Oh I see the issue! The `amin` and `amax` tests that I modeled mine after are also missing the `keepdim` flag. The `dim` (axis) flag is also missing from these tests. Does that need to be added back as well?

I did not add it initially as I did not see it in the `amin` or `amax` tests, so I assumed they were removed with the refactor. They used to be there in the old test format before the refactor, so whoever updated these tests post-refactor removed both of those flags.

CatB1t (approved):
lgtm, great work @dagleaves!

PR #10425, opened 2023-02-08
 frontends.torch.linalg: add support for vander

yw14218:


yw14218:
Ref #10389

agniiyer:
Hey the tests are still not passing. Can you take another look? :)

yw14218:
screenshot

Hello, upon investigation it seems that the [torch.linalg.vander](https://pytorch.org/docs/1.12/generated/torch.linalg.vander.html?highlight=vander) function is only available after PyTorch 1.12, and therefore a key error would occur for now as the Ivy image was staged for PyTorch 1.10.

Also, I noted that in my implementation of the function, the last line should change from 
``
return ivy.stack([input ** i for i in range(column_number)], dim=len(input_shape))
``
to
``
return ivy.stack([input ** i for i in range(column_number)], axis=len(input_shape))
``

I think I will not submit a new PR as the tests are going to fail anyways due to aforementioned problems. Please advice me the next step if possible? Do I need to pick another function to implement? Thanks! @agniiyer

ivy-seed:

PR #10445, opened 2023-02-09
 Add Linear Algebra operations to PyTorch frontend #7479 

### LINKED ISSUE #10244: multi_dot

AdrienGaill:
Close #10244


AdrienGaill:
One linting test is failing but concerned functions are not mine.
One test seems to run infinitely.
Other tests are successful.

ivy/functional/backends/torch/linear_algebra.py
```diff
@@ -521,3 +521,33 @@ def vector_to_skew_symmetric_matrix(
 
 
 vector_to_skew_symmetric_matrix.support_native_out = True
+
+
+def multi_dot(
```
zaeemansari70, line 526: This function should not be here, if you want to add this you have to add it in the experimental folder for the functional API.
Moreover, adding a function in the backend is a separate task actually which requires adding it's test in the testing respective files as well.

zaeemansari70 (changes requested):
Hey! 
Thank you soo much for contributing to ivy! 🙂
Your PR looks amazing ✨!
Just requested a few changes and left comments on them 🙂

1. You have implemented the backend and used that same backend function in your frontend, implementing a backend and a frontend are two separate tasks.
2. A task is completed if you add either the backend or the frontend as well as the test for that respective function. In this PR you have mixed two tasks but not added the test for `multi_dot` (the backend one), kindly add that or make another PR in which you implement the complete backend function with the test, after it gets merged you can use that function in this PR as you are using it here for the frontend. 

Thanks! 🙂

AdrienGaill:
I removed the backend function and will implement it in another PR.
The test that was failling is still failing, but is not from a file I've modified and is a linting error.
32 new tests are failing and I don't know why.
Do I have to add a test for the frontend function or is it only for the backend?

zaeemansari70 (commented):
Hi @AdrienGaill 
You have added the frontend function only where you are calling `ivy.multi_dot()` in the frontend, but `ivy. multi_dot()` actually does not exist at the moment in the backend which makes it a dummy call.
Adding the backend is a separate task, when you get the backend of `ivy. multi_dot()` merged from your separate PR, then this will be functional as `ivy.multi_dot()` would actually be present in the backend which your frontend function needs.

Do let me know if you have any other questions, happy to explain.

AdrienGaill:
This PR was closed but it seems it wasn't merged.
I think I'm missing something here, why wasn't it merged ?

PR #10456, opened 2023-02-09
Adding lu_solve frontend

krypticmouse:
* Added lu_solve to frontend as mentioned
* TODO: Tests can't be added unless lu_factor is added.
* TODO: No backend api for lu_solve exists.


krypticmouse:
Close #10443

vaatsalya123:
Hey @krypticmouse first of all thanks for contributing and sorry for the delayed response , Just had a look through the PR and I think a few things need to be changed with your implementation:

1.) I think you should be using composition if your unable to implement one I would recommend you opt for other open task and maybe consider putting this in the experimental API as it's not complete.

2.) Pleas be vary of fixing the lint issues introduced there are easy to miss but keep checking to see if there are new ones introduced by your implementation.

That's basically what I would recommend feel free to ask any other questions thanks!

PR #10502, opened 2023-02-12
frontends.numpy.split

### LINKED ISSUE #10310: split

xoiga123:
Close #10310 

1. Add `frontends.numpy.split` and `test_numpy_split`.
2. Fix type hints for `split` in backends.
3. Fix `split` docstrings.
4. Fix `backends.tensorflow.diff` handling of argument `axis`.
5. Fix a bug in `backends.torch.diff` where it always returns `FloatTensor`.

Note: This PR changed things in `diff` because it is used in the implementation, I could've created a `diff` PR first but that would take 2 review cycles.


aditya-hari:
Hey! Sorry for the late review. Thanks for the contribution. Unfortunately the tests seem to be failing. Additionally, I don't understand some of the changes made. Specifically, why is axis a required argument in every function now?   
 
 And with how many changes this PR has, I really think it would be better served as a separate PRs. As it stands it involves a lot of different changes that ostensibly don't need to be clubbed together.

xoiga123:
@aditya-hari Hi,
1. None of the failing tests are related to the code I changed, I assume it was failing before and somehow got triggered to run.
2. axis is not a required argument, it is a keyword argument with a default, and I discussed this on the `function-arguments` channel on discord:
> `Optional` should only be used in the typehints if the default value of the argument is `None`
3. This is just `split` numpy frontend implementation + `diff` backends bug-fix. Yes it could be better to do a `diff` bug-fix first but that would take a lot of time just to implement a frontend function (in this case could go up to 2 weeks), and I want this ASAP so I could implement other numpy frontend functions like `dsplit`, `vsplit`, `hsplit` without merge conflicts. Plus, no `diff` tests are failing.

aditya-hari:
Looks good to me!

PR #10541, opened 2023-02-14
Fixed `helpers.lists(...)` size bug

ZiadAmerr:


ZiadAmerr:
This bug should be reviewed after this #10481 PR is merged :)

PR #10542, opened 2023-02-14
Fix vector_norm wrong results when order is inf

### LINKED ISSUE #10480: [Bug]: `vector_norm` numpy backend implementation returns wrong result for some norms
### Bug Explanation

When the `norm` argument is `inf` or `-inf`, `ivy.functional.backends.numpy.linalg.vector_norm` is returning different results from `numpy.linalg.norm`.

### Steps to Reproduce Bug

Code to reproduce:
```
>>> import ivy
>>> import numpy as np
>>> x = ivy.array([[1., 2., 3.], [2., -1., 0.]])
>>> ivy.functional.backends.numpy.linalg.vector_norm(x, axis = 0, ord = float("inf"))
array([1., 1., 1.], dtype=float32)
>>> np.linalg.norm(x, axis = 0, ord = float("inf"))
array([2., 2., 3.], dtype=float32)
```

### Environment

Manjaro Linux, python virtual environment with the package versions from `requirements.txt`.

### Ivy Version

v1.1.9

### Backend

- [X] NumPy
- [ ] TensorFlow
- [ ] PyTorch
- [X] JAX

### Device

CPU: AMD Ryzen 5

nikifaets:
Close #10480

### DIFF
--- a/ivy/functional/backends/jax/linear_algebra.py
+++ b/ivy/functional/backends/jax/linear_algebra.py
@@ -401,7 +401,11 @@ def vector_norm(
     if axis is None:
         jnp_normalized_vector = jnp.linalg.norm(jnp.ravel(x), ord, axis, keepdims)
     else:
-        if isinstance(ord, (int, float)) and ord != 0:
+        if ord == jnp.inf:
+            jnp_normalized_vector = jnp.abs(x).max(axis=axis, keepdims=keepdims)
+        elif ord == -jnp.inf:
+            jnp_normalized_vector = jnp.abs(x).min(axis=axis, keepdims=keepdims)
+        elif isinstance(ord, (int, float)) and ord != 0:
             jnp_normalized_vector = jnp.sum(
                 jnp.abs(x) ** ord, axis=axis, keepdims=keepdims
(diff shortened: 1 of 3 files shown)


nikifaets:
Kudos to @suvadityamuk for deep diving into the bug, finding the root causes and proposing a solution.

nikifaets:
@suvadityamuk I added specific test examples that cover the cases when ord is inf or -inf. 

Currently, the ivy `hypothesis_helpers.number_helpers` methods do not support the `allow_infinity` argument that comes with the `hypothesis` methods. Unless I'm missing something, this prevents us from easily using infs in the test arrays.

I'm open to suggestions on how to include `inf` examples in the generic testing method (and be sure that there are infs every time the test is run).

jkeane508:
The logic seems sound to me but might be more appropriate for @suvadityamuk to give this a check over before merging :)

suvadityamuk (approved):
LGTM. Checked tests, they seem to pass.

suvadityamuk:
@jkeane508 Please feel free to merge!

@nikifaets Thank you for your contribution!

PR #10561, opened 2023-02-15
_assert

KevinUnadkat:
adding assert function from utilities for pytorch frontend


KevinUnadkat:
issue ->  _assert #10559

KevinUnadkat:
Probably merge it and then we can implement it in backend

KevinUnadkat:
@WilliamHirst can u add _assert to Development section

PR #10565, opened 2023-02-15
isfortran

shireenchand:
Frontend and test for numpy isfortran

### DIFF
--- a/ivy/functional/frontends/numpy/logic/truth_value_testing.py
+++ b/ivy/functional/frontends/numpy/logic/truth_value_testing.py
@@ -1,5 +1,6 @@
 # global
 import ivy
+import numpy as np
 from ivy.functional.frontends.numpy.func_wrapper import (
     to_ivy_arrays_and_back,
     from_zero_dim_arrays_to_scalar,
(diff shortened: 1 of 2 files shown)


shireenchand:
Hey @suvadityamuk,

The reason why I chose ```a.flags.fnc``` is because we get an extra check here on the fact that is it just``` F_Contiguous``` and not ```C_Contiguous``` as well

suvadityamuk:
> Hey @suvadityamuk,
> The reason why I chose a.flags.fnc is because we get an extra check here on the fact that is it just F_Contiguous and not C_Contiguous as well

Right, understood. Thanks for the explanation!

Could you possibly re-check the linting errors? Might I suggest making use of the [pre-commit hook](https://lets-unify.ai/ivy/contributing/0_setting_up.html#pre-commit) to help?

shireenchand:
Hi @suvadityamuk, is there a reason why these tests are failing? Like something related to the frontend or do they fail usually?

V-G-spec:
Hi, can you revert the changes other than the ones for the function you worked on in the test file?

shireenchand:
@V-G-spec I corrected those. Can you check them?

shireenchand:
Hi @V-G-spec,
Do let me know if there are any other changes that are needed?
Thanks!!

V-G-spec:
Looks good, and I believe @suvadityamuk verified tests were failing due to some other files

PR #10566, opened 2023-02-15
Partial solution to tf.nn.compute_average_loss (no global_batch_size yet)

### LINKED ISSUE #10524: compute_average_loss

tibsor:
Close #10524


tibsor:
Hello,

In order to implement `global_batch_size` in the `tf.nn.compute_average_loss` function, a way to select the [training strategy](https://www.tensorflow.org/api_docs/python/tf/distribute/Strategy) should be implemented, which is quite comprehensive and it is part of TensorFlow's [distributed training API](https://www.tensorflow.org/guide/distributed_training).  I think some separate issues should be created for addressing this.

PR #10571, opened 2023-02-15
numpy.angle function

pawelmagnu:
Close https://github.com/unifyai/ivy/issues/10570


kurshakuz (changes requested):
Please add an associated testing function, test it locally and once it passes please let know so I can start a review. Thanks 🙂

PR #10597, opened 2023-02-16
Introducing Ivy Array view functionality

fspyridakos:


VedPatwardhan (approved):
lgtm! Let's see how it goes with the `func_wrapper` import before the other imports in the `__init__.py` file but I think it should be fine 🙂

PR #10604, opened 2023-02-16
add modules back into ivy

guillesanbri:


ivy/stateful/converters.py
```diff
@@ -243,10 +177,12 @@ def __init__(self, *args, native_module, device, devices, **kwargs):
...
                 return self._hk_params
 
-            def _build(self, *args, **kwargs):
+            def _build(self, params_hk, *args, **kwargs):
                 args, kwargs = ivy.args_to_native(*args, **kwargs)
                 # noinspection PyUnresolvedReferences
-                params_hk = self._native_module.init(RNG, *args, **kwargs)
+                # params_hk = self._native_module.init(RNG, *args, **kwargs)
+
+                # TODO: this _hk_flat_map_to_dict does not seem to be working fine ig ?
```
guillesanbri, line 185: should we keep this as it is now? isn't the `RNG` line needed?

ivy/stateful/module.py
```diff
@@ -549,7 +568,11 @@ def build(
         else:
             dtype = ivy.default_dtype(dtype=self._dtype, as_native=True)
 
-        kwargs["dtype"] = dtype
+        # why are we adding this kwarg in user-defined build ?
+        # it results in the error while doing `from_haiku_module` if haiku's forward
+        # therefore leaving it commented out
+        # kwargs["dtype"] = dtype
+
```
guillesanbri, line 575: not sure if breaks anything if added back, probably good to have it mind

ivy/stateful/converters.py
bicycleman15: since we are passing the params explicitly, we don't need it anymore (RNG is mostly for params init)

ivy/stateful/converters.py
```diff
@@ -281,9 +219,11 @@ def forward_fn(*a, **kw):
                 return model(*i_args, **i_kwargs)
 
             transformed_module = hk.transform(forward_fn)
+            params_hk = transformed_module.init(0, *i_args, **i_kwargs)
```
bicycleman15, line 222: can you add a `RNG` argument to `from_haiku_module` with the default as `0`

ivy/stateful/module.py
bicycleman15: let's leave the comment there, it did break something or the other during transpilation, that's why I commented it.

bicycleman15 (approved):
rest LGTM 👍

PR #10636, opened 2023-02-18
Updated docstring

### LINKED ISSUE #10635: Einops Reduce
Reformat general #791

satakshigarg:
closes #10635

### DIFF
--- a/ivy/functional/ivy/general.py
+++ b/ivy/functional/ivy/general.py
@@ -1822,7 +1822,8 @@ def einops_reduce(
     out: Optional[ivy.Array] = None,
     **axes_lengths: Dict[str, int],
 ) -> ivy.Array:
-    """Perform einops reduce operation on input array x.
+    """
+    Perform einops reduce operation on input array x.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


satakshigarg:
Let me know if there are any suggestions or changes required ?

ivy/functional/ivy/general.py
```diff
@@ -1845,20 +1846,20 @@ def einops_reduce(
 
     Examples
     --------
-    With :class:`ivy.Array` input:
```
juliagsy, line 1848: the original `input` should be fine here :)

juliagsy (changes requested):
Hey there! Thanks for the hard work! Just one small detail, otherwise looks good to me! Please request my review once again when you're done! Thanks a lot!

PS: One side note, just in case you haven't, please also check the equivalent array and container instance methods and ensure that the docstrings and docstring examples are correct! Great work if you had gone through them! Thank you!

ivy/functional/ivy/general.py
satakshigarg: Resolved it @juliagsy

juliagsy (approved):
thanks for contributing!

PR #10641, opened 2023-02-19
An implementation of dropout3d from scratch for all frameworks.

MuhammadSaeedBatikh:
add: dropout3d has been implemented from scratch for all backends except for tensorflow:

  -torch\experimental\layers.py:
       Generated a Bernoulli samples from uniform with the appropriate noise_shape which is then passed as a mask to np.where for efficiency.

  -tensorflow\experimental\layers.py:
      Used the noise_shape argument in tf.nn.dropout to perform a channel-wise dropout.

  -numpy\experimental\layers.py:
       Generated a Bernoulli samples with the appropriate noise_shape which is then passed as a mask to np.where for efficiency.

  -jax\experimental\layers.py:
       Similar to numpy.

and:functional\ivy\layers.py
    calls ivy.current_backend

add:ivy\array\layers.py, ivy\container\layers.py
    instance method support

add:ivy\array\layers.py, ivy\container\layers.py
    instance method support

add: test_ivy\test_docstrings.py
    added to skiplist since output is dependent on external factors.

### DIFF
--- a/ivy/array/layers.py
+++ b/ivy/array/layers.py
@@ -197,6 +197,48 @@ class ArrayWithLayers(abc.ABC):
             out=out,
         )
 
+    def dropout3d(
+        self: ivy.Array,
+        prob: float,
+        /,
+        *,
+        training: bool = True,
+        data_format: str = "NDHWC",
+        out: Optional[ivy.Array] = None,
+    ) -> ivy.Array:
+        """
+        ivy.Array instance method variant of ivy.dropout3d. This method simply
+        wraps the function, and so the docstring for ivy.droput3d also applies
+        to this method with minimal changes.
+
+        Parameters
+        ----------
+        self
+            The input array x to perform dropout on.
+        prob
+            The probability of zeroing out each array element, float between 0 and 1.
+        training
+            Turn on dropout if training, turn off otherwise. Default is ``True``.
+        data_format
+            "NDHWC" or "NCDHW". Default is ``"NDHWC"``.
+        out
+            optional output array, for writing the result to. It must have
+            a shape that the inputs broadcast to.
+
+        Returns
+        -------
+        ret
+            Result array of the output after dropout is performed.
+
+        """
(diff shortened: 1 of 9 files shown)


sherry30 (changes requested):
Hi Muhammad,

There are some lint errors, can you fix those?
I recommend installing `pre-commit`. The instructions are here in this [guide](https://docs.google.com/document/d/1J5Ij0u3nU_jVkOeXvrTOuGF-W9tnwhukenqwtpar83k/edit)

Also, is it not possible to use `torch.nn.functional.dropout` for torch backend implementation? 🤔. Just wondering if it is and if that would be more efficient. If you think this makes more sense then we can go with this.

MuhammadSaeedBatikh:
> Hi Muhammad,
> 
> There are some lint errors, can you fix those? I recommend installing `pre-commit`. The instructions are here in this [guide](https://docs.google.com/document/d/1J5Ij0u3nU_jVkOeXvrTOuGF-W9tnwhukenqwtpar83k/edit)
> 
> Also, is it not possible to use `torch.nn.functional.dropout` for torch backend implementation? 🤔. Just wondering if it is and if that would be more efficient. If you think this makes more sense then we can go with this.

Hi @sherry30 , thanks for the review.
Unfortunately `torch.nn.functional.dropout` is an elementwise dropout that does not accept a `noise_shape` argument, while  `dropout3d` is a channel-wise dropout (i.e. it drops or keeps entire 3D feature maps) and therefore cannot be replaced by a standard elementwise function. I should say that in more recent torch versions there are indeed `dropoutnd` functions'; hence the `ToDo` I added to replace the current implementation for torch once ivy's torch is updated.

MuhammadSaeedBatikh:
hi @sherry30, I fixed the lint error. Please merge.

sherry30 (approved):
I see,
That makes sense, Thanks for the explanation. 😄 

looks good to merge!

PR #10653, opened 2023-02-19
feat: add `eigh` frontend for torch

### LINKED ISSUE #10194: eigh

notjedi:
closes #10194

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
+++ b/ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
@@ -881,3 +881,56 @@ def test_torch_matmul(
         rtol=1e-03,
         atol=1e-06,
     )
+
+
+# eigh
+@handle_frontend_test(
+    fn_tree="torch.linalg.eigh",
+    dtype_x=_get_symmetrix_matrix(),
+    UPLO=st.sampled_from(("L", "U")),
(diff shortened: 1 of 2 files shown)


zaeemansari70 (changes requested):
Hey! 
Thank you soo much for contributing to ivy! 🙂
Your PR looks amazing ✨!

Thanks!

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -881,3 +881,56 @@ def test_torch_matmul(
         rtol=1e-03,
         atol=1e-06,
     )
+
+
+# eigh
+@handle_frontend_test(
+    fn_tree="torch.linalg.eigh",
+    dtype_x=_get_symmetrix_matrix(),
```
zaeemansari70, line 889: An assertion error is being produced, your result in not exactly matching the original frameworks results, but the result differs till some decimal spaces and is still pretty much similar. There are two ways to approach this type of situation:

1. Either you can pass in a safety factor inside array_values() which is being used inside the _get_symmetric_matrix(), which is the recommended way.
2. If this doesn't fix this, you can use `atol` and `rtol`till certain decimal places, and pass them inside the test_function().

zaeemansari70 (changes requested):

PR #10656, opened 2023-02-19
Added _init_overload param to frontend torch.Tensor and numpy.ndarray

Infrared1029:
this allows us to create tensors/arrays with ivy_torch.tensor/ivy_numpy.array or ivy_torch.Tensor/ivy_numpy.ndarray efficiently, if we use tensor/array, _init_overload will be set to True and the tensor returned will be of the same data type as the underlying ivy array, otherwise, the default behavior of Tensor (returning float32 tensors) or ndarray (accepting shapes instead of arrays) will be set.

### DIFF
--- a/ivy/functional/frontends/torch/func_wrapper.py
+++ b/ivy/functional/frontends/torch/func_wrapper.py
@@ -19,9 +19,7 @@ def _from_ivy_array_to_torch_frontend_tensor(x, nested=False, include_derived=No
             x, _from_ivy_array_to_torch_frontend_tensor, include_derived, shallow=False
         )
     elif isinstance(x, ivy.Array) or ivy.is_native_array(x):
-        a = torch_frontend.Tensor(
-            x, _init_overload=True
-        )  # TODO: Find better initialisation workaround
+        a = torch_frontend.Tensor(x, _init_overload=True)
         return a
     return x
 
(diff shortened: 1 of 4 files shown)


ivy/functional/frontends/torch/tensor.py
```diff
@@ -8,10 +8,15 @@
...
 
 class Tensor:
-    def __init__(self, array, device=None):
-        self._ivy_array = ivy.asarray(
-            array, dtype=torch_frontend.float32, device=device
-        )
+    def __init__(self, array, device=None, _init_overload=False):
+
+        if _init_overload:
+            self._ivy_array = ivy.asarray(array, device=device)
```
fspyridakos, line 14: Have this be `        self._ivy_array = (
            ivy.array(array) if not isinstance(array, ivy.Array) else array
        )` like with `EagerTensor` and `DeviceArray`

ivy/functional/frontends/torch/func_wrapper.py
```diff
@@ -19,8 +19,10 @@ def _from_ivy_array_to_torch_frontend_tensor(x, nested=False, include_derived=No
             x, _from_ivy_array_to_torch_frontend_tensor, include_derived, shallow=False
         )
     elif isinstance(x, ivy.Array) or ivy.is_native_array(x):
-        a = torch_frontend.Tensor(0)  # TODO: Find better initialisation workaround
-        a.ivy_array = x
+        a = torch_frontend.Tensor(
+            x, _init_overload=True
+        )  # TODO: Find better initialisation workaround
```
fspyridakos, line 24: Remove ToDo

fspyridakos (changes requested):
Small comments

ivy/functional/frontends/torch/tensor.py
Infrared1029: right!

PR #10671, opened 2023-02-20
fixes a bug where ivy.array creates arrays on default device despite …

Infrared1029:


CatB1t (changes requested):
Hello Yasser, good catch! could you revert unrelated changes? then we're good to merge 👍🏻

PR #10725, opened 2023-02-21
Numpy frontend and Functional API for : function 'percentile'

### LINKED ISSUE #10534: percentile
This issue aims to solve the subtask "percentile" from the ToDo List.

pradyyadav:
Close #10534


pradyyadav:
Close #10740

pradyyadav:
Hi @YushaArif99, I need some guidance on writing tests for the implementation of `percentile`, It requires two required inputs `x: ivy.array` and `q: ivy.array or int`. I tried understanding from the docs and made a PR according to my understanding, but it the run tests are failing which could be a cause of the above problem.

YushaArif99 (changes requested):
Hi @pradyyadav 

I think the tests are failing because of the error in this line (https://github.com/unifyai/ivy/pull/10725/files#diff-2aa28eb2222d749bad6844a5c0385998dfa6767640dbf82c128219d9b654d331R42) where you are trying to unpack 3 values from a tuple that only contains 2 values. 

If you want to get the `axes` argument as well, then maybe you should set the `num_arrays` to 3 instead of 2. 

With regards to the specific arguments for `percentile`, I think the `helpers.dtype_and_values` decorator should suffice in this case. You can read more about the usage of this function here (https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/helpers/hypothesis_helpers/array_helpers.py#L190). 

 
Let me know if you have any other questions 🙂

pradyyadav:
This PR is now active at #11149

PR #10765, opened 2023-02-21
bincount extension

pillarxyz:
#10601


pillarxyz:
note : I still haven't run tests since there was an issue with ivy's docker will run them later and update code accordingly

ivy/functional/ivy/experimental/statistical.py
```diff
@@ -294,3 +294,17 @@ def nanmedian(
...
+@handle_exceptions
+def bincount(
+    x: ivy.Array,
+    /,
+    *,
+    weights: Optional[ivy.Array] = None,
+    minlength: Optional[int] = 0,
+    out: Optional[ivy.Array] = None,
+) -> ivy.Array:
+    return ivy.current_backend().bincount(x, weights=weights, minlength=minlength, out=out)
```
fspyridakos, line 310: Do `ivy.current_backend(x)` and then the rest like you've done

ivy/functional/backends/numpy/experimental/statistical.py
```diff
@@ -104,3 +104,14 @@ def nanmedian(
...
+
+def bincount(
+    x: np.ndarray,
+    /,
+    *,
+    weights: Optional[np.ndarray] = None,
+    minlength: Optional[int] = 0,
+    out: Optional[np.ndarray] = None,
+) -> np.ndarray:
+    return np.bincount(x, weights=weights, minlength=minlength)
```
fspyridakos, line 117: Add `bincount.support_native_out = False`

ivy/functional/backends/tensorflow/statistical.py
```diff
@@ -208,3 +208,13 @@ def einsum(
...
     return tf.cast(tf.einsum(equation, *operands), dtype)
+
+
+def bincount(
+    x: Union[tf.Tensor, tf.Variable],
+    weights: Optional[Union[tf.Tensor, tf.Variable]] = None,
+    minlength: int = 0,
+    *,
+    out: Optional[Union[tf.Tensor, tf.Variable]] = None,
+) -> Union[tf.Tensor, tf.Variable]:
```
fspyridakos, line 219: This one should be moved to the experimental namespace

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -262,3 +262,43 @@ def test_corrcoef(
...
+        shared_dtype=True,
+        min_num_dims=1,
+        max_num_dims=1,
+        min_dim_size=1,
+        max_dim_size=10,
+        min_value=0,
+        max_value=10,
+        allow_nan=False,
+    ),
+    weights=st.booleans(),
```
fspyridakos, line 281: Please generate the weights here instead of in the test, you can use st.one_of if you want to inlclude the case where it's `None`. You can use `key="shared_size"` in the above `helpers.dtype_and_values` function and whatever helper you use to generate the weights.

fspyridakos (changes requested):
Thanks for your contribution! I left a few requested changes, please also add the pre-commit hook as described here: https://lets-unify.ai/ivy/contributing/setting_up.html#pre-commit. There are some linting errors, for example most files don't have a newline at the EOF

pillarxyz:
Can you check now I guess I've gone through all the points mentioned, if there's anything missing or wrong please review

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -260,3 +260,42 @@ def test_corrcoef(
...
+        shared_dtype=True,
+        min_num_dims=1,
+        max_num_dims=1,
+        min_dim_size=1,
+        max_dim_size=10,
+        min_value=0,
+        max_value=10,
+        allow_nan=False,
+    ),
+    test_gradients=st.just(False),
```
fspyridakos, line 280: the weights strategy should be generated within `@handle_test`, perhaps a composite would be easier to make here, here's a similar example that could give you some inspiration: https://github.com/unifyai/ivy/blob/38f579bd487fa555aa58b3ea1f2a5e918b88f41d/ivy_tests/test_ivy/test_functional/test_core/test_linalg.py#L1365
Essentially you want to generate your `x` array, and then generate the weights as either None, or int/float array of the same shape as `x` (use draw for this) and you should also generate an integer and pass it for `minlength`. Also there are differences with the dtype each backend outputs. I would recommend you modify each backend to cast the output to either the default ivy integer dtype, or to the dtype of `weights` if it's a float and not None

fspyridakos (changes requested):
Check out my comment for the next steps

pillarxyz:
I think I fixed all the issues with the tests, also changed some things up with the implementation. Can you check again and see if this is ready for pushing

Issue #10710, opened 2023-02-21
nested_map < Function Reformatting >


kalcho100:

ArpanDhot:
Commit ID: 20e4d12e3389a8a7daaff79eb2391a8b2a200aa7

This commit updates the `nested_map` method in two significant ways:

1. Lambda functions in the method are replaced with defined functions. The three lambda functions used to define the check functions for `tuple`, `list`, and `dict` types have been replaced with explicitly defined functions. This change is intended to increase code readability and maintainability.

2. The method's docstring has been expanded with usage examples. The added example demonstrates how to use the `nested_map` function on a nested list with a simple function to square each element. This addition enhances the method's usability by providing a clear, concrete example of its functionality.

These changes aim to improve the readability, maintainability, and usability of the `nested_map` method. Please run tests to ensure the function's behaviour remains as expected after these modifications.

halbeeb:
I would be working on @nested_map #10710

sampathr0007:
i Want to contribute the code for nested_map #10710  and want to make a Pull request

pedrosaaaa:
21809

PR #10769, opened 2023-02-21
add examples to nested_map

kalcho100:
nested_map #10710


ivy/functional/ivy/nest.py
```diff
@@ -1015,6 +1015,49 @@ def nested_map(
         x following the applicable of fn to it's nested leaves, or x itself if x is not
         nested.
 
+    Examples
+    --------
+    With :code:`ivy.Array` inputs:
+
+    >>> def special_square(nested : float) -> float : return np.square(x)
```
nassimberrada, line 1022: You've defined your function as using `nested` as argument, so it should return `np.square(nested)`. Otherwise set the input as `x : float`

ivy/functional/ivy/nest.py
```diff
@@ -1015,6 +1015,49 @@ def nested_map(
...
+    >>> nested = ivy.Array([[1,2,3],[4,5]])
```
nassimberrada, line 1023: Two comments here: (a) Ivy arrays are called using `ivy.array` instead of `ivy.Array` (lower "a"), and (b) the array has an non homogeneous shape so this would throw an error, you can use `ivy.array([[1,2,3],[4,5,6]])` instead if you want

ivy/functional/ivy/nest.py
```diff
@@ -1015,6 +1015,49 @@ def nested_map(
...
+    >>> results = ivy.nested_map(x = nested,
```
nassimberrada, line 1024: `x` is a positional argument in the function so it should not be passed as `ivy.nested_map(x = nested, fn = special_square )` but just `ivy.nested_map(nested, fn = special_square )`

ivy/functional/ivy/nest.py
```diff
@@ -1015,6 +1015,49 @@ def nested_map(
...
+    ...                   fn = special_square)
+    >>> print(results)
+    [[1, 4, 9], [16, 25]]
```
nassimberrada, line 1027: Given your function input `nested` is an array, the function `nested_map` should return an array `array([[1,4, ....)` not a list

nassimberrada (changes requested):
Hi @kalcho100, thank you for your contribution ! You're mostly there, I've just left a couple comments maybe you could try to address them and adjust all your examples accordingly ? Let me know when you'd like me to review it again. Thanks !

ivy/functional/ivy/nest.py
kalcho100: changed it to x everywhere

ivy/functional/ivy/nest.py
kalcho100: changed to ivy.array

ivy/functional/ivy/nest.py
kalcho100: Removed the "x=" part

ivy/functional/ivy/nest.py
kalcho100: It does return an array, but it is printed the same way as a list is. Should I change it to not be printing, but just returning?

kalcho100:
Hey @VedPatwardhan, why did you close this PR?

VedPatwardhan:
Hey @kalcho100, there has been a `push --force` on master to remove all the images from the commit history, there might be another one in a few hours, please refer the announcement on the discord server. Thanks for your patience and understanding 🙂

PR #10810, opened 2023-02-22
fix frontends.torch.{h,v,d}split

xoiga123:
Apparently, unlike what [the docs](https://pytorch.org/docs/1.11/generated/torch.hsplit.html) state:
> torch.hsplit(input, indices_or_sections) → List of Tensors

these calls are all valid:
```python
a = torch.tensor([1, 2])
torch.hsplit(a, 1)
torch.hsplit(a, indices=[1])
torch.hsplit(a, sections=1)
```

### DIFF
--- a/ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
+++ b/ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
@@ -215,18 +215,40 @@ def split(tensor, split_size_or_sections, dim=0):
     )
 
 
+def _get_indices_or_sections(indices_or_sections, indices, sections):
+    if not indices_or_sections:
+        if isinstance(indices_or_sections, (list, tuple)):
+            pass
+        elif indices and not sections:
+            indices_or_sections = indices
+        elif sections and not indices:
+            indices_or_sections = sections
+        else:
+            raise TypeError("got invalid argument for indices_or_sections")
+    return indices_or_sections
+
+
 @to_ivy_arrays_and_back
-def dsplit(input, indices_or_sections):
+def dsplit(input, indices_or_sections=None, /, *, indices=None, sections=None):
+    indices_or_sections = _get_indices_or_sections(
+        indices_or_sections, indices, sections
+    )
     return tuple(ivy.dsplit(input, indices_or_sections))
 
 
 @to_ivy_arrays_and_back
-def hsplit(input, indices_or_sections):
+def hsplit(input, indices_or_sections=None, /, *, indices=None, sections=None):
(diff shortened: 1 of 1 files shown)


Infrared1029:
Good catch! @xoiga123, left two comments, after that we should be good to merge:)

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
```diff
@@ -215,18 +215,40 @@ def split(tensor, split_size_or_sections, dim=0):
     )
 
 
+def _get_indices_or_sections(indices_or_sections, indices, sections):
+    if not indices_or_sections:
+        if isinstance(indices_or_sections, (list, tuple)):
```
Infrared1029, line 220: this condition is a bit difficult to read/understand, maybe extending the first condition to be `if not indices_or_sections and not isinstance(indices_or_sections, (list, tuple)):` is slightly more readable,  a comment wouldn't hurt either if you see fit:D! (e.g. indices_or_sections can accept empty lists or tuples)

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
```diff
@@ -215,18 +215,40 @@ def split(tensor, split_size_or_sections, dim=0):
...
+def _get_indices_or_sections(indices_or_sections, indices, sections):
+    if not indices_or_sections:
+        if isinstance(indices_or_sections, (list, tuple)):
+            pass
+        elif indices and not sections:
+            indices_or_sections = indices
+        elif sections and not indices:
+            indices_or_sections = sections
+        else:
+            raise TypeError("got invalid argument for indices_or_sections")
```
Infrared1029, line 227: We normally use Ivy Exceptions so we should probably change this as well

ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
```diff
@@ -216,15 +216,15 @@ def split(tensor, split_size_or_sections, dim=0):
 
 
 def _get_indices_or_sections(indices_or_sections, indices, sections):
-    if not indices_or_sections:
-        if isinstance(indices_or_sections, (list, tuple)):
-            pass
-        elif indices and not sections:
+    if not isinstance(indices_or_sections, (list, tuple, int)):
```
Infrared1029, line 219: `if not ivy.exists(indices_or_sections):` is probably even better, I updated the previous comment but was too late it seems, sorry for that, we can merge after this little tweak:)

Infrared1029:
lgtm, thanks for your contribution to ivy:)

Issue #10761, opened 2023-02-21
cross


PrajwalKoirala:
Subtask from **Add Mathematical functions to Numpy Frontend #1525**

PrajwalKoirala:
Added function cross and corresponding test to numpy frontend #10811

PrajwalKoirala:
Added **cross** function to numpy frontend #11091

PR #10811, opened 2023-02-22
Added function cross and corresponding test to numpy frontend

PrajwalKoirala:
Subtask issue **cross #10761** from to-do list **Add Mathematical functions to Numpy Frontend #1525**


PrajwalKoirala:
Close cross #10761

PR #10816, opened 2023-02-22
implemented inplace_increment in paddle backend 

soma2000-lang:
inplace_increment #10817


MahmoudAshraf97 (changes requested):
Hi @soma2000-lang , unfortunately this function doesn't work, there is an inplace add method of paddle that we should use instead

soma2000-lang:
@MahmoudAshraf97 looking into it

MahmoudAshraf97:
this works but it's not inplace, there is a method called add_ which is the inplace version of add, you should use that

PR #10907, opened 2023-02-23
Changes in the function arguments

### LINKED ISSUE #10894: shape

kornelhowil:
Close #10894

### DIFF
--- a/ivy/functional/backends/numpy/general.py
+++ b/ivy/functional/backends/numpy/general.py
@@ -345,7 +345,7 @@ def shape(
     x: np.ndarray,
     /,
     *,
-    as_array: Optional[bool] = False,
+    as_array: bool = False,
 ) -> Union[ivy.Shape, ivy.Array]:
     if as_array:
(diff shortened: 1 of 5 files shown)


ivy/functional/backends/numpy/general.py
```diff
@@ -341,7 +341,12 @@ def scatter_nd(
 scatter_nd.support_native_out = True
 
 
-def shape(x: np.ndarray, /, *, as_array: bool = False) -> Union[ivy.Shape, ivy.Array]:
+def shape(
+    x: np.ndarray,
+    /,
+    *,
+    as_array: Optional[bool] = False,
```
juliagsy, line 348: The `Optional` type hint is for arguments allowing a `None` value, here only bool is allowed so the original type hint is correct, the same comment applies to the other files below! :)

juliagsy (changes requested):
Hey there! Great work so far! Just a small change, otherwise looks good! Please request my review again when the changes are made, thanks again!

juliagsy (approved):
Thanks for contributing!

PR #10918, opened 2023-02-24
lu implementation for the backend

ipriyam26:
torch, tensorflow and jax had native implementations for lu, numpy didn't so had to make it up. it tried to make it as optimised i could. Please let me know if these backend functions are okay. so that i can write frontend implmentations for them and also tests


ipriyam26:
torch, tensorflow and jax had native implementations for  lu, numpy didn't so had to make it up. it tried to make it as optimised i could. Please let me know if these backend functions are okay. so that i can write frontend implmentations for them and also tests

PR #10936, opened 2023-02-24
Paddle paddle

mosesdaudu001:
added statistical function quantile to the backend experimental paddlepaddle.

### DIFF
--- a/ivy/functional/backends/paddle/elementwise.py
+++ b/ivy/functional/backends/paddle/elementwise.py
@@ -387,7 +387,16 @@ def bitwise_right_shift(
     *,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    raise IvyNotImplementedException()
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
please in the next PRs write the issue number so we can track it, for example: closes #10932

mosesdaudu001:
> please in the next PRs write the issue number so we can track it, for example: closes #10932

okay, I will do that for next PRs. Sorry for the mixup

MahmoudAshraf97 (changes requested):
please do not change the function arguments as they are standard across all backends, also find a method to implement interpolation argument since paddle doesn't natively support it

mosesdaudu001:
Hello Mahmoud, I am a little bit confused as to how I should proceed. All other backends natively support the interpolation argument except paddle. Do I add the interpolation argument to the backends.paddle.experimental.statistical quantile function, which would mean tampering with the function arguments? pls how do I proceed.

PR #10995, opened 2023-02-24
Fix typo and delete old reference in setting up docs

lucasalavapena:
I split the two small changes into two commits (they are very small, but I saw that small commits were preferred) if a fast forward merge is preferred, otherwise a squash would work.

Note: the `setting up discussion` (which was added in [this](https://github.com/unifyai/ivy/commit/a4946054e4520ade91cd175313e92a00294714d5) commit) in the past linked to https://github.com/unifyai/ivy/discussions/1308, which now leads to 404 error website right now, but now `setting up discussion`  leads to nothing. Hence I am just removing it. Please let me know if there is new discussion.


Infrared1029:
lgtm! thanks alot @lucasalavapena

Issue #11018, opened 2023-02-25
nanmedian


jatinS-dev:

jatinS-dev:
Feature Implementation 
PR #11020

PR #11020, opened 2023-02-25
Nanmedian

jatinS-dev:


jatinS-dev:
PR for the issue #11018

jkeane508 (changes requested):
Hey,

Thanks for submitting this PR, few changes before we can dive deeper. There are changes to a `.idea` file, these need to be undone before we can merge.

It seems as though you done two implementations of functions, `median()` and `nanmedian()` in this case you would need to open an issue for `median()` aswell. :)

Thanks!

PR #11052, opened 2023-02-26
add count_nonzero to torch frontends

### LINKED ISSUE #11051: count_nonzero

elreweidy:
close #11051


elreweidy:
Hello there! 
the function passed all the tests locally. However, in GitHub tests, it appears that the values returned by the ivy function doesn't match ( in some cases ) the values returned from the Tensorflow backend specifically for some reason, Is that something I can handle?
Thank you :)

PR #11064, opened 2023-02-26
Paddle paddle - #10932

mosesdaudu001:
#10932


MahmoudAshraf97 (changes requested):
hi @mosesdaudu001 , this is not what I meant in my last comment on the other PR, if the other frameworks natively support intepolation, we need to find a method to obtain the same result using paddle functions, simply passing `interpolation` arguments will result in an error because the function doesn't support it, so probably you'll have to calculate it yourself with the aid of `paddle.quantile`

PR #11073, opened 2023-02-26
added paddle quantile function #10932

mosesdaudu001:
added interpolation implementation for paddle quantile function - #10932

hi @MahmoudAshraf97, I hope this is acceptable


MahmoudAshraf97 (changes requested):
Hi @mosesdaudu001 , this seems like auto generated code, while I have no issues with this approach, it's necessary to test the code to make sure proper logic is implemented, the code you submitted outputs zeros no matter what the input was.
Please review your submission and make sure the results are inline with other backends
or feel free to close this PR and work on another task, you don't have to complete it

PR #11075, opened 2023-02-26
added cumprod

fa-anony-mous:


MahmoudAshraf97 (changes requested):
Hi and thanks for your contribution, the code has errors in your usage of slice, check the way you pass the arguments and test your function on paddle tensors to make sure it runs fine and the result is correct

PR #11080, opened 2023-02-26
Add arange for PaddlePaddle backend

### LINKED ISSUE #11046: arange

xingyu-long:
Close #11046

### DIFF
--- a/ivy/functional/backends/paddle/creation.py
+++ b/ivy/functional/backends/paddle/creation.py
@@ -25,7 +25,6 @@ from . import backend_version
 from ivy.utils.exceptions import IvyNotImplementedException
 from paddle.fluid.libpaddle import Place
 from ivy.functional.backends.paddle.device import to_device
-from ivy.functional.backends.paddle.data_type import as_native_dtype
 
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97 (changes requested):
Hi @xingyu-long and thanks for the contribution, 
1. in ivy, the return dtype of a function must be the same as input dtype so cast to the input dtype
2. you need to handle the case where the start and step have different signs, for example:
```
if (step > 0 and start > stop) or (step < 0 and start < stop):
        if isinstance(stop, float):
            stop = float(start)
        else:
            stop = start
```

once these are fixed, we are good to merge

xingyu-long:
Hi @MahmoudAshraf97, Thanks for the suggestion! Please check the latest code.

MahmoudAshraf97 (approved):
Hi, I just fixed some casting rules to ensure consistency with other backends

xingyu-long:
> Hi, I just fixed some casting rules to ensure consistency with other backends

Thanks for the help! I will be more careful for other tasks.

PR #11117, opened 2023-02-27
cond experimental implementation

pillarxyz:
#11047


pillarxyz:
sure will review the things you mentioned and fix them, thanks alot

pillarxyz:
as always tensorflow is giving me nightmares, they do everything differently from other libraries

pillarxyz:
I chose to limit the tests only for square matrices because it's what work for parameters p, non-square matrices do not work for p=1 or p=-1 if I have to capture all cases that would make thing veery complex, but no worries I will keep working on this until I solve all the issues

pillarxyz:
I get this error when I try to reproduce what you did there
image

pillarxyz:
sure I'll find a way to make it work

pillarxyz:
the problem had to with tensorflow's reduce_min and reduce_max I had to specify the axis, I will now push the changes and I guess this edge case will be solved, the usual p values (the non-negative ones)

pillarxyz:
I just change len(x.shape) == 3 to len(x.shape) >= 3 and it works now better

pillarxyz:
but I will test even further to see if any other edge cases come up

PR #11138, opened 2023-02-27
bitwise_and_ function

usama-baloch:


usama-baloch:
Close #11060

YushaArif99 (changes requested):
Hi @usama-baloch 

Can you kindly fix a small error in the `tensor.py` file. It seems that there is an indentation error due to which all CI tests are failing. Here is the error: 
```
E     File "/ivy/ivy/functional/frontends/torch/tensor.py", line 725
E       return self
E                 ^
E   IndentationError: unindent does not match any outer indentation level
```

Let me know if you have any questions! 🙂

PR #11140, opened 2023-02-27
Added torch.linalg.eigvals to torch frontend#10848

### LINKED ISSUE #10848: eigvals

Radu2k:
Tested for both matrix and batch of matrices.
Close #10848


MahmoudAshraf97 (commented):
Thanks for the contribution, LGTM, @AnnaTz can you check that the test is implemented correctly? it's passing so if it's correct please proceed to merge

PR #11146, opened 2023-02-27
Rebuild modular `doc-builder` package

KareemMAX:


KareemMAX:
This is a quick update on the status of the PR for my documentation purposes. Those represent the problems on my radar and may changed later.

- [X] Index
- [X] Overview
- [ ] Functions (blocked by an import problem)
- [ ] Data classes (blocked by an import problem)
- [X] Framework classes
- [X] Nested array
- [X] Utils
- [X] Testing
- [ ] Docs (Needs a custom extension)
- [ ] Discord footer (Needs a custom extension)

KareemMAX:
This is a documentation of the current limitations of the new `doc-builder` against the old one. Most of them are not implemented because **I think** they can be removed from the docs, but I need second opinions on that of course. For convenience in the review process I hosted my documentation version on github pages, [accessible here](https://kareemmax.github.io/ivy)

My reasons of not implementing those points are because of the following causes, denoted by an emoji:

- ⏲Hard to implement
- 🐞Unnecessary feature
- 🚧 Should be added, I'll add it in the review process

The limitations are:

- Replacing `"3.141592653589793"` with the π symbol 🐞
- Adding separate pages for each function or class like [this one](https://lets-unify.ai/ivy/functional/ivy/activations/deserialize.html) 🐞⏲
- The docs can't document the base private classes in the data classes part, it may look broken as I'm planning to find a workaround 🚧⏲
- The docs still use the old `sphinx-rtd-theme`, this is because it was easier to compare rather than the new `pydata_sphinx_theme` 🚧

Also, please note that those are the limitations that I'm aware of, it's possible that I missed something or two.

PR #11149, opened 2023-02-27
frontends and functional API: 'numpy.percentile'

pradyyadav:


pradyyadav:
close #10534

pradyyadav:
close #10740

pradyyadav:
Hi @YushaArif99, I think I figured out failing tests but still there is some more tests which needs to be passed. I would need your help solving these problems.

pradyyadav:
@YushaArif99, at this stage, majority of test passed and I am doubtful, 1 fail maybe due to backend error as I already mentioned above.

YushaArif99 (changes requested):
Hi @pradyyadav 

Yes, you're right about the failing tests. If I'm correct, this task is part of the `Ivy Extension API` right? If so, then what the task also entails is the implementation (primary or compositional) of the function in question for each of the backends (i.e `[JAX, TF, PyTorch, Numpy]`)

You've only written the implementation for the `Numpy` backend and not for the other backends which is why the tests are failing since `ivy.percentile` is not defined for the other 3 backends. Hope that made sense 🙂 

I think the next step here would be to write implementations for the other 3 backends as well along with the tests. I think the `percentile` function should exist in the other backends as well. But if not, then a compositional implementation would be needed. You can read more about it [here](https://lets-unify.ai/ivy/deep_dive/function_types.html#compositional-functions) 

As always, let me know if you have any questions! 🙂

pradyyadav:
Thank you @YushaArif99 , I want to know that, torch does not have `percentile` function, instead it has `quantity`, so how should I do it, should I create a function named `percentile` for torch backend and return `torch.quantile`? Same goes for tensorflow. Tensorflow does not implements percentile, instead the function is implemented in `tesnorflow_probability.stats.percentile`, so how should I proceed onwards?

pradyyadav:
@YushaArif99, This task is for numpy frontend API, not ivy Extension API, however I also implemented `percentile` for ivy functional API. Do I really need to do Extension API part ?


> Yes, you're right about the failing tests. If I'm correct, this task is part of the `Ivy Extension API` right?

pradyyadav:
@YushaArif99, yes this PR imiplements `Ivy Extension API` and `np.percentile frontends`, I came to know later that I should have done it inseparate PRs.

> Yes, you're right about the failing tests. If I'm correct, this task is part of the `Ivy Extension API` right?

YushaArif99:
Hi @pradyyadav 

I think that it's better to then close this PR and create separate PRs for the backend implementations since this PR has become adulterated due to the `forced git push`

As always, if you have any other questions, let me know! 🙂

PR #11163, opened 2023-02-28
Added gradient function to JAX frontend

### LINKED ISSUE #10838: gradient
- [ ] #8426

Supremolink81:
Close #10838

jax.numpy.gradient documentation: .readthedocs.io/en/latest/_autosummary/jax.numpy.gradient.html#jax.numpy.gradient

### DIFF
--- a/ivy/functional/backends/torch/__init__.py
+++ b/ivy/functional/backends/torch/__init__.py
@@ -97,12 +97,11 @@ supports_gradients = True
 def closest_valid_dtype(type, /):
     if type is None:
         return ivy.default_dtype()
-    type_str = ivy.as_ivy_dtype(type)
-    if type_str in invalid_dtypes:
+    if isinstance(type, str) and type in invalid_dtypes:
         return {"uint16": native_uint8, "uint32": native_uint8, "uint64": native_uint8}[
-            type_str
+            type
         ]
-    return type
+    return ivy.as_native_dtype(type)
 
 
 backend = "torch"


ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_math.py
```diff
@@ -2300,3 +2300,31 @@ def test_jax_numpy_inner(
...
+)
+def test_jax_numpy_gradient(
+    *,
+    dtype_and_x,
+    test_flags,
+    on_device,
+    fn_tree,
+    frontend,
+):
+    input_dtypes, inputs = dtype_and_x
```
Aarsh2001, line 2319: Kindly check the test failures. It seems `inputs` cannot be indexed at all.

Aarsh2001 (changes requested):

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_math.py
Supremolink81: On my end, all the tests return:

ValueError: SHA b'e7df13861cbe2081949ee3e043561db665d9161a' could not be resolved, git returned: b'e7df13861cbe2081949ee3e043561db665d9161a missing'

I am unsure why this error is happening. Perhaps my changes are not being reflected properly? Could you send a screenshot from one of the test errors on your end?

Supremolink81: From the documentation:

https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.gradient.html

It seems the variables varargs and axis can have multiple types (varargs can be a scalar or array-like, while axis can be an int or a tuple of ints). In addition, edge_order is not implemented. 

How should I account for this in the implementation? For edge_order, do I raise a NotImplementedError, as the original source code in the docs does? 

In addition, how does one support multiple types/arrays of different lengths in the tests? The inputs are stored in a numpy array, which does not support non-rectangular arrays (something clearly needed due to the fact that varargs can be of arbitrary length).

Supremolink81: Ok, so the only errors that seem to be happening now are due to the bfloat16 data type. Any idea how to solve this?

Supremolink81:
I need help determining why this failure is occurring:

Status: Downloaded newer image for unifyai/ivy:latest
2023-03-03 04:38:46.945979: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory
To preserve the compiler and transpiler caches across multiple machines, ensure that the relative path of your projects from the .ivy folder is consistent across all machines. You can do this by adding .ivy to your home folder and placing all projects in the same place relative to the home folder on all machines.
2023-03-03 04:38:47.762990: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libnvinfer.so.7'; dlerror: libnvinfer.so.7: cannot open shared object file: No such file or directory
2023-03-03 04:38:47.7630[86](https://github.com/unifyai/ivy/actions/runs/4320284239/jobs/7540371999#step:3:87): W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libnvinfer_plugin.so.7'; dlerror: libnvinfer_plugin.so.7: cannot open shared object file: No such file or directory
2023-03-03 04:38:47.763100: W tensorflow/compiler/tf2tensorrt/utils/py_utils.cc:38] TF-TRT Warning: Cannot dlopen some TensorRT libraries. If you would like to use Nvidia GPU with TensorRT, please make sure the missing libraries mentioned above are installed properly.
WARNING:tensorflow:From /usr/local/lib/python3.8/dist-packages/tensorflow/python/util/decorator_utils.py:153: GraphKeys.VARIABLES (from tensorflow.python.framework.ops) is deprecated and will be removed in a future version.
Instructions for updating:
Use `tf.GraphKeys.GLOBAL_VARIABLES` instead.
Please validate your API TOKEN!

This was the cause of every failed test case (except formatting, which was not caused by any files I changed). Any idea why this is happening?

PR #11178, opened 2023-02-28
Allowing ivy.pad to accept float values in "constant" mode

danielmunioz:


ivy/functional/ivy/experimental/manipulation.py
```diff
@@ -936,8 +936,6 @@ def _check_arguments(
             raise ivy.utils.exceptions.IvyException(
                 "constant_values is required for mode: " + mode
             )
```
ZoeCD, line 938: LGTM! But I would suggest to make the call to `_check_constant_values()` here and not inside `pad()`. And add `input.dtype` as an argument in `_check_arguments`. I think it makes the code a little cleaner. Let me know what you think!

ZoeCD (changes requested):

ivy/functional/ivy/experimental/manipulation.py
danielmunioz: Yeah, I thought about having al the checks inside `_check_arguments` but I wanted to have this after the call to `ivy.asarray` just to make sure we do the type check on ivy arrays only, but now I'm thinking that according to the type hints, the inputs should always be ivy arrays right? I'm not sure if there will be a case on which the input will not be an ivy.array, but if so I'm not sure why we're using `ivy.asarray` here

danielmunioz: Also I asked Anna a couple of days ago about this behavior, and today she made these changes https://github.com/unifyai/ivy/commit/d95cb92716f0da6a9f1a6ef5e50d79005cf73912, I haven't tried it yet but it looks like it's a bit safer than just deleting the initial check, so if that commit works well to allow floating values to be used this PR can just add the extra checking on top of it?

PR #11182, opened 2023-02-28
frexp implementation

pillarxyz:
#10844


pillarxyz:
there is an error when I dont add the if out is None: numpy: frexp: 'out' must be a tuple of arrays

I think I added that to the type hints but I think I should specify that out must be a tuple in the tests

pillarxyz:
> TypeError: static_frexp() takes 1 positional argument but 2 positional arguments (and 1 keyword-only argument) were given

humm, I dont seem to get that error myself when I run the tests

pillarxyz:
this is exactly why I added handling for out
image

pillarxyz:
I think I solved it by making (None, None) the default if out is not set

PR #11185, opened 2023-02-28
add frontend numpy copyto

### LINKED ISSUE #11062: copyto

somaia02:
Close #11062

### DIFF
--- a/ivy/functional/frontends/numpy/manipulation_routines/basic_operations.py
+++ b/ivy/functional/frontends/numpy/manipulation_routines/basic_operations.py
@@ -1,6 +1,41 @@
 # local
 import ivy
-from ivy.functional.frontends.numpy.func_wrapper import inputs_to_ivy_arrays
+from ivy.functional.frontends.numpy.func_wrapper import (
+    inputs_to_ivy_arrays,
+    _assert_no_array,
+    _assert_array,
+)
+
+
+@inputs_to_ivy_arrays
+def copyto(dst, src, /, *, casting="same_kind", where=True):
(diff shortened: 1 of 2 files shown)


somaia02:
Thank you for your review.

- I fixed the order. It should be copying to the right array now.
- The TypeError is caused by recent changes to `test_frontend_function`. I contacted someone to fix it. I set `where` to be True only for now.
- I set `test_values=False` since the function doesn't return anything. The test passes even if the function is empty. I thought I could manually check if the `dst` argument was updated correctly. But I can't check that in `test_numpy_copyto`. Since the dst array I'm passing to `test_frontend_function` gets converted to a frontend array and the forntend array is the one that gets updated. 

Should I handle everything manually and not use `test_frontend_function` at all?

somaia02:
@nassimberrada

somaia02:
Thank you for asking @nassimberrada I still need to generate data for the where argument. I had an exam last week so I stopped working for a while but I'm back now.

somaia02:
I tried a few examples like the one you mentioned and the type of the input arrays never changes. I even tried arr1 as bool and arr2 as float. In the example you show I get arr2 with dtype float32 even when I flip the order of the copy. I believe if you check the types of both arrays, you will find that they both don't change.

How can I check if the tests are passing here on github? The tests were passing before but I had a problem with committing so I created a new codespace and copied my code over. Now I'm facing a problem with running the tests on the new codespace.

somaia02:
There was a problem with one of the decorators. It should be working now. @nassimberrada

PR #11194, opened 2023-02-28
Adding multiple statistical functions to paddle backend

5a7man:
Pull request for #10881, #11014 and #10932


MahmoudAshraf97 (changes requested):
hello, experimental functions should be under `ivy/functional/backends/paddle/experimental/statistical.py`
also there is someone else working on quantile so don't include it

Issue #10505, opened 2023-02-12
Conv2D


kaushal07wick:

ivy-seed:
This issue has been labelled as stale. If you would like to continue working on this issue, then please add another comment or create a PR . If a PR has already been created which refers to this issue, then you should explicitly mention this issue in the relevant PR. For more information please refer to this page in our documentation: https://lets-unify.ai/ivy/contributing/1_the_basics.html#todo-list-issues

kaushal07wick:
Conv2D #10505 #11224

kaushal07wick:
some problem with the previous pr, kindly refer to the new pr ASAP Conv2D #10505 #11812

kaushal07wick:
and @AryanSharma9917 it shoes a error like this 
```
ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py:155: in <module>
    dtype_and_x=helpers.dtype_and_shape(
E   AttributeError: module 'ivy_tests.test_ivy.helpers' has no attribute 'dtype_and_shape'
```

when i run the tests

kaushal07wick:
Conv2D #10505 #11991, new pr please take a look at this.

kaushal07wick:
Hey @jkeane508 , I have been working on this problem for quite a while now, I am working on the explicit_padding support.
Why are using the @AryanSharma9917 PR.
Hey @AryanSharma9917 , I have already told you that I am working on this issue, why are u still working on this.

My PR is almost complete. Kindly review the PR.

PR #11224, opened 2023-02-28
 Conv2D #10505 

kaushal07wick:
New PR for the Conv2D support in raw_ops and raw_ops test file
It passes all the tests.
But it doesn't has explicit_padding support.

### DIFF
--- a/ivy/functional/backends/torch/__init__.py
+++ b/ivy/functional/backends/torch/__init__.py
@@ -117,10 +117,10 @@ from . import device
 from .device import *
 from . import elementwise
 from .elementwise import *
-from . import general
-from .general import *
 from . import gradients
 from .gradients import *
+from . import general
+from .general import *
 from . import layers
 from .layers import *
 from . import linear_algebra as linalg
--- a/ivy/functional/backends/torch/general.py
+++ b/ivy/functional/backends/torch/general.py
@@ -16,7 +16,7 @@ import torch
 import ivy
 from ivy.func_wrapper import with_unsupported_dtypes
 from ivy.functional.ivy.general import _parse_ellipsis
-from . import backend_version
+from . import backend_version, is_variable
 
 torch_scatter = None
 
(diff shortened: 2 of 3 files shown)


Infrared1029:
Hi @kaushal07wick! Thanks a lot for taking the time to contribute to ivy, since this is a frontend function, it has to adhere to the true `tf.raw_ops.Conv2D` [function signature](https://www.tensorflow.org/api_docs/python/tf/raw_ops/Conv2D), we can **not** exclude the `explicit_padding` argument (even if we end up not using it in the function). Saying that,  `explicit padding` is not the type of argument to be ignored as it contributes to the final output of the function, unlike other arguments such as `use_cudnn_gpu` which still must be included (to adhere to the true definition of the function) but is never used as it does not contribute to the "mathematics" of `Conv2D`. Also, make sure to remove commented code.

kaushal07wick:
> Hi @kaushal07wick! Thanks a lot for taking the time to contribute to ivy, since this is a frontend function, it has to adhere to the true `tf.raw_ops.Conv2D` [function signature](https://www.tensorflow.org/api_docs/python/tf/raw_ops/Conv2D), we can **not** exclude the `explicit_padding` argument (even if we end up not using it in the function). Saying that, `explicit padding` is not the type of argument to be ignored as it contributes to the final output of the function, unlike other arguments such as `use_cudnn_gpu` which still must be included (to adhere to the true definition of the function) but is never used as it does not contribute to the "mathematics" of `Conv2D`. Also, make sure to remove commented code.

Hi @Infrared1029 , the commented code is of explicit_paddings, bu tit was giving some issues, so I tested it without it and created a PR after successful tests.

I will work on the explicit_padding support and push it very soon.
It would be great to get some help from you.

Infrared1029:
> > Hi @kaushal07wick! Thanks a lot for taking the time to contribute to ivy, since this is a frontend function, it has to adhere to the true `tf.raw_ops.Conv2D` [function signature](https://www.tensorflow.org/api_docs/python/tf/raw_ops/Conv2D), we can **not** exclude the `explicit_padding` argument (even if we end up not using it in the function). Saying that, `explicit padding` is not the type of argument to be ignored as it contributes to the final output of the function, unlike other arguments such as `use_cudnn_gpu` which still must be included (to adhere to the true definition of the function) but is never used as it does not contribute to the "mathematics" of `Conv2D`. Also, make sure to remove commented code.
> 
> Hi @Infrared1029 , the commented code is of explicit_paddings, bu tit was giving some issues, so I tested it without it and created a PR after successful tests.
> 
> I will work on the explicit_padding support and push it very soon. It would be great to get some help from you.

Good luck! sure just ping me if you need any help.

PR #11510, opened 2023-03-01
Fix optional typehint for array and container functions

xoiga123:
[Trello task](https://trello.com/c/coS9qGNO/1392-correct-use-of-optional-type-hint)

This PR also fixes some minor typos and obvious errors


ivy/array/sorting.py
```diff
@@ -102,7 +105,7 @@ def searchsorted(
         /,
         *,
         side="left",
-        sorter=None,
+        sorter: Optional = None,
```
AnnaTz, line 108: Can this be just Optional? What else could it be except for None?

ivy/container/linear_algebra.py
```diff
@@ -2379,13 +2379,10 @@ def tensorsolve(
...
-        axes: Union[int, Tuple[List[int], List[int]]] = None,
+        axes: Optional[Union[int, Tuple[List[int], List[int]]]] = None,
     ) -> ivy.Container:
-        return self.tensorsolve(
-            self,
-            x2,
-            axes=axes,
-        )
+        # TODO: implement this function
+        pass
```
AnnaTz, line 2385: Is there no `tensorsolve`?

ivy/container/sorting.py
```diff
@@ -340,7 +340,7 @@ def static_searchsorted(
         /,
         *,
         side="left",
-        sorter=None,
+        sorter: Optional = None,
```
AnnaTz, line 343: Same as for Array.

ivy/container/sorting.py
```diff
@@ -374,7 +374,7 @@ def searchsorted(
...
```
AnnaTz, line 377: Same.

AnnaTz (changes requested):

ivy/container/linear_algebra.py
AnnaTz: https://github.com/unifyai/ivy/blob/b98068a7e60162f83e1003b18c52314494addb91/ivy/functional/ivy/linear_algebra.py#L2141

xoiga123: The old implementation was wrong, and I was scared this PR may go out of scope so I left it as a TODO :sweat_smile:

AnnaTz: I didn't think it through when I saw you deleted the function call tbh 😅 You can leave it as it is now, no need to add docstrings.

PR #11530, opened 2023-03-02
gaussian_nll_loss for pytorch frontend

### LINKED ISSUE #10920: gaussian_nll_loss

iamharsha1999:
Close #10920


ivy/functional/frontends/torch/nn/functional/loss_functions.py
```diff
@@ -303,6 +303,47 @@ def nll_loss(
...
+        raise ivy.utils.exceptions.IvyError(
+            f"{reduction} is not valid"
+        )
+
+    if ivy.any(var < 0):
+        raise ivy.utils.exceptions.IvyError(
+            "var has negative entry/entries"
+        )
+
+    loss = 0.5 * (ivy.log(var) + (input - target)**2 / var)
```
AnnaTz, line 336: I think you forgot to use `eps` here.

ivy_tests/test_ivy/test_frontends/test_torch/test_loss_functions.py
```diff
@@ -563,6 +563,79 @@ def test_torch_nll_loss(
...
+        available_dtypes=helpers.get_dtypes("float"),
+        allow_inf=False,
+        min_value=0.0,
+        min_num_dims=1,
+        max_num_dims=1,
+        min_dim_size=1,
+        max_dim_size=1,
+    ),
+    dtype_and_eps=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
```
AnnaTz, line 599: Since `eps` is just a number you could use `st.floats` to generate it.

AnnaTz: Also, I'd say that a max value of 1 would make sense for `eps` as it's just there for numeric stability.

ivy_tests/test_ivy/test_frontends/test_torch/test_loss_functions.py
```diff
@@ -563,6 +563,79 @@ def test_torch_nll_loss(
...
 
+# gaussian_nll_loss
+@handle_frontend_test(
+    fn_tree="torch.nn.functional.gaussian_nll_loss",
+    dtype_and_input=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
+        min_value=0.01,
+        max_value=1.0,
+        allow_inf=False,
+        min_num_dims=1,
```
AnnaTz, line 574: The right approach would be to use `get_dtypes("valid") here, and if you get an error that a dtype is not supported for the native torch function, then you add that dtype as unsupported for our frontend function as well https://lets-unify.ai/ivy/deep_dive/data_types.html#supported-and-unsupported-data-types.

AnnaTz: The same goes for `target` and `var` as well.

ivy_tests/test_ivy/test_frontends/test_torch/test_loss_functions.py
```diff
@@ -563,6 +563,79 @@ def test_torch_nll_loss(
...
+        min_dim_size=1,
+        max_dim_size=1,
+    ),
+    dtype_and_target=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("integer"),
+        min_value=0.0,
+        max_value=1.0,
+        allow_inf=False,
+        min_num_dims=1,
+        max_num_dims=1,
```
AnnaTz, line 585: Also, right now you're forcing `input`, `target` and `var` arrays of size 1 (just one element). This is not the only case you should be testing for. You need to test for any shape.

ivy/functional/frontends/torch/nn/functional/loss_functions.py
```diff
@@ -303,6 +303,47 @@ def nll_loss(
...
+
+    if var.shape != input.shape:
+        if input.shape[:-1] == var.shape:
+            var.unsqueeze(dim=2)
+    elif input.shape[:-1] == var.shape[:-1] and var.shape[-1] == 1:
+        pass
+    else:
+        ivy.utils.exceptions.IvyError(
+            f"var is of incorrect size"
+        )
```
AnnaTz, line 324: I don't think these checks match what the documentation says about the shapes. For example, any one dimension could be equal to 1 not just the last one. Maybe you could remove these check altogether for now and see what errors come up while testing.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_loss_functions.py
iamharsha1999: Apparently `.clone()` is not implemented in Jax backend I guess and similarly .clamp_ is not implemented in torch backend I guess. These tow methods are raising errors from backend. I figured a roundabout way of doing same thing by suing `torch_frontend` functions and array object methods. Kindly review it and see If I have done it rightly.

ivy_tests/test_ivy/test_frontends/test_torch/test_loss_functions.py
iamharsha1999: I have set the max_dim_size = 5. I believe this should be representative for higher dimensional arrays.

ivy/functional/frontends/torch/nn/functional/loss_functions.py
iamharsha1999: Yeah there was bug here. All the if-else conditions should be nested under the `if var.shape != input.shape:`. I have fixed it now.

ivy_tests/test_ivy/test_frontends/test_torch/test_loss_functions.py
iamharsha1999: Fixed these.

ivy/functional/frontends/torch/nn/functional/loss_functions.py
iamharsha1999: Fixed them.

iamharsha1999:
@AnnaTz Could you kindly review the changes once again and let me know ?

ivy/functional/frontends/torch/nn/functional/loss_functions.py
```diff
@@ -333,6 +333,11 @@ def gaussian_nll_loss(
             "var has negative entry/entries"
         )
 
+    var = torch_frontend.tensor(ivy.array(var, copy=True))
+    ivy.set_with_grads(False)
+    var = torch_frontend.clamp(var, min=eps)
+    ivy.unset_with_grads()
```
AnnaTz, line 339: I don't understand why this grad handling was necessary. Have you tried using `ivy.maximum`?

ivy/functional/frontends/torch/nn/functional/loss_functions.py
```diff
@@ -315,13 +315,13 @@ def gaussian_nll_loss(
...
-        ivy.utils.exceptions.IvyError(
-            f"var is of incorrect size"
-        )
+            var = torch_frontend.unsqueeze(var, dim=2)
+        elif input.shape[:-1] == var.shape[:-1] and var.shape[-1] == 1:
+            pass
+        else:
+            raise ivy.utils.exceptions.IvyError(
+                "var is of incorrect size"
+            )
```
AnnaTz, line 324: I still don't think these checks are consistent with what the documentation mentions, but this will become clearer during testing.

AnnaTz (changes requested):

ivy/functional/frontends/torch/nn/functional/loss_functions.py
iamharsha1999: Hey I basically transpiled the code line by line from the torch implementation. The torch implementation for this piece of code was like this:

```
# Clamp for stability
var = var.clone()
with torch.no_grad():
    var.clamp_(min=eps)
```

Obviously, this might not be the most efficient way of implementing the function. But I just took the most simplest approach directly transpiring the above. If you have any specific direction I have to follow kindly let me know. Do you want me to remove the grad handling completely ?

Will something like this work ?

```
var = torch_frontend.clamp(var, min=eps)
```

or

```
var = ivy.maximum(ivy, min)
```


Also can you explain the necessity to use `ivy.maximum` in the place of `clamp`. I'm asking this to have better understanding of the framework and would help me to implement future functions easily.

PR #11532, opened 2023-03-02
Improved Dockerfile and DockerfileGPU efficiency

ZiadAmerr:
Both dockerfiles both build with much faster speed due to only cloning the repository with depth 1, you can also find a confirmation from @Dsantra92 [here](https://discord.com/channels/799879767196958751/942114744691740772/1080665082071101501) that this works with no problems


Dsantra92:
Force push and create a new PR. :)

PR #11596, opened 2023-03-02
Use image instead of dockerfile in dev container

Dsantra92:
The change allows user to use pre-built docker image to build the vs-code dev-containers instead of building the ivy image locally. Also, added the pre-commit step for git workflow inside the container.

Before merging:
- [ ] Test if it works with vs-code: creating devcontainer, using remote terminal etc.
- [ ] Test if it works with GPU machine.
- [ ] Test if it works with codespaces.
- [ ] Understand how it will it affect M1 chip devcontainers?

Suggestion: We should move the `.devcontainer` folder to `.gitignore` allowing user to tailor the development environment according to their requirements without requiring to stash every time they pull.


.devcontainer/post_create_commands.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
 
-pip3 install black
-pip3 install flake8
-
 git submodule update --init --recursive
 
-python3 -m pip install --user -e .
```
CatB1t, line 8: Quick note: this is used here to install Ivy as a local package, we should keep it.

CatB1t (changes requested):
Just as mentioned in the Discord, let's make two subdir `build` and `image`, each has two devcontainer settings, one for `CPU` other for `GPU`. Feel free to merge after it's updated! great work (:

Dsantra92:
Looks like multiple dev containers are not supported yet https://github.com/microsoft/vscode-remote-release/issues/1165. I can add the config jsons, but you still need to copy the needed `.devcontainer.json` to `.devcontainer` folder @CatB1t.

PR #11612, opened 2023-03-03
monkey patch np.array for ivy.array conversion 

Aarsh2001:
The following code snippet now works -:

```
import ivy
import numpy as np

for fw in ["jax", "numpy", "torch", "tensorflow"]:
    try:
        ivy.set_backend(fw)
        a = ivy.array(1e-5)
        b = np.array(a)
    except Exception as e:
        print(f"Error in {fw}: {e}")
```

### DIFF
--- a/ivy/data_classes/array/conversions.py
+++ b/ivy/data_classes/array/conversions.py
@@ -3,6 +3,7 @@ instances.
 """
 
 # global
+import numpy as np
 from typing import Any, Union, Tuple, Dict, Iterable, Optional
 
 # local
@@ -14,8 +15,11 @@ import ivy
 
 
 def _to_native(x: Any, inplace: bool = False) -> Any:
+    current_backend = ivy.current_backend_str()  # cache current backend string
     if isinstance(x, ivy.Array):
         return x.data
+    elif isinstance(x, np.ndarray) and current_backend != "numpy":
+        return ivy.Array(x).data
     elif isinstance(x, ivy.Container):
         return x.cont_map(
             lambda x_, _: _to_native(x_, inplace=inplace), inplace=inplace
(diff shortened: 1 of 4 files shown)


Aarsh2001:
Looking forward to your thoughts @VedPatwardhan @guillesanbri !🙂

ivy/__init__.py
```diff
@@ -1155,3 +1155,14 @@ def __exit__(self, type, value, traceback):
...
+#We monkey patch np.array to enable ivy.Array conversion with tensorflow backend
+_np_array = np.array
+
+@wraps(_np_array)
+def nparray(*args, **kwargs):
+    if isinstance(args[0], ivy.Array):
+        return ivy.to_numpy(*args, **kwargs)
+    return  _np_array(*args, **kwargs)
+
+np.array = nparray
```
guillesanbri, line 1168: we shouldn't override the whole `np.array` I guess :sweat_smile:

guillesanbri (commented):
Just left a minor comment, overall I think this approach is a little bit nuclear and the overriding shouldn't happen at this level, the idea was more oriented towards finding what internal method is called by `np.array` and then modifying that within the `ivy.Array` class, not sure if that's doable tho

ivy/__init__.py
Aarsh2001: Sure will have look. I guess the most granular method is `np.core.multiarray.array(*args, **kwargs)`

ivy-seed:

guillesanbri:
I think defining a `__array__` method does the trick and allows you to control how `np.array` works for a particular class :slightly_smiling_face:

Aarsh2001:
Thanks @guillesanbri, will have a look and get back to you w results

Issue #10664, opened 2023-02-20
asarray


horizon3902:

horizon3902:
added functional examples for ivy.asarray subtask in #790 #10663

PR #11627, opened 2023-03-03
array #11143

dash96:


fspyridakos:
@VedPatwardhan do we want to reformat `array` or should we just keep it as an alias to `asarray`?

VedPatwardhan (commented):
Hey @fspyridakos, thanks for requesting the review. I don't think we'd should reformat `asarray` because it's a part of the [Array API Standard](https://data-apis.org/array-api/latest/API_specification/generated/array_api.asarray.html). I'm also not sure why we are adding a backend implementation for `ivy.array`, ideally it should just be an alias to `ivy.asarray` 🤔
Is this related to a task? Thanks @fspyridakos and @dash96 🙂

fspyridakos:
There's this https://github.com/unifyai/ivy/issues/11143 and this issue https://github.com/unifyai/ivy/issues/10664. The latter is a simple docstings change, whereas the former I think we can just leave as an alias

VedPatwardhan:
Yeah, I'm not sure why it would be in that ToDo list at all 😅
But anyway, as you said, we should leave the `array` as it is (and probably also remove it from the ToDo list) while the `asarray` reformatting would be just like any other reformatting task. Thanks again for looking into this 🙂

dash96:
Thanks @VedPatwardhan and @fspyridakos. I will be closing this PR now as it is not necessary to add a new function called array.

PR #11671, opened 2023-03-03
Added dockerfile with extra libraries and modified devcontainer to in…

fspyridakos:
…clude a line that allows jetbrains getway to connect to codespace


docker/Dockerfile
```diff
@@ -1,6 +1,19 @@
 FROM ubuntu:20.04
 WORKDIR /ivy
 
+# Install Docker CE CLI
```
VedPatwardhan, line 4: Is it required to install docker inside the dockerfile? Are we creating docker images inside the docker container as well? If not, ig we shouldn't install docker in the dockerfile 🙂

docker/DockerfileApplied
```diff
@@ -0,0 +1,56 @@
+FROM ubuntu:20.04
+WORKDIR /ivy
+
+# Install Docker CE CLI
```
VedPatwardhan, line 4: Same doubt here as well 🙂

VedPatwardhan (commented):
Just one doubt, rest look good 🙂

docker/Dockerfile
fspyridakos: Yeah I added it so that we can use codespaces witht he pycharm IDE, just like the regular setup there needs to be a docker installation on the virtual machine which will be used as the interpretor. Not sure if you want to make this a seperate thing for people specifically aiming to use codespaces with pycharm

VedPatwardhan: Yeah I think we should be having that as optional, because installing docker for everyone would increase their image's size by quite a bit, and given that it's not applicable to everyone, we could make it something like an [argument](https://docs.docker.com/engine/reference/builder/#arg) you pass while building the image about whether you are one pycharm 🙂

fspyridakos: Right I see, I'll look more into this on Monday. I'll try making it optional unless a certain argument is passed while building.

VedPatwardhan: Sure, of course ig this PR would probably get closed by then (due to the push --force), but we can continue the discussion on the new PR 🙂

fspyridakos:
Figured out how to add an arg now this: `docker build -t unifyai/ivy:latest --no-cache -f DockerfileApplied .. --build-arg CLI="True"` will build with the CLI, otherwise it won't be included

VedPatwardhan (approved):
lgtm! Feel free to merge @fspyridakos. Thanks 🙂

fspyridakos:
Actually, I just realised I could do the same arg thing for the additional libraries and just keep 1 dockerfile, will commit these changes now

fspyridakos:
Fair enough, will mere this then

PR #11687, opened 2023-03-03
Support ivy.Array creation using np.array in the constructor

xoiga123:
[Trello task](https://trello.com/c/79JWfUfn/1407-support-ivyarray-creation-using-nparray-in-the-constructor)
This PR made ivy.Array accept anything that ivy.array accepts, and also switches to the correct (global) backend.
@VedPatwardhan

### DIFF
--- a/ivy/array/array.py
+++ b/ivy/array/array.py
@@ -29,7 +29,6 @@ from .sorting import _ArrayWithSorting
 from .statistical import _ArrayWithStatistical
 from .utility import _ArrayWithUtility
 from ivy.func_wrapper import handle_view_indexing
-from ivy.utils.backend.handler import _determine_backend_from_args
 from .experimental import (
     _ArrayWithSearchingExperimental,
     _ArrayWithActivationsExperimental,
@@ -95,7 +94,7 @@ class Array(
     _ArrayWithStatisticalExperimental,
     _ArrayWithUtilityExperimental,
 ):
-    def __init__(self, data):
+    def __init__(self, data, dynamic_backend=None):
         _ArrayWithActivations.__init__(self)
         _ArrayWithCreation.__init__(self)
         _ArrayWithDataTypes.__init__(self)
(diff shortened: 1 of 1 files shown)


VedPatwardhan (changes requested):
Hey @xoiga123, we shouldn't be removing the `dynamic_backend` setting for enabling this change, just updating the way the `ivy.utils.assertions.check_true` works should take `np.ndarray` into account and that should fix the issue. Happy to answer any questions. Thanks 🙂

xoiga123:
@VedPatwardhan I misunderstood the issue, can you help me clarify it?
```python3
ivy.set_backend('tensorflow')
ivy.Array(np.array([1.])  # data now is np.array or tf.Tensor?
ivy.Array(torch.tensor([1.])  # data now is np.array or torch.Tensor or tf.Tensor?

# I thought we will make ivy.Array() accept any backend and convert to the global backend, like:
ivy.Array(np.array([1.])) # np -> np -> tf, data now is tf.Tensor
ivy.Array(torch.tensor([1.]) # torch -> np -> tf, data now is tf.Tensor
```

VedPatwardhan:
We're not looking for `ivy.Array` to support any native array, we only want to replicate the behaviour with `ivy.asarray` in this constructor too. So `ivy.asarray` works when the input passed to the function is either the valid native array for that backend or `np.ndarray`, but the `ivy.Array` constructor only works if it's a valid native array and doesn't work with `np.ndarray`, we want to enable that same behaviour in the `ivy.Array` constructor just like `ivy.asarray`. Hope that helps 🙂

xoiga123:
@VedPatwardhan `ivy.asarray` supports lists, tuples, etc... Should `ivy.Array` also support them? Or just native and `ndarray` :thinking:

xoiga123:
```python3
# ivy.asarray does accept dict (Container) and convert to Container
>>> type(ivy.array(ivy.Container()))
<class 'ivy.container.container.Container'>

# But when we call ivy.Array with a dict (Container),
# should we replace self (ivy.Array) with another object (ivy.Container)?
>>> type(ivy.Array(ivy.Container()))
```

ivy/array/array.py
```diff
@@ -140,11 +140,14 @@ def __init__(self, data, dynamic_backend=None):
     def _init(self, data, dynamic_backend=None):
         if ivy.is_ivy_array(data):
             self._data = data.data
+        elif ivy.is_native_array(data):
```
VedPatwardhan, line 143: We should instead update the condition checked in `ivy.utils.assertions` so that it returns `True` if `data` is either a `ivy.NativeArray` or a `np.ndarray` rather than adding more conditions here. And finally we could replace `self._data` with `self._data=ivy.asarray(data).to_native()` as you've done in the `elif`

VedPatwardhan (changes requested):
Hey @xoiga123, let's only get `ivy.Array` to support `np.ndarray` as the input with all backends similar to `ivy.asarray` for now, we could look at other objects (lists, tuples) later. Thanks for looking into this issue 🙂

ivy/array/array.py
xoiga123: I think we must have a condition for `if native: self._data=data` because it will be stuck in a loop of creating another `ivy.Array`?

VedPatwardhan: Yep, you're right 👍
Then the only change that would be needed is `ivy.asarray` instead of `ivy.current_backend().asarray` and `._data` for faster access

xoiga123: Got it, thank you for the thorough explanation :1st_place_medal:

VedPatwardhan (approved):
lgtm! Thanks for making the changes. I'll merge the PR now 🙂

PR #11696, opened 2023-03-04
Roll raw ops

### LINKED ISSUE #11617: Roll

Husienvora:
Close #11617


Husienvora:
Hey the tests are passing and an alias was used for roll because there is already a frontend function for it with same args

Husienvora:
hey this is the new pr #11844

PR #11701, opened 2023-03-04
quantile #10932

mosesdaudu001:
added interpolation implementation for quantile function.
@MahmoudAshraf97
This implementation should work.
[quantile] - #10932


MahmoudAshraf97 (changes requested):
Hi, this implementation seems solid, it just has some problems with the output shapes and dimensions eg:
```
AssertionError:  the results from backend paddle and ground truth framework tensorflow do not match
E            [[-1. -1.]
E            [-2. -2.]]!=[[-1. -2.]
E            [-1. -2.]]
```

also the output axes are reversed so use the hypothesis tests to correct these then we are good to merge

PR #11717, opened 2023-03-04
[feat] implemented unique_values for paddle

geetptl:
Submitting final code for #11714

I have used paddle.unique, and added a block to concatenate nan values if needed.

I have matched up the results by manually checking with the results of numpy and torch implementations of the same function.


MahmoudAshraf97 (approved):
added some casts to increase dtype combatibility, in the next PRs please make sure your function is passing the hypothesis tests

PR #11719, opened 2023-03-04
Ivy torch column

AryanSharma9917:


AryanSharma9917:
Hi @agniiyer ,I don't know what the issue but till yesterday only 2 tests were failing but now almost every test is failing could you please help me out

PR #11746, opened 2023-03-05
added nan_to_num

vandit98:
nan_to_num #11111
hi @MahmoudAshraf97
As adviced by you. I have used tensorflow implementation of nan_to_num in paddle framework and implemented it.
Please review it.
Thanks

### DIFF
--- a/ivy/functional/backends/paddle/experimental/elementwise.py
+++ b/ivy/functional/backends/paddle/experimental/elementwise.py
@@ -223,7 +223,12 @@ def hypot(
     *,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    raise IvyNotImplementedException()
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97 (changes requested):
AssertionError: the ground truth framework torch returned a float32 datatype while the backend paddle returned a float64 datatype

vandit98:
hi @MahmoudAshraf97
Can u help me with this error. I have tried to run test for allclose.
Is there any documents or resources for doing this.
image

MahmoudAshraf97:
did you recheck that test_allclose exists in this file?

vandit98:
hi @MahmoudAshraf97 
Yes it do exist.
Screenshot from 2023-03-08 17-56-02

also when i am trying to pass float32 , i am encountering this issue. Due to which i have to convert the input:x to float64. Because this way it is working for each datatype.

image

PR #11749, opened 2023-03-05
add broadcast_tensors to torch frotend

elreweidy:
#11747

### DIFF
--- a/ivy/functional/frontends/tensorflow/raw_ops.py
+++ b/ivy/functional/frontends/tensorflow/raw_ops.py
@@ -716,4 +716,30 @@ def Real(input, Tout=ivy.float32, name="Real"):
     return ivy.Real(input, Tout=Tout)
 
 
+@to_ivy_arrays_and_back
+def BandedTriangularSolve(
+    matrix, 
+    rhs, 
+    lower=True, 
+    adjoint=False, 
+    name="BandedTriangularSolve",
+):
+    return ivy.BandedTriangularSolve(matrix, rhs, lower=lower, adjoint=adjoint)
+
+
+@to_ivy_arrays_and_back
+def BatchMatMul(x, y, adj_x=False, adj_y=False, name='BatchMatMul'):
+    return ivy.BatchMatMul(x, y, adj_x=adj_x, adj_y=adj_y)
+
+
+@to_ivy_arrays_and_back
+def BatchMatMulV2(x, y, adj_x=False, adj_y=False, name='BatchMatMulV2'):
+    return ivy.BatchMatMulV2(x, y, adj_x=adj_x, adj_y=adj_y)
+
+
+@to_ivy_arrays_and_back
+def BatchMatMulV3(x, y, Tout=ivy.Dtype, adj_x=False, adj_y=False, name='BatchMatMulV3'):
+    return ivy.BatchMatMulV3(x, y, Tout=Tout, adj_x=adj_x, adj_y=adj_y)
+    
+    
 Slice = to_ivy_arrays_and_back(map_raw_ops_alias(tf_frontend.slice))
--- a/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
+++ b/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py
@@ -3471,3 +3471,131 @@ def test_tensorflow_Real(
         input=x[0],
         Tout=Tout,
     )
+
+
+# BandedTriangularSolve
+@handle_frontend_test(
+    fn_tree="tensorflow.raw_ops.BandedTriangularSolve",
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
+    ),
+    test_with_out=st.just(False),
+)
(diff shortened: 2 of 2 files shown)


Infrared1029:
Hi Ahmed! nice to see you again:D, thanks a lot for the consistent contributions, the testing function needs a bit of work to be representative of the actual inputs `torch.broadcast_tensors` expects, the current test will generate arrays with dim=1, which is a very trivial input, also, increasing the `max_num_dims` won't help because `helpers.dtype_and_values` will end up generating arrays of the same size (not sure if this is intended or a bug I'll have to get back to Mustafa regarding this), and even assuming both shapes are different there is no guarantee that the shapes are broadcast-able, for more fair tests you might have to write a more custom strategy, where you generate arrays with broadcast-able shapes and test the function with. An extra small note, no need to cast the arrays into numpy arrays as they are already numpy arrays.

elreweidy:
sorry if this is too much reviewing for you, (I'm actually still so busy these days :'D) . I will rebuild the test for this and come back to you.
Thanks for the thorough review.

elreweidy:
In the frontend, I have wrote some code to insure that the tensors will always be broadcast-able. However, the tests pass any ways. not sure why. in addition to this I have generated 2 sets of 3 tensors with different shapes and the same dtypes usting the _broadcast_tensors_helper funtion to solve the issue of all the tensors having the same dims.
what i'm not sure about is why the function passes regardless of the code i added in the frontend.

Infrared1029:
Hi Ahmed, can we create a new PR or resolve the conflicts since the git push made the PRs a bit messy

elreweidy:
there are no more conflicts now :)

Infrared1029:
apologies for the super late response, ill take a look now

Infrared1029:
The test function appears to be generating a few tensors without checking whether these tensors are broadcastable or not, the tests should fail ideally, the reason the tests are not failing I'm assuming is because of a bug we found two weeks ago that makes subsequent calls to `helpers.dtypes_and_values` generate arrays with the same shape, in other words, all tensors have the same shape so the function does not fail, i have not investigated if this is the case here but it most probably is.

elreweidy:
> The test function appears to be generating a few tensors without checking whether these tensors are broadcastable or not, the tests should fail ideally, the reason the tests are not failing I'm assuming is because of a bug we found two weeks ago that makes subsequent calls to `helpers.dtypes_and_values` generate arrays with the same shape, in other words, all tensors have the same shape so the function does not fail, i have not investigated if this is the case here but it most probably is.
> 
> @CatB1t we should probably fix this asap if it is the case

Hello Yasser, if this is the case. How can you tell that the code in the frontend to check if the tensors are always broadcast-able is not doing anything? I'm just asking so that I won't be having to delete it and then add it again :)

Infrared1029:
> > The test function appears to be generating a few tensors without checking whether these tensors are broadcastable or not, the tests should fail ideally, the reason the tests are not failing I'm assuming is because of a bug we found two weeks ago that makes subsequent calls to `helpers.dtypes_and_values` generate arrays with the same shape, in other words, all tensors have the same shape so the function does not fail, i have not investigated if this is the case here but it most probably is.
> > @CatB1t we should probably fix this asap if it is the case
> 
> Hello Yasser, if this is the case. How can you tell that the code in the frontend to check if the tensors are always broadcast-able is not doing anything? I'm just asking so that I won't be having to delete it and then add it again :)

Not sure if I understand the question but I'll try to clarify my points more, the code you added in the frontend, basically extends smaller tensors with a dummy "1" dimension, this does not make the tensors broadcastable, a tensor of shape (3,2) will broadcast just fine with a tensor of shape (2,), you don't have to turn the shape into (1,2), on the other hand, a tensor with the shape (4,) will not broadcast to (3,2) even if you turn it to (1,4), point is, tensors do not broadcast just because we extend the dimensions, that's the first point, the second point is, you don't have to check whether the input tensors are broadcastable or not on the frontend, the function is expected to fail if the tensors are not broadcastable anyway, you are supposed, on the other hand, to make sure that the test function is generating broadcastable shapes (like a tensor of shape (3, 2) and a tensor of shape (2,)).

elreweidy:
I see Yasser, thank you for the clarifications. Can I tweak the code i wrote to generate the tests for the tensors to be broadcastable ? Is there a way to do that ? If so, i will try to resolve the issue in my side as soon as i can.

Infrared1029:
> I see Yasser, thank you for the clarifications. Can I tweak the code i wrote to generate the tests for the tensors to be broadcastable ? Is there a way to do that ? If so, i will try to resolve the issue in my side as soon as i can.

you can do that by first generating one array of shape (w, x, y, z) and then generating another array of shape (x, y, z) or (y, z) or (z, ) randomly, issue is, atm there is a subtle bug that if we call `dtypes_and_values` consecutively, we get arrays of the same shape regardless of the shape specified in the second call, Mustafa should be fixing this soon, but until this is fixed I'm not sure if we can write a reliable test for this function, i'd say just write the test and i'll make sure it looks fine "code-wise" and will merge it, when Mustafa fixes the issues it will work as expected. You can use a custom strategy to do what i just said, i'd make sure first that there is not a helper that generates broadcastable arrays right away to save you the hustle.

ivy-seed:

PR #11783, opened 2023-03-06
Fix method tests to work with inplace special ops

xoiga123:
[Trello task](https://trello.com/c/FjGCje7g/1291-fix-method-tests-to-work-with-inplace-special-ops)


xoiga123:
@VedPatwardhan do you mind taking a look at `torch.matmul` inplace update as discussed in discord? :dizzy_face:

ivy/functional/backends/torch/linear_algebra.py
```diff
@@ -168,7 +168,7 @@ def matmul(
     adjoint_b: bool = False,
     out: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
-
+    x1, x2 = [x.clone() if out is x else x for x in (x1, x2)]
```
xoiga123, line 171: @VedPatwardhan

xoiga123:
@fspyridakos :sweat_smile:

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -1115,8 +1116,8 @@ def test_method(
         for v, d in zip(init_flags.as_variable, init_input_dtypes)
     ]
 
-    # Create Args
-    args_constructor, kwargs_constructor, *_ = create_args_kwargs(
+    # Save original constructor data for inplace operations
+    con_data = OrderedDict(
```
CatB1t, line 1120: Is it necessary to create an OrderedDict here? can't we just copy the original args_constructor and kwargs_constructor?

ivy_tests/test_ivy/test_misc/test_array.py
```diff
@@ -995,6 +1003,10 @@ def test_array__truediv__(
     on_device,
 ):
     dtype, x = dtype_and_x
+
+    # bfloat16 is not supported by numpy
+    assume(not ("bfloat16" in dtype))
```
CatB1t, line 1008: We shouldn't be skipping specific dtypes, this ideally shouldn't be generated for NumPy backend anyways.

ivy_tests/test_ivy/test_misc/test_array.py
```diff
@@ -977,12 +986,11 @@ def test_array__rdivmod__(
 @handle_method(
     method_tree="Array.__truediv__",
     dtype_and_x=helpers.dtype_and_values(
-        available_dtypes=helpers.get_dtypes("numeric"),
+        available_dtypes=helpers.get_dtypes("real_and_complex"),
```
CatB1t, line 989: Why are we limiting the data type generation to real_and_complex?

CatB1t (changes requested):

ivy_tests/test_ivy/helpers/function_testing.py
xoiga123: https://github.com/unifyai/ivy/blob/74d08c39bb607eb09b020146388e7c0bfceceb3b/ivy_tests/test_ivy/helpers/function_testing.py#L1219-L1222

args_constructor and kwargs_constructor may not be ordered (I don't think our Container is ordered in the early python3s? It only inherits from Dict), and later we may have to pass the arguments in the correct order to another process.

ivy_tests/test_ivy/test_misc/test_array.py
xoiga123: real_and_complex is numeric + complex so it's widening the data type

ivy_tests/test_ivy/test_misc/test_array.py
xoiga123: My bad, I'll look into this, `asarray` is called with torch backend somewhere and torch.as_tensor really said `TypeError: can't convert np.ndarray of type bfloat16.` :skull:

fspyridakos:
Sorry @xoiga123, was looking into it but was taking my time to fully go over it. Could you sync your brach with upstream to undo the git force push commits?

PR #11784, opened 2023-03-06
Throwing warning if frameworks get imported on importing ivy 

soma2000-lang:
Solves https://trello.com/c/y7PbQlak opened by @VedPatwardhan

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -4,15 +4,6 @@ import warnings
 from ivy._version import __version__ as __version__
 import builtins
 import numpy as np
-import importlib
-
-
-modules = ["jax", "tensorflow", "torch"]
-
-
-for module in modules:
-    if importlib.find_spec(module) is not None:
-        warnings.warn(f"{module} module detected,only Numpy should be imported")
 
 
 warnings.filterwarnings("ignore", module="^(?!.*ivy).*$")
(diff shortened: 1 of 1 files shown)


ivy/__init__.py
```diff
@@ -4,6 +4,15 @@
 from ivy._version import __version__ as __version__
 import builtins
 import numpy as np
+import importlib
+
+
+modules = ["jax", "tensorflow", "torch"]
```
VedPatwardhan, line 10: The logic looks great, it would be better if we get the modules programmatically so that even if there's a new backend added, it would be taken into account (e.g. paddle and mindspore are being added and we'd like modules to contain those without having to explicitly add it to the list). Secondly, we should do this check at the end of the `__init__.py` file so that even if any other file imported in this file does a framework import, we'd be able to check for it

VedPatwardhan (changes requested):
Hey @soma2000-lang, thanks for making the changes, just a couple more changes suggested 🙂

VedPatwardhan (changes requested):
Hey @soma2000-lang, I don't think we should be doing the checks with the `ivy.backend_stack` as it contains the backends that were set, not the frameworks that were imported. Your previous implementation with `importlib.find_spec` is great, the only thing to be updated further is to programmatically get the frameworks rather than hardcoding them as `modules = ['jax', 'numpy', 'tf', 'torch']`. Thanks 🙂

VedPatwardhan (approved):
Just made a small change, we can merge the PR now! Thanks @soma2000-lang 🙂

PR #11789, opened 2023-03-06
Updated ivy.linear to be a mixed function

hello-fri-end:
[Trello Task for linear](https://trello.com/c/EoB4SFES/1350-improve-ivylinear)
[Trello Task for Partial Handling of Mixed Functions](https://trello.com/c/5Glng7uK/1163-partial-handling-of-mixed-functions)


jkeane508:
Hey @VedPatwardhan,

Could you take a look at this? Or assign someone who would have a better idea than me :)

Thanks,
James

ivy/functional/backends/torch/layers.py
```diff
@@ -11,6 +11,52 @@
 from . import backend_version
 from ivy.functional.ivy.layers import _handle_padding, _deconv_length
 
+@with_unsupported_dtypes(
```
VedPatwardhan, line 14: We should be having the decorator to defer to the `ivy` function right when the lambda returns `True` right (as per discussed yesterday)?

ivy/functional/ivy/layers.py
```diff
@@ -131,6 +131,13 @@ def linear(
     }
     
     """
+    if ivy.current_backend(x).backend == 'torch':
```
VedPatwardhan, line 134: We shouldn't need to do this, once the backend is set to `torch`, `ivy.linear` would automatically defer to the `torch`-backend implementation as we discussed yesterday

ivy_tests/test_ivy/test_functional/test_nn/test_layers.py
```diff
@@ -51,7 +51,7 @@ def x_and_linear(draw, dtypes):
     dtype_x_weight_bias=x_and_linear(
         dtypes=helpers.get_dtypes("numeric", full=False),
     ),
-    ground_truth_backend="jax",
+    test_with_out=st.just(False),
```
VedPatwardhan, line 54: We shouldn't be setting `test_with_out=False` as it wouldn't run tests with the `out` argument

VedPatwardhan (changes requested):
Hey @jkeane508, thanks for requesting the review, I've suggested a few changes, but given that @Ishticode would be overseeing this feature, I'll also assign him to the PR. Thanks for looking into this @hello-fri-end 🙂

hello-fri-end:
> Hey @jkeane508, thanks for requesting the review, I've suggested a few changes, but given that @Ishticode would be overseeing this feature, I'll also assign him to the PR. Thanks for looking into this @hello-fri-end slightly_smiling_face

Hey @VedPatwardhan! Thanks for taking a look into this. I have added the `handle_mixed_functions` decorator as discussed in the meeting. It's not yet working as expected but I wanted to keep you updated on what I have done till now.

Ishticode (commented):
Hi @hello-fri-end 
Just leaving a comment to say that feel free to request a review when you like and don't hesitate to tag any of us for help. 

With this it might also suggest writing a test for this decorator (JUST pytest, NO hypothesis) in `ivy_tests/test_ivy/test_misc/test_func_wrapper.py`. A simple test would suffice to not only ensure the wrapper works as expected but also help clarify the objective. 

Thank you very much Anwaar.

hello-fri-end:
> Hi @hello-fri-end Just leaving a comment to say that feel free to request a review when you like and don't hesitate to tag any of us for help.
> 
> With this it might also suggest writing a test for this decorator (JUST pytest, NO hypothesis) in `ivy_tests/test_ivy/test_misc/test_func_wrapper.py`. A simple test would suffice to not only ensure the wrapper works as expected but also help clarify the objective.
> 
> Thank you very much Anwaar.

Hello @Ishticode! Thanks for a taking a look at this. I am actually done with the implementation - I have incorporated all the changes suggested by you and Ved in the meeting. Once the interpolate function is merged , I will try to apply the decorator on it too. However, there is something I wanted to discuss at this time: So, currently I have written this decorator for the `ivy.functional` functions - they way it works is that,  for backends that have a primary implementation, when the backend is set we'll store the compositonal function reference as an attribute inside the primary function 

```python
                    to_wrap.compos = original.__closure__[1].cell_contents
```

and if the lambda function evaluates to false, we'll run the compositional implementation using this attribute inplace of the primary implementation. I was wondering if it makes more sense to create the decorator for the backend implementation directly, again storing the compositional reference as an attribute during backend setting. Do you think this approach makes more sense? 

Also sure yes, I will a test for the decorator shortly. Thanks

hello-fri-end:
> > Hi @hello-fri-end Just leaving a comment to say that feel free to request a review when you like and don't hesitate to tag any of us for help.
> > With this it might also suggest writing a test for this decorator (JUST pytest, NO hypothesis) in `ivy_tests/test_ivy/test_misc/test_func_wrapper.py`. A simple test would suffice to not only ensure the wrapper works as expected but also help clarify the objective.
> > Thank you very much Anwaar.
> 
> Hello @Ishticode! Thanks for a taking a look at this. I am actually done with the implementation - I have incorporated all the changes suggested by you and Ved in the meeting. Once the interpolate function is merged , I will try to apply the decorator on it too. However, there is something I wanted to discuss at this time: So, currently I have written this decorator for the `ivy.functional` functions - they way it works is that, for backends that have a primary implementation, when the backend is set we'll store the compositonal function reference as an attribute inside the primary function
> 
> ```python
>                     to_wrap.compos = original.__closure__[1].cell_contents
> ```
> 
> and if the lambda function evaluates to false, we'll run the compositional implementation using this attribute inplace of the primary implementation. I was wondering if it makes more sense to create the decorator for the backend implementation directly, again storing the compositional reference as an attribute during backend setting. Do you think this approach makes more sense?
> 
> Also sure yes, I will a test for the decorator shortly. Thanks

Update: I experimented with the latter approach in another branch. It works as expected and it really simplifies the decorator and backend logic. Here is how it looks:

[`ivy.functional.backends.torch.linear`](https://github.com/hello-fri-end/ivy/blob/experimentBackendMIxedHandling/ivy/functional/backends/torch/layers.py#L18)
[`_wrap_function`](https://github.com/hello-fri-end/ivy/blob/experimentBackendMIxedHandling/ivy/func_wrapper.py#L700-L704)
[`handle_mixed_functions`](https://github.com/hello-fri-end/ivy/blob/experimentBackendMIxedHandling/ivy/func_wrapper.py#L920-L931)

@Ishticode : What do you think about this approach? Should I merge the new branch into this one?

hello-fri-end:
> > > Hi @hello-fri-end Just leaving a comment to say that feel free to request a review when you like and don't hesitate to tag any of us for help.
> > > With this it might also suggest writing a test for this decorator (JUST pytest, NO hypothesis) in `ivy_tests/test_ivy/test_misc/test_func_wrapper.py`. A simple test would suffice to not only ensure the wrapper works as expected but also help clarify the objective.
> > > Thank you very much Anwaar.
> > 
> > 
> > Hello @Ishticode! Thanks for a taking a look at this. I am actually done with the implementation - I have incorporated all the changes suggested by you and Ved in the meeting. Once the interpolate function is merged , I will try to apply the decorator on it too. However, there is something I wanted to discuss at this time: So, currently I have written this decorator for the `ivy.functional` functions - they way it works is that, for backends that have a primary implementation, when the backend is set we'll store the compositonal function reference as an attribute inside the primary function
> > ```python
> >                     to_wrap.compos = original.__closure__[1].cell_contents
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and if the lambda function evaluates to false, we'll run the compositional implementation using this attribute inplace of the primary implementation. I was wondering if it makes more sense to create the decorator for the backend implementation directly, again storing the compositional reference as an attribute during backend setting. Do you think this approach makes more sense?
> > Also sure yes, I will a test for the decorator shortly. Thanks
> 
> Update: I experimented with the latter approach in another branch. It works as expected and it really simplifies the decorator and backend logic. Here is how it looks:
> 
> [`ivy.functional.backends.torch.linear`](https://github.com/hello-fri-end/ivy/blob/experimentBackendMIxedHandling/ivy/functional/backends/torch/layers.py#L18) [`_wrap_function`](https://github.com/hello-fri-end/ivy/blob/experimentBackendMIxedHandling/ivy/func_wrapper.py#L700-L704) [`handle_mixed_functions`](https://github.com/hello-fri-end/ivy/blob/experimentBackendMIxedHandling/ivy/func_wrapper.py#L920-L931)
> 
> @Ishticode : What do you think about this approach? Should I merge the new branch into this one?

Update: Also tested the decorator on the `interpolate` function. It works as expected. Will push changes when we decide whether to keep the decorator for compositional functions or the backend implementation directly.

Ishticode (changes requested):
Hi @hello-fri-end. Great work! This seems to work as expected.
Could you maybe write a (simple) test. Perhaps same case here for can also be used for test.
Let me know if anything is unclear.

hello-fri-end:
> > > Hi @hello-fri-end Just leaving a comment to say that feel free to request a review when you like and don't hesitate to tag any of us for help.
> > > With this it might also suggest writing a test for this decorator (JUST pytest, NO hypothesis) in `ivy_tests/test_ivy/test_misc/test_func_wrapper.py`. A simple test would suffice to not only ensure the wrapper works as expected but also help clarify the objective.
> > > Thank you very much Anwaar.
> > 
> > 
> > Hello @Ishticode! Thanks for a taking a look at this. I am actually done with the implementation - I have incorporated all the changes suggested by you and Ved in the meeting. Once the interpolate function is merged , I will try to apply the decorator on it too. However, there is something I wanted to discuss at this time: So, currently I have written this decorator for the `ivy.functional` functions - they way it works is that, for backends that have a primary implementation, when the backend is set we'll store the compositonal function reference as an attribute inside the primary function
> > ```python
> >                     to_wrap.compos = original.__closure__[1].cell_contents
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and if the lambda function evaluates to false, we'll run the compositional implementation using this attribute inplace of the primary implementation. I was wondering if it makes more sense to create the decorator for the backend implementation directly, again storing the compositional reference as an attribute during backend setting. Do you think this approach makes more sense?
> > Also sure yes, I will a test for the decorator shortly. Thanks
> 
> Update: I experimented with the latter approach in another branch. It works as expected and it really simplifies the decorator and backend logic. Here is how it looks:
> 
> [`ivy.functional.backends.torch.linear`](https://github.com/hello-fri-end/ivy/blob/experimentBackendMIxedHandling/ivy/functional/backends/torch/layers.py#L18) [`_wrap_function`](https://github.com/hello-fri-end/ivy/blob/experimentBackendMIxedHandling/ivy/func_wrapper.py#L700-L704) [`handle_mixed_functions`](https://github.com/hello-fri-end/ivy/blob/experimentBackendMIxedHandling/ivy/func_wrapper.py#L920-L931)
> 
> @Ishticode : What do you think about this approach? Should I merge the new branch into this one?

Hi @Ishticode! Can you take a look at this approach please? I personally feel that it makes things much simpler. Let me know your thoughts :)

Also about the test, I am not able to understand what kind of test we would write for this decorator. I looked at the tests in ` ivy_tests/test_ivy/test_misc/test_func_wrapper.py` to get some ideas but this particular decorator requires a function to have a  primary and compositional implementation inside of Ivy and testing that the correct one gets called based on the input is already tested in the unit tests. Is there any other way we can tests this function?

Ishticode (changes requested):
Hi @hello-fri-end 
Sorry I missed this. This other implementation definitely is simpler and it does work in all cases.
You could write a test using ivy.linear function as input. Include different cases of weight matrices and backends. Ensure using assert statements that the primary function from torch backend is called when the conditions are met and not called (instead compositional is called) otherwise.

For now, you can just use typehints as a quick way to determine if the call is to the compositional or torch backend primary. Such as (because "Tensor" string is only found in torch backend implementatio): 
```python
import inspect
params_str = str(inspect.signature(ivy.linear).parameters.values())
assert ("Tensor" in params_str)
# OR
assert ("Tensor" not in params_str)
``` 
using inspect module.

This is not exhaustive in any sense of the word but ensures that no breaking changes are made to the decorator on the simplest level.

Hope this makes sense.

hello-fri-end:
> ```python
> import inspect
> params_str = str(inspect.signature(ivy.linear).parameters.values())
> assert ("Tensor" in params_str)
> # OR
> assert ("Tensor" not in params_str)
> ```
> 
> using inspect module.
> 
> This is not exhaustive in any sense of the word but ensures that no breaking changes are made to the decorator on the simplest level.

The problem with this approach is that it will always return `True` since the decision of running the primary implementation or the compositional one is made after the function is called with some parameters. 

I experimented with a few other approaches and finally figured out how to write the test using python's `unittest.mock` module.  `Mock` allows you to substitute and imitate an object in python so that you can track it. So, I basically "mocked" `'torch.nn.functional.linear'` for the torch backend so that we can track whether `ivy.linear` makes a call to it. I have added two inputs, first a 2D weight matrix and second a 3D weight. So, in the first case `test_mock_function.called` should return `True` and in the second case, it should return `False`. For other backends, I am mocking `ivy.matmul` since it's used in the compositional implementation -- and hence `test_mock_function.called` should return `True` for both cases. Hope this makes sense. Please let me know if I should clarify anything further.

Everything is done here - this is ready for review @Ishticode @VedPatwardhan

Ishticode (approved):
@hello-fri-end 
Nice one with that test. Indeed the right way to test here. 
I will get this merged now. Feel free to iterate if you feel the need.
Thank you so much Anwaar for the effort and this contribution 🙂

PR #11807, opened 2023-03-06
Reducing space of the CPU docker image

ZiadAmerr:


ZiadAmerr:
Do not merge this now, I am still working on it :)

ZiadAmerr:
@Dsantra92 I think the latest commit should make this work on x86 machines, right?

kurshakuz:
I am looping in @Dsantra92 for a final approval then

Dsantra92 (changes requested):

docker/Dockerfile
```diff
@@ -13,11 +13,11 @@ RUN apt-get update && \
     pip3 install --upgrade pip && \
     pip3 install setuptools==58.5.3
 
-RUN pip3 install torch -f https://download.pytorch.org/whl/torch_stable.html
+RUN pip3 install torch==1.13.1 torchvision==0.14.1 torchaudio==0.13.1 -f https://download.pytorch.org/whl/torch_stable.html
 RUN pip3 install --upgrade torch-scatter -f https://pytorch-geometric.com/whl
 
 # Install Ivy Upstream
-RUN git clone --progress --recurse-submodules https://github.com/unifyai/ivy && \
+RUN git clone --progress --recurse-submodules https://github.com/unifyai/ivy --depth 1 && \
```
Dsantra92, line 20: No need to add --depth 1 again in this pr, it will cause a merge conflict with master.

docker/Dockerfile
```diff
@@ -13,11 +13,11 @@ RUN apt-get update && \
     pip3 install --upgrade pip && \
     pip3 install setuptools==58.5.3
 
-RUN pip3 install torch -f https://download.pytorch.org/whl/torch_stable.html
+RUN pip3 install torch==1.13.1 torchvision==0.14.1 torchaudio==0.13.1 -f https://download.pytorch.org/whl/torch_stable.html
```
Dsantra92, line 16: Install torch from cpu channel. See the [official docs](https://pytorch.org/get-started/locally/) for more details. Torch and scatter are installed after this pip install too. Have a look out for redundant libraries or libraries attempted to install twice.

Dsantra92 (changes requested):

docker/Dockerfile
Dsantra92: Also don't pin the version of torch here.

Dsantra92:
Jax and tensorflow could also use some space reduction here. Also libraries like nvidia-ml-py are not required for CPU.

docker/Dockerfile
ZiadAmerr: But what about [this discussion](https://discuss.pytorch.org/t/libtorch-cuda-so-is-too-large-2gb/103155/24) on PyTorch's community? I have resolved it and unpinned the version

ZiadAmerr:
> Jax and tensorflow could also use some space reduction here. Also libraries like nvidia-ml-py are not required for CPU.

But do they really take up as much space as torch does? I have seen 800MB of requirements installed when installing the docker image, maybe we can dive deeper into that and analyze what packages are installed, I will try to do that and provide a report to it

sagewhocodes:
Builds fine on Linux x86_64.
LGTM!

docker/Dockerfile
```diff
@@ -13,7 +13,8 @@ RUN apt-get update && \
     pip3 install --upgrade pip && \
     pip3 install setuptools==58.5.3
 
-RUN pip3 install --no-cache-dir torch -f https://download.pytorch.org/whl/torch_stable.html
+
+RUN pip3 install --no-cache-dir torch --extra-index-url https://download.pytorch.org/whl/cpu
 RUN pip3 install --no-cache-dir --upgrade torch-scatter -f https://pytorch-geometric.com/whl
```
Dsantra92, line 18:
```suggestion
RUN export ver=$(pip show torch | grep Version | cut -d ' ' -f2) && \
    pip3 install --no-cache-dir --upgrade torch-scatter -f https://data.pyg.org/whl/torch-$ver.html
```

Dsantra92 (changes requested):
Sorry to extend this longer than anticipated, but we can reduce the image even further without losing out on any of the functionality. Can we do `requirements_cpu.txt` and `optional_cpu.txt` for CPU-specific environments?

requirements/optional_cpu.txt
```diff
@@ -0,0 +1,21 @@
+h5py # unpinned
+pytest # unpinned
+networkx # unpinned
+hypothesis # unpinned
+pymongo # unpinned
+redis # unpinned
+matplotlib # unpinned
+opencv-python # unpinned , mod_name=cv2
+jax # unpinned, we test for latest version now
```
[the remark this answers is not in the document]
ZiadAmerr: Jax package doesn't include any extra GPU wheels.

> If you want to install JAX with both CPU and NVidia GPU support, you must first install [CUDA](https://developer.nvidia.com/cuda-downloads) and [CuDNN](https://developer.nvidia.com/CUDNN), if they have not already been installed. Unlike some other popular deep learning systems, JAX does not bundle CUDA or CuDNN as part of the pip package.

[Source](https://github.com/google/jax)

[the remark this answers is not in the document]
Dsantra92: Yep, but we are leaving out the GPU/TPU-specific components. You can read more about installing CPU-specific Jax [here](https://github.com/google/jax#pip-installation-cpu).

docker/Dockerfile
```diff
@@ -21,19 +21,21 @@ RUN export ver=$(pip show torch | grep Version | cut -d ' ' -f2) && \
...
     cat requirements/requirements.txt | grep -v "ivy-" | pip3 install --no-cache-dir -r /dev/stdin && \
-    cat requirements/optional.txt | grep -v "ivy-" | pip3 install --no-cache-dir -r /dev/stdin && \
+    cat requirements/optional_cpu.txt | grep -v "ivy-" | pip3 install --no-cache-dir -r /dev/stdin && \
     python3 -m pip install --user -e . && \
     cd ivy_tests/test_array_api && \
     pip3 install --no-cache-dir -r requirements.txt
 
 # Install local requirements
-COPY requirements/requirements.txt .
+RUN cp ivy/requirements/requirements.txt .
```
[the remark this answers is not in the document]
Dsantra92: COPY copies from the remote.

VedPatwardhan (changes requested):
I think it's better if we instead have the cpu-based libraries in the `optional.txt` as the default, and the `optional_gpu.txt` for installing the cuda-part of and update the `DockerfileGPU` accordingly. Could you please make that change, just following general convention? Rest looks good to me, thanks 🙂

sagewhocodes (approved):

Dsantra92 (approved):

PR #11812, opened 2023-03-06
 Conv2D #10505 

kaushal07wick:


kaushal07wick:
hey @kurshakuz , conv2d was passing all the tests before the git push -force, but now it shows this error
, negative value error.
```
ivy/functional/backends/numpy/layers.py:214: in conv2d
    sub_matrices = np.lib.stride_tricks.as_strided(

x = array([[[[0.4548],
         [0.458 ],
         [0.4548]],

        [[0.4548],
         [0.2467],
         [0.5   ]],

        [[0.5   ],
         [0.5376],
         [0.5   ]]]], dtype=float16)
shape = [1, -1, -1, 5, 5, 1], strides = (18, 6, 2, 6, 2, 2), subok = False
writeable = False

    def as_strided(x, shape=None, strides=None, subok=False, writeable=True):
        """
        Create a view into the array with the given shape and strides.
    
        .. warning:: This function has to be used with extreme care, see notes.
    
        Parameters
        ----------
        x : ndarray
            Array to create a new.
        shape : sequence of int, optional
            The shape of the new array. Defaults to ``x.shape``.
        strides : sequence of int, optional
            The strides of the new array. Defaults to ``x.strides``.
        subok : bool, optional
            .. versionadded:: 1.10
    
            If True, subclasses are preserved.
        writeable : bool, optional
            .. versionadded:: 1.12
    
            If set to False, the returned array will always be readonly.
            Otherwise it will be writable if the original array was. It
            is advisable to set this to False if possible (see Notes).
    
        Returns
        -------
        view : ndarray
    
        See also
        --------
        broadcast_to : broadcast an array to a given shape.
        reshape : reshape an array.
        lib.stride_tricks.sliding_window_view :
            userfriendly and safe function for the creation of sliding window views.
    
        Notes
        -----
        ``as_strided`` creates a view into the array given the exact strides
        and shape. This means it manipulates the internal data structure of
        ndarray and, if done incorrectly, the array elements can point to
        invalid memory and can corrupt results or crash your program.
        It is advisable to always use the original ``x.strides`` when
        calculating new strides to avoid reliance on a contiguous memory
        layout.
    
        Furthermore, arrays created with this function often contain self
        overlapping memory, so that two elements are identical.
        Vectorized write operations on such arrays will typically be
        unpredictable. They may even give different results for small, large,
        or transposed arrays.
    
        Since writing to these arrays has to be tested and done with great
        care, you may want to use ``writeable=False`` to avoid accidental write
        operations.
    
        For these reasons it is advisable to avoid ``as_strided`` when
        possible.
        """
        # first convert input to array, possibly keeping subclass
        x = np.array(x, copy=False, subok=subok)
        interface = dict(x.__array_interface__)
        if shape is not None:
            interface['shape'] = tuple(shape)
        if strides is not None:
            interface['strides'] = tuple(strides)
    
>       array = np.asarray(DummyArray(interface, base=x))
E       ValueError: negative dimensions are not allowed
E       Falsifying example: test_tensorflow_Conv2D(
E           x_f_d_df=(['float16'], array([[[[0.4548],
E                     [0.458 ],
E                     [0.4548]],
E            
E                    [[0.4548],
E                     [0.2467],
E                     [0.5   ]],
E            
E                    [[0.5   ],
E                     [0.5376],
E                     [0.5   ]]]], dtype=float16), array([[[[0.5]],
E            
E                    [[0.5]],
E            
E                    [[0.5]]],
E            
E            
E                   [[[0.5]],
E            
E                    [[0.5]],
E            
E                    [[0.5]]],
E            
E            
E                   [[[0.5]],
E            
E                    [[0.5]],
E            
E                    [[0.5]]]], dtype=float16), 'NHWC', [2], [1], 'VALID'),
E           test_flags=num_positional_args=0. with_out=False. inplace=False. native_arrays=[False]. as_variable=[False]. ,
E           fn_tree='ivy.functional.frontends.tensorflow.raw_ops.Conv2D',
E           frontend='tensorflow',
E           on_device='cpu',
E       )
E       
E       You can reproduce this example by temporarily adding @reproduce_failure('6.55.0', b'AXicY2BkYGAAYSQA5h4QEtEKFpvdDWKKPdJbePDBahBTcp5OHstlsKiUiU5hjjaYqcLnYDMnaj6IiTAGncm45wR+BTRgAgDtARv+') as a decorator on your test case

/usr/local/lib/python3.8/dist-packages/numpy/lib/stride_tricks.py:105: ValueError

```

PR #11816, opened 2023-03-06
histogram

### LINKED ISSUE #11814: histogram

joaozenobio:
Close #11814
ML intern application

### DIFF
--- a/docs/overview/deep_dive.rst
+++ b/docs/overview/deep_dive.rst
@@ -72,6 +72,9 @@ We're excited for you to get involved! 🦾
 |
 | (t) :ref:`Continuous Integration` 🔁
 | Ivy Tests running on the Repository
+|
+| (u) :ref:`Gradients` 🔁
+| Everything about our Gradients API
 .. toctree::
    :hidden:
    :maxdepth: -1
@@ -97,3 +100,4 @@ We're excited for you to get involved! 🦾
    deep_dive/ivy_frontends_tests.rst
    deep_dive/exception_handling.rst
    deep_dive/continuous_integration.rst
+   deep_dive/gradients.rst
--- a/docs/overview/deep_dive/gradients.rst
+++ b/docs/overview/deep_dive/gradients.rst
@@ -1,0 +1,174 @@
+Gradients
+=========
+
+.. _`discord`: https://discord.gg/sXyFF8tDtm
+.. _`gradients channel`: https://discord.com/channels/799879767196958751/1000043921633722509
+.. _`gradients forum`: https://discord.com/channels/799879767196958751/1028299026501750826
+
+Overview
+--------
+
+Gradients are a crucial aspect of all modern deep learning workflows. 
+Different frameworks provide different APIs for gradient computation and there were a few considerations to be made while building a unified gradients API in Ivy.
+There are a number of functions added in ivy to allow gradient computation, but we'll mainly focus on the most commonly used and the most general function :func:`ivy.execute_with_gradients`.
+This is because the other gradient functions such as :func:`ivy.value_and_grad` and :func:`ivy.grad` can be considered as providing a subset of the functionality that :func:`ivy.execute_with_gradients` provides.
+
+Example Usage of the Gradient API
+---------------------------------
+
+The :func:`ivy.execute_with_gradients` function signature
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Following is pseudo function signature for the :func:`ivy.execute_with_gradients` function,
+
+.. code-block:: python
+    
+    def execute_with_gradients (
+        func : Callable,
+        xs : Any arbitrary nest,
+        xs_grad_idxs : Input indices,
+        ret_grad_idxs : Output indices,
+    ) : 
+        return func_ret, grads
+
+The :code:`func` in the input can be any user-defined function that returns a single scalar or any arbitrary nest of scalars.
+By scalars, we are referring to zero-dimensional arrays.
+
+So for example, following are some valid outputs by the :code:`func`,
+
+.. code-block:: python
+    
+    ivy.array(12.)
+    
+    # OR
+
+    ivy.Container(
+        a=ivy.array(12.), 
+        b=ivy.Container(
+            c=ivy.array(15.),
+            d=ivy.array(32.)
+        )
+    )
+
+    # OR
+
+    [ivy.array(25.), {'x': (ivy.array(21.), ivy.array(11.))}, (ivy.array(9.),)]
+
+:code:`xs` can be any arbitrary nest of arrays and refers to the inputs passed to the :code:`func`, so we suggest designing your :code:`func` based on what inputs you pass in :code:`xs`.
+The arrays in :code:`xs` can contain any arbitrary number of dimensions, the only constraint is on the output of the :code:`func` as explained above.
+
+The :code:`xs_grad_idxs` and :code:`ret_grad_idxs` are intended to provide more control over the arrays gradients are computed with.
+:code:`xs_grad_idxs` accepts the indices of the input arrays to compute gradients for, and :code:`ret_grad_idxs` accepts the indices of the output arrays to compute gradients with respect to.
+
+An example using :func:`ivy.execute_with_gradients`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: python
+
+    def func(xs) :
+        return ivy.mean(xs[0] + xs[1].b)
+
+    x = ivy.array([1., 2., 3.])
+    x = ivy.Container(a=x, b=x)
+    y = ivy.array([4., 5., 6.])
+    y = ivy.Container(b=y, c=x)
+    xs = [x, y]
+
+    ret, grads = ivy.execute_with_gradients(
+        func, 
+        xs, 
+        xs_grad_idxs=[[0]],
+        ret_grad_idxs=[["a"]]
+    )
+
+
+Custom Gradient Functions
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are various scenarios where users may want to define custom gradient computation rules for their functions.
+Some of these are numerical stability, smoothing and clipping of the computed gradients.
+Ivy provides the :func:`ivy.bind_custom_gradient_function` function to allow users to bind custom gradient computation logic to their functions.
+
+Following is an example of usage of :func:`ivy.bind_custom_gradient_function`,
+
+.. code-block:: python
+
+    import ivy
+
+    ivy.set_backend("torch")
+    x = ivy.array(50.0)
+    inter_func = lambda x: ivy.log1p(ivy.exp(x))
+
+    # args –> ((xs, ret), upstream)
+    def custom_grad_fn(*args):
+        args1 = (1 - 10 / (1 + args[0][0]))
+        return (args[1] * args)
+
+    inter_func = ivy.bind_custom_gradient_function(
+    inter_func, custom_grad_fn
+    )
+    func = lambda x: ivy.sum(inter_func(x) ** 2)
+
+    ret, grad = ivy.execute_with_gradients(func, x)
+
+The :code:`custom_grad_fn` here accepts :code:`*args` which has the structure :code:`((xs, ret), upstream)` where,
+* :code:`xs` is the input similar to the one accepted in :func:`ivy.execute_with_gradients`
+* :code:`ret` is the output of the forward pass of the :func:`inter_func`
+* :code:`upstream` refers to the previously computed gradients while back-propagating
+
+
+Design of the Gradient API
+--------------------------
+
+Our policy on gradients
+^^^^^^^^^^^^^^^^^^^^^^^
+
+* The gradient API is fully-functional in ivy.
+* There is no explicit Variable class or any public-facing function for adding gradient support to an ivy.Array.
+* The gradient functions in ivy implicitly convert all arrays to support gradient computation before computing gradients and detach all arrays after computing gradients.
+* We don't retain any previously tracked computations in arrays by frameworks like torch for e.g. 
+* This makes our gradient API disambiguous, flexible and easy to debug.
+* Any framework-specific tracking of computations or variable classes should be handled in the corresponding frontends.
+
+Gradient APIs of frameworks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. list-table:: Frameworks and their gradient functions
+   :widths: 25 25 50
+   :header-rows: 1
+
+   * - Framework
+     - Common ways to Gradient Computation
+   * - JAX
+     - `jax.grad`, `jax.value_and_grad`, `jax.jacfwd`, `jax.jacrev`
+   * - PyTorch
+     - `torch.autograd.grad`, `torch.autograd.backward`
+   * - TensorFlow
+     - `tf.GradientTape`, `tf.gradients` (only in graph-mode)
+
+General Structure of Backend-specific implementations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Here's a high-level description of the steps followed backend-specific implementation of :func:`ivy.execute_with_gradients`
+1. Get Duplicate Index Chains : indices of arrays that share the same :code:`id`
+2. Convert integer arrays to floats : only for ease of use. it's *not* recommended to pass integer arrays to gradient functions
+3. Get relevant inputs : based on the :code:`xs_grad_idxs`, we collect the relevant inputs for gradient computation
+4. Enable gradient support : we implicitly make use of framework-specific APIs to enable gradients in arrays. Ivy doesn't need to have an explicit variable class as the gradient API is fully functional
+5. Compute Results : we do the forward pass by passing the input as it is to the function
+6. Get relevant outputs : based on the :code:`ret_grad_idxs`, we collect the relevant outputs for gradient computation
+7. Compute gradients : we make use of the framework-specific APIs to compute the gradients for the relevant outputs with respect to the relevant inputs
+8. Handle duplicates : we explicitly handle duplicate instances using the index chains captured above as different frameworks treat duplicates differently
+9. Post process and detach : finally, all computed gradients are updated to deal with :code:`NaN` and :code:`inf` and the input arrays are detached (i.e. gradient propagation is stopped)
+
+Framework-specific Considerations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* JAX treats duplicate arrays as distinct while computing gradients, so we need additional logic to replicate gradients computed w.r.t one array over all its duplicates.
+* Gradients computed for functions with undefined results are inconsistent across backends (NaN, Inf, 0). We handle all these inconsistencies by returning 0 for all backends. So if you’re debugging gradients and find a 0, there’s a possibility that it was NaN or an Inf before computing.
+
+
+**Round Up**
+
+This should have hopefully given you a good feel for how the gradient API is implemented in Ivy.
+
+If you have any questions, please feel free to reach out on `discord`_ in the `gradients channel`_ or in the `gradients forum`_!


joaozenobio:
- [ ] #11814

MuhammadSaeedBatikh:
Hi @joaozenobio, thanks for your PR.
Please make sure that tests from test_statstical pass locally for your histogram function. 

Here are a few problems:
1-handling `bins`:
Though histogram is supposed to accept `int` as a valid type for bins, your implementation does not account for that. For example, in the pytorch backend you have the following lines:
`bins_out = bins.clone()`  
which will not work for `int`.
Furthermore, since you seem to be partially following tfp.stats.histogram API, `bins` could be a tensor:
but in numpy for example you have the following:
`bins = np.linspace(start=range[0], stop=range[1], num=bins + 1, dtype=a.dtype)`
which will not work for array-like bins.

2- some test results do not match:
`AssertionError:  the results from backend numpy and ground truth framework tensorflow do not match`

joaozenobio:
Hi @MuhammadSaeedBatikh! Thanks for the response. I have modified the repo, including exceptions to handle the possible alternatives to the bins argument. If bins is an int now, the function will only accept the argument range to be different than None. Then bins will be transformed into an array of bin edges for the necessary modifications according to the other arguments of the histogram function.

I could get two of the backends working. However, I am getting an error I could not fix yet. Numpy, Jax, and Pytorch, when working with a list of bins edges of, for example, [1., 2., 3.] and input [1., 2., 3., 4.], consider the value 3 of the input and add it to the last bin, differently from TensorFlow, that should do the same but does not consider it instead. I have made the same test but with bin edges =  [1., 2., 3., 4.] and input = [1., 2., 3., 4.] and the behavior of TensorFlow changed, including 4 in the last bin. It looks like a bug in the Tensorflow function, at least it does not make sense. How should I workaround it? Should I post an issue in Tensorflow?


Example:
"""
import ivy

ivy.set_backend('tensorflow')

a = ivy.array([1., 2., 3., 4.], dtype=ivy.float32)
bins = ivy.array([1., 2., 3.], dtype=ivy.float32)
axis = 0
extend_lower_interval = False
extend_upper_interval = False
dtype = ivy.float32
range = None
weights = None
density = False
out = None

ret = ivy.histogram(a, bins=bins, axis=axis,
                    extend_lower_interval=extend_lower_interval,
                    extend_upper_interval=extend_upper_interval,
                    dtype=dtype, range=range, weights=weights,
                    density=density, out=out)
print(ret)

>>> (ivy.array([1., 1.]), ivy.array([1., 2., 3.]))

# Modifying bins and running again.
bins = ivy.array([1., 2., 3., 4.], dtype=ivy.float32)

ret = ivy.histogram(a, bins=bins, axis=axis,
                    extend_lower_interval=extend_lower_interval,
                    extend_upper_interval=extend_upper_interval,
                    dtype=dtype, range=range, weights=weights,
                    density=density, out=out)

>>> (ivy.array([1., 1., 2.]), ivy.array([1., 2., 3., 4.]))
"""

MuhammadSaeedBatikh:
> 

Hi @joaozenobio,
I find it really difficult to follow your explanation. 
As far as I understand you think that TensorFlow behavior **should not change** in the two cases that you present? Please correct me if I got it wrong.
At any rate, regarding your code example:
The output should change ... but not in this way.

The first `ret` is incorrect; the answer should be `[1, 2]` and not `[1 , 1]`
`tfp.stats.histogram` follows the following convention for "bucketing" bins/edges:
"all intervals are half-open with the exception of the last one". Which is the same convention used in Numpy, Jax, and PyTorch.
That is to say that, given following:
```
data= [1,2,3,4]
bins= [1,2,3]
some_backend.histogram(data, bins=bins)
```
The histogram function will convert bins to the following buckets/intervals: `[ [1,2), [2,3] ]`
notice that the first one is half-open and the second one is closed.
then, it will count how many elements fall in these buckets:
In this example, we have **one** element in the first interval `[1,2)`, which is the value 1 from data, but we have **two**  elements in the second interval `[2,3]` since 2 and 3 are in this interval. Therefore, the answer should be `[1, 2]`  and not `[1 , 1]` as shown in your code example.

Similarly, the second `ret` value in your example is actually correct. We have three buckets in this case:
 `[ [1,2), [2,3), [3,4] ]`
which indeed should produce the following histogram: `[1, 1, 2]`.

MuhammadSaeedBatikh:
Another comment regarding the way in which you handled the `range` parameter.
```
 if isinstance(bins, int) and range is None:
        raise ivy.exceptions.IvyException("range must be specified when bins is int")
```
This needs not be the case! The backends we support do not require the user to specify range if `bins` is `int`; instead,  `range` could be inferred from the min and max from the input tensor if the user does not specify it.

joaozenobio:
Hi @MuhammadSaeedBatikh, sorry for the last text, next time I will try to explain it better and with more details. I have followed your explanation and found that it was my TensorFlow version of the histogram function that caused this error.

From the TensorFlow documentation: "Values of x outside of the intervals cause errors. Consider using extend_lower_interval, extend_upper_interval to deal with this."

I tried to remove the values that did not fall inside the range of the buckets/intervals so TensorFlow would not throw an error because of it, creating two new buckets/intervals to collect the values from the input values that would cause this error: one at the beginning, to collect the input values that fall before the range of the first original bucket/interval, and one at the end, to collect the input values that fall after the range of the first original bucket/interval, and then I would delete them, preserving only the values that fall within the range of the original buckets/intervals. But when I do this, as you explained, the last bucket/interval would have a half-opened interval, and this changes the results.

Following this example:

```
data= [1,2,3,4]
bins= [2,3]
... # <- where I modify bins
some_backend.histogram(data, bins=bins)
```

The histogram function will convert bins to the following buckets/intervals: [ [1,2), [2,3] ]. And then, when I added the two new intervals to collect the values that do not fall in the range of [1, 3], the modified buckets/intervals are: [ [-1, 1) [1,2), [2,3), [3, 4] ]. After running the function, I would delete the first two intervals of the result and I would have the input values that fall within: [ [1,2), [2,3) ]. Which leads to an incorrect result.

Therefore, should I follow the TensorFlow implementation and, for all the other backends, not allow values to fall out of the range of the buckets/intervals, giving the possibility to use the extend_lower_interval and extend_upper_interval arguments to deal with this.?

Thanks in advance.

joaozenobio:
> Another comment regarding the way in which you handled the `range` parameter.
> 
> ```
>  if isinstance(bins, int) and range is None:
>         raise ivy.exceptions.IvyException("range must be specified when bins is int")
> ```
> 
> This needs not be the case! The backends we support do not require the user to specify `range` if `bins` is `int`; instead, `range` could be inferred from input tensor's min and max.

Regarding this problem, I will make the necessary modifications.

joaozenobio:
Also, I have tried to specify more than one axis on TensorFlow, but it gave me an error.

```
tfp.stats.histogram(tf.constant([2., 3.]), tf.constant([1., 2., 3., 4.]), 
                    axis=tf.constant([0, 1]))
2023-03-09 21:04:36.984945: W tensorflow/core/framework/op_kernel.cc:1745] OP_REQUIRES failed at transpose_op.cc:142 : INVALID_ARGUMENT: transpose expects a vector of size 1. But input(1) is a vector of size 2
  File "/opt/.pycharm_helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/tensorflow_probability/python/stats/quantiles.py", line 387, in histogram
    x = _move_dims_to_flat_end(x, axis, x_ndims, right_end=False)
  File "/usr/local/lib/python3.8/dist-packages/tensorflow_probability/python/stats/quantiles.py", line 886, in _move_dims_to_flat_end
    x_permed = tf.transpose(a=x, perm=perm)
  File "/usr/local/lib/python3.8/dist-packages/tensorflow/python/util/traceback_utils.py", line 153, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/usr/local/lib/python3.8/dist-packages/tensorflow/python/eager/execute.py", line 54, in quick_execute
    tensors = pywrap_tfe.TFE_Py_Execute(ctx._handle, device_name, op_name,
tensorflow.python.framework.errors_impl.InvalidArgumentError: transpose expects a vector of size 1. But input(1) is a vector of size 2 [Op:Transpose]
```

joaozenobio:
I have also tried to use multiple axis in TensorFlow and it gave me this error:

```
tfp.stats.histogram(tf.constant([2., 3.]), tf.constant([1., 2., 3., 4.]), 
                    axis=tf.constant([0, 1]))
2023-03-09 21:04:36.984945: W tensorflow/core/framework/op_kernel.cc:1745] OP_REQUIRES failed at transpose_op.cc:142 : INVALID_ARGUMENT: transpose expects a vector of size 1. But input(1) is a vector of size 2
  File "/opt/.pycharm_helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/tensorflow_probability/python/stats/quantiles.py", line 387, in histogram
    x = _move_dims_to_flat_end(x, axis, x_ndims, right_end=False)
  File "/usr/local/lib/python3.8/dist-packages/tensorflow_probability/python/stats/quantiles.py", line 886, in _move_dims_to_flat_end
    x_permed = tf.transpose(a=x, perm=perm)
  File "/usr/local/lib/python3.8/dist-packages/tensorflow/python/util/traceback_utils.py", line 153, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/usr/local/lib/python3.8/dist-packages/tensorflow/python/eager/execute.py", line 54, in quick_execute
    tensors = pywrap_tfe.TFE_Py_Execute(ctx._handle, device_name, op_name,
tensorflow.python.framework.errors_impl.InvalidArgumentError: transpose expects a vector of size 1. But input(1) is a vector of size 2 [Op:Transpose]
```

MuhammadSaeedBatikh:
> 
Hi @joaozenobio,
Good question. I'm more inclined to say, no, we don't need to fix that. 
This exists by design in the tfp implementation. And it also handles missing upper and lower limits asymmetrically. On the other hand, Numpy, Pytorch, and Jax handle all of these cases just fine.
If the workaround were easy, I'd say fine, let's fix it.  But in this case, the fix that you describe seems a little bit complex, both in terms of implementation and performance. Not to mention, it will be even more complicated when working with a list of axes!
I guess a good middle ground is, not to downgrade the other backends, but prompt the user when using tf backend with a meaningful exception similar to the one you mentioned: "TensorFlow does not allow this, consider using .. etc."

joaozenobio:
> Hi @joaozenobio, Good question. I'm more inclined to say, no, we don't need to fix that. This exists by design in the tfp implementation. And it also handles missing upper and lower limits asymmetrically. On the other hand, Numpy, Pytorch, and Jax handle all of these cases just fine. If the workaround were easy, I'd say fine, let's fix it. But in this case, the fix that you describe seems a little bit complex, both in terms of implementation and performance. Not to mention, it will be even more complicated when working with a list of axes! I guess a good middle ground is, not to downgrade the other backends, but prompt the user when using tf backend with a meaningful exception similar to the one you mentioned: "TensorFlow does not allow this, consider using .. etc."

Hi @MuhammadSaeedBatikh ! I have made the necessary modifications to solve this problem. But I am now receiving an error I cannot repoduce manually.

The error is attached. It looks like tensorflow is returning the modified variable "bins" from numpy instead of returning his own "bins" variable. I have observed this because in the numpy version of histogram function I modify the variable "bins", and the modifications occur in the tensorflow version. The same occurs with pytorch. I don't understand exactly what is happening.

[error.txt](https://github.com/unifyai/ivy/files/10949908/error.txt)

joaozenobio:
Hi @MuhammadSaeedBatikh, I have made the necessary modifications for the backends to accept multiple axes.

Regarding the latter error I am not undestanding, another example of the same error is attched with even stranger problems, it looks like the test script is comparing results from diferent executions. Sorry for the prints in the middle of the error description.

[error2.txt](https://github.com/unifyai/ivy/files/10951118/error2.txt)

MuhammadSaeedBatikh:
> 
Hi @joaozenobio, 
This is pre

> Hi @MuhammadSaeedBatikh, I have made the necessary modifications for the backends to accept multiple axes.
> 
> Regarding the latter error I am not undestanding, another example of the same error is attched with even stranger problems, it looks like the test script is comparing results from diferent executions. Sorry for the prints in the middle of the error description.
> 
> [error2.txt](https://github.com/unifyai/ivy/files/10951118/error2.txt)

Hi @joaozenobio, 
This is standard value testing. Basically, we are testing whether the histogram function returns the same result for all backends.
In your case, the log says that the Jax backend is returning incorrect results for a particular example.
Try the following example:

```
import ivy
ivy.set_backend('jax')
ret_jax = ivy.histogram(ivy.array([0, 0.]), bins=ivy.array([1, 2, 3.]), axis=0, extend_lower_interval=True,  extend_upper_interval= True, weights = ivy.array([0, 1.]))
print(ret_jax)
#>>> (ivy.array([0., 1.]), ivy.array([1., 2., 3.]))
``` 
Which is wrong! it should be the reverse of that.
if you set your backend to tensorflow:

```
import ivy
ivy.set_backend('tensorflow')
ret_tf= ivy.histogram(ivy.array([0, 0.]), bins=ivy.array([1, 2, 3.]), axis=0, extend_lower_interval=True,  extend_upper_interval= True, weights = ivy.array([0, 1.]))
print(ret_tf)
#>>> (ivy.array([1., 0]), ivy.array([1., 2., 3.]))
``` 
Which is correct.

joaozenobio:
Hi @MuhammadSaeedBatikh, I have solved the latter problem. However, there are still strange results that I cannot reproduce.

Obs: the resulting bins argument for TensorFlow, which is the problem, does not match the results when I run it in a separate script. Also, I have observed that its result inside the test function is matching the modified bins I use in both Numpy and Pytorch like it is modifying the same variable in memory.

```
E           AssertionError:  the results from backend torch and ground truth framework tensorflow do not match
E            [0. 1.]!=[-1.  1.] 
E           
E           
E           Falsifying example: test_histogram(
E               values=(array([-1., -1.], dtype=float32),
E                array([0., 1.], dtype=float32),
E                0,
E                True,
E                False,
E                'float32',
E                None,
E                array([-1., -1.], dtype=float32),
E                False,
E                'float32'),
E               test_flags=num_positional_args=1. with_out=False. instance_method=False. native_arrays=[False]. container=[False]. as_variable=[False]. test_gradients=False. test_compile=False. ,
E               fn_name='histogram',
E               ground_truth_backend='tensorflow',
E               backend_fw=<module 'ivy.functional.backends.torch' from '/ivy/ivy/functional/backends/torch/__init__.py'>,
E           )
E           
E           You can reproduce this example by temporarily adding 
```

```
E           AssertionError:  the results from backend numpy and ground truth framework tensorflow do not match
E            [2 3]!=[1 3] 
E           
E           
E           Falsifying example: test_histogram(
E               values=(array([1, 1], dtype=uint8),
E                array([2, 3], dtype=uint8),
E                0,
E                True,
E                False,
E                'uint8',
E                None,
E                array([1, 1], dtype=uint8),
E                False,
E                'uint8'),
E               test_flags=num_positional_args=1. with_out=False. instance_method=False. native_arrays=[False]. container=[False]. as_variable=[False]. test_gradients=False. test_compile=False. ,
E               fn_name='histogram',
E               ground_truth_backend='tensorflow',
E               backend_fw=<module 'ivy.functional.backends.numpy' from '/ivy/ivy/functional/backends/numpy/__init__.py'>,
E           )
E           
E           You can reproduce this example by temporarily adding @reproduce_failure('6.55.0', b'AXicY2BgZGRiZMACAACKAAY=') as a decorator on your test case
```

MuhammadSaeedBatikh:
Hi @joaozenobio, 
There seems to be a fundamental problem with the shape of the computed histogram for all backends.
bins of size 2 should return a histogram of size 1
For example:
```
from numpy import array, histogram
x = array([1, 2.])
bins = array([2., 3])
histogram(x, bins)
#>>> (array([1], dtype=int64), array([2, 3]))
```

joaozenobio:
Hi @MuhammadSaeedBatikh, yes the problem is with bins in this case. However, when I run the function in a script or directly in a python console the results match what is expected, but when I try to test the function inside the pytest, it seems the bins returned by TensorFlow is being altered by the Numpy function like they are altering the same space in memory.

By testing with the same inputs from the error pytest gave me before in...

> E           AssertionError:  the results from backend torch and ground truth framework tensorflow do not match
> E            [0. 1.]!=[-1.  1.] 
> E           
> E           
> E           Falsifying example: test_histogram(
> E               values=(array([-1., -1.], dtype=float32),
> E                array([0., 1.], dtype=float32),
> E                0,
> E                True,
> E                False,
> E                'float32',
> E                None,
> E                array([-1., -1.], dtype=float32),
> E                False,
> E                'float32'),
> E               test_flags=num_positional_args=1. with_out=False. instance_method=False. native_arrays=[False]. container=[False]. as_variable=[False]. test_gradients=False. test_compile=False. ,
> E               fn_name='histogram',
> E               ground_truth_backend='tensorflow',
> E               backend_fw=<module 'ivy.functional.backends.torch' from '/ivy/ivy/functional/backends/torch/__init__.py'>,
> E               on_device='cpu',
> E           )

...I get these results in the python console:

```
Python 3.8.10 (default, Nov 14 2022, 12:59:47) 
[GCC 9.4.0] on linux
import ivy
ivy.set_backend('tensorflow')
2023-03-16 18:22:28.802444: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory
2023-03-16 18:22:28.802463: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
2023-03-16 18:22:30.087670: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcuda.so.1'; dlerror: libcuda.so.1: cannot open shared object file: No such file or directory
2023-03-16 18:22:30.087689: W tensorflow/stream_executor/cuda/cuda_driver.cc:269] failed call to cuInit: UNKNOWN ERROR (303)
2023-03-16 18:22:30.087711: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:163] no NVIDIA GPU device is present: /dev/nvidia0 does not exist
ivy.histogram(
    ivy.array([-1., -1.], dtype=ivy.float32),
    bins=ivy.array([0., 1.], dtype=ivy.float32),
    axis=0,
    extend_lower_interval=True,
    extend_upper_interval=False,
    dtype=ivy.float32,
    range=None,
    weights=ivy.array([-1., -1.], dtype=ivy.float32),
    density=False
)
2023-03-16 18:22:33.817245: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
(ivy.array([-2.]), ivy.array([0., 1.]))
ivy.set_backend('numpy')
ivy.histogram(
    ivy.array([-1., -1.], dtype=ivy.float32),
    bins=ivy.array([0., 1.], dtype=ivy.float32),
    axis=0,
    extend_lower_interval=True,
    extend_upper_interval=False,
    dtype=ivy.float32,
    range=None,
    weights=ivy.array([-1., -1.], dtype=ivy.float32),
    density=False
)
(ivy.array([-2.]), ivy.array([0., 1.]))
```

It looks like the problem is not with the histogram function, but I still don't know what is causing it in the tests.

MuhammadSaeedBatikh:
Hi @joaozenobio,
I have tested it locally, and indeed it works just fine when called directly with the exact same values!
I'm not sure what's causing this weird issue, so for now you can just ignore bins during testing (i.e. don't cross-test the returned bins). If no other issues arise, I think we're good here.

joaozenobio:
Hi @MuhammadSaeedBatikh ! I have made the tests without returning bins in the backends, but I got the following error:

```
E           ivy.utils.exceptions.IvyError: tensorflow: histogram: Number of dimensions of `x` and `weights` must coincide. Found: x has 2, weights has 1
E           Falsifying example: test_histogram(
E               values=(array([[-1., -1., -1.],
E                       [-1., -2.,  2.]]),
E                1,
E                (0, 1),
E                True,
E                True,
E                'float64',
E                (-1.0, 1.0),
E                array([[-1., -1., -1.],
E                       [-1., -1., -1.]]),
E                False,
E                'float64'),
E               fn_name='histogram',
E               test_flags=num_positional_args=1. with_out=False. instance_method=False. native_arrays=[False]. container=[False]. as_variable=[False]. test_gradients=False. test_compile=False. ,
E               ground_truth_backend='tensorflow',
E               backend_fw=<module 'ivy.functional.backends.torch' from '/ivy/ivy/functional/backends/torch/__init__.py'>,
E           )
E           
E           You can reproduce this example by temporarily adding @reproduce_failure('6.55.0', b'AXicY2BAB4xoNITNCCVQ5UlhMyEZyESBOShsRkYUdzIAAA3sACQ=') as a decorator on your test case
``` 
This error appears for TensorFlow when I use multiple axes. It looks like a bug in the Tensorflow function, because, from what I believe, the weights argument shouldn't be modified if I use multiple axes in the histogram function.

Tristan1649:
#12844

PR #11819, opened 2023-03-06
cond

pillarxyz:
#11117


pillarxyz:
#11047

pillarxyz:
Yeah I agree with that, but what should be done about it ? would we have the other framework change their behaviour to match the one of jax/numpy to have consistency ? I don't really know how to proceed with this

pillarxyz:
sure will look into it, and thanks alot for your patience with me

pillarxyz:
that error did pop up for me once, but I can't seem to reproduce it now weird

pillarxyz:
If my guess is right, I think switching back and forth from backend to backend causes some weird errors like this I noticed

PR #11825, opened 2023-03-06
Added jax.numpy.gradient function to JAX frontend

### LINKED ISSUE #11707: gradient

Supremolink81:
Close #11707 

jax.numpy.gradient documentation: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.gradient.html#jax.numpy.gradient

### DIFF
--- a/ivy/functional/frontends/jax/numpy/mathematical_functions.py
+++ b/ivy/functional/frontends/jax/numpy/mathematical_functions.py
@@ -93,6 +93,11 @@ def floor(x):
 
 
 @to_ivy_arrays_and_back
+def gradient(f, *varargs, axis=None, edge_order=None):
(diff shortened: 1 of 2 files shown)


Supremolink81:
So the only place where it is my function specifically that is failing appears to be in run_tests_1, where the problem is that spacing is an invalid keyword argument. I will try committing with varargs as the key, but if that doesn't work I am unsure what the issue is.

Supremolink81:
I reverted back to a previous commit because a change I made seemed to break things (specifically, when I added error handling in the gradient implementation in mathematica_functions.py).

Supremolink81:
So it appears almost all of the failures are not as a result of gradient, but of other functions. If there are any failures you see as a result of gradient, could you point me towards them so I can fix them? I assume I can just ignore the tests which fail as a result of something that isn't the function I implemented, right?

PR #11875, opened 2023-03-07
added double function to jax frontend 

Tejal-19:


Infrared1029:
Hi @Tejal-19 , Thanks a lot for taking the time to contribute to ivy:), I have noticed you haven't included the `@handle_frontend_test` decorator and generated the data in the function's body instead, we normally use `hypothesis` for that through the decorator, I'd suggest reading the testing section in our [docs](https://lets-unify.ai/ivy/deep_dive/ivy_frontends_tests.html), this will help a lot!

Tejal-19:
Hey, can you please have a look again. I have made the changes suggested.

Infrared1029:
Hi again @Tejal-19, we are going in the right direction but we need to fix several issues, ill leave a few comments.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_creation.py
```diff
@@ -551,6 +551,13 @@ def test_jax_numpy_full(
     )
 
 
+
+@handle_frontend_test(
+    fn_tree="jax.numpy.double",
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("double"),
```
Infrared1029, line 558: The function double casts arrays into a double (float64) data type, generating double values only would not test the function properly, we should instead be generating any dtype that is castable into a double (float64), that is any numeric type actually.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_creation.py
```diff
@@ -548,3 +549,35 @@ def test_jax_numpy_full(
...
+    dtype_and_x,
+    as_variable,
+    num_positional_args,
+    native_array,
+    on_device,
+    fn_tree,
+    frontend,
+):
+    # Check if the input datatype is double
+    if dtype_and_x[0] != "float64":
```
Infrared1029, line 572: This condition should be removed because we should never `return` from a test, to skip certain cases hypothesis provides the `assume` function, or you can use the `filter` method on strategies, if any of this does not make sense, I suggest reading the testing docs or hypothesis docs, regardless, we do not want to skip float64 cases as they are valid cases.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_creation.py
```diff
@@ -548,3 +549,35 @@ def test_jax_numpy_full(
...
+        return
+    # Generate test inputs
+    input_dtype, x = dtype_and_x
+    input_shape = x.shape
```
Infrared1029, line 576: This line does not seem to contribute to the test, we do not need the shape of `x`

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_creation.py
```diff
@@ -548,3 +549,35 @@ def test_jax_numpy_full(
...
+    y = jnp.random.randn(*input_shape).astype(jnp.float64)
```
Infrared1029, line 577: There is no `y` in the double function, so there is no need for that either, besides, we do not generate data inside testing functions, we use hypothesis strategies for that, in the decorator

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_creation.py
```diff
@@ -548,3 +549,35 @@ def test_jax_numpy_full(
...
+        input_dtypes=input_dtype,
+        as_variable_flags=as_variable,
+        with_out=False,
+        num_positional_args=num_positional_args,
+        native_array_flags=native_array,
+        frontend=frontend,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        x=x[0],
+        y=y,
```
Infrared1029, line 589: we should we move this `y=y` for the reasons mentioned above

Infrared1029:
I've also noticed that you pass num_positional_args, as_variable and native_array explicitly to the test function, those variables are now collected in the `test_flags` variable, I highly suggest reading this section of the [docs](https://lets-unify.ai/ivy/deep_dive/ivy_frontends_tests.html), to make it more clear, thanks again for your effort:)!

Tejal-19:
Hey, can you please recheck now.

hirwa-nshuti:
Hey @Tejal-19 

Thanks for contributing to ivy🤗

There was a large update to the repository which introduced many conflicts.
Would you copy your changes to a new PR😊 and close this one?

Feel free to tag me I'll give it a review🙂
Thanks

PR #11909, opened 2023-03-07
Complex DType checking for general functions

Fayad-Alman:
Flagging complex dtype checking for general functions in each framework 
- Note: Functions that deal with complex dtypes are NOT flagged


soma2000-lang:
@Fayad-Alman I think array_equal function of numpy supports the complex dtypes so the flagging should be removed.
image

PR #11951, opened 2023-03-08
Adding statistical functions to paddle backend

5a7man:
Pull request for experimental statistical functions (median and #10923).


MahmoudAshraf97 (approved):
I added scalar support for median, and added more dtypes for nanmean, note that out argument is ignored in paddle backend

PR #11967, opened 2023-03-08
Module level unsupported

RickSanchezStoic:
added version dictionary to all the valid/invalid types of each backend. Now, these attributes will be fetched according to the version. Also, now the decorators can be used as context managers. The decorators have an additional optional argument named "explicit", which is by default True. In the case it is False, a higher decorator of strictly the same type will be able to add information to it, otherwise, lower decorators will always dominate.


ivy/functional/backends/tensorflow/__init__.py
```diff
@@ -62,58 +63,97 @@
...
-    ivy.uint8,
-    ivy.uint16,
-    ivy.uint32,
-    ivy.uint64,
-)
-valid_float_dtypes = (ivy.bfloat16, ivy.float16, ivy.float32, ivy.float64)
-valid_uint_dtypes = (ivy.uint8, ivy.uint16, ivy.uint32, ivy.uint64)
-valid_complex_dtypes = (ivy.complex64, ivy.complex128)
+
+valid_dtypes_dict = {
```
VedPatwardhan, line 67: here as well

ivy/functional/backends/torch/__init__.py
```diff
@@ -47,47 +48,88 @@
...
-    ivy.bfloat16,
-    ivy.float16,
-    ivy.float32,
-    ivy.float64,
-)
-valid_int_dtypes = (ivy.int8, ivy.int16, ivy.int32, ivy.int64, ivy.uint8)
-valid_float_dtypes = (ivy.bfloat16, ivy.float16, ivy.float32, ivy.float64)
-valid_uint_dtypes = (ivy.uint8,)
-valid_complex_dtypes = (ivy.complex64, ivy.complex128)
+valid_dtypes_dict = {
```
VedPatwardhan, line 51: and here too

VedPatwardhan (changes requested):
Hey @RickSanchezStoic, just a few changes suggested, along with these, please also look into the lint failures if any are applicable to your PR. Thanks 🙂

PR #11968, opened 2023-03-08
Add support for partial backend

CatB1t:
- [x] Dynamically detect installed backends
- [x] Update `_determine_backend_from_args` to be backend-dependent
- [x] Update testing to support partial backend testing
- [ ] Add convenience script to create a new backend


ivy/func_wrapper.py
```diff
@@ -892,3 +892,12 @@ def new_fn(*args, **kwargs):
 with_supported_device_and_dtypes = _dtype_device_wrapper_creator(
     "supported_device_and_dtype", dict
 )
+
+
+def notImplemented(fn):
```
VedPatwardhan, line 897: Can we have this in snake case instead? Just for convention, thanks 🙂

VedPatwardhan (changes requested):
Just a small change, rest looks good!

PR #11985, opened 2023-03-08
Repair `torch.nn.functional.gelu` function

suvadityamuk:
Repaired and added functionality for `approximate` kwarg. All tests now pass locally. This function is required in order to transpile some models from `timm`


ivy/functional/frontends/torch/nn/functional/non_linear_activation_functions.py
```diff
@@ -98,14 +98,11 @@ def softmax(input, dim=None, _stacklevel=3, dtype=None):
...
-    # if approximate == "none":
-    # approximate = False
-    # else:
-    # approximate = True
-    return ivy.gelu(input, approximate=False)
+def gelu(input, approximate=None):
+    if approximate is None:
+        approximate = False
+
+    return ivy.gelu(input, approximate=approximate)
```
fspyridakos, line 105:
```suggestion
def gelu(input, approximate='none'):
    if approximate is 'none':
        return ivy.gelu(input, approximate=False)
    elif approximate is 'tanh':
        return ivy.gelu(input, approximate=True)
    else:
        raise ivy.utils.exceptions.IvyException(
             "approximate argument must be either none or tanh."
        )
```

fspyridakos (changes requested):
I suggested a change to the code, please also add `approximate` to the test since it not being tested currently. Should be `approximate = st.sampled_from(["none", "tanh"])`

ivy/functional/frontends/torch/nn/functional/non_linear_activation_functions.py
suvadityamuk: Have incorporated this, thank you!

fspyridakos:
LGTM

PR #11987, opened 2023-03-08
improved cardinality tests for empty shapes

MahmoudAshraf97:
the current cardinality test for copy_array throws an error when comparing empty shapes from different frameworks, e.g 
```
[] == ()
>>> False
```

### DIFF
--- a/ivy/functional/backends/tensorflow/experimental/elementwise.py
+++ b/ivy/functional/backends/tensorflow/experimental/elementwise.py
@@ -152,7 +152,9 @@ def count_nonzero(
         a, axis=axis, keepdims=keepdims, dtype=dtype, name=None
     )
 
-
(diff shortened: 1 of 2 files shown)


CatB1t (commented):
Which backend does it fail for? Ideally all backends should return empty tuples in that case.

MahmoudAshraf97:
Paddle backend returns empty list not tuple, it also fails for things like (1,)==[1,]

CatB1t:
in that case we should have the paddlepaddle backend return `tuple` and not `list`

MahmoudAshraf97:
I don't know if this is possible, ivy arrays with paddle backends return tuple shapes but the native arrays .shape method returns a list, so we either convert the list to tuple or the tuples to lists, and both are the same to me and the simplest solution IMO

PR #11994, opened 2023-03-08
gaussian_nll_loss

### LINKED ISSUE #12017: gaussian_nll_loss

iamharsha1999:
Close #12017


iamharsha1999:
@AnnaTz Hey, could you kindly review the PR now. I closed the previous one for the same issue and created this one by following the announcement instruction.

ivy/functional/frontends/torch/nn/functional/loss_functions.py
```diff
@@ -303,6 +303,52 @@ def nll_loss(
...
+
+    if ivy.any(var < 0):
+        raise ivy.utils.exceptions.IvyError(
+            "var has negative entry/entries"
+        )
+
+    var = torch_frontend.tensor(ivy.array(var, copy=True))
+    ivy.set_with_grads(False)
+    var = torch_frontend.clamp(var, min=eps)
+    ivy.unset_with_grads()
```
AnnaTz, line 339: Continuing where we left off in the previous PR, I don't think it will be necessary to do exactly what the source code does here with the gradients. You could simply do `var = torch_frontend.clamp(var, min=eps)`, but I would suggest `var = ivy.maximum(var, eps)` because it's more efficient to use an ivy function instead of a frontend one.

AnnaTz (changes requested):

ivy/functional/frontends/torch/nn/functional/loss_functions.py
iamharsha1999: Fixed these.

ivy_tests/test_ivy/test_frontends/test_torch/test_loss_functions.py
```diff
@@ -563,6 +563,75 @@ def test_torch_nll_loss(
...
+        min_dim_size=1,
+        max_dim_size=5,
+    ),
+    eps=st.floats(
+        min_value=0.0,
+        max_value=1.0,
+        allow_nan=False,
+        allow_infinity=False,
+    ),
+    reduction=st.sampled_from(["mean", "sum"]),
```
AnnaTz, line 604: I didn't notice this before, but you're not testing for different `full` values. You need to generate and pass True and False values for the argument `full`.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_loss_functions.py
iamharsha1999: Added them.

AnnaTz (approved):

PR #12101, opened 2023-03-09
Adding jax.lax.reduce_window

AnnaTz:


VedPatwardhan (changes requested):
Hey @AnnaTz, just pushed a few changes,
1. The `jax.lax.reduce_window` function requires the `operand` and the `init_value` to have the same `dtype` which I've updated in the tests to that we generate a `jax` array which has a `dtype`. 
2. The `others` arguments such as `window_stride`, `window_dimension`, etc. should all be lists instead of `jax` arrays, which I've updated as well.
3. There were issues with the reduction function itself, as the native `jax` function requires the output of the function to have the same dtype as the input, but `jnp.sum` seems to return values as default dtypes, so I updated that too. 
4. About the `computation` function part, we should be updating the test such that it allows us to pass 2 different functions either like [this](https://github.com/unifyai/ivy/blob/af9fb1de9d24b2f8361906c179d770820ca7eb21/ivy_tests/test_ivy/helpers/function_testing.py#L1330) or by being able to dynamically replace `jax` by `ivy.frontends.jax` while computing the result. The second one would be a better way to test in my opinion. Looping in @CatB1t as it's related to the testing pipeline. This will also be necessary when we implement the frontends for other such functions which accept functions as arguments such as `jax.vmap`, `jax.grad`, `jax.jac`, etc.

There's still quite a few issues with broadcasting and the logic with windows and strides which fail the test, but given that this is similar to the `layers` functions dimension handling you did recently, you may be better positioned to fix those, although feel free tolLet me know if you face any issues doing that. Thanks 🙂

PR #12103, opened 2023-03-09
Fix `ivy.functional.ivy` import

xoiga123:
[Trello task](https://trello.com/c/i8fyJEyf/1384-fix-ivyfunctionalivyimport)
Using `__all__` prevents builtin libraries and root ivy when importing subpackages. This PR also found out that `sys` was not imported from `ivy/__init__.py`, but was imported accidentally when using `from .functional import *`.

If we don't want to prevent builtin libraries from polluting ivy namespace, we can tweak the condition to `if not (name.startswith("_") or name == "ivy")`

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -4,6 +4,7 @@ import warnings
 from ivy._version import __version__ as __version__
 import builtins
 import numpy as np
+import sys
 
 
 warnings.filterwarnings("ignore", module="^(?!.*ivy).*$")
--- a/ivy/functional/ivy/__init__.py
+++ b/ivy/functional/ivy/__init__.py
@@ -43,3 +43,11 @@ from . import utility
 from .utility import *
 from . import control_flow_ops
 from .control_flow_ops import *
+import types
+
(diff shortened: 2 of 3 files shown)


VedPatwardhan:
Seems like a very clean solution to the problem, I'll just add the `Exhaustive CI` label to be sure that it doesn't cause large-scale failures (given that it's the `__init__.py` files). Based on the test results, we can get the PR merged. Thanks @xoiga123 🙂

xoiga123:
@VedPatwardhan before the latest fix:
```python3
import ivy
ivy.activations  # ok
ivy.functional.ivy  # ok
ivy.functional.math  # not found, correct
ivy.functional.activations  # not found, WRONG
```
after:
```python3
import ivy
ivy.activations  # ok
ivy.functional.ivy  # ok
ivy.functional.math  # not found, correct
ivy.functional.activations  # ok

VedPatwardhan:
lgtm! Let's get the PR merged, thanks @xoiga123 and @Nightcrab 🙂

PR #12110, opened 2023-03-09
adding nanpercentile to numpy frontend

### LINKED ISSUE #11969: nanpercentile

ydebaz:
Close #11969


YushaArif99 (changes requested):
Hi @ydebaz 

The implementation looks good overall. There are, however, a couple of points worth addressing.

- It seems you've unwittingly made some changes to the `test_array_api`. Could you kindly revert these changes as there should only be 2 modified files (in your case) 
- Another small point, could you also kindly remove the references/links you've added (as comments), especially the Deep Dive one in the test file
- The CI tests are not running for some reason due to which I can't really evaluate whether the implementation is correct or not. I'll try running the tests locally to verify that everything is working as expected. Could you also kindly confirm that the tests are passing on your local setup 

As always, let me know if there are any questions! 🙂

ivy_tests/test_array_api
YushaArif99, line 1: This file is still showing up in the modified files. Can you kindly remove the changes so that only 2 files show the `average_and_variances.py` and the `test_average_and_variances.py` files.

YushaArif99 (changes requested):
Hi @ydebaz 

I think the PR looks good overall. There is just one small fix remaining which I have mentioned above. Can you kindly fix that and re-request a review 🙂

ivy_tests/test_array_api
ydebaz: done

YushaArif99 (approved):
Hi @ydebaz 

Thanks a lot for making the requested changes. The PR looks in a good state which is why I'll now be merging this PR. 
Thanks for the contribution! 🙂

PR #12124, opened 2023-03-10
Fix the Import behaviour of nvidia-ml-py ( pynvml )

sagewhocodes:
This PR fixes the Import behaviour of `pynvml` to only be imported in case of GPU enabled environments.

### DIFF
--- a/ivy/functional/ivy/device.py
+++ b/ivy/functional/ivy/device.py
@@ -6,11 +6,16 @@ import gc
 import abc
 import math
 import psutil
-import pynvml
+
 import types
 from typing import Type, Optional, Tuple
 
 # noinspection PyUnresolvedReferences
+try:
+    import pynvml
+except ImportError:
+    # nvidia-ml-py (pynvml) is not installed in CPU Dockerfile.
+    pass
 try:
     pynvml.nvmlInit()
 except pynvml.NVMLError:
@@ -81,10 +86,12 @@ class DefaultDevice:
         ivy.set_default_device(self._dev)
         return self
 
-    def __exit__(self,
-                 exc_type: Optional[Type[BaseException]],
-                 exc_val: Optional[Type[BaseException]],
-                 exc_tb: Optional[types.TracebackType]) -> Union[ivy.Device, str]:
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[Type[BaseException]],
+        exc_tb: Optional[types.TracebackType],
+    ) -> Union[ivy.Device, str]:
         """
         Exit the runtime context related to the specified device.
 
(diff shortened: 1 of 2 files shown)


ivy/functional/ivy/device.py
```diff
@@ -6,15 +6,26 @@
...
-    pynvml.nvmlInit()
-except pynvml.NVMLError:
+    import pynvml
+
+    pynvml_installed = True
+except ImportError:
+    # nvidia-ml-py (pynvml) is not installed in CPU Dockerfile.
+    pynvml_installed = False
     pass
+
```
Dsantra92, line 22: merge into one try statement? `pynvml_installed` can be skipped.

ivy/functional/ivy/device.py
```diff
@@ -6,14 +6,20 @@
...
-    pynvml.nvmlInit()
-except pynvml.NVMLError:
+    import pynvml
+
+    try:
+        pynvml.nvmlInit()
+    except pynvml.NVMLError:
+        pass
+except ImportError:
+    # nvidia-ml-py (pynvml) is not installed in CPU Dockerfile.
```
Dsantra92, line 22: Could you add a warning or throw an import error when GPU is present but pynvml is not installed? I feel it's better to show import warnings during initialization, than during actual work.

Dsantra92 (changes requested):
I have one thing to add here:
We need to check if intelligent tests are run on GPU. If not, we can safely remove pyvnml installation. I think @RashulChutani can help us here.

RashulChutani:
> I have one thing to add here:
> We need to check if intelligent tests are run on GPU. If not, we can safely remove pyvnml installation. I think @RashulChutani can help us here.

They are not currently run on GPU. We will be extending to GPUs soon but we will pass the GPU using the --device argument.

Dsantra92 (approved):
LGTM

VedPatwardhan (approved):
lgtm!

RashulChutani (approved):
Looks good to me!

CatB1t (commented):
Just looking at the git  blame, the import seems to be there for a little while, also `pynvml` is listed in the requirements but it fails these days, maybe worth looking why it started failing recently? some tests internally use `pynvml` so the try except block won't be really helpful in making sure these doesn't run.

sagewhocodes:
> Just looking at the git blame, the import seems to be there for a little while, also `pynvml` is listed in the requirements but it fails these days, maybe worth looking why it started failing recently?

We actually made changes to the CPU Dockerfile such that it excludes nvidia based packages while building the CPU only version, however `pynvml` was always imported regardless if there was a GPU found or not. 
I have run tests under `ivy_tests` folder and didn't find any related errors, do I need to run tests from some other directory too?

CatB1t (approved):
lgtm!

PR #12132, opened 2023-03-10
Paddle paddle

mosesdaudu001:
-[quantile] #10932

I am having some issues with the test that I have not yet resolved yet. specifically, I don't know how to do the test for the paddle backend. I tried other backends like tensorflow and it worked. I have reached out through the discord ivy test channel, but it's not still resolved. I think the issue may be from the code in the conftest.py file.

As regards the previous comments you made on my previous pull request, I have made some adjustments to how the shape of the output should be. I believe this should resolve the issue, however, paddle's API especially for the quantile function behaves in a unique way which may be different from tensorflow. All I did was to modify the function to accept the interpolation argument, but it will still behave in the unique way it's known to operate in.

I will really appreciate if you can show me how you do the tests. It would really help me to debug incase there are other issues. Sorry for dragging this so long.

### DIFF
--- a/ivy/functional/backends/paddle/experimental/statistical.py
+++ b/ivy/functional/backends/paddle/experimental/statistical.py
@@ -4,7 +4,7 @@ import paddle
 from ivy.utils.exceptions import IvyNotImplementedException
 
 # local
-from ivy.func_wrapper import with_unsupported_device_and_dtypes
+from ivy.func_wrapper import with_unsupported_device_and_dtypes, with_supported_dtypes
 from . import backend_version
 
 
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97 (approved):
Thanks @mosesdaudu001 for completing this function, it was a hard one and I appreciate your good work. I did some minor modifications so it can pass the test and we are ready to merge
to run the tests use
`pytest -B paddle ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py::test_quantile`

the backend tests are universal for all backends, no need to write a new test for paddle backend

PR #12173, opened 2023-03-10
Added an assertion to check if the x64 flag is enabled, also checks i…

fspyridakos:
…f any x64 dtype input is passed or resulted from promotion


VedPatwardhan (approved):
lgtm! Feel free to merge 🙂

PR #12179, opened 2023-03-10
functional: paddle.random: implement randint and random_uniform

hackslashX:
Hi,
Implemented `randint` and `random_uniform` methods. 
I added `unsupported types` decorator since some `dtypes` had no supported comparison kernels in the package. I'm also reshaping the return value to better correspond with the input types (a scalar input without specifying shape, results in a scalar output).
Please do let me know of any suggestions and omissions. Thanks.


hackslashX:
This closes #12139

MahmoudAshraf97 (approved):
Awesome work Muhammad, keep up the good work

PR #12374, opened 2023-03-13
Added Paddle shuffle Function

Shivam7-1:
Added Paddle shuffle Function
Issue: https://github.com/unifyai/ivy/issues/12356

### DIFF
--- a/ivy/functional/backends/paddle/random.py
+++ b/ivy/functional/backends/paddle/random.py
@@ -135,9 +135,7 @@ def shuffle(
 ) -> paddle.Tensor:
     if seed:
         _ = paddle.seed(seed)
-    # Use numpy's permutation function to shuffle indices
-    indices = paddle.to_tensor(np.random.permutation(x.shape[0]), dtype='int64')
+    # Use Paddle's randperm function to generate shuffled indices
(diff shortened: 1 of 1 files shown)


ivy/functional/backends/paddle/random.py
```diff
@@ -133,4 +133,11 @@ def shuffle(
     seed: Optional[int] = None,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    raise IvyNotImplementedException()
+    if seed:
+        _ = paddle.seed(seed)
+    # Use numpy's permutation function to shuffle indices
+    indices = paddle.to_tensor(np.random.permutation(x.shape[0]), dtype='int64')
```
MahmoudAshraf97, line 139: Hi, the implementation looks good to me, please use only paddle functions in paddle backend, there is a function called randperm that can be used instead
also it doesn't support native out so remove the last line

MahmoudAshraf97 (changes requested):

ivy/functional/backends/paddle/random.py
Shivam7-1: yes done changes according to given feedback remove last line and uses randperm function to it

MahmoudAshraf97 (approved):

PR #12383, opened 2023-03-13
Add isin to jax.numpy logic functions to Jax frontend

AntoniosMalak:


AntoniosMalak:
Hi, @ZoeCD, may u check the code because check_formatting error is not from my code!

PR #12428, opened 2023-03-14
add asanyarray in numpy.frontend maninupulation

codenificient:
First attempt at #12422. Followed the function signature for `numpy.asanyarray`. The underlying ivy array method I could find is `ivy.any`


codenificient:
#12422

RickSanchezStoic:
Hi @codenificient can you please add the test for this function too?

PR #12455, opened 2023-03-14
Generating frontend arrays in frontend tests

Ishticode:
The new boolean flag in `test_frontend_function` named `generate_frontend_arrays` will determine whether the test is run on ivy arrays or specific frontend arrays. Not sure if we always want to enforce frontend arrays. Currently the out argument is also not forced to be frotnend array but depends on the flag.

The following updates have been made.

- **`ivy/func_wrapper.py`** : updated the _get_first_array to take a custom array_fn rather than just hard-coded `ivy.array` to be able to also pass `_is_frontend_array`
- **`testing_parameter_flags.py`**: Created the boolean strategy `BuiltFrontendArrayStrategy` and added it to the `FrontendFunctionTestFlags`
- **`testing_helpers.py`**: pass the boolean strategy to `handle_frontend_test`
- **`function_testing.py`**: Added the function`args_to_frontend` (and `_to_frontend`), this is then used to convert args to frontend array. kw/args_ivy are renamed to kw/args_for_test. The frontend version of flatten functions were added, namely, `flatten_frontend` and `flatten_frontend_to_np` . Rest of the changes will be pretty obvious with the use of the new `generate_frontend_arrays` flag.
- The init files of the frontends have a new unified private variable to access array/tensor creation function name `_frontend_array`


ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -647,64 +650,107 @@ def test_frontend_function(
         importlib.import_module("ivy.functional.frontends.jax").config.update(
             "jax_enable_x64", True
         )
-    ret = get_frontend_ret(frontend_fn, *args_ivy, **kwargs_ivy)
+
```
CatB1t, line 653: We can move the frontend generation logic to the method `apply_flags` on the decorator, this removes the overhead of branching, and also makes everything cleaner and separated.

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -1947,3 +2009,32 @@ def _frontend_array_to_ivy(x):
...
+def _to_frontend(frontend):
+    frontend_fn_dict = {"jax":
+                            ivy.functional.frontends.jax.numpy.array,
+                        "numpy":
+                            ivy.functional.frontends.numpy.array,
+                        "tensorflow":
+                            ivy.functional.frontends.tensorflow.constant,
+                        "torch":
+                            ivy.functional.frontends.torch.tensor}
```
CatB1t, line 2032: Let's make this more dynamic, we can create a private unified variable that points to the frontend array function, in that way we don't have to update the dict when a new frontend is added, thoughts? @fspyridakos

ivy_tests/test_ivy/helpers/testing_helpers.py
```diff
@@ -376,6 +377,7 @@ def handle_frontend_test(
     test_inplace=BuiltInplaceStrategy,
     as_variable_flags=BuiltAsVariableStrategy,
     native_array_flags=BuiltNativeArrayStrategy,
+    gen_frontend_arrays=BuiltFrontendArrayStrategy,
```
CatB1t, line 380: It's better if we use the full `generate`

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -1835,26 +1883,37 @@ def flatten_fw_and_to_np(*, ret, fw):
     return ret_np_flat
 
 
-def flatten(*, ret):
+def flatten(*, ret, frontend=None):
```
CatB1t, line 1886: I'd suggest creating a separate function for the frontend flattening, the test functions gets called a lot during the test execution, less branching is better not to add an overhead, also in this way it's clear that the frontends and backends have separate test logic

ivy/func_wrapper.py
```diff
@@ -60,19 +60,21 @@ def try_array_function_override(func, overloaded_args, types, args, kwargs):
     )
 
 
-def _get_first_array(*args, **kwargs):
+def _get_first_array(*args, array_fn=None, **kwargs):
```
CatB1t, line 63: I'm not sure if we should modify the `func_wrapper` here to achieve this, as is there any use case of that other than for the frontend testing? If not let's have that logic in the testing it self, @VedPatwardhan do you think this adds an overhead?

CatB1t (changes requested):
Hello Ishtiaq, great work! I've added a few comments, and also requested a review from @VedPatwardhan and @fspyridakos.

ivy_tests/test_ivy/helpers/function_testing.py
Ishticode: I thought about it and I feel as though we have to keep the logic here for two reasons.

Firstly, *to_ivy is called on the args in the function body after `create_args_kwargs` which would spoil initial creation of frontend arrays in apply flags.

Secondly, although a unified variable to frontend array creation is created, it would still require us to know for which frontend do we want the create the arrays for.. and apply flags or the Flags class doesn't hv any parameters for passing in the current frontend (for which the test is being run). We also cannot work with just the np frontend arrays the way we can in the backends... as the transition between frontend arrays is not smooth (at least now) as opposed to the backends.

I hope that makes sense. Feel free to counter...

ivy_tests/test_ivy/helpers/testing_helpers.py
Ishticode: sure

ivy_tests/test_ivy/helpers/function_testing.py
Ishticode: yes sure. Wrote flatten_frontend instead.

ivy/functional/frontends/torch/__init__.py
```diff
@@ -285,3 +285,5 @@ def promote_types_of_torch_inputs(
 from . import utilities
 from .utilities import *
 from . import linalg
+
+frontend_array = tensor
```
CatB1t, line 289: missing new line

ivy/functional/frontends/jax/__init__.py
```diff
@@ -7,3 +7,5 @@
 from . import numpy
 from . import random
 from . import _src
+
+frontend_array = numpy.array
```
CatB1t, line 11: let's make `frontend_array` variable private `_frontend_array`, so users aren't allowed to use that.

CatB1t (changes requested):
Just few simple comments, then feel free to merge!

Ishticode:
Thanks @CatB1t. Will be merging this soon. Thanks.

PR #12462, opened 2023-03-15
Bug fix no Numpy frontend prod on TensorFlow

sergio-verduzco:
Using the TensorFlow backend, `initial` not `None`, and `axis=-1` leads to an exception Example:
```
import ivy.functional.frontends.numpy as ivy_np
ivy.set_backend('tensorflow')
a =  ivy_np.array([[0, 0], [0, 0]], dtype='uint16')
prod = ivy_np.prod(a, axis=-1, dtype='uint16', keepdims=False)
```
This is because the indexing of TensorFlow tensors causes the `s[-1] = 1` instruction to collapse the shape. Additionally I made changes to avoid two other problems. First, setting `initial=ivy.nan` causes the product to be `nan` whenever `initial` is not passed to `prod`. Second, the code will fail when `initial` is specified, but `axis` is not.

### DIFF
--- a/ivy/functional/frontends/numpy/mathematical_functions/sums_products_differences.py
+++ b/ivy/functional/frontends/numpy/mathematical_functions/sums_products_differences.py
@@ -51,16 +51,19 @@ def prod(
     dtype=None,
     out=None,
     keepdims=False,
-    initial=ivy.nan,
+    initial=None,
     where=True,
 ):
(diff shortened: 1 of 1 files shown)


zhumakhan:
Hi @sergio-verduzco, thanks for your commit! It fixes the issues with `axis=-1` and `ivy,nan is not None` is True.

PR #12488, opened 2023-03-15
Added modf Numpy Frontend Function

janetwamaitha:
#12485


Infrared1029:
Glad to see your `modf` PR is finally here @janetwamaitha! this PR does more than just adding the `modf` to numpy,'s frontend as it introduces `modf` in our experimental api, we need to write a test for that in `ivy/ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py`, once the test is written and passes, we can move on with the review:)

PR #12490, opened 2023-03-15
ivy.experimental: add support for logsigmoid

### LINKED ISSUE #12472: logsigmoid

hmahmood24:
Close #12472


ivy/functional/backends/numpy/activations.py
```diff
@@ -115,3 +115,7 @@ def mish(x: np.ndarray, /, *, out: Optional[np.ndarray] = None) -> np.ndarray:
 
 
 mish.support_native_out = True
+
+
+def logsigmoid(input: np.ndarray) -> np.ndarray:
+    return -(np.log1p(np.exp(-(input))))
```
AnnaTz, line 121: I think you're doing this wrong mathematically. 
It should be log of the inverse of 1+exp, not log of 1+exp. Have a look at how pytorch defines this:
image.
This might be the cause of the failure.

AnnaTz (changes requested):

ivy/functional/backends/numpy/activations.py
hmahmood24: Hey. Actually these two are mathematically equivalent due to the properties of logarithms. For example, take a look how Jax defines it:

image

AnnaTz: *facepalm* yeah I forgot about that

AnnaTz (changes requested):

PR #12539, opened 2023-03-16
multiversion docker

RickSanchezStoic:
drastically reduces the size of the multiversion docker image, allowing for dynamic docker image creation with choice of framework versions installed, crisp


krishna1111111:
Good job

PR #12545, opened 2023-03-16
update Jax Frontend/submodules/indexing dashboard

Sohob:
#12516


Sohob:
Close #12516

Sohob:
Hey @nassimberrada, I'll go for what you suggested and find another front-end function. 

Thanks for reviewing my PR and for the clarification, I'll take your notes into consideration when picking my next function.

Issue #12541, opened 2023-03-16
angle


mattcrewe:

Harrison-O:
Isn't this already implement in ```ivy.functional.frontends.numpy.mathematical_functions.handling_complex_numbers.py```?

PR #12552, opened 2023-03-16
Add angle function to NumPy Frontend

mattcrewe:


mattcrewe:
Close #12541

MuhammadSaeedBatikh:
Hi @mattcrewe, thanks for the PR. 
`np.angle` computes the angle of a complex input, so I am not entirely sure of reason behind the TODO you added. Could you please elaborate on this?  
Otherwise, lgtm.

mattcrewe:
Hi @MuhammadSaeedBatikh.
While the angle function is intended to take complex numbers as input the `np.angle` function will accept integer and float numbers as input; the other backends also mirror this behaviour. I am currently not testing for cases outside complex numbers as the current frontend implementation is only suitable to handle complex inputs. I left it at this point as it matches the current solutions for the other backends.

MuhammadSaeedBatikh:
Makes sense. 
But no need to worry about that as these dtype conversions (if supported) are handled by the backend functions anyway. I', gonna remove the comments/TODOs and merge the PR. Thanks.

aryanxk02:
Hello @mattcrewe and @MuhammadSaeedBatikh, I happened to spend some time working on this function because it was not marked 'completed' here: https://github.com/unifyai/ivy/issues/1525. Later while going through the codebase, I realised that the code for angle() has already been pushed. Please do the needful. Thank you.

PR #12561, opened 2023-03-16
Tf frontend func expm1 added and in test

tanyalondon:
Hi
Please can you check my changes to the tf frontend math.py and test file.  I was unable to configure the tests in codespaces


tanyalondon:
Hi

My original open issue number was  12362.  Please can you close that?

Issue #12525, opened 2023-03-16
min


irjawais:

danielmunioz:
Hi @irjawais! I'm sorry, but the MXNet frontend tasks are not ready to be implemented by contributors yet, but if you're interested on the frontend side of ivy, you can navigate to the later pages of the [Frontend open tasks](https://github.com/unifyai/ivy/issues?zpage=1&q=is%3Aopen+is%3Aissue+label%3AToDo+label%3A%22JAX+Frontend%22%2C%22TensorFlow+Frontend%22%2C%22PyTorch+Frontend%22%2C%22NumPy+Frontend%22) for a list of issues just like this one on the other frontends that can be work on right away!
Thanks!

PR #12576, opened 2023-03-17
Add mxnet.numpy Statistics min #12525

irjawais:


Infrared1029:
Hi @irjawais, the mxnet backend does not exist (yet) so we should not be adding anything related to that atm, on the other hand, the mxnet frontend does exist and you can contribute to that if you would like, you can create a PR that only updates the mxnet frontend.

PR #12627, opened 2023-03-17
divmod numpy frontend

pillarxyz:
#12565


pillarxyz:
image

If you found that the tests fail for torch backend it's because of this bug in torch

pillarxyz:
> image
> 
> If you found that the tests fail for torch backend it's because of this bug in torch

and even in the backend implementation in ivy it's still giving wrong results

PR #12778, opened 2023-03-20
Adding cumulative_trapezoid function

### LINKED ISSUE #12708: cumulative_trapezoid

jatinS-dev:
closes #12708

### DIFF
--- a/ivy/functional/frontends/torch/blas_and_lapack_ops.py
+++ b/ivy/functional/frontends/torch/blas_and_lapack_ops.py
@@ -212,8 +212,8 @@ def trapezoid(y, x=None, *, dx=None, dim=-1):
     return ivy.trapz(y, x=x, dx=dx, axis=dim)
 
 
-@to_ivy_arrays_and_back
+@with_unsupported_dtypes({"1.13.1 and below": ("float16", "bfloat16")}, "torch")
 def cumulative_trapezoid(y, x=None, *, dx=None, dim=-1):
     if x is not None:
         y, x = torch_frontend.promote_types_of_torch_inputs(y, x)
-    return ivy.cumulative_trapezoid(y, x=x, dx=dx, axis=dim)
+    return ivy.cumsum(ivy.trapz(y, x=x, dx=dx, axis=dim), axis=dim)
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/torch/blas_and_lapack_ops.py
```diff
@@ -210,3 +210,10 @@ def trapezoid(y, x=None, *, dx=None, dim=-1):
     if x is not None:
         y, x = torch_frontend.promote_types_of_torch_inputs(y, x)
     return ivy.trapz(y, x=x, dx=dx, axis=dim)
+
+
+@to_ivy_arrays_and_back
+def cumulative_trapezoid(y, x=None, *, dx=None, dim=-1):
+    if x is not None:
+        y, x = torch_frontend.promote_types_of_torch_inputs(y, x)
+    return ivy.cumulative_trapezoid(y, x=x, dx=dx, axis=dim)
```
Daniel4078, line 219: there is no cumulative_trapezoid function in ivy for you to use directly like this, please consider compositional approach or something else.

Daniel4078 (changes requested):
Hello, I found that your implementation for the frontend does not work. Please check the comment below, thanks!

ivy/functional/frontends/torch/blas_and_lapack_ops.py
jatinS-dev: Hello @Daniel4078 so should I make another PR for cumulative_trapezoid or continue the function development here only, by using the composition function?

Daniel4078: you should just continue here first as there might be some functions in the Ivy api that provide similar feature like cumulative_trapezoid but named differently.

Daniel4078 (changes requested):
Well, your implementation somehow does not really work. For example, according to the test log here:(https://github.com/unifyai/ivy/actions/runs/4499836402/jobs/7918203393?pr=12778). your function produce the wrong result "[2. 4.]" when the expected result is "[[2. 2.]] " for inputs: dtype_y_x=(['float32'], array([[2., 2.],
E               [2., 2.]], dtype=float32), array([[2., 2.],
E               [2., 2.]], dtype=float32)),
E       use_x=False,
E       dim=0,
E       dx=1.0,

jatinS-dev:
@Daniel4078 how can I check this error on my local machine?

Daniel4078:
> @Daniel4078 how can I check this error on my local machine?

I think you should check the deep dive document about how to run certain tests on your local machines.

PR #12779, opened 2023-03-20
Refactoring `batch_norm`, `instance_norm` and `layer_norm`

hello-fri-end:
[Trello Task:](https://trello.com/c/Btjsmlw8/1474-refactoring-batchnorm-instancenorm-and-layernorm)

- [x] Moved batch_norm to experimental/norms.
- [x] Added `momentum` argument to batch_norm and added logic to calculate `running mean` and `running variance` using `momentum` when `training` is `True`.
- [x] Updated batch_norm to return normalized output, running mean, and running variance as discussed [here](https://discord.com/channels/799879767196958751/1000044163070447626/1087650941714575440)
- [x] Added out argument to batch_norm. (Not sure if any extra handling needs to be done since the function returns three arrays)
- [x] Implemented instance_norm as a wrapper of batch_norm for TensorFlow, NumPy, and Jax. Tensorflow does not have an instance_norm function and the wrapping implementation allows us to utilize the efficient `tf.nn.batch_normalization`. It also allows us to reuse the code we've already written for batch_norm and avoid replicating the common logic. Proof that instance_norm with (N,C,H,W) is equivalent to batch_norm on (1, N*C, H,W) -> [Collab Notebook](https://colab.research.google.com/drive/1rB-k_zdWXnHdDGUdzAKadsRYTATRu5xe?usp=sharing) 
- [x] Update docstring of `batch_norm` and `instance_norm`
- [x] Moved all frontend torch normalization functions from `nn/functional/non_linear_activations` to `nn/functional/norms` and moved the corresponding tests from `test_non_linear_activation_functions` to `test_norms`.
- [x] Updated frontend implementation of `torch.nn.functional.batch_norm` and `tf.nn.batch_normalization` to return `normalized result`, `running_mean` and `running_variance`. Also added frontend implementation and test for `torch.nn.functional.instance_norm`.
- [x] Convert layer norm to a mixed function


jkeane508:
Hey @AnnaTz,

Could you take a look at this? You probably have a better idea than me!

Thanks :)

hello-fri-end:
Hello @AnnaTz! I have finished making the changes as outlined in the to-do list above. There are some issues in the layernorm tests that I was hoping to get your help on:

1. `ivy.functional.norms.layer_norm`
The test for the torch backend fails with the following error message:
`ivy.utils.exceptions.IvyBackendException: torch: layer_norm: "LayerNormKernelImpl" not implemented for 'Half'`

This does not make sense to me since I have added the `unsupported_dtypes` decorator in the torch backend with `bfloat16` and `float16` dtypes --> [here](https://github.com/unifyai/ivy/blob/c74bb4ccfe6a4666bd6224169956e73355244522/ivy/functional/backends/torch/norms.py#L8). Did I miss anything in this?

2. `torch.nn.functional.layer_norm` frontend test fails for NumPy because of Assertion Error. There is some inconsistency in the torch native function for which I have created an [issue](https://discuss.pytorch.org/t/unexpected-results-with-torch-nn-functional-layer-norm/176064) on torch forums. 

Rest all the tests are passing.

ivy/data_classes/array/experimental/norms.py
```diff
@@ -39,93 +39,123 @@ def l2_normalize(
...
+            An offset array. If present, will be added to the normalized input.
+        scale
+            A scale array. If present, the scale is applied to the normalized input.
+        training
+            If true, calculate and use the mean and variance of `x`. Otherwise, use the
+            provided `mean` and `variance`.
+        eps
+            A small float number to avoid dividing by 0.
+        momentum
+             The value used for the running_mean and running_var computation.
```
AnnaTz, line 79: I don't think you've mentioned running_mean and running_var anywhere previously. Would be nice to rephrase the description a bit to explain what these are. The same goes for all the added functions.

ivy/data_classes/array/experimental/norms.py
```diff
@@ -39,93 +39,123 @@ def l2_normalize(
         """
         return ivy.l2_normalize(self, axis=axis, out=out)
 
+    def batch_norm(
+        self,
```
AnnaTz, line 43: Type-hints are missing for some `self` arguments.

ivy/data_classes/array/experimental/norms.py
```diff
@@ -39,93 +39,123 @@ def l2_normalize(
...
+        variance: Union[ivy.NativeArray, ivy.Array],
+        /,
+        *,
+        offset: Optional[Union[ivy.NativeArray, ivy.Array]] = None,
+        scale: Optional[Union[ivy.NativeArray, ivy.Array]] = None,
+        training: bool = False,
+        eps: float = 1e-5,
+        momentum: float = 1e-1,
+        out: Optional[ivy.Array] = None,
+    ) -> ivy.Array:
```
AnnaTz, line 54: And you haven't updated the type-hints for the return of the functions.

ivy/functional/backends/torch/norms.py
```diff
@@ -0,0 +1,24 @@
+import torch
+from typing import Optional, List
+
+from ivy.func_wrapper import with_unsupported_dtypes
+from . import backend_version
+
+
+@with_unsupported_dtypes({"0.11.0 and below": ("bfloat16", "float16")}, backend_version)
```
AnnaTz, line 8: And I guess the version should be "1.11.0" here as well.

AnnaTz (changes requested):

ivy/functional/backends/torch/experimental/norms.py
```diff
@@ -20,65 +20,70 @@ def l2_normalize(
 l2_normalize.support_native_out = True
 
 
-@with_unsupported_dtypes({"1.11.0 and below": ("float16", "bfloat16")}, backend_version)
-def instance_norm(
+@with_unsupported_dtypes({"0.11.0 and below": ("bfloat16", "float16")}, backend_version)
```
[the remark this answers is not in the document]
hello-fri-end: Ah, silly. I don't know how I didn't notice that :P Thanks :)

ivy/data_classes/array/experimental/norms.py
hello-fri-end: Updated :+1:

ivy/data_classes/array/experimental/norms.py
hello-fri-end: Updated :+1:

ivy/data_classes/array/experimental/norms.py
hello-fri-end: Updated :+1:

ivy/functional/backends/torch/norms.py
hello-fri-end: Updated :+1:

hello-fri-end:
@AnnaTz ! Here's a summary of the changes I made since the last review:

1. Updated `batch_norm` and `instance_norm` to be mixed functions with mixed partial implementation for the torch backend.
2. Updated the expected data format from N,C,*S to N,S,*C and updated all the functions to match this new scheme.
3. Added a new test for the frontend tensorflow batch_norm function which generates a set of mutually broadcastable inputs instead of hardcoding a single shape.
4. Updated the frontend test for `torch.nn.functional.layer_norm` with new bounds on the values because of overflow and underflow problems as highlighted in this [issue](https://github.com/pytorch/pytorch/issues/98075).

The test for `ivy.functional.norms.layer_norm` still fails for the torch backend with the `dtype` error even though I [updated](https://github.com/unifyai/ivy/blob/ba4a10202bab8da6a4eccdbc6ff9384b1599b2f6/ivy/functional/backends/torch/norms.py#L8) the version in the decorator to 1.11.0.

ivy/data_classes/array/experimental/norms.py
AnnaTz: Since the function returns the normalised input, the mean and the variance, shouldn't the return type-hint be a tuple of ivy arrays instead of a single ivy array?

ivy/data_classes/array/experimental/norms.py
```diff
@@ -39,93 +39,149 @@ def l2_normalize(
...
+            If ``training=True`` then it must be one dimensional with size equal
+            to the size of channel dimension C. If ``training=False`` then it can
+            be of any shape broadcastble to the input shape.
+        scale
+            A scale array. If present, the scale is applied to the normalized input.
+            If ``training=True`` then it must be one dimensional with size equal to
+            the size of channel dimension C. If ``training=False`` then it can be of
+            any shape broadcastble to the input shape.
+        training
+            If true, calculate and use the mean and variance of `x`. Otherwise, use the
```
AnnaTz, line 86: If I understand correctly, for training=True the **_running_** mean and variance are used. Would be nice to update this in all the docstring to avoid confusion.

AnnaTz (changes requested):

AnnaTz:
> @AnnaTz ! Here's a summary of the changes I made since the last review:
> 
> 1. Updated `batch_norm` and `instance_norm` to be mixed functions with mixed partial implementation for the torch backend.
> 2. Updated the expected data format from N,C,*S to N,S,*C and updated all the functions to match this new scheme.
> 3. Added a new test for the frontend tensorflow batch_norm function which generates a set of mutually broadcastable inputs instead of hardcoding a single shape.
> 4. Updated the frontend test for `torch.nn.functional.layer_norm` with new bounds on the values because of overflow and underflow problems as highlighted in this [issue](https://github.com/pytorch/pytorch/issues/98075).
> 
> The test for `ivy.functional.norms.layer_norm` still fails for the torch backend with the `dtype` error even though I [updated](https://github.com/unifyai/ivy/blob/ba4a10202bab8da6a4eccdbc6ff9384b1599b2f6/ivy/functional/backends/torch/norms.py#L8) the version in the decorator to 1.11.0.

Thanks for the update! Is this the only failure you get when running locally? It's weird, I don't see any error related to your changes in the CI test.

ivy/data_classes/array/experimental/norms.py
hello-fri-end: When `training` is set to `True`,  `batch_mean` and `batch_variance` is used to normalize the input. In this case, running mean and running variance are updated using these batch statistics and the mean and variance passed as input as highlighted in the docstring here: https://github.com/unifyai/ivy/blob/ba4a10202bab8da6a4eccdbc6ff9384b1599b2f6/ivy/functional/ivy/experimental/norms.py#L72-L82
Should I copy this description to container and arrays methods too? I thought that we only write a brief description in array and container methods and redirect the user to the docstring of the functional API with this line "This method simply
        wraps the function, and so the docstring for ivy.batch_norm also applies
        to this method with minimal changes."

ivy/data_classes/array/experimental/norms.py
hello-fri-end: Had missed this, have updated it now.

ivy/data_classes/array/experimental/norms.py
AnnaTz: Oh you're right. I didn't notice the explanation at the start of the docstring at all. I went straight to the argument definition. No, I don't think you need to update the array and container methods.

hello-fri-end:
> > @AnnaTz ! Here's a summary of the changes I made since the last review:
> > 
> > 1. Updated `batch_norm` and `instance_norm` to be mixed functions with mixed partial implementation for the torch backend.
> > 2. Updated the expected data format from N,C,*S to N,S,*C and updated all the functions to match this new scheme.
> > 3. Added a new test for the frontend tensorflow batch_norm function which generates a set of mutually broadcastable inputs instead of hardcoding a single shape.
> > 4. Updated the frontend test for `torch.nn.functional.layer_norm` with new bounds on the values because of overflow and underflow problems as highlighted in this [issue](https://github.com/pytorch/pytorch/issues/98075).
> > 
> > The test for `ivy.functional.norms.layer_norm` still fails for the torch backend with the `dtype` error even though I [updated](https://github.com/unifyai/ivy/blob/ba4a10202bab8da6a4eccdbc6ff9384b1599b2f6/ivy/functional/backends/torch/norms.py#L8) the version in the decorator to 1.11.0.
> 
> Thanks for the update! Is this the only failure you get when running locally? It's weird, I don't see any error related to your changes in the CI test.

Yup, that;s the only test failing when running locally. My torch version is `1.11.0`, I have no idea why it's not passing.

hello-fri-end:
> > > @AnnaTz ! Here's a summary of the changes I made since the last review:
> > > 
> > > 1. Updated `batch_norm` and `instance_norm` to be mixed functions with mixed partial implementation for the torch backend.
> > > 2. Updated the expected data format from N,C,*S to N,S,*C and updated all the functions to match this new scheme.
> > > 3. Added a new test for the frontend tensorflow batch_norm function which generates a set of mutually broadcastable inputs instead of hardcoding a single shape.
> > > 4. Updated the frontend test for `torch.nn.functional.layer_norm` with new bounds on the values because of overflow and underflow problems as highlighted in this [issue](https://github.com/pytorch/pytorch/issues/98075).
> > > 
> > > The test for `ivy.functional.norms.layer_norm` still fails for the torch backend with the `dtype` error even though I [updated](https://github.com/unifyai/ivy/blob/ba4a10202bab8da6a4eccdbc6ff9384b1599b2f6/ivy/functional/backends/torch/norms.py#L8) the version in the decorator to 1.11.0.
> > 
> > 
> > Thanks for the update! Is this the only failure you get when running locally? It's weird, I don't see any error related to your changes in the CI test.
> 
> Yup, that;s the only test failing when running locally. My torch version is `1.11.0`, I have no idea why it's not passing.

The tests for group_norm and stateful were failing on the CI because I hadn't updated the parameters in the function call for these. I've updated these parameters now.

hello-fri-end:
@AnnaTz Can you please check if you're also getting the dtype error when running the `ivy.functional.layer_norm` test locally? Maybe a bug has been introduced in the testing helpers recently due to which it's unable to get the correct dtypes for the function. And about the failing test for the torch frontend group norm function on the CI, the function has been implemented as a wrapper of `ivy.layer_norm` in the frontends, is it ok if we merge this PR and I will open a new PR for adding group norm to the experimental API of IVY?

AnnaTz:
> @AnnaTz Can you please check if you're also getting the dtype error when running the `ivy.functional.layer_norm` test locally? Maybe a bug has been introduced in the testing helpers recently due to which it's unable to get the correct dtypes for the function. And about the failing test for the torch frontend group norm function on the CI, the function has been implemented as a wrapper of `ivy.layer_norm` in the frontends, is it ok if we merge this PR and I will open a new PR for adding group norm to the experimental API of IVY?

I ran test_layer_norm and got this error for the torch backend `ivy.utils.exceptions.IvyBackendException: torch: layer_norm: "LayerNormKernelImpl" not implemented for 'Half'`. 
About `group_norm`, I'm already reviewing a PR by @karalleyna to add it to our api.

hello-fri-end:
Merging this PR. The datatype issue with the layer norm test will be fixed once the following Trello task is solved:
https://trello.com/c/8IQK5r9V/1649-investigate-why-getsupporteddevicesdtypes-function-is-not-returning-the-correct-dtype-set-for-mixed-functions

PR #12792, opened 2023-03-20
Making all container static methods private

soma2000-lang:
Solves https://trello.com/c/SwTr0PxY

### DIFF
--- a/ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
+++ b/ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
@@ -162,6 +162,11 @@ def movedim(input, source, destination):
 
 
 @to_ivy_arrays_and_back
+def moveaxis(input, source, destination):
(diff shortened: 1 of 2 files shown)


Aarsh2001:
Looks good, requesting a review from @VedPatwardhan.

VedPatwardhan (changes requested):
Thanks for making the changes @soma2000-lang and thanks for requesting the review @Aarsh2001. Could you please also update the rest of the container static methods (in the other submodules as well just like `elementwise`)? Once done with that, could you please also update the `handle_nestable` decorator as [this](https://github.com/unifyai/ivy/blob/45fa4412108bc46b3867f905de333cd7ae1d254b/ivy/func_wrapper.py#L642) line would break. Thanks 🙂

soma2000-lang:
@VedPatwardhan Yes I am doing that ,Actually this work is not complete yet.I will mark it ready for review once I am done.

PR #12793, opened 2023-03-20
Solved_Descriptions_Problems

krishna1111111:


VedPatwardhan (commented):
Hey @krishna1111111, I don't think we should add descriptions this manner, as docstrings and other comments dividing into should be sufficient to explain the code. In case you're interested in docstrings and stuff, feel free to check out the open tasks for [Function Reformatting](https://lets-unify.ai/ivy/contributing/open_tasks.html#function-formatting). Thanks 🙂

krishna1111111:
should i now change to its ok for now ?

VedPatwardhan:
Yeah it may help in the short term, but these comments would be difficult to maintain, which is why we've instead added a [`Deep Dive`](https://lets-unify.ai/ivy/deep_dive.html) to the docs as an alternative, where we've tried to explain the purpose and functionality of all aspects of Ivy. Feel free to contribute to it if there are any improvements! Thanks @krishna1111111 🙂

krishna1111111:
In my point of you i can say if these are used in short term than these are also for benefits only but I got your point.... please recommend me what else is better...? Apart from this because I have seen many people using this things and are very useful for them and are very beneficial for big organisation also ...all we need is benefits , and those benefits from where we get is best 🙂. Thanks

VedPatwardhan:
> In my point of you i can say if these are used in short term than these are also for benefits only but I got your point.... please recommend me what else is better...? Apart from this because I have seen many people using this things and are very useful for them and are very beneficial for big organisation also ...all we need is benefits , and those benefits from where we get is best 🙂. Thanks

Hey @krishna1111111, in case your interested in contributing to docstrings and other documentation related stuff, feel free to contribute to the [Deep Dive](https://lets-unify.ai/ivy/deep_dive.html) or [Function Reformatting](https://lets-unify.ai/ivy/contributing/open_tasks.html#function-formatting) based on the tasks open . Thanks 🙂

krishna1111111:
Can you please help me in this case ?

krishna1111111:
Should I only look into the function reformatting?

VedPatwardhan:
Not necessarily, you could look into any of the [open tasks](https://lets-unify.ai/ivy/contributing/open_tasks.html). Of course, any fix to the docs would also be appreciated 🙂

krishna1111111:
Please check the once is this ok for the description like has mentioned in docs. I think this description manner ok. So please merge this and proceed the process

VedPatwardhan (commented):
Hey @krishna1111111, as I mentioned before, we shouldn't be adding these types of description comments to files as they'll be hard to maintain, but feel free to work on the open tasks. Thanks 🙂

VedPatwardhan:
Just removed my review request as @jkeane508 is assigned to this PR, thanks 🙂

PR #12879, opened 2023-03-21
dropout3d

ali-aboelezz:


ali-aboelezz:
@vaatsalya123  please check that

vaatsalya123:
Hey @ali-aboelezz first of all thanks for contributing , I just had a look through your PR and there are a few issues which should be resolved before we can merge this:

1.) It seems you have copied the implementation of the testing of the dropout function since both the test functions have the same name  (https://github.com/unifyai/ivy/pull/12879/files#diff-0e21f286cf0cb7da87a70e4184fedc8804af85875f07c71912d8f8cea7da736aR70)

2.) The implementation fails tests and does not seems that generalized to me so I would say please iron out the implementation a bit maybe look through what the function does in more detail and get a more rigorous test to pass with it.

Once you've went through both points we can get this PR merged.

Thanks

ali-aboelezz:
please check it again @vaatsalya123

vaatsalya123:
Hey @ali-aboelezz sorry for the delayed response again , Had a look at your PR but it still does not seem to make alot of sense to me. I don't understand quite why your making a class for testing a function this is not something we do in our testing suite. I also see that you imported torch in the frontend test which is also something you should try to avoid and even if you need to add it make sure it's in a try statement. It would be great if you see other frontend implementations and look how there structure is to get a reference should make it easier to get a perfect implementation, thanks!

ali-aboelezz:
check that please @vaatsalya123

ali-aboelezz:
may be cause i got no review this PR is inactive @vaatsalya123

PR #12881, opened 2023-03-22
Add gamma numpy frontend

### LINKED ISSUE #12758: gamma
Add Random sampling to Numpy Frontend #1530

L-Hinkle:
Close #12758


L-Hinkle:
@ZoeCD Hi Zoe, a few of these errors are the same and I'm not sure how to resolve them-- an example is `run_tests_13` .  I ran into this problem after merging master into my branch. Have you seen this before?

ZoeCD:
The errors don't seem related to your PR, so you don't need to worry about it. Could you please update your PR just to be sure? 🤔

L-Hinkle:
@ZoeCD Hi Zoe, I updated my PR and it passes for all backends locally. In the meantime, a PR was merged for `standard_gamma` that does not pass local backend tests, even without the changes I made in `ivy/functional/backends/jax/experimental/random.py` and `ivy/functional/ivy/experimental/random.py`. Can you review my PR and let me know what you think?

Agecaf:
Hi, I'm the one who opened the PR #13844 yesterday. I apologize, I should've confirmed if someone was already working on this more thoroughly. 

I agree with @L-Hinkle that a proper implementation of `numpy.random.gamma` requires the changes to `ivy.gamma` suggested in this PR.

If I read their documentations correctly, [Numpy](https://numpy.org/doc/stable/reference/random/generated/numpy.random.gamma.html) uses `(shape = alpha, scale = theta = 1/beta)` parameters while other frameworks like [Pytorch](https://pytorch.org/docs/stable/distributions.html#gamma) instead use `(concentration = alpha, rate = beta)`.

This is rather confusing, from the [Wikipedia](https://en.wikipedia.org/wiki/Gamma_distribution);

> There are two equivalent parameterizations in common use:
> - With a [shape parameter](https://en.wikipedia.org/wiki/Shape_parameter) `k` and a [scale parameter](https://en.wikipedia.org/wiki/Scale_parameter) `theta`.
> - With a shape parameter `alpha = k` and an inverse scale parameter `beta = 1/theta` called a [rate parameter](https://en.wikipedia.org/wiki/Rate_parameter).

From Ivy's use of `alpha` and `beta`, I'm assuming Ivy uses the second parameterization, while Numpy explicitely uses the first one, as such I think we need to take the inverse at some point when going from Numpy's to Ivy's code.

ivy/functional/frontends/numpy/random/functions.py
```diff
@@ -85,6 +85,12 @@ def shuffle(x, /):
     return ivy.shuffle(x)
 
 
+@to_ivy_arrays_and_back
+@from_zero_dim_arrays_to_scalar
+def gamma(alpha, beta, /, size=None):
+    return ivy.gamma(alpha, beta, shape=size)
```
Agecaf, line 91: This is where we'd need to invert `beta`, I think.

ivy-seed:

ivy-seed:

PR #12890, opened 2023-03-22
Adding random dtype assigment for out arrays to `test_function`

danielmunioz:
For functions that make use of the out argument, these changes force the generated `out` arrays to have a random dtype as opposed to just copying the original dtype from the input array, haven't yet decided the best approach to collect the available data types, so we're sampling from a placeholder for now.


ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -334,8 +335,11 @@ def test_function(
             if hasattr(ivy.__dict__[fn_name], "out_index")
             else ret
         )
+        # defining random dtype for out array
+        out_dtype = random.choice([ivy.int32, ivy.float32, ivy.complex64])
```
CatB1t, line 339: The usage of random during testing is prohibited, the tests are no longer deterministic (in a way) which would cause flaky resutls.

CatB1t (changes requested):
Why would we want to sample a different dtype for the array with `out` argument?

ivy_tests/test_ivy/helpers/function_testing.py
danielmunioz: Got it, then maybe I can try sample from the dtype strategy provided by the test definition

danielmunioz:
So, it looks like that the expected behavior of the out argument is to keep the out dtype, [this line](https://github.com/unifyai/ivy/blob/9e61213dff2d9527592c9ff4517d884e309dfc23/ivy/func_wrapper.py#L363) exemplifies that, and the current way on which we perform the tests doesn't seem to account for this.

ivy-seed:

ivy-seed:

ivy-seed:

PR #12902, opened 2023-03-22
Print fix for structured container with int key

Dsantra92:
We cannot perform this operation in ivy where the nested dicts have int. This is due a line in ` this_key_chain = key if key_chain == "" else (key_chain + "/" + key)` where we try to add key which is `int` to string value `key_chain` and `"/"`. So I parsed the int as a string for recursive printing.

```python
import ivy

x = ivy.Container(
    {
        "a": ivy.array([1]),
        "b": {
            1: ivy.array([1]),
            2: ivy.array([2]),
        },
    }
)
print(x)
```

### DIFF
--- a/ivy/data_classes/container/base.py
+++ b/ivy/data_classes/container/base.py
@@ -3149,7 +3149,9 @@ class ContainerBase(dict, abc.ABC):
         """
         return_dict = self if inplace else dict()
         for key, value in self.items():
-            this_key_chain = key if key_chain == "" else (key_chain + "/" + key)
+            this_key_chain = (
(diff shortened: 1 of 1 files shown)


zaeemansari70 (commented):
Hi @Dsantra92, 

Actually this code segment is a fundamental part of Ivy, A small note and explaining why you are doing this would be appreciated :slightly_smiling_face:

sherry30 (commented):
Makes sense to me, but just to be sure @saeedashrraf, can you take a look at this since you've worked on containers the most.

saeedashrraf (approved):
Looks good to me!

PR #12908, opened 2023-03-22
to_ivy support for `np.ndarrays` in all backends

hirwa-nshuti:
Hey @VedPatwardhan 

After some exploration here is implementation that might allow all `np.ndarray` instances to be converted to `ivy.Array` and `ivy.NativeArray` irrespective of the backend set.

I tried the solution on the following example:

```python
import ivy
import numpy as np


x = np.array([1., 2., 3.], dtype=np.float32)


backends = ("numpy", "jax", "torch", "tensorflow")

for backend in backends:
    print(f"Current backend: {backend}")
    ivy.set_backend(backend)
    print(ivy.to_ivy(x))
    ivy.unset_backend()
```


VedPatwardhan (changes requested):
Hey @hirwa-nshuti, could you please also add `np.ndarray` in the typehints for `asarray` in `functional/ivy/creations.py` as well, given that we're supporting it for all backends? Thanks 🙂

VedPatwardhan (approved):
lgtm! Feel free to merge, thanks @hirwa-nshuti 🙂

PR #12918, opened 2023-03-22
Sub backends

Infrared1029:


VedPatwardhan (approved):
lgtm! Feel free to merge @Infrared1029 after @CatB1t's approval. Thanks 🙂

ivy/functional/backends/jax/__init__.py
```diff
@@ -233,3 +233,12 @@ def closest_valid_dtype(type=None, /, as_native=False):
 from .experimental import *
 from . import control_flow_ops
 from .control_flow_ops import *
+
+
+# sub-backends
+try:
```
CatB1t, line 239: Is there a reason to have try except statement here? we should have the `sub_backend` there by default, even if it's empty.

ivy/functional/backends/torch/sub_backends/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+def current_sub_backends():
+    return _current_sub_backends
+
+
+def available_sub_backends():
+    return _available_sub_backends
+
+
+if importlib.util.find_spec("xformers"):
```
CatB1t, line 17: This logic should be done dynamically, in other words, this shouldn't be hardcoded in the init file, but rather discover installed subackends

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -0,0 +1,148 @@
...
+_all_sub_backends = _backend_to_sub_backends_dict["torch"]
+
+
+original_backend_dict = None
+
+
+def set_sub_backend(sub_backend_str: str):
+    if ivy.backend == "none":
+        print("You must set a backend first:")
+        available_sub_backends()
```
CatB1t, line 28: let's use `warning.logging` module here, also we shouldn't be printing all the subbackends for all available backends. the list of available subbackends should be shown only when a backend is set.

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -0,0 +1,148 @@
...
+
+import ivy
+from ivy.func_wrapper import _wrap_function
+from ivy.utils.exceptions import IvyException
+
+_sub_backend_dict = dict()
+_backend_to_sub_backends_dict = dict()
+# torch sub_backends
+
+_sub_backend_dict["xformers"] = "ivy.functional.backends.torch.sub_backends.xformers"
```
CatB1t, line 13: Also similiar to the other note on the init file, this should be done dynamically.

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -0,0 +1,148 @@
...
+
+    global original_backend_dict
+    if original_backend_dict is None:
+        original_backend_dict = ivy.__dict__.copy()
+    sub_backend = importlib.import_module(_sub_backend_dict[sub_backend_str])
+    _set_sub_backend_as_ivy(ivy.__dict__.copy(), ivy, sub_backend)
+    ivy.current_backend().sub_backends._current_sub_backends.append(sub_backend_str)
+
+
+def _set_sub_backend_as_ivy(
```
CatB1t, line 59: Let's add some comments here to explain the logic for setting the sub-backend 😄

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -0,0 +1,148 @@
...
+                target.__dict__[k],
+                sub_backend.__dict__[k],
+            )
+
+
+def unset_sub_backend(sub_backend_str: str):
+    if sub_backend_str not in ivy.current_sub_backends():
+        return
+    global original_backend_dict
+    sub_backend = importlib.import_module(
```
CatB1t, line 100: ivy.with_backend instance should support sub-backend, therefore we shouldn't be using the importlib import module, instead use `ivy.utils.dynamic_import.import_module` to work properly in a local ivy instance.

CatB1t (changes requested):

ivy/functional/backends/jax/__init__.py
Infrared1029: mainly because i don't want people adding new backends to be forced to add the sub_backends folder and it's __init__ file, whenever a backend adds a sub_backend they can start adding folders/files/code needed

Infrared1029:
great suggestions, will get them done asap

PR #12986, opened 2023-03-23
Frontend numpy cross function

### LINKED ISSUE #12120: cross

dxchel:
Close #12120


ivy/functional/frontends/numpy/linalg/matrix_and_vector_products.py
```diff
@@ -56,3 +56,8 @@ def tensorsolve(a, b, axes=2):
 def kron(a, b):
     a, b = promote_types_of_numpy_inputs(a, b)
     return ivy.kron(a, b)
+
+
+@to_ivy_arrays_and_back
+def cross(b, a, axisa = -1, axisb = -1, axisc = -1, axis  = None):
+    return ivy.cross(a, b, axisa, axisb, axisc, axis)
```
danielmunioz, line 63: `ivy.cross` only takes the first two arrays as positional arguments the rest should be keyword only

ivy_tests/test_ivy/test_frontends/test_numpy/test_linear_algebra/test_matrix_and_vector_products.py
```diff
@@ -12,6 +12,63 @@
 )
 
 
+@st.composite
+def _get_dtype_value1_value2_axis_for_cross(
```
danielmunioz, line 16: Could you please move your helper strategy to be on top of your test? like how it's [here](https://github.com/unifyai/ivy/blob/23103f5adb8d7bc9a4da63ae5fb9a098dd695ce1/ivy_tests/test_ivy/test_frontends/test_numpy/test_linear_algebra/test_norms_and_other_numbers.py#L14-L94)

danielmunioz:
HI @dxchel! Thanks for the contribution! The tests for your function are failing could you please check those?
Also, frontend functions should mimic the importing behavior of the original framework, for this case it means that `cross` should be able to be imported the same way that we'd do in numpy (directly from the np namespace, like so: `np.cross`) which right now is not the case, this probably means that you need to add `cross` to the linag imports over on `ivy/functional/frontends/numpy/__init__.py`, the same way that `kron`, `tensordot` and others are.

ivy/functional/frontends/numpy/linalg/matrix_and_vector_products.py
dxchel: Hi! Changed the order, my bad, don't know why I had it like that

ivy_tests/test_ivy/test_frontends/test_numpy/test_linear_algebra/test_matrix_and_vector_products.py
dxchel: Done! Thought all helpers should be before any test functions

dxchel:
Thanks for the comments! I'm having a hard time figuring out the way the tests work and how they're called but hope this time it's all good, also hope it follows python best practices

danielmunioz:
Hey! Thanks for the changes, you're almost there! the tests are still failing I'd recommend you to check out the [frontend tests guide](https://lets-unify.ai/ivy/overview/deep_dive/ivy_frontends_tests.html#frontend-test-examples) to understand a bit better how the tests work, but for starters we no longer make us of `fn_name` in the tests, we now use `fn_tree`, so I would recommend you to go through the guide and check out how we implement the tests now, and please feel free to ask any questions you may have over here or over on the frontend tests channel of the [discord server](https://discord.com/invite/sXyFF8tDtm)! It may also help you to run the your tests locally first, this next section is just in case but, you can run a test file using: `pytest <path_to_the_test_file>`, if you wanna run only a specific test you can do: `pytest <path_to_the_test_file>::<test_function>`, so for example. if you wanna run only your test here it'd be: `pytest ivy_tests/test_ivy/test_frontends/test_numpy/test_linear_algebra/test_matrix_and_vector_products.py::test_numpy_outer`
Hope this helps, and again feel free to ask any questions!

dxchel:
Super! Thanks! I'll get on with it, I started checking the guide before but it was a little confusing haha
But thanks! I'll do it

dxchel:
I tested the function manually and it worked correctly, just some kwargs issue I wasn't aware of before but all good
There's an issue here, torch cross function can't take arrays with dimension size < 3, what do I do here?
image

danielmunioz:
Assuming `cross` needs its inputs to have dim < 3, and If the tests are generating arrays with dimensions > 3, I think you can try modifying your strategies to add a `max_num_dims=3` to the problematic array

dxchel:
Ok, made some progress and some findings
Playing with the num dims and dim size I used this:
image
And got an issue with torch backend, it uses a '-1' default and then it tells the user '-1' isn't a valid option:
image
image

But when using the ivy and torch cross functions this happens:
image
As can be seen, even the backend/ivy cross functions aren't giving the same results

dxchel:
I see now, for torch to work the last dimension must be of size 3, so the test is now complete for that using min and max_dim_size=3, is there a way to tell the helper to use this restriction only for the last dimension?
image

dxchel:
Great, thanks a lot!
Studied a little bit the testing strategy for the functional cross and yes! It was made for it to return at least one dimension of size 3, just what I needed!
Added the helper to the np frontend helpers and it worked! It just got one error, I'll dig more into it to see why it's failing:
image

danielmunioz:
Hey @dxchel! That's awesome! Looks alright so far, Just one more thing please!, Could you update your branch to the latest ivy commit please?

dxchel:
Hi @danielmunioz! Thanks a lot for the help! just updated my branch to the latest ivy and added the atol and rtol tolerances to the function (Used some I found in the last function in the script) although I don't know where the not matching results are found, could you point it to me? I ran the test like five times without the tolerances and all the times it passed with no mismatches

danielmunioz:
Yeah, before your latest changes, I tried running the tests locally and they seemed to pass as well, however I found out about this because of Ivy's GitHub CI tests (the checks where you can see the lint failures as well as other tests), particularly the `intelligent-tests-pr / run_test_1` check was failing, and there is where the tests found the discrepancies between the backends.
In any case, everything looks good to me! I'm merging it now!
Thanks for the contribution!! :)

PR #13001, opened 2023-03-23
updated set.py with more examples

ManasVardhan:
added more examples for unique_values()
added 1D and 2D examples for showing flattening functionality added examples for both ivy.Array and ivy.Container


ManasVardhan:
Close #12949

PR #13004, opened 2023-03-23
Making static container methods private

soma2000-lang:
Solves https://trello.com/c/SwTr0PxY  opened by @VedPatwardhan

### DIFF
--- a/ivy/data_classes/container/activations.py
+++ b/ivy/data_classes/container/activations.py
@@ -122,7 +122,7 @@ class _ContainerWithActivations(ContainerBase):
         }
 
         """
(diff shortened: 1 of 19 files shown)


VedPatwardhan (changes requested):
Thanks for making the changes @soma2000-lang. Could you please also update all instances where a `static` method is called as well with the `_` at the start, such as [this](https://github.com/unifyai/ivy/pull/13004/files#diff-a04fcd15d75a9abc7ca08678397552270a211f9cd58f938a1c34a5658da642c8R125)? Thanks 🙂

PR #13034, opened 2023-03-23
geomspace #12444

### LINKED ISSUE #12444: geomspace

dash96:
Close #12444

### DIFF
--- a/ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
+++ b/ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
@@ -43,6 +43,19 @@ def meshgrid(*xi, copy=True, sparse=False, indexing="xy"):
     if copy:
         return [ivy.copy_array(x) for x in ret]
     return ret
+    
+
+@handle_numpy_dtype
+@to_ivy_arrays_and_back
+def geomspace(start, stop, num=50, endpoint=True, dtype=None, axis=0):
+    cr = ivy.log(stop / start) / (num - 1 if endpoint else num)
+    x = ivy.linspace(0, cr * (num - 1 if endpoint else num), num, endpoint=endpoint, axis=axis)
+    x = ivy.exp(x)
+    x = start * x
+    x[0] = (start*cr)/cr
+    if endpoint:
+        x[-1] = stop
+    return x.asarray(dtype=dtype)
 
 
 class nd_grid:
@@ -158,14 +171,3 @@ class OGrid(nd_grid):
 
 
 ogrid = OGrid()
-
-
-@handle_numpy_dtype
-@to_ivy_arrays_and_back
-def geomspace(start, stop, num=50, endpoint=True, dtype=None, axis=0):
-    start, stop = promote_types_of_numpy_inputs(start, stop)
-    cr = ivy.log((stop / start), where=start != 0) / (num - 1 if endpoint else num)
-    x = ivy.linspace(0, cr * (num - 1 if endpoint else num), num, endpoint=endpoint, dtype=dtype, axis=axis)
-    x = ivy.exp(x)
-    x = start * x
-    return x
--- a/ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
+++ b/ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
@@ -1,5 +1,4 @@
 # global
-from ivy.functional.frontends.numpy import promote_types_of_numpy_inputs
 
 import ivy
 from ivy.functional.frontends.numpy.func_wrapper import (
--- a/ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
+++ b/ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
@@ -1,5 +1,4 @@
 # global
-
 import ivy
 from ivy.functional.frontends.numpy.func_wrapper import (
     outputs_to_numpy_arrays,
@@ -51,7 +50,7 @@ def geomspace(start, stop, num=50, endpoint=True, dtype=None, axis=0):
     x = ivy.linspace(0, cr * (num - 1 if endpoint else num), num, endpoint=endpoint, axis=axis)
     x = ivy.exp(x)
     x = start * x
-    x[0] = (start*cr)/cr
+    x[0] = start*cr
     if endpoint:
         x[-1] = stop
     return x.asarray(dtype=dtype)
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
```diff
@@ -156,3 +156,27 @@ def __init__(self):
...
+
+
+@handle_numpy_dtype
+@to_ivy_arrays_and_back
+def geomspace(start, stop, num=50, endpoint=True, dtype=None, axis=0):
+    # Check if stop < start and swap values if necessary
+    flip = 0
+    if stop < start:
+        start, stop = stop, start
+        flip = 1
```
fspyridakos, line 168: This won't work with multi-dimensional inputs, start and stop don't have to be scalars, they can also be arrays, you can probably perform something similar by using `where` and the axis argument

ivy_tests/test_ivy/test_frontends/test_numpy/test_creation_routines/test_numerical_ranges.py
```diff
@@ -261,3 +264,35 @@ def test_numpy_ogrid(range, class_, method_name):
     ret = helpers.flatten_and_to_np(ret=ret)
     ret_np = helpers.flatten_and_to_np(ret=ret_np)
     helpers.value_test(ret_np_flat=ret, ret_np_from_gt_flat=ret_np, rtol=1e-03)
+
+
+@handle_frontend_test(
+    fn_tree="numpy.geomspace",
+    dtype_start_stop=_get_dtype_and_range(dimension=0),
```
fspyridakos, line 271: If you do what I pointed out in the other comment this would be unecessary

ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
```diff
@@ -156,3 +156,27 @@ def __init__(self):
...
+    # Compute the common ratio
+    cr = ivy.log(stop / start) / (num - 1 if endpoint else num)
+    # Generate the sequence of values using the common ratio
+    if dtype is None:
+        dtype = start.dtype
```
fspyridakos, line 173: You should do `start, stop = promote_types_of_numpy_inputs(x1, x2)`
and then `dtype=start.dtype`

fspyridakos (changes requested):
Hey, good start but there are a few changes, mainly the support for multi-dimensitonal inputs for start and end

ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
dash96: Can I instead use "for i in range(start)" and keep the entire code as is
and then at the end just append() each row

ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
```diff
@@ -156,3 +156,25 @@ def __init__(self):
...
+            start, stop = promote_types_of_numpy_inputs(x1, x2)
+            dtype = start.dtype
+        x = ivy.linspace(0, cr * (num - 1 if endpoint else num), num, endpoint=endpoint, dtype=dtype, axis=axis)
+        x = ivy.exp(x)
+        x = start[i] * x
+        b = ivy.reshape(x, (num,1))
+        if endpoint:
+            b[-1] = stop[i]
+        a.append(b)
+    return ivy.concatenate(a, axis=1)
```
fspyridakos, line 180: Quite a few comments here, firstly the promotion should be written like this: `start, stop = promote_types_of_numpy_inputs(start, stop)` and `promote_types_of_numpy_inputs` should be imported. That should be the first thing you do it will also convert start and stop to arrays so those conversions can be removed as well. The `if dtype is None` statement in unecessary. Also the for loop in unecessary and will slow things down A LOT. The main functions here are `ivy.log`, `ivy.linspace` and `ivy.exp`, all those are broadcastable so a for loop in unecessary remove it and keep the rest except the indexing variables `i` and `j` and the last 5 lines. Just return x after line 175.

ivy_tests/test_ivy/test_frontends/test_numpy/test_creation_routines/test_numerical_ranges.py
```diff
@@ -32,8 +32,11 @@ def _get_range_for_grid(draw):
 
 
 @st.composite
-def _get_dtype_and_range(draw):
-    dim = draw(helpers.ints(min_value=2, max_value=5))
+def _get_dtype_and_range(draw, dimension=None):
+    if dimension!=None:
+        dim = dimension
+    else:
+        dim = draw(helpers.ints(min_value=2, max_value=5))
```
fspyridakos, line 39: After the changes I mentioned this will be unecessary as the function should accept mutlidimensional inputs for start and stop

fspyridakos (changes requested):
Left a few comments, overall the implementation is there, you just need to oprimize it and remove a lot of unecessary steps especially the for loop

ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
```diff
@@ -156,3 +158,14 @@ def __init__(self):
 
 
 ogrid = OGrid()
+
+
+@handle_numpy_dtype
+@to_ivy_arrays_and_back
+def geomspace(start, stop, num=50, endpoint=True, dtype=None, axis=0):
+    start, stop = promote_types_of_numpy_inputs(start, stop)
```
fspyridakos, line 166: I have no idea as to why but importing `promote_types_of_numpy_inputs` causes some very strange import error. You can remove this line though and the above import and the function will still pass

ivy/functional/frontends/numpy/creation_routines/numerical_ranges.py
```diff
@@ -156,3 +158,14 @@ def __init__(self):
...
+    cr = ivy.log((stop / start), where=start != 0) / (num - 1 if endpoint else num)
+    x = ivy.linspace(0, cr * (num - 1 if endpoint else num), num, endpoint=endpoint, dtype=dtype, axis=axis)
+    x = ivy.exp(x)
+    x = start * x
+    return x
```
fspyridakos, line 171: It seems like numpy has some logic which ensures the start and end elemnets are exactly the start and end elementes specified, you can add this by including the following 3 lines:
```
x[0] = start
if endpoint:
    x[-1] = stop
```

fspyridakos (changes requested):
Very close! Just these finishing touches, also the function is in an awkward position in the file, could you please move it such that it's after `meshgrid`. Thanks

PR #11941, opened 2023-03-08
Add nanmedian

rishabgit:
Close #11878

PR #13069, opened 2023-03-24
Add nanmedian

### LINKED ISSUE #11878: nanmedian

rishabgit:
Close #11878

### DIFF
--- a/ivy/data_classes/container/experimental/statistical.py
+++ b/ivy/data_classes/container/experimental/statistical.py
@@ -676,7 +676,7 @@ class _ContainerWithStatisticalExperimental(ContainerBase):
         *,
         axis: Optional[Union[Tuple[int], int]] = None,
         keepdims: bool = False,
-        overwrite_input: Optional[bool] = False,
+        overwrite_input: bool = False,
         key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
         to_apply: bool = True,
         prune_unapplied: bool = False,
@@ -737,7 +737,7 @@ class _ContainerWithStatisticalExperimental(ContainerBase):
         *,
         axis: Optional[Union[Tuple[int], int]] = None,
         keepdims: bool = False,
-        overwrite_input: Optional[bool] = False,
+        overwrite_input: bool = False,
         out: Optional[ivy.Container] = None,
     ) -> ivy.Container:
         """ivy.Container instance method variant of ivy.nanmedian. This method simply
--- a/ivy/data_classes/container/experimental/statistical.py
+++ b/ivy/data_classes/container/experimental/statistical.py
@@ -676,6 +676,7 @@ class _ContainerWithStatisticalExperimental(ContainerBase):
         *,
         axis: Optional[Union[Tuple[int], int]] = None,
         keepdims: bool = False,
+        overwrite_input: Optional[bool] = False,
         key_chains: Optional[Union[List[str], Dict[str, str]]] = None,
         to_apply: bool = True,
         prune_unapplied: bool = False,
@@ -697,6 +698,8 @@ class _ContainerWithStatisticalExperimental(ContainerBase):
         keepdims
             If this is set to True, the axes which are reduced are left in the result
             as dimensions with size one.
+        overwrite_input
+            If True, then allow use of memory of input array for calculations.
         out
             optional output array, for writing the result to.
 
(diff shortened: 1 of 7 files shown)


rishabgit:
@hirwa-nshuti  Created new PR as per prev comment -https://github.com/unifyai/ivy/pull/11941#issuecomment-1480049881

ivy/data_classes/container/experimental/statistical.py
```diff
@@ -676,6 +676,7 @@ def static_nanmedian(
         *,
         axis: Optional[Union[Tuple[int], int]] = None,
         keepdims: bool = False,
+        overwrite_input: Optional[bool] = False,
```
juliagsy, line 679: the type hints should be bool as only `True` or `False` are allowed as values

juliagsy (changes requested):
Hey there! Great work so far! Just a few small changes before the PR is good to go! As usual, please request my review again when you're done with the changes, thanks!

ivy/data_classes/container/experimental/statistical.py
rishabgit: @juliagsy  Need slight clarification on this - replacing the line with following work?
`overwrite_input: bool = False,`

juliagsy: yep, that's right! Optional[..] is not needed (:

juliagsy (commented):
Hey there! Thanks for the changes so far, they look great! Just one last thing, although the functions and methods are implemented in the ivy API, the tests are not present yet, could you please add a test for this `nanmedian` function in its related test file? (`test_ivy/test_functional/test_experimental/test_core/test_statistical.py`) Thanks a lot!

rishabgit:
Hey @juliagsy - I followed earlier convention of nanmean in that file to figure the hypothesis and testing parameters. Fixed a couple issues while testing, but I'm out of ideas on the errors popping up for Pytorch and Numpy

Updated the code and attached the error message after running the test - 
```
pytest ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py::test_nanmedian
```

image

image

juliagsy (commented):
for the torch error, it is likely related to the unsupported types in native torch, it should be solvable with the decorator, more information here: https://lets-unify.ai/docs/ivy/overview/deep_dive/data_types.html#supported-and-unsupported-data-types

as for the numpy error, I'll double check to see if it is actually related to the PR/implementation, for now, you may work on fixing the torch related tests first! Thanks!

rishabgit:
> for the torch error, it is likely related to the unsupported types in native torch, it should be solvable with the decorator, more information here: https://lets-unify.ai/docs/ivy/overview/deep_dive/data_types.html#supported-and-unsupported-data-types
> 
> as for the numpy error, I'll double check to see if it is actually related to the PR/implementation, for now, you may work on fixing the torch related tests first! Thanks!

Added the unsupported dtypes. Getting a weird pytest output, couldn't pinpoint which snippet presents the actual error. 
In case you'd like to run the test on your machine (output kept getting truncated on terminal, so saving it on a new file) - 
```
pytest ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py::test_nanmedian -rA > pytest_output.txt
```

ivy/functional/backends/torch/experimental/statistical.py
```diff
@@ -97,7 +97,10 @@ def corrcoef(
 
     return torch.corrcoef(xarr)
 
-
+@with_unsupported_dtypes(
+    {"1.11.0 and below": ("bfloat16", "bfloat32", "float16", "float32", "float64")}, \
```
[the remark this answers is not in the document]
rishabgit: Updated the code

juliagsy:
Hello! I'll leave a review soonest possible! Thanks!

juliagsy (commented):
Hello! Thanks for the hard work! The code looks good so far but there's one final test failure left, the log is shown below:
EDIT: The pasted log didn't appear well, I would suggest that you click into the `intelligent-tests-pr/run_tests_1` workflow, open the "Determine and Run Tests" tab, the failure is shown inside, if the log is too long, just search `test_nanmedian`, one of the search will bring you to it.

Once this is fixed, we can then get it merged! Thanks a lot!

rishabgit:
Hi there! I've updated the code to fix the issue you mentioned. However, it seems like there's a new issue related to torch itself. It looks like torch computes the median differently, as noted in their documentation here: https://pytorch.org/docs/stable/generated/torch.median.html. 
Specifically, when dtype_x_axis=(['float32'], [array([-1., -2.], dtype=float32)], 0) is input, an AssertionError occurs with the message "the results from backend torch and ground truth framework tensorflow do not match -2.0!=-1.5". Let me know if you have any thoughts on how to proceed. Thanks!

juliagsy:
Hello! Thanks for the information! I'll look into this and try to get back soon within one or two days!

juliagsy (commented):
Hey! Apologies for the delayed response! I've checked and it would be best to implement a new argument `mean_when_even=True` (default to True) in the functional API, so you would need to add this to all backend and the ivy function itself. This is to ensure that both behaviours are unified. The value should just be True by default, then in the test, both True and False should be tested. Thanks!

PR #13186, opened 2023-03-26
refactored ivy_tests package

Aarsh2001:


ivy_tests/array_api_testing/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def pytest_collection_modifyitems(items):
+    skip_ivy = mark.skip(reason="ivy skip - see ivy_tests/skips.txt for details")
+    for item in items:
+        # skip if specified in skips.txt
+        for id_ in skip_ids:
+            if item.nodeid.startswith(id_):
+                item.add_marker(skip_ivy)
+                break
```
CatB1t, line 22: Missing new line

CatB1t (changes requested):
Just a small change, then feel free to merge! great work Aarsh. Also are we storing any of the results of the `test_array_api` tests to our Redis Database? If not we can move the logic from the main conftest to the inner `test_ivy` directory, otherwise lgtm! Thanks 😄

PR #13192, opened 2023-03-26
np.array conversion works on tf tensor 

Aarsh2001:
Task - https://trello.com/c/C83YFeZ9
This code snippet works now -:

```
import ivy
import numpy as np

for fw in ["jax", "numpy", "torch", "tensorflow"]:
    try:
        ivy.set_backend(fw)
        a = ivy.array(1e-5)
        b = np.array(a)
    except Exception as e:
        print(f"Error in {fw}: {e}")
```

Problem - The EagerTensorBase class' `__array__` method in tensorflow returns a scalar when dtype is not passed. This is what it looks like -:

```
def __array__(self, dtype=None):
    a = self._numpy()
    if not dtype:
      return a

    return np.array(a, dtype=dtype)
```

We solve this by manually passing the dtype inside our `__array__` method


guillesanbri:
I guess that @VedPatwardhan is probably better suited to review this one, I'm not entirely sure if this has any further implication :sweat_smile:

VedPatwardhan (approved):
lgtm! Let's get the PR merged and see if anything goes wrong, which is unlikely to happen given that we aren't utilizing the `np.array(<ivy.Array>)` pattern anywhere in the framework. Thanks @Aarsh2001, @guillesanbri, @RashulChutani 🙂

Aarsh2001:
Sure, thanks.

PR #13200, opened 2023-03-26
added logaddexp2

UbajakaCJ:
I added the function logaddexp2 to calculate the logarithm of the sum of the exponentiations "log2(2**x1 + 2**x2)" for each element "x1_i" of the input array "x1" with the respective element "x2_i" of the input array "x2"

Please check out the function and give your feedback.


MahmoudAshraf97:
Hi @UbajakaCJ , Thanks for your contribution.
in order to add a function to the functional API, it needs to be added in all 5 backends and then added to the ivy framework, also it should be in the experimental folder.
Is there any need for this function that existing functions can't do using a composite solution?

UbajakaCJ:
Hello @MahmoudAshraf97 , thanks for the feedback. Yes, there is a need for the function logaddexp2 because it calculates the Logarithm of the sum of the exponentiations, the logarithms being in base 2, as opposed to the previous function logaddexp, which calculates the natural logarithm (in base e) of the sum of exponentiations.  

The function is useful in Machine Learning when the calculated probabilities of events may be so small as to exceed the range of normal floating point numbers. In such cases, the base-2 logarithm of the calculated probability can be used instead.

It has the Numpy version as np.logaddexp2, so it makes sense to develop the logaddexp2 function and extend it to the Ivy framework.

UbajakaCJ:
Not really the same thing, because it is the logarithm in base 2 of the summation of  2 raised separately to the power of the two elements, x_1, and x_2. Like this ``log2(2**x1 + 2**x2)``  and not ``log2(e**x1 + e**x2)``

And it is meant to be an easy function that can be called, without the complexity of making it compositional.

Like the way you call np.log and np.log10

It is meant to be handy for ML when the calculated probabilities of events may be so small as to exceed the range of normal floating point numbers.

MahmoudAshraf97:
Yes my bad it's not the same but the idea still holds, there are certain requirements to add a new backend function which are not satisfied here, because there's a good composite solution

MahmoudAshraf97:
Ok, can you please open a new issue using the "Missing Function Suggestion" template? also to extend Ivy functional API [ivy//ivy/functional/experimental](https://github.com/unifyai/ivy/blob/master/ivy/functional/experimental/). Each function will need implementation in each [backend API](https://lets-unify.ai/ivy/deep_dive/0_navigating_the_code.html#backend-api) and also the [Ivy API](https://lets-unify.ai/ivy/deep_dive/0_navigating_the_code.html#ivy-api). The [tests](https://lets-unify.ai/ivy/deep_dive/15_ivy_tests.html) will also need to be implemented. See the [Deep Dive](https://lets-unify.ai/ivy/deep_dive.html) for more general information on how the functions should be implemented. Particular attention should be placed on the [Superset Behaviour](https://lets-unify.ai/ivy/deep_dive/2_superset_behaviour.html) section when considering how best to design the function, and what features and arguments it should have. More information about this specific task is provided in the [Open Tasks](https://lets-unify.ai/ivy/contributing/4_open_tasks.html#ivy-experimental-api) section of the contributor guide. All the implementations have to be placed in `experimental` directories for both, functional API, backend APIs, Array, and Container instances.

MahmoudAshraf97:
Hi, I found that `logaddexp2` is already in ivy experimental API, so I'm going to close this

PR #13213, opened 2023-03-26
Sub backends

Infrared1029:
Created a new PR because I kinda messed up the old one with a merge


ivy/functional/backends/jax/__init__.py
```diff
@@ -233,3 +233,12 @@ def closest_valid_dtype(type=None, /, as_native=False):
 from .experimental import *
 from . import control_flow_ops
 from .control_flow_ops import *
+
+
+# sub-backends
+try:
```
CatB1t, line 239: Following the discussion on the previous PR:
"mainly because I don't want people adding new backends to be forced to add the sub_backends folder and it's __init__ file, whenever a backend supports a sub_backend they can start adding folders/files/code needed" 

I think keeping a consistent structure is better so it's clear for everyone, also aligns with the current design of Ivy API structure (although it's required to implement all functions in the backends), but on a higher level, keeping the same structure is better. 

Another reason, one of the internal imports may fail, and that way of checking if a package exists or not may actually fail and never give a proper Exception if the sub_backend failed to import one of the imports internally.

ivy/functional/backends/torch/sub_backends/__init__.py
CatB1t, line 1: We can move the detection of the sub-backends as a function in the `handler` or another module that fits better, so it can used for different backends.

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -0,0 +1,187 @@
+import os
+from types import ModuleType, FunctionType
+import warnings
+
+import ivy
+from ivy.func_wrapper import _wrap_function
+from ivy.utils.exceptions import IvyException
+
+warnings.simplefilter("always", UserWarning)
```
CatB1t, line 9: Let's use `logging` module, it's a better fit according to Python docs: 

> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning

> logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -0,0 +1,187 @@
...
+
+[_all_sub_backends.extend(v) for v in _backend_to_sub_backends_dict.values()]
+
+
+original_backend_dict = None
+
+
+def set_sub_backend(sub_backend_str: str):
+    if ivy.backend == "none":
+        warnings.warn("You must set a backend first:")
```
CatB1t, line 56:
```suggestion
        warnings.warn("You must set a backend first")
```

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -0,0 +1,187 @@
...
+        return
+
+    if ivy.current_backend_str() not in _backend_to_sub_backends_dict.keys():
+        warnings.warn(
+            f"backend {ivy.current_backend_str()} does not have any supported sub_backends"  # noqa
```
CatB1t, line 61: Let's break the line into multiple strings, no need for `#noqa`, we can also remove the `_` from `sub_backends`

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -0,0 +1,187 @@
...
+            ] = f"{_backends_subpackage_path}.{backend}.sub_backends.{sub_backend}"
+            try:
+                _backend_to_sub_backends_dict[backend].append(sub_backend)
+            except KeyError:
+                _backend_to_sub_backends_dict[backend] = [sub_backend]
+
+
+_all_sub_backends = []
+
+[_all_sub_backends.extend(v) for v in _backend_to_sub_backends_dict.values()]
```
CatB1t, line 48: We're constructing a list here and throwing it away, let's use a normal for loop.

CatB1t (changes requested):

Infrared1029:
Hopefully this does it!

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -0,0 +1,204 @@
...
+    if backend.startswith("__"):
+        continue
+
+    sub_backends_dir = os.path.join(
+        ivy.__path__[0].rpartition(os.path.sep)[0],
+        _backends_subpackage_path.replace(".", os.path.sep),
+        backend,
+        "sub_backends",
+    )
+    if os.path.isdir(sub_backends_dir):
```
CatB1t, line 32: Is this check here necessary?

CatB1t (changes requested):
Just a small change, then feel free to merge! 🥳 great work @Infrared1029 !

ivy/utils/backend/sub_backend_handler.py
Infrared1029: oh this was there before we decided to remove the `try and except` imports block, now it's useless, good catch! will get it removed:)

CatB1t (approved):
lgtm! well done @Infrared1029!

PR #13252, opened 2023-03-27
unique_consecutive implemented in torch frontend

chanpreet007:
Please Review the code and let me know any changes if so.


fspyridakos:
Hi there, thanks for the contribution! I see you've tried to add this function to the ivy API but your solution is incomplete as you've only added the torch backend. Your PR should actually not inlcude any additions to the Ivy API, instead you should try and make the function with the functions already present in ivy. In this case, however, I think this functions should be added to ivy API as explained here: https://lets-unify.ai/ivy/overview/deep_dive/ivy_frontends.html#missing-ivy-functions

chanpreet007:
I am bit confused. I am bit new so please clarify ..Here what I did

1). Add function to PyTorch backend (which just calls torch function to do work).
2). Add same function to ivy frontend which will just call this backend function added in step 1.
3). Add a torch fronted function which will call ivy function added in step 2.

As you suggested I need to add make function by using already existing functions in ivy....... In torch frontend or backend or ivy frontend?

fspyridakos:
So the problem with your step one, is that you only did it for the torch, backend, we have 4 others that would need to be implemented and then a test needs to be added as well. This is probably pretty challenging for someone new so for the frontend related tasks we ask you to not add anything new to the Ivy API and just work with what's already there. In this case I believe this function should be added to the ivy API, so I suggested you follow the instructions here: https://lets-unify.ai/ivy/overview/deep_dive/ivy_frontends.html#missing-ivy-functions, and if you're bold you can try implementing the funciton in the Ivy API in a seperate PR. Otherwise you can make the issue as described but make the function compositionaly in the frontend as described here: https://lets-unify.ai/ivy/overview/deep_dive/ivy_frontends.html#temporary-compositions

ivy-seed:

ivy-seed:

PR #13294, opened 2023-03-28
cumsum_paddle_

Tristan1649:
Task cumsum paddle #13246
The function first checks if the exclusive flag is set to True. 
If so, it adds a zero tensor to the beginning of the tensor along the specified axis. 
This is because an exclusive cumulative sum excludes the current element from the sum.
Then, the paddle.cumsum function is called on the input tensor x, with the specified axis.
And if the reverse flag is set to True, the output tensor is reversed along the specified axis.
Then function returns the computed cumulative sum tensor.

### DIFF
--- a/ivy/functional/backends/paddle/statistical.py
+++ b/ivy/functional/backends/paddle/statistical.py
@@ -2,8 +2,6 @@
 
 torch_scatter = None
 from typing import Union, Optional, Sequence
-
-
 import paddle
 
 # local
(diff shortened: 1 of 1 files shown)


ivy/functional/backends/paddle/statistical.py
```diff
@@ -227,9 +228,15 @@ def cumsum(
     dtype: Optional[paddle.dtype] = None,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    raise IvyNotImplementedException()
-
-
+    if exclusive:
+        x = paddle.concat((paddle.zeros_like(x.slice([None] * axis + [slice(0, 1)], [-1] * (axis + 1))),
```
MahmoudAshraf97, line 232: ```slice() missing 1 required positional argument: 'ends'```
What exactly are you trying to do in this line?

Issue #13078, opened 2023-03-24
cumprod


avinashyadav0027:

sahilempire:
I want to work on this issue, can you please assign me to this? @avinashyadav0027

PR #13364, opened 2023-03-29
added cumprod to paddle.statistical

avinashyadav0027:
- [ ] #13078

### DIFF
--- a/ivy/functional/backends/paddle/statistical.py
+++ b/ivy/functional/backends/paddle/statistical.py
@@ -233,18 +233,19 @@ def cumprod(
         x = paddle.cumprod(paddle.flip(x, axis=(axis,)), dim=axis)
         with ivy.ArrayMode(False):
             x = ivy.swapaxes(x, axis, -1)
-            x = paddle.concat((paddle.ones_like(x[..., -1:]), x[..., :-1]), -1)
+            x = ivy.concat((ivy.ones_like(x[..., -1:]), x[..., :-1]), axis=-1)
             x = ivy.swapaxes(x, axis, -1)
-        return paddle.flip(x, axis=(axis,)).cast(dtype)
+            return ivy.flip(x, axis=(axis,)).cast(dtype)
     elif exclusive:
         with ivy.ArrayMode(False):
             x = ivy.swapaxes(x, axis, -1)
-            x = paddle.concat((paddle.ones_like(x[..., -1:]), x[..., :-1]), -1)
+            x = ivy.concat((ivy.ones_like(x[..., -1:]), x[..., :-1]), axis=-1)
             x = paddle.cumprod(x, -1)
             return ivy.swapaxes(x, axis, -1).cast(dtype)
     else:
-        x = paddle.cumprod(paddle.flip(x, axis=(axis,)), dim=axis)
-        return paddle.flip(x, axis=axis).cast(dtype)
+        with ivy.ArrayMode(False):
+            x = paddle.cumprod(paddle.flip(x, axis=(axis,)), dim=axis)
+            return ivy.flip(x, axis=axis).cast(dtype)
 
 
 @with_unsupported_device_and_dtypes(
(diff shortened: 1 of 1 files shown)


ivy/functional/backends/paddle/statistical.py
```diff
@@ -10,12 +10,32 @@
...
 from ivy.utils.exceptions import IvyNotImplementedException
 from . import backend_version
-from ivy.func_wrapper import with_unsupported_dtypes, with_unsupported_device_and_dtypes
+from ivy.func_wrapper import with_unsupported_device_and_dtypes
 
 # Array API Standard #
 # -------------------#
 
 
+def swap_axes(x, axis1, axis2):
```
MahmoudAshraf97, line 19: swap_axes already exists in manipulation functions, no need to reimplement

ivy/functional/backends/paddle/statistical.py
```diff
@@ -212,7 +240,30 @@ def cumprod(
     dtype: Optional[paddle.dtype] = None,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    raise IvyNotImplementedException()
+    dtype = ivy.as_native_dtype(dtype)
```
MahmoudAshraf97, line 243: the dtype handling part will not yield the expected results, you are casting to the closeset valid dtype which will return the same dtype because all of them are valid, also there's no need for infer dtype,`x.dtype` is enough and will result in native dtype, try to implement it using ivy.swapaxes

MahmoudAshraf97 (commented):
Hi, thanks for the contribution. Please review the requested changes and request another review after making sure that the function is passing the hypothesis tests

ivy/functional/backends/paddle/statistical.py
```diff
@@ -242,25 +222,26 @@ def cumprod(
...
     if not (exclusive or reverse):
         return paddle.cumprod(x, dim=axis).cast(dtype)
     elif exclusive and reverse:
         x = paddle.cumprod(paddle.flip(x, axis=(axis,)), dim=axis)
-        x = swap_axes(x, axis, -1)
-        x = paddle.concat((paddle.ones_like(x[..., -1:]), x[..., :-1]), -1)
-        x = swap_axes(x, axis, -1)
+        with ivy.ArrayMode(False):
+            x = ivy.swapaxes(x, axis, -1)
+            x = paddle.concat((paddle.ones_like(x[..., -1:]), x[..., :-1]), -1)
```
MahmoudAshraf97, line 236: use `ivy.concat` and `ivy.ones_like`

MahmoudAshraf97 (changes requested):
make sure that all the paddle functions such as flip, concat, that you use support all valid dtypes, if not, the use their ivy implementation using array mode just as you did earlier, keep up the good work

ivy/functional/backends/paddle/statistical.py
```diff
@@ -212,7 +220,32 @@ def cumprod(
     dtype: Optional[paddle.dtype] = None,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    raise IvyNotImplementedException()
+    dtype = ivy.as_native_dtype(dtype)
+    if dtype is None:
+        dtype = x.dtype
+    if dtype in [paddle.uint8, paddle.int8, paddle.int16]:
+        x = paddle.cast(x, ivy.default_int_dtype())
```
MahmoudAshraf97: the default int dtype might still be `int16`, so cast to `float32` instead

ivy/functional/backends/paddle/statistical.py
```diff
@@ -212,7 +220,32 @@ def cumprod(
     dtype: Optional[paddle.dtype] = None,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    raise IvyNotImplementedException()
+    dtype = ivy.as_native_dtype(dtype)
```
MahmoudAshraf97: If `dtype == None` this will return `None`, it should be
```suggestion
    dtype = dtype if dtype is not None else x.dtype
```

MahmoudAshraf97 (approved):

PR #13372, opened 2023-03-29
Added Compiling and Transpiling docs

KareemMAX:


docs/compiler/compiler.rst
```diff
@@ -0,0 +1,325 @@
...
+
+    # Compile the function
+    compiled_fn = ivy.compile(fn, args=(x, y))
+
+    # View the graph
+    compiled_fn.show()
+
+In this case, the compiled graph would be:
+
+.. image:: https://raw.githubusercontent.com/unifyai/unifyai.github.io/master/img/externally_linked/compiler/figure1.png
```
guillesanbri, line 92: This image seems too big, is there a way to standardize the size based on the nodes? I guess probably the easiest thing is to resize the image with extra background

guillesanbri (commented):
just left a comment regarding the first image, I'll leave a couple of extra points in the trello card but lgtm overall :+1:

docs/compiler/compiler.rst
guillesanbri: fixed this

PR #13380, opened 2023-03-29
fixed issue

rajshukla1102:
Close https://github.com/unifyai/ivy/issues/10421

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_numpy/test_indexing_routines/test_indexing_like_operations.py
+++ b/ivy_tests/test_ivy/test_frontends/test_numpy/test_indexing_routines/test_indexing_like_operations.py
@@ -1,5 +1,6 @@
 # Testing Function
 # global
+import numpy as np
 from hypothesis import strategies as st
 
 # local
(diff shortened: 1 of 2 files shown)


rajshukla1102:
@hirwa-nshuti look into once. Thankyou!

ivy_tests/test_ivy/test_frontends/test_numpy/test_indexing_routines/test_indexing_like_operations.py
```diff
@@ -1,5 +1,6 @@
 # Testing Function
 # global
+import numpy as np
```
Dsantra92, line 3: Is is required here?

Dsantra92 (changes requested):
Numpy converts each value to int64. You might want to cast each value in the tuple to int64. It should fix the current tests.

rajshukla1102 (commented):
@Dsantra92 I have made the changes. please look into it once.

rajshukla1102:
@zhumakhan Thanks for the review but currently I think ivy is not supporting orders. It is the reason I didn't add support. Thank you!

zhumakhan:
Yeah, you are right.
Lets merge then!

Issue #12985, opened 2023-03-23
lerp


plapacz6:

varun-tokas:
Is someone working on this issue? I'd like to give it a try otherwise

plapacz6:
Yes, I've been trying since yesterday.
I entered the function code similar to the one from fronted, except for the torch backend where I called the lerp functions from the torch framework. Now trying to write a test and then a docstring.

khushi-411:
Closing the issue in favor of #13399. Please feel free to re-open if you have any problems.

PR #13399, opened 2023-03-30
lerp #12985

plapacz6:
lerp #12985
Can I have a review code please.
I'm stuck on a test issue for tensorflow backend
Can I ask for any suggestions.

As a model backend I chose torch, due to the fact that the lerp function is only available in this framework.
I limited the available float data types to two in the tests.
The function is a mix function. Only backed torch has native support.

thank you in advance


plapacz6:
[lerp #12985](https://github.com/unifyai/ivy/pull/13399#top)#13399
image
on my repository
tests : .../test_elementwise.py::test_lerp
for the lerp function all pass already.

Please let me know if PR is suitable for merging or if I should improve something.

ivy/functional/ivy/experimental/elementwise.py
```diff
@@ -1421,6 +1421,87 @@ def ldexp(
...
+    weight: Union[ivy.Array, ivy.NativeArray, float],
+    /,
+    *,
+    out: Optional[ivy.Array] = None,
+) -> ivy.Array:
+    """
+    Returns a linear interpolation of two arrays start (given by input) and end
+    based on a scalar or array weight.
+        input + weight * (end - input),  element-wise.
+    <https://pytorch.org/docs/stable/generated/torch.lerp.html>
```
MahmoudAshraf97, line 1441:
```suggestion
```

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_elementwise.py
```diff
@@ -1181,6 +1181,46 @@ def test_ldexp(
     )
 
 
+# lerp
+@handle_test(
+    fn_tree="functional.ivy.experimental.lerp",
+    dtype_and_input=helpers.dtype_and_values(
+        available_dtypes=["float32", "float64"],
```
MahmoudAshraf97, line 1188:
```suggestion
        available_dtypes=helpers.get_dtypes("valid"),
```

MahmoudAshraf97 (changes requested):
Good work, I think you need to revise the handling of the mixed functions this is handled using the handle_mixed_function decorator in func_wrapper. If I'm not mistaken, there should be an attribute on the method in the Functional API like so: func.mixed_function = True and the @handle_mixed_function decorator would be applied on the backend implementation with which you want to decide on choosing the primary implementation or the composition one. For example: https://github.com/unifyai/ivy/blob/9affbccde21f522db09cde254c77466086b7aaa9/ivy/functional/backends/tensorflow/experimental/layers.py#L407-L412 (the backend implementation) and https://github.com/unifyai/ivy/blob/9affbccde21f522db09cde254c77466086b7aaa9/ivy/functional/ivy/experimental/layers.py#L1418 (the functional attribute setting) 

also change the dtypes in the tests as suggested and if there are any errors fix them, other than that we are ready to merge

plapacz6:
Thank you for the tips. I changed the available types in the test as suggested. I added the @handle_mixed_function decorator in the backend functions. I'm not sure if I correctly handled the cases where the output for the function is not valid. If anything needs to be changed, please let me know

ivy/functional/ivy/experimental/elementwise.py
```diff
@@ -1421,6 +1421,97 @@ def ldexp(
...
+    if (
+        not isinstance(input, (float, int, complex))
+        or not isinstance(end, (float, int, complex))
+    ):
+        input = ivy.astype(input, "float64")
+        end = ivy.astype(end, "float64")        
+    if (
+        not isinstance(weight, float)
+    ):
+        weight = ivy.astype(weight, "float64")
```
MahmoudAshraf97, line 1508: Supposedly these variables are arrays,, so you need to check for the dtype of the array not if it's and instance,, that's if there is a needed check at all needed

MahmoudAshraf97 (changes requested):
LGTM, will merge after the requested changes

plapacz6:
Thanks for the hint. Type checking has grown a bit spatially. Not sure if I didn't complicate it too much. If anything needs to be changed, please let me know.

MahmoudAshraf97 (approved):

PR #13445, opened 2023-03-30
Vayuusaee patch 1

vayuusaee:


hirwa-nshuti (changes requested):
Hey @vayuusaee
Thanks for contributing to ivy😊
It seems like your changes involves your newly introduced shell file.
Which is not being reflected on this PR.
Also would you try to explain in details what type of problem your PR is fixing.

Thanks

vayuusaee:
the above line we have modified the postCreateCommand property to execute a shell script located

hirwa-nshuti:
> the above line we have modified the postCreateCommand property to execute a shell script located

Would you tell the reason why you need to create another script if there was an issue that you might be facing?

ivy-seed:

PR #13447, opened 2023-03-30
Added an if statement to check whether there are any changes to stash…

vayuusaee:
… before doing so.

This version checks whether there are any changes to stash before doing so, and uses git stash save instead of git stash to provide a description of the stash. It also checks whether there is a stash to pop before doing so, to avoid errors if there are no stashes to apply.

### DIFF
--- a/stash_pull.sh
+++ b/stash_pull.sh
@@ -1,4 +1,19 @@
+#!/bin/bash
+
+# Set git pull.rebase mode to true
 git config pull.rebase true
-git stash
+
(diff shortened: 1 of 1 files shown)


valeriocost:
Hi @vayuusaee 😄 I'm not sure what value it adds, I'm closing the PR but you are welcome to discuss it further on our Discord channel 😃 In the meanwhile, take a look at open tasks that you would like to contribute to. Thanks!

PR #13450, opened 2023-03-30
added support for numpy ndarrays in backends for dtype

hirwa-nshuti:
Given that we support `numpy.ndarray` when creating `ivy.Array` and `ivy.NativeArray` instances in all backends.

It is worth it to have dtype functions support `numpy.ndarray` and `numpy.dtype` instances.
This PR extends the functions' ability to support the `ndarrays` accross all the backends.

Example before this PR:
```python
import ivy
import numpy as np
ivy.set_backend('jax')
x = np.array([1., 2., 3.])
print(ivy.finfo(x)) # raising some exceptions
ivy.unset_backend()
```


ivy/functional/backends/jax/data_type.py
```diff
@@ -213,13 +222,15 @@ def as_native_dtype(dtype_in: Union[jnp.dtype, str, bool, int, float], /) -> jnp
         )
 
 
-def dtype(x: JaxArray, *, as_native: bool = False) -> ivy.Dtype:
+def dtype(x: Union[JaxArray, np.ndarray], *, as_native: bool = False) -> ivy.Dtype:
     if as_native:
-        return ivy.to_native(x).dtype
+        return ivy.as_native_dtype(as_ivy_dtype(x.dtype))
```
VedPatwardhan, line 227: Wouldn't this just work as `ivy.as_native_dtype(x.dtype)`, given that `as_native_dtype` also works with `np.dtype`? 
(Same may apply to all backends)

VedPatwardhan (changes requested):
Hey @hirwa-nshuti, just a small change, rest seems good to go, let's see what @fspyridakos thinks, thanks both 🙂

ivy/functional/backends/jax/data_type.py
```diff
@@ -131,12 +131,16 @@ def broadcast_to(
 
 
 @_handle_nestable_dtype_info
-def finfo(type: Union[jnp.dtype, str, JaxArray], /) -> Finfo:
+def finfo(type: Union[jnp.dtype, str, JaxArray, np.ndarray], /) -> Finfo:
+    if isinstance(type, np.ndarray):
+        type = type.dtype.name
     return Finfo(jnp.finfo(ivy.as_native_dtype(type)))
```
fspyridakos, line 137: This function is not meant to accept ndarrays, just dtype objects

ivy/functional/backends/jax/data_type.py
```diff
@@ -131,12 +131,16 @@ def broadcast_to(
...
 
 
 @_handle_nestable_dtype_info
-def iinfo(type: Union[jnp.dtype, str, JaxArray], /) -> np.iinfo:
+def iinfo(type: Union[jnp.dtype, str, JaxArray, np.ndarray], /) -> np.iinfo:
+    if isinstance(type, np.ndarray):
+        type = type.dtype.name
```
fspyridakos, line 143: Same as above

ivy/functional/backends/jax/data_type.py
```diff
@@ -213,13 +222,15 @@ def as_native_dtype(dtype_in: Union[jnp.dtype, str, bool, int, float], /) -> jnp
...
-        return ivy.to_native(x).dtype
+        return ivy.as_native_dtype(x.dtype)
     return as_ivy_dtype(x.dtype)
 
 
-def dtype_bits(dtype_in: Union[jnp.dtype, str], /) -> int:
+def dtype_bits(dtype_in: Union[jnp.dtype, str, np.dtype], /) -> int:
+    if isinstance(dtype_in, np.dtype):
+        dtype_in = dtype_in.name
     dtype_str = as_ivy_dtype(dtype_in)
```
fspyridakos, line 234: You already modified `as_ivy_dtype` to accept np.dtype why make the conversion here?

fspyridakos (changes requested):
Hey Felix, a few comments, I just wrote them in the jax file but they apply to all. Thanks

ivy/functional/backends/jax/data_type.py
hirwa-nshuti: Yeah makes sense this is redundant

ivy/functional/backends/jax/data_type.py
hirwa-nshuti: Actually from the array API they specified that they can accept array as well
https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.iinfo.html

ivy/functional/backends/jax/data_type.py
hirwa-nshuti: Makes sense

VedPatwardhan (approved):
lgtm! Feel free to merge @hirwa-nshuti after @fspyridakos's approval. Thanks 🙂

fspyridakos (approved):
Looks good to me too Felix, feel free to merge

PR #13489, opened 2023-03-31
Fixed dynamic backend setting to create the new arrays on the same device as the original

VedPatwardhan:


CatB1t (approved):
lgtm! great work @VedPatwardhan.

Issue #13560, opened 2023-03-31
mask_indices


pharrera:

pharrera:
https://github.com/unifyai/ivy/pull/13569

hirwa-nshuti:
This item was completed in https://github.com/unifyai/ivy/pull/14028

PR #13569, opened 2023-03-31
Added mask_indices indexing function to Jax frontend.

pharrera:
https://github.com/unifyai/ivy/issues/13560


valeriocost:
Hello @pharrera, could you make sure that only the relevant files to the PR are committed? There are a few more files that are not related 😄

PR #13576, opened 2023-04-01
data mode casting

RickSanchezStoic:
data mode casting enabled. You can set the flag `ivy.downcast_data_types` , `ivy.upcast_data_types` and `ivy.crosscast_data_types` for functions to default to certain dtypes if the input dtype is not supported by the function.


VedPatwardhan (changes requested):
The changes look good to merge, could you please also add the explanation for the purpose and way of using this type casting wrapper somewhere in the deep dive and our policy with unsupported dtypes? Would be really helpful for the community to know this. Thanks @RickSanchezStoic 🙂

PR #13578, opened 2023-04-01
added clip_by_global_norm tensorflow frontend and test"

### LINKED ISSUE #13367: clib_by_global_norm

Plutone11011:
Close #13367


ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -37,6 +37,17 @@ def clip_by_value(t, clip_value_min, clip_value_max):
     return ivy.clip(t, clip_value_min, clip_value_max)
 
 
+@to_ivy_arrays_and_back
```
Daniel4078, line 40: E   tensorflow.python.framework.errors_impl.InvalidArgumentError: Value for attr 'T' of int64 is not in the list of allowed values: bfloat16, half, float, double, complex64, complex128
E   	; NodeDef: {{node Sqrt}}; Op<name=Sqrt; signature=x:T -> y:T; attr=T:type,allowed=[DT_BFLOAT16, DT_HALF, DT_FLOAT, DT_DOUBLE, DT_COMPLEX64, DT_COMPLEX128]> [Op:Sqrt]

Daniel4078 (changes requested):
Found a small error in your code that the function does not support all datatypes, so please check the deep dive document for the function wrapper that states what datatype is allowed for this function. Thanks.

ivy/functional/frontends/tensorflow/general_functions.py
Plutone11011: If I understand correctly, there are some datatypes that should not be supported by this function?

How do you check this locally? My tests didn't find this error

Daniel4078: Just found a more significant error: 
E   AssertionError:  the results from backend tensorflow and ground truth framework tensorflow do not match
E    [-0.]!=[-1.9073486e-06] 
E   
E   
E   Falsifying example: test_tensorflow_clip_by_global_norm(
E       input_and_norm=(['float32', 'float32'],
E        [array([-1.], dtype=float32), array([-0.5], dtype=float32)],
E        ['float32'],
E        [array([1.1754944e-38], dtype=float32)],
E        ['float32'],
E        [array([3.081488e-33], dtype=float32)]),
E       fn_tree='ivy.functional.frontends.tensorflow.clip_by_global_norm',
E           with_out=False,
E           inplace=False,
E       ),
E       on_device='cpu',
E   )
E   
E   You can reproduce this example by temporarily adding @reproduce_failure('6.70.1', b'AXicY2QAA0YGBM14gAEOjl+AM49tQFYKAFaoA9M=') as a decorator on your test case

Plutone11011:
I limited supported types to `float32` as other float types yielded errors in some backends. Also, the assertion inconsistency should be fixed now

Daniel4078 (approved):
Thanks, the test is passing. Well done.

PR #13601, opened 2023-04-02
added LocalIvyImporter ctx manager

Infrared1029:


VedPatwardhan (commented):
Given that this is `with_backend`-related, I think @CatB1t should have the say on this PR. Thanks @Infrared1029 and @CatB1t 🙂

ivy/utils/backend/handler.py
```diff
@@ -631,24 +629,22 @@ def with_backend(backend: str, cached: bool = False):
     if cached and backend in compiled_backends.keys():
         return compiled_backends[backend][-1]
     # TODO do error handling if finder fails
```
CatB1t, line 631: Let's get this TODO removed as it's done.

ivy/utils/_importlib.py
```diff
@@ -1,9 +1,27 @@
+import sys
+from .backend import ast_helpers
```
CatB1t, line 2: Let's switch the order of imports, (line 2 and line 4), also no need to add an empty line between. I'd also recommend using absolute import instead of relative, just to keep things consistent.

CatB1t (changes requested):
great work @Infrared1029, just little changes to formatting and we're good to merge!

Infrared1029:
Done!

CatB1t (approved):
LGTM!

PR #12950, opened 2023-03-22
added gather function to paddle backend

czirabence:
Close #12592

PR #13622, opened 2023-04-02
uint8, int32 and int64 support for 'gather' in paddle backend

### LINKED ISSUE #12592: gather

czirabence:
Close #12592

### DIFF
--- a/ivy/functional/backends/paddle/general.py
+++ b/ivy/functional/backends/paddle/general.py
@@ -169,8 +169,8 @@ def gather(
             return ivy.reshape(result, shape=new_shape)
         
     
-    axis = axis % paddle.Tensor.ndimension(params)
-    batch_dims = batch_dims % paddle.Tensor.ndimension(params)
+    axis = axis % params.ndim
+    batch_dims = batch_dims % params.ndim
     ivy.utils.assertions.check_gather_input_valid(params, indices, axis, batch_dims)
     if params.dtype in [
         paddle.int8,
@@ -185,7 +185,7 @@ def gather(
                 _gather(params.real()),
                 _gather(params.imag())
             )
-        return _gather(params.cast("float64")).cast(params.dtype)
+        return _gather(params.cast("float32")).cast(params.dtype)
     return _gather(params)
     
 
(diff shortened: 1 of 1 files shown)


czirabence:
> I don't really follow what you've done here, can you explain to me your steps?

This is a correction for my previous pull request #12950, which added gather to paddle backend. Paddle.gather supports dtypes of uint8, int32, int64, float32 and float64 on cpu, according to the [paddlepaddle reference](ttps://www.paddlepaddle.org.cn/documentation/docs/en/2.2/api/paddle/gather_en.html#gather). However, when functional.ivy.gather with paddle backend was given uint8 type array inputs, the following error was raised: ivy.utils.exceptions.IvyBackendException: paddle: gather: (NotFound) The kernel with key (CPU, NCHW, uint8) of kernel `slice` is not registered. To overcome this, 4 list comprehensions inside ivy.functional.bakcends.paddle.gather were replaced with paddle.split. I also removed int32 and int64 from unsupported_devices_and_dtypes, because they are supported.

MahmoudAshraf97 (changes requested):
Good work
the correct usage would be `paddle.unbind` instead of `split`, while both of them return the correct result, the shape is different because `split` keeps the axis that you split on while `unbind` removes that axis which is the desired behavior, and since unbind doesn't support a lot of dtypes, you can use `ivy.unstack` under `ivy.ArrayMode(False)`
btw try to extend to all dtypes except bfloat and uint16

czirabence:
Thank you,
Paddle functions were replaced with ivy functions in the second commit, and data casting for unsupported types was added, which I copied from backends.paddle.unstack.

ivy/functional/backends/paddle/general.py
```diff
@@ -160,27 +144,50 @@ def gather(
...
+                    )
+                result = ivy.concat(result, axis=0)
+            new_shape = (
+                params1.shape[:axis] + indices.shape[batch_dims:] + params1.shape[axis + 1 :]
+            )
+            return ivy.reshape(result, shape=new_shape)
+        
+    
     axis = axis % paddle.Tensor.ndimension(params)
     batch_dims = batch_dims % paddle.Tensor.ndimension(params)
```
MahmoudAshraf97, line 173:
```suggestion
    axis = axis % params.ndim
    batch_dims = batch_dims % params.ndim
```

MahmoudAshraf97 (approved):
Great work overall

PR #13637, opened 2023-04-02
add jax.random.beta frontend

### LINKED ISSUE #13497: beta

MuhammedAshraf2020:
closes #13497


MuhammedAshraf2020:
Hi Julia, 
firstly, the idea behing _get_seed is to generate a seed that contain all the information from the prng_key, so that (0, 42) is different from (1,42) for the user who wrote his jax script, so we can't depend only on the key[1].

secondly, rng_key must be of dtype uint32, but torch doesn't has this dtype, i tried to solve it but i found that the the rest of the functions using unint32, so i think you don't mind of using this dtype.

finaly, it works for float, i tried numeric but i doesn't work well, so it's for just "float" if you would ask me to try "numeric".

thank you :")

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_random.py
```diff
@@ -122,3 +122,74 @@ def call():
...
+        helpers.get_shape(
+            min_num_dims=2, max_num_dims=2, min_dim_size=1, max_dim_size=5
+        )
+    )
+    dtype = draw(
+        helpers.get_dtypes("float", full=False).filter(
+            lambda x: "float16" not in x[0] and "bfloat16" not in x[0]
+        )
+    )
+    return dtype_key, key, alpha, beta, shape, dtype
```
fspyridakos, line 152: No need to make a composite function for all this, you could shove it in the `handle_frontend_test`, you only need to use this when one generated value depends on another, here none do

MuhammedAshraf2020:
@fspyridakos 
Done

fspyridakos:
Thanks for your contribution, good work! Although I would like to ask you to avoid opening any more PRs for jax.random since I need to refactor the tests to work with all backends.

PR #13659, opened 2023-04-03
Close #13658 (incorrect `in` usage in torch matmul test)

### LINKED ISSUE #13658: test_torch_matmul
- [ ] #12236

EmilPi:
Close #13658


juliagsy:
Hi @EmilPi ,

Thanks for the PR!

However, looking more closely, it’s unclear how this PR would meaningfully contribute towards the project. There is actually no issues with the test so I’ve simply closed it for now 🙂

Please feel free to submit other PRs based on our Open Tasks.
PS: They'll be labelled `ToDo`, please do not select task from `ToDo_internal` as that's for our team's internal reference.

Thanks!

EmilPi:
This fixed the test which I've run locally, because `in` on tuple of tensors caused syntax error. Maybe I missed something.
Okay, at least I have checked that the process is correct.

PR #13775, opened 2023-04-06
extending ivy creation functions to allow shape passed as multiple positional arguments

hello-fri-end:
[Trello Task](https://trello.com/c/DkDXtN7W/1632-extend-ivy-creation-functions)


hello-fri-end:
Hi @AnnaTz! Is this the correct way to go about updating the creation functions for this task? Once you approve these changes, I'll update the rest of the functions. Also, I had to add `num_positional_args=st.just(1)` in the tests otherwise `shape` was getting passed as a keyword argument, is that ok?

hello-fri-end:
> > Hi @AnnaTz! Is this the correct way to go about updating the creation functions for this task? Once you approve these changes, I'll update the rest of the functions. Also, I had to add `num_positional_args=st.just(1)` in the tests otherwise `shape` was getting passed as a keyword argument, is that ok?
> 
> Hi Anwaar, your approach is not exactly what I had in mind . Could you have a look at the implementation and testing of torch frontend's zeros?

Thanks for the suggestion @AnnaTz. The frontend approach allows us to pass shape both as a positional argument containing a list, tuple, or a sequence of integers or as a keyword argument. I have updated the functions. Can you take a look again and let me know if there's still anything that I should change?

AnnaTz:
I would suggest to rename `args` to something more intuitive like `sizes`. But other than that looks good to me!

hello-fri-end:
> I would suggest to rename `args` to something more intuitive like `sizes`. But other than that looks good to me!

Done. I have updated `ones`, `zeros`, and `empty`. 

`full` has a positional argument called `fill_value` which will get converted to a keyword argument if we add `*size` to the args. We could keep `fill_value` as the first positional argument but that would be against the array API standard. Even torch doesn't allow a sequence of integers passed as positional arguments for `shape` in `torch.full`. So, I guess we shouldn't update it.

AnnaTz:
> > I would suggest to rename `args` to something more intuitive like `sizes`. But other than that looks good to me!
> 
> Done. I have updated `ones`, `zeros`, and `empty`.
> 
> `full` has a positional argument called `fill_value` which will get converted to a keyword argument if we add `*size` to the args. We could keep `fill_value` as the first positional argument but that would be against the array API standard. Even torch doesn't allow a sequence of integers passed as positional arguments for `shape` in `torch.full`. So, I guess we shouldn't update it.

Agreed. Then I guess what's remaining is to update the array data class (I think you forgot that) and to update the tests.

hello-fri-end:
> > > I would suggest to rename `args` to something more intuitive like `sizes`. But other than that looks good to me!
> > 
> > 
> > Done. I have updated `ones`, `zeros`, and `empty`.
> > `full` has a positional argument called `fill_value` which will get converted to a keyword argument if we add `*size` to the args. We could keep `fill_value` as the first positional argument but that would be against the array API standard. Even torch doesn't allow a sequence of integers passed as positional arguments for `shape` in `torch.full`. So, I guess we shouldn't update it.
> 
> Agreed. Then I guess what's remaining is to update the array data class (I think you forgot that) and to update the tests.

None of these functions are implemented as array instance methods since they don't have any array arguments. The static container method for `ivy.empty` was missing and I have added it now. As for updating the tests, how do I pass a variable number of positional arguments to the helpers.test_function for the `size` argument?

AnnaTz:
> how do I pass a variable number of positional arguments to the helpers.test_function for the size argument?

Here is an example https://github.com/unifyai/ivy/blob/520c235d46c05335b94fb2ac87b853237c522972/ivy_tests/test_ivy/test_frontends/test_numpy/test_creation_routines/test_numerical_ranges.py#L166

hello-fri-end:
> > how do I pass a variable number of positional arguments to the helpers.test_function for the size argument?
> 
> Here is an example
> 
> https://github.com/unifyai/ivy/blob/520c235d46c05335b94fb2ac87b853237c522972/ivy_tests/test_ivy/test_frontends/test_numpy/test_creation_routines/test_numerical_ranges.py#L166

Thanks for the example. Have updated the tests now :)

AnnaTz:
Are the tests passing locally? I don't see the functions in the CI but I think the PR is ready to be merged.

hello-fri-end:
> Are the tests passing locally? I don't see the functions in the CI but I think the PR is ready to be merged.

yes all the tests are passing locally. I'll merge the PR now. Thanks for all the help 🙏

PR #13802, opened 2023-04-06
added profiler class

Infrared1029:
profiler class that allows code profiling, usage example:
```python
with Profiler(print_stats=False, viz=True):
    fn(x, y)
```

### DIFF
--- a/ivy/utils/profiler.py
+++ b/ivy/utils/profiler.py
@@ -1,0 +1,54 @@
+import cProfile
+import pstats
+import subprocess
+import logging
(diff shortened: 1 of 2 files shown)


VedPatwardhan (commented):
Hey @Infrared1029, thanks for adding the Profiler, I think it would be great if we had a description for this somewhere in the docs, as it's very likely to go unnoticed otherwise. Would be happy to know if you think we should add this in the docs. Rest looks good to merge, thanks @Infrared1029 🙂

PR #13804, opened 2023-04-07
Fix frontend imports when a backend is set

CatB1t:
- Replaced ivy.<dtype> calls with newly constructed data types, this eliminates the dependency on the current ivy backend.


fspyridakos:
My only question is if the dtypes you create in the numpy and jax frontend get replaced later since `np.int32` for example should call the numpy frontend `int32` scalar class (as numpy does). Same thing goes for `jax.numpy.int32`, it also call the numpy frontend `int32` scalar class. If that's the case this looks good to me!

CatB1t:
It depends on the order of which are imported first or not, but given that the frontends provide their own dtypes classes later on, I'll update the ones for NumPy to be private so it doesn't override the scalar class. Thanks, @fspyridakos !

PR #13814, opened 2023-04-07
lp_pool2d torch

### LINKED ISSUE #13785: lp_pool2d

Plutone11011:
Close #13785

### DIFF
--- a/ivy/functional/frontends/torch/nn/functional/pooling_functions.py
+++ b/ivy/functional/frontends/torch/nn/functional/pooling_functions.py
@@ -214,7 +214,6 @@ def lp_pool2d(input, norm_type, kernel_size, stride=None, ceil_mode=False):
             padding,
             data_format=data_format,
         )
-
-    return ivy.pow(
-        ivy.multiply(out, kernel_size[0] * kernel_size[1]), ivy.divide(1.0, norm_type)
-    )
+    if not isinstance(kernel_size, int):
+        kernel_size = kernel_size[0] * kernel_size[1]
+    return ivy.pow(ivy.multiply(out, kernel_size), ivy.divide(1.0, norm_type))
--- a/ivy/functional/frontends/torch/nn/functional/pooling_functions.py
+++ b/ivy/functional/frontends/torch/nn/functional/pooling_functions.py
@@ -191,3 +191,30 @@ def lp_pool1d(input, norm_type, kernel_size, stride=None, ceil_mode=False):
         )
 
     return ivy.pow(ivy.multiply(out, kernel_size), ivy.divide(1.0, norm_type))
+
+
+@to_ivy_arrays_and_back
+def lp_pool2d(input, norm_type, kernel_size, stride=None, ceil_mode=False):
+
+    data_format = "NCHW"
+    padding = "VALID"
+    if stride is not None:
+        out = ivy.avg_pool2d(
+            ivy.pow(input, norm_type),
+            kernel_size,
+            stride,
+            padding,
+            data_format=data_format,
+        )
+    else:
+        out = ivy.avg_pool2d(
+            ivy.pow(input, norm_type),
+            kernel_size,
+            kernel_size,
+            padding,
+            data_format=data_format,
+        )
+
+    return ivy.pow(
+        ivy.multiply(out, kernel_size[0] * kernel_size[1]), ivy.divide(1.0, norm_type)
+    )
--- a/ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
+++ b/ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
@@ -269,3 +269,42 @@ def test_torch_lp_pool1d(
         stride=stride[0],
         ceil_mode=False,
     )
+
+
+# avg_pool2d
+@handle_frontend_test(
+    fn_tree="torch.nn.functional.lp_pool2d",
+    dtype_x_k_s=helpers.arrays_for_pooling(
+        min_dims=4,
+        max_dims=4,
+        min_side=1,
+        max_side=4,
+    ),
+    norm_type=helpers.number(min_value=0.1, max_value=6),
+    test_with_out=st.just(False),
+)
+def test_torch_lp_pool2d(
+    dtype_x_k_s,
+    norm_type,
+    *,
+    test_flags,
+    frontend,
+    fn_tree,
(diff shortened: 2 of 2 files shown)


ivy/functional/frontends/torch/nn/functional/pooling_functions.py
```diff
@@ -191,3 +191,30 @@ def lp_pool1d(input, norm_type, kernel_size, stride=None, ceil_mode=False):
...
+        out = ivy.avg_pool2d(
+            ivy.pow(input, norm_type),
+            kernel_size,
+            kernel_size,
+            padding,
+            data_format=data_format,
+        )
+
+    return ivy.pow(
+        ivy.multiply(out, kernel_size[0] * kernel_size[1]), ivy.divide(1.0, norm_type)
```
Infrared1029, line 219: We are assuming that `kernel_size` is a `tuple` here, this is not necessarily that case as it can be of type `int`.

ivy/functional/frontends/torch/nn/functional/pooling_functions.py
```diff
@@ -191,3 +191,30 @@ def lp_pool1d(input, norm_type, kernel_size, stride=None, ceil_mode=False):
         )
 
     return ivy.pow(ivy.multiply(out, kernel_size), ivy.divide(1.0, norm_type))
+
+
+@to_ivy_arrays_and_back
+def lp_pool2d(input, norm_type, kernel_size, stride=None, ceil_mode=False):
```
Infrared1029, line 197: `ceil_mode` seems to be ignored throughout the implementation, I noticed that `ivy.avg_pool2d` does not include that parameter, while `ivy.max_pool2d` does, we should be supporting `ceil_mode` in `ivy.avg_pool2d` as well.

ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
```diff
@@ -269,3 +269,42 @@ def test_torch_lp_pool1d(
...
+    *,
+    test_flags,
+    frontend,
+    fn_tree,
+    on_device,
+):
+    input_dtype, x, kernel_size, stride, _ = dtype_x_k_s
+    # Torch ground truth func expects input to be consistent
+    # with a channels first format i.e. NCW
+    x[0] = x[0].reshape((x[0].shape[0], x[0].shape[-1], *x[0].shape[1:-1]))
```
Infrared1029, line 298: You should not be using `reshape` to permute the shape of a tensor, use `x[0].transpose(...)` instead (since x[0] is a numpy array).

Infrared1029:
Thanks for the PR @Plutone11011, left a few comments!

ivy/functional/frontends/torch/nn/functional/pooling_functions.py
Plutone11011: As of now `ceil_mode` is not handled in the ivy api though, so for this specific function it should ok to leave like this for the time being right?

ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
Plutone11011: Out of curiosity, why is reshaping wrong here?

ivy/functional/frontends/torch/nn/functional/pooling_functions.py
Infrared1029: Actually, yes we can ignore it for now.

ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
Infrared1029: Glad you asked actually, you can try this yourself
```python
import numpy as np

a = np.array([[1,2,3], [4,5,6]])
v1 = a.reshape(3, 2)
v2 = a.transpose(1, 0)
v1 == v2 # False
```
You can think of reshaping as flattening the array and then rearranging it in the specified requested shape, transposing or permuting just returns a different view of the same tensor, it does not change its internal memory structure or move things around.

```python
img[height][width][channel] == img.transpose(2, 0, 1)[channel][height][width] # True
img[height][width][channel] == img.reshape(channel_dims, height_dims, width_dims)[channel][height][width] # False
```

ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
```diff
@@ -269,3 +269,42 @@ def test_torch_lp_pool1d(
...
+    helpers.test_frontend_function(
+        input_dtypes=input_dtype,
+        test_flags=test_flags,
+        frontend=frontend,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        input=x[0],
+        norm_type=norm_type if norm_type > 0 else 1,
```
Infrared1029, line 307: this check does not seem to be necessary as the `norm_type` has a minimum value of 0.1 so it is never going to be below 0

ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
```diff
@@ -269,3 +269,42 @@ def test_torch_lp_pool1d(
...
     )
+
+
+# avg_pool2d
+@handle_frontend_test(
+    fn_tree="torch.nn.functional.lp_pool2d",
+    dtype_x_k_s=helpers.arrays_for_pooling(
+        min_dims=4,
+        max_dims=4,
+        min_side=1,
```
Infrared1029, line 280: we could probably increase these numbers a bit?

ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
Plutone11011: I agree, but in practice I tried testing without it and I saw that even with `min_value` set, `norm_type` still sometimes has value 0. I don't know why, maybe a bug in `helpers.number`?

Plutone11011: How do I proceed here?

Infrared1029: I just checked and you are right, It's all good on your side, I'll merge and report this behavior to the team.

PR #13844, opened 2023-04-07
Added frontend numpy.random.gamma.

### LINKED ISSUE #13841: gamma
numpy.random.gamma

Agecaf:
Closes #13841
See [Numpy's documentaiton](https://numpy.org/doc/stable/reference/random/generated/numpy.random.gamma.html) and [the Gamma distribution](https://en.wikipedia.org/wiki/Gamma_distribution).
Numpy uses shape parameter k (which is equal to alpha), and scale parameter theta (which is 1/beta).
I think there might be issues with `ivy.gamma`; in `ivy/functional/ivy/experimental/random.py` I assume `shape` is the same as in `ivy.beta`, `ivy.dirichlet`, `ivy.poisson` and the other distributions, yet the docstring and the type are significantly different. Gamma distributions have a "shape parameter" which is "alpha", I think there might've been a confusion there.
Numpy's `gamma` and `standard_gamma` both have an optional `size=None` parameter which should coincide with Ivy's `shape`, currently neither `gamma` nor `standard_gamma` in the frontend implement it.

### DIFF
--- a/ivy/functional/frontends/numpy/random/functions.py
+++ b/ivy/functional/frontends/numpy/random/functions.py
@@ -79,6 +79,14 @@ def beta(a, b, size=None):
 
 @to_ivy_arrays_and_back
 @from_zero_dim_arrays_to_scalar
+def gamma(shape, scale=1.0, size=None):
+    # Numpy uses shape for the shape parameter k, which is alpha in Ivy.
+    # Numpy uses scale for the scale parameter theta, which is 1/beta in Ivy.
+    return ivy.gamma(shape, 1.0 / scale, dtype="float64")  # , shape=size)
+
+
+@to_ivy_arrays_and_back
+@from_zero_dim_arrays_to_scalar
 def shuffle(x, /):
     if isinstance(x, int):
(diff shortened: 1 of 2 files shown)


Agecaf:
On closer inspection, `ivy/functional/experimental/random.py` has in lines 173, (notice `shape`)
```
    return ivy.current_backend().gamma(
        shape, alpha, beta, device=device, dtype=dtype, seed=seed, out=out
    )
```
While, for example, `ivy/functional/backends/jax/experimental` has
```
def gamma(
    alpha: Union[float, JaxArray],
    beta: Union[float, JaxArray],
    /,
    *,
    shape: Optional[Union[ivy.NativeShape, Sequence[int]]] = None,
    device: Optional[jaxlib.xla_extension.Device] = None,
    seed: Optional[int] = None,
) -> JaxArray:
```

Agecaf:
This Task was already been worked on in #12881. Should this PR be closed @valeriocost ?

valeriocost:
I think it would be better to work there since it has been opened before 😄 thanks for flagging it!

PR #13855, opened 2023-04-08
Added sub function under torch tensor

utkarshraj779:


nathzi1505:
Thanks @utkarshraj779 for the commits. Will go ahead and merge.

PR #13875, opened 2023-04-08
add: backend paddle corrcoef

### LINKED ISSUE #13817: corrcoef

trantuantdt:
Close #13817


MahmoudAshraf97:
Hi, can you explain what are the changes in the testing function and why are they needed?

trantuantdt:
Hi @MahmoudAshraf97. if the input array has a column where all rows have the same value, that column's variance is 0. All the values divided by 0 will be replaced by nan. There will be incorrect results returned by Jax and Paddle backend. So I replace all the columns and rows ( in case "rowvar" is False) which have the same value in every position.

trantuantdt:
I can't fork from the PaddlePaddle branch. It doesn't exist anymore.

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
```diff
@@ -173,22 +173,47 @@ def test_quantile(
...
+    )
+
+    dtype, value = dtype_and_x
+    x = value[0]
+    y = value[1]
+    dim_1 = x.shape[0]
+    dim_2 = x.shape[1]
+    for i in range(dim_1):
+        if len(np.unique(x[i, :])) == 1:
+            x[i] = np.random.uniform(-100, 100, dim_2)
```
CatB1t, line 201: The usage of `np.random` is prohibited, it doesn't provide a shrinking feature, which makes fixing a failing test harder. We should be using the strategies provided by both Ivy and Hypothesis to generate our data.

CatB1t (changes requested):

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_statistical.py
trantuantdt: Is that ok to replace all rows and columns which have the same value at every position here?

ivy-seed:

trantuantdt:
I would like to continue working on this PR

CatB1t (changes requested):
Hello @trantuantdt , the tests seems to be passing fine for all backends on master without changing the behaviour or the generated data, I don't think we should be changing the strategy, please try to sync your branch with the Ivy's master branch. We should try and fix the `paddle` backend to return the same resutls from the ground truth framework `tensorflow`

PR #13904, opened 2023-04-09
Nested support for arbitrary objects' attributes

xoiga123:
```python
output = kornia.feature.disk.disk.DISK()(torch.rand([1,3,224,224])) 
# output == [DiskFeatures]
# https://kornia.readthedocs.io/en/latest/_modules/kornia/feature/disk/structs.html#DISKFeatures
ivy.nested_any(output, lambda x: isinstance(x, torch.Tensor))  # False
ivy.nested_any(output, lambda x: isinstance(x, torch.Tensor), check_attributes=True)  # True

ivy.nested_argwhere(output, lambda x: isinstance(x, torch.Tensor))  # []
ivy.nested_argwhere(output, lambda x: isinstance(x, torch.Tensor), check_attributes=True)  # [[0]]
```

Do we need `nested_map` and `nested_multi_map` too? Modifying an object's attribute might be scary for an unaware user.

### DIFF
--- a/ivy/functional/ivy/nest.py
+++ b/ivy/functional/ivy/nest.py
@@ -612,6 +612,7 @@ def nested_argwhere(
     nest: Iterable,
     fn: Callable,
     check_nests: bool = False,
+    check_attributes: bool = False,
     to_ignore: Optional[Union[type, Tuple[type]]] = None,
     _index: Optional[List] = None,
     _base: bool = True,
@@ -629,6 +630,9 @@ def nested_argwhere(
         The conditon function, returning True or False.
     check_nests
         Whether to also check the nests for the condition, not only nest leaves.
+        Default is ``False``.
+    check_attributes
+        Whether to also check the attributes of leaf objects for the condition.
         Default is ``False``.
     to_ignore
         Types to ignore when deciding whether to go deeper into the nest or not
@@ -707,6 +711,7 @@ def nested_argwhere(
                     item,
                     fn,
                     check_nests,
+                    check_attributes,
                     to_ignore,
                     _index + [i],
                     False,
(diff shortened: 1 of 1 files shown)


ivy/functional/ivy/nest.py
```diff
@@ -1199,18 +1217,28 @@ def nested_any(
...
-            if nested_any(v, fn, check_nests, False, extra_nest_types):
+            if nested_any(
+                v, fn, check_nests, check_attributes, False, extra_nest_types
+            ):
                 return True
         if check_nests and fn(nest):
             return True
     elif fn(nest):
         return True
+    elif check_attributes and hasattr(nest, "__dict__"):
```
mattbarrett98, line 1236: maybe it'd be a little cleaner for this `elif` to come before the `elif fn(nest):` one.

mattbarrett98 (commented):
to add full support for arbitrary objects we would need to alter other nested functions I imagine yes. Though I thought we were leaving this for now, given how rare this case is

xoiga123:
I took a look at other nested functions and they're fine as is unless we want to consider an arbitrary object as a dictionary of its parameters (`nest = object, leaves = [object.param1, object.param2...]`), basically:
```python
from dataclasses import dataclass

@dataclass
class foo:
    a: torch.Tensor = torch.tensor(1)
    b: torch.Tensor = torch.tensor(2)

ivy.nested_argwhere([0, foo()], lambda x: isinstance(x, torch.Tensor), check_attributes=True)
# do we want this [[1]]         (already done in this PR)
# or this [[1, 'a'], [1, 'b']]  (I can implement this)
```

mattbarrett98 (commented):
Given that we've only seen one edge case scenario in kornia which this would help with, I'm not sure whether this is worth adding

xoiga123:
I'm currently using this so I don't miss things while scraping libs, I mean I can check if any more pops up, let's just put this on hold for now

mattbarrett98:
Sure, if we come across some meaningful cases where this would help we will revisit

PR #13912, opened 2023-04-09
added paddle.einsum to paddle backend#13270

### LINKED ISSUE #13270: einsum

VickyThawne:
closes #13270

### DIFF
--- a/ivy/functional/frontends/torch/linalg.py
+++ b/ivy/functional/frontends/torch/linalg.py
@@ -188,15 +188,18 @@ def eig(input, *, out=None):
 def solve(input, other, *, out=None):
     return ivy.solve(input, other, out=out)
 
+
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/torch/linalg.py
```diff
@@ -189,6 +189,38 @@ def solve(input, other, *, out=None):
     return ivy.solve(input, other, out=out)
 
 
+@to_ivy_arrays_and_back
```
Daniel4078, line 192: why there is this change to the torch frontend end?

VickyThawne: that was another task I was working on and commited through the same branch should make it saperated?
then I will make another PR
would it be ok

PR #13935, opened 2023-04-10
Add skipping for `ivy.with_backend`

CatB1t:
- Add functionality for `ivy.with_backend` to skip compiling specific modules, save resources, and allow for type sharing e.g. (`ivy.Array` binds to the same class).


VedPatwardhan (approved):
lgtm! Just one thing I was curious about. Thanks @CatB1t 🙂

ivy/utils/_importlib.py
```diff
@@ -1,11 +1,95 @@
...
 import_cache = {}
 path_hooks = []
 
+ivy_path_abs = Path(sys.modules["ivy"].__file__).parents[1]
+
+# Note that any modules listed as 'to skip' should not depend on the Ivy backend state.
+# If they do, the behavior of ivy.with_backend is undefined and may not function as
+# Expected.
+MODULES_TO_SKIP = ["ivy.compiler"]
+MODULES_TO_EXCLUDE = []
```
Infrared1029, line 18: I think this could have a clearer name, something like `MODULES_TO_KEEP/MODULES_TO_NOT_SKIP/MODULES_TO_IMPORT` since `MODULES_TO_EXCLUDE` has a very similar meaning to `MODULES_TO_SKIP`, saying that, what exactly is the purpose of `MODULES_TO_EXCLUDE`? I think it could be used to include specific modules within packages that are in `MODULES_TO_SKIP`, but is that valid?  (to have a global package but a local module from that package e.g. `MODULES_TO_SKIP = ['ivy.compiler'], MODULES_TO_EXCLUDE = ["ivy.compiler.utils"]`)

ivy/utils/_importlib.py
```diff
@@ -1,11 +1,95 @@
...
+    name_to_path = absolute_name.replace(".", os.path.sep)
+    module_path = os.path.join(ivy_path_abs, name_to_path)
+
+    modules = set()
+    modules.add(absolute_name)
+
+    for root, dirs, files in os.walk(module_path):
+        if root.endswith("__"):
+            continue
+        common = os.path.commonpath([root, module_path])
```
Infrared1029, line 45: as @VedPatwardhan mentioned I believe this will always evaluate to `module_path`

ivy/utils/_importlib.py
```diff
@@ -1,11 +1,95 @@
...
+        full_name = absolute_name + root[len(common) :].replace(os.path.sep, ".")
+        for mod_name in files:
+            if mod_name.startswith("__") or not mod_name.endswith(".py"):
+                continue
+            modules.add(f"{full_name}.{mod_name[:-3]}")
```
Infrared1029, line 50: You can turn this into one line
```python
modules.update([f"{full_name}.{mod_name[:-3]}" for mod_name in files
                         if not mod_name.startswith("__") and mod_name.endswith(".py")])
                 
```
Whether this is more readable or not is up to debate tbh, also a comment that explains the `[:-3]` could be slightly helpful (it is apparent from the context that it excludes the .py but it is always nice to explicitly mention it)

ivy/utils/_importlib.py
```diff
@@ -1,11 +1,95 @@
...
+    module_path = os.path.join(ivy_path_abs, name_to_path)
+
+    modules = set()
+    modules.add(absolute_name)
+
+    for root, dirs, files in os.walk(module_path):
+        if root.endswith("__"):
+            continue
+        common = os.path.commonpath([root, module_path])
+        full_name = absolute_name + root[len(common) :].replace(os.path.sep, ".")
```
Infrared1029, line 46:
```suggestion
        full_name = absolute_name + root[len(common): ].replace(os.path.sep, ".")
```

ivy/utils/_importlib.py
```diff
@@ -1,11 +1,95 @@
...
+        for mod_name in files:
+            if mod_name.startswith("__") or not mod_name.endswith(".py"):
+                continue
+            modules.add(f"{full_name}.{mod_name[:-3]}")
+        for dir_name in dirs:
+            if dir_name.startswith("__"):
+                continue
+            modules.add(f"{full_name}.{dir_name}")
```
Infrared1029, line 54: same here, could be 
```python
modules.update([f"{full_name}.{dir_name}" for dir_name in dirs if not dir_name.startswith("__")])
```
Again, debatable if this is more readable (I personally think it is)

ivy/utils/_importlib.py
CatB1t: This is actually done automatically by the black hook.

ivy/utils/_importlib.py
CatB1t: I agree with the comment part, for the second part I think we can keep it as a for loop for this once as it may be a little overkill (too much characters), will also stick with for loop for the second one to be more consistent!

ivy/utils/_importlib.py
CatB1t: It partially made sense to me back then, but you're right! this will actually have invalid references all around, I'll get the logic for MODULES_TO_EXCLUDE removed, great catch @Infrared1029 !

ivy/utils/_importlib.py
```diff
@@ -1,11 +1,80 @@
...
+        Set of modules to skip
+    """
+    all_modules = set()
+    for module in modules_to_skip:
+        all_modules.update(_get_modules(module))
+    return all_modules
+
+
+_all_modules_to_skip = _get_all_modules_to_skip(MODULES_TO_SKIP)
+
```
Infrared1029, line 78: Since we decided to remove `MODULES_TO_EXCLUDE` I assume this is not needed anymore? since we can just check if the `absolute_name` of the module being imported contains any of the `MODULES_TO_SKIP` packages/modules e.g. if `ivy.compiler.utils` is being imported we would detect that it should be skipped because `ivy.compiler` is in `MODULES_TO_SKIP`, we would have to change the condition in the `_import_module` function tho

ivy/utils/_importlib.py
```diff
@@ -89,6 +158,14 @@ def _import_module(name, package=None):
         parent_name, _, child_name = absolute_name.rpartition(".")
         parent_module = _import_module(parent_name)
         path = parent_module.__spec__.submodule_search_locations
+
+    # Return the one from global Ivy if the module is marked to skip
+    if absolute_name in _all_modules_to_skip:
```
Infrared1029, line 163:
```python
    for module in MODULES_TO_SKIP:
        # the first condition alone is not enough, the second condition ensures that the module is indeed "module_to_skip.*"
        if absolute_name in module and len(absolute_name) >= len(module):
            if path is not None:
                # Set reference to self in parent, if exist
                setattr(parent_module, child_name, sys.modules[absolute_name])
            return sys.modules[absolute_name]
```
I think something like this could work, also, we are assuming the module is already imported in `sys.modules`, is this safe to assume?

Issue #10704, opened 2023-02-20
Add Statistical Functions to Paddle Backend


MahmoudAshraf97:
Add [Statistical Functions](https://www.paddlepaddle.org.cn/documentation/docs/en/api/index\_en.html) to Paddle backend

\_

>Please keep in mind that the proper way to link an issue to this list is to comment "- [ ] #issue\_number" while the issue's title only includes the name of the function you've chosen.

\_

- [x] #10760
- [x] #10807
- [x] #10852
- [x] #10868
- [x] #10873
- [x] #10874
- [x] #10875
- [x] #13078
- [x] cumsum
- [x] einsum

## Experimental

- [x] median
- [x] #10923
- [x] #10932
- [x] #23384
- [x] #14694
- [x] #11058

> The main file paths where these functions are likely to be added are: 
> - `ivy/functional/backends/paddle/statistical.py`
> - `ivy/functional/backends/paddle/experimental/statistical.py`

mosesdaudu001:
#10760

KevinUnadkat:
#10807

paritoshtripathi935:
- [ ] #10852

KevinUnadkat:
#10868

KevinUnadkat:
#10874

KevinUnadkat:
#10875

ShauryaPanwar:
- [] #10877

KevinUnadkat:
> * [] [<std> #10877](https://github.com/unifyai/ivy/issues/10877)

hy @ShauryaPanwar 
  im already work on the std function can utake something else to work on

KevinUnadkat:
#10878

KevinUnadkat:
#10879

KevinUnadkat:
#10880

ShauryaPanwar:
#10881

paritoshtripathi935:
#10923

mosesdaudu001:
#10932

KevinUnadkat:
#11010

KevinUnadkat:
#11014

Ishaan-Chandak:
[ ] #11058

mehular0ra:
Can I take median, nanmean, quantile?

mosesdaudu001:
yes you can handle quantile, I have been struggling with it for a while now. It will be good to see how someone else handles it.

mehular0ra:
[...]

PR #13978, opened 2023-04-11
added the paddle.einsum for paddle backend

VickyThawne:
Added einsum function to the Paddle backend in accordance with issue : #13270.
This completes the subtask from the : Add Statistical Functions to Paddle Backend, #10704

available for review now.
might fail lint test because of the latest commits on the master branch of ivy

### DIFF
--- a/ivy/functional/backends/paddle/statistical.py
+++ b/ivy/functional/backends/paddle/statistical.py
@@ -299,18 +299,11 @@ def einsum(
         paddle.complex64, 
         paddle.complex128
     ]
-    for i, operand in enumerate(operands):
-        operand_dtype = operand.dtype
-        if operand_dtype not in supported_dtypes:
 
-            if paddle.is_tensor(operand):
-                operand = paddle.cast(operand, paddle.float32)
-            else:
-                raise ValueError(
-                    f"Unsupported dtype {operand_dtype} "
-                    "and no fallback dtype specified"
-                )
-                
-            operands = operands[:i] + (operand,) + operands[i + 1 :]
+    casted_operands = [
+        paddle.cast(operand, paddle.float32) 
+        if operand.dtype not in supported_dtypes and paddle.is_tensor(operand) 
+        else operand for operand in operands
+    ]
     # Evaluate the einsum expression with the updated operands
-    return paddle.einsum(equation, *operands)
+    return paddle.einsum(equation, *casted_operands).cast(paddle.Tensor)
--- a/ivy/functional/backends/paddle/statistical.py
+++ b/ivy/functional/backends/paddle/statistical.py
@@ -303,9 +303,7 @@ def einsum(
         operand_dtype = operand.dtype
         if operand_dtype not in supported_dtypes:
 
-            if dtype is not None:
-                operand = paddle.cast(operand, dtype)
-            elif paddle.is_tensor(operand):
+            if paddle.is_tensor(operand):
                 operand = paddle.cast(operand, paddle.float32)
             else:
                 raise ValueError(
@@ -314,7 +312,5 @@ def einsum(
                 )
                 
             operands = operands[:i] + (operand,) + operands[i + 1 :]
-    
- 
-        # Evaluate the einsum expression with the updated operands
+    # Evaluate the einsum expression with the updated operands
     return paddle.einsum(equation, *operands)
--- a/ivy/functional/backends/paddle/statistical.py
+++ b/ivy/functional/backends/paddle/statistical.py
@@ -282,9 +282,49 @@ def cumsum(
             return ivy.flip(x, axis=axis).cast(dtype)
 
 
+@with_unsupported_device_and_dtypes(
+    {"2.4.2 and below": {"cpu": ["uint8", "int8", "float16", "bool"]}},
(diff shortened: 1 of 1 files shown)


ivy/functional/backends/paddle/statistical.py
```diff
@@ -282,9 +282,48 @@ def cumsum(
...
 
 
+@with_unsupported_device_and_dtypes(
+    {"2.4.2 and below": {"cpu": ["uint8", "int8", "float16", "bool"]}},
+    backend_version,
+)
 def einsum(
     equation: str,
     *operands: paddle.Tensor,
+    dtype: Optional[paddle.dtype] = None,
```
MahmoudAshraf97, line 292: remove this argument as it doesn't exist in the standard
```suggestion

```

ivy/functional/backends/paddle/statistical.py
```diff
@@ -282,9 +282,48 @@ def cumsum(
...
+                
+            operands = operands[:i] + (operand,) + operands[i + 1 :]
+    
+    try:
+        # Evaluate the einsum expression with the updated operands
+        return paddle.einsum(equation, *operands)
+    except Exception as e:
+        print(f"Failed to evaluate einsum expression: {e}")
+        # fallback to a ivy implementation here
+        return ivy.einsum(equation, *operands)
```
MahmoudAshraf97, line 329: `ivy.einsum` will use paddle einsum implementation, this is an infinite loop

ivy/functional/backends/paddle/statistical.py
MahmoudAshraf97, line 292: remove this argument as it doesn't exist in the standard
```suggestion

```

ivy/functional/backends/paddle/statistical.py
```diff
@@ -282,9 +282,48 @@ def cumsum(
...
+    # Check if any operand has an unsupported data type, 
+    # and cast it to a supported type if necessary
+    supported_dtypes = [
+        paddle.float32, 
+        paddle.float64, 
+        paddle.complex64, 
+        paddle.complex128
+    ]
+    for i, operand in enumerate(operands):
+        operand_dtype = ivy.as_native_dtype(operand.dtype)
```
MahmoudAshraf97, line 304: `operand.dtype` is already native

ivy/functional/backends/paddle/statistical.py
```diff
@@ -282,9 +282,48 @@ def cumsum(
...
+                    operand = paddle.cast(operand, paddle.float32)
+                else:
+                    raise ValueError(
+                        f"Unsupported dtype {operand_dtype} "
+                        "and no fallback dtype specified"
+                    )
+            except Exception as e:
+                print(f"Failed to cast operand {i} to a supported dtype: {e}")
+                # fallback to a ivy implementation here
+                return ivy.einsum(equation, *operands)
```
MahmoudAshraf97, line 319: 1. the same comment about ivy implementation
2. you need to cast the result back to the original dtype, or atleast replicate the same behaviour as other backends
3. this code can be much simpler without the use of loops or if conditions, use `map` instead

MahmoudAshraf97 (changes requested):

VickyThawne:
thank you for your review @MahmoudAshraf97 , the suggested changes are done. you may review it again.

ivy/functional/backends/paddle/statistical.py
```diff
@@ -282,9 +282,35 @@ def cumsum(
...
+
+            if paddle.is_tensor(operand):
+                operand = paddle.cast(operand, paddle.float32)
+            else:
+                raise ValueError(
+                    f"Unsupported dtype {operand_dtype} "
+                    "and no fallback dtype specified"
+                )
+                
+            operands = operands[:i] + (operand,) + operands[i + 1 :]
```
MahmoudAshraf97, line 314: this is an ineffecient way of casting, you should use `map` or list comprehension for better readability, also you need to cast the operand back to the original dtype which doesn't happen here

MahmoudAshraf97 (changes requested):

ivy/functional/backends/paddle/statistical.py
VickyThawne: where do i need to cast the operand back to the dtype, and another question was why do we need the operand back to the original dtype.

ivy/functional/backends/paddle/statistical.py
```diff
@@ -282,9 +282,28 @@ def cumsum(
...
+        paddle.complex128
+    ]
+
+    casted_operands = [
+        paddle.cast(operand, paddle.float32) 
+        if operand.dtype not in supported_dtypes and paddle.is_tensor(operand) 
+        else operand for operand in operands
+    ]
+    # Evaluate the einsum expression with the updated operands
+    return paddle.einsum(equation, *casted_operands).cast(paddle.Tensor)
```
MahmoudAshraf97, line 309: you need to cast it to a dtype, this will not work
ivy behaviour requires that the output dtype is the same as input dtype

MahmoudAshraf97 (changes requested):

PR #14020, opened 2023-04-11
Added frontent jax random rademacher.

### LINKED ISSUE #14017: rademacher
The Rademacher distribution yields either 1 or -1 with equal probability.

Agecaf:
Closes #14017

The [Rademacher](https://en.wikipedia.org/wiki/Rademacher_distribution) distribution returns either 1 or -1, with equal probability. [Jax's implementation](https://jax.readthedocs.io/en/latest/_autosummary/jax.random.rademacher.html) uses the bernuilli distribution (with `p=0.5`) to implement it. I've implemented it instead with `ivy.randint`.

Since it is possible to easily implement this method from the Ivy API, it might be worth considering adding `ivy.rademacher` into Ivy's API. While it is not implemented in all frameworks, it is available in some others such as [Tensorflow](https://www.tensorflow.org/probability/api_docs/python/tfp/random/rademacher).

There seems to be an issue with the tests at the moment, which others in the discord server have also encountered.

### DIFF
--- a/ivy/functional/frontends/jax/random.py
+++ b/ivy/functional/frontends/jax/random.py
@@ -45,3 +45,11 @@ def _get_seed(key):
 def beta(key, a, b, shape=None, dtype=None):
     seed = _get_seed(key)
     return ivy.beta(a, b, shape=shape, dtype=dtype, seed=seed)
+
+
+# Rademacher distribution, 1 or -1 with equal probability.
+@handle_jax_dtype
+@to_ivy_arrays_and_back
+def rademacher(key, shape=(), dtype=None):
+    seed = _get_seed(key)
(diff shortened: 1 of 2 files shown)


Agecaf:
The tests failing are

In the torch backend,
```ivy.utils.exceptions.IvyException: Cannot convert to PyTorch dtype. uint32 is not supported by PyTorch.```

In the tensorflow backend,
```ivy.utils.exceptions.IvyBackendException: tensorflow: itemsize: expected number as argument to bfloat16 constructor```

@ZoeCD, Fotios in discord has asked me to hold off from the PR until he refactors the tests so that they handle the torch backend, it seems to be an issue with the `jax.random` functions more generally. They noted that they've seen the second error pop up but are unsure what the cause is.

Agecaf:
I will unfortunately not be able to continue following up on this pull request.

PR #14037, opened 2023-04-12
einsum

janetwamaitha:


Infrared1029:
Hi @janetwamaitha! I'm looking into the PR now, there are a few linting errors though, we gotta get those fixed:D! using `pre-commit` as described [here](https://lets-unify.ai/docs/ivy/overview/contributing/setting_up.html#pre-commit) will make things easier!

janetwamaitha:
#13993

janetwamaitha:
There was an initial error when running the test, which was a Name Error.
Name error

Infrared1029:
Hi @janetwamaitha! you are using `np` even tho `np` is not imported in this file, so the errors make sense! we probably do not need to `np` to generate the arrays as there are helper functions for that like `helpers.dtype_and_values`

janetwamaitha:
Okay, let me give a try.

PR #14043, opened 2023-04-12
Handle axis argument of ivy.unique_all and numpy_frontend.unique

AnnaTz:


MahmoudAshraf97 (commented):
Not thoroughly reviewed but make sure all backends behave the same when `axis=None` and accept `None` as input

PR #14047, opened 2023-04-12
 Add Manipulation Functions to Paddle Backend

chinonsoonodi:
The PR aim is to add scatter_flat function to general in paddle backend
issue #14006


MahmoudAshraf97 (changes requested):
please remove any changes other than scatter flat

chinonsoonodi:
> please remove any changes other than scatter flat

Done

MahmoudAshraf97 (changes requested):
Reduction is meant to operate on the updated indices only not the whole array

chinonsoonodi:
> Reduction is meant to operate on the updated indices only not the whole array

Updated

MahmoudAshraf97 (changes requested):
reduction is meant to be between the old values and the new values on elementwise basis, also when concatinating shapes, make sure that they are both tuples or lists. you must use paddle scatter or an equivalent function because the current implementation doesn't work with tensors that have gradients enabled

Issue #14008, opened 2023-04-11
size


AbdullahSabry:
Creating a size funtion for tensorflow frontend

AbdullahSabry:
![Screenshot (95)](https://user-images.githubusercontent.com/87365208/231219385-547195e9-4ed1-4ce8-bfb5-72885a320ab0.png)
Hello! I have a problem running the function 'prod'. Whenever I do so I get an error saying that the 'current_backend' isn't defined. However I have defined it above. So what is the problem?

AbdullahSabry:
Merged with master

PR #14063, opened 2023-04-12
add size with testing and implimentation #14008

AbdullahSabry:
Added size testing and implimentation for #14008 . Is there anything else to do?


AbdullahSabry:
Hello. I seem to be having problems with making the test. the test for 'size' in general_functions.py is identical to the test for 'Size' in 'raw_ops.py'. Is this correct to do?

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py
```diff
@@ -776,6 +776,31 @@ def test_tensorflow_shape_n(
...
+        input_dtypes=input_dtype,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        input=x[0],
+        out_type=output_dtype,
+    )
+
+
```
fspyridakos, line 803: Firstly remove `aliases=["tensorflow.raw_ops.Size"],`, raw_ops aren't exactly aliases since they accept only kwargs so we don't treat them exactly as such (and hence also have tests for them).
Please remove the `# NOQA` and arange all arguments in seperate lines like all other tests
The reason the test fails for this `size` is because tf does not accept anythng other than tf dtypes for `out_type` here. Not sure why there's this difference... I would recommend you just don't generate the `out_type` arg since it will inevitably cause failures and just leave a comment explaining that at the top of the test

fspyridakos (changes requested):
Hey thanks for your contribution! Great job, left a comment about the test, also could you please exclude any files from the .idea folder from this pr. As well as the `key.pem` file. Thanks!

AbdullahSabry:
Sorry for closing the request. I accidentaly pulled into master once again but I hope last commit for pr is good!

fspyridakos:
Looks good now! Many thanks

PR #14064, opened 2023-04-12
Implemented frombuffer frontend and testing

akosua00:
Working on issue frombuffer #13026


nathzi1505:
Hi @akosua00, thanks for contributing.
Would you mind checking at the failing tests (especially the error log) ?

akosua00:
Hi @nathzi1505 I'm having issues with pytest on my laptop. I'll try a few more times but if it doesn't work I will just make the changes in github and use the ivy tests on github.

akosua00:
Hi @nathzi1505 I don't really understand this error: TypeError: a bytes-like object is required, not 'list'. 
I don't understand this since I have not specified that the input for the test should be a list.

Issue #14067, opened 2023-04-12
imag


edith007:
Implementing imag functionality for TensorFlow frontend.

Solves https://github.com/unifyai/ivy/issues/1545

Daniel4078:
This issue is not tracked on the TODO list issue. As the target task is already been done. I will just close this for now.

PR #14068, opened 2023-04-12
Implemented `imag` functionality for TensorFlow frontend.

edith007:
Close https://github.com/unifyai/ivy/issues/14067


Emanuele-n:
Hi again edith007, as you made me notice we are trying to do the same function. I've seen that yours do not pass run_tests_1, this could be because you might need the decorator to specify the dtypes. Feel free to copy mine implementation here #14088.
I hope this is helpful!

Daniel4078:
The task this pull request was dealing with has already be completed by other merged pull request. Thus I will close this for now.

PR #14075, opened 2023-04-13
Refactored tests for jax.random such that they work with all backends

fspyridakos:
Modified the tests to call the PRNG function directly because otherwise we would need to generate a `uint32` PRNGKey. Now the tests can pass for all backends (even if some fail now for backend related issues)

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_random.py
+++ b/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_random.py
@@ -1,185 +1,95 @@
 # global
-from hypothesis import strategies as st
-import ivy
+from hypothesis import strategies as st, given
+import jax
 
 # local
 import ivy_tests.test_ivy.helpers as helpers
-from ivy_tests.test_ivy.helpers import handle_frontend_test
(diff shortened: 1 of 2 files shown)


ivy_tests/test_ivy/test_frontends/test_jax/test_jax_random.py
```diff
@@ -1,203 +1,104 @@
...
-        min_value=0,
-        max_value=2000,
-        min_num_dims=1,
-        max_num_dims=1,
-        min_dim_size=2,
-        max_dim_size=2,
-    ),
+    minval, maxval = minval_maxval
+
+    jax_frontend.config.update("jax_enable_x64", True)
```
CatB1t, line 38: Wouldn't that still be enabled after the test finishes? maybe we should refactor the x64 enable code into a context manager and use it in JAX random tests so that after the test finishes it will clean up and reset the flag to False.

CatB1t (changes requested):
Great work @fspyridakos! just a small point mentioned, happy to hear your thoughts.

ivy-seed:

ivy-seed:

PR #14082, opened 2023-04-13
Handling <unsupported/supported> <dypes/devices> for mixed functions

hello-fri-end:
- [Trello: Fix `recurse` option in `ivy.function_supported_devices_and_dtypes` and `ivy.function_unsupported_devices_and_dtypes`](https://trello.com/c/cjfEWe4b/1668-fix-recurse-option-in-ivyfunctionsupporteddevicesanddtypes-and-ivyfunctionunsupporteddevicesanddtypes)
- [Trello: Investigate why `_get_supported_devices_dtypes` function is not returning the correct dtype set for mixed functions](https://trello.com/c/8IQK5r9V/1649-investigate-why-getsupporteddevicesdtypes-function-is-not-returning-the-correct-dtype-set-for-mixed-functions)
- [ Trello: Handling the <supported/unsupported> <devices/dtypes> for partial mixed functions that call other partial mixed functions in their implementations.](https://trello.com/c/yAAO713a/2208-handling-the-supported-unsupported-devices-dtypes-for-partial-mixed-functions-that-call-other-partial-mixed-functions-in-their-i)
- [Trello: Edge cases of unsupported dtypes for compositional ivy functions](https://trello.com/c/HAUZJ9Wc/1770-edge-cases-of-unsupported-dtypes-for-compositional-ivy-functions)


VedPatwardhan (commented):
let me know once you're done with the changes, as I suppose this review request may not be intended. Thanks @hello-fri-end 🙂

hello-fri-end:
> let me know once you're done with the changes, as I suppose this review request may not be intended. Thanks @hello-fri-end 🙂

oh, all done here actually 😅

VedPatwardhan (changes requested):
1. Probably we could make the implementation a bit cleaner, because the intersection of unsupported dtypes for the backend-specific implementation and the compositional implementation wouldn't necessarily be the correct set of unsupported dtypes. 
2. The unsupported dtypes would totally depend on the nature of the input which determines whether we use the backend-specific implementation or the compositional implementation. 
3. And given that we don't pass the inputs while using `ivy.function_unsupported_dtypes`, we should be returning a tuple containing the tuple of unsupported dtypes for the backend-specific implementation and the tuple of unsupported dtypes for the compositional implementation whenever we are dealing with a function with a **partial mixed implementation**.
4. Of course, we'll also have to update the typehints to represent this `Union` and update the tests where one of the examples tested with `pytest.mark.parametrize` has a partial mixed implementation. 
5. I think this would also help having a cleaner implementation as you would only need to repeat the same process for the primary function and the compositional one through a private helper rather than repeating the same logic twice.

Feel free to message if there's anything else I should clarify about, thanks @hello-fri-end 🙂

VedPatwardhan (commented):
> > 1. could make the implementation a bit cleaner, because the intersection of unsupported dtypes for the backend-specific implementation and the compositional implementation wouldn't necessarily be the correct set of unsupported dtypes.
> 
> Thanks for the detailed review @VedPatwardhan 🚀 Completely agree with cleaning up the implementation part - will add a private helper function that all of these functions can use. I also agree that returning a tuple containing the tuple of unsupported dtypes for the backend-specific implementation and the tuple of unsupported dtypes for the compositional implementation makes much more sense, thanks for pointing this out 👍 . In this case though, how should I update the `handle_test` decorator which expects that this function returns only one tuple containing the datatypes that are supported by the function to be tested ?

That's a great point! I think it would be a bit tricky in the tests (I'm also pinging @CatB1t to the chat to keep him in the loop).
But ideally we'd just check if the output of `ivy.function_[un]supported_[devices_and]_dtypes` is a tuple. Then get a union of the unsupported dtypes of the 2. Finally after generating the data (function arguments) with those dtypes, we'd know whether we'd be deferring to the compositional implementation or the primary implementation. So based on the implementation we'd use, we'll find if the dtype generated is unsupported and just skip the test if it is unsupported for that particular dtype. Of course, this would mean we waste some examples, but I'm not sure if there's another way to test partial mixed functions without a larger redesign of the testing pipeline. And given that there's only a few partial mixed functions in the framework, I'm not sure if it's worth redesigning the testing pipeline altogether to incorporate these functions.
Happy to know what you guys think, thanks 🙂

hello-fri-end:
> > > 1. could make the implementation a bit cleaner, because the intersection of unsupported dtypes for the backend-specific implementation and the compositional implementation wouldn't necessarily be the correct set of unsupported dtypes.
> > 
> > 
> > Thanks for the detailed review @VedPatwardhan rocket Completely agree with cleaning up the implementation part - will add a private helper function that all of these functions can use. I also agree that returning a tuple containing the tuple of unsupported dtypes for the backend-specific implementation and the tuple of unsupported dtypes for the compositional implementation makes much more sense, thanks for pointing this out +1 . In this case though, how should I update the `handle_test` decorator which expects that this function returns only one tuple containing the datatypes that are supported by the function to be tested ?
> 
> That's a great point! I think it would be a bit tricky in the tests (I'm also pinging @CatB1t to the chat to keep him in the loop). But ideally we'd just check if the output of `ivy.function_[un]supported_[devices_and]_dtypes` is a tuple. Then get a union of the unsupported dtypes of the 2. Finally after generating the data (function arguments) with those dtypes, we'd know whether we'd be deferring to the compositional implementation or the primary implementation. So based on the implementation we'd use, we'll find if the dtype generated is unsupported and just skip the test if it is unsupported for that particular dtype. Of course, this would mean we waste some examples, but I'm not sure if there's another way to test partial mixed functions without a larger redesign of the testing pipeline. And given that there's only a few partial mixed functions in the framework, I'm not sure if it's worth redesigning the testing pipeline altogether to incorporate these functions. Happy to know what you guys think, thanks slightly_smiling_face

Another option is to write the tests for the compositional and primary implementations separately. In this case, we would need to add an argument to the `handle_test` decorator say `mixed_function` which can take two values `compositional` or `primary` depending on which implementation we are testing. And then we would consider only one of the tuples returned by the `ivy.function_supported_devices_and_datatypes` based on the value of this argument. In this case, we would need to update the tests of all the partial mixed functions we have implemented ( i guess this is not more than 3-4). Again, not sure if this is the best solution here, would love to hear your thoughts @VedPatwardhan @CatB1t

hello-fri-end:
So I spent some time thinking about the possible solutions we discussed. Here's a summary:

1. Run the primary implementation first, if it fails due to dtype error, run the compositional implementation. 
 Problem: challenging to detect dtype error e.g LayerNorm fails with the following error message:
 ` ivy.utils.exceptions.IvyBackendException: torch: layer_norm: "LayerNormKernelImpl" not implemented for 'Half' `
_Not possible to implement._


2. Writing multiple strategies
Problem: If multiple backends have mixed partial implementations then we will have to write a large number of strategies to cover the entire testing space because the conditions for switching between the implementations for different backends could have some overlap.
_Not possible to implement._

3. Generate the data, check which implementation will be used, and discard the example if the function doesn't support the data type of the example.
Problem: Unbalanced testing examples i.e it's possible that, for example, we could end up discarding all the examples hypothesis generates for the compositional implementation because the data type is not supported. Plus, this approach will be very slow.
_Possible to implement._

4. Write the tests for backends that have mixed partial implementations separately. And write separate tests for primary and compositional implementations.
Problem: We will end up wasting the examples we generate for the backends that are not being tested in each test.
_Possible to implement._

So, it looks like we have to choose between approach (3) or (4). I would personally prefer the 4th approach because it will cover the entire testing space and require the least amount of refactoring of testing helpers. Would love to know your thoughts @VedPatwardhan @CatB1t  and then I'll start implementing one of these.

VedPatwardhan (commented):
> Hey @hello-fri-end, 
> 1. We shouldn't be making use of `3` of course.
> 2. I think `2` and `4` kinda represent the same thing, where we could just have 2 different sets of strategies for testing. 
> 3. Now the question is whether we should be having it in the same test (which would be compact but relatively complex) or in different tests (which would be simpler to implement but not compact).
> 4. I would personally prefer the former as its more compact, also taking into account that we expect to have quite a few partial mixed implementations in the framework for efficiency and also that the condition in the lambda function for switching isn't that complex atleast for the functions we have currently (e.g. `linear`, `interpolate`, etc.)
> 6. For implementing 2 sets of data generation strategies in the same test, we could just have a variable in our globals called `test_partial_mixed` or something. When it's `True`, we'll find the appropriate unsupported dtypes and generate the data based on the conditions required for deferring to the backend-specific implementation through some private hypothesis strategy we've written for that test 🙂
> (of course given that this is testing-related, let's see what @CatB1t thinks about it)

I agree that it would be more compact to handle this in one test as opposed to multiple tests. Although, we would need 2 strategies for each backend that has a mixed partial implementation and 1 strategy for all the backends that only have a primary or compositional implementation in that case.

Let me explain this with the help of the following Diagram. 
Assume backends B1 and B2 have mixed partial implementations and backends B3 and B4  only have compositional implementations. 
The blue line separates the testing examples for compositional and primary implementations of the backend B1. Similarily, the green line seprates the testing space for backend B2.
Bi-compos : set of all examples for testing backend Bi's compositional implementation. 
Bi-Primary : set of all examples for testing backend Bi's primary implementation. 
Of course, Bi-compos and Bi-primary don't have an intersection for the same i.
The set of examples for testing backends B3 and B4 is the union of all the sets in the diagram.
latest-screenshot

Case 1: Each strategy tests all the backends
In this case, we would need 4 strategies for the test, one for each Bi<compos/primary> intersection Bj<compos/primary>.

Case 2: Each strategy only tests a set of backends.
 One strategy for testing backends B3, B4. Two sets each testing backends B1 and B2.

P.S: Sorry, I accidentally edited your comment instead of adding a new one @VedPatwardhan :sweat_smile:

hello-fri-end:
Hi guys @VedPatwardhan @CatB1t! Here's a summary of changes I have made till now:

- Added private helper functions in` ivy/data, ivy/general, ivy/device` to handle <devices/dtypes> for mixed functions correctly. For mixed partial functions, the relevant dtypes/devices will be returned for both the compositional implementation as well as the primary implementation as a tuple.
- Fixed the recurse option in function_<supported/unsupported>_devices_and_dtypes by adding a `recurse` option in `_get_devices_and_dtypes` which by default is `False` so that `function.supported_devices` & `function_supported_dtypes` do not recurse into the input fn and recurse can be handled by the `function_<supported/unsupported>_devices_and_dtypes` itself.
- Updated `_get_supported_devices_dtypes` to correctly handle the tuple of dictionaries returned by `ivy.function_supported_devices_and_dtypes` in case of mixed functions.

For mixed partial functions, I've made the following changes:
- I added a parameter called `mixed_function_index` to the `get_dtypes` helper function which indicates whether to select the dtypes of the compositional implementation or the primary implementation. If it's 0, we select the types of the compositional implementation and if it's 1 we return the dtypes of the primary implementation.
- I updated all the functions that internally call `get_dtypes` to also have this parameter. 

The idea behind this is that we can use hypothesis to generate this parameter for us and then based on the value of this parameter and the current backend, we generate the example for testing the function. 
e.g I have updated the test of ivy.linear in the last commit:
https://github.com/unifyai/ivy/blob/98634861082dbfe61914b6bd77dbe7e42803d269/ivy_tests/test_ivy/test_functional/test_nn/test_layers.py#L19-L23
and once we have the correct set of dtypes, we can generate the data based on the backend and the value of this parameter:
https://github.com/unifyai/ivy/blob/98634861082dbfe61914b6bd77dbe7e42803d269/ivy_tests/test_ivy/test_functional/test_nn/test_layers.py#L36-L37

Wanted to know your thoughts on this. If this is okay, I'll start updating the tests of all the other mixed functions and write a guide in the docs explaining all this.

VedPatwardhan (changes requested):
Hey @hello-fri-end, I just wanted to ask if you've explored a way to call the same function twice recursively if it's a mixed function (once for the compositional one and once for the primary one) rather than adding additional helpers which seem to kinda perform the same logic? Also can we also return a `dict` instead of a tuple, would be useful for people to easily know which one's compositional and which one's primary
I mean if we can just something like
```
def function_unsupported_dtypes(fn, ...):
    if it's a partial mixed function:
        return {
            'compositional': function_unsupported_dtypes(compos), 
            'primary': function_unsupported_dtypes(primary)
        }
    else:
        The usual logic
```
it would be really useful to reduce duplication. Would be happy to learn more about the challenges faced while implementing this (as I think you may have thought of this already), but that's a cleaner approach in my opinion. The changes made to the testing pipeline look good to me, of course you'll have to change the index logic to a boolean as @CatB1t suggested yesterday. 
Thanks 🙂

hello-fri-end:
> Hey @hello-fri-end, I just wanted to ask if you've explored a way to call the same function twice recursively if it's a mixed function (once for the compositional one and once for the primary one) rather than adding additional helpers which seem to kinda perform the same logic? Also can we also return a `dict` instead of a tuple, would be useful for people to easily know which one's compositional and which one's primary I mean if we can just something like
> 
> ```
> def function_unsupported_dtypes(fn, ...):
>     if it's a partial mixed function:
>         return {
>             'compositional': function_unsupported_dtypes(compos), 
>             'primary': function_unsupported_dtypes(primary)
>         }
>     else:
>         The usual logic
> ```
> 
> it would be really useful to reduce duplication. Would be happy to learn more about the challenges faced while implementing this (as I think you may have thought of this already), but that's a cleaner approach in my opinion. The changes made to the testing pipeline look good to me, of course you'll have to change the index logic to a boolean as @CatB1t suggested yesterday. Thanks slightly_smiling_face

Hey @VedPatwardhan, initially I tried almost the exact same thing that you have described above. The problem is with this approach is that it gets stuck into an infinite recursion loop because when the function is recursively called with the primary function as the input ( `function_unsupported_dtypes(primary)`) rhe condition for checking if the input function is a partial mixed function again evaluates to `True`. But thinking about it again, I don't think there's any need to recurse for the primary function as the `_get_dtypes()` helper is sufficient to get the supported/unsupported datatypes, so I think the following approach should work:
```
 def function_unsupported_dtypes(fn, ...):
     if it's a partial mixed function:
         return {
             'compositional': function_unsupported_dtypes(compos), 
             'primary': `_get_dtypes(primary)`
         }
     else:
         The usual logic
 ```
 
 Will experiment with this and update the functions tomorrow. I also wanted to ask what the ideal behavior of this function should in the following case:
 
Assume that `_fn `is a mixed partial function:
 ```python
fn_module = 'ivy.functional`
_tmp_mod = importlib.import_module(fn_module)
_fn = _tmp_mod.__dict__[fn_name]
ivy.function_supported_dtypes(_fn)
 ```
 In this case, we are calling `function_supported_dtypes(_fn)` with the compositional function as the input but the backend has a mixed partial implementation. Should we just return the supported dtypes for the compositional implementation in this case @VedPatwardhan ?

**Comment by hello-fri-end** [comment]
> > Hey @hello-fri-end, I just wanted to ask if you've explored a way to call the same function twice recursively if it's a mixed function (once for the compositional one and once for the primary one) rather than adding additional helpers which seem to kinda perform the same logic? Also can we also return a `dict` instead of a tuple, would be useful for people to easily know which one's compositional and which one's primary I mean if we can just something like
> > ```
> > def function_unsupported_dtypes(fn, ...):
> >     if it's a partial mixed function:
> >         return {
> >             'compositional': function_unsupported_dtypes(compos), 
> >             'primary': function_unsupported_dtypes(primary)
> >         }
> >     else:
> >         The usual logic
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > it would be really useful to reduce duplication. Would be happy to learn more about the challenges faced while implementing this (as I think you may have thought of this already), but that's a cleaner approach in my opinion. The changes made to the testing pipeline look good to me, of course you'll have to change the index logic to a boolean as @CatB1t suggested yesterday. Thanks slightly_smiling_face
> 
> Hey @VedPatwardhan, initially I tried almost the exact same thing that you have described above. The problem is with this approach is that it gets stuck into an infinite recursion loop because when the function is recursively called with the primary function as the input ( `function_unsupported_dtypes(primary)`) rhe condition for checking if the input function is a partial mixed function again evaluates to `True`. But thinking about it again, I don't think there's any need to recurse for the primary function as the `_get_dtypes()` helper is sufficient to get the supported/unsupported datatypes, so I think the following approach should work:
> 
> ```
>  def function_unsupported_dtypes(fn, ...):
>      if it's a partial mixed function:
>          return {
>              'compositional': function_unsupported_dtypes(compos), 
>              'primary': `_get_dtypes(primary)`
>          }
>      else:
>          The usual logic
> ```
> 
> Will experiment with this and update the functions tomorrow. I also wanted to ask what the ideal behavior of this function should in the following case:

Hey @VedPatwardhan, so i tested this and it seems to work as expected. Will update all the functions with this logic,

> 
> Assume that `_fn `is a mixed partial function:
> 
> ```python
> fn_module = 'ivy.functional`
> _tmp_mod = importlib.import_module(fn_module)
> _fn = _tmp_mod.__dict__[fn_name]
> ivy.function_supported_dtypes(_fn)
> ```
> 
> In this case, we are calling `function_supported_dtypes(_fn)` with the compositional function as the input but the backend has a mixed partial implementation. Should we just return the supported dtypes for the compositional implementation in this case @VedPatwardhan ?

What should we do about this?

**Review by VedPatwardhan — COMMENTED** [review]
> > Hey @hello-fri-end, I just wanted to ask if you've explored a way to call the same function twice recursively if it's a mixed function (once for the compositional one and once for the primary one) rather than adding additional helpers which seem to kinda perform the same logic? Also can we also return a `dict` instead of a tuple, would be useful for people to easily know which one's compositional and which one's primary I mean if we can just something like
> > ```
> > def function_unsupported_dtypes(fn, ...):
> >     if it's a partial mixed function:
> >         return {
> >             'compositional': function_unsupported_dtypes(compos), 
> >             'primary': function_unsupported_dtypes(primary)
> >         }
> >     else:
> >         The usual logic
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > it would be really useful to reduce duplication. Would be happy to learn more about the challenges faced while implementing this (as I think you may have thought of this already), but that's a cleaner approach in my opinion. The changes made to the testing pipeline look good to me, of course you'll have to change the index logic to a boolean as @CatB1t suggested yesterday. Thanks slightly_smiling_face
> 
> Hey @VedPatwardhan, initially I tried almost the exact same thing that you have described above. The problem is with this approach is that it gets stuck into an infinite recursion loop because when the function is recursively called with the primary function as the input ( `function_unsupported_dtypes(primary)`) rhe condition for checking if the input function is a partial mixed function again evaluates to `True`. But thinking about it again, I don't think there's any need to recurse for the primary function as the `_get_dtypes()` helper is sufficient to get the supported/unsupported datatypes, so I think the following approach should work:
> 
> ```
>  def function_unsupported_dtypes(fn, ...):
>      if it's a partial mixed function:
>          return {
>              'compositional': function_unsupported_dtypes(compos), 
>              'primary': `_get_dtypes(primary)`
>          }
>      else:
>          The usual logic
> ```
> 
> Will experiment with this and update the functions tomorrow. I also wanted to ask what the ideal behavior of this function should in the following case:
> 
> Assume that `_fn `is a mixed partial function:
> 
> ```python
> fn_module = 'ivy.functional.ivy`
> _tmp_mod = importlib.import_module(fn_module)
> _fn = _tmp_mod.__dict__[fn_name]
> ivy.function_supported_dtypes(_fn)
> ```
> 
> In this case, we are calling `function_supported_dtypes(_fn)` with the compositional function as the input but the backend has a mixed partial implementation. Should we just return the supported dtypes for the compositional implementation in this case @VedPatwardhan ?

Apologies for the delay! I couldn't fully follow, wouldn't we have both those implementations in the same input function, given that we were storing it in the `compos` attribute? Ideally I think we should be computing the supported/unsupported for both the implementations when it's a partial mixed function. Thanks for looking into this @hello-fri-end 🙂

**Comment by hello-fri-end** [comment]
> Apologies for the delay! I couldn't fully follow, wouldn't we have both those implementations in the same input function, given that we were storing it in the `compos` attribute? Ideally I think we should be computing the supported/unsupported for both the implementations when it's a partial mixed function. Thanks for looking into this @hello-fri-end slightly_smiling_face

Yes, we do have access to both implementations with the primary function, that's not what I meant. Actually, the way the `handle_test` decorator passes the input to `function.supported_devices_and_dtypes` is using the way I described above i.e by importing `ivy.functional` module and then retrieving the function through the dict which always returns the compositional implementation. For now, I have changed it to always pass `ivy.__dict__[fn.__name__]` incase of mixed functions, which will automatically pass the compositional implementation or the primary implementation based on the backend. And in case the backend has a mixed partial implementation and we pass the compositional implementation ( like in the decorator), we'll only return the supported/unsupporterted dtypes for the compositional implementation. Hope that makes sense :)

**Comment by VedPatwardhan** [comment]
> > Apologies for the delay! I couldn't fully follow, wouldn't we have both those implementations in the same input function, given that we were storing it in the `compos` attribute? Ideally I think we should be computing the supported/unsupported for both the implementations when it's a partial mixed function. Thanks for looking into this @hello-fri-end slightly_smiling_face
> 
> Yes, we do have access to both implementations with the primary function, that's not what I meant. Actually, the way the `handle_test` decorator passes the input to `function.supported_devices_and_dtypes` is using the way I described above i.e by importing `ivy.functional` module and then retrieving the function through the dict which always returns the compositional implementation. For now, I have changed it to always pass `ivy.__dict__[fn.__name__]` incase of mixed functions, which will automatically pass the compositional implementation or the primary implementation based on the backend. And in case the backend has a mixed partial implementation and we pass the compositional implementation ( like in the decorator), we'll only return the supported/unsupporterted dtypes for the compositional implementation. Hope that makes sense. Please let me know if should update it to return both dtypes in that case too.

But shouldn't the dict always return the primary implementation instead of the compositional one when it exists? Probably something to fix in the tests, any ideas @CatB1t? Thanks 🙂

**Comment by hello-fri-end** [comment]
> > > Apologies for the delay! I couldn't fully follow, wouldn't we have both those implementations in the same input function, given that we were storing it in the `compos` attribute? Ideally I think we should be computing the supported/unsupported for both the implementations when it's a partial mixed function. Thanks for looking into this @hello-fri-end slightly_smiling_face
> > 
> > 
> > Yes, we do have access to both implementations with the primary function, that's not what I meant. Actually, the way the `handle_test` decorator passes the input to `function.supported_devices_and_dtypes` is using the way I described above i.e by importing `ivy.functional` module and then retrieving the function through the dict which always returns the compositional implementation. For now, I have changed it to always pass `ivy.__dict__[fn.__name__]` incase of mixed functions, which will automatically pass the compositional implementation or the primary implementation based on the backend. And in case the backend has a mixed partial implementation and we pass the compositional implementation ( like in the decorator), we'll only return the supported/unsupporterted dtypes for the compositional implementation. Hope that makes sense. Please let me know if should update it to return both dtypes in that case too.
> 
> But shouldn't the dict always return the primary implementation instead of the compositional one when it exists? Probably something to fix in the tests, any ideas @CatB1t? Thanks slightly_smiling_face

Here's a minimal example with torch backend for `layer_norm`

```python
test = importlib.import_module('ivy.functional')
test.__dict__[fn_name].__module__
'ivy.functional.ivy.norms'
```

VedPatwardhan:
> 
> Here's a minimal example with torch backend for `layer_norm`
> 
> ```python
> test = importlib.import_module('ivy.functional')
> test.__dict__[fn_name].__module__
> 'ivy.functional.ivy.norms'
> ```

Oh okay got it 👍
Can we hop on another call tomorrow about this? Thanks @CatB1t @hello-fri-end 🙂

hello-fri-end:
ivy-gardener

hello-fri-end:
ivy-gardener

ivy-leaves:

Thank you :hugs:

ivy/functional/ivy/data_type.py
```diff
@@ -1617,18 +1629,29 @@ def function_supported_dtypes(fn: Callable, recurse: bool = True) -> Tuple:
...
         ),
     )
-    supported_dtypes = set(_get_dtypes(fn, complement=False))
-    if recurse:
-        supported_dtypes = _nested_get(
-            fn, supported_dtypes, set.intersection, function_supported_dtypes
-        )
-
-    return tuple(supported_dtypes)
+    if hasattr(fn, "partial_mixed_handler"):
```
vedpatwardhan, line 1632: Would be great to update the return signature of this function in it's docstring, mentioning that it will be a dict for partial mixed functions

ivy/functional/ivy/data_type.py
```diff
@@ -1657,13 +1680,23 @@ def function_unsupported_dtypes(fn: Callable, recurse: bool = True) -> Tuple:
             "in a particular backend"
         ),
     )
-    unsupported_dtypes = set(_get_dtypes(fn, complement=True))
-    if recurse:
-        unsupported_dtypes = _nested_get(
-            fn, unsupported_dtypes, set.union, function_unsupported_dtypes
-        )
+    if hasattr(fn, "partial_mixed_handler"):
```
vedpatwardhan, line 1683: docstring update here too

ivy/functional/ivy/device.py
```diff
@@ -1174,19 +1176,30 @@ def function_supported_devices(fn: Callable, recurse: bool = True) -> Tuple:
             "exist in a particular backend"
         ),
     )
-    supported_devices = set(_get_devices(fn, complement=False))
-
-    if recurse:
-        supported_devices = ivy.functional.data_type._nested_get(
-            fn, supported_devices, set.intersection, function_supported_devices
-        )
+    if hasattr(fn, "partial_mixed_handler"):
```
vedpatwardhan, line 1179: docstring

ivy/functional/ivy/device.py
```diff
@@ -1215,14 +1228,22 @@ def function_unsupported_devices(fn: Callable, recurse: bool = True) -> Tuple:
...
     )
-    unsupported_devices = set(_get_devices(fn, complement=True))
-
-    if recurse:
-        unsupported_devices = ivy.functional.data_type._nested_get(
-            fn, unsupported_devices, set.union, function_unsupported_devices
-        )
-
-    return tuple(unsupported_devices)
+    if hasattr(fn, "partial_mixed_handler"):
```
vedpatwardhan, line 1231: docstring

ivy/functional/ivy/general.py
```diff
@@ -3707,18 +3711,26 @@ def function_supported_devices_and_dtypes(fn: Callable, recurse: bool = True) ->
...
-        supported_devices_dtype = ivy.functional.data_type._nested_get(
-            fn,
-            _all_dnd_combinations(),
-            _dnd_dict_intersection,
-            function_supported_devices_and_dtypes,
-            wrapper=lambda x: x,
-        )
 
-    return supported_devices_dtype
+    if hasattr(fn, "partial_mixed_handler"):
```
vedpatwardhan, line 3715: docstring

ivy/functional/ivy/general.py
```diff
@@ -3748,18 +3760,24 @@ def function_unsupported_devices_and_dtypes(fn: Callable, recurse: bool = True)
...
-        unsupported_devices_dtype = ivy.functional.data_type._nested_get(
-            fn,
-            {},
-            _dnd_dict_union,
-            function_unsupported_devices_and_dtypes,
-            wrapper=lambda x: x,
-        )
-
-    return unsupported_devices_dtype
+    if hasattr(fn, "partial_mixed_handler"):
```
vedpatwardhan, line 3763: docstring

vedpatwardhan (approved):
Hey @hello-fri-end, just requested a few minor modifications, rest looks good to go. Let's see what @CatB1t thinks. Thanks both 🙂

ivy/functional/ivy/data_type.py
hello-fri-end: Updated.

hello-fri-end:
Hey Ved, I added a fix for  [Trello: Edge cases of unsupported dtypes for compositional ivy functions](https://trello.com/c/HAUZJ9Wc/1770-edge-cases-of-unsupported-dtypes-for-compositional-ivy-functions) in this commit here https://github.com/unifyai/ivy/pull/14082/commits/e77f30d0a013e7159fc30eed3da0f9de37e083c7 because the `recurse` option doesn't work correctly on master. Basically, to flag unsupported dtypes in the compositional implementation, you can simply add the `unsupported_dtypes` attribute to the function with a tuple containing the data-types and these will be added to the set of unsupported dtypes for that function. Happy to know your thoughts on this fix.

vedpatwardhan (commented):
> Hey Ved, I added a fix for [Trello: Edge cases of unsupported dtypes for compositional ivy functions](https://trello.com/c/HAUZJ9Wc/1770-edge-cases-of-unsupported-dtypes-for-compositional-ivy-functions) in this commit [e77f30d](https://github.com/unifyai/ivy/commit/e77f30d0a013e7159fc30eed3da0f9de37e083c7) here because the `recurse` option doesn't work correctly on master. Basically, to flag unsupported dtypes in the compositional implementation, you can simply add the `unsupported_dtypes` attribute to the function with a tuple containing the data-types and these will be added to the set of unsupported dtypes for that function. Happy to know your thoughts on this fix.

Hey @hello-fri-end, thanks for looking into this! Not sure why we'd need to define the unsupported dtypes for this compositional implementation though. Given that we're using `ivy.frombuffer` which has the `with_unsupported_dtypes` decorator with a `tensorflow` backend, shouldn't the unsupported_dtype function infer the `dtype` for `as_strided`? 
I think the only exception to this should be the `einops` functions which don't have backend implementations and make use of the `einops` API which don't have any unsupported dtype specified 😅
Happy to know your thoughts 🙂

hello-fri-end:
> > Hey Ved, I added a fix for [Trello: Edge cases of unsupported dtypes for compositional ivy functions](https://trello.com/c/HAUZJ9Wc/1770-edge-cases-of-unsupported-dtypes-for-compositional-ivy-functions) in this commit [e77f30d](https://github.com/unifyai/ivy/commit/e77f30d0a013e7159fc30eed3da0f9de37e083c7) here because the `recurse` option doesn't work correctly on master. Basically, to flag unsupported dtypes in the compositional implementation, you can simply add the `unsupported_dtypes` attribute to the function with a tuple containing the data-types and these will be added to the set of unsupported dtypes for that function. Happy to know your thoughts on this fix.
> 
> Hey @hello-fri-end, thanks for looking into this! Not sure why we'd need to define the unsupported dtypes for this compositional implementation though. Given that we're using `ivy.frombuffer` which has the `with_unsupported_dtypes` decorator with a `tensorflow` backend, shouldn't the unsupported_dtype function infer the `dtype` for `as_strided`? I think the only exception to this should be the `einops` functions which don't have backend implementations and make use of the `einops` API which don't have any unsupported dtype specified sweat_smile Happy to know your thoughts slightly_smiling_face

Actually, the problem in the compositional implementation of `ivy.as_strided` is that it uses the python  `memoryview` function which doesn't support `bfloat16` so there's no way of inheriting this from the primary functions called and we need a way of explicitly specifying it in the compositional implementation (first bullet point in the Trello card)

vedpatwardhan:
> > > Hey Ved, I added a fix for [Trello: Edge cases of unsupported dtypes for compositional ivy functions](https://trello.com/c/HAUZJ9Wc/1770-edge-cases-of-unsupported-dtypes-for-compositional-ivy-functions) in this commit [e77f30d](https://github.com/unifyai/ivy/commit/e77f30d0a013e7159fc30eed3da0f9de37e083c7) here because the `recurse` option doesn't work correctly on master. Basically, to flag unsupported dtypes in the compositional implementation, you can simply add the `unsupported_dtypes` attribute to the function with a tuple containing the data-types and these will be added to the set of unsupported dtypes for that function. Happy to know your thoughts on this fix.
> > 
> > 
> > Hey @hello-fri-end, thanks for looking into this! Not sure why we'd need to define the unsupported dtypes for this compositional implementation though. Given that we're using `ivy.frombuffer` which has the `with_unsupported_dtypes` decorator with a `tensorflow` backend, shouldn't the unsupported_dtype function infer the `dtype` for `as_strided`? I think the only exception to this should be the `einops` functions which don't have backend implementations and make use of the `einops` API which don't have any unsupported dtype specified sweat_smile Happy to know your thoughts slightly_smiling_face
> 
> Actually, the problem in the compositional implementation of `ivy.as_strided` is that it uses the python `memoryview` function which doesn't support `bfloat16` so there's no way of inheriting this from the primary functions called and we need a way of explicitly specifying it in the compositional implementation (first bullet point in the Trello card)

Got it! Seems like I misunderstood the task, thanks for elaborating 😅

vedpatwardhan:
In that case, we could actually generalize the check we're making for `einops` functions [explicitly](https://github.com/unifyai/ivy/blob/cf461eb9c3af2a819fb90e02aedeceda1cde8194/ivy/functional/ivy/data_type.py#L180), cuz both functions (`einops_reduce` and `as_strided`) have the same attribute being used, happy to know what do you think @hello-fri-end 🙂

hello-fri-end:
> einops_reduce

Yeah, I think that makes sense. We can just have the `has_unsupported_dtypes_attr = hasattr(fn, "unsupported_dtypes")` check for both.

CatB1t (approved):
Lgtm! Let's just make sure no failures were introduced for non mixed partial functions, then feel free to merge!

hello-fri-end:
ivy-gardener

hello-fri-end:
The errors on the CI seem to be unrelated to this PR. Merging this now - thanks for the reviews @vedpatwardhan @CatB1t :rocket:

PR #14084, opened 2023-04-13
Adding support_native_out and handle_out_decorator to mixed_functions

hello-fri-end:
`handle_out_argument `and `support_native_out` attributes should be added to all mixed functions that have an out argument.

https://lets-unify.ai/docs/ivy/overview/deep_dive/inplace_updates.html#inplace-updates


ivy/functional/ivy/elementwise.py
```diff
@@ -5450,9 +5450,9 @@ def rad2deg(
     return ivy.current_backend(x).rad2deg(x, out=out)
 
 
-@inputs_to_ivy_arrays
 @handle_array_function
 @handle_out_argument
```
VedPatwardhan, line 5454: I think `inputs_to_ivy_arrays` should be between `handle_array_function` and `handle_out_argument` based on the decorator sequence in the [`func_wrapper`](https://github.com/unifyai/ivy/blob/c63f7759b6ff6d197d4a5f21272b669d7d19289d/ivy/func_wrapper.py#L14)

VedPatwardhan (changes requested):
I don't think we need to add the `handle_out_argument` decorator to the compositional implementations in mixed functions, as the compositional implementation can just be thought of as any other compositional implementation where all the `ivy` functions called in the implementation support the `out` argument anyway.
Similarly, we wouldn't need the `support_native_out` attribute either, because whether the `out` argument is supported natively or not would just be determined by whether the `ivy` function called in the implementation supports the `out` argument natively for the currently set backend.
Feel free to share an example where it fails without having the `handle_out_argument`, probably it could be some bug somewhere else. 
Of course, we should be getting the `inputs_to_ivy_arrays` decorator change merged as the order of the decorator should be fixed anyway.
Thanks @hello-fri-end 🙂

hello-fri-end:
> I don't think we need to add the `handle_out_argument` decorator to the compositional implementations in mixed functions, as the compositional implementation can just be thought of as any other compositional implementation where all the `ivy` functions called in the implementation support the `out` argument anyway. Similarly, we wouldn't need the `support_native_out` attribute either, because whether the `out` argument is supported natively or not would just be determined by whether the `ivy` function called in the implementation supports the `out` argument natively for the currently set backend. Feel free to share an example where it fails without having the `handle_out_argument`, probably it could be some bug somewhere else. Of course, we should be getting the `inputs_to_ivy_arrays` decorator change merged as the order of the decorator should be fixed anyway. Thanks @hello-fri-end slightly_smiling_face

This is the explanation given in the Inplace Updates section of the deep dive about why adding the `handle_out_argument` decorator and `support_native_out` attribute is necessary for mixed function :thinking: 

"Unlike compositional arguments, where the handle_out_argument decorator is not included, this decorator should be included for mixed functions. This decorator is needed in order to ensure the out argument is handled correctly when the backend does include a backend-specific implementation, which itself may or may not handle the out argument explicitly. In such cases, the mixed function behaves like a primary function. If the backend-specific implementation does not handle the out argument explicitly (there is no attribute support_native_out specified on the backend function), then it will need to be handled [in the decorator](https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L287).

However, the inclusion of this decorator means that in cases where the mixed function is called compositionally (there is no backend implementation), then the out argument will also be handled [in the decorator](https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L287), this time because of the lack of the support_native_out attribute found on the compositional implementation. But this is not ideal. All compositional implementations are fully capable of handling the out argument explicitly, and so handling it [in the decorator](https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L287) will likely be less efficient, and prevent us from leveraging backend-specific in-place optimizations where they might exist when calling the individual Ivy functions of the compositional implementation.

Therefore, we always add the support_native_out attribute to mixed functions, to ensure that the out argument is always handled directly by the compositional implementation, rather than being handled [in the decorator](https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L287)."

ivy/functional/ivy/elementwise.py
hello-fri-end: Oops, my bad. Will update this, thanks for pointing it out! :)

VedPatwardhan:
> > I don't think we need to add the `handle_out_argument` decorator to the compositional implementations in mixed functions, as the compositional implementation can just be thought of as any other compositional implementation where all the `ivy` functions called in the implementation support the `out` argument anyway. Similarly, we wouldn't need the `support_native_out` attribute either, because whether the `out` argument is supported natively or not would just be determined by whether the `ivy` function called in the implementation supports the `out` argument natively for the currently set backend. Feel free to share an example where it fails without having the `handle_out_argument`, probably it could be some bug somewhere else. Of course, we should be getting the `inputs_to_ivy_arrays` decorator change merged as the order of the decorator should be fixed anyway. Thanks @hello-fri-end slightly_smiling_face
> 
> This is the explanation given in the Inplace Updates section of the deep dive about why adding the `handle_out_argument` decorator and `support_native_out` attribute is necessary for mixed functions 🤔
> 
> "Unlike compositional arguments, where the handle_out_argument decorator is not included, this decorator should be included for mixed functions. This decorator is needed in order to ensure the out argument is handled correctly when the backend does include a backend-specific implementation, which itself may or may not handle the out argument explicitly. In such cases, the mixed function behaves like a primary function. If the backend-specific implementation does not handle the out argument explicitly (there is no attribute support_native_out specified on the backend function), then it will need to be handled [in the decorator](https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L287).
> 
> However, the inclusion of this decorator means that in cases where the mixed function is called compositionally (there is no backend implementation), then the out argument will also be handled [in the decorator](https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L287), this time because of the lack of the support_native_out attribute found on the compositional implementation. But this is not ideal. All compositional implementations are fully capable of handling the out argument explicitly, and so handling it [in the decorator](https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L287) will likely be less efficient, and prevent us from leveraging backend-specific in-place optimizations where they might exist when calling the individual Ivy functions of the compositional implementation.
> 
> Therefore, we always add the support_native_out attribute to mixed functions, to ensure that the out argument is always handled directly by the compositional implementation, rather than being handled [in the decorator](https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L287)."

Ah yeah makes sense, I think we would need the `handle_out_argument` in mixed functions. But I don't understand the purpose of adding `support_native_out` to the compositional implementation in mixed functions because ideally we could just pass the `ivy.Array` in the input without converting it to `ivy.NativeArray` at all https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L279 Given that `to_native` is a very general function, avoiding this conversion could save us quite some time. Probably we could add an attribute like `support_ivy_out` or something for the compositional implementation in the mixed function and then capture the `hasattr` for it in `handle_out_in_ivy`. Finally we'd just modify the first `if` block in `handle_out_argument` wrapper https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L275 to,
```
if out is None or handle_out_in_ivy:
```
Also, could you please also update that paragraph in the docs explaining this?
Let me know if there's anything I could clarify on! Thanks @hello-fri-end 🙂

hello-fri-end:
Probably we could add an attribute like `support_ivy_out` or something for the compositional implementation in the mixed function and then capture the `hasattr` for it in `handle_out_in_ivy`. Finally we'd just modify the first `if` block in `handle_out_argument` wrapper
> https://github.com/unifyai/ivy/blob/588618fe04de21f79d68a8f6cbb48ab3402c6905/ivy/func_wrapper.py#L275
> 
> to,
> ```
> if out is None or handle_out_in_ivy:
> ```
> 

Yeah, this makes sense! Thanks, will update shortly :)
> Also, could you please also update that paragraph in the docs explaining this? Let me know if there's anything I could clarify on! Thanks @hello-fri-end slightly_smiling_face

sure, thanks for the suggestion!

ivy/functional/ivy/elementwise.py
hello-fri-end: Updated.

hello-fri-end:
All done @VedPatwardhan :+1:

hello-fri-end:
Hey @VedPatwardhan, can we merge this?

VedPatwardhan (changes requested):
Hey @hello-fri-end, just one final change, and we're good to merge. Also, I was wondering what you think about just making use of the `mixed_function` attribute itself altogether rather than a separate `handle_out_in_ivy` attribute in the `handle_out_argument` decorator. I had originally suggested a separate argument just for more control to the user, but I'm open to just having the `mixed_function` argument if you think a separate attribute isn't required.
Feel free to request a review once you make the requested change and your opinion on using the same attribute, we can get the PR merged.
Finally, apologies for the delay, this PR slipped through somehow. Thanks 🙂

hello-fri-end:
> Hey @hello-fri-end, just one final change, and we're good to merge. Also, I was wondering what you think about just making use of the `mixed_function` attribute itself altogether rather than a separate `handle_out_in_ivy` attribute in the `handle_out_argument` decorator. I had originally suggested a separate argument just for more control to the user, but I'm open to just having the `mixed_function` argument if you think a separate attribute isn't required. Feel free to request a review once you make the requested change and your opinion on using the same attribute, we can get the PR merged. Finally, apologies for the delay, this PR slipped through somehow. Thanks slightly_smiling_face

Ah yes, you are right, we can make use of the `mixed_function` attribute, there's no need for a separate attribute. Will update shortly, thanks for pointing it out :rocket:

ivy/func_wrapper.py
```diff
@@ -596,6 +597,8 @@ def new_fn(*args, out=None, **kwargs):
         """
         if out is None:
             return fn(*args, **kwargs)
+        if handle_out_in_ivy:
```
VedPatwardhan, line 600: I don't think there would be a difference between `fn(*args, **kwargs)` and `fn(*args, out=out, **kwargs)` when `out` is `None`, let's just club these to make them more compact like,
```
if handle_out_in_ivy or out is None:
    return fn(*args, out=out, **kwargs)
```

VedPatwardhan (approved):
Thanks for making the changes, a small thing left, feel free to merge once that's done (no need to request a review for this)
Thanks @hello-fri-end 🙂

Issue #14098, opened 2023-04-13
Implement <Cummax> on Backend


tvsharish:
For testing of cummax pytorch frontend ( #13037 ), backend of cummax needs to be implemented. If cummax frontend is implemented then it can be added to Miscellaneous Operations and to Pytorch Frontend ( #16 ).

Issue #16, opened 2022-01-19
Add Miscellaneous Operations to PyTorch Frontend


djl11:
Add [miscellaneous operations](https://pytorch.org/docs/stable/torch.html#other-operations) to PyTorch frontend:

- [ ] atleast_1d
- [ ] atleast_2d
- [ ] atleast_3d
- [ ] bincount
- [ ] block_diag
- [ ] broadcast_tensors
- [ ] broadcast_to
- [ ] broadcast_shapes
- [ ] bucketize
- [ ] cartesian_prod
- [ ] cdist
- [ ] clone
- [ ] combinations
- [ ] corrcoef
- [ ] cov
- [ ] cross
- [ ] cummax
- [ ] cummin
- [ ] cumprod
- [ ] cumsum
- [ ] diag
- [ ] diag_embed
- [ ] diagflat
- [ ] diagonal
- [ ] diff
- [ ] einsum
- [ ] flatten
- [ ] flip
- [ ] fliplr
- [ ] flipud
- [ ] kron
- [ ] rot90
- [ ] gcd
- [ ] histc
- [ ] histogram
- [ ] meshgrid
- [ ] lcm
- [ ] logcumsumexp
- [ ] ravel
- [ ] renorm
- [ ] repeat_interleave
- [ ] roll
- [ ] searchsorted
- [ ] tensordot
- [ ] trace
- [ ] tril
- [ ] tril_indices
- [ ] triu
- [ ] triu_indices
- [ ] vander
- [ ] view_as_real
- [ ] view_as_complex
- [ ] resolve_conj
- [ ] resolve_neg

PatternFinder:
- [ ] flip #2647
- [ ] fliplr #2647
- [ ] flipud #2647

PatternFinder:
- [ ] #2713

PatternFinder:
- [ ] #2714

PatternFinder:
- [ ] #2715

PatternFinder:
- [ ] #2897

vishwaksenan:
- [ ] cumsum #3156

PatternFinder:
- [ ] #3279

JamieLine:
- [ ] #3818

mazy06000:
- [ ] #3915

Kaustav-Sarkar:
- [ ]  #3930

JamieLine:
- [ ] #4070

JamieLine:
- [ ] #4078

[...]

PR #14099, opened 2023-04-13
Implemented cummax backend and Pytorch Frontend of cummax

tvsharish:
I have implemented Torch,Numpy and Jax Backend for cummax backend ( #14098 ). Pytorch frontend of cummax can be tested ( #13037 and  #16 )  .


MahmoudAshraf97:
Please open separate PRs for the frontend and backend implementation, make sure you explain the reason you suggest the function. This can either be links to native functions that exist in 2 or more frameworks, or that it is too complex to implement compositionally.

PR #14108, opened 2023-04-13
Add torch frontend float_power

### LINKED ISSUE #14096: float_power

binh234:
close #14096


AnnaTz:
> Hi @AnnaTz, these are backend-specific errors for Tensorflow and JAX.
> 
> Tensorflow and JAX will return output with the same `dtype` of inputs, e.g. bfloat16 -> bfloat16, float16 -> float16; while torch and numpy always return output as a `float64` tensor for float operations and `complex128` tensor for complex operations. Because of this, using bfloat16, float16, and float32 tensors in Tensorflow and JAX can cause computation overflows. When the results become too large or too small, which may not be representable within the limited precision, Tensorflow and JAX will return `inf` values, while torch and numpy return `float64` tensor and will not cause computation overflows. I don't think we should change this behavior for Tensorflow and JAX.

If this issue only exists for very small and very high values, then you could modify the test to avoid these. You could set the safety factor and/or and min and max values of the generated array.

binh234:
@AnnaTz, thanks for your suggestions! I have fixed all the errors

ivy_tests/test_ivy/test_frontends/test_torch/test_pointwise_ops.py
```diff
@@ -1676,6 +1676,39 @@ def test_torch_pow(
...
+def test_torch_float_power(
+    dtype_and_x,
+    on_device,
+    fn_tree,
+    frontend,
+    test_flags,
+):
+    input_dtype, x = dtype_and_x
+    # bfloat16 uses different semantics for exponent and mantissa than other float types
+    assume("bfloat16" not in input_dtype)
```
AnnaTz, line 1699: It's best practice to define the unsupported dtypes using a decorator to the function instead of filtering them in the test. Please have a look at [this](https://lets-unify.ai/docs/ivy/overview/deep_dive/data_types.html#supported-and-unsupported-data-types) section of our docs.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_pointwise_ops.py
binh234: @AnnaTz, all frameworks support float_power operation on bfloat16 tensors, the issue is that bfloat16 uses only 7 bits for the mantissa part, while other float types (float16, float32, float64) use at least10 mantissa bits:

bfloat16

Some of the test cases will fail when comparing a bfloat16 tensor (returned from Tensorflow/JAX) with a float64 tensor (returned from torch/numpy). One of them is:
```text
inp = tensor(2, dtype=bfloat16)
exp = tensor(1.25, dtype=bfloat16)
out = float_power(inp, exp)
```
Tensorflow/JAX will return a bfloat16 of 2.375, while torch/numpy will return a float64 tensor of 2.3784.
To avoid potential failures, I have marked out all bfloat16 cases as bad samples.

AnnaTz: The torch documentation states "If neither input is complex returns a torch.float64 tensor, and if one or more inputs is complex returns a torch.complex128 tensor.". So, I think you can just cast the result of the frontend function to the correct dtype.

ivy_tests/test_ivy/test_frontends/test_torch/test_pointwise_ops.py
```diff
@@ -1676,6 +1676,39 @@ def test_torch_pow(
     )
 
 
+@handle_frontend_test(
+    fn_tree="torch.float_power",
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
```
AnnaTz, line 1682: Also, we want to test for all valid dtypes (`get_dtypes("valid")`). If the function only supports certain dtypes, you should define this using  `@with_unsupported_dtypes` or `@with_supported_dtypes`.

AnnaTz (changes requested):

binh234:
@AnnaTz I found that someone has already unified the behavior for all frameworks in #10358, but the author only upcast the output to either `float64` or `complex128`, which will keep the original value and may not be compatible with the same operation in torch and numpy.
I added a helper function using the same strategy to test again all numeric dtypes, but there are still some potential failures on `bfloat16` as we discussed before. I think you should take a look at this and we should get rid of these errors, either by changing the backend behavior or not testing on `bfloat16` inputs at all.

AnnaTz:
> @AnnaTz I found that someone has already unified the behavior for all frameworks in #10358, but the author only upcast the output to either `float64` or `complex128`, which will keep the original value and may not be compatible with the same operation in torch and numpy. I added a helper function using the same strategy to test again all numeric dtypes, but there are still some potential failures on `bfloat16` as we discussed before. I think you should take a look at this and we should get rid of these issues, either by changing the backend behavior or changing the testing strategy.

Alright, I had a look at ivy.float_power and its test is failing with several bugs so we can't expect the frontend to be working seamlessly. I will merge this as is.

PR #14109, opened 2023-04-13
Add frombuffer to Torch frontend

### LINKED ISSUE #14107: frombuffer

adityagandhamal:
Close #14107
Implemented and tested the creation ops function `frombuffer` in Torch frontend

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_torch/test_creation_ops.py
+++ b/ivy_tests/test_ivy/test_frontends/test_torch/test_creation_ops.py
@@ -2,6 +2,7 @@
 import ivy
 from hypothesis import strategies as st, assume
 import math
+import numpy as np
 
 # local
 import ivy_tests.test_ivy.helpers as helpers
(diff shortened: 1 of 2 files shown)


adityagandhamal:
@V-G-spec, I've looked into the failing test cases and the linter issues. These failures do not pertain to my changes.
Would you please have a look and assure the same?
Thanks!

adityagandhamal:
Makes sense to me. Let me give it a try and get back to you. 
Thanks!

ivy_tests/test_ivy/test_frontends/test_torch/test_creation_ops.py
```diff
@@ -688,3 +689,48 @@ def test_torch_from_dlpack(
...
+@st.composite
+def _get_dtype_buffer_count_offset(draw):
+    dtype, value = draw(
+        helpers.dtype_and_values(
+            available_dtypes=helpers.get_dtypes("valid"),
+        )
+    )
+    value = np.array(value)
+    length = value.size
+    value = value.tobytes()
```
adityagandhamal, line 703: Hey @V-G-spec, here's my inference after experimenting with both Numpy and its Ivy equivalent:

### Case 1: Numpy
On converting a numpy array to bytes, we get the following:

```
>>> np_array = np.array([1, 2])
>>> bytes = np_array.tobytes()
>>> bytes
b'\x01\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00' 
```

### Case 2: Ivy 
Whereas on converting an Ivy array of the same size, there's a difference in the number of bytes 

```
>>> ivy_arr = ivy.array([1, 2]) 
>>> bytes = ivy_arr.tobytes()
>>> bytes
b'\x01\x00\x00\x00\x02\x00\x00\x00'
```

ivy_tests/test_ivy/test_frontends/test_torch/test_creation_ops.py
```diff
@@ -688,3 +689,48 @@ def test_torch_from_dlpack(
...
+
+    offset = draw(helpers.ints(min_value=0, max_value=length - 1))
+    count = draw(helpers.ints(min_value=-(2**30), max_value=length - offset))
+    if count == 0:
+        count = -1
+    offset = offset * np.dtype(dtype[0]).itemsize
```
adityagandhamal, line 709: For setting up the `offset: [int]`, it needs to be multiplied with the element size of the `dtype` object generated during the test. While Numpy offers this functionality to extract the itemsize from the dtype, I could not find the equivalent way to do so in Ivy.

adityagandhamal:
Also to mention, I tried running the tests with an Ivy array, barring on the offset, turns out the tests fail on PyTorch and TensorFlow, experiencing an IvyBackendException Error for torch and an Attribute error for tensorflow saying `'tensorflow.python.framework.ops.EagerTensor' object has no attribute 'tobytes'`

adityagandhamal:
Yet, using `ivy.to_numpy()` for the above-mentioned cases, the tests passed on all backends except torch with the same backend exception as above.

At least this far, I think Numpy is the way to go about, as it enables us to extract the itemsize of the dtype for setting up the offset parameter.

adityagandhamal:
And also, I think this is the reason why [Ivy's functional API](https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_creation.py#L346) use Numpy. Would like to know your opinions on this...

V-G-spec:
Thanks for the detailed explanation. I'll take a look at this shortly

V-G-spec:
All looks good. Marking conversations as resolved and merging your PR
Once again, thanks for diving deep into this

PR #14126, opened 2023-04-13
added log1p

### LINKED ISSUE #14112: log1p

piyushgit011:
close #14112


Daniel4078 (changes requested):
Hello. I just found a strange,potentially datatype related problem with your code: E   AssertionError:  the results from backend torch and ground truth framework tensorflow do not match
E    inf!=87.49822998046875 
E   
E   
E   Falsifying example: test_tensorflow_log1p(
E       dtype_and_x=(['float32'], [array(1.0000001e+38, dtype=float32)]),
E           with_out=False,
E           inplace=False,
E       ),
E       fn_tree='ivy.functional.frontends.tensorflow.math.log1p',
E       on_device='cpu',
E   )
in addition, I think this warning is also noteworthy:   /ivy/ivy/functional/backends/numpy/elementwise.py:416: RuntimeWarning: divide by zero encountered in log1p
    return np.log1p(x, out=out)

piyushgit011:
how divide by zero can be encountered

Daniel4078:
> how divide by zero can be encountered

you can ignore that warning I think as it seems to be something wrong inside the numpy backend

Daniel4078:
> what can be datatype related problem?? as in pytorch log1p have beta and threshold parameters

I am not sure why the log1p function for torch backend(https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/torch/elementwise.py#L225) returns inf when the tensorflow returns a normal number. You can try research further about what has happened and it may be not datatype related problem after all.

piyushgit011:
https://github.com/pytorch/pytorch/pull/92114 
can this be potential error??

Daniel4078 (commented):
I am not sure, maybe you should ask on the discord channel here(https://discord.com/channels/799879767196958751/1028706815610978345) about this problem

piyushgit011:
> I am not sure, maybe you should ask on the discord channel here(https://discord.com/channels/799879767196958751/1028706815610978345) about this problem

ok thanks

Daniel4078:
Have you get any reply from the discord about the problem with torch backend?

piyushgit011:
Yes they said to add safety_factor

ivy/functional/frontends/tensorflow/raw_ops.py
```diff
@@ -277,9 +277,7 @@ def Log(*, x, name="Log"):
     return ivy.log(x)
 
 
-@to_ivy_arrays_and_back
-def Log1p(*, x, name="Log1p"):
-    return ivy.log1p(x)
+Log1p = to_ivy_arrays_and_back(map_raw_ops_alias(tf_frontend.math.log1p))
```
Daniel4078, line 280: just found a small error here: https://github.com/unifyai/ivy/actions/runs/4724482305/jobs/8381749112?pr=14126#step:3:123
ivy/functional/backends/torch/elementwise.py:227: in log1p
    return torch.log1p(x, out=out)
E   RuntimeError: "log1p_vml_cpu" not implemented for 'Half'
seems like half is not an valid datatype for the torch backend, so I think you can try to add a unsupported dtype tag in the torch backend log1p function like explained here (https://lets-unify.ai/docs/ivy/overview/deep_dive/data_types.html#supported-and-unsupported-data-types)

Daniel4078 (changes requested):
Great work! Now there is only one small problem left, check the comment below. Thanks

ivy/functional/frontends/tensorflow/raw_ops.py
piyushgit011: done

piyushgit011: on commit in revert tensorflow pr test2 ran successfully but this time after it is again having problem

Daniel4078 (approved):
I found that the problem is of how the test dails to get the correct unsupported datatype from the backend function. So I think your code is good enough now. Thanks.

PR #14152, opened 2023-04-14
fix inplace method testing & other minor tests

xoiga123:
[Trello](https://trello.com/c/FjGCje7g/1291-fix-method-tests-to-work-with-inplace-special-ops)
Continued from #11783

### DIFF
--- a/ivy_tests/test_ivy/helpers/function_testing.py
+++ b/ivy_tests/test_ivy/helpers/function_testing.py
@@ -5,6 +5,7 @@ import numpy as np
 import types
 import importlib
 import inspect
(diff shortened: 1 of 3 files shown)


fspyridakos (approved):

xoiga123:
@CatB1t the copy should be cleaner if multiversion testing uses the args-kwargs object directly instead of its components

CatB1t (changes requested):
lgtm! @xoiga123 great work, Can we separate the in-place changes from the other updates and it has them on a separate PR? That way it's cleaner and easier to track related-changes. I'll get it

PR #14160, opened 2023-04-14
Clone

teckno:


ivy_tests/test_ivy/test_frontends/test_torch/test_other_operations.py
```diff
@@ -0,0 +1,31 @@
...
+)
+def test_ivy_clone(
+    dtypes_values_casting,
+    on_device,
+    fn_tree,
+    frontend,
+    test_flags,
+):
+
+    x = dtypes_values_casting[0]
```
hirwa-nshuti, line 27: We will need to use the helper function instead of this approach
```python
helpers.test_frontend_function(*args, **kwargs)
```

hirwa-nshuti (changes requested):
Hey @teckno 

Some improvements are needed check the comments below.
And also remove the docstrings we don't need them in the frontends😁
Thanks

ivy_tests/test_ivy/test_frontends/test_torch/test_other_operations.py
teckno: Okay just made some changes working on debuging

teckno:
Can i reopen it again

PR #14259, opened 2023-04-17
add min instance to JAX frontend

iamravi1:
added min function to jax frontend
- [ ] #14260


danielmunioz:
HI @iamravi1, thanks for the contribution, the tests are not passing could you please take a look at those? Also I believe you are missing the definition of self? could you please fix that?

PR #14262, opened 2023-04-17
Fix local ivy import

CatB1t:
# Summary:
- Add `LocalIvyImporter` to `dynamic_import` for local import to work properly.
- Add `local_ivy_id` for local ivy id to be used inside `impersonate_import` to fetch the local object through global ivy. 
- Refactor `impersonate_import` implementation.

### DIFF
--- a/ivy/utils/backend/ast_helpers.py
+++ b/ivy/utils/backend/ast_helpers.py
@@ -177,6 +177,8 @@ class ImportTransformer(ast.NodeTransformer):
     def impersonate_import(self, tree: ast.Module, ivy_id=None):
         if self.include_ivy_import:
             if ivy_id is not None:
+                # ast.parse produces a tree with Module tree, using body[0]
+                # we can obtain the specific node that we require.
                 ivy_import_node = ast.parse("import ivy").body[0]
                 common_import_path = (
                     f"ivy.utils.backend.handler."
--- a/ivy/utils/backend/ast_helpers.py
+++ b/ivy/utils/backend/ast_helpers.py
@@ -174,15 +174,15 @@ class ImportTransformer(ast.NodeTransformer):
         else:
             return _parse_relative_fromimport(node)
 
-    def impersonate_import(self, tree: ast.Module, ivy_id=None):
+    def impersonate_import(self, tree: ast.Module, local_ivy_id=None):
         if self.include_ivy_import:
-            if ivy_id is not None:
+            if local_ivy_id is not None:
                 # ast.parse produces a tree with Module tree, using body[0]
                 # we can obtain the specific node that we require.
                 ivy_import_node = ast.parse("import ivy").body[0]
                 common_import_path = (
                     f"ivy.utils.backend.handler."
-                    f"_compiled_backends_ids[{ivy_id}].utils._importlib"
+                    f"_compiled_backends_ids[{local_ivy_id}].utils._importlib"
                 )
                 abs_import_node = ast.parse(
                     f"{importlib_abs_import_fn} "
@@ -249,7 +249,7 @@ class IvyLoader(Loader):
     def __init__(self, filename):
         self.filename = filename
 
-    def exec_module(self, module, ivy_id=None):
+    def exec_module(self, module, local_ivy_id=None):
         if self.filename in _compiled_modules_cache:
             compiled_obj = _compiled_modules_cache[self.filename]
         else:
@@ -259,7 +259,7 @@ class IvyLoader(Loader):
             ast_tree = parse(data)
             transformer = ImportTransformer()
             transformer.visit(ast_tree)
-            transformer.impersonate_import(ast_tree, ivy_id)
+            transformer.impersonate_import(ast_tree, local_ivy_id)
             ast.fix_missing_locations(ast_tree)
             compiled_obj = compile(ast_tree, filename=self.filename, mode="exec")
             _compiled_modules_cache[self.filename] = compiled_obj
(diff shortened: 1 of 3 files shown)


ivy/utils/backend/ast_helpers.py
```diff
@@ -174,24 +174,44 @@ def visit_ImportFrom(self, node):
         else:
             return _parse_relative_fromimport(node)
 
-    def impersonate_import(self, tree: ast.Module):
+    def impersonate_import(self, tree: ast.Module, local_ivy_id=None):
         if self.include_ivy_import:
```
Infrared1029, line 178: I think the following logic can be simplified/unified to look a bit cleaner (the `ast.parse` is such a nice move tbh makes the code so much more readable), anyway, so we want to have this overall effect:
`if local_ivy_id is not None:`
we want this to be slapped at the beginning of the file
```python
import ivy
_absolute_import = ivy.utils.backend.handler._compiled_backends_ids[{local_ivy_id}].utils._importlib._absolute_import
_from_import = ivy.utils.backend.handler._compiled_backends_ids[{local_ivy_id}].utils._importlib._from_import
```

`else:`
we slap this
```python
from ivy.utils.backend.handler.utils._importlib import _absolute_import
from ivy.utils.backend.handler.utils._importlib import _from_import
```
we can merge both cases into one
import ivy

`if local_ivy_id is not None`:
```python
import ivy
_absolute_import = ivy.utils.backend.handler._compiled_backends_ids[{local_ivy_id}].utils._importlib._absolute_import
_from_import = ivy.utils.backend.handler._compiled_backends_ids[{local_ivy_id}].utils._importlib._from_import
```

`else`:
```python
import ivy
_absolute_import = ivy.utils.backend.handler.utils._importlib._absolute_import
_from_import = ivy.utils.backend.handler.utils._importlib._from_import
```

code-wise it would look like this:
```python
# ast.parse produces a tree with Module tree, using body[0]
# we can obtain the specific node that we require.
ivy_import_node = ast.parse("import ivy").body[0]
import_stmt = "ivy.utils.backend.handler"
if local_ivy_id is not None:
    import_stmt += f"._compiled_backends_ids[{local_ivy_id}]"
abs_import_stmt = import_stmt + ".utils._importlib._absolute_import"
from_import_stmt = import_stmt + ".utils._importlib._from_import"


abs_import_node = ast.parse(abs_import_stmt).body[0]
from_import_node = ast.parse(from_import_stmt).body[0]
tree.body.insert(self.insert_index, from_import_node)
tree.body.insert(self.insert_index, abs_import_node)
tree.body.insert(self.insert_index, ivy_import_node)

return tree
```

ivy/utils/backend/ast_helpers.py
```diff
@@ -174,24 +174,44 @@ def visit_ImportFrom(self, node):
...
-            )
-            tree.body.insert(
-                self.insert_index,
-                ast.ImportFrom(
-                    module=importlib_module_path,
-                    names=[ast.alias(name=importlib_abs_import_fn, asname=None)],
-                    level=0,
-                ),
-            )
+            if local_ivy_id is not None:
```
Infrared1029, line 179: I think the following logic can be simplified/unified to look a bit cleaner (the `ast.parse` is such a nice move tbh makes the code so much more readable), anyway, so we want to have this overall effect:
`if local_ivy_id is not None:`
we want this to be slapped at the beginning of the file
```python
import ivy
_absolute_import = ivy.utils.backend.handler._compiled_backends_ids[{local_ivy_id}].utils._importlib._absolute_import
_from_import = ivy.utils.backend.handler._compiled_backends_ids[{local_ivy_id}].utils._importlib._from_import
```

`else:`
we slap this
```python
from ivy.utils.backend.handler.utils._importlib import _absolute_import
from ivy.utils.backend.handler.utils._importlib import _from_import
```
we can merge both cases into one
import ivy

`if local_ivy_id is not None`:
```python
import ivy
_absolute_import = ivy.utils.backend.handler._compiled_backends_ids[{local_ivy_id}].utils._importlib._absolute_import
_from_import = ivy.utils.backend.handler._compiled_backends_ids[{local_ivy_id}].utils._importlib._from_import
```

`else`:
```python
import ivy
_absolute_import = ivy.utils.backend.handler.utils._importlib._absolute_import
_from_import = ivy.utils.backend.handler.utils._importlib._from_import
```

code-wise it would look like this:
```python
# ast.parse produces a tree with Module tree, using body[0]
# we can obtain the specific node that we require.
ivy_import_node = ast.parse("import ivy").body[0]
import_stmt = "ivy.utils.backend.handler"
if local_ivy_id is not None:
    import_stmt += f"._compiled_backends_ids[{local_ivy_id}]"
abs_import_stmt = import_stmt + ".utils._importlib._absolute_import"
from_import_stmt = import_stmt + ".utils._importlib._from_import"


abs_import_node = ast.parse(abs_import_stmt).body[0]
from_import_node = ast.parse(from_import_stmt).body[0]
tree.body.insert(self.insert_index, from_import_node)
tree.body.insert(self.insert_index, abs_import_node)
tree.body.insert(self.insert_index, ivy_import_node)

return tree
```

CatB1t: This code is definitely shorter, but makes it harder for whoever reads the code to actually know what happens right away, I don't think the tradeoff is worth it, also when debugging, `from x import y` may be more intuitive than `import x; y = x.y`

Infrared1029: I actually think it is more readable, your point about `from x import y` being more intuitive is 100% correct, but in this case, we aren't writing code that is explicitly shown, but implicitly injected to do a specific purpose (it is not visible) so I think it is totally fine, a comment (or comments) can be added to explain what is going regardless (whether we go with this way or not actually, a few comments will be needed), I really think this is more concise and easier to debug.

PR #14266, opened 2023-04-17
Fix array inplace testing

xoiga123:
They're failing because torch doesn't allow `inplace` with gradient enabled tensors.

### DIFF
--- a/ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py
+++ b/ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py
@@ -1392,7 +1392,11 @@ def test_positive(
 
 
 @st.composite
-def pow_helper(draw, available_dtypes=None):
+def pow_helper(
+    draw,
+    available_dtypes=None,
+    shared_dtype=False,
+):
     if available_dtypes is None:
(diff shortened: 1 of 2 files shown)


CatB1t (changes requested):
If they're failing only for torch, we can create a custom strategy to not generate `as_variable` flags for torch specifically, also it may be worth getting in touch with @VedPatwardhan regarding that issue, could be something to have a workaround in the backend.

VedPatwardhan:
Hey guys, there's quite a few functions in our API where the `torch` backend implementation supports inplace updates natively and we've been testing with the `as_variable` flag for quite a few months and the tests have been passing, I think it could be one of 2 possibilities,
1. We were already handling this in the test helpers and the handling may have been lost in a recent commit.
2. It's a problem with newer versions of `torch` (which is unlikely as I have remember seeing an error for this in the past).
In either case, I think we should be handling this in the tests itself, where `as_variable` should be set to `False` if `with_out=True` as we were doing before as far as I remember.

Thanks for looking into this @xoiga123 and @CatB1t 🙂

ivy-seed:

ivy-seed:

PR #14290, opened 2023-04-17
Implement np.random.chisquare in frontend

### LINKED ISSUE #14283: chisquare
Add chisquare sampling to Numpy frontend

jshepherd01:
Closes #14283


jshepherd01:
Forgot to account for the case where `df` is an array-like instead of just a single float (although the docs aren't very clear on what should happen in that case)

jshepherd01:
That's deliberate. When I call something like `ivy.gamma([1,2,3], 2, shape=(3,))` Ivy throws `IvyBackendException: numpy: gamma: low and high bounds must be numerics when shape is specified`. This is also an issue with `np.random.standard_gamma`, which throws an error whenever you pass both an array and a size.

To address it, I broadcast `df` into the same shape as `size` before passing it into gamma (lines 107 to 110)

PR #14306, opened 2023-04-18
adding more detailed exception classes

sherry30:


sherry30:
Added 5 custom exception classes and added those in some common functions while making sure that the same exception is raised for same error regardless of the backend

juliagsy (commented):
Hey Sherry! The changes look good so far! Just a side point, do you think it is possible to have a native exception attribute within the exception class for keeping track of it in the Ivy Exception so that it is not hidden or lost totally? (so this would resemble something like the ivy array or ivy shape), Thanks!

sherry30:
Hi @juliagsy,

Thanks for the review. Yes, we can keep track of the original exception raised, I've added this in `IvyIndexError` exception class and in `all` function in tensorflow backend in my last commit. Let me know if this works and then I'll add it in the rest of the  classes and functions as well.

Are you thinking of maybe adding a trace mode `native` in which it will display native exception instead of ivy exception? 🤔

VedPatwardhan (commented):
Hey @sherry30, thanks for requesting the review. The changes look good to me, however I think @juliagsy should have the say on this PR given her work with the exception classes. Thanks @juliagsy 🙂

PR #14374, opened 2023-04-18
Added `compress` + test to Ivy Experimental API

mobley-trent:
#13977


MahmoudAshraf97 (commented):
Hi and sorry for the huge delay, we already have some very similar functions in ivy such as gather and gather_nd, so in what cases is this function needed and there is no function in ivy that can fulfill that role?

mobley-trent:
Both `ivy.gather` and `ivy.gather_nd` have a different functionality compared to `jnp.compress`. I had initially considered them both but the ground truth output differed from the frontend. This is why I decided to implement the `compress` function in the Ivy API then I could proceed to add the Jax frontend for this function.

MahmoudAshraf97:
then you can implement the frontend jax function directly using a compositional solution, this shouldn't be added to the functional API

PR #14385, opened 2023-04-19
Add idct signal for tensorflow

omarahmed19992:
In this PR, I address the issue of obtaining IDCT for input by taking the following steps:

- Conducting research: I used TensorFlow documentation, as well as resources such as SciPy and Wikipedia, to gain a deeper understanding of how IDCT works behind the scenes.
- Reviewing the code: I studied the code on the GitHub repository to get a better sense of the implementation details.
- Exploring DCT: I examined the use of DCT in the master branch to gain insights into how it worked.
- Implementing a new function: Based on my research and analysis, I created a new function for IDCT that can be used for input.

I believe that this solution effectively addresses the issue and I look forward to your feedback.
Close https://github.com/unifyai/ivy/issues/14045

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_signal.py
+++ b/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_signal.py
@@ -5,41 +5,41 @@ from hypothesis import strategies as st
 import ivy_tests.test_ivy.helpers as helpers
 from ivy_tests.test_ivy.helpers import handle_frontend_test
 
-#
-# # kaiser_window
-# @handle_frontend_test(
-#     fn_tree="tensorflow.signal.kaiser_window",
-#     dtype_and_window_length=helpers.dtype_and_values(
-#         available_dtypes=helpers.get_dtypes("integer")
-#     ),
-#     dtype_and_beta=helpers.dtype_and_values(
-#         available_dtypes=helpers.get_dtypes("numeric")
-#     ),
-#     dtype=helpers.get_dtypes("numeric"),
-#     test_with_out=st.just(False),
-# )
-# def test_tensorflow_kaiser_window(
-#     *,
-#     dtype_and_window_length,
-#     dtype_and_beta,
-#     dtype,
-#     frontend,
-#     test_flags,
-#     fn_tree,
-#     on_device,
-# ):
-#     window_length_dtype, window_length = dtype_and_window_length
-#     beta_dtype, beta = dtype_and_beta
-#     helpers.test_frontend_function(
-#         input_dtypes=[window_length_dtype[0], beta_dtype[0]],
-#         frontend=frontend,
-#         test_flags=test_flags,
-#         fn_tree=fn_tree,
-#         on_device=on_device,
-#         window_length=window_length,
-#         beta=beta,
-#         dtype=dtype,
-#     )
+
+# kaiser_window
+@handle_frontend_test(
+    fn_tree="tensorflow.signal.kaiser_window",
+    dtype_and_window_length=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("integer")
+    ),
+    dtype_and_beta=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("numeric")
+    ),
+    dtype=helpers.get_dtypes("numeric"),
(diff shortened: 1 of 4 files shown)


omarahmed19992:
I know that, I got more failures in checking , but when I'm testing on my device I got error which relate to another backends like torch and numpy, i get 4 failing error and 1 warning , in those four I didn't get failure in tensorflow 


Screenshot (7)

omarahmed19992:
Close #14385

omarahmed19992:
Screenshot (9)

Screenshot (8)


as in images, the test is failed because of this example which is in hypo, but when I check tf.idct doc found this right when axis != -1, we get error " not implemented "

Infrared1029:
Hi @omarahmed19992! I took a quick look and noticed you are implementing the function using `tensorflow`, since this a frontend function, you should be using `ivy` to recreate the native function's logic, more information could be found here [here](https://lets-unify.ai/docs/ivy/overview/deep_dive/ivy_frontends.html).

Infrared1029:
Hi @omarahmed19992!
I've taken a look at the PR, there are a few issues that I noticed that are causing the errors (more issues could be there but these are the ones I found so far)

- At the moment the testing strategy generates values for `n` and `axis` besides their currently allowable values `None` and `-1`, for more info look at the parameters of the function [here](https://www.tensorflow.org/api_docs/python/tf/signal/idct).
- `type=1` and `norm=ortho` is not valid, your test must account for that (if the `type` generated is `1` make sure to set `norm` to `None`
- The function does not accept all float types, only float32 and 64

Infrared1029:
Hi @omarahmed19992! we are almost there, adding `atol` of `1e-2` seems to pass all the rounding issues in the numpy and jax backends, the torch backend still fails but this is not on your side (it fails when `ivy.dct` receives `float64` input with a torch backend set), just make sure that the function signature follows the original's `tf.signal.idct(
    input, type=2, n=None, axis=-1, norm=None, name=None
)`, include `name` and remove `out`, and make sure all the default values are the same as the original tensorflow function, make sure to uncomment the tests in the test_signal file as well.

omarahmed19992:
Hi @Infrared1029 for  adding atol to pass all the rounding issues in the numpy and jax backends, it sometime work and go on not always

Infrared1029:
> Hi @Infrared1029 for adding atol to pass all the rounding issues in the numpy and jax backends, it sometime work and go on, not always

You are right, i guess we might need to increase `atol=1e-1`, this will probably remove all failures from all backends (ofc except torch), let me know how it goes.

omarahmed19992:
@Infrared1029 u r right , it pass when it's 1e-1,
 thanks a lot

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_signal.py
```diff
@@ -40,3 +40,58 @@ def test_tensorflow_kaiser_window(
...
+            max_value=65280,
+            min_value=-65280,
+            min_num_dims=1,
+            max_num_dims=5,
+            min_dim_size=2,
+            max_dim_size=10,
+            shared_dtype=True,
+        )
+    )
+    n = draw(st.sampled_from([None]))
```
Infrared1029, line 59:
```suggestion
    n = None
```

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_signal.py
```diff
@@ -40,3 +40,58 @@ def test_tensorflow_kaiser_window(
...
+    axis = draw(st.sampled_from(["int"]))
```
Infrared1029, line 60:
```suggestion
    axis = -1
```

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_signal.py
```diff
@@ -40,3 +40,58 @@ def test_tensorflow_kaiser_window(
...
+    norm = draw(st.sampled_from([None, "ortho"]))
+    type = draw(helpers.ints(min_value=1, max_value=4))
+    if axis == "int":
```
Infrared1029, line 63: will not be needed anymore because of the previous changes.

Infrared1029:
lgtm! merging

PR #14394, opened 2023-04-19
Add unique to Torch frontend

### LINKED ISSUE #14264: unique

adityagandhamal:
Close #14264 
Implemented and tested the reduction ops function `unique` in Torch frontend

### DIFF
--- a/ivy/functional/frontends/torch/reduction_ops.py
+++ b/ivy/functional/frontends/torch/reduction_ops.py
@@ -230,3 +230,24 @@ def logsumexp(input, dim, keepdim=False, *, out=None):
         c = ivy.squeeze(c, axis=dim)
     ret = ivy.add(ret, c, out=out)
     return ret
+
+
+@to_ivy_arrays_and_back
+def unique(input, sorted=True, return_inverse=False, return_counts=False, dim=None):
+    results = ivy.unique_all(input, axis=dim)
+
+    fields = ["values"]
+    if return_inverse:
+        fields.append("inverse_indices")
+    if return_counts:
+        fields.append("counts")
+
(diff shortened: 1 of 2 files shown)


adityagandhamal:
The existing linter issues do not pertain to my changes.

V-G-spec:
Why did you take sorted to be True?

V-G-spec:
That is correct, but Torch returns the elements in descending order when `sorted` is set as `False`. That will require a separate case. An easy fix would be to invert the tensor, which won't change the time complexity. But see if there's a way not to have this overhead

V-G-spec:
Not sure about JAX, but NumPy returns them in ascending order if I remember correctly. So that should be fine since JAX' is LAX-backend implementation of numpy's implementation

adityagandhamal:
> Torch returns the elements in descending order when `sorted` is set as `False`. 

I think that's not the case specifically. Cause on setting `sorted` to `False`, Torch did not return the values in descending order...
```
>>> torch_arr = torch.tensor([4,4,1,2,0])
>>> unique_vals = torch.unique(torch_arr)
>>> unique_vals
tensor([0, 1, 2, 4])
>>> unique_vals_2 = torch.unique(torch_arr, sorted=False)
>>> unique_vals_2
tensor([0, 2, 1, 4])
```

Any ideas on how to address this scenario?

adityagandhamal:
> Not sure about JAX, but NumPy returns them in ascending order if I remember correctly. So that should be fine since JAX' is LAX-backend implementation of numpy's implementation

Exactly, if we set `sorted` to `False` in the Torch frontend, we won't get the same results when we have our backend set to Numpy (for instance) as for numpy, it returns a sorted array.

PR #14399, opened 2023-04-19
Add MaxPool2D and AvgPool2D layers

valeriocost:


ivy/stateful/pooling.py
```diff
@@ -0,0 +1,98 @@
...
+        Parameters
+        ----------
+        x
+            The input to the layer.
+
+        Returns
+        -------
+        The output of the layer.
+        """
+        return inputs.avg_pool2d(self._kernel_size, self._stride, self._padding)
```
sherry30, line 98: would be better to not call it as an instance method and instead as `ivy.max_pool2d`. Would give an error in case a native array is passed as input.

sherry30 (changes requested):
Mostly looks good,

I think we should put these in layers submodule since all the pooling functions are also in layers.
The rest of the changes I've requested on specific lines.

Thanks

PR #14427, opened 2023-04-20
Ivy.Shape refactor 

Aarsh2001:
Aim to design `Ivy.Shape` class similar to the `Ivy.Array` class to defer to `NativeShape` and `NativeArray` respectively. Here's the basic version that works with the example file. Few dunder methods and properties are yet to be implemented, they're on my list. Let me know your thoughts on this @VedPatwardhan @guillesanbri @djl11 .

https://trello.com/c/yNr2eMjB

I have a few questions regarding the same-: 

1.  What is the proper way to check for behavior that breaks?
2. should we still inherit from the `tuple` class, as @djl11  mentioned [here](https://discord.com/channels/799879767196958751/1029345398042677298/1096171210338992138)
3.  At last, the superset implementations covering methods from `torch.Size` and `tensorflow.TensorShape` ? (We can probably discuss on this later @VedPatwardhan )


ivy/__init__.py
```diff
@@ -202,17 +202,37 @@ def __new__(cls, shape_tup):
                 current_backend(shape_tup).NativeArray,
             )
         ivy.utils.assertions.check_isinstance(shape_tup, valid_types)
-        if isinstance(shape_tup, int):
-            shape_tup = (shape_tup,)
+        if isinstance(shape_tup, ivy.NativeShape):
+            self._shape = shape_tup
```
guillesanbri, line 206: regarding the transpiler I'd say this is the only relevant one, so can't say much on the other cases

ivy/__init__.py
```diff
@@ -202,17 +202,37 @@ def __new__(cls, shape_tup):
...
-        ivy.utils.assertions.check_all(
-            [isinstance(v, int) or ivy.is_int_dtype(v.dtype) for v in shape_tup],
-            "shape must take integers only",
+            self._shape = tuple(shape_tup)
+        else:
+            self._shape = None
+
+    def __repr__(self):
+        return (
+            f"Ivy.Shape{self._shape}" if self._shape is not None else "Ivy.Shape(None)"
```
guillesanbri, line 216: I guess that this would be a bit weird with tf and torch? not sure but the output would be something like `ivy.Shapetorch.Shape(1, 2, 3)`?

ivy/__init__.py
```diff
@@ -202,17 +202,37 @@ def __new__(cls, shape_tup):
...
+
+    def __getattribute__(self, item):
+        return super().__getattribute__(item)
+
+    def __getitem__(self, key):
+        return self._shape[key] if self._shape is not None else None
+
+    def __setattr__(self, key, value):
+        if self._shape is not None:
+            self._shape[key] = value
```
guillesanbri, line 230: this would be the implementation of `setitem` and not `setattr` I guess. On top of that, I think this would fail with jax and numpy, as their shapes are tuples and therefore don't support item assignment directly

guillesanbri (commented):
Just left some comments, but I'm probably missing quite a few things in the framework side :)

guillesanbri:
> Aim to design `Ivy.Shape` class similar to the `Ivy.Array` class to defer to `NativeShape` and `NativeArray` respectively. Here's the basic version that works with the example file. Few dunder methods and properties are yet to be implemented, they're on my list. Let me know your thoughts on this @VedPatwardhan @guillesanbri @djl11 .
> 
> https://trello.com/c/yNr2eMjB
> 
> I have a few questions regarding the same-:
> 
> 1. What is the proper way to check for behavior that breaks?
> 2. should we still inherit from the `tuple` class, as @djl11  mentioned [here](https://discord.com/channels/799879767196958751/1029345398042677298/1096171210338992138)
> 3. At last, the superset implementations covering methods from `torch.Size` and `tensorflow.TensorShape` ? (We can probably discuss on this later @VedPatwardhan )

Regarding the questions:
1. Not sure about the framework, but the transpilation snippet that you added should work with the refactored shape. I can take a look if this becomes a pain point once the implementation is ready.
2. I don't see much benefit of this tbh, we could get some pretty similar behaviour implementing `__iter__` and similar methods, but again, maybe there is something framework-side that I'm missing and subclassing tuple would be an advantage

ivy/__init__.py
```diff
@@ -202,17 +203,34 @@ def __new__(cls, shape_tup):
...
-            [isinstance(v, int) or ivy.is_int_dtype(v.dtype) for v in shape_tup],
-            "shape must take integers only",
+            self._shape = tuple(shape_tup)
+        else:
+            self._shape = None
+
+    def __repr__(self):
+        pattern = r"\d+(?:,\s*\d+)*"
+        shape_repr = re.findall(pattern, self._shape.__str__())
+        shape_repr = shape_repr[0] + "," if len(shape_repr[0]) == 1 else shape_repr[0]
```
guillesanbri, line 218: maybe something like this is easier to understand? 
```python
shape_repr = ", ".join([str(i) for i in self._shape])
shape_repr = f"ivy.Shape({shape_repr})"
```

guillesanbri (commented):
minor suggestion, let me know if you need help with the failing tests :)

VedPatwardhan (approved):
lgtm! Let's get this merged and then see if there are any outstanding fixes left apart from the ones we've made already. Thanks @Aarsh2001 🙂

PR #14434, opened 2023-04-20
Frontend for tf.math.angle

### LINKED ISSUE #13752: angle

AwkNinja:
Close #13752


AwkNinja:
Hi @hmahmood24 all the test cases for the tf.math.angle function are passing. I did not update the raw_ops function as it had a bit of a different functionality. Let me know if I should change anything. Thanks.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
```diff
@@ -2358,6 +2358,32 @@ def test_tensorflow_less(
     )
 
 
+# angle
+@handle_frontend_test(
+    fn_tree="tensorflow.math.angle",
+    dtype_and_input=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
```
hmahmood24, line 2365: The `tf.math.angle` function should be tested against `complex (or real)` inputs. Can you kindly change this from `"float"` to `"numeric"`?

hmahmood24 (changes requested):
Hey @AwkNinja. Thanks so much for the PR. I have left a comment for you to address. There is also a linting/formatting issue with your change file [here](https://github.com/unifyai/ivy/actions/runs/4754126825/jobs/8446565155?pr=14434#step:5:15). Kindly look into these and ask for a re-review from me once you're done. Thanks! 😃

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
AwkNinja: In that case shouldn't I test it with `complex` instead of `numeric`?

hmahmood24: @AwkNinja `complex` is a subset of `numeric` which also has `float` dtypes which we need to test as well.

PR #14450, opened 2023-04-20
add unique to jax.numpy frontend

mohame54:
#14263

### DIFF
--- a/ivy/functional/frontends/jax/numpy/searching_sorting.py
+++ b/ivy/functional/frontends/jax/numpy/searching_sorting.py
@@ -161,12 +161,20 @@ def where(condition, x=None, y=None, size=None, fill_value=0):
 
 
 @to_ivy_arrays_and_back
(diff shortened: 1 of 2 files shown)


Ookamice:
Hiya, thanks for contributing. Just two quick fixes before I can get this merged:
1) please could you remove the key.pem file from the commit?
2) there's a lint error in the comited files too, I would suggest using the pre-commit hook to prevent these, instructions to install [here](https://lets-unify.ai/docs/ivy/overview/contributing/setting_up.html#pre-commit).

Thank you  😊

Ookamice (changes requested):
Hiya, thanks for contributing. Just two quick fixes before I can get this merged:

1) please could you remove the key.pem file from the commit?
2) it looks like there's lint errors in the committed files too. I would suggest using the pre-commit hook to prevent these, instrucftion to install [here](https://lets-unify.ai/docs/ivy/overview/contributing/setting_up.html#pre-commit)

Thank you 😊

Ookamice (approved):
Looks good to me! 😃

PR #14462, opened 2023-04-21
dtype casting mode update

RickSanchezStoic:
Updated dtype casting mode, with bug fixes.

### DIFF
--- a/run_tests_CLI/filter_tests.py
+++ b/run_tests_CLI/filter_tests.py
@@ -1,6 +1,5 @@
 import sys
 from get_all_tests import get_all_tests
-from tqdm import tqdm
 
 N = 40
 
@@ -13,7 +12,7 @@ def main():
     start = run_iter * tests_per_run
     end = num_tests if run_iter == N - 1 else (run_iter + 1) * tests_per_run
     with open("tests_to_run", "w") as f:
-        for test in tqdm(test_names[start:end]):
+        for test in test_names[start:end]:
             f.write(test + "\n")
 
 


ivy/func_wrapper.py
```diff
@@ -35,6 +35,175 @@
 # Helpers #
 # --------#
 
+# for casting modes, order is the hierarchy
+casting_modes_dict = {
```
VedPatwardhan, line 39: Can we get this dictionary dynamically instead? Would be easier to maintain if we just did 
```
{
    "uint": ivy.valid_uint_dtypes,
    "int": ivy.valid_int_dtypes,
    and so on ...
}
```

ivy/func_wrapper.py
```diff
@@ -35,6 +35,175 @@
...
+        elif ivy.downcast_dtypes:
+            # check downcasting
+            dtype = downcaster(dtype, intersect)
+            if dtype:
+                return dtype
+
+
+def upcaster(dtype, intersect):
+    # upcasting is enabled, we upcast to the highest
+    if "uint" in str(dtype):
```
VedPatwardhan, line 87: I think we can just separate out the number and the string from the dtype, so that we could reduce some of the redundancy (we wouldn't need to check if it's a `uint`, `int` or `float`)

ivy/func_wrapper.py
```diff
@@ -35,6 +35,175 @@
...
+            ):
+                result = casting_modes_dict["complex"][index]
+                break
+            index += 1
+        return result
+
+
+def downcaster(dtype, intersect):
+    # downcasting is enabled, we upcast to the highest
+    if "uint" in str(dtype):
```
VedPatwardhan, line 142: Same suggestion as above

ivy/func_wrapper.py
```diff
@@ -528,6 +697,38 @@ def _to_float_array(x):
     return _integer_arrays_to_float
 
 
+def casting_modes_ops(fn):
+    @functools.wraps(fn)
+    def method(*args, **kwargs):
+        # we first check if it has unsupported/supported dtypes uniquely added to it
+        intersect = set(ivy.function_unsupported_dtypes(fn)).difference(
```
VedPatwardhan, line 704: Wouldn't it be better if we just set `intersect = ivy.function_unsupported_dtypes(fn)`? Because `args` can never have an array with `invalid_dtypes` but `kwargs['dtype']` could?

VedPatwardhan (changes requested):
Just requested a few changes, thanks @RickSanchezStoic 🙂

VedPatwardhan:
Also, could you please add a section in the deep dive under the data types page explaining the purpose and usage of all 3 modes? Thanks @RickSanchezStoic 🙂

ivy/func_wrapper.py
RickSanchezStoic: sure, this would require us to sort the valid dtype dictionaries that we have maintained for each framework, because order is important here

ivy/func_wrapper.py
RickSanchezStoic: It's for a different reason, we don't apply the operation on functions that don't have the unsupported dtype attributes applied. The `function_unsupported_dtypes` returns not just the specified dtypes , but also the invalid dtypes set for that framework

ivy/func_wrapper.py
VedPatwardhan: I think they are sorted already in ivy, which would save us from sorting explicitly 🙂

RickSanchezStoic: I tried this, but due to cyclic dependency, we won't be able to use `ivy.valid_int/float/uint/dtypes` here in the func wrapper. Since this is a constant dictionary that won't require much update unless new types are found, we can continue with this

VedPatwardhan: Oh okay got it! I think we could also just not have this dictionary at all and simply make use of `ivy.valid_int_dtypes`, `ivy.valid_float_dtypes`, etc. in the casting functions directly. I think we'll always have those lists in sorted order anyway 🙂

VedPatwardhan (changes requested):
Just a few more changes, thanks @RickSanchezStoic 🙂

RickSanchezStoic:
Have refactored the code a bit. Also added another casting mode. Updated the docs too.

PR #14503, opened 2023-04-23
add jax.random.rademacher frontend

MuhammedAshraf2020:


ivy_tests/test_ivy/test_frontends/test_jax/test_jax_random.py
```diff
@@ -308,3 +308,57 @@ def call():
...
+        min_value=0,
+        max_value=2000,
+        min_num_dims=1,
+        max_num_dims=1,
+        min_dim_size=2,
+        max_dim_size=2,
+    ),
+    shape=helpers.get_shape(
+        min_num_dims=2, max_num_dims=2, min_dim_size=1, max_dim_size=5
+    ),
```
KareemMAX, line 326: I think this is not covering the entire input space, can you change that?

KareemMAX (changes requested):
LGTM, just change the hypothesis strategy and we are good to go with the final tests

PR #14506, opened 2023-04-23
add jax.random.randint frontend

MuhammedAshraf2020:


KareemMAX (changes requested):
LGTM, just a minor fix and we are good to go with the final tests

KareemMAX (changes requested):
You didn't use `helpers.get_bounds()` please use them

PR #14516, opened 2023-04-24
Update layers.py

Chetangavali:
In interpolate function of tensorflow/experimental/layers.py file there is union with only one argument. In interpolate function's mode argument I think there is no need of union.


Chetangavali:
#14516

Chetangavali:
Sorry I accidentally closed this earlier.

RickSanchezStoic:
LGTM!

Chetangavali:
Thanks

PR #14524, opened 2023-04-24
Using paddle.assign for inplace_decrement paddle backend function

Sam-Armstrong:


MahmoudAshraf97 (changes requested):

ivy/functional/backends/paddle/general.py
```diff
@@ -228,7 +228,7 @@ def inplace_decrement(
     val: Union[ivy.Array, paddle.Tensor],
 ) -> ivy.Array:
     (x_native, val_native), _ = ivy.args_to_native(x, val)
-    x_native -= val_native
+    paddle.assign(x_native - val_native, x_native)
     if ivy.is_ivy_array(x):
         x.data = x_native
```
MahmoudAshraf97, line 233: Hi sam, there is a minor issue with this approach is that it doesn't support some dtypes
`[paddle.int8, paddle.uint8, paddle.float16, paddle.bool]`,
to check which dtypes are supported for a certain piece of code use this:
```python
dtypes_list = ["int8", "int16", "int32", "int64", "uint8", "float16", "float32", "float64", "complex64", "complex128", "bool",]
unsupported = []
for dtype in dtypes_list:
    try:
        a = paddle.to_tensor([[1.,2,3]],dtype=dtype)
        # insert the function you want to test in the following line
        a-a
    except:
        # use this to generate native dtype
        unsupported.append(a.dtype)
        # use this to generate string dtype
        #unsupported.append(dtype)
print(unsupported)
```
I suggest changing it to the following
```suggestion
    (x_native, val_native), _ = ivy.args_to_native(x, val)
    if ivy.is_ivy_array(x):
        target = x.data
    else:
        target = x
    with ivy.ArrayMode(False):
        return paddle.assign(ivy.subtract(x_native,val_native), target)
```
what is happening here is that i'm using an ivy function inside array mode which removes all the wrappers and conversions to ivy arrays and using only native functions, when you check the implementation of `ivy.subtract` you'll find that the unsupported dtypes are supported using casting

MahmoudAshraf97 (changes requested):

MahmoudAshraf97 (approved):

PR #14527, opened 2023-04-24
fix ivy.ones, ivy.zeros and ivy.empty to correctly unpack size when backend is not set

hello-fri-end:


ivy/functional/backends/jax/creation.py
```diff
@@ -95,10 +95,12 @@ def empty(
     device: jaxlib.xla_extension.Device,
     out: Optional[JaxArray] = None,
 ) -> JaxArray:
-    if size and shape:
-        raise TypeError("empty() got multiple values for argument 'shape'")
+    if len(size)!=0:
+        size = size[0] if isinstance(size[0], (tuple, list)) else size
+    if len(size)!=0 and shape:
+        raise TypeError("ones() got multiple values for argument 'shape'")
```
AnnaTz, line 101: You forgot to change the name of the functions in the error messages 😅

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -249,7 +249,6 @@ def test_asarray(
     )
 
 
-# empty
 # TODO: Fix container and instance methods
```
AnnaTz, line 252: I think this comment can be removed.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -751,7 +746,6 @@ def test_triu(
     ),
     dtype=helpers.get_dtypes("numeric", full=False),
     size=helpers.ints(min_value=1, max_value=3),
```
AnnaTz, line 748: It seems to me that the way you are using `size` to test is not correct. `size` is an integer, then you convert it to a 1-element tuple, and then you iterate through it to produce the `x{i}` which in practice will always be a single `x0`.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
hello-fri-end: I just verified this and it turns out you're right. The documentation for `helpers.ints` is misleading though as it says this helper function draws an arbitrarily sized list of integers. Replacing with `helpers.lists` should do the job though,

AnnaTz: Tbh you don't even need to generate a size. You could just get a non-None shape and a bool variable that will select whether to unpack or not.

hello-fri-end: Thanks for the suggestion. Have updated the tests now :)

AnnaTz (approved):

PR #14603, opened 2023-04-25
Added function 'cond' to tf_frontend with testing for #14575

AbdullahSabry:
#14575 
Hello,
I have a test unable to run on my machine for the numpy backend for the function 'cond', however all the other tests are successful. I believe there is something wrong with the backend, but not sure. Tell me what you think.

Thank you!

### DIFF
--- a/ivy/functional/frontends/tensorflow/general_functions.py
+++ b/ivy/functional/frontends/tensorflow/general_functions.py
@@ -68,6 +68,24 @@ def clip_by_norm(t, clip_norm, axes=None):
     return t_clip
 
 
+def cond(pred, true_fn=None, false_fn=None, name=None):
+    if true_fn is None:
+        raise TypeError("cond(): 'true_fn' argument required")
+    if false_fn is None:
+        raise TypeError("cond(): 'false_fn' argument required")
+
+    if not callable(true_fn):
+        raise TypeError("'true_fn' must be callable.")
+    if not callable(false_fn):
+        raise TypeError("'false_fn' must be callable.")
+
+    if pred:
+        return true_fn()
+
+    if not pred:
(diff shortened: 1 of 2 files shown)


AbdullahSabry:
I'm not really sure why 26 tests are failing from my 2 changed files. Did I do something wrong?

AbdullahSabry:
Hi @hirwa-nshuti 
I believe the problem with all these tests failing has to do with the fact that the function 'cond' from tensorflow is a duplicate from the function 'cond' from ivy.experimental.linear_algebra which has completely different characteristics. Meaning there is a conflict between them. 

I rechecked the docs on instructions on how to resolve this but couldn't find anything. It seems like the tests fail before even putting in the test for 'cond'; meaning just the decleration of it is causing problems. 

Can you check this out and tell me what you think?

Thanks!

ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -68,6 +68,11 @@ def clip_by_norm(t, clip_norm, axes=None):
     return t_clip
 
 
+@to_ivy_arrays_and_back
+def cond(pred, true_fn=None, false_fn=None, name=None):
+    return ivy.if_else(pred, true_fn, false_fn, ivy.array([]))
```
hirwa-nshuti, line 73: Ivy functional API doesn't have this function implemented you'll need to implement a composition for it

hirwa-nshuti (changes requested):
Hey @AbdullahSabry 

Thanks for contributing to ivy😊
There a few changes needed to be made on this PR.
Please check the comments below

AbdullahSabry:
Removed Changes temporarely

PR #14622, opened 2023-04-26
Add Cond to Pytorch frontend Linear Algebra #14591

### LINKED ISSUE #14591: cond

OmarJabri7:
Added Function cond with unit test for Linear Algebra Pytorch.
close #14591


ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -410,6 +410,25 @@ def test_torch_eigvalsh(
     )
 
 
+@handle_frontend_test(
+    fn_tree="torch.linalg.cond",
+    aliases=["torch.cond"],
```
hirwa-nshuti, line 415: This alias doesn't exist as well

hirwa-nshuti (changes requested):
Hey @OmarJabri7 

Thanks for the PR 😊
Some changes are needed check the comments below.

Thanks

OmarJabri7:
Hello @hirwa-nshuti , I applied some changes however I still see an assertion error np.nan!= np.inf when comparing ground truth.
Can that be due to the order of my parameters ?

OmarJabri7:
@hirwa-nshuti I think we can now:
close #14591

OmarJabri7:
close #14591

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -514,8 +555,8 @@ def test_torch_cross(
         frontend=frontend,
         test_flags=test_flags,
         fn_tree=fn_tree,
-        rtol=1e-2,
-        atol=1e-3,
+        rtol=1e-9,
```
hirwa-nshuti, line 558: Would you revert these changes

OmarJabri7: @hirwa-nshuti  sure, I will reput them. I asserted now in the actual function.
The problem is that I want to exclude non singular matrices.
What do you think about the new changes?

OmarJabri7: @hirwa-nshuti 
I fixed the issues, my tests are failing only on checks that are not related to my code changes at all.
What can I do in this scenario?
Can we close #14622  #14591

hirwa-nshuti (approved):
LGTM!
Thanks for contributing😊

PR #14632, opened 2023-04-26
potential fix for reverse mode addition with ivy.Shape

Aarsh2001:


Aarsh2001:
> Hi @Aarsh2001 ,Just a question to increase my knowledge in this matter ,but why do we require the reverse mode for ivy.shape?

This is because all the backends natively support reverse mode addition for their `shape` classes, and so should we 🙂. For eg-:

`(1,)` + some `torch.Size()` object/ `TensorShape` object, will produce a valid output.

ivy/__init__.py
```diff
@@ -221,13 +221,22 @@ def __repr__(self):
...
         except TypeError:
-            self._shape = self._shape + list(other)
+            if reverse:
+                self._shape = list(other) + self._shape
+            else:
+                self._shape = self._shape + list(other)
         return self
 
+    def __radd__(self, other):
+        return self.__add__(other, reverse=True)
```
VedPatwardhan, line 238: Let's just add the same logic in the reverse way, rather than adding a new argument to `__add__`. Would just make the code more readable 🙂

VedPatwardhan (changes requested):
Hey @Aarsh2001, just requested a small change. Rest looks good to go 🙂

ivy/__init__.py
Aarsh2001: sounds good

PR #14633, opened 2023-04-26
Frontend for tf.math.angle

### LINKED ISSUE #13752: angle

AwkNinja:
Close #13752

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
+++ b/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
@@ -2393,7 +2393,7 @@ def test_tensorflow_less(
 @handle_frontend_test(
     fn_tree="tensorflow.math.angle",
     dtype_and_input=helpers.dtype_and_values(
-        available_dtypes=["float64", "complex64" ,"complex128"],
+        available_dtypes=["float64", "complex64", "complex128"],
     ),
 )
 def test_tensorflow_angle(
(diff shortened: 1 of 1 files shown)


AwkNinja:
Hi @hmahmood24 I accidentally closed my last PR but continuing our conversation, I basically tried the `numeric` dtype for the test and it results in failing all the tests because they also contain integers. So as you mentioned we need to test the function on float and complex dtypes. I also found out that the np.angle and jnp.angle explicitly return a float64, so if we generate test data using float16 or float32 the tests for both tests frameworks fail. So what I have done right now is test with float64, complex64 and complex128. Currently the tests are passing. Let me know what you think.

AwkNinja:
For reference this is my last PR that I accidentally closed https://github.com/unifyai/ivy/pull/14434/

hmahmood24:
> Hi @hmahmood24 I accidentally closed my last PR but continuing our conversation, I basically tried the `numeric` dtype for the test and it results in failing all the tests because they also contain `integers`. So as you mentioned we need to test the function on `float` and `complex` dtypes. I also found out that the `np.angle` and `jnp.angle` explicitly return a `float64`, so if we generate test data using `float16` or `float32` the tests for both tests frameworks fail. So what I have done right now is test with `float64`, `complex64` and `complex128`. Currently the tests are passing. Let me know what you think.

@AwkNinja Right, that makes sense. Although the `integer` types should be properly flagged as unsupported in the `tensorflow` backend function for `angle`, thanks for letting me know! For now, this is fine. I'll go ahead and merge the PR now. Thanks so much for an amazing contribution! 🚀

PR #14644, opened 2023-04-27
Added multiply to torch frontend

guitld:


guitld:
This PR reffers to  issue #14573

ivy/functional/frontends/torch/tensor.py
```diff
@@ -991,3 +991,7 @@ def fill_(self, value):
 
     def nonzero(self):
         return torch_frontend.nonzero(self)
+
+    @with_unsupported_dtypes({"1.11.0 and below": ("bfloat16",)}, "torch")
+    def multiply(self, other, *, out=None):
+        return torch_frontend.multiply(self._ivy_array, other, out=out)
```
YushaArif99, line 997: I don't think there's a need to do `self._ivy_array` here. Simply calling the frontend function by passing `self` should work as the frontend function's decorators will take care of converting the inputs to and from `ivy arrays` to `frontend array` etc.

.idea/ivy.iml
YushaArif99, line 1: This file should not be modified

.idea/misc.xml
YushaArif99, line 1: same for this file as well

.idea/vcs.xml
YushaArif99, line 1: same for this file as well

YushaArif99 (changes requested):
Hi @guitld 

Thanks for making the PR. Could you kindly address the abovementioned points? Also, there are a couple of merge conflicts that will need to be resolved. Could you kindly also resolve them as this will allow the CI tests to run? Once all the relevant tests are passing, you can re-request a review 🙂

guitld:
Hi @YushaArif99 , just commited the changes you suggested! Thx for the review :)
I don't know why it's failing in the check formatting test though

YushaArif99 (changes requested):
HI @guitld 

The `.idea` files are still showing in the modified files. You can try using `git revert` to revert the commit that modified these files. If that doesnt work, one solution could be to reset your HEAD to point to the master branch using `git reset --hard origin`. This will reset your branch's HEAD to point to the master branch and so your local changes would get removed. Its best to keep a backup before you use `git reset`. Once this is done, you can then do the usual `git add->git commit->git push` to create a new commit. **NOTE:** its best to manually add the files using `git add <file path>` rather than `git add .` as the latter might add unwanted files (eg: the `idea` files) to the staging area.

Hope that helps 🙂

guitld:
Hi @YushaArif99

I think .idea is not modified anymore. Also, the conflicts are resolved.
Could you review it again please? :)

YushaArif99 (approved):
Hi @guitld 

Thanks for removing the `.idea` files. The CI tests are also passing now which means we can merge this PR 🙂

PR #14665, opened 2023-04-27
Adding the where argument

soma2000-lang:
I have tried to achieve the superset behavior for only a single elementwise (abs()) function.I will do the rest once this is done.The tests are currently failing ,which I am trying to fix.However it  will be great if @VedPatwardhan  can check the rest of the implemntation is correct or not.

Task link-https://trello.com/c/gqDVhX0M


VedPatwardhan (changes requested):
Hey @soma2000-lang, apologies for the delay! 
I don't think we should be implementing the `where` argument in this manner, so basically the function `abs` would look like,
```
def abs(
    x: Union[float, ivy.Array, ivy.NativeArray],
    /,
    *,
    where: Optional[ivy.Array] = None,
):
```
and `where` would be a boolean mask, such that we apply the function only to the values that are `True` in the boolean mask.
We shouldn't make any other changes to the `array` or `container` instance or static methods or even the `abs` function in `functional/ivy/elementwise.py` except for adding the `where` argument. We should instead implement the logic that makes use of the `where` mask in the backend-specific implementation through native functions of the framework if possible to maximize efficiency or using `ivy` functions otherwise. 
Happy to answer any other questions, thanks 🙂

soma2000-lang:
@VedPatwardhan Really Sorry,I got confused looking at the implementation of jax's abs() function using the where argument.

ivy/functional/backends/jax/elementwise.py
```diff
@@ -12,7 +12,13 @@
 from . import backend_version
 
 
-def abs(x: Union[float, JaxArray], /, *, out: Optional[JaxArray] = None) -> JaxArray:
+def abs(
+    x: Union[float, JaxArray],
+    /,
+    *,
+    out: Optional[JaxArray] = None,
```
VedPatwardhan, line 19: following the convention we've followed through the codebase, `where` should come before `out`. This applies to all backend implementations and the array and container instance and static methods.

ivy/functional/backends/jax/elementwise.py
```diff
@@ -12,7 +12,13 @@
...
+    where: Optional[ivy.Array] = None,
```
VedPatwardhan, line 20: Following the `numpy` convention, they've set the default value of `where` as `True`, hence it can't be typehinted as `Optional`. It should be `Union[bool, JaxArray]` instead of `Optional[ivy.Array]`

ivy/functional/backends/numpy/elementwise.py
```diff
@@ -12,8 +12,14 @@
...
 @_scalar_output_to_0d_array
 def abs(
-    x: Union[float, np.ndarray], /, *, out: Optional[np.ndarray] = None
+    x: Union[float, np.ndarray],
+    /,
+    *,
+    out: Optional[np.ndarray] = None,
+    where: Optional[Union[bool, np.ndarray]] = True,
 ) -> np.ndarray:
+    if where:
```
VedPatwardhan, line 21: I think we can just make one function call to `np.absolute`, given that it supports the `out` argument natively

ivy/functional/backends/tensorflow/elementwise.py
```diff
@@ -14,12 +14,15 @@ def abs(
...
     out: Optional[Union[tf.Tensor, tf.Variable]] = None,
+    where: Optional[Union[bool, tf.Tensor, tf.Variable]] = True,
 ) -> Union[tf.Tensor, tf.Variable]:
     if not tf.is_tensor(x):
         x = tf.convert_to_tensor(x)
     x_dtype = ivy.dtype(x)
     if any(("uint" in x_dtype, "bool" in x_dtype)):
         return x
+    if where:
+        return tf.abs(x, where=where)
```
VedPatwardhan, line 25: `tf.abs` doesn't accept the `where` argument natively, we'll have to use an `ivy.where` composition to implement that behaviour

ivy/functional/backends/torch/elementwise.py
```diff
@@ -521,11 +521,17 @@ def trunc(x: torch.Tensor, /, *, out: Optional[torch.Tensor] = None) -> torch.Te
...
+    x: Union[float, torch.Tensor],
+    /,
+    *,
+    out: Optional[torch.Tensor] = None,
+    where: Optional[Union[bool, torch.Tensor]] = True,
 ) -> torch.Tensor:
     x = _cast_for_unary_op(x)
     if x.dtype is torch.bool:
         return x
+    if where:
```
VedPatwardhan, line 533: `torch.abs` doesn't accept a `where` argument either

ivy/functional/frontends/jax/numpy/mathematical_functions.py
```diff
@@ -16,8 +16,8 @@ def sign(x, /):
 
 
 @to_ivy_arrays_and_back
-def absolute(x, /):
-    return ivy.abs(x)
+def absolute(x, /, where):
+    return ivy.abs(x, where=where)
```
VedPatwardhan, line 20: We shouldn't be changing the frontends in this PR, `jax.numpy.absolute` doesn't accept a `where` argument anyway

ivy/functional/ivy/elementwise.py
```diff
@@ -30,7 +30,8 @@ def abs(
     /,
     *,
     out: Optional[ivy.Array] = None,
-) -> ivy.Array:
+    where: Optional[ivy.Array] = False,
```
VedPatwardhan, line 33: The default value for `where` should be `True`

VedPatwardhan (changes requested):
Hey @soma2000-lang, thanks for making the changes. I've requested a few more, also the changes apply to all backend-specific implementations and array and container methods for `abs` where applicable. Thanks 🙂

ivy/functional/backends/numpy/elementwise.py
```diff
@@ -12,8 +12,14 @@
...
 @_scalar_output_to_0d_array
 def abs(
-    x: Union[float, np.ndarray], /, *, out: Optional[np.ndarray] = None
+    x: Union[float, np.ndarray],
+    /,
+    *,
+    where: Union[bool, np.ndarray] = True,
+    out: Optional[np.ndarray] = None,
 ) -> np.ndarray:
+    if where & np.absolute(x, out=out):
```
VedPatwardhan, line 21: `where` in this case wouldn't need to be in the `if` statement, `where` is supposed to indicate which elements we would convert to their absolute values rather than whether we would apply the function to the array or not

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -827,6 +831,8 @@ def abs(
         paddle.float16,
         paddle.bool,
     ]:
+        if where:
```
VedPatwardhan, line 834: same here, we should be doing something like this instead
```
ivy.where(where, paddle.abs(x.cast("float32")).cast(x.dtype), x)
```

ivy/functional/backends/tensorflow/elementwise.py
```diff
@@ -13,13 +13,16 @@ def abs(
...
     *,
+    where: Union[bool, tf.Tensor, tf.Variable] = True,
     out: Optional[Union[tf.Tensor, tf.Variable]] = None,
 ) -> Union[tf.Tensor, tf.Variable]:
     if not tf.is_tensor(x):
         x = tf.convert_to_tensor(x)
     x_dtype = ivy.dtype(x)
     if any(("uint" in x_dtype, "bool" in x_dtype)):
         return x
+    if where:
```
VedPatwardhan, line 24: same as above

VedPatwardhan (changes requested):
Hey @soma2000-lang, thanks for making the changes, just suggested a few more of them. Thanks for looking into this 🙂

ivy/functional/backends/jax/elementwise.py
```diff
@@ -12,7 +12,13 @@
 from . import backend_version
 
 
-def abs(x: Union[float, JaxArray], /, *, out: Optional[JaxArray] = None) -> JaxArray:
+def abs(
+    x: Union[float, JaxArray],
+    /,
+    *,
+    where: Union[bool, JaxArray] = None,
```
VedPatwardhan, line 19: The default value should be `True` instead of `None`, just like the other backend implementations

ivy/functional/backends/jax/elementwise.py
```diff
@@ -12,7 +12,13 @@
...
+    out: Optional[JaxArray] = None,
+) -> JaxArray:
```
VedPatwardhan, line 21: We should be making use of the `where` argument inside the implementation as the boolean mask applied

VedPatwardhan (changes requested):
Hey @soma2000-lang, thanks for making the changes and apologies for the delay in reviewing the PR. Just suggested a couple more and we're done with the `abs` function 🙂

ivy/functional/backends/jax/elementwise.py
```diff
@@ -12,11 +12,17 @@
...
+    x: Union[float, JaxArray],
+    /,
+    *,
+    where: Union[bool, JaxArray] = True,
+    out: Optional[JaxArray] = None,
+) -> JaxArray:
     if "bool" in str(x.dtype):
         return x
     # jnp.where is used for consistent gradients
-    return jnp.where(x != 0, jnp.absolute(x), 0)
```
VedPatwardhan, line 19: `x != 0` should still be used, we'd need an additional `ivy.where` for using the `mask`.
It could be something like,
```
ivy.where(where, jnp.where(x != 0, jnp.absolute(x), 0), x)`
```

VedPatwardhan (changes requested):
One last change, feel free to make a start on the other functions once that's done. Thanks @soma2000-lang 🙂

PR #14712, opened 2023-04-28
Fixes tf_gather 

akshatvishu:
Fixed `axis= None`  case as we can't perform the modulo operation.


xoiga123:
Hi @akshatvishu, great catch! Still, I have some questions :thinking:. I assume `axis=None` can only happen when you use tensorflow frontend, because it's defined as
```python
def gather(params, indices, axis=None, batch_dims=0, name=None):
    return ivy.gather(params, indices, axis=axis, batch_dims=batch_dims)
```

Now, we either have to fix all the backends (handle `axis=None` being passed in explicitly), or we can fix the tensorflow frontend like
```python
def gather(params, indices, axis=None, batch_dims=0, name=None):
    return ivy.gather(params, indices, axis=axis if axis else 0, batch_dims=batch_dims)
```
because it is mentioned in the [tensorflow docs](https://www.tensorflow.org/api_docs/python/tf/gather):
> gather can select slices across any axis depending on the axis argument (which defaults to 0)
You can test for this yourself:
```python
params = tf.constant([[0, 1.0, 2.0],
                      [10.0, 11.0, 12.0],
                      [20.0, 21.0, 22.0],
                      [30.0, 31.0, 32.0]])
tf.gather(params, indices=[3,1]).numpy()
# array([[30., 31., 32.],
#        [10., 11., 12.]], dtype=float32)
tf.gather(params, indices=[3,1], axis=0).numpy()
# array([[30., 31., 32.],
#        [10., 11., 12.]], dtype=float32)
```

akshatvishu:
Hey @xoiga123 ; This error was caught by  matt barrett and we discussed this in the [#ivy_frontends discord channel](https://discord.com/channels/799879767196958751/998782045494976522/1101096286293348412); and I think you're also correct that this case of `axis=None` will only arise when we use  use tensorflow frontend.Mostly , Matt highlighted that the behavior of  `ivy_tf_frontend_gather` was not matching the behavior of `tf_gather`  hence, I propose these changes !

xoiga123:
@akshatvishu Oh yea I see. In that case, isn't it better to modify the tensorflow frontend like I suggested, as it's something that can only occur when using the frontend? Consider your proposed changes:
```python
ivy.functional.backends.tensorflow.gather(tensor, indices, axis=None)  # this works
ivy.functional.backends.torch.gather(tensor, indices, axis=None)  # this fails -> inconsistency
```

akshatvishu:
> @akshatvishu Oh yea I see. In that case, isn't it better to modify the tensorflow frontend like I suggested, as it's something that can only occur when using the frontend? Consider your proposed changes:
> 
> ```python
> ivy.functional.backends.tensorflow.gather(tensor, indices, axis=None)  # this works
> ivy.functional.backends.torch.gather(tensor, indices, axis=None)  # this fails -> inconsistency
> ```

I thought i I've modified the `gather()` for tenor-flow only. Or does modifying `tf_gather` located at `backend/tensorflow/general.py` will cause the inconsistency as you mentioned ? If that the case then I will instead modify the `tf_gather()` at front-end instead of backend. Sorry for my naive questions as still trying to understand how Ivy works @xoiga123

xoiga123:
@akshatvishu Modifying the backend is still a valid solution and your code is good :+1:, but you **must** modify all the other backends (torch, jax, numpy) as well to keep consistency. So it's better to just modify the tensorflow frontend in my opinion.

akshatvishu:
> @akshatvishu Modifying the backend is still a valid solution and your code is good +1, but you **must** modify all the other backends (torch, jax, numpy) as well to keep consistency. So it's better to just modify the tensorflow frontend in my opinion.

Thank you so much for your feedback. I will then revert the changes made to the tf_backend file and instead do the changes at-the frontend. Also, should I push the changes in this PR only or shall I close this and open a new PR ? @xoiga123

xoiga123:
@akshatvishu Sorry I took some time between replies, yes you can push directly to this PR :muscle: Thank you

akshatvishu:
Hey @xoiga123  I added the code to handle `axis=None` case in `gather()` located at path `ivy/ivy/functional/frontends/tensorflow/general_functions.py`; is this the correct file to make the changes?

xoiga123:
LGTM! Thank you for contributing :hugs:

xoiga123 (approved):

akshatvishu:
> LGTM! Thank you for contributing hugs

Thank you for your review and approval!🙌🤗

PR #14757, opened 2023-04-30
tensorflow frontend random stateless_normal , poisson

### LINKED ISSUE #14532: stateless_normal

### LINKED ISSUE #14533: stateless_poisson

### LINKED ISSUE #14534: stateless_gamma

NouranYoussef:
closes #14533
closes #14532


NouranYoussef:
I have issues testing the stateless_gamma provided, the tf.random.stateless_gamma function specifies that :
alpha : Tensor. The concentration parameter of the gamma distribution. Must be broadcastable with beta, and broadcastable with the rightmost dimensions of shape.
beta: Tensor. The inverse scale parameter of the gamma distribution. Must be broadcastable with alpha and broadcastable with the rightmost dimensions of shape.

,However, when I pass a shape argument to ivy.gamma, I receive the error:

image

As the function expects alpha and beta to be numeric values when shape is specified.

image

and when i don't pass the input shape var to the wrapped ivy.gamma , i still got an error:

image

Daniel4078 (changes requested):
I think that this quote from the stateless_gamma's tensorflow doc may answer your question(https://www.tensorflow.org/api_docs/python/tf/random/stateless_gamma) that you will need to change how you generate the shape argument:
A slight difference exists in the interpretation of the shape parameter between stateless_gamma and gamma: in gamma, the shape is always prepended to the shape of the broadcast of alpha with beta; whereas in stateless_gamma the shape parameter must always encompass the shapes of each of alpha and beta (which must broadcast together to match the trailing dimensions of shape).
by the way, you should keep the shape argument in the version you committed so that the CI checks can show the error you found to me as well.

NouranYoussef:
> I think that this quote from the stateless_gamma's tensorflow doc may answer your question(https://www.tensorflow.org/api_docs/python/tf/random/stateless_gamma) that you will need to change how you generate the shape argument: A slight difference exists in the interpretation of the shape parameter between stateless_gamma and gamma: in gamma, the shape is always prepended to the shape of the broadcast of alpha with beta; whereas in stateless_gamma the shape parameter must always encompass the shapes of each of alpha and beta (which must broadcast together to match the trailing dimensions of shape). by the way, you should keep the shape argument in the version you committed so that the CI checks can show the error you found to me as well.

Thank you for your review. I've kept the shape argument in my latest commit. Could you please check it?

Daniel4078 (commented):
So I found that the tensorflow backend implementation of gamma(https://github.com/unifyai/ivy/blob/5e12db9bd4d9b4bef84e8a5c8e722e9db28c4651/ivy/functional/backends/tensorflow/experimental/random.py#L78) somehow requires alpha and beta to be ints not floats, which I think is wrong. Thus I think it is not your problem here that the tests fails and you should try report this problem to the functions/experimental discord channel.

NouranYoussef:
> So I found that the tensorflow backend implementation of gamma(
> 
> https://github.com/unifyai/ivy/blob/5e12db9bd4d9b4bef84e8a5c8e722e9db28c4651/ivy/functional/backends/tensorflow/experimental/random.py#L78
> 
> ) somehow requires alpha and beta to be ints not floats, which I think is wrong. Thus I think it is not your problem here that the tests fails and you should try report this problem to the functions/experimental discord channel.

ok I think i will make another separate pr for stateless_gamma once i manage to resolve the issues , i've remove stateless_gamma from this pr in my last update can you check it and let me now if this pr needs anything to change, or if you've any further feedback. thank you for your help

Daniel4078 (approved):
Looks good to me, hope that you can soon have a working gamma backend to work with.

PR #14760, opened 2023-04-30
Adding the  missing special methods

soma2000-lang:
Solves https://trello.com/c/QDPhbgM6

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -239,8 +239,59 @@ class Shape:
         self._shape = self._shape * other
         return self
 
+    def __rmul__(self, other):
+        self._shape = other * self._shape
+        return self
+
+    def __bool__(self):
+        return self._shape.__bool__()
+
+    def __div__(self, other):
+        return self._shape // other
+
+    def __mod__(self, other):
+        return self._shape % other
+
+    def __rdiv__(self, other):
+        return other // self._shape
+
+    def __rmod__(self, other):
+        return other % self._shape
+
+    def __reduce__(self):
+        return (self._shape,)
+
+    def as_dimension(self, other):
+        if isinstance(other, self._shape):
+            return other
+        else:
+            return self._shape
+
+    def __sub__(self, other):
+        try:
+            self._shape = self._shape - other
+        except TypeError:
+            self._shape = self._shape - list(other)
+        return self
+
+    def __rsub__(self, other):
+        try:
+            self._shape = other - self._shape
+        except TypeError:
+            self._shape = list(other) - self._shape
+        return self
+
     def __eq__(self, other):
         return self._shape == other
+
+    def __int__(self):
+        if hasattr(self._shape, "__int__"):
+            res = self._shape.__int__()
+        else:
+            res = int(self._shape)
+        if res is NotImplemented:
+            return res
+        return to_ivy(res)
 
     def __ge__(self, other):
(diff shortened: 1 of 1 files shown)


Aarsh2001 (commented):
Hey @soma2000-lang, we have updated the task a bit. Kindly have a look at the description, thanks 🙂

soma2000-lang:
Hi @Aarsh2001 Just  had some silly doubt.

1. Could not  find the native shape class for the torch.size.

2. We need to add the special methods to init py file right?

Aarsh2001 (commented):
Hey @soma2000-lang, you need to check the methods by printing `dir(torch.Size)` and `dir(tf.TensorShape)`. So it should output something like this(for torch) -:

Screenshot 2023-05-03 at 5 21 22 PM

I guess the way they implement `torch.Size` is a bit different, and it is implemented as a subclass of std::vector<int64_t>, where it overrides several methods to provide additional functionality for working with tensor sizes. You can find the base file [here](https://github.com/pytorch/pytorch/blob/main/torch/csrc/Size.cpp). But it don't think you need to refer it for the implementations. 

What you should do instead is -:
1. check for the intersection of `dunder` methods in `tf.TensorShape` and `torch.Size()`. This is priority number 1
2. implement all other property/class methods as supersets. So these methods would be which don't have a preceding or succeeding underscore. For example `numel` and `count` in the screenshot above. You can open another PR for this. 

Hope this clarifies, if not, feel free to reach out to me 🙂

soma2000-lang:
image

Aarsh2001:
> Hi @Aarsh2001 I have a doubt ,I cannot find reference to some dunder methods like '''**reduce_ex** ''' and ""**format**""" I am also taking reference from [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/tensor_shape.py)
> 
> For the property methods I will open another PR

Hey @soma2000-lang, the `__reduce_ex` method is used for serialising/deserialising objects. I don't think we need to implement this, cause we would be defaulting to the `NativeShape` class anyway. They have handled it in their implementation🙂. Here's an example I ran, which works for `ivy.Shape` -:

```
import ivy
import pickle

ivy.set_backend("tensorflow")
x = ivy.array([2., 3.])
y = ivy.Shape(x.shape)

with open('my_dict.pkl', 'wb') as f:
    # Serialize the dictionary to the file using pickle.dump()
    pickle.dump(y, f)
```
This can be extended  to other backends as well. I don't think we need the `__format__` method either, and a few others like `__init_subclass__`, `__subclasshook__`. I think this PR is a good merge 🙂, requesting a final review from @VedPatwardhan

ivy/__init__.py
```diff
@@ -263,6 +314,29 @@ def __getitem__(self, key):
...
         return len(self._shape) if self._shape is not None else 0
 
+    def __delattr__(self, item):
+        return super().__delattr__(item)
+
+    def __hash__(self):
+        return hash(self._shape)
+
+    def __setattr__(self, item):
+        return super().__setattr__(item)
```
Aarsh2001, line 324: Also, let's remove this, we don't need it

Aarsh2001 (changes requested):

VedPatwardhan (approved):
looks good to merge, thanks @soma2000-lang @Aarsh2001 🙂

PR #14786, opened 2023-05-01
Added Instance Method short to torch frontend

fahad12398:
#14779


Ookamice (changes requested):
Hiya, thank you for contributing. 😊

It appears there's a test that is failing due to bfloat16 using torch ground truth and jax backend. Could you try adding that to the @with_unsupported_dtypes decorator and see if that fixes it? 

Other than that, everything looks good 👍

fahad12398:
> Hiya, thank you for contributing. 😊
> 
> It appears there's a test that is failing due to bfloat16 using torch ground truth and jax backend. Could you try adding that to the @with_unsupported_dtypes decorator and see if that fixes it?
> 
> Other than that, everything looks good 👍

ok, will do that. Thank you.

fahad12398:
I have looked into the issue. The issue is with jax and jax backend return the value which is not consistent with other frameworks when the value is negative number and dtype is bfloat16. So, may be it is the problem with tests and that can be prevented by either passing 'valid' or 'integer'. or should i use `@with_unsupported_dtypes`.
The Code:
image

The Output:
image

Ookamice:
Hiya, I think the point your brought up is very good and the example given is super helpful. This appears to be a pretty common issue when I search around on the Ivy Discord Server.

For best practise, I believe using `valid` as the dtype is the way to go with unsupported dtype decorator used for data types that the backend doesn't support. Since I believe the jax backend does support bfloat16 but it is an underflow issue, I would be more inclined to add a min and max on the test case instead. However, feel free to play around with it as I am not 100% sure myself.

Other than that, everything looks good and once you make the changes, I'll get it merged 😄😄

fahad12398:
> Hiya, I think the point your brought up is very good and the example given is super helpful. This appears to be a pretty common issue when I search around on the Ivy Discord Server.
> 
> For best practise, I believe using `valid` as the dtype is the way to go with unsupported dtype decorator used for data types that the backend doesn't support. Since I believe the jax backend does support bfloat16 but it is an underflow issue, I would be more inclined to add a min and max on the test case instead. However, feel free to play around with it as I am not 100% sure myself.
> 
> Other than that, everything looks good and once you make the changes, I'll get it merged 😄😄

ok, I have already added min and max to the test case and the tests were passing. Now, There are conflicts in the files I have changed. So, Resolving them. After that, The PR can be merged.

fahad12398:
> > Hiya, I think the point your brought up is very good and the example given is super helpful. This appears to be a pretty common issue when I search around on the Ivy Discord Server.
> > For best practise, I believe using `valid` as the dtype is the way to go with unsupported dtype decorator used for data types that the backend doesn't support. Since I believe the jax backend does support bfloat16 but it is an underflow issue, I would be more inclined to add a min and max on the test case instead. However, feel free to play around with it as I am not 100% sure myself.
> > Other than that, everything looks good and once you make the changes, I'll get it merged 😄😄
> 
> ok, I have already added min and max to the test case and the tests were passing. Now, There are conflicts in the files I have changed. So, Resolving them. After that, The PR can be merged.

Done with the changes.

Ookamice:
LGTM!!😁

Thank you for your contributions.

PR #14789, opened 2023-05-01
Fixes dimension issue of linespace torch backend

usama-baloch:


ivy/functional/backends/tensorflow/creation.py
```diff
@@ -285,8 +285,10 @@ def linspace(
     if axis is None:
         axis = -1
     with tf.device(device):
-        start = tf.cast(tf.constant(start), dtype=dtype)
-        stop = tf.cast(tf.constant(stop), dtype=dtype)
+        start = tf.cast(start, dtype=dtype)
```
YushaArif99, line 288: isnt it better to simply to `tf.constant(tf.cast(start, dtype=dtype))` instead

YushaArif99 (changes requested):

PR #14791, opened 2023-05-01
Added handler_mixed_function decorator and fixed some issues in pad f…

usama-baloch:
1: Added the decorator handle_mixed_function on PyTorch and TensorFlow backend, it is working correctly and calling the native implementation whenever it fulfills the condition that I added in the handler parameter. 

2: I didn't add the decorator in Jax's backend implementation because the implementation is already checking all the conditions and applying the pad function which passes every corner case.

3: ivy's pad function was failing on some corner cases, so fixed the issue for that cases also and tested it again, Now working properly.


YushaArif99:
@AnnaTz could you kindly also take a look at this PR given that you've already worked on the `ivy.pad` function previously 🙂

ivy/functional/backends/torch/experimental/manipulation.py
```diff
@@ -328,3 +330,105 @@ def concat_from_sequence(
...
+                elem_val, elem_has_multiple_values = _check_tuple(elem)
+                if elem_has_multiple_values:
+                    has_multiple_values = True
+                size += elem_val
+            return size, has_multiple_values
+    else:
+        return 1, False
+
+
+def _check(*args, **kwargs):
```
YushaArif99, line 370: this logic seems to be duplicated. Is it possible to refactor it maybe in a helper file?

ivy/functional/backends/tensorflow/experimental/manipulation.py
```diff
@@ -271,3 +273,89 @@ def concat_from_sequence(
...
+        else:
+            pad = args[1]
+            inp = args[0]
+            if isinstance(pad, tuple):
+                if _check_dimension(inp, pad, mode):
+                    return True
+                else:
+                    return False
+            elif isinstance(pad, int):
+                return False
```
AnnaTz, line 342: Could you explain why you are making these checks for `constant_values` and `pad`? I was expecting that the only thing to check for tensorflow would be the `mode`.

ivy/functional/backends/torch/experimental/manipulation.py
```diff
@@ -328,3 +330,105 @@ def concat_from_sequence(
...
+            c = kwargs['constant_values']
+            val, cond = _check_tuple(c)
+            if cond is False:
+                return False
+            else:
+                kwargs['constant_values'] = val
+                return True
+        elif _can_handle_padding(inp.shape, pad, mode):
+            return True
+        else:
```
AnnaTz, line 398: I don't think this is the correct approach for the `constant_values` and `pad`. You don't need to check these in the decorator to determine whether the backend will be used. You can convert them to the format that torch is expecting inside the backend function. This way we can use the native function for all the torch supported mode cases.

AnnaTz (changes requested):

ivy/functional/backends/torch/experimental/manipulation.py
usama-baloch: to reduce the redundancy, I imported the helper function _check_tuple in TensorFlow backend from PyTorch backend

ivy/functional/backends/tensorflow/experimental/manipulation.py
usama-baloch: In the documentation of the Tensorflow, we also need to check the dimensions of specific modes (SYMMETRIC AND REFLECT) and if the dimensions of the pad meet the criteria that are given in the Tensorflow documentation then we can use native implementation. if it does not meet the criteria then it raises an error so that's why I added this logic that it should call the compositional implementation of the pad when the modes are symmetric or reflect and if it does not meet the criteria.  Here is the documentation of the tf.pad that explains this condition: [Click Here](https://www.tensorflow.org/api_docs/python/tf/pad)

ivy/functional/backends/torch/experimental/manipulation.py
usama-baloch: If I apply this to the implementation of the native function then it does not work properly I think because these checks of tuples and dimensions return boolean True and False to represent the handler whether to call the compositional form or native form. so if we implement this into the TensorFlow or torch backend implementation then how it will call the compositional function at that time into the native implementation? on the other hand, we also need to use native implementation whenever it needs, so if we add this logic into the native implementation then it can be the same as compositional implementation because if we do this then it looks like we are refactoring these dimensions to meet the criteria I guess, so how would it differ from that?

ivy/functional/ivy/experimental/manipulation.py
```diff
@@ -939,6 +947,12 @@ def _check_arguments(
         ],
         message="the provided mode is not supported",
     )
+
+    if type(pad_width) == int:
+        pad_width = (pad_width,)
+    if type(pad_width) != tuple:
+        pad_width = tuple(pad_width)
```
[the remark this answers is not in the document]
usama-baloch: Yes, you are right, I remove these checks because it handled by the _to_pairs.

ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_manipulation.py
```diff
@@ -551,7 +551,7 @@ def _pad_helper(draw):
             min_num_dims=1,
             min_value=-100,
             max_value=100,
-        ).filter(lambda x: x[0][0] not in ["float16", "bfloat16"])
+        ).filter(lambda x: x[0][0] not in ["float16", "bfloat16", "complex64", "complex128"])
```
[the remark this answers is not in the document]
usama-baloch: Sure, Now Added!

ivy/functional/ivy/experimental/manipulation.py
```diff
@@ -710,11 +710,19 @@ def _set_pad_area(padded, axis, width_pair, value_pair):
...
+            padded[left_slice] = value_pair[0]
     if width_pair[1] > 0:
         right_slice = _slice_at_axis(
             slice(padded.shape[axis] - width_pair[1], None), axis
         )
         padded[right_slice] = value_pair[1]
+        if isinstance(value_pair[1], ivy.Array):
+            padded[right_slice] = value_pair[1].data
+        else:
+            padded[right_slice] = value_pair[1]
```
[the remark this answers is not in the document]
usama-baloch: When I was testing the pad function on multiple test cases. sometimes it raises errors like this: 

self = ivy.array([0]), args = (dtype('uint8'),), kwargs = {}

    def __array__(self, *args, **kwargs):
        args, kwargs = args_to_native(*args, **kwargs)
>       return self._data.__array__(*args, dtype=self.dtype, **kwargs)
E       TypeError: __array__() got multiple values for argument 'dtype'

on these two lines:
`padded[right_slice] = value_pair[1]`
`padded[left_slice] = value_pair[0]`

I debugged this many times and tried to find the root cause of this error and check why it is happening. I find out that sometimes the _set_pad_area helper function gets the value_pair as ivy array (not most of the time because most of the time value_pair is just a list) and I don't know why it gets this sometimes but the reason for the error, when it gets ivy array, was that it already has dtype in the *args and it again assigning the dtype using self.dtype which raises an error of "multiple values of dtype arguments". on the other hand, when the value pair is list then it does not have multiple dtype so it works fine,

I think when the value_pair is ivy array then it has two dtypes when we run this line: padded[left_slice] = value_pair[0] or padded[right_slice] = value_pair[1], and it gets confused to add the first dtype or second one. So that's why I only fetch the data part of the ivy array which returns a native array and then we store it into padded and then it gets only one dtype.

Maybe we can find a better solution for this but at the moment this solution works fine and does not fail on these corner cases

ivy/functional/backends/tensorflow/experimental/manipulation.py
AnnaTz: Right, tensorflow always has to introduce weird checks and conditions 😅

ivy/functional/ivy/experimental/manipulation.py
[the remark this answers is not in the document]
AnnaTz: Thanks for explaining this! Could you give me an example set of inputs that cause this error. I'd like to have a look at it myself and push a fix directly to master. There might be a more elegant solution.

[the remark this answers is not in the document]
AnnaTz: Nevermind, I just got the error after running `test_pad` a couple more times.

ivy/functional/backends/tensorflow/experimental/manipulation.py
usama-baloch: yes, it has weird conditions 😅

usama-baloch:
Hi @AnnaTz, during the testing phase of the pad function in the torch and TensorFlow backend, I was getting some weird errors from linspace function (linsapce used in the _get_linear_ramps helper function in ivy compositional implementation). I debugged it and find out that linspace was not returning the output that we expect on some of the test cases and because of the wrong outputs from the linspace function, the pad function also returns the wrong output or raise an error during the _set_pad_area function. So here is one of the example that was causing error:

When the backend is torch:

Example:

The Input data was this in the pad function: 
        `ivy.array([[[0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0]],

                       [[0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0]],

                       [[0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0]],

                       [[0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0]],

                       [[0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0],
                        [0, 0, 0, 0, 0]]]),
        pad_width=1,
        mode="linear_ramp",
        stat_length=2,
        constant_values=0,
        end_values=0,
        reflect_type="even" `


Error: `could not broadcast input array from shape (1,7,6) into shape (1,6,7)`


The root problem came from linsapce function's helper function `linspace_helper`, the linspace should return (2,6,7) on this input that I mentioned above but it returns (2,7,6) which causes an error when it came to _set_pad_area helper function because we are storing data of value_pair which have a shape of (1,6,7) in this input and expects the same size from padded[left_slice/right_slice] which is not because it is (1,7,6). So after spending some time debugging this function, I find out that this line `res = torch.transpose(res, axis, -1)` is changing the shape in the wrong way (which means most of the time it does not work properly but sometimes works only).  So there are many inputs where it returns wrong shapes and we get broadcast error in pad testing, So I changed the logic and tested it multiple times and on these test cases also where the old logic fails, Now it works properly and the whole pad function also works well. 

On Tensorflow backend:
there was a minor issue in linspace function of tensorflow when it came to converting different inputs into TensorFlow tensors, It did not recognize the data type to choose when it tried to convert some inputs into tensor on this line: `start = tf.constant(start, dtype=dtype) or stop = tf.constant(stop, dtype=dtype)` because some times we have start and stop with different dtype and the dtype that we explicitly stored in tf.constant was different, so it does not understand which to choose, So before doing this I ensure that the dtype of the start or stop that we are converting must be same to the dtype that explicitly given

ivy/functional/backends/tensorflow/creation.py
```diff
@@ -285,6 +285,8 @@ def linspace(
     if axis is None:
         axis = -1
     with tf.device(device):
+        start = tf.cast(start, dtype=dtype)
+        stop = tf.cast(stop, dtype=dtype)
         start = tf.cast(tf.constant(start), dtype=dtype)
         stop = tf.cast(tf.constant(stop), dtype=dtype)
```
AnnaTz, line 291: I had already fixed this in the two lines below yours 😅 If the test passes with your fix, we should keep it instead of mine as it's simpler. If so, please commit it directly to master so we can keep this PR focused on pad.

AnnaTz (changes requested):

ivy/functional/backends/torch/experimental/manipulation.py
AnnaTz: It would be more efficient to call a helper function to convert the padding and the constant value to the correct form inside the backend and then call the native function. If we just call the compositional implementation everything will be done manually, which is much slower.

ivy/functional/backends/tensorflow/creation.py
usama-baloch: Sure, I'll commit this to the master 👍

ivy/functional/backends/torch/experimental/manipulation.py
usama-baloch: Torch only supports multi-dimensional padding forms in its backend, if we directly fetch the single dimension padding into torch, it raises an error on any mode. padding should be multi-dimensional in all the modes torch supports. it is really hard to convert the padding value into the correct dimension to call native implementation all time that's why I implemented this check helper function to check these padding forms and call the compositional when it torch backend does not support the constant and padding values. yes, it is less efficient than the native implementation because sometimes it calls compositional implementation but it also fulfills the work of the decorator to check whether to call the compositional or native implementation on some instructions. to check torch supports which kind of padding and constant values check this: [Click Here](https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html)

AnnaTz: Could you give me an example of what you mean by single dimension vs multi dimension padding? I'm pretty sure it is possible as I've done conversions like this for the torch backend of the convolution functions.

AnnaTz: Also, please sync with master to resolve the conflicts.

AnnaTz: Torch's pad function expects the pad widths in a different format than numpy, jax and tensorflow, but they can all support all dimensionality cases.
Here is an example:
The input array is 3D, torch gets a 4D padding tuple and numpy gets a 3D tuple. The results are equivalent.
```
import torch
import torch.nn.functional as F
import numpy as np

t4d = torch.empty(3, 3, 4)
p3d = (2, 1, 3, 3)
out = F.pad(t4d, p3d, "constant", 0)
print(out.shape)
# torch.Size([3, 9, 7])

t = np.empty((3, 3, 4))
p = ((0, 0), (3, 3), (2, 1))
out = np.pad(t, p, 'constant', constant_values=0)
print(out.shape)
# (3, 9, 7)
```

YushaArif99: Hi @AnnaTz. Apologies for stepping in so late 😅. Although the conversation has gotten quite lengthy, if I followed correctly, I think what @usama-baloch is referring to w.r.t. the limited support for multidimensional padding is relevant for the other 3 modes of torch(i.e. `reflect`, `replicate`, and `circular`). *constant padding* is supported for arbitrary dimensions as is mentioned in the documentation [here](https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html#:~:text=Constant%20padding%20is%20implemented%20for%20arbitrary%20dimensions.) which is why I think your example worked correctly. However, if the inputs are in a format that's not supported natively by torch, (for instance input is a 6-7 D tensor or the padding tuple is of odd length), in such cases, modifying the inputs to make it compatible with the native function whilst also guaranteeing identical results would become non-trivial. In such cases, it's better to resort to the compositional implementation, which is where the role of the `handle_mixed_functions` decorator comes in. 
  

@usama-baloch, with regards to the `_can_handle_padding` helper, it seems you're returning `True` when `mode=circular`. But shouldn't this be for `mode=constant` case? If I'm not mistaken, doesn't the `F.pad` function in torch impose some extra restrictions on the `pad` tuple and the `input` tensor for the other 3 modes so you cant always return `True` right?

usama-baloch: @YushaArif99 Yes, this is exactly what I'm trying to say.

@AnnaTz  @YushaArif99, with regards to the issue regarding the `_can_handle_padding` helper, I'm currently working on it and there are some cases that are failing.

ivy/functional/backends/torch/experimental/manipulation.py
```diff
@@ -358,3 +359,105 @@ def unique_consecutive(
...
+    if mode == 'circular':
+        return True
+    if isinstance(pad, int):
+        return False
+    for i in range(n_dims):
+        if not isinstance(pad[2 * i], int) or not isinstance(pad[2 * i + 1], int):
+            return False
+        if input_shape[i] + pad[2*i] + pad[2*i+1] < 0:
+            return False
+    return True
```
AnnaTz, line 379: To be more specific on the point I was trying to make about converting the padding to the correct format so we can call the native functions in more cases:
Here, if the padding is not a tuple of integers, the function returns False. However, if the padding is an integer or a tuple of tuples (as expected by ivy.pad) you can just convert it to a tuple of integers like the one torch is expecting (as shown in the example code I wrote). 
Have you tried debugging to see if the code ever gets inside the torch backend during ivy pad's test?

AnnaTz (changes requested):

ivy/functional/backends/torch/experimental/manipulation.py
usama-baloch: 1) I have made some changes that needed to be in the `_can_handle_padding` helper function of the pad function in the torch backend, I also added an extra check at the beginning of the `_can_handle_padding` helper function which checks `whether it is tuple of tuple then convert it into a tuple of integers`. 

2) I did not add the integer case that you suggested, the reason is that when I tested it then it didn't return the same value when we convert the integer into a tuple and apply the padding function.

3) I have one question about the implementation of the torch pad function: while I was testing some examples in `torch.nn.functional.pad` which is fully native and I find out that the output of the torch backend is different then the output of the numpy backend. 

Here is this example: 
Torch backend:
`inp = torch.tensor([[1., 2., 3.], [4., 5., 6.], [7., 8., 9.]])
pad = (1,1)
mode='constant' 
res = torch.nn.functional.pad(inp, pad, mode, value=0)`

res: 
`tensor([[0., 1., 2., 3., 0.],
        [0., 4., 5., 6., 0.],
        [0., 7., 8., 9., 0.]])`

Numpy Backend:

`array = np.array([[1., 2., 3.], [4., 5., 6.], [7., 8., 9.]])
pad_width = (1,1)
mode='constant'

res = np.pad(array, pad_width, mode)`

res: 

`array([[0., 0., 0., 0., 0.],
       [0., 1., 2., 3., 0.],
       [0., 4., 5., 6., 0.],
       [0., 7., 8., 9., 0.],
       [0., 0., 0., 0., 0.]])`

so, I find out in the docs of torch: [Docs](https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html) about torch applying padding in a bit different way, it has `pad = (padding_left, padding_right, padding_top, padding_bottom, padding_front, padding_back)padding_front,padding_back)`. additionally when I converted pad=(1,1) to pad=(1,1,1,1) in torch backend for testing purpose then it returns the same answer as numpy backend returns when the input's pad=(1,1) in numpy. 

Now I'm stuck on this point if we directly call the native implementation ( when it is needed to call the native one ) in the torch backend but it returns the answer which does not match the desirable output because of the internal functionality of the torch's pad function. then this becomes irrelevant. So, what can we do now? any suggestions? 

I think we should keep the compositional implementation for these test cases, it may be inefficient but it can provide users the answers they want, but what do you think about it @AnnaTz?

AnnaTz: It is fairly simple to convert the pad widths to torch's expected format inside the torch backend implementation. Please have a look at how our torch frontend does the inverse of this (converting pad widths from torch format to ivy/numpy format). A similar approach should be followed to handle the integer pad width cases.

ivy-seed:

ivy-leaves:

Thank you :hugs:

PR #14836, opened 2023-05-02
min_max bound strategies and refactoring of ints strategy use

Ishticode:
Three strategies have been created `min_max_bound`, `exclude_min_max` and the `combo min_max_bound_exclusion`.
These reduce the `min_value`, `max_value`, `exclude_min`, and `exclude_max` parameters down to one strategy.
These can be passed as values or strategies.
This PR refactors the helpers.ints strategy to use the `min_max_bound` strategy instead of `min_value` and `max_value`. Due to large refactoring required, others will be refactored in separate PR while giving us a chance to modify the approach.
I don't think use of a strategy should be forced for min and max values.


ivy_tests/test_ivy/helpers/hypothesis_helpers/number_helpers.py
```diff
@@ -13,6 +13,31 @@
...
+
+@st.composite
+def min_max_bound_exclusion(draw,
+                            min_value=None,
+                            max_value=None,
+                            exclude_min=True,
+                            exclude_max=True):
+    min_val, max_val = draw(min_max_bound(min_value=min_value, max_value=max_value))
+    min_ex, max_ex = draw(exclude_min_max(exclude_min=exclude_min, exclude_max=exclude_max))
+    return min_val, max_val, min_ex, max_ex
```
CatB1t, line 38: I think it's acceptable to stay with the older implemenation of `ints`, as those parameters are limited on what value that can generate, we don't have to add more complexity by using functions that just returns the `st.just` value of the paramter.

CatB1t (changes requested):

ivy_tests/test_ivy/helpers/hypothesis_helpers/number_helpers.py
Ishticode: Hi @CatB1t 
Thankyou for the review and I see your point. But I think this seems a bit needless complexity because almost all existing use of ints strategy had known values specified. Imagine a scenario where you want to generate axes values with the max not exceeding the dimension of arbitrary arrays generated by another shared strategy.  It will be possible to strategize the min max values based on the data being generated alongside ints as shared strategies. Although this can be done in current setting but without the parameter decreasing and I think these sub strategies might have more use than shown here in this PR.
Let me know what you think. 🙂

PR #14837, opened 2023-05-02
Consistent handling of devices

Dsantra92:
Some of the device handling was not handled uniformly. The changes in this pr can be summarized as:

1. `torch.tensor.to` will follow the original function argument more strictly.
2. The device handling like `None` value for device will be handled in the backend.
3. Tensorflow and jax can have `cpu` indexes given that the indexes are more than 0. Otherwise it's plain simple 'cpu'.
4. Indexed device with cpu like 'cpu:1' from jax will be converted to non-indexed version 'cpu' when used with `ivy.as_ivy_dev`.
5. There is a new framework function `ivy.is_native_device` similar to `ivy.is_native_dtype`.
6. The `ivy.to_native_device` and `ivy.to_ivy_device` throws error if the input is not correct instead of returning the unchanged value.
7. `None` is not treated as ivy.Device.

To be resolved: Where do we handle `cuda`? This is problem because `cuda` is a kind of valid device for torch. Most device conversion allow ivy.NativeDevice. Though this is the not the same as `torch.device('cuda')` but can still partly qualify as torch device. I would love to hear opinion on this.

These changes can break potentially break some other code!


ivy/__init__.py
```diff
@@ -77,7 +77,12 @@ def __new__(cls, dev_str):
         if dev_str != "":
             ivy.utils.assertions.check_elem_in_list(dev_str[0:3], ["gpu", "tpu", "cpu"])
             if dev_str != "cpu":
-                # ivy.assertions.check_equal(dev_str[3], ":")
+                ivy.assertions.check_equal(
```
VedPatwardhan, line 80: Rather than enforcing the use of device_id, can we instead assume it to be `0` when not passed? We should then be raising a warning saying that the device id wasn't passed in the input string, so is being considered as `0`.

VedPatwardhan (changes requested):
Hey @Dsantra92, thanks for looking into this, just requested a small change. After this change, let's also run a bunch of tests to see if it breaks something and get it merged. Thanks 🙂

ivy/functional/ivy/device.py
```diff
@@ -378,6 +378,39 @@ def as_native_dev(device: Union[ivy.Device, ivy.NativeDevice], /) -> ivy.NativeD
...
+
+    >>> import torch
+    >>> device = torch.device("cuda")
+    >>> device
+    device(type='cuda')
+
+    >>> ivy.is_native_dev(device)
+    True
+    """
+    return ivy.current_backend().is_native_device(device)
```
AnnaTz, line 411: Since you are adding a new function, I think you should also implement a test for it.

ivy/functional/ivy/device.py
```diff
@@ -378,6 +378,39 @@ def as_native_dev(device: Union[ivy.Device, ivy.NativeDevice], /) -> ivy.NativeD
...
+
```
AnnaTz, line 412: Also, the tests for the device-related functions you modified might have started failing now. You will probably need to update the assertions there to reflect the new expected behavior.

ivy/functional/backends/paddle/device.py
```diff
@@ -71,6 +71,12 @@ def as_native_dev(
         return paddle.fluid.libpaddle.CUDAPlace()
 
 
+def is_native_device(device: Place, /):
```
AnnaTz, line 74: If I'm not mistaken, the name should be `is_native_dev`.

AnnaTz (changes requested):

ivy/functional/backends/paddle/device.py
```diff
@@ -19,62 +21,104 @@
...
+def is_native_dev(device, /):
+    return isinstance(
+        device,
+        (
+            core.Place,
+            core.XPUPlace,
+            core.CPUPlace,
+            core.CUDAPinnedPlace,
+            core.CUDAPlace,
+            core.IPUPlace,
```
VedPatwardhan, line 104: As discussed with @djl11 yesterday, we're currently holding off the support for IPUs and XPUs, so `cpu`, `gpu` and `tpu` would be the only devices supported in `ivy` for now. Same applies to the `torch` backend as well

VedPatwardhan (changes requested):
Hey @Dsantra92, just a small change, rest looks good to go. The changes to the frontends also seem okay to me, but @AnnaTz is in a better position to review it. Thanks both 🙂

ivy/functional/backends/paddle/device.py
Dsantra92: I forgot this pr existed. We are not supporting ipu in Ivy, for supporting ivy we need to support it as an `ivy.Device`. We are just verifying if the backend ipu device is a device and allowing them to run properly in the native backend.

Dsantra92:
I don't think any of the tests are related specifially to device methods.

AnnaTz (approved):
Yeah I think the CI checks are having issues again. Generally the PR looks good to me and I think we can merge it.

ivy/functional/backends/jax/device.py
```diff
@@ -37,10 +38,14 @@ def dev(
     as_native: bool = False,
 ) -> Union[ivy.Device, jaxlib.xla_extension.Device]:
     if isinstance(x, jax.interpreters.partial_eval.DynamicJaxprTracer):
-        return ""
+        logging.warn(
```
vedpatwardhan, line 41: I think this should be an error right? If it fails silently and returns an empty string, that might be a bit hard to debug where we tried to access the device of `DynamicJaxprTracer`. Unless this is something native jax users expect, I think we should just raise an exception with the same message

vedpatwardhan (changes requested):
Hey @Dsantra92, thanks for making the changes, just one change (I'm not sure why we were returning `""` in the first place, would be good to dive a bit into the git blame). Rest seems good to go 🙂

ivy/functional/backends/jax/device.py
vedpatwardhan: Oh okay makes sense, let's add a `ToDo` comment here then, sounds good!

ivy/functional/backends/jax/device.py
```diff
@@ -50,53 +57,69 @@ def dev(
...
+    if is_native_dev(device):
         return device
-    dev_split = ivy.Device(device).split(":")
-    device = dev_split[0]
-    if len(dev_split) > 1:
-        idx = int(dev_split[1])
+    if isinstance(device, str):
+        dev_split = ivy.Device(device).split(":")
+        device = dev_split[0]
+        if len(dev_split) > 1:
```
vedpatwardhan, line 105: can we also get the arbitrary devices thing working as well as discussed for this function and other functions in the backends?

ivy/functional/backends/paddle/device.py
```diff
@@ -17,62 +19,104 @@
...
+    if device is not None:
+        native_dev = as_native_dev(device)
+        if not dev(x, as_native=True)._equals(native_dev):  # type: ignore
+            # TODO: check memory leak because ret is copying the original tensor
+            return paddle.to_tensor(x, place=native_dev, stop_gradient=x.stop_gradient)
+    return x
 
 
-def as_ivy_dev(device: Place, /):
+def as_ivy_dev(device, /):
```
vedpatwardhan, line 42: ig the typehint was removed by mistake?

vedpatwardhan (changes requested):
Thanks for making the changes @Dsantra92, we can get the PR merged after these are done based on the CI results 🙂🚀

ivy/functional/backends/paddle/device.py
vedpatwardhan: Yep sure, but we shouldn't allow it to run properly if device was of the `core.XPUPlace` type for e.g.

ivy/functional/backends/paddle/device.py
Dsantra92: No, not really. Paddle Place is not the base type for all devices. There are more device types like CPUPlace and CUDAPlace which are native devices. So Place isn't the class it supports.

ivy/functional/backends/paddle/device.py
Dsantra92: I mean it's a native device. We are just verifying if it's one, we are not supporting it in ivy or converting it.

ivy/functional/backends/jax/device.py
Dsantra92: I think that is more of a device placement issue, not a converting one. We shouldn't be handling arbitrary device logic in `to_ivy_device` or `to_native_device` but rather during device placement (which makes me wonder if https://github.com/unifyai/ivy/pull/16101 is a good case for this). Maybe a function `is_real_device` or `to_nearest_real_device` is more suitable for such operations.

vedpatwardhan: Oh okay, makes sense! Thanks for clarifying, yeah we can handle this, not necessarily in the device shifting decorators but later. I'm not sure about the utility of `is_real_device` or `to_nearest_real_device`, so probably it could just be a private helper 👍

ivy/functional/backends/paddle/device.py
vedpatwardhan: Oh okay, ig then we should be doing a `Union[]` of relevant place types as the type-hint here 🙂

ivy-leaves:

Thank you :hugs:

vedpatwardhan (approved):
lgtm! I haven't dived into the changes to the frontends, but the changes to the framework are good to merge. Thanks @Dsantra92 🙂

PR #14842, opened 2023-05-03
Adding tests for jax frontend's _src

Fayad-Alman:
Your task is to write tests for the functions of this frontend module https://github.com/unifyai/ivy/tree/master/ivy/functional/frontends/jax/_src.
It would be ideal to use hypothesis as much as possible, but in complex cases you will probably have to resort to plain assertions.


Fayad-Alman:
Hi @AnnaTz 
Could you check my PR and see why the first test is failing?
I understand that the flatten() function is receiving invalid arguments, but I'm not sure how to fix this.
Thanks!

AnnaTz:
> Hi @AnnaTz Could you check my PR and see why the first test is failing? I understand that the flatten() function is receiving invalid arguments, but I'm not sure how to fix this. Thanks!

The error is occurring inside the native jax function (the one we are using as ground truth). This is probably because you are feeding it the wrong type of arguments. Afaik these are the expected arguments
`tree`: The nested data structure to flatten.
`is_leaf`: A function that takes a node in the data structure and returns True if it is a leaf node or False otherwise. If None, then the default behavior is to consider any node that is not a container (i.e., not a tuple, list, or dictionary) to be a leaf node.

AnnaTz:
You might need to be a bit creative to test these tree-related functions. I don't think feeding the function with a simple array will be enough to properly test it. I would suggest testing on a more complex nested structure. For example, you could generate a dictionary of lists of tuples to be the `tree` and defining `is_leaf` to consider tuples as leafs.

Fayad-Alman:
@AnnaTz Mind taking a look at the new changes?

AnnaTz:
> @AnnaTz Mind taking a look at the new changes?

Could you please try creating a nested structure to test this function as I suggested above? @abdulasiraj has made some progress on this kind of testing. Maybe he could give you some hints.

Fayad-Alman:
@AnnaTz My commits have switched to a new PR: https://github.com/unifyai/ivy/pull/15381

PR #14860, opened 2023-05-03
eq/lt/gt/ge/le fix for handling tuple/list cases in ivy.Shape

Aarsh2001:


ivy/__init__.py
```diff
@@ -192,6 +192,15 @@ def can_cast(self, to):
         return can_cast(self, to)
 
 
+def shape_casting_helper(ivy_shape, other):
```
VedPatwardhan, line 195: Can this be implemented as a private static method of the `Shape` class instead?

VedPatwardhan (changes requested):
Hey @Aarsh2001, just a small change, rest looks good to go 🙂

VedPatwardhan (approved):
lgtm! Feel free to merge, thanks @Aarsh2001 🙂

PR #14891, opened 2023-05-04
Jax frontend - Implemented while_loop function and it's test.

### LINKED ISSUE #14613: while_loop

ayush0O7:
Close #14613


ayush0O7:
Hello Ivy team this is my second PR :-)

1. I have not used ivy.while_loop since it doesn’t satisfy type requirements for init_val. Ivy’s while_loop requires init_val to be an iterable while jax.lax.while_loop accepts init_val to be a scalar as well.
2. I have limited the range of min_value and max_value to prevent looping for an excessively large number of iterations. The range includes both positive and negative values.
3. For the test functions I am making sure that all the real values inside x are >=0.
4. The function any_negative_real checks whether ANY of the real parts are less than 0. I earlier tried things like (x<0).any() and (x.real()<0).any() but they were giving some or the other error hence I implemented the any_negative_real function.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_control_flow_operators.py
```diff
@@ -153,3 +153,51 @@ def _test_body_fn(x, y):
...
+                if isinstance(elem, (int, float)) and elem < 0:
+                    return True
+                elif isinstance(elem, complex):
+                    return False
+                elif isinstance(elem, (list, tuple)):
+                    if any_negative_real(elem):
+                        return True
+            return False
+
+        return any_negative_real(x)
```
ZiadAmerr, line 188:
```suggestion
    def _test_cond_fn(x):
        for elem in x:
            if isinstance(elem, (int, float)) and elem < 0:
                return True
            elif isinstance(elem, complex):
                return False
            elif isinstance(elem, (list, tuple)):
                if any_negative_real(elem):
                    return True
        return False
```

Minor refactor, do you think this is more organized? `any_negative_real(x)` is used only once, I don't see the point of wrapping it as a function and calling it once. If you do have a specific reason behind this, please let me know :)

ZiadAmerr (changes requested):

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_control_flow_operators.py
ayush0O7: any_negative_real(x) can be called multiple times, let me explain.

Since the number of dimensions x can have be upto 5(Refer : dtype_and_values function) so to access the values(int, float, complex) stored in x we need to recursively travel x until we reach a leaf node which contains the values(int, float, complex).

Basically, any_negative_real(x) is a function which recurses over a multi-dimensional array until it reaches the leaf node containing the values(int, float or complex). Upon reaching leaf nodes it checks whether any real number is negative. 

Example : Assume we have a 3-d array [[[1, 2], [3, 4]], [[5, 6],[7, 8]]]. To access the element "1" any_negative_real(x) would be called 3 times , removing one square bracket at each iteration. 

first call:         x = [[[1, 2], [3, 4]], [[5, 6],[7, 8]]]
second call :  elem = [[1, 2], [3, 4]]
third call :       elem = [1, 2] 

during third call values 1 , 2 both will be compared to 0.

If you have any questions then please tell me i will be happy to answer :-)

ZiadAmerr: Oh sorry, I did not notice the recursive call! My fault, great work :)

ayush0O7:
Hey @ZiadAmerr do you have any more suggestions ? Otherwise we can merge this.

ZiadAmerr (approved):

ZiadAmerr:
Great work, thanks @ayush0O7! :)

PR #14892, opened 2023-05-04
Fixing out argument handling for mixed functions

hello-fri-end:
https://trello.com/c/SOop47HS/1852-fix-the-out-argument-of-mixed-functions


hello-fri-end:
Hey @VedPatwardhan! Instead of creating another decorator just for adding the `mixed_function` attribute so that it gets added to the innermost function, we can simply modify the `handle_out_in_ivy` variable inside the `handle_out_argument` decorator to check if `fn`'s module has `ivy.functional.ivy` as a substring which would imply that it's a compositional function and can handle the out argument directly. Incase, the mixed partial function uses the primary function,  the `support_native_out` attribute will decide how to handle the out argument as usual. Please let me know what you think about this :)

VedPatwardhan (commented):
Let's discuss this on a call, as I think flagging a function as mixed explicitly would help users debug issues in their code. Thanks 🙂

hello-fri-end:
Hey @VedPatwardhan, I've updated all the functions as discussed yesterday. Could you please take a look at the changes? Once you approve these, I'll update the docs.

ivy/func_wrapper.py
```diff
@@ -777,7 +777,7 @@ def _infer_device(*args, device=None, **kwargs):
 
 def handle_out_argument(fn: Callable) -> Callable:
     handle_out_in_backend = hasattr(fn, "support_native_out")
-    handle_out_in_ivy = hasattr(fn, "mixed_function")
+    is_compositional_function = "ivy.functional.ivy" in fn.__module__
```
VedPatwardhan, line 780: it would be great if we had a smaller variable name here 😅

ivy/func_wrapper.py
```diff
@@ -962,34 +962,44 @@ def _wrap_function(
         # set attributes
         for attr in original.__dict__.keys():
             # private attribute or decorator
-            if attr.startswith("_") or hasattr(ivy, attr) or attr == "handles_out_arg":
+            if (
+                attr.startswith("_")
+                or hasattr(ivy, attr)
+                or (not compositional and attr == "mixed_function")
```
VedPatwardhan, line 968: given that we won't have the `mixed_function` attribute, let's remove this check

ivy/func_wrapper.py
```diff
@@ -962,34 +962,44 @@ def _wrap_function(
...
-            # for the handle_mixed_function decorator
-            if to_wrap != original:
-                to_wrap.compos = original
-            for attr in to_replace[compositional]:
-                setattr(original, attr, True)
+        # to_native_arrays_and_back must be added
+        # to the primary implementation (if it exists)
+        # of mixed functions.
+        if original != to_wrap:
+            if not (
```
VedPatwardhan, line 980: I think we could simply check if `inputs_to_ivy_arrays` has been applied rather than checking if the other 2 aren't applied, as every compositional implementation should have the `inputs_to_ivy_arrays` decorator

ivy/func_wrapper.py
```diff
@@ -962,34 +962,44 @@ def _wrap_function(
...
+                hasattr(original, "output_to_ivy_arrays")
+                and hasattr(original, "input_to_native_arrays")
+            ) and not hasattr(original, "to_native_arrays_and_back"):
+                # add as an attribute temporarily so that decorators can be applied
+                # in the correct sequence below
+                setattr(original, "to_native_arrays_and_back", True)
```
VedPatwardhan, line 986: Ideally, we should keep the `original` function unchanged (and not apply `setattr` or `delattr`, and just modify `to_wrap`

VedPatwardhan (changes requested):
Hey @hello-fri-end, just suggested a few changes, happy to discuss anything in detail. Thanks 🙂

ivy/func_wrapper.py
hello-fri-end: Sure, done :sweat_smile:

ivy/func_wrapper.py
hello-fri-end: Oops, had missed it. Have removed it now.

ivy/func_wrapper.py
hello-fri-end: Oh, you're right! If primary implementation exists ( i.e `to_wrap != original`)  and `original` has an `input_to_ivy_arrays` attribute then that means it's a mixed function. Have updated it now.

ivy/func_wrapper.py
hello-fri-end: Added the attribute in another way. Please take a look :)

VedPatwardhan (changes requested):
Hey @hello-fri-end, the changes look good to me, could you please update all related sections in the deep dive? Thanks 🙂

hello-fri-end:
> Hey @hello-fri-end, the changes look good to me, could you please update all related sections in the deep dive? Thanks slightly_smiling_face

Done, have updated the deepdive! Please let me know if I missed anything.

VedPatwardhan (commented):
Hey @hello-fri-end, `test_interpolate` seems to be failing on [run_tests (33)](https://github.com/unifyai/ivy/actions/runs/5042660329/jobs/9043525978?pr=14892#logs), could you please check if the failure is related to the changes made? Thanks 🙂

hello-fri-end:
> Hey @hello-fri-end, `test_interpolate` seems to be failing on [run_tests (33)](https://github.com/unifyai/ivy/actions/runs/5042660329/jobs/9043525978?pr=14892#logs), could you please check if the failure is related to the changes made? Thanks slightly_smiling_face

Hey @VedPatwardhan, interpolate fails on master as well with the same error. Sherry is currently looking into it.

VedPatwardhan (approved):
> > Hey @hello-fri-end, `test_interpolate` seems to be failing on [run_tests (33)](https://github.com/unifyai/ivy/actions/runs/5042660329/jobs/9043525978?pr=14892#logs), could you please check if the failure is related to the changes made? Thanks slightly_smiling_face
> 
> Hey @VedPatwardhan, interpolate fails on master as well with the same error. Sherry is currently looking into it.

Sounds good, feel free to merge the PR if that's the case, thanks @hello-fri-end 🙂

PR #14928, opened 2023-05-04
Added ivy.array lint in docstrings

KareemMAX:
This PR adds [`unifyai/lint-hook`](https://github.com/unifyai/lint-hook) bot to the linting configuration, in which it scans for `ivy.array()` in the output of a docstring example and formats it


ivy/data_classes/array/array.py
```diff
@@ -908,12 +908,20 @@ def __gt__(self, other):
         >>> z = x > y
         >>> print(z)
         {
-            a: ivy.array([[True, False, False],
-                          [False, False, False],
-                          [False, False, False]]),
-            b: ivy.array([[True, True, False],
-                          [False, False, False],
-                          [False, False, False]])
+            a: ivy.array(
```
VedPatwardhan, line 911: I think the array above was originally formatted correctly, given that all frameworks format it that way. Same applies to other similar instances.

VedPatwardhan (commented):
Not sure if the PR was intended to be reviewed, but just wanted to bring a case into consideration. Thanks @KareemMAX 🙂

CatB1t (changes requested):
Agree to @VedPatwardhan point, also we should ideally make sure when the linter is setup, that it can handle code and string arrays differenetly, so arrays existing in code should be formatted but not the ones in the docstring.

ivy-seed:

ivy-seed:

PR #14939, opened 2023-05-05
Multicuda

RickSanchezStoic:
Introduces directory-level installation of frameworks and corresponding dependencies to avoid dependency conflict.


RickSanchezStoic:
Dockerfile:

1. Directory Based framework and dependency installation enabled.
2. Switch to Debian: Buster from Ubuntu:20

DockerfileGOUMultiCuda:

1. GPU support with multiple cuda and cudnn versions enabled.
2. Switch to Debian:Buster from Ubuntu:20
3. Directory Based framework and dependency installation enabled.

PR #14941, opened 2023-05-05
Add torch.norm function

### LINKED ISSUE #14816: torch.norm

paulaehab:
Close #14816

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_torch/test_reduction_ops.py
+++ b/ivy_tests/test_ivy/test_frontends/test_torch/test_reduction_ops.py
@@ -838,3 +838,69 @@ def test_torch_unique(
         return_counts=return_counts,
         dim=axis,
     )
+
+
+@st.composite
+def _get_axis_and_p(draw):
+    force_tuple_axis = True
+
+    p = draw(st.sampled_from(["fro", "nuc", 1, -1, 2, -2]))
+    if p == "fro" or p == "nuc":
+        force_tuple_axis = True
+        min_axes_size = 2
+
+    else:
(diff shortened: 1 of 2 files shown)


ivy_tests/test_ivy/test_frontends/test_torch/test_reduction_ops.py
```diff
@@ -838,3 +838,69 @@ def test_torch_unique(
...
+    p, values = p_dtype_x_axis
+    input_dtype, x, axis = values
+
+    helpers.test_frontend_function(
+        input_dtypes=input_dtype,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        rtol=1e-01,
```
AnnaTz, line 900: Looks good to me! Only thing is the `rtol` 10% is very high. Isn't the test passing without this?

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_reduction_ops.py
paulaehab: @AnnaTz
 yes because when I change it it say the ground truth between frame works not equal but when I test it it give the same result I don't know why

AnnaTz: Then I guess there is some inconsistency between torch and our ivy functions. I don't think this can be fixed in this PR. I'd say you can merge this now.

AnnaTz (approved):

PR #14943, opened 2023-05-05
add data_format arg to ivy.MaxPool2D and ivy.AvgPool2D

zhumakhan:
Added data_format argument to ivy.MaxPool2D and ivy.AvgPool2D


zhumakhan:
Currenlty this has only lint checks failing and not related to changes made in this PR

PR #14950, opened 2023-05-05
Add torch.tesnor.fmod_ function

### LINKED ISSUE #14946: torch.Tensor.fmod_

paulaehab:
Close #14946


sherry30 (changes requested):
Hi Paula,

Looks good for the most part, but the instance methods in torch tensor ending with `_` like `fmod_` are all in place methods , you can checkout other similar inplace instance methods ending with `_` in this class to get an idea how they're implemented. 😄 

Thanks

paulaehab:
@sherry30 Sorry for this mistake I implement it but using torch_frontend.fmod because fmod function not implemented yet in torch.tensor to call it and Iadded TODO to implement this function

PR #14961, opened 2023-05-06
shortend minkowski funtion

raj921:


raj921:
#14961

RickSanchezStoic:
@raj921 Can you please try elaborating a bit on what issues this fixes and what exact changes you have made? Thanks!

raj921:
I reduced this code 17 loc to 8 loc  removing extra variables in the last line and directly returning the result and writing and reducing conditional loop in one line to remove extra eilf statement

RickSanchezStoic:
@raj921 seems like you have made too many changes to other files too which were probably not needed. Can you look into that and make sure that only the required file which I believe is `distance.py` gets changed? Thanks!

raj921:
yes i already deleted very file i changed expect distance.py

RickSanchezStoic:
Hey, @raj921 that's not exactly how it works. If you delete those files, they will be deleted from the repo too. What you can do is copy and save the changes that you have made (I guess in `distance.py` ) to somewhere else and then make a fresh clone from ivy and then create a new pr. Make sure that you add only those files that you have made changes to while making a commit. using `git add filename` . Let me know if you have any doubts.

PR #14973, opened 2023-05-06
Extend ivy functional API

YasCoMa:
It is related to the issue "Extend Ivy Functional API #3856"
- [x] nanmin
-  [x] nanmax

### DIFF
--- a/ivy/functional/backends/jax/statistical.py
+++ b/ivy/functional/backends/jax/statistical.py
@@ -135,6 +135,29 @@ def var(
     )
 
 
+def cov(
+    x: JaxArray,
+    /,
+    *,
+    y: Optional[JaxArray],
+    rowvar: Optional[bool] = True,
+    bias: Optional[bool] = False,
+    ddof: Optional[int] = None,
+    fweights: Optional[Union[int, Sequence[int]]],
+    aweights: Union[Sequence[int], Sequence[float]],
+) -> JaxArray:
+    ret = jnp.cov(
+        x,
+        y=y,
+        rowvar=rowvar,
+        bias=bias,
+        ddof=ddof,
+        fweights=fweights,
+        aweights=aweights,
+    )
+    return ivy.astype(ret, x.dtype, copy=False)
+
+
 # Extra #
 # ------#
 
--- a/ivy/functional/backends/jax/statistical.py
+++ b/ivy/functional/backends/jax/statistical.py
@@ -135,29 +135,6 @@ def var(
     )
 
 
-def cov(
-    x: JaxArray,
-    /,
-    *,
-    y: Optional[JaxArray],
-    rowvar: Optional[bool] = True,
-    bias: Optional[bool] = False,
-    ddof: Optional[int] = None,
-    fweights: Optional[Union[int, Sequence[int]]],
-    aweights: Union[Sequence[int], Sequence[float]],
-) -> JaxArray:
-    ret = jnp.cov(
-        x,
-        y=y,
-        rowvar=rowvar,
-        bias=bias,
-        ddof=ddof,
-        fweights=fweights,
-        aweights=aweights,
-    )
-    return ret
-
-
 # Extra #
 # ------#
 
--- a/ivy/functional/backends/jax/statistical.py
+++ b/ivy/functional/backends/jax/statistical.py
@@ -155,7 +155,7 @@ def cov(
         fweights=fweights,
         aweights=aweights,
     )
-    return ivy.astype(ret, x.dtype, copy=False)
+    return ret
 
 
 # Extra #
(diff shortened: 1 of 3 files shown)


zaeemansari70 (changes requested):
Hi @YasCoMa ,
Requested some changes :slightly_smiling_face:

ivy/functional/backends/jax/statistical.py
```diff
@@ -135,6 +135,29 @@ def var(
...
+    ret = jnp.cov(
+        x,
+        y=y,
+        rowvar=rowvar,
+        bias=bias,
+        ddof=ddof,
+        fweights=fweights,
+        aweights=aweights,
+    )
+    return ivy.astype(ret, x.dtype, copy=False)
```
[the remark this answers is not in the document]
YasCoMa: I just made the change and pushed it. thanks

ivy/functional/backends/jax/statistical.py
```diff
@@ -135,6 +135,29 @@ def var(
...
+    /,
+    *,
+    y: Optional[JaxArray],
+    rowvar: Optional[bool] = True,
+    bias: Optional[bool] = False,
+    ddof: Optional[int] = None,
+    fweights: Optional[Union[int, Sequence[int]]],
+    aweights: Union[Sequence[int], Sequence[float]],
+) -> JaxArray:
+    ret = jnp.cov(
```
zaeemansari70, line 149: As it is the frontend of a specific framework, which is `jax` in this case, the implementation should be only in `ivy`, which means that we can only use Ivy functions.

zaeemansari70 (changes requested):
@YasCoMa , Apoloogies I overlooked your PR :sweat_smile: , 
I thought that you implemented the backend mistakenly I requested some incorrect changes. 

I've requested the correct changes below :slightly_smiling_face: 
Do let me know if you have any question, happy to answer :slightly_smiling_face: 

Thanks!

ivy/functional/backends/jax/statistical.py
YasCoMa: No, previously you understood right! At first, i intended to implement the frontend cov function for jax, but i saw that there was no backend function for it. Then, i track the dependencies and built for my first pull the backend function for it. Late, after pr accepted and merged, i can implement the ivy "middle end" and then the respective frontend function.

zaeemansari70: @YasCoMa , This PR of yours is linked to a frontend issue `cov`, also the backend function does exist, you might me looking in the `statistical` submodule :slightly_smiling_face: 
kindly take a look at this https://github.com/unifyai/ivy/blob/27772879639877c2f1553ae5b5dfbebc8cfc71fe/ivy/functional/backends/jax/experimental/linear_algebra.py#L153

YasCoMa: Thanks, i really did not check the experimental section before creating a new one. I removed my implementation from the statistical in functional. And i implemented nanmin and nanmax that were not checked in the issue neither in the experimental/statistical

zaeemansari70 (commented):
@AnnaTz 
Do we need these in our frontends?
If yes, then we should ask her to implement these two separately in the respective backends via separate PR's.
Just need a suggestion from you on this
Thanks!

AnnaTz:
> @AnnaTz Do we need these in our frontends? If yes, then we should ask her to implement these two separately in the respective backends via separate PR's. Just need a suggestion from you on this Thanks!

Yes, I think it makes sense to add these to our api. `nanmean`, `nanmedian` and `nanquantiIe` are already in the `Extend Ivy Functional API` ToDo list. I'll add these two as well.
@YasCoMa please link this PR to the subtasks of the ToDo list I mentioned.

zaeemansari70 (commented):
@YasCoMa 
If you want to work on these functions you have to work on extending our Ivy's API and implementing them in our backends, a good suggestion would be to work with separate PR's for both of these functions :slightly_smiling_face: 

YasCoMa:
> @YasCoMa If you want to work on these functions you have to work on extending our Ivy's API and implementing them in our backends, a good suggestion would be to work with separate PR's for both of these functions slightly_smiling_face Thanks! slightly_smiling_face

When i send the pr it takes the whole code, how can i separate them in two?
The next step is to make them in the ivy or for the other backends?

zaeemansari70:
@YasCoMa , I'd suggest go through the deep dive section in the docs https://unify.ai/docs/ivy/overview/deep_dive.html to get a good idea how to implement the backend. Also, you have to make 2 separate PR's for both the functions, you have to add them in the backend experimental folder :slightly_smiling_face: 

Feel free to close this PR and make new ones.
Thanks! :+1: :slightly_smiling_face:

PR #14974, opened 2023-05-06
polydiv

sondertg:


sondertg:
close #13886

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_math.py
```diff
@@ -2748,6 +2748,40 @@ def test_jax_numpy_polyint(
...
 
 
+# polydiv
+@handle_frontend_test(
+    fn_tree="jax.numpy.polydiv",
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
+        num_arrays=2,
+        min_num_dims=1,
+        max_num_dims=1,
```
AnnaTz, line 2758: I think `u` and `v` can be multidimensional arrays, so the test should not only be checking for `max_num_dims=1`.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_math.py
sondertg: I removed `max_num_dims` limit but there is a problem with dtype or tolerance. I checked `polymul` , `polysub` and `polyadd`, followed exact logic of original jax implementation. Can't find why tests are failing.

AnnaTz: The error I see right now in the CI is due to the data type not being the correct one. You can solve this by casting the result.

ivy/functional/frontends/jax/numpy/mathematical_functions.py
```diff
@@ -624,6 +624,31 @@ def polyint(p, m=1, k=None):
...
+@with_unsupported_dtypes(
+    {"0.3.14 and below": ("float16",)},
+    "jax",
+)
+@to_ivy_arrays_and_back
+def polydiv(u, v, *, trim_leading_zeros=False):
+    dtypeu = u.dtype
+    dtypev = v.dtype
+    u = ivy.asarray(u)
+    v = ivy.asarray(v)
```
AnnaTz, line 636: I see in the CI checks there is an error caused by these `ivy.asarray` calls. Could you explain why you think they are needed? `u` and `v` are already converted to ivy arrays in `@to_ivy_arrays_and_back`.

ivy/functional/frontends/jax/numpy/mathematical_functions.py
```diff
@@ -624,6 +624,31 @@ def polyint(p, m=1, k=None):
...
+    scale = 1. / v[0]
+    q = ivy.zeros(max(m - n + 1, 1), dtype=u.dtype)
+    for k in range(0, m - n + 1):
+        d = scale * u[k]
+        q[k] = d
+        u[k:k+n+1] += (-d * v)
+    if trim_leading_zeros:
+        u = trim_zeros(u, trim='f')
+    q = ivy.asarray(q, dtype=dtypev)
+    u = ivy.asarray(u, dtype=dtypeu)
```
AnnaTz, line 648: Same here. `q` and `u` are already ivy arrays.

ivy/functional/frontends/jax/numpy/mathematical_functions.py
```diff
@@ -624,6 +624,31 @@ def polyint(p, m=1, k=None):
...
+    return q, u
+
```
AnnaTz, line 650: There is also an assertion error in the CI checks indicating something is not right in the polydiv calculation. Please look into that. I would suggest running the test locally as it will help make debugging faster.

AnnaTz (changes requested):

ivy/functional/frontends/jax/numpy/mathematical_functions.py
sondertg: I added those to fix an issue with tensorflow. It was returning two different dtypes but I removed them and realized it was my mistake. I'll check logic but as far as I remember, just followed the original. Thanks for explanations.

ivy/functional/frontends/jax/numpy/mathematical_functions.py
sondertg: Yes, I thought it is related to tolerance or dtypes(`[1] != [-0.0]`). I am working on it.

sondertg: I couldn't find `polydiv` at other backends. I checked again and can't find any problem with logic. However, I need to check which functions be called from different frameworks so that I can tell the problem. (their results are in different formats 
```py
import numpy as np
import tensorflow as tf
import torch as pt

# my implementation
def polydiv(u, v, *, trim_leading_zeros=False):
    m = len(u) - 1
    n = len(v) - 1
    scale = 1. / v[0]
    q = np.zeros((max(m - n + 1, 1),), dtype=u.dtype)
    for k in range(0, m - n + 1):
        d = scale * u[k]
        q[k] = d
        u[k:k+n+1] -= (d * v)
    if trim_leading_zeros:
        u = np.trim_zeros(u, trim='f')
    return q, u


x = np.array([-1.0,-1.0], dtype='float64')
y = np.array([-1.0,-1.0], dtype='float64')

a = pt.tensor([-1.0,-1.0])
b = pt.tensor([-1.0,-1.0])

print(polydiv(x, y)) # quotient, remainder
print(pt.div(a,b)) # quotient
tf.divide(x,y) # remainder
```
Result from above code:
```
(array([1.]), array([0., 0.]))
tensor([1., 1.])
tf.Tensor([-0. -0.], shape=(2,), dtype=float64)
```

AnnaTz: There is no point in comparing results of functions in other frameworks. This is a frontend function. It should only be compared to the corresponding original one, in this case `jax.numpy.polydiv`.

sondertg: it's weird. If I didn't totally misunderstand the `polydiv` function, jax returning wrong result. I wasn't able to install jax but tried it on google colab and it seems it returns wrong result.

Reference:
https://numpy.org/doc/stable/reference/generated/numpy.polydiv.html

Our failure:

(-x-1 ) / (-x-1) 

This is 1. Am I missing something?


firefox_j9iAqx8rwD

AnnaTz: I too would expect the result to be 1.0 with 0 remainder. This is strange.

AnnaTz: I would suggest the following:
1. Check if `numpy.polydiv` returns the same result. `jax.numpy.polydiv` is supposed to mimic `numpy.polydiv` , so if the result is different, it would be a hint that there is a bug in jax's source code.
2. Copy-paste jax's (and maybe numpy's) source code and run the debugger step-by-step through those functions and your function. This way we can investigate where the different result stems from.

sondertg: `numpy.polydiv` returns zero too as quotient.  I just submitted a bug report at numpy repository (Probably won't make any difference for us.). I'll debug it but even if we find the problem, how can we proceed? Isn't our implementation the one which returns correct result?

sondertg: Let me try to explain what I did. First of all I realized the code I tried on colab has a problem. When I return `u` with my `polydiv` implementation, it changes the original input and that's why `np.poydiv` or `jax.polydiv` returns wrong result if I run them with my implementation.
However, something similiar happens inside our tests.  Now I'll explain why I think that. This is my implementation:
```py
@with_unsupported_dtypes(
    {"0.3.14 and below": ("float16",)},
    "jax",
)
@to_ivy_arrays_and_back
def polydiv(u, v, *, trim_leading_zeros=False):
#    if (u == ivy.array([-1.])).all() and (v == ivy.array([-1.])).all():
#        return [ivy.array([-0.], dtype=u.dtype), ivy.array([0.], dtype=u.dtype)]
#    if (u == ivy.array([-1., -1.])).all() and (v == ivy.array([-1., -1.])).all():
#        return [ivy.array([-0.], dtype=u.dtype), ivy.array([0., 0.], dtype=u.dtype)]
    u_arr, v_arr = ivy.promote_types_of_inputs(u, v)
    m = len(u_arr) - 1
    n = len(v_arr) - 1
    scale = 1. / v_arr[0]
    q = ivy.zeros((max(m - n + 1, 1),))
    for k in range(0, m - n + 1):
        d = scale * u_arr[k]
        q[k] = d
        u_arr[k:k+n+1] -= d * v_arr
    if trim_leading_zeros:
        u_arr = ivy.trim_zeros(u_arr, trim='f')
    return q, u_arr
```
I just didn't change input arrays so I can exclude that failure. However this didn't work.
input for u and v:
```
E            [array([-1.], dtype=float32), array([-1.], dtype=float32)]),
```
and the error:
```
E       len(ret_np_flat) != len(ret_np_from_gt_flat):
E       
E       ret_np_flat:
E       
E       [array([1.])]
E       
E       ret_np_from_gt_flat:
E       
E       [array([-0.], dtype=float32), array([0.], dtype=float32)]
```

If I am not wrong, our implementation returns 1 as expected. Than I wrote the part commented out at above. If we run test with removing `#` for first if block:
```
E           AssertionError:  the results from backend numpy and ground truth framework jax do not match
E            [-0.]!=[1.] 
```
Now test complains our result is wrong.


Also, I tried to vary tolerances and min/max values. Moreover, I debugged by breakpoints and used `print` s for logging. 

```py
x = 0
@with_unsupported_dtypes(
    {"0.3.14 and below": ("float16",)},
    "jax",
)
@to_ivy_arrays_and_back
def polydiv(u, v, *, trim_leading_zeros=False):
    global x
    x += 1
    print(x)
    print(f"input arrays u: {u} and v: {v}")
    u_arr, v_arr = ivy.promote_types_of_inputs(u, v)
    m = len(u_arr) - 1
    n = len(v_arr) - 1
    print(f"promoted input arrays u_arr: {u_arr} and v_arr: {v_arr}")
    scale = 1. / v_arr[0]
    q = ivy.zeros((max(m - n + 1, 1),), dtype=u_arr.dtype)
    for k in range(0, m - n + 1):
        d = scale * u_arr[k]
        q[k] = d
        u_arr[k:k+n+1] -= (d * v_arr)
    if trim_leading_zeros:
        u_arr = trim_zeros(u_arr, trim='f')
    print(f"output arrays q: {q} and u_arr {u_arr}")
    return q, u_arr
```

This shows me that my function works correct.  Logs are below from numpy:
```
1
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
input arrays u_arr: [-1.] and v_arr: [-1.]
output arrays q: ivy.array([1.]) and u_arr [0.]
2
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
input arrays u_arr: [-1.] and v_arr: [-1.]
output arrays q: ivy.array([1.]) and u_arr [0.]
3
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
input arrays u_arr: [-1.] and v_arr: [-1.]
output arrays q: ivy.array([1.]) and u_arr [0.]
4
input arrays u: ivy.array([-1.]) and v: ivy.array([-2.])
input arrays u_arr: [-1.] and v_arr: [-2.]
output arrays q: ivy.array([0.5]) and u_arr [0.]
5
input arrays u: ivy.array([-1.]) and v: ivy.array([-2.])
input arrays u_arr: [-1.] and v_arr: [-2.]
output arrays q: ivy.array([0.5]) and u_arr [0.]
6
input arrays u: ivy.array([-1.]) and v: ivy.array([-2.])
input arrays u_arr: [-1.] and v_arr: [-2.]
output arrays q: ivy.array([0.5]) and u_arr [0.]
7
input arrays u: ivy.array([-1.]) and v: ivy.array([-2.11664224e-06])
input arrays u_arr: [-1.] and v_arr: [-2.1166422e-06]
output arrays q: ivy.array([472446.4062]) and u_arr [0.]
8
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
input arrays u_arr: [-1.] and v_arr: [-1.]
output arrays q: ivy.array([1.]) and u_arr [0.]
9
input arrays u: ivy.array([-1.]) and v: ivy.array([-257.])
input arrays u_arr: [-1.] and v_arr: [-257.]
output arrays q: ivy.array([0.00389105]) and u_arr [0.]
10
input arrays u: ivy.array([-257.]) and v: ivy.array([-1.])
input arrays u_arr: [-257.] and v_arr: [-1.]
output arrays q: ivy.array([257.]) and u_arr [0.]
11
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
input arrays u_arr: [-1.] and v_arr: [-1.]
output arrays q: ivy.array([1.]) and u_arr [0.]
12
input arrays u: ivy.array([-1.]) and v: ivy.array([1.])
input arrays u_arr: [-1.] and v_arr: [1.]
output arrays q: ivy.array([-1.]) and u_arr [0.]
13
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
input arrays u_arr: [-1.] and v_arr: [-1.]
output arrays q: ivy.array([1.]) and u_arr [0.]
14
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
input arrays u_arr: [-1.] and v_arr: [-1.]
output arrays q: ivy.array([1.]) and u_arr [0.]
15
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
input arrays u_arr: [-1.] and v_arr: [-1.]
output arrays q: ivy.array([1.]) and u_arr [0.]
```

Lastly, I think problem is related to the loop I used. Below part can't return anything from `jax` tests. The original implementation of the function uses these:
https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html

I changed that block:
```py
  for k in range(0, m-n+1):
    d = scale * u_arr[k]
    q = q.at[k].set(d)
    u_arr = u_arr.at[k:k+n+1].add(-d*v_arr)
```
to
```py
    for k in range(0, m - n + 1):
        d = scale * u_arr[k]
        q[k] = d
        u_arr[k:k+n+1] -= (d * v_arr)
```

`jax` part of tests returns the `TypeError` below:
```
>     raise TypeError(msg.format(type(self)))
E     TypeError: '<class 'jaxlib.xla_extension.ArrayImpl'>' object does not support item assignment. JAX arrays are immutable. Instead of ``x[idx] = y``, use ``x = x.at[idx].set(y)`` or another .at[] method: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html
```

Both of these two approaches do the same thing however, jax arrays are immutable so indexing doesn't work.
For reference, original implementation:
```py
@_wraps(np.polydiv, lax_description=_LEADING_ZEROS_DOC)
def polydiv(u: ArrayLike, v: ArrayLike, *, trim_leading_zeros: bool = False) -> Tuple[Array, Array]:
  check_arraylike("polydiv", u, v)
  u_arr, v_arr = promote_dtypes_inexact(u, v)
  m = len(u_arr) - 1
  n = len(v_arr) - 1
  scale = 1. / v_arr[0]
  q: Array = zeros(max(m - n + 1, 1), dtype = u_arr.dtype) # force same dtype
  for k in range(0, m-n+1):
    d = scale * u_arr[k]
    q = q.at[k].set(d)
    u_arr = u_arr.at[k:k+n+1].add(-d*v_arr)
  if trim_leading_zeros:
    # use the square root of finfo(dtype) to approximate the absolute tolerance used in numpy
    u_arr = trim_zeros_tol(u_arr, tol=sqrt(finfo(u_arr.dtype).eps), trim='f')
  return q, u_arr
```

Note: I am not testing `trim_leading_zeros` just to investigate problem with a simpler aspect.

AnnaTz: Thanks for explaining in detail!
So, there is the error that the ground truth returns 0 instead of 1 but when you manually set the code to return 0 in that case there is an error that the ground truth returns 1? 😅 Could this be related to rounding depending on the data type?
About the item assignment error, in which function does this occur? I guess `u_arr` is an ivy array, so the error must be coming up in the jax backend of some ivy function.
Also why do you think the for loop is in blame for the incorrect returned values? Did the assertion error get fixed for the rest of the backends when you changed that assignment line?

sondertg: > Thanks for explaining in detail! So, there is the error that the ground truth returns 0 instead of 1, but when you manually set the code to return 0 in that case the error says that the ground truth returns 1? 😅 Could this be related to rounding depending on the data type? 

- Exactly. Since I thought this could be a problem, I played with min/max values and tolerances. Varied them to both edges ( decresed and increased tolerances/min, max values). I am not sure if this enough to eliminate that possibility.

> About the item assignment error, in which function does this occur? I guess `u_arr` is an ivy array, so the error must be coming up in the jax backend of some ivy function.

- While implementing those functions I usually fix errors from `numpy` parts without checking rest of tests. And once I solve errors on there, all tests are usually passing. For this case, I just realized `jax` part fails at each case because of the reason explained below. I don't know which Ivy function throws the error. Probably some fancy wrapper on there that I have no idea what's happening behind. 🤣 Here's the stack info from  the error:
```
test_jax_numpy_math.py:2775: in test_jax_numpy_polydiv
../../helpers/function_testing.py:671: in test_frontend_function
    ret = get_frontend_ret(
../../helpers/function_testing.py:2002: in get_frontend_ret
    ret = frontend_fn(*args, **kwargs)
/ivy/ivy/functional/frontends/jax/func_wrapper.py:102: in _outputs_to_frontend_arrays_jax
    ret = fn(*args, **kwargs)
/ivy/ivy/functional/frontends/jax/func_wrapper.py:79: in _inputs_to_ivy_arrays_jax
/ivy/ivy/functional/frontends/jax/numpy/mathematical_functions.py:645: in polydiv
    u_arr[k:k+n+1] -= (d * v_arr)
```

> Also why do you think the for loop is in blame for the incorrect returned values? Did the assertion error get fixed for the rest of the backends when you changed that assignment line?

- Because jax part of the tests never outputs with my logging implementation.(Logs below)  At item assignment part they throw the error I mentioned. Jax arrays are immutable. Jax implementation of the `polydiv` function uses that approach:
https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html
I also imported jax and used `at` inside the loop but than our function returns nothing. (empty array)
So, to solve this, I probably need to use inplace updates but don't know where to look at. As far as I remember our framework also has that feature, I'll check docs when I am able to.
```
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
promoted input arrays u_arr: [-1.] and v_arr: [-1.]
17
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
promoted input arrays u_arr: [-1.] and v_arr: [-1.]
18
input arrays u: ivy.array([-1.]) and v: ivy.array([-1.])
promoted input arrays u_arr: [-1.] and v_arr: [-1.]
19
.
.
.
```

If you have anything that can lead me to a right direction, I can follow. Thanks!

ivy/functional/frontends/jax/numpy/mathematical_functions.py
```diff
@@ -623,6 +623,25 @@ def polyint(p, m=1, k=None):
...
+    n = len(v) - 1
+    scale = 1. / v[0]
+    q = ivy.zeros((max(m - n + 1, 1),), dtype=u.dtype)
+    for k in range(0, m - n + 1):
+        d = scale * u[k]
+        q[k] = d
+        u[k:k+n+1] -= (d * v)
+    if trim_leading_zeros:
+        u = trim_zeros(u, trim='f')
+    return q, u
```
AnnaTz, line 642: Hi @sondertg, I wasn't sure what else to suggest so I debugged this myself locally and I found out what's wrong.
When you assign new values to `u`, the `args`/`kwargs` inside `test_frontend_function` (essentially the `u` and `v` variables that get fed to your function) also get updated. So eventually the arguments passed to the ground truth call are not the same as the ones passed to the frontend call, thus the results are not the same.
I'm not sure if this is expected behavior for our testing routine, as I haven't seen this happening before. I will look into this further.
Long story short, the problem goes away if you do `r = ivy.copy_array(u)` and then apply the new value assignments and zero trimming on `r`.

AnnaTz (changes requested):

ivy/functional/frontends/jax/numpy/mathematical_functions.py
sondertg: Hi @AnnaTz , thanks for your help. Now, original `numpy` implementation does not contain `trim_leading_zero`. Jax version somehow has control over tolerance of that array passed to the function. After searching a bit, if we can't control tolerance of our 'trim_leading_zero' function, we can't pass those tests. Jax version uses this:
```py
trim_zeros_tol(u_arr, tol=sqrt(finfo(u_arr.dtype).eps), trim='f')
# and implementation of that function at the source code of jax:
def trim_zeros_tol(filt, tol, trim='fb'):
  filt = core.concrete_or_error(asarray, filt,
    "Error arose in the `filt` argument of trim_zeros_tol()")
  nz = (ufuncs.abs(filt) < tol)
  if reductions.all(nz):
    return empty(0, _dtype(filt))
  start = argmin(nz) if 'f' in trim.lower() else 0
  end = argmin(nz[::-1]) if 'b' in trim.lower() else 0
  return filt[start:len(filt) - end]
```
Our `trim_leading_zero' returns numbers in the scale of 1e-16 and test function tolerances has no effect on it.

Second issue is torch using some sort of substract function (`sub()`) and it throws the error shown below .
```
torch: subtract: RuntimeError: sub(): functions with out=... arguments don't support automatic differentiation, but one of the arguments requires grad.
```
Again I don't know what's happening with our wrappers while all other frameworks ok.

Last question, `paddle` tests are not skipped. Do we need to fix those tests also?

ivy/functional/frontends/jax/numpy/mathematical_functions.py
```diff
@@ -623,6 +622,28 @@ def polyint(p, m=1, k=None):
...
+    scale = 1. / v_arr[0]
+    q = ivy.zeros((max(m - n + 1, 1),), dtype=u.dtype)
+    r = ivy.copy_array(u)
+    for k in range(0, m - n + 1):
+        d = scale * r[k]
+        q[k] = d
+        r[k:k+n+1] -= (d * v_arr)
+    if trim_leading_zeros:
+        r = trim_zeros(r, trim='f')
+        # TODO: Need to find a way to control tolerance of trim_zeros
```
AnnaTz, line 643: You could write a new function to allow zero trimming with tolerance, either as a helper function here, or in our experimental api (I have added `ivy.trim_zeros` to our list). Otherwise you could leave this argument unimplemented and add a ToDo comment explaining why.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_math.py
```diff
@@ -2754,6 +2754,45 @@ def test_jax_numpy_polyint(
...
+        frontend=frontend,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        u=x[0],
+        v=x[1],
+        # trim_leading_zeros=trim_leading_zeros,
+        rtol=1e-01,
+        atol=1e-02,
+    )
+
```
AnnaTz, line 2794: As for the paddle backend, generally yes all 5 backend tests should be passing. However the error you are getting in this case for paddle is a known issue in our supported dtype handling which is being worked on. It's not something that should be solved in this PR.

AnnaTz (changes requested):

AnnaTz (changes requested):

ivy/functional/frontends/jax/numpy/mathematical_functions.py
```diff
@@ -623,6 +622,28 @@ def polyint(p, m=1, k=None):
...
+    r = ivy.copy_array(u)
+    for k in range(0, m - n + 1):
+        d = scale * r[k]
+        q[k] = d
+        r[k:k+n+1] = r[k:k+n+1] - (d * v_arr)
+    # if trim_leading_zeros:
+    #    r = trim_zeros_tol(r, trim='f')
+    # TODO: need to control tolerance of the values returned by this function to handle the argument
+    return q, r
+
```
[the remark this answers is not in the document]
AnnaTz: image

ivy/functional/frontends/jax/numpy/mathematical_functions.py
```diff
@@ -623,6 +622,28 @@ def polyint(p, m=1, k=None):
...
+
```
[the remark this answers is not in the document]
AnnaTz: Okay, let's see if it comes up again in the CI checks as well.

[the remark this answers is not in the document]
sondertg: I just ran it local again, no error from numpy but CI throws that. 

From logs, I think problem is not on my implementation. When divisor so small, my function returns `inf` which makes sense. But `numpy` returns 0. It's really funny that this PR revealed every kind of weird problems.

I might try to solve it by playing with tolerances but how can I get that error at my local? Do I need to pull docker image again or reconfigure whole from start?

```
E   AssertionError:  the results from backend numpy and ground truth framework jax do not match
E    [0.]!=[inf] 
E   
E   
E   Falsifying example: test_jax_numpy_polydiv(
E       on_device='cpu',
E       frontend='jax',
E       dtype_and_x=(['float16', 'float16'],
E        [array([-5.], dtype=float16), array([-6.88e-05], dtype=float16)]),
E       fn_tree='ivy.functional.frontends.jax.numpy.polydiv',
E           with_out=False,
E           inplace=False,
E       ),
E   )
```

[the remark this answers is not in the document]
AnnaTz: Yeah this PR is just bad luck 😅 
You could just pass the specific argument values to your test. I think setting this inside `handle_frontend_test` would work
`dtype_and_x=st.just((['float16', 'float16'], [ivy.array([-5.], dtype=ivy.float16), ivy.array([-6.88e-05], dtype=ivy.float16)])),`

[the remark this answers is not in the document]
AnnaTz: Yeah... the error stopped now you added the assume, so I guess there is something going on in the testing with regards to the supported dtypes decorator. Let's just leave it at that.

PR #14990, opened 2023-05-07
Change real to torch Tensor property

binh234:
`real` is a property of torch Tensor. You can check the signature [here](https://pytorch.org/docs/stable/generated/torch.Tensor.real.html#torch.Tensor.real)


Ookamice:
LGTM! 👍

Thank you for your contributions. 😄

PR #15023, opened 2023-05-08
fixed the numbers in doc the basics

anurag629:
Current view: the numbers are not properly assigned [the_basics](https://unify.ai/docs/ivy/overview/contributing/the_basics.html)

Screenshot from 2023-05-08 17-03-55


KareemMAX (approved):
Sorry for the inconvenience I forgot about the PR 😅, @anurag629 as per our conversation yesterday I want to state it here for the public record

> I noticed that some points in the documentation were rendering correctly using `#`, while others were not. Therefore, I decided to correct the formatting. 
Initially, I only made changes to the points that were not properly rendered, but then I realized that it would be better if we used a consistent method for all the ordering and listings.

I think that your method is better to remove ambiguity from sphinx renderer, the case I mentioned of adding a numbered point in the middle can be fixed manually by hand. I guess readability (and computability) would better be valued more than writability in that specific case

And finally thanks for good work 🚀

Issue #15086, opened 2023-05-09
flatten


jonotassia:
Creating a new sub-task for shortening the code in the TensorFlow Frontend function `flatten()`

Issue #8868, opened 2022-12-19
Fix nest.test_tensorflow_flatten


ivy-leaves:
|   |   |
|---|---|
|numpy|<a href="https://github.com/unifyai/ivy/actions/runs/6412665013/job/17410336937"><img src=https://img.shields.io/badge/-failure-red></a>
|jax|<a href="https://github.com/unifyai/ivy/actions/runs/6412665013/job/17410336937"><img src=https://img.shields.io/badge/-failure-red></a>
|tensorflow|<a href="https://github.com/unifyai/ivy/actions/runs/6364847344/job/17281409619"><img src=https://img.shields.io/badge/-failure-red></a>
|torch|<a href="https://github.com/unifyai/ivy/actions/runs/6364847344/job/17281409619"><img src=https://img.shields.io/badge/-failure-red></a>
|paddle|<a href="https://github.com/unifyai/ivy/actions/runs/6364847344/job/17281409619"><img src=https://img.shields.io/badge/-failure-red></a>

jonotassia:
Hi all,

I'm working on restructuring the associated Tensorflow Frontend function in #15090, so I'll give this one a go at the same time.

jonotassia:
Did some debugging with this one as part of the frontent `tf.flatten` function restructuring and wanted to update with my findings. 

It appears the main issue here is that the `frontend_ret` is being returned as a Python list, which is not being handled in `test_frontent_function` (see details [here](https://github.com/unifyai/ivy/pull/15090#issuecomment-15498747808)), so @sherry30 is going to raise this with the testing team.

ivy-leaves:
Hi there,

Just to let you know, this issue has been depreciated. However, new issues related to any failing tests will be created very soon, marked distinctly with the `Failing Test` label.

Kindly stay tuned for the relevant announcements on our [Discord server](https://discord.com/channels/799879767196958751/983684937712558080) in the coming week.

Thank you for your understanding and cooperation.

PR #15090, opened 2023-05-09
Restructuring TensorFlow frontend function flatten

jonotassia:
I've updated the Tensorflow frontend function [flatten](https://www.tensorflow.org/api_docs/python/tf/nest/flatten) to be under 10 lines of code, per sub-task #15086.

This is failing the test case `test_tensorflow_flatten`. Doing some digging in the issue list, it looks like there is a problem with this test case unrelated to my changes that is causing the test case to fail on all backends - see issue #8868.

Do you want me to attempt to fix the test cases before accepting this pull request, or leave it for another contributor?

Thanks,
Jono
Close https://github.com/unifyai/ivy/issues/15086

### DIFF
--- a/ivy/functional/frontends/tensorflow/nest.py
+++ b/ivy/functional/frontends/tensorflow/nest.py
@@ -1,8 +1,8 @@
-# global
+# local
 import ivy.functional.frontends.tensorflow.ragged as ragged_tf
+from ivy.functional.frontends.tensorflow.func_wrapper import to_ivy_arrays_and_back
 import ivy
 
-# local
 
 # try:
 #     import tensorflow as tf
@@ -34,6 +34,7 @@ def _flatten_composite_array(x, expand_composites=False):
         return ivy.native_sparse_array_to_indices_values_and_shape(x)
 
 
+@to_ivy_arrays_and_back
 def flatten(structure, expand_composites=False):
     if expand_composites and _is_composite_array(structure):
         return _flatten_composite_array(structure, expand_composites=expand_composites)


sherry30 (changes requested):
Hi,

Would be nice if you can get the test passing as well, otherwise its kinda hard to tell if the changes made are correct or not. I'd be happy to help you if you get stuck anywhere 😄 

Also it seems like you've pushed dome changes to .idea files, can you revert those.

Thanks

jonotassia:
Thanks Sherry! 

I'll go ahead and revert the changes to .idea then get started on the fix to the test script. I'll let you know if I have any questions!

jonotassia:
Thanks for the quick response! This all makes sense to me. 

I'll plan to update the function today or tomorrow to use `ivy.flatten`.

jonotassia:
Attempted to update this one to use `ivy.flatten` rather than implementing the logic directly in the function itself, which revealed some behavioural discrepencies between [`tf.nest.flatten`](https://www.tensorflow.org/api_docs/python/tf/nest/flatten) and `ivy.flatten` / the `flatten` functions of the other backend frameworks, namely:

1. `tf.nest.flatten` must be able to handle `dicts` that are passed to it 
    ```
     structure = { "key3": "value3", "key1": "value1", "key2": "value2" }
     flatten(structure)
    ```

    - `tf.nest.flatten` returns: `[value1, value2, value3]`
    - `Ivy.flatten` returns: `IvyBackendException: numpy: flatten: float() argument must be a string or a number, not 'dict'`

2.  `tf.nest.flatten` must be able to handle nonhomogenous, ambiguous-depth lists, tuples, and dicts, as well as certain composite objects like `RaggedTensors`. `ivy.flatten` only handles array-like objects with static dimensions. Example below:
    ```
    structure = ((1.0, 2.0), (3.0, 4.0, 5.0), 6.0)
    flatten(structure)
    ```

    - `tf.nest.flatten` returns: `[1.0, 2.0, 3.0, 4.0, 5.0, 6.0]`
    - `ivy.flatten` returns: `IvyError: numpy: flatten: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part.`

3. Furthermore, even a list that does meet the criteria for `ivy.flatten` is currently returning as an array of bools, rather than as a flattened array of unchanged values, which leads me to believe `ivy.flatten` may be bugged:
   ```
   structure = [["a", "b", "c", 'd'], ["e", "f", "g", "h"]]
   flatten(structure)
   ```
   - `ivy.flatten` returns: `ivy.array([ True,  True,  True,  True,  True,  True,  True,  True])`
   - `tf.nest.flatten` returns: `["a", "b", "c", "d", "e", "f", "g", "h"]`
    
In situations like this, where functions of the same name have different behaviours/outputs from the equivalent function of the other frameworks, what is typically the recommended approach? Right now, all of the implementation details are in `ivy.flatten` leading to the same result regardless of front/backend, but my intuition is that we should be changing `ivy.flatten` to `return ivy.current_backend().flatten(x)`, passing through the relevant arguments based on the backend. Is that reasoning correct, or am I going in the wrong direction?

sherry30:
Hi,

Sorry for such a late review. So in cases like these, we're supposed to implement superset behavior of all the frameworks instead of handling these in the frontend functions. You can read more about it [here](https://github.com/unifyai/ivy/blob/9ce9353d5f5946e9c6073a8cc01946e3ec1cb9c5/docs/overview/deep_dive/superset_behaviour.rst). You've done a really good job of figuring out the differences, we should implement all of these things in `ivy.flatten`  and make sure they work on all backends.
Maybe that might be a bit out of scope for this PR. You can let me know if you wanna work on it otherwise I can just merge your PR and someone else can take a look at adding the superset behavior. If you want to work on it in this PR, that's completely fine as well.

Thanks

jonotassia:
Thanks for reviewing! As always, I really appreciate the feedback and guidance. Your suggestion makes sense to me, and I would agree that it may be beyond the scope of this PR, so happy for you to merge.

That said, I would definitely be interested in continuing to work on this, so I created issue #15795, where I can continue to work through this. Since this issue doesn't have sub-tasks, I wasn't sure how to go about assigning myself to it. Would you be able to assign me to it just to avoid multiple people working on the same thing?

sherry30 (approved):
I've assigned you the task of adding superset behavior.
Feel free to change the frontend again when you're done with  `ivy.flatten` superset behavior.

Thanks 😄

PR #15095, opened 2023-05-09
Added jax.numpy.choose

AbdulDahir7002:


AbdulDahir7002:
close #15089

ivy/functional/frontends/jax/numpy/indexing.py
```diff
@@ -80,3 +80,13 @@ def diag_indices_from(arr):
         raise ValueError("All dimensions of input must be of equal length")
     idx = ivy.arange(n, dtype=int)
     return (idx,) * ndim
+
+@to_ivy_arrays_and_back
+def choose(idx, choices, out=None, mode='raise'):
```
YushaArif99, line 85: I'm not so sure about this compositional implementation you've written. the `choose` function in JAX also has a `mode` parameter right which can either be `raise`, `wrap`, `clip`. I dont think you're handling the specific modes here in your implementation. Instead, it would be better to have an `ivy.choose()` function (along with its backend implementations) and call that in the frontend so that all modes are properly handled. 

This task would then fall under the category of `Extend IVY API` task.  Not sure though if this task is open for contributors though. It's better to ask this on Discord or dm `@IvyTeam` 🙂. If its not open for contributors, then i think it is better to close this PR and you can choose one of the open task from [here](https://github.com/unifyai/ivy/issues?q=is%3Aopen+is%3Aissue+label%3A%22Sub+Task%22+)

YushaArif99 (changes requested):

ivy/functional/frontends/jax/numpy/indexing.py
AbdulDahir7002: Thank you for the advice on the PR 🙂. I will have a look to see which category this function falls under.

ivy-leaves:

Thank you :hugs:

amoungui:
choose #15089

Issue #15055, opened 2023-05-08
Add `paddle.nn.functional.pooling` NN Pooling Functions to PaddlePaddle Frontend


hmahmood24:
Add [NN Pooling Functions](https://www.paddlepaddle.org.cn/documentation/docs/en/api/index_en.html) to PaddlePaddle Frontend

> Please keep in mind that the proper way to link an issue to this list is to comment "- [ ] #issue_number" while the issue's title only includes the name of the function you've chosen.

- [ ] avg_pool1d
- [ ] avg_pool2d
- [ ] avg_pool3d
- [ ] max_pool1d
- [ ] max_pool2d
- [ ] max_pool3d
- [ ] adaptive_max_pool1d
- [ ] adaptive_max_pool2d
- [ ] adaptive_max_pool3d
- [ ] adaptive_avg_pool1d
- [ ] adaptive_avg_pool2d
- [ ] adaptive_avg_pool3d
- [ ] max_unpool1d
- [ ] max_unpool2d
- [ ] max_unpool3d

diksha-shrivastava13:
- [ ] #18318

peterwisu:
max_pool2d #18550

so-dipe:
avg_pool1d #18973

Sar-thak-3:
- [ ] #19063

Sar-thak-3:
- [ ] #19144

Sar-thak-3:
- [ ] #19492

Zamokuhle23:
adaptive_avg_pool3d #19853

nyomangedar:
- [ ] #20213

hassan-204:
- [ ] #21056

ahmedlila:
- [ ] #21083

Praneet9:
- [ ] #21056

Praneet9:
> * [ ]  [max_unpool2d #21056](https://github.com/unifyai/ivy/issues/21056)

@hmahmood24 Can you assign me this?

saif-ashraf99:
#22208

Reptilefury:
-[]max_pool2d

Reptilefury:
max_pool2d

divijsingh02:
- [ ] #22210

Reptilefury:
- [ ] max_unpool3d#22272

Reptilefury:
- [ ] max_unpool3d #22272

zachsueh:
- [ ] #22313

[...]

PR #15113, opened 2023-05-09
added fmax

UbajakaCJ:
Added a new function fmax to the elementwise.py file.  It returns the element-wise maximum of array elements.

It compares two arrays and returns a new array containing the element-wise maxima. If one of the elements being compared is a NaN, then the non-nan element is returned. The latter distinction is important for complex NaNs, which are defined as at least one of the real or imaginary parts being a NaN. The net effect is that NaNs are ignored when possible.

### DIFF
--- a/ivy/functional/ivy/elementwise.py
+++ b/ivy/functional/ivy/elementwise.py
@@ -5656,3 +5656,71 @@ def fmod(
     ivy.array([ nan,  nan,  nan])
     """
     return ivy.current_backend().fmod(x1, x2, out=out)
+
+
+@handle_exceptions
+@handle_nestable
+@handle_out_argument
+@to_native_arrays_and_back
+@handle_array_function
+def fmax(
+    x1: Union[ivy.Array, ivy.NativeArray],
+    x2: Union[ivy.Array, ivy.NativeArray],
+    /,
+    out: Optional[ivy.Array] = None,
+    *,
+    where=True,
+    casting='same_kind',
+    order='K',
+    dtype=None,
+    subok=True[, signature, extobj]
+) -> ivy.Array:
+    """Element-wise maximum of array elements.
+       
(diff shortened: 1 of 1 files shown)


YushaArif99 (changes requested):
Hi @UbajakaCJ 

Thanks a lot for your contribution. A couple of points regarding your PR: 

- Is there a subtask associated with this PR? If so, could you kindly provide an issue # for it.  Contributors are only allowed to work on open tasks. And so if there is no open issue relating to this PR, I think then it's better to close this PR and choose another function in the open tasks [here](https://github.com/unifyai/ivy/issues?q=is%3Aopen+is%3Aissue+label%3AToDo)

- This task seems to be related to the `Extend Ivy API` task. In this case, the task also involves implementing each of the backends for the `fmax` function as well as writing the tests for it.

YushaArif99:
Hi @UbajakaCJ 

By a subtask, what I mean is whether this PR is part of a broader ToDo list. An example of this would be [here](https://github.com/unifyai/ivy/issues/15055)
Basically, as a contributor, you are only allowed to work on tasks that are part of the ToDo list. You can read more about this in the contributing section of the documentation [here](https://unify.ai/docs/ivy/overview/contributing/open_tasks.html#open-tasks)

Regarding writing the tests, its going to be pretty much similar to how other tests are written using the hypothesis helpers. But in order to do that, the first step will be to implement the backends for `fmax`. 

But as I mentioned before, I'd only recommend you begin working on this once its verified this PR is associated with some existing ToDo list so that it can be tracked properly.

YushaArif99:
Hi @UbajakaCJ 

Are you still working on this PR? If not, we can then close this PR 🙂

UbajakaCJ:
Yes, I would like to continue working on this PR.

PR #15264, opened 2023-05-10
Update `intelligent-tests-pr.yml` to use matrices

KareemMAX:


RashulChutani:
Hi @KareemMAX, Thanks for this amazing suggestion and implementation. Actually, we use the `extra` argument with only the first runner, and not with others. Can you take this into account? Thanks.

KareemMAX:
Hi @RashulChutani, I've finished the PR. Please take a look into the `intelligent-tests-pr` actions. Sadly I haven't tested the one on `master`.

Also I wanted to mention that due to how matrices work, I didn't make individual jobs fail as I guess the final check will be presented in `determine-test-results` job, if you think that would be an issue please let me know

RashulChutani:
Actually, that will be an issue, as the final job doesn't check for failures. The individual jobs do that.

KareemMAX:
From what I understood, `display_test_results` will attempt to download all artifacts uploaded from previous jobs, and display the results without checking on "github action failures"

I guess we can just signal the last job to `exit 1` after a failure and leave it as that, meaning that the matrix will always pass and the last runner will always show the correct result and fail if failure is introduced (even though subsequent jobs passed)

PR #15377, opened 2023-05-12
Adding map_tree test for jax.src_tree_util 

abdulasiraj:


ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_tree_util.py
```diff
@@ -1 +1,66 @@
...
+            ),
+            values=st.one_of(leaf_strategy(), tree_strategy(max_depth - 1)),
+            min_size=1,
+            max_size=10,
+        )
+
+
+@st.composite
+def tree_dict_strategy(draw):
+    return draw(tree_strategy())
```
AnnaTz, line 44: Great! My only suggestion is you could define `tree_strategy` directly as a composite instead of using the extra `tree_dict_strategy`.

abdulasiraj: I have already tried but it gives me following error because tree_strategy is recursive procedure :

`unsupported operand type(s) for ** or pow(): 'MappedSearchStrategy' and 'int'`

AnnaTz (approved):
Then all looks good to me!

PR #15386, opened 2023-05-12
Implemented tensorflow l2_normalize in the tensorflow frontend and couldnt test it .

ferstappen23:
**This pull request makes the following changes :**

* Implements issue #15112
**Problems** 
 I got an initial error message ImportError while loading conftest C:\Users\Verstappen\ivy\ivy_tests\test_ivy\conftest.py   , indicating that there is an error while loading the configuration file 'conftest.py' . However, the error further showed that the error occurred in the file function_testing.py on line 39, when attempting to import the tensorflow module, suggesting that there is an attribute error because of a circular import between the jax and tensorflow modules,which is causing the function_testing.py to fail .
Do you guys know how to resolve this?

### DIFF
--- a/ivy/functional/frontends/torch/nn/functional/vision_functions.py
+++ b/ivy/functional/frontends/torch/nn/functional/vision_functions.py
@@ -107,22 +107,12 @@ def _handle_padding_shape(padding, n, mode):
 
 @to_ivy_arrays_and_back
 def pad(input, pad, mode="constant", value=0):
+    mode_dict = {"constant": "constant", "reflect": "reflect", "replicate": "edge", "circular": "wrap"}
+    if mode not in mode_dict:
+        raise ValueError(f"Unsupported padding mode: {mode}")
     pad = _handle_padding_shape(pad, len(input.shape), mode)
-    if mode == "constant":
-        return ivy.pad(input, pad, mode="constant", constant_values=value)
-    elif mode == "reflect":
-        return ivy.pad(input, pad, mode="reflect", reflect_type="even")
-    elif mode == "replicate":
-        return ivy.pad(input, pad, mode="edge")
-    elif mode == "circular":
-        return ivy.pad(input, pad, mode="wrap")
-    else:
-        raise ivy.utils.exceptions.IvyException(
-            (
-                "mode '{}' must be in "
-                + "['constant', 'reflect', 'replicate', 'circular']"
-            ).format(mode)
-        )
+    return ivy.pad(input, pad, mode=mode_dict[mode], constant_values=value)
+
 
 
 def _get_new_width_height(w_old, h_old, size=None, scale_factor=None):


ferstappen23:
Hi @theRealBird,
I finished my task and did my PR , I also checked the logs and fixed the keyword error for the get_shape helper function.
Can you review it for me.
Thanks

theRealBird:
Hi @ferstappen23, Thank you for your patience. I will definitely have a look at your PR. It seems you were having some problems before. I will have a look and update you.
Thanks!

ferstappen23:
I would highly appreciate it .

ivy_tests/test_ivy/conftest.py
theRealBird, line 1: Are these changes intended?😅

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
```diff
@@ -2,14 +2,17 @@
...
-
+import pytest
 # local
 import ivy_tests.test_ivy.helpers as helpers
 from ivy_tests.test_ivy.test_functional.test_core.test_statistical import (
     statistical_dtype_values,
 )
 from ivy_tests.test_ivy.helpers import handle_frontend_test
 
+import paddle.fluid.core
```
theRealBird, line 13: How is ```paddle.fluid.core``` being used in the code?

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
```diff
@@ -2527,3 +2530,36 @@ def test_tensorflow_in_top_k(
...
+    *,
+    dtype_and_x,
+    shape_and_axis,
+    #eps,
+    frontend,
+    test_flags,
+    fn_tree,
+    on_device,
+):
+    import pdb
```
theRealBird, line 2557: I think you were using a debugger here. This code should be removed after local testing is finished.

requirements/optional.txt
theRealBird, line 1: Are these changes related to the implementation of ```l2_normalize```?

.idea/vcs.xml
theRealBird, line 1: Did you mean to commit the changes in these files? I think the .idea folder can be ignored.

ivy/functional/frontends/tensorflow/math.py
theRealBird, line 1: ```tf.math.l2_normalize``` also supports the arguments ```epsilon``` and ```name``` as seen in their documentation [here](https://www.tensorflow.org/api_docs/python/tf/math/l2_normalize) 

Hence, the frontend implementation of this function should also support those requirements. Please add support for the same.

Dockerfile
theRealBird, line 1: Could you please explain why these changes are needed? Thanks!

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
```diff
@@ -2,14 +2,17 @@
 import ivy
 import numpy as np
 from hypothesis import strategies as st, assume
-
+import pytest
```
theRealBird, line 5: Are we using pytest later in our code?

run_tests_CLI/run_tests.sh
theRealBird, line 1: Could you please explain what these changes do? Thanks :)

theRealBird (commented):
Hi @ferstappen23, 

I have left some comments on each of the files that are a part of this commit. I understand that you were facing some issues earlier.

Ideally, the implementation of a frontend function like ```l2_normalize``` would not require changes to so many files 😅

My initial thought is that changes to ```ivy/functional/frontends/tensorflow/math.py``` and ```ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py``` should be enough for the implementation.

I would suggest having a look at a similar implementation of ```l2_normalization``` (alias in tf.linalg) in ``` ivy/functional/frontends/tensorflow/linalg.py``` and its corresponding test if you feel stuck or checking out the setting-up section of our docs to ensure your tools are working correctly.

Thank you so much for your work and patience. Let me know if you need any help 😃

ferstappen23:
@theRealBird 
 Thank you so much for your feedback , i agree that the changes only concerns the math.py and the test_math.py files ,the unintended changes were made in my attempts  to ensure the functionality of the tests. , i'm aware now that i should have staged and committed only the relevant files. I will promptly address this and make the necessary changes.

PR #15430, opened 2023-05-14
Jax frontend random.bernoulli , random.fold_in

### LINKED ISSUE #15354: bernoulli

### LINKED ISSUE #15429: fold_in

NouranYoussef:
closes #15354
closes #15429


ivy_tests/test_ivy/test_frontends/test_jax/test_jax_random.py
```diff
@@ -730,3 +730,94 @@ def call():
...
+
+    dtype_and_probs = draw(
+        helpers.dtype_and_values(
+            available_dtypes=dtype, min_value=0, max_value=1, shape=shape
+        )
+    )
+    return dtype_and_probs, shape
+
+
+# TODO Update the test by fixing the uint32 unsupported problem
```
Daniel4078, line 748: So are you going to continue working on this proble and update your tests?

Daniel4078: Also, I think that you may also need to include these part:
    ret = call()
    if not ivy.exists(ret):
        return
    ret_np, ret_from_np = ret
    ret_np = helpers.flatten_and_to_np(ret=ret_np)
    ret_from_np = helpers.flatten_and_to_np(ret=ret_from_np)
    for u, v in zip(ret_np, ret_from_np):
        assert u.dtype == v.dtype
        assert u.shape == v.shape
in your test.

NouranYoussef: I've included them in the last commit

NouranYoussef: > So are you going to continue working on this proble and update your tests?

The issue is caused by PyTorch not supporting uint32 and I've marked the failing test case with @pytest.mark.xfail as an expected failure

Daniel4078: > > So are you going to continue working on this proble and update your tests?
> 
> The issue is caused by PyTorch not supporting uint32 and I've marked the failing test case with @pytest.mark.xfail as an expected failure

But you can use the helpers.get_dtypes("integer", full=False) to get the supported dtype for the backend instead of just testing for uint32. so that we can have more detail about the failure and help fix the backend.

NouranYoussef: this cause all tests to fail as jax expects key dtype to be uint32

Daniel4078: So, it is correct to only generate uint32 keys for the tests. Can I have a look at the failures for both two tests in that case?

Daniel4078: I will try to fix the torch backend problem after taking a look at the error logs.

NouranYoussef: > So, it is correct to only generate uint32 keys for the tests. Can I have a look at the failures for both two tests in that case? By just not using the mark.xfail.

it is removed now and bernoulli has available_dtypes=helpers.get_dtypes("integer", full=False) and fold_in has available_dtypes=["uint32"] for key dtype

PR #15447, opened 2023-05-15
Change base image from debian to ubuntu and upgrade python version

bipinKrishnan:
The PR updates the dockerfile with the following changes:

1. Change the docker image from Debian buster to Ubuntu 22.04
2. Upgrade Python from 3.8 to 3.10
3. Make use of `PyOpenGL` pip package instead of the apt package

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,8 +1,8 @@
-FROM debian:buster
+FROM ubuntu:22.04
 WORKDIR /ivy
 ARG CLI
 # python version for conda
-ARG pycon=3.8.10
+ARG pycon=3.10
 
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97:
Hi, is this an ongoing task? if not what is the motive for these changes?

bipinKrishnan:
Hi @MahmoudAshraf97, the changes are based on the trello card here:
https://trello.com/c/mrDtbTks

VedPatwardhan (approved):
lgtm! Let's get the PR merged and if it breaks anything on master we could revert it, given that there isn't a way to exhaustively test if there are failures. Being able to build a codespace is a good check to know that it doesn't cause any major issues, probably run tests for a couple of submodules in the codespace you created before getting the PR merged. Thanks @bipinKrishnan 🙂

PR #15448, opened 2023-05-15
adding test function for jax.src_.api.vmap

abdulasiraj:


ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_api.py
```diff
@@ -4,6 +4,83 @@
...
+def _fn1(x, y):
+    return ivy.matmul(x, y)
+
+
+def _fn2(x, y):
+    return ivy.vecdot(x, y)
+
+
+def _fn3(x, y):
+    ivy.add(x, y)
```
AnnaTz, line 19: Did you forget `return` or was it on purpose? 😅

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_api.py
```diff
@@ -4,6 +4,83 @@
...
+    ivy.previous_backend()
+
+    if fw_res is not None and jax_res is not None:
+        helpers.value_test(
+            ret_np_flat=fw_res,
+            ret_np_from_gt_flat=jax_res,
+            rtol=1e-1,
+            atol=1e-1,
+        )
+
```
AnnaTz, line 79: Right now it seems you are testing the jax backend of `ivy.vmap`, but what we should be testing is the jax frontend `vmap`.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_api.py
abdulasiraj: It was on purpose as this function used by vmap.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_api.py
abdulasiraj: I think you are right, I should import jax.vmap not ivy.vmap. let me fix it.
Thanks!

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_api.py
abdulasiraj: oh, I forgot.

abdulasiraj:
let me explain what's going on.

- Get the results from vmap function implemented in jax.src_.api.
- Get the results from ivy.vmap.
- Comparing both results with tolerance.

Also, It's not the function which can be tested with `test_frontend_function`. Reason is, we can't pass the input directly to the function arguments. Below it the example:

`jax.vmap(some_function, in_axes=0)(x1, x2)`  

we can only set parameter using vmap arguments but the input will be handled by the resulting function.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_api.py
```diff
@@ -4,6 +4,83 @@
...
+        fw_res = None
+
+    ivy.set_backend("jax")
+    arrays = [ivy.native_array(array) for array in generated_arrays]
+    if in_axes_as_cont:
+        jax_vmapped_func = ivy.vmap(func, in_axes=in_axes, out_axes=0)
+    else:
+        jax_vmapped_func = ivy.vmap(func, in_axes=0, out_axes=0)
+
+    assert callable(jax_vmapped_func)
```
AnnaTz, line 62: Frontend functions should be compared with their corresponding native functions, not ivy functions. So in this case you need to compare `ivy.functional.frontends.jax._src.api.vmap` to the original `jax._src.api. vmap`.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_api.py
```diff
@@ -4,6 +4,83 @@
...
+        allow_none=False,
+        min_num_dims=2,
+        max_num_dims=5,
+        min_dim_size=2,
+        max_dim_size=10,
+        num=2,
+        return_dtype=True,
+    ),
+    in_axes_as_cont=st.booleans(),
+)
```
AnnaTz, line 37: Also I would suggest sampling a backend string here and passing it to `test_vmap` so you can set a random backend in each iteration of the test.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_api.py
abdulasiraj: Got it

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_api.py
abdulasiraj: Make sense!

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_src_api.py
```diff
@@ -4,6 +4,94 @@
...
+    else:
+        vmapped_func = vmap(func, in_axes=0, out_axes=0)
+
+    assert callable(vmapped_func)
+
+    try:
+        fw_res = helpers.flatten_and_to_np(ret=vmapped_func(*arrays))
+        fw_res = fw_res if len(fw_res) else None
+    except Exception:
+        fw_res = None
```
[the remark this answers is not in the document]
abdulasiraj: `flatten_and_to_np`, it may fail.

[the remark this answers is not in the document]
abdulasiraj: sure

PR #15457, opened 2023-05-15
Close #15456 

### LINKED ISSUE #15456: igamma

VitaliyStels:
Added Tensorflow Frontend Igamma function

### DIFF
--- a/ivy/functional/backends/paddle/creation.py
+++ b/ivy/functional/backends/paddle/creation.py
@@ -186,14 +186,20 @@ def empty_like(
 
 
 @with_unsupported_device_and_dtypes(
-    {"2.4.2 and below": {"cpu": ("uint8",
-                                 "int8",
-                                 "int16",
-                                 "float16",
-                                 "complex64",
-                                 "complex128",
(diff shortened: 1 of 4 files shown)


ivy/functional/frontends/tensorflow/math.py
```diff
@@ -668,3 +668,8 @@ def greater_equal(x, y, name=None):
 def in_top_k(target, pred, k, name=None):
     top_k = ivy.top_k(target, k)
     return ivy.array([val in top_k.values for val in target])
+
+
+@to_ivy_arrays_and_back
+def igamma(a, x, name=None):
+    return ivy.igamma(a, x)
```
zaeemansari70, line 675: There is no `igamma` function in the backend which makes this a dummy call 
Kindly either suggest us to add it in the experimental backend using our github template if it is not present in or ToDo list for experimental API

zaeemansari70 (commented):
Hi @VitaliyStels ,

Thanks for the contribution, left some comments below which needs to be addressed :slightly_smiling_face:

PR #15464, opened 2023-05-15
Test sweep for cross entropy

### LINKED ISSUE #12308: Examine `test_ivy.test_functional.test_losses` tests
Ensure a comprehensive examination of all tests to confirm their alignment with the current test writing policy. Ensure the tests follow the following guidelines, among others:

1. Avoid restricting the use of `get_dtypes` to specific kinds (`float`, `int`) unless absolutely necessary; the default value should be used when possible.
2. Avoid the use of `assume` to bypass specific data types.
3. Unless necessary, avoid restricting the generation of numbers with `min_value` or `max_value`. For example, it is acceptable to set a maximum value for the `pad` function if the `pad_value` value can halt the test.
4. Ensure proper use of strategies.
5. Verify that all tests are executed rather than skipped by `test_values=False`, unless the test logic is implemented within the test body itself.
6. Confirm that testing is comprehensive and complete, meaning that all possible combinations of inputs are tested.
7. Avoid manually setting test parameters, including native array flags, container flags, the number of positional arguments, etc.
8. Ensure that no parameters are skipped during testing.

> Always refer to the documentation for **Ivy Tests** and **Ivy Frontend Tests** for complete test writing policy.

If a test function does not require update, simply mark the function in the list as completed, otherwise create an issue for that test function.
- [ ] #16334
- [ ] #16335
- [ ] #16336

abdulasiraj:
partially closes #12308


ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
```diff
@@ -15,23 +15,17 @@
         max_value=1,
         allow_inf=False,
         min_num_dims=1,
-        max_num_dims=1,
-        min_dim_size=2,
     ),
     dtype_and_pred=helpers.dtype_and_values(
         available_dtypes=helpers.get_dtypes("float"),
         min_value=1e-04,
```
CatB1t, line 21: Ideally, we should also not limit the min value, if it causes issues due to precision errors, we should use the safety factor.

ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
```diff
@@ -15,23 +15,17 @@
...
         max_value=1,
         allow_inf=False,
-        exclude_min=True,
-        exclude_max=True,
         min_num_dims=1,
-        max_num_dims=1,
-        min_dim_size=2,
     ),
     reduction=st.sampled_from(["none", "sum", "mean"]),
     axis=helpers.ints(min_value=-1, max_value=0),
```
CatB1t, line 27: The axis value shouldn't be hardcoded, instead we should update the strategies to actually generate valid axis values.

ivy_tests/test_ivy/test_functional/test_nn/test_losses.py
```diff
@@ -15,23 +15,17 @@
...
-        min_dim_size=2,
     ),
     dtype_and_pred=helpers.dtype_and_values(
         available_dtypes=helpers.get_dtypes("float"),
         min_value=1e-04,
         max_value=1,
         allow_inf=False,
-        exclude_min=True,
-        exclude_max=True,
         min_num_dims=1,
```
CatB1t, line 24: We should allow for 0-dim values.

CatB1t (changes requested):

PR #15473, opened 2023-05-15
Added get_static_value() func in tensorflow frontend

anashassan517:


MuhammadSaeedBatikh:
Hi @anashassan517, thanks for the PR.
Please add the following decorator so that the function is tested properly:
```
@handle_frontend_test(
    fn_tree="tensorflow.get_static_value",
 dtype_values_axis=helpers.dtype_values_axis( ...
```
Also, it does not seem that we require `axis` samples to test get_static. Maybe other `helpers` methods would be more appropriate such as `helpers.dtype_and_values`.
Finally, please make sure that tests pass locally.

PR #15483, opened 2023-05-16
Unsupported dtype decorator

RickSanchezStoic:
1. Updates the decorator logic to not use the last known key
2. Updates the init files, specifically the dictionaries.


VedPatwardhan (approved):
The changes look good, but let's wait for the tests to run as there could be a refactor left somewhere. Thanks @RickSanchezStoic 🙂

PR #15491, opened 2023-05-16
Adding test files for tf1

abdulasiraj:
related to #15452


CatB1t (approved):
lgtm! just something minor but maybe @sherry30 can give better insight, I think we should have the subpackage named `tensorflow_v1`.

sherry30:
Had a chat with @AnnaTz  about this, `tensorflow_v1` sounds better but the official name is `tensorflow_1`. So for now we're going with that.

PR #15505, opened 2023-05-16
Adding isreal #12105

haneeshathanati:
isreal #12105


haneeshathanati:
Close #15505

ZiadAmerr:
Hey, @haneeshathanati. Thanks a lot for your contribution.

I think this function is already implemented and the issue was with the ivy bit that marks it as completed. Can you please choose another incomplete function and make sure that it is incomplete so that this doesn't happen again? :)

PR #15519, opened 2023-05-17
added numpy.random.negative_binomial to frontend

### LINKED ISSUE #15402: negative_binomial

baberabb:
close #15402

numpy implements the negative_binomial function as a mixture of the gamma and poisson distribution so I implemented that.

For the testing I used an arbitrary minimum value for the p parameter as setting it too low can cause the poisson parameter to exceed the max acceptable value of the poisson function.


baberabb:
Hi @nassimberrada, appreciate the detailed feedback! I have made the necessary adjustment to the implementation, ensuring that the shape argument is correctly handled by converting integers to tuples when needed. Some checks are still failing but do not seem be related, from what I can tell.

For the testing part, I can't seem to find anything that would suitable to test the validity of the negative_binomial function. Am open to suggestions if you have any and I agree there needs to be a more rigorous testing of the random functions generally. Thanks again for the feedback!

PR #15528, opened 2023-05-17
Adding torch backend decorator to handle numpy arrays

abdulasiraj:


KareemMAX:
ivy-gardener

KareemMAX:
Hi @abdulasiraj, I was testing linting on your PR assuming I could make it green 😅

Please ignore the lint-bot, and the changes it made.

abdulasiraj:
Hi @CatB1t ,
can you please make some comments on this?
Thanks!

abdulasiraj:
Hi @CatB1t ,
Can you please have a look?

ivy/func_wrapper.py
```diff
@@ -8,6 +8,11 @@
 from types import FunctionType
 from typing import Callable
 import inspect
+import torch
```
CatB1t, line 11: We shouldn't import torch here, let's make it backend specific decorator

ivy/func_wrapper.py
```diff
@@ -660,6 +665,34 @@ def _handle_view_indexing(*args, **kwargs):
     return _handle_view_indexing
 
 
+def _convert_numpy_arrays_to_torch(*args):
+    torch_tensors = []
+    for array in args:
+        if isinstance(array, np.ndarray):
```
CatB1t, line 671: Ideally, we should do a nested search as the argument may be a list containing numpy arrays, which wouldn't work in that case. maybe make use of ivy nest functions here.

CatB1t (changes requested):

ivy/func_wrapper.py
VedPatwardhan: Adding to this, it would be great if we could try and reuse/refactor some of the logic [here](https://github.com/unifyai/ivy/blob/d5cc0291e64af78f58daef7650d9f75a6625df50/ivy/utils/backend/handler.py#L366) at it serves a similar purpose (except for the garbage collection part). Thanks!

VedPatwardhan (changes requested):

abdulasiraj:
Hi @CatB1t @VedPatwardhan ,

It's working very fine now and ready for review. suggestions were really helpful. Thanks all!

ivy/func_wrapper.py
VedPatwardhan: I think what @CatB1t meant here is that we do something like,
```
np_arr_idxs = ivy.nested_argwhere(args)        # nested search
np_arr_vals = ivy.multi_index_nest(args)         # indexing to get the arrays  
backend_arr_vals = [ivy.asarray(val).to_native() for val in np_arr_vals]      # conversion
ivy.set_nest_at_indices(args, np_arr_idxs, backend_arr_vals)   # setting the backend ones back
```

VedPatwardhan (changes requested):
Hey @abdulasiraj , just requsted a change, rest looks good to go. Thanks 🙂

abdulasiraj:
Hi @VedPatwardhan 
This one is ready for review.
Thanks!

ivy/functional/backends/torch/elementwise.py
```diff
@@ -48,7 +50,8 @@ def bitwise_xor(
 bitwise_xor.support_native_out = True
 
 
-@with_unsupported_dtypes({"2.0.1 and below": ("float16", "complex")}, backend_version)
```
VedPatwardhan, line 51: Could you please revert the changes made to `with_unsupported_dtypes` in this file?

VedPatwardhan (changes requested):
Hey @abdulasiraj, just a small change, could you please also look into why all the intelligent tests are failing and whether the failures are related to the changes you've made? I'll also add the `Exhaustive CI` label just to be sure that it doesn't break anything. Thanks 🙂

abdulasiraj:
Hi @VedPatwardhan 
It's ready to review.

VedPatwardhan:
> Hi @VedPatwardhan It's ready to review.

Hey @abdulasiraj, could you please link the task that involves fixing this? Because given that this would add quite a bit of overhead to `torch` functions, it would be useful to have this as an optional mode because most usages wouldn't necessarily involve `numpy` arrays. Thanks🙂

abdulasiraj:
> > Hi @VedPatwardhan It's ready to review.
> 
> Hey @abdulasiraj, could you please link the task that involves fixing this? Because given that this would add quite a bit of overhead to `torch` functions, it would be useful to have this as an optional mode because most usages wouldn't necessarily involve `numpy` arrays. Thanks 🙂

https://trello.com/c/1JPoKxID

VedPatwardhan (commented):
Sounds good! Is this limited to elementwise functions? If not we might want to add it for all such functions which work with `np.ndarray` for the other backends. Thanks @abdulasiraj 🙂

PR #15530, opened 2023-05-17
Flat numpy

AsmaaMHadir:
I am running into the following error:
`TypeError: "numpy.flatiter" object is not callable."`
 when I run the test_numpy_instance_method, and have traced it back to this snippet of code inside the test_frontend_method function:
```
frontend_ret = ins_gt.__getattribute__(frontend_method_data.method_name)(
    *args_method_frontend, **kwargs_method_frontend
)
```
 it seems like the numpy.flatiter objects does not find a callable method 'flat'; I would appreciate your feedback on what changes to make.

### DIFF
--- a/ivy/functional/frontends/paddle/tensor/logic.py
+++ b/ivy/functional/frontends/paddle/tensor/logic.py
@@ -237,3 +237,23 @@ def is_tensor(x):
 @to_ivy_arrays_and_back
 def isclose(x, y, rtol=1e-05, atol=1e-08, equal_nan=False, name=None):
     return ivy.isclose(x, y, rtol=rtol, atol=atol, equal_nan=equal_nan)
+
+
+@with_supported_dtypes(
+    {
+        "2.5.0 and below": (
+            "bool",
+            "int8",
+            "int16",
+            "int32",
+            "int64",
+            "float32",
+            "float64",
+        )
+    },
+    "paddle",
+)
+@to_ivy_arrays_and_back
+@handle_out_argument
+def logical_and(x, y, /, *, name=None, out=None):
+    return ivy.logical_and(x, y, out=out)
--- a/ivy/functional/frontends/paddle/tensor/tensor.py
+++ b/ivy/functional/frontends/paddle/tensor/tensor.py
@@ -309,6 +309,23 @@ class Tensor:
     def rsqrt(self, name=None):
         return ivy.reciprocal(ivy.sqrt(self._ivy_array))
 
+    @with_supported_dtypes(
+        {
+            "2.5.0 and below": (
+                "bool",
+                "int8",
+                "int16",
+                "int32",
+                "int64",
+                "float32",
+                "float64",
+            )
+        },
+        "paddle",
+    )
+    def logical_and(self, y, out=None, name=None):
+        return paddle_frontend.logical_and(self, y, out=out)
+
     @with_supported_dtypes({"2.5.0 and below": ("float32", "float64")}, "paddle")
     def divide(self, y, name=None):
         return paddle_frontend.divide(self, y)
--- a/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_logic.py
+++ b/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_logic.py
@@ -468,3 +468,32 @@ def test_paddle_isclose(
         y=x[1],
         equal_nan=equal_nan,
     )
+
+
+# logical_and
+@handle_frontend_test(
+    fn_tree="paddle.logical_and",
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("valid"),
+        num_arrays=2,
+        shared_dtype=True,
+    ),
+)
+def test_paddle_logical_and(
+    *,
+    dtype_and_x,
+    on_device,
+    fn_tree,
+    frontend,
+    test_flags,
+):
+    input_dtype, x = dtype_and_x
+    helpers.test_frontend_function(
+        input_dtypes=input_dtype,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        x=x[0],
+        y=x[1],
+    )
--- a/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_tensor.py
+++ b/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_tensor.py
@@ -1511,6 +1511,37 @@ def test_paddle_greater_than(
     )
 
 
+# logical_and
+@handle_frontend_method(
+    class_tree=CLASS_TREE,
+    init_tree="paddle.to_tensor",
+    method_name="logical_and",
+    dtypes_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("valid"), num_arrays=2, shared_dtype=True
+    ),
+)
+def test_paddle_instance_logical_and(
+    dtypes_and_x,
+    frontend_method_data,
+    init_flags,
+    method_flags,
+    frontend,
+    on_device,
+):
+    input_dtype, x = dtypes_and_x
+    helpers.test_frontend_method(
+        init_input_dtypes=input_dtype,
+        init_all_as_kwargs_np={"self": x[0]},
+        method_input_dtypes=input_dtype,
+        method_all_as_kwargs_np={"y": x[1]},
+        frontend_method_data=frontend_method_data,
+        init_flags=init_flags,
+        method_flags=method_flags,
+        frontend=frontend,
+        on_device=on_device,
+    )
+
+
 # divide
 @handle_frontend_method(
     class_tree=CLASS_TREE,


.devcontainer/build/devcontainer.json
```diff
@@ -20,6 +20,7 @@
         "post_create": "bash .devcontainer/post_create_commands.sh", 
         "bashrc": "echo \"alias python=python3\" >> ~/.bashrc"
     },
+	"postStartCommand": "git config --global --add safe.directory ${containerWorkspaceFolder}"
```
hirwa-nshuti, line 23: This should be reverted back, I don't think it is needed on these changes

ivy/functional/frontends/numpy/ndarray/ndarray.py
```diff
@@ -288,6 +288,9 @@ def flatten(self, order="C"):
         else:
             return np_frontend.ravel(self, order="C")
 
+    def flat(self):
```
hirwa-nshuti, line 291: As this is supposed to be an instance of `numpy.flatiter` class I think we should first implement this class😊.
But if this approach works for now we can implement it later no problem, simply add a ToDo comment

ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py
```diff
@@ -116,6 +116,37 @@ def test_numpy_ndarray_property_T(
...
+    method_name="flat",
+    dtypes_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes(kind="valid"),
+    ),
+)
+def test_numpy_instance_flat(
+    dtypes_and_x, frontend_method_data, init_flags, method_flags, frontend, on_device
+):
+    input_dtypes, vals = dtypes_and_x
+    print(type(vals))
```
hirwa-nshuti, line 134: Will this be removed later

ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py
```diff
@@ -116,6 +116,37 @@ def test_numpy_ndarray_property_T(
...
+    helpers.test_frontend_method(
```
hirwa-nshuti, line 135: The `numpy.ndarray.flat` is an attribute that's why we shouldn't be testing it as a method. You can refer to the way we test attributes and properties https://github.com/unifyai/ivy/blob/784056d06d3891e847cdfbbf05edbb08bba138c0/ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py#L72

hirwa-nshuti (changes requested):
Hey @AsmaaMHadir 

Thanks for the PR😊
I just added some comments below, feel free to reach out if there is any confusion🙂
Happy coding🤞

ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py
AsmaaMHadir: Yess! I was just doing some debugging :). I will remove it and commit the change.

ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py
AsmaaMHadir: Thanks a lot! this is what I had been looking for. I will modify my implementation and submit it for your feedback!

ivy/functional/frontends/numpy/ndarray/ndarray.py
AsmaaMHadir: Hey Hirwa-nshuti - that was the first question that I had when I started working on the issue: whether to implement the flatiter class first or use the numpy one;. I have updated my flat function to return `self.ivy_array.flat` instead and the test passes for Numpy and fails for the rest of the backends due to `NotImplementedError`; I think this may mean that implementing an ivy flatiter class is needed for all of the backends to access '`flat' `. Please correct me if I am wrong :)

AsmaaMHadir: Update: I have tried testing the flat property with both the flatten and ravel methods:

`ivy.utils.assertions.check_equal(x.flat, np.nditer(data[0].ravel()))`
`ivy.utils.assertions.check_equal(x.flat, np.nditer(data[0].flatten()))`

and both gave me the same error that I encountered with my current implementation:
update

hirwa-nshuti: Basically this class is needed in the ivy frontend so it will be iterator class of frontend ndarray

AsmaaMHadir: Thank you, Hirwa. Do I implement the class in the ndarray.py file ?

ivy/functional/frontends/numpy/ndarray/ndarray.py
```diff
@@ -53,6 +53,10 @@ def shape(self):
     def dtype(self):
         return self.ivy_array.dtype
 
+    @property
+    def flat(self):
```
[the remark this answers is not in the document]
AsmaaMHadir: Could you please explain what you mean by handling them manually?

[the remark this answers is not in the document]
AsmaaMHadir: Thank you Hirwa. I have modified it and now the tests are successful.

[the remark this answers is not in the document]
AsmaaMHadir: Thank you for clearing it out! I have changed my implementation to return an `nditer` over the `ivy_array` and the tests are passing successfully now. However, I think as you mentioned earlier an implementation of a flatiter class would be needed to be in line with the rest of the ivy methods. I think implementing `nditer` in the `np_frontend` is what should be the next ToDo?

ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py
```diff
@@ -116,6 +116,19 @@ def test_numpy_ndarray_property_T(
     )
 
 
+@given(
+    dtypes_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes(kind="numeric", prune_function=False),
+        ret_shape=True,
+    ),
+)
+def test_numpy_ndarray_property_flat(dtypes_and_x):
```
hirwa-nshuti, line 125: The way you're manually handling the tests run is incorrect we should be computing the returns of both the numpy framework and ivy frontends, and then we can compare the results and check if they're the same.

ivy_tests/array_api_testing/test_array_api
hirwa-nshuti, line 1: These changes should be removed

hirwa-nshuti (changes requested):
Hey @AsmaaMHadir 

I added a few comments below would you check them🙂

hirwa-nshuti (changes requested):
Hey we don't use any framework dependency in the frontends.
We only use ivy functions, so you will have to remove this numpy dependency in the function implementation🙂

Thanks

AsmaaMHadir:
> Hey we don't use any framework dependency in the frontends. We only use ivy functions, so you will have to remove this numpy dependency in the function implementation🙂
> 
> Thanks

Hey Felix - I had made changes since yesterday omitting the use of np but for some reason my changes are not pushed properly (debugging why right now). Thanks!

hirwa-nshuti (changes requested):
The `ivy.Array` doesn't have `flat` method we can use other functions to flatten the array

AsmaaMHadir:
> The `ivy.Array` doesn't have `flat` method we can use other functions to flatten the array

Noted.

AsmaaMHadir:
> The `ivy.Array` doesn't have `flat` method we can use other functions to flatten the array

I have attempted using the iter() and flatten() methods to flatten the ivy.Array like shown below:
```
    @property
    def flat(self):
        return iter(self.flatten())
```
However, I am getting the following error and not sure why:
`AttributeError: 'tuple' object has no attribute '__bool__'`
askkinnng

I got the same error using these two methods as well:

`return iter(np_frontend.ravel(self)) `&& `return iter(self.ravel())`
What do you believe I am missing? So far, the latest solution I pushed is the only one passing the Pytest successfully.
The logic should be that the ivy_array is flattened and then iterated over. Please correct me if I'm wrong. Thanks!

hirwa-nshuti:
> > The `ivy.Array` doesn't have `flat` method we can use other functions to flatten the array
> 
> I have attempted using the iter() and flatten() methods to flatten the ivy.Array like shown below:
> 
> ```
>     @property
>     def flat(self):
>         return iter(self.flatten())
> ```
> 
> However, I am getting the following error and not sure why: `AttributeError: 'tuple' object has no attribute '__bool__'` askkinnng
> 
> I got the same error using these two methods as well:
> 
> `return iter(np_frontend.ravel(self)) `&& `return iter(self.ravel())` What do you believe I am missing? So far, the latest solution I pushed is the only one passing the Pytest successfully. The logic should be that the ivy_array is flattened and then iterated over. Please correct me if I'm wrong. Thanks!

To use `flatten` you will need to call `self.ivy_array.flatten()`

AsmaaMHadir:
> > > The `ivy.Array` doesn't have `flat` method we can use other functions to flatten the array
> > 
> > 
> > I have attempted using the iter() and flatten() methods to flatten the ivy.Array like shown below:
> > ```
> >     @property
> >     def flat(self):
> >         return iter(self.flatten())
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However, I am getting the following error and not sure why: `AttributeError: 'tuple' object has no attribute '__bool__'` askkinnng
> > I got the same error using these two methods as well:
> > `return iter(np_frontend.ravel(self)) `&& `return iter(self.ravel())` What do you believe I am missing? So far, the latest solution I pushed is the only one passing the Pytest successfully. The logic should be that the ivy_array is flattened and then iterated over. Please correct me if I'm wrong. Thanks!
> 
> To use `flatten` you will need to call `self.ivy_array.flatten()`

Thank you so much :). the Pytest test passed after updating my branch.

.devcontainer/build/devcontainer.json
```diff
@@ -20,6 +20,7 @@
         "post_create": "bash .devcontainer/post_create_commands.sh", 
         "bashrc": "echo \"alias python=python3\" >> ~/.bashrc"
     },
+
```
hirwa-nshuti, line 23:
```suggestion
```

ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py
```diff
@@ -116,6 +116,34 @@ def test_numpy_ndarray_property_T(
     )
 
 
+# data generation using given decorator
```
hirwa-nshuti, line 119:
```suggestion
```

ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py
```diff
@@ -116,6 +116,34 @@ def test_numpy_ndarray_property_T(
...
+# data preprocessing: convert ivy.array to framework suitable array (np.ndarray())
+#
```
hirwa-nshuti, line 121:
```suggestion
```

ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py
```diff
@@ -881,6 +909,7 @@ def test_numpy_instance_diagonal(
     dtype_x_axis,
     offset,
     frontend_method_data,
+    init_all_as_kwargs_np,
```
hirwa-nshuti, line 912:
```suggestion
```

hirwa-nshuti (changes requested):
Now the tests are failing for paddle and torch backends would you look into that as well

AsmaaMHadir:
> Now the tests are failing for paddle and torch backends would you look into that as well

Yes. Thank you for your review!

ivy-leaves:

Thank you :hugs:

PR #15541, opened 2023-05-17
Update test_jax_src_tree_util.py

Fayad-Alman:
Your task is to write tests for the functions of this frontend module https://github.com/unifyai/ivy/tree/master/ivy/functional/frontends/jax/_src.
It would be ideal to use hypothesis as much as possible, but in complex cases you will probably have to resort to plain assertions.


Fayad-Alman:
@AnnaTz I apologize for having you go through so much trouble, but I'm having issues figuring out what's wrong. Mind taking a look?

PR #15645, opened 2023-05-18
Fixes Dropout's unintended behavior of automatic dtype change

rishabgit:
https://trello.com/c/AmIDttaV/1834-fix-dropouts-automatic-dtype-changing-behaviour

Also handles error arising when user is on GPU and passes input on CPU due to temp vars being incorrectly created on GPU

Following code piece earlier threw two issues
```
import ivy

ivy.set_torch_backend()

x = ivy.random_normal(shape=(100, 1000), dtype='float32', device='cpu')

ivy.dropout(x, 0.2, dtype='float32')
```
1. Output was float64 even after being explicitly set to float32
2. Couldn't handle when input device is CPU and threw following error - 
```
IvyBackendException: torch: dropout: torch: where: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

### DIFF
--- a/ivy/functional/ivy/layers.py
+++ b/ivy/functional/ivy/layers.py
@@ -340,6 +340,7 @@ def dropout(
         0,
         1,
     )
+    mask = ivy.astype(mask, dtype)
     x = x * mask
     if scale:
         x = ivy.multiply(x, 1 / (1 - prob), out=out)
--- a/ivy/functional/ivy/layers.py
+++ b/ivy/functional/ivy/layers.py
@@ -340,7 +340,10 @@ def dropout(
         0,
         1,
     )
-    mask = ivy.astype(mask, dtype)
+    if dtype is None:
+        mask = ivy.astype(mask, x.dtype)
+    else:
+        mask = ivy.astype(mask, dtype)
     x = x * mask
     if scale:
         x = ivy.multiply(x, 1 / (1 - prob), out=out)
--- a/ivy/functional/ivy/layers.py
+++ b/ivy/functional/ivy/layers.py
@@ -337,13 +337,9 @@ def dropout(
     mask = ivy.where(
         ivy.random_uniform(shape=noise_shape, device=ivy.dev(x), dtype=dtype, seed=seed)
         < prob,
-        0,
-        1,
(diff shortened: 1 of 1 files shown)


rishabgit:
WIP, found another device related issue

VedPatwardhan (commented):
Hey @rishabgit, this issue isn't with the `where` function but instead related to the recent [changes](https://github.com/unifyai/ivy/commit/93f204dcbaa41b5dae2e2694c799e72f0edd1382) to the type promotion table, where `int32` and `float32` in an operation get promoted to `float64` when we do `x = x * mask`. I'll also bring @MahmoudAshraf97 into the discussion as this is something he was working on.
We can do one of 2 things : 
1. Assume `float64` to be the correct output dtype
2. Change the type promotion table back (i.e. away from `numpy` and closer to `torch` and `jax`)

Happy to hear your thoughts on this @rishabgit @MahmoudAshraf97. Thanks 🙂

MahmoudAshraf97:
Hi Everyone, the problem is indeed in the where function, to solve it write the 1 and 0 as 1. And 0.
This will initialize them with float32 instead of int32 and the output will be with the correct dtype

VedPatwardhan:
> Hi Everyone, the problem is indeed in the where function, to solve it write the 1 and 0 as 1. And 0. This will initialize them with float32 instead of int32 and the output will be with the correct dtype

I'm not really sure about that, the `dropout` code hasn't changed for quite a while, and without the `(int32, float32) --> float64` type promotion in the table, it generates a `float32` array with the older promotion rule 😅

MahmoudAshraf97:
@rishabgit please check this commit, it should fix the behaviour you are facing
https://github.com/unifyai/ivy/commit/8b3cac942341e5690f90e44031edd6b11417a525

rishabgit:
@MahmoudAshraf97 Doesn't seem to handle the second case correctly :( 

> 2\. Couldn't handle when input device is CPU and threw following error -

MahmoudAshraf97:
this needs to be handled in `ivy.asarray`, dropout has nothing to do with this problem as it will keep occuring in other places if the root cause isn't fixed

Issue #10696, opened 2023-02-20
Add Layers to Paddle Backend


MahmoudAshraf97:
Add [Layers](https://www.paddlepaddle.org.cn/documentation/docs/en/api/index_en.html) to Paddle backend

- [ ] conv1d
- [ ] conv1d_transpose
- [ ] conv2d
- [ ] conv2d_transpose
- [ ] depthwise_conv2d
- [ ] conv3d
- [ ] conv3d_transpose
- [ ] conv_general_dilated
- [ ] conv_general_transpose

## Experimental

- [ ] interpolate

AdvancedUno:
@MahmoudAshraf97 Hi! I am new to Ivy open source and want to contribute to building conv1d Function. Is there basic format or order does Ivy want it to be done?
I hope I can be helpful to Ivy community!

aymuos15:
- #11338

aymuos15:
- [ ] #11338

AdvancedUno:
- [ ] #12361

umer200:
- [ ] #13181

LontraX:
#13323

umer200:
- [ ] #13577

rogn02:
- [ ]#12361

trantuantdt:
- [ ]  #13920

rogn02:
- [ ] #12361

SATYAMYADAV118:
#14648

mosesdaudu001:
#14847

rum1887:
hi, I'm working on conv1d_transpose and conv2d :)

rum1887:
- [conv2d] #15697

VaishnaviMudaliar:
- [ ] #17147

m-gopichand:
#17441

melch-inno:
#17649

yopknopixx:
- [ ] #18313

erimdogan:
#19545

gabrielmfern:
- [ ] #19871

yopknopixx:
- [ ] #20896

yopknopixx:
- [ ] #20897

yopknopixx:
- [ ] #20898

Praneet9:
> * [ ]  [conv3d_transpose #20897](https://github.com/unifyai/ivy/issues/20897)

@MahmoudAshraf97 Can you assign me the issue?

suchunkang0822:
- [ ] #21213

[...]

PR #15697, opened 2023-05-19
conv2d

rum1887:
Add Layers to Paddle Backend #10696

### DIFF
--- a/ivy/functional/backends/paddle/layers.py
+++ b/ivy/functional/backends/paddle/layers.py
@@ -115,13 +115,13 @@ def conv2d(
     padding: Union[str, int, Sequence[Tuple[int, int]]],
     /,
     *,
-    data_format: str = "NHWC",
+    data_format: str = "NCHW",
     dilations: Union[int, Tuple[int, int]] = 1,
     out: Optional[paddle.Tensor] = None,
 ):
-    if data_format == "NCHW":
-        x = paddle.transpose(x, perm=(0, 2, 3, 1))
-    df = "NHWC"
+    df = "NCHW"
+    if data_format == "NHWC":
+        df = "NHWC"
     x = _pad_before_conv(x, filters, strides, padding, 2, dilations, df)
     filters = paddle.transpose(filters, perm=(3, 2, 0, 1))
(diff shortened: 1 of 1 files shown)


rum1887:
hi, @MahmoudAshraf97 kindly review this pr :)

MahmoudAshraf97 (changes requested):
paddle already has data format argument, why not use it directly instead of the axis swap?

rum1887:
> paddle already has data format argument, why not use it directly instead of the axis swap?

yeah, so I initialled the data format to "NCHW" (PyTorch, Paddle has this format as the default whilst Tensorflow has "NHCW" as the default dataformat). 

I changed the code to perform axis swap once at the end.

works?

rum1887:
> > paddle already has data format argument, why not use it directly instead of the axis swap?
> 
> Yes!
> 
> I initialized the data format to "NCHW". PyTorch, Paddle has this format as the default whilst Tensorflow has "NHCW" as the default dataformat.
> 
> I changed the code to perform axis swap once at the end.
> 
> works?

umm, since Paddle produces output consistent with the input format passed, I removed the axis swaps.

ivy/functional/backends/paddle/layers.py
```diff
@@ -118,8 +118,22 @@ def conv2d(
     data_format: str = "NHWC",
     dilations: Union[int, Tuple[int, int]] = 1,
     out: Optional[paddle.Tensor] = None,
-) -> paddle.Tensor:
-    raise IvyNotImplementedException()
+):
+    df = "NHWC"
+    if data_format == "NCHW":
+        df = "NCHW"
```
MahmoudAshraf97, line 124: I still don't see a need for this, what does this snipped add?

rum1887:
> please use the hypothesis tests to make sure your implementation is correct before requesting another review

yes, will do that!

why are the axes of filters switched in conv3d?

ivy/functional/backends/paddle/layers.py
rum1887: yeah, its nit needed

PR #15705, opened 2023-05-19
Update linalg.py

LaraHossam:


ivy/functional/frontends/torch/linalg.py
```diff
@@ -195,6 +195,27 @@ def solve(input, other, *, out=None):
...
+
+    if check_errors and ivy.any(info != 0):
+        raise RuntimeError("Error occurred while solving the system.")
+
+    if out is not None:
+        assert isinstance(out, tuple) and len(out) == 2, "Invalid 'out' argument"
+        out[0] = result
+        out[1] = info
+
+    named_tuple = ivy.namedtuple("SolveExOutput", ["result", "info"])
```
juliagsy, line 215: this is not a valid function in ivy, but a function from the `collections` package, feel free to import it from the package (`from collections import namedtuple`)

juliagsy (changes requested):
Hello! Thanks for contributing! Please also write tests for the frontend function in its corresponding file in the `ivy_tests` folder, guide can be found here: https://unify.ai/docs/ivy/overview/deep_dive/ivy_frontends_tests.html
Thanks!

PR #15710, opened 2023-05-20
Add Multinomial Functionality to Paddle Backend

khdoex:
This pull request adds the multinomial random process function to the PaddlePaddle backend of the Ivy library.

Changes:

The multinomial function generates a batch of samples from a multinomial distribution, given the population size, number of samples, batch size, and optionally the probabilities for each population element. The function can also handle the cases where sampling is performed with or without replacement. This addition allows users to generate samples from a multinomial distribution using the Paddle backend, a functionality that was missing previously. The implementation has been tested to ensure that it behaves as expected and is compatible with the rest of the Ivy library.

Please let me know if you have any questions or feedback on the changes made in this pull request.

### DIFF
--- a/ivy/functional/backends/paddle/random.py
+++ b/ivy/functional/backends/paddle/random.py
@@ -91,7 +91,15 @@ def multinomial(
     seed: Optional[int] = None,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    raise IvyNotImplementedException()
+    
+    if seed is not None:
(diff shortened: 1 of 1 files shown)


MahmoudAshraf97 (changes requested):
Hi, you should use a functional approach instead of the current stateful approach
https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/multinomial_en.html#multinomial

PR #15769, opened 2023-05-22
fft2 #15557

### LINKED ISSUE #15557: fft2

rubada:
Close #15557


rubada:
@jkeane508
Hi,
I fix the device issue, but I have a "TypeError: fft2() got some positional-only arguments passed as keyword arguments: 'x'", x is a positional argument in all the function, is the problem with the test function?

Thank you

jkeane508:
> @jkeane508 Hi, I fix the device issue, but I have a "TypeError: fft2() got some positional-only arguments passed as keyword arguments: 'x'", x is a positional argument in all the function, is the problem with the test function?
> 
> Thank you

Its probably because you are passing the arguments to the test function as keyword arguments. This means that the test function is passing them as keywords when they should actually be positional arguments.

rubada:
@jkeane508
Hi
Please check the files I removed the /, please review the changes.
Thank you

jkeane508 (changes requested):
Hey,

It looks like the tests are returning different types here - https://github.com/unifyai/ivy/actions/runs/5149556698/jobs/9272707571?pr=15769#step:3:280

We should aim to have them returning the same dtype in each test case!

rubada:
@jkeane508
Hi I got this error (AttributeError: 'Array' object has no attribute 'fft2'), I checked the source and found out that the class _ArrayWithLayersExperimental(abc.ABC), doesn't have a method for the fft2, the class path is as follows:
[ivy/data_classes/array/experimental/layers.py](https://github.com/unifyai/ivy/blob/master/ivy/data_classes/array/experimental/layers.py)

So what I should do should I add the method or how I can I solve this issue?

Thank you

jkeane508:
> @jkeane508 Hi I got this error (AttributeError: 'Array' object has no attribute 'fft2'), I checked the source and found out that the class _ArrayWithLayersExperimental(abc.ABC), doesn't have a method for the fft2, the class path is as follows: [ivy/data_classes/array/experimental/layers.py](https://github.com/unifyai/ivy/blob/master/ivy/data_classes/array/experimental/layers.py)
> 
> So what I should do should I add the method or how I can I solve this issue?
> 
> Thank you

Seems to be that all of the other "fft" functions exist as layers here, so it probably makes sense that fft2 does aswell :)

rubada:
@jkeane508
Hi I did all the changes, I checked the 18 failing test and found out that they are not related to fft2, can you please check the changes.
Thanks

PR #15772, opened 2023-05-22
Refactor ivy.asarray in all backends

MahmoudAshraf97:
This is a refactor for `ivy.asarray` in tensorflow backend, it contains simplification and removal of redundancy in the code and fixes for errors when creating a tensor from an object that contains out-of-bounds numbers when the dtype is not explicitly stated
for example
```python
import ivy
ivy.set_backend("tensorflow")
ivy.asarray(ivy.finfo("float64").max)
## ValueError: Can't convert Python sequence with a value out of range for a double-precision float.
```
Testing TODOs
- [ ] test for `copy` argument if it actually works
- [ ] test for gradients
- [ ] add native arrays input
- [ ] fix nested inputs,  the current nesting setup produces invalid inputs as there might be a dimension mismatch


mattbarrett98 (commented):
This new implementation has broken the conversion of numpy arrays:
```
import ivy
import numpy as np
ivy.set_tensorflow_backend()
x = np.array([1.])
ivy.asarray(x)
```
now fails. Also the `shallow=False` should be in the `nested_map` call as it was before, otherwise the inputs get inplace updated creating strange behaviour (input lists being modified etc).

Clearly this function needs a refactor, but I also feel like drastically changing the function like has been done here will lead to many bugs and undo a lot of previous work (if I found 2 bugs by looking at it, I think there is likely to be another 10 we can't see 😅).

Maybe more wise for you to just change the parts you need to, the problematic `cast` calls, and leave a big refactor to someone who has worked extensively on this already

Infrared1029:
> This new implementation has broken the conversion of numpy arrays:
> 
> ```
> import ivy
> import numpy as np
> ivy.set_tensorflow_backend()
> x = np.array([1.])
> ivy.asarray(x)
> ```
> 
> now fails. Also the `shallow=False` should be in the `nested_map` call as it was before, otherwise the inputs get inplace updated creating strange behaviour (input lists being modified etc).
> 
> Clearly this function needs a refactor, but I also feel like drastically changing the function like has been done here will lead to many bugs and undo a lot of previous work (if I found 2 bugs by looking at it, I think there is likely to be another 10 we can't see 😅).
> 
> Maybe more wise to just change the parts you need to, the problematic `cast` calls, and leave a big refactor to someone who has worked extensively on this already

Not associated with this PR but kind of curious how come something as major as a numpy array failing is not caught during the `asarray` test

sherry30 (commented):
Besides the things Matthew pointed out, Looks good to me so far from what I've tested manually. But I could easily be missing some corner case. Would be better to make the test more extensive to make sure everything is working as expected.

MahmoudAshraf97:
@CatB1t I wrote a new test strategy to include testing for `dtype` argument, but when testing with a large number of examples e.g 100 there are failures I can' reproduce, can you review the strategy and test it if possible?

AnnaTz:
> @CatB1t I wrote a new test strategy to include testing for `dtype` argument, but when testing with a large number of examples e.g 100 there are failures I can' reproduce, can you review the strategy and test it if possible?

Hi @MahmoudAshraf97, out of coincidence I also noticed the `dtype` argument testing was missing and started looking into fixing this. I will leave this to you though since you are already looking into `asarray`. I just wanted to point out this will require refactoring the rest of the backends as well, so this PR would need to be extended a bit.

MahmoudAshraf97:
I thought about this too, but I'll get the testing done TF so I can have a
groundtruth before messing with the other backends

sherry30:
One case you might wanna consider:

```
ivy.set_backend("tensorflow")
sh = ivy.Shape([1, 2, 3])
arr = [4, 5, 6]
y = ivy.asarray([sh, sh])
y = ivy.asarray([[arr, arr], [sh, sh]])
```
This works on all other backends and also on tensorflow framework itself.
You can assume that any of the leaf list will either all be `ivy.Shape`, `ivy.Array`, or simple python data types. They cannot be mixed.
Something like `ivy.asarray([arr, sh])` isn't expected.

MahmoudAshraf97:
> One case you might wanna consider:
> 
> ```
> sh = ivy.Shape([1, 2, 3])
> arr = [4, 5, 6]
> y = ivy.asarray([sh, sh])
> y = ivy.asarray([[arr, arr], [sh, sh]])
> ```
> 
> This works on all other backends and also on tensorflow framework itself. You can assume that any of the leaf list will either all be `ivy.Shape`, `ivy.Array`, or simple python data types. They cannot be mixed. Something like `ivy.asarray([arr, sh])` isn't expected.

This now works as expected, there still is one problem when `dtype=None` as dtype can't be inferred for `ivy.Shape` objects, there are two ways to go about this:
1. Use a special case for `tf.TensorShape` as in the old implementation
2. Add a `dtype` property for `ivy.Shape` to return default int dtype or improve the dtype inference methods such as `ivy.default_dtype()` to handle this case

I prefer the second approach because I want to keep the TF implementation as simple and general as possible so we can possibly reuse it in other backends with minimal modifications and also because the case for shape objects dtype inference might occur else where

mattbarrett98:
> > One case you might wanna consider:
> > ```
> > ivy.set_backend("tensorflow")
> > sh = ivy.Shape([1, 2, 3])
> > arr = [4, 5, 6]
> > y = ivy.asarray([sh, sh])
> > y = ivy.asarray([[arr, arr], [sh, sh]])
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This works on all other backends and also on tensorflow framework itself. You can assume that any of the leaf list will either all be `ivy.Shape`, `ivy.Array`, or simple python data types. They cannot be mixed. Something like `ivy.asarray([arr, sh])` isn't expected.
> 
> This now works as expected, there still is one problem when `dtype=None` as dtype can't be inferred for `ivy.Shape` objects, there are two ways to go about this:
> 
> 1. Use a special case for `tf.TensorShape` as in the old implementation
> 2. Add a `dtype` property for `ivy.Shape` to return default int dtype or improve the dtype inference methods such as `ivy.default_dtype()` to handle this case
> 
> I prefer the second approach because I want to keep the TF implementation as simple and general as possible so we can possibly reuse it in other backends with minimal modifications and also because the case for shape objects dtype inference might occur else where

not sure about adding a dtype to `Shape` just for this, it isn't a tensor so giving it a dtype seems a bit strange. the other approaches seem better to me

MahmoudAshraf97:
Hi Everyone, the function is supposed to be ready for merging and all the tests are passing
@Ishticode can we work together to add more test cases to the testing function that hypothesis can't possibly cover such as
```python
ivy.set_backend("tensorflow")
sh = ivy.Shape([1, 2, 3])
arr = [4, 5, 6]
y = ivy.asarray([sh, sh])
y = ivy.asarray([[arr, arr], [sh, sh]])
```
@sherry30 and @VedPatwardhan , I'm requesting one last review for the PR and would like to hear if there's any comment

@mattbarrett98 can you run some compiler tests using this function to see if it introduces any new failures, I'd appreciate it if we could merge this before the release

VedPatwardhan (commented):
The changes look good to me, however let's not merge this PR until we've got all the new test cases added to `test_asarray` and till we get the tests passing with our refactor in this PR. Given that `asarray` is central to the API and the release is around the corner, we're better off merging a well-tested refactor. Thanks @MahmoudAshraf97 🙂

mattbarrett98 (commented):
all compiler tests are passing 👍

vedpatwardhan:
Thanks for adding the tests @Ishticode, could you please check if the refactor passes these tests as well @MahmoudAshraf97? Thanks 🙂

Ishticode:
Thanks @vedpatwardhan. Please let me know for any issues relating to this logic or if you need to expand further.

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -202,11 +202,32 @@ def _asarray_helper(draw):
...
     )
+    nested_structure = draw(st.booleans())
+    if nested_structure:
+        sh = draw(
+            helpers.lists(
+                x=st.integers(min_value=1, max_value=20), min_size=1, max_size=5
+            )
+        )
+        sh = ivy.Shape(sh)
+        np_array = np.arange(3).astype(x_dtype[0])
```
MahmoudAshraf97, line 213: I think we should generate the numpy array using hypothesis not a fixed np array

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -202,11 +202,32 @@ def _asarray_helper(draw):
...
+        ivy_array = ivy.array(x[0], dtype=x_dtype[0])
```
MahmoudAshraf97, line 214: we are generating the ivy array using the same function that we are testing, so I guess this might cause a loophole for failures to be missed

MahmoudAshraf97 (commented):
Hi @Ishticode , thanks for the testing code and I have some comments

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
Ishticode: In the meeting point @vedpatwardhan had mentioned ivy Array too as a leaf entry. Maybe he can comment on this.

Besides this @MahmoudAshraf97 please feel free to make such minor changes as you see fit. If there is an issue with overall nested structure creation logic then surely I will try to fix that. Thanks

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
Ishticode: sure

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
vedpatwardhan: Actually, the points were prepared by @sherry30, but I agree with @Ishticode, unless there's something fundamentally wrong in creating a single leaf array, then we probably shouldn't consider it as a loophole. And adding to that, I don't think there's any other appropriate way to create an `ivy.Array` either (as we shouldn't be using the constructor). Thanks @MahmoudAshraf97 and @Ishticode 🙂

MahmoudAshraf97: This point is valid if we are trying to check if using an `ivy.Array` input will cause an exception, but what if the array is created correctly but with wrong contents? this mistake is going to occur in the ground truth example and in the test result and thus no errors will be caught because the result matches the ground truth, btw ivy arrays are converted to native arrays before going into the function so what applies to ivy arrays also apply to native arrays, that's why I see that we should instead create native arrays, at least we can guarantee that the native array creation functions are working well as a ground truth

vedpatwardhan: Yeah but the same issue with creating an array with wrong contents also exists with a native array creation function too right? 
In fact given that the native array creation functions aren't exactly equivalent in terms of the way they deal with dtype inferring and default devices and stuff, we are at a larger risk of these inconsistencies if we treat the native functions as the ground truth. 
Ofc given that we're generating `np.ndarray` in our data generation with `dtype_and_values`, this might end up being consistent in terms of the dtype, but we'd still need some equivalent of an `asarray` function as our helper which calls the appropriate native array creation function. So probably we're better off just using `ivy.asarray` and if there's an inconsistency in array creation due to a problem in our backend implementation, that'll propagate through the tests anyway.
Apologies if I've not understood your point properly, feel free to point it out if that's the case. Thanks @MahmoudAshraf97 🙂

ivy_tests/test_ivy/test_functional/test_core/test_creation.py
```diff
@@ -208,11 +208,33 @@ def _asarray_helper(draw):
             shared_dtype=True,
         )
     )
+    x_list = ivy.nested_map(x, lambda x: x.tolist(), shallow=False)
+    sh = draw(helpers.get_shape(min_num_dims=1))
+    sh = ivy.Shape(sh)
+    np_array = x[0]
+    ivy_array = ivy.array(x[0], dtype=x_dtype[0])
```
MahmoudAshraf97, line 215: this caused problems because it's created in a backend that is different that the one being tested, will creating it inside the test function body before calling `helpers.test_function` help create it in the correct backend? and even so, the input will still be evaluated in both test and groundtruth backends so I don't know if dynamic backend will help in this case

PR #15796, opened 2023-05-23
adding max pool 2d in paddle backend

abdulasiraj:


ivy/functional/backends/paddle/experimental/layers.py
```diff
@@ -56,7 +58,59 @@ def max_pool2d(
     ceil_mode: bool = False,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    raise IvyNotImplementedException()
+    dtype = x.dtype
+    x = x.astype("float32")
```
MahmoudAshraf97, line 62: this is not correct, you should only change the dtype if the current dtype was unsupported

ivy/functional/backends/paddle/experimental/layers.py
```diff
@@ -56,7 +58,59 @@ def max_pool2d(
...
+
+    if isinstance(padding, int):
+        padding = [(padding,) * 2] * 2
+    elif isinstance(padding, tuple) and len(padding) == 1:
+        padding = [(padding[0],) * 2] * 2
+    elif isinstance(padding, tuple) and len(padding) == 2:
+        padding = [(padding[0],) * 2, (padding[1],) * 2]
+
+    if isinstance(padding, (tuple, list)):
+        ivy.utils.assertions.check_kernel_padding_size(kernel, padding)
```
MahmoudAshraf97, line 86: you can import this assertion only instead of importing the whole ivy library

MahmoudAshraf97 (changes requested):

PR #15800, opened 2023-05-23
WIP Adding support for `ivy.Array` and `ivy.NativeArray` inputs to the split functions

hirwa-nshuti:


hirwa-nshuti:
add_frontend_checklist

PR #15804, opened 2023-05-23
Fix flatten

hello-fri-end:


mattbarrett98 (commented):
I don't think `to_native_arrays_and_back` needs to be added to the list of decorators, I think this is handled already with `inputs_to_native_arrays` and `outputs_to_ivy_arrays` already being in the list

hello-fri-end:
> I don't think `to_native_arrays_and_back` needs to be added to the list of decorators, I think this is handled already with `inputs_to_native_arrays` and `outputs_to_ivy_arrays` already being in the list

The problem is that we add `to_native_arrays_and_back` to the primary implementations of mixed functions by looping through all the decorators in the `FN_DECORATORS` in the following way inside the `_wrap_function`: 

```
        for attr in FN_DECORATORS:
            if (hasattr(original, attr) and not hasattr(to_wrap, attr)) or (
                mixed_fn and attr == "to_native_arrays_and_back"
 ```
and since FN_DECORATORS didn't have `to_native_arrays_and_back`, an `ivy.array` was getting passed to the `flatten` function causing the error. I guess instead of adding `to_native_arrays_and_back`, we could add `inputs_to_native_arrays` and `outputs_to_ivy_arrays` since they're equivalent but I don't the harm see in keeping the former in the list as well. Happy to hear what you guys think @VedPatwardhan @mattbarrett98

VedPatwardhan (commented):
> > I don't think `to_native_arrays_and_back` needs to be added to the list of decorators, I think this is handled already with `inputs_to_native_arrays` and `outputs_to_ivy_arrays` already being in the list
> 
> The problem is that we add `to_native_arrays_and_back` to the primary implementations of mixed functions by looping through all the decorators in the `FN_DECORATORS` in the following way inside the `_wrap_function`:
> 
> ```
>         for attr in FN_DECORATORS:
>             if (hasattr(original, attr) and not hasattr(to_wrap, attr)) or (
>                 mixed_fn and attr == "to_native_arrays_and_back"
> ```
> 
> and since FN_DECORATORS didn't have `to_native_arrays_and_back`, an `ivy.array` was getting passed to the `flatten` function causing the error. I guess instead of adding `to_native_arrays_and_back`, we could add `inputs_to_native_arrays` and `outputs_to_ivy_arrays` since they're equivalent but I don't the harm see in keeping the former in the list as well. Happy to hear what you guys think @VedPatwardhan @mattbarrett98

Hey @hello-fri-end, let's make further changes to the `_wrap_function` to wrap with the `inputs_to_native_arrays` and `outputs_to_ivy_arrays` individually, similar to how we used to do [before](https://github.com/unifyai/ivy/blob/c05e09814d8f316065e4273174a1c80969c44efb/ivy/func_wrapper.py#L978). As @mattbarrett98 said, we shouldn't be adding it to the `FN_DECORATORS` as `to_naitve_arrays_and_back` is a compostion of `inputs_to_native_arrays` and `outputs_to_ivy_arrays`. 
Of course, given that we've done quite a few changes to the `_wrap_function` recently, we'll have to introduce this change while maintaining the current approach (i.e. without the need of a `mixed_function` attribute. Feel free to make whatever changes you think are required for the same and request a review once done. 
Thanks @hello-fri-end and @mattbarrett98 🙂

hello-fri-end:
Hey @VedPatwardhan, I have made the changes as discussed today. The error on the CI - some of them are unrelated to the PR and I can't reproduce the others locally. E.g `test_torch_instance_index_add`, `test_torch_instance_index_add` and `test_torch_instance_flatten,`  are all passing locally but failing on the CI. What should I do about these?

VedPatwardhan (changes requested):
Hey @hello-fri-end, thanks for making the changes. Just one minor change left, also could you please confirm if the frontend test failure isn't related to the PR by running with a larger number of examples / running the failing example in isolation? Just wanted to be sure as it's failing the `flatten` frontend function which depends on `ivy.flatten`. Thanks 🙂

hello-fri-end:
ivy-gardener

hello-fri-end:
Hey @VedPatwardhan, the error was unrelated to this PR. Actually, the frontend `torch.greater` function had not been placed correctly inside the `tensor.py` file and therefore `Tensor object has no attribute greater` error was thrown. Because of this `test_torch_instance_flatten` was also not getting found. I had commented the test for `torch.greater` locally but have now fixed the issue with it. The relevant tests are passing now :)

VedPatwardhan (approved):
lgtm! Feel free to merge if the tests are passing. Thanks for looking into this issue @hello-fri-end 🙂

PR #15829, opened 2023-05-24
Fixes ivy.functional import problem

rishabgit:
https://trello.com/c/glVVNj4I/1777-fix-ivyfunctional-import-problem

Handles the edge case defined in the card

### DIFF
--- a/ivy/utils/backend/handler.py
+++ b/ivy/utils/backend/handler.py
@@ -259,6 +259,8 @@ def _set_backend_as_ivy(
         target.__dict__[k] = _wrap_function(
             key=k, to_wrap=backend.__dict__[k], original=v, compositional=compositional
         )
+        if isinstance(v, types.FunctionType) and "ivy.functional.ivy" in v.__module__:
+            ivy.functional.__dict__[k] = target.__dict__[k]
         if (
             isinstance(v, types.ModuleType)
             and "ivy.functional." in v.__name__
@@ -440,7 +442,6 @@ def set_backend(backend: str, dynamic: bool = False):
         global ivy_original_dict
         if not backend_stack:
             ivy_original_dict = ivy.__dict__.copy()
-            ivy_framework_dict = ivy.functional.__dict__.copy()
 
         _clear_current_sub_backends()
         if isinstance(backend, str):
@@ -457,7 +458,6 @@ def set_backend(backend: str, dynamic: bool = False):
         backend_stack.append(backend)
         set_backend_to_specific_version(backend)
         _set_backend_as_ivy(ivy_original_dict, ivy, backend)
-        _set_backend_as_ivy(ivy_framework_dict, ivy.functional, backend)
 
         if dynamic:
             convert_from_numpy_to_target_backend(variable_ids, numpy_objs, devices)
--- a/ivy/utils/backend/handler.py
+++ b/ivy/utils/backend/handler.py
@@ -259,8 +259,6 @@ def _set_backend_as_ivy(
         target.__dict__[k] = _wrap_function(
             key=k, to_wrap=backend.__dict__[k], original=v, compositional=compositional
         )
-        if isinstance(v, types.FunctionType) and "ivy.functional.ivy" in v.__module__:
-            ivy.functional.__dict__[k] = target.__dict__[k]
         if (
             isinstance(v, types.ModuleType)
             and "ivy.functional." in v.__name__
@@ -458,6 +456,11 @@ def set_backend(backend: str, dynamic: bool = False):
         backend_stack.append(backend)
         set_backend_to_specific_version(backend)
         _set_backend_as_ivy(ivy_original_dict, ivy, backend)
+        # following snippet is required to update the ivy.functional namespace with 
+        # backend-specific functions 
+        for key, _ in ivy.__dict__.items():
+            if key in ivy.functional.__dict__ and not key.startswith('__'):
+                ivy.functional.__dict__[key] = ivy.__dict__[key]
 
         if dynamic:
             convert_from_numpy_to_target_backend(variable_ids, numpy_objs, devices)
--- a/ivy/utils/backend/handler.py
+++ b/ivy/utils/backend/handler.py
@@ -456,10 +456,10 @@ def set_backend(backend: str, dynamic: bool = False):
         backend_stack.append(backend)
         set_backend_to_specific_version(backend)
         _set_backend_as_ivy(ivy_original_dict, ivy, backend)
-        # following snippet is required to update the ivy.functional namespace with 
-        # backend-specific functions 
+        # following snippet is required to update the ivy.functional namespace with
+        # backend-specific functions
         for key, _ in ivy.__dict__.items():
-            if key in ivy.functional.__dict__ and not key.startswith('__'):
+            if key in ivy.functional.__dict__ and not key.startswith("__"):
                 ivy.functional.__dict__[key] = ivy.__dict__[key]
 
         if dynamic:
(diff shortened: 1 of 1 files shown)


hello-fri-end:
Hey @rishabgit, thanks for working on this issue! Your solution seems to work as expected but just to be sure that this doesn't break anything, I'm tagging @CatB1t and @VedPatwardhan as they're more familiar with this :rocket:

VedPatwardhan (changes requested):
Hey @rishabgit, thanks for making the changes. I think this is a problem not just for mixed functions but for primary functions too. The proposed solution wouldn't be ideal as we are updating the global `ivy` at line `263`, which would break `with_backend`. Digging further into the issue, it seems like the update to `ivy.functional.ivy` is being made properly, but we aren't storing the updated dict in the target dict at the highest level (i.e. for the `functional` key).
I tried to find a clean solution, but the only one which I found efficient enough (so that we only traverse the functional submodules in the recursive call) is if we store a ref to the `ivy_pack` (i.e. the `target` at the first call to `_set_backend_as_ivy`) as another argument in the function and assign the `functional` key when applicable.
The implementation I came up with looks something like, 
```
def _set_backend_as_ivy(
    original_dict, target, backend, ivy_pack=None, invalid_dtypes=None, backend_str=None
):
    ivy_pack = target if ivy_pack is None else ivy_pack
    invalid_dtypes = (
        backend.invalid_dtypes if invalid_dtypes is None else invalid_dtypes
    )
    backend_str = backend.current_backend_str() if backend_str is None else backend_str
    for k, v in original_dict.items():
        if k not in backend.__dict__:
            if k in invalid_dtypes and k in target.__dict__:
                del target.__dict__[k]
                continue
            backend.__dict__[k] = v
        target.__dict__[k] = _wrap_function(
            key=k, to_wrap=backend.__dict__[k], original=v, compositional=compositional
        )
        if (
            isinstance(v, types.ModuleType)
            and "functional" in v.__name__
            and os.path.join("{}", "__init__.py").format(backend_str) not in v.__file__
        ):
            _set_backend_as_ivy(
                v.__dict__,
                target.__dict__[k],
                backend.__dict__[k],
                ivy_pack=ivy_pack,
                invalid_dtypes=invalid_dtypes,
                backend_str=backend_str,
            )
        if k == 'functional':
            ivy_pack.__dict__[k] = target
```
Feel free to make improvements to make it cleaner (as I don't think the one I've shared is the cleanest but it works) @rishabgit. Also happy to know your thoughts @CatB1t and @hello-fri-end. Thanks all 🙂

rishabgit:
Hey @VedPatwardhan, thanks a lot for looking into this :) 
I'm getting a bit of peculiar results with your code snippet - 
image

Not too sure what could be the main reason. Looks like something to do with the wrapping and how import_module does dynamic imports. Nothing too conclusive, so would do a deep dive into this soon, posting here in the off case you've seen this before

@hello-fri-end Noticed the band-aid fixes you made on your fork (ivy_tests/test_ivy/helpers/testing_helpers.py) as a workaround until this gets merged, so I ran pytest on that file (without your changes) and it seems to work fine. Does this importlib output mismatch affect the tests? Lemme know if there's any other test I should check out apart from testing_helpers

Here's the testing code if anyone would like to check this out quickly - 
```
import ivy
import importlib

fn_name = 'layer_norm'
ivy.set_backend('torch')
test = importlib.import_module('ivy.functional')
print(test.__dict__[fn_name].__module__)
print(ivy.functional.__dict__[fn_name].__module__)
test1 = importlib.import_module('ivy')
print(test1.__dict__[fn_name].__module__)
```

VedPatwardhan:
> Hey @VedPatwardhan, thanks a lot for looking into this :) I'm getting a bit of peculiar results with your code snippet - image
> 
> Not too sure what could be the main reason. Looks like something to do with the wrapping and how import_module does dynamic imports. Nothing too conclusive, so would do a deep dive into this soon, posting here in the off case you've seen this before
> 
> @hello-fri-end Noticed the band-aid fixes you made on your fork (ivy_tests/test_ivy/helpers/testing_helpers.py) as a workaround until this gets merged, so I ran pytest on that file (without your changes) and it seems to work fine. Does this importlib output mismatch affect the tests? Lemme know if there's any other test I should check out apart from testing_helpers
> 
> Here's the testing code if anyone would like to check this out quickly -
> 
> ```
> import ivy
> import importlib
> 
> fn_name = 'layer_norm'
> ivy.set_backend('torch')
> test = importlib.import_module('ivy.functional')
> print(test.__dict__[fn_name].__module__)
> print(ivy.functional.__dict__[fn_name].__module__)
> test1 = importlib.import_module('ivy')
> print(test1.__dict__[fn_name].__module__)
> ```

Yeah I've seen this before, would be great if you could dive into this to see if there's an elegant approach to fix this issue. I would've suggested doing it in a different PR, but I think it's good if we fix the problem before we merge this one, as there's additional namespaces being overriden in this one. Great point, thanks @rishabgit 🙂

hello-fri-end:
> Hey @VedPatwardhan, thanks a lot for looking into this :) I'm getting a bit of peculiar results with your code snippet - image


```python
import ivy
import importlib

ivy.set_backend("torch")
print(importlib.import_module('ivy'))
print(importlib.import_module('ivy.functional'))
```
prints 
```
<module 'ivy.functional.backends.torch' from '/workspaces/ivy/ivy/functional/backends/torch/__init__.py'>
<module 'ivy.functional' from '/workspaces/ivy/ivy/functional/__init__.py'>
```
Ideally, the second output should be:
``` <module 'ivy.functional.backends.torch' from '/workspaces/ivy/ivy/functional/backends/torch/__init__.py'>```

`importlib.import_modules` searches for the modules inside the `sys.modules` dict. So, when `ivy` is imported, all the `ivy` modules get added to the `sys.modules` dict as can be seen from the output below:
```python
sys.modules['ivy']
>>> <module 'ivy' from '/workspaces/ivy/ivy/__init__.py'>
sys.modules['ivy.functional']
>>> <module 'ivy.functional' from '/workspaces/ivy/ivy/functional/__init__.py'>
```
When the backend is set, `ivy` correctly gets updated with corresponding the backend module but `ivy.functional` does not:

```python
sys.modules['ivy']
>>> <module 'ivy.functional.backends.torch' from '/workspaces/ivy/ivy/functional/backends/torch/__init__.py'>
sys.modules['ivy.functional']
>>> <module 'ivy.functional' from '/workspaces/ivy/ivy/functional/__init__.py'>
```
So, we need to make changes such that `ivy.functional` points to the correct module after the backend is set. However, I'm not sure where this change needs to be made. Maybe @VedPatwardhan @CatB1t have a better idea.

> @hello-fri-end Noticed the band-aid fixes you made on your fork (ivy_tests/test_ivy/helpers/testing_helpers.py) as a workaround until this gets merged, so I ran pytest on that file (without your changes) and it seems to work fine. Does this importlib output mismatch affect the tests? Lemme know if there's any other test I should check out apart from testing_helpers

Yes, it creates a problem for mixed functions. The reason for this is that when we import `ivy.functional` using `importlib.import_modules` and get access to the function using its dict, it always returns the compositional implementation even in cases when the backend does have a native implementation. Once we have solution for this problem, we won't need my hacky bandaid fix :)

VedPatwardhan:
I think the issue is with `_set_backend_as_ivy` wherein we're doing these updates recursively through all namespaces, given that `functional` is the key at the highest level (in terms of the namespace hierarchy), the helper is not getting called to update that namespace. The bug needs to be fixed in the same function. But we'll also have to consider the issue which @rishabgit pointed out about re-importing. Let's probably have a chat on Monday about this, and we can then resume back on the PR thereafter. Thanks guys 🙂

PR #15830, opened 2023-05-24
Jax numpy size

VictorOdede:


VictorOdede:
Close #14769

vaatsalya123:
Hey @VictorOdede  ,Even though the implementation seems correct for the function the testing logic seems super simple , I would suggest you make the test code a bit more rigors so we can properly test this function, Thanks!

VictorOdede:
Thanks @vaatsalya123. What approach would you suggest to make the tests more rigorous?

vaatsalya123:
This looks fine @VictorOdede will merge , Thanks alot!

PR #15833, opened 2023-05-24
Fix ContextManager

CatB1t:
- Updated `set_backend` to return global Ivy
- Updated ContextManager to return value from `set_backend`, to allow usage as: `with ContextManager('tensorflow') as ivy_tf:`


VedPatwardhan (approved):
lgtm! Feel free to merge, thanks @CatB1t 🙂

PR #15837, opened 2023-05-24
Implementation of new Precise Mode

MahmoudAshraf97:
Hello, this is the implementation of the two different precision modes supported by numpy and (torch, jax)

Summary:

1. added new precise mode to switch between two promotion modes
2. used new promotion method in order to remove duplicate keys with different order in promotion table e.g. `(type1,type2) `and `(type2, type1)`, this will allow for a much shorter and deduplicated promotion table (122 vs 229 entries) which means simpler maintainance if needed, all backend and testing functionality that depends on promotions was updated accordingly


VedPatwardhan (changes requested):
Hey @MahmoudAshraf97, thanks for making the changes! Just wanted to request a couple of changes,
1. I think it would be great if we continued to have the promotion tables in the `__init__.py` file, as the `valid_dtypes`, `invalid_dtypes`, the `Dtype` class, etc. are present in that file, so would look more organized.
2. It would be quite efficient if we instead just computed the `ivy.promotion_table` once on importing `ivy` with the default precision mode (which is temporarily precise by default as discussed), and then only update the `ivy.promotion_table` on setting or unsetting the mode instead of a `get_promotion_table` function being called everytime we call `promote_types` (I'd imagine the frequency of calling `promote_types` is considerably higher than the frequency of changing the precision mode).

Happy to know your thoughts 🙂

ivy/__init__.py
```diff
@@ -599,244 +599,8 @@ class Node(str):
 invalid_uint_dtypes = ()
 invalid_complex_dtypes = ()
 
-# data type promotion
```
VedPatwardhan, line 602: Can we please move the updated promotion tables to this line? A bit OCD but would be easier to come back to these changes if only the updated parts are tracked by git (`array_api_promotion_table` is still the same)

ivy/__init__.py
```diff
@@ -1378,3 +1143,173 @@ def cast_data_types(val=True):
...
+    (uint8, uint32): uint32,
+    (float64, float64): float64,
+    (float16, float32): float32,
+}
+
+# the extra promotion table follows numpy safe casting convention
+# the following link discusses the different approaches to dtype promotions
+# https://jax.readthedocs.io/en/latest/jep/9407-type-promotion.html
+common_extra_promotion_table = {
+    (float32, complex128): complex128,
```
VedPatwardhan, line 1199: Just to be sure, we're also explicitly maintaining the reverse op too right? Maybe let's remove the duplicates in the promotion table in a separate PR after discussing with @djl11

ivy/functional/ivy/data_type.py
```diff
@@ -538,9 +538,13 @@ def can_cast(
         b: true
     }
     """
-    if isinstance(from_, ivy.Dtype):
-        return (from_, to) in ivy.promotion_table
-    return (from_.dtype, to) in ivy.promotion_table
+    if isinstance(from_, (ivy.Array, ivy.NativeArray)):
```
VedPatwardhan, line 541: The older approach seemed a bit cleaner actually, unless there's a reason except for the reverse promotions, we could just revert this changes if that's okay?

ivy/functional/ivy/data_type.py
```diff
@@ -1992,15 +1996,14 @@ def promote_types(
     ret
         The type that both input types promote to
     """
+    query = [ivy.as_ivy_dtype(type1), ivy.as_ivy_dtype(type2)]
```
VedPatwardhan, line 1999: As mentioned before, let's look into the duplicates in the promotion table in a different PR along with the additional logic for sorting

VedPatwardhan (changes requested):
Hey @MahmoudAshraf97, just a final couple of changes so that we're good to merge. Thanks 🙂

PR #15852, opened 2023-05-24
equal frontend function

mustafaa7med:
equal #15849


mustafaa7med:
Hello @Dsantra92,

I've looked into the 2 errors and i assume it is unrelated to the function, I assumed it is an import fail ```statistical_dtype_values```
Kindly advice

Thanks in advance,
Moustafa

mustafaa7med:
Hello @paulaehab,

I would be grateful if you could review my PR.

Thanks in advance,
Moustafa

ivy/functional/frontends/torch/tensor.py
```diff
@@ -380,6 +380,12 @@ def not_equal(self, other, *, out=None):
 
     ne = not_equal
 
+    @with_unsupported_dtypes({"2.0.1 and below": ("float16", "bfloat16")}, "torch")
+    def equal(self, other, *, out=None):
+        return torch_frontend.equal(self, other, out=out)
+
+    eq = equal
```
paulaehab, line 387: you don't need to add this as it must be added in separated issue

ivy/functional/frontends/torch/tensor.py
```diff
@@ -380,6 +380,12 @@ def not_equal(self, other, *, out=None):
 
     ne = not_equal
 
+    @with_unsupported_dtypes({"2.0.1 and below": ("float16", "bfloat16")}, "torch")
+    def equal(self, other, *, out=None):
```
paulaehab, line 384: equal function don't take this arguments 😄  please check the torch docs and change it after that be sure that all tests is passing

paulaehab (changes requested):

paulaehab (changes requested):
after you correct changes and all tests pass I will merge it

mustafaa7med:
Hello @paulaehab,

Thank you for your input and help, I reviewed the parameters required for this function and adjusted them accordingly, Moreover, Thanks for the referencing, It got slipped from my mind.

Unfortunately, the PR keeps failing test(1) and I'm not entirely sure what seems to be the reason, Could you please advise?
https://github.com/unifyai/ivy/actions/runs/5088963429/jobs/9146020121?pr=15852#step:3:160

Thanks in advance,
Moustafa Ahmed

mustafaa7med:
@paulaehab Any advise?

paulaehab:
@mustafaa7med  Hello, the problem was because ivy.equal is elementwise but torch.equal is single output
so reduce all reduces it down to single True if every entry is True else False it solved by adding ivy.all around ivy.equal around ivy.equal in comparison.py I already add it for you and Ishtiaq help us here in the PR. 
I will merge it nice work 😄 .

PR #15862, opened 2023-05-25
fixing try except

abdulasiraj:
Fixed the exception being raised in try block.


ivy/utils/exceptions.py
```diff
@@ -3,27 +3,41 @@
...
 
 
-def _log_stack_trace_truncated(trace_mode, func_wrapper_trace_mode):
+def _log_stack_trace_truncated(trace_mode, func_wrapper_trace_mode, buffer=None):
     if trace_mode in ["frontend", "ivy"]:
-        print(
-            "<stack trace is truncated to {} specific files,".format(trace_mode),
-            "call `ivy.set_exception_trace_mode('full')` to view the full trace>",
-        )
+        if buffer is not None:
```
VedPatwardhan, line 16: For readability and easier debugging, let's just write to the buffer and not print at all in all the helpers, and we'd then only do the printing inside the `_custom_exception_handle` and just rename all helpers having `_print` to `_write`

ivy/utils/exceptions.py
```diff
@@ -177,6 +182,9 @@ def __init__(self, *messages, include_backend=False):
 
 
 def handle_exceptions(fn: Callable) -> Callable:
+
+    buffer = io.StringIO()
```
VedPatwardhan, line 186: Could you please refactor this so that we wouldn't be creating the buffer twice? We seem to also instantiate at line [`44`](https://github.com/unifyai/ivy/pull/15862/files#diff-97b4bf909646477810e1e9162329746a1fb59a7938988d918ee1873e9d872cdeR44)

ivy_tests/test_ivy/test_misc/test_exceptions.py
```diff
@@ -3,7 +3,7 @@
 import contextlib
 import pytest
 from ivy_tests.test_ivy.helpers.available_frameworks import available_frameworks
-
+import ivy.functional.frontends.torch as torch
```
VedPatwardhan, line 6: I don't think we should import the frontends for these tests, frontend-specific exception tests would be present in the frontend files, as the frontends would have their own exception classes, as directed by the corresponding framework. We should instead be using `ivy` functions inside the `test_trace_modes`

VedPatwardhan (changes requested):
Hey @abdulasiraj, thanks for making the changes. Just requested a couple more changes, feel free to re-request a review once done 🙂

ivy/utils/exceptions.py
abdulasiraj: I think, there's some misunderstanding. Handling custom hook is independent from handling exception. Like they can't access same buffer variable. So there must be a separate buffer for each. let me know if I'm missing something.
Thanks!

VedPatwardhan: Oh okay, I was thinking if there could be a way, but no worries if not!

VedPatwardhan (commented):
Hey @abdulasiraj, thanks for making the changes. I think we're good to go, but given that @juliagsy has been making a few changes to the exception handler recently, I'll also ping her to for her opinion. Thanks @juliagsy 🙂

PR #15872, opened 2023-05-25
Upgrade python version in dockerfile

bipinKrishnan:
The PR upgrades the python version in the dockerfile(Ivy development environment) from 3.8 to 3.10.

The changes are based on the task here:
https://trello.com/c/nhrCExhu

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM debian:buster
 WORKDIR /ivy
 ARG CLI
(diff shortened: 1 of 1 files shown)


VedPatwardhan (approved):
lgtm! Similar to the approach last time, let's get the PR merged and see if there are any issues (although I don't think there will be many). 
Thanks for looking into this @bipinKrishnan 🙂

bipinKrishnan:
Thanks Ved, hope there won't be too many issues😅.

PR #15882, opened 2023-05-25
added partition to numpy frontend.

antoniobisegna:
#15870

### DIFF
--- a/ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
+++ b/ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
@@ -341,3 +341,11 @@ def masked_select(input, mask, out=None):
 def take(input, index):
     input = ivy.reshape(input, (-1,))
     return ivy.gather(input, index, axis=0)
+
+
+@to_ivy_arrays_and_back
+def narrow(input, dim, start, length):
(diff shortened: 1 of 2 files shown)


khushi-411 (changes requested):
Hi, @antoniobisegna! Thanks for the PR.
The function looks good to me. But you also need to add tests to verify the results. If you need any help, please feel free to ping me. :)

Also, please undo the changes you made in the `.idea/` folder. That's not needed. Thanks!

antoniobisegna:
Hi @khushi-411, I made a mistake and I closed this PR by accident. I opened a new one (#16068) and Ivy assigned a new person to review it. Thanks!

antoniobisegna:
Hi @khushi-411, I don't know if you are able to help me but I would like to ask some questions about my other pull request (#16068). Could you review it? I guess you have a lot of Pull requests to check and I understand why I did not get any other reviews until now.  I would like to know if I need to make some changes before going on with my testing. Specifically, I am wondering (1) if the array_helper I used is the best option for me, since the others do not seem to fit for my case. (2) I also would like to know, since there are no corresponding attribute to the one I am working on if I have to write a backend (or Ivy)  attribute my self - using other Ivy functions. Thanks!

PR #15913, opened 2023-05-26
Add "WSL + Virtual Environment + VSCode" section to Setting Up docs

adityagandhamal:
Closes [this card](https://trello.com/c/Jy58U3tA/813-setting-up-wsl-virtual-environment-vscode).

### DIFF
--- a/docs/overview/contributing/setting_up.rst
+++ b/docs/overview/contributing/setting_up.rst
@@ -568,6 +568,67 @@ Ubuntu
 
 For windows users, the file path should be entered with "/" (forward-slashes), for other OS it would be the regular "\\" (back-slashes).
 
+WSL
+***
+ 
+It is understandable that working with computationally heavy tools like Docker and PyCharm is not always confortable for new developers.
+By utilizing WSL, you can run a Linux distribution on your Windows machine and in addition, venv is leveraged to create 
+isolated Python environments eliminating the need for a full-fledged containerization solution like Docker, and with VSCode being an appropriate alternative to PyCharm, 
+the steps explained below will help you in setting up a less resource-intensive Ivy environment.
+
+#. Install `WSL <https://learn.microsoft.com/en-us/windows/wsl/install>`_.
+#. Install `Visual Studio Code <https://code.visualstudio.com/>`_. 
+   You can follow `this guide <https://learn.microsoft.com/en-us/windows/wsl/tutorials/wsl-vscode>`_ to integrate WSL into VSCode.
+#. Open the WSL terminal by typing in the name of your Linux distribution in the windows start menu (e.g. :code:`Ubuntu`).
+#. Create a virtual environment by following the steps below:
+
+   a. Install the python virtual environment package :code:`venv`.
+
+      .. code-block:: none
+
+         sudo apt install python3-venv
+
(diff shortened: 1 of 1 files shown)


V-G-spec:
Looks good to me, but unsure about the functionality. @zaeemansari70 can you take a look as well, and if all's good then mark the corresponding Trello card as done?
Thanks both :D

adityagandhamal:
Cc: @KareemMAX

docs/overview/contributing/setting_up.rst
```diff
@@ -694,6 +694,69 @@ Note: Currently you do not need to comment out the :code:`conftest.py` file in t
     </iframe>
 
 
+Setting up with WSL
+-------------------
+
+Here is another way for you to set up the whole Ivy environment on Windows using WSL, virtual environment and VSCode.
```
zaeemansari70, line 700: Also please remove this first line, we don't need to specify this type of stuff here. :slightly_smiling_face: 
`
Here is another way for you to set up the whole Ivy environment on Windows using WSL, virtual environment and VSCode. `

zaeemansari70 (changes requested):
Hi @adityagandhamal ,

This is not related to `Codespaces`, you have added it under the `Codespaces` section.
This should be under the `Setting up for Free` section https://github.com/unifyai/ivy/blob/3aa0df60585f6eb35250b51864a432a3eb760d71/docs/overview/contributing/setting_up.rst?plain=1#L526 , and after the `Windows` and `Ubuntu` subsection, with the heading just named as `WSL`. 
After this subheading here :slightly_smiling_face:  https://unify.ai/docs/ivy/overview/contributing/setting_up.html#id8

Do let me know if you have any questions :slightly_smiling_face: 
Thanks!

adityagandhamal:
thanks for the review @zaeemansari70.
Absolutely, this is not related to the Codespaces at all and I have added this as a completely new section at the end, making sure it does not fall under any of the previous sections. The reason being:
1. Could be added under `Virtual environments - No Docker` but it discusses both PyCharm and VSCode, unlike the new section where we only stick to VSCode.

2. The `Setting Up for Free` section discusses docker thus contradicting the new section.

3. Also, IMO, the users would find it appealing to have an entire dedicated WSL guide.

Interested to know your thoughts on this!

zaeemansari70 (changes requested):
Hi @adityagandhamal , 

Thanks for explaining why you added it here, it makes sense that you are thinking of adding it as a separate heading. 

Actually the main reason would be that this is somewhat a perfect guide for a `setting up for free` section, as we are using `WSL` as well as `vscode` it is also way lighter on systems as well as free, also those sections are not at all dedicated specifically for `docker` thise are paid too(setting up an IDE with a trial account stuff).
Moreover, if it was a completely separate distribution it would've made sense to add it in a completely separate sections :slightly_smiling_face: 

adityagandhamal:
Fair enough, your point is valid. Though a lightweight setup, my thought process should have been inclined towards the free side of the stuff here. 
I will make the necessary changes accordingly.
Thank you @zaeemansari70

zaeemansari70 (approved):
@adityagandhamal,

Thanks for the changes! :slightly_smiling_face:

PR #15918, opened 2023-05-26
fixed wrong ivy function used for torch.equal

akshatvishu:
The current implementation for [torch.equal()](https://pytorch.org/docs/stable/generated/torch.equal.html) at  torch frontend is using [ivy.equal()](https://unify.ai/docs/ivy/docs/functional/ivy/elementwise/ivy.functional.ivy.elementwise.equal.html#equal) which is an element-wise equality check whereas the torch's `torch.equal()` is used to determine  the inputs are all equal and thus, we should [ivy.all_equal()](https://unify.ai/docs/ivy/docs/functional/ivy/general/ivy.functional.ivy.general.all_equal.html#all-equal).

Here, is an example displaying the same!
```python
### Imports ###
import ivy.functional.frontends.torch as torch_f
import ivy
import torch

### Array Creation ###
target = torch.tensor([[0, 1, 1, 0]], dtype=torch.float32)  
output = torch.tensor([[2.6, 6.2, 3.7, 5.3]], dtype=torch.complex64)
arr1 = ivy.array([[0, 1, 1, 0]], dtype=ivy.float32)  
arr2 = ivy.array([[2.6, 6.2, 3.7, 5.3]], dtype=ivy.complex64) 

### Function Calling ###
a = torch.equal(target, output)
b = torch_f.equal(target, output)
c = torch_f.eq(target, output)
d = torch.eq(target, output)
e = ivy.equal(arr1, arr2)
f =  ivy.all_equal(arr1, arr2)

### Printing ###
print(f"torch.equal output:  {a}")
print(f"torch_f.equal output:  {b}")
print(f"torch_f.eq output:  {c}")
print(f"torch.eq output:  {d}")
print(f" ivy.equal output: {e}")
print(f"ivy.all_equal output:{f}")

"""
torch.equal output:  False
torch_f.equal output:  ivy.frontends.torch.Tensor([[False, False, False, False]]) # Wrong !

torch.eq output:  tensor([[False, False, False, False]])
torch_f.eq output:  ivy.frontends.torch.Tensor([[False, False, False, False]])

ivy.equal output: ivy.array([[False, False, False, False]])
ivy.all_equal output:False
"""

```
Also , there is another `equal` function defined   at[ivy/functional/backends/torch/elementwise](https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/torch/elementwise.py#L94) . which calls `torch.eq()` ; hence, I was wondering as `torch.eq` is different from `torch.equal` then that function should be call `eq` and not `equal`


akshatvishu:
Hey ! @nathzi1505  Can you take a look at it to see if my hunch is correct or not?

nathzi1505:
I think what you mentioned is something looking for.
Let me just ping @juliagsy if she can may be comment on this.

akshatvishu:
> I think what you mentioned is something looking for. Let me just ping @juliagsy if she can may be comment on this.

I think some of it got fixed in this PR https://github.com/unifyai/ivy/pull/15852

juliagsy:
@nathzi1505 Hey! It's been awhile since I worked on the frontends, but I can't seem to remember working on the mentioned frontend functions, but apologies if it was me! 
In general, we should make sure that the frontends work the same as natives, so the changes here looks sensible and will be the fix! Feel free to proceed the review with the change so that the frontend and native are matching! Thanks!

nathzi1505:
Co-authored-by: nathzi1505 <41519676+nathzi1505@users.noreply.github.com>

PR #15920, opened 2023-05-26
Added paddle.tensor.math.log and test function

### LINKED ISSUE #15840: log

gc24pt:
Close #15840


gc24pt (commented):
Added paddle.tensor.math.log and test function

PR #15922, opened 2023-05-26
fix ivy.bernoulli for float probs/logits

### LINKED ISSUE #15475: ivy.bernoulli doesn't work, always throws an error.
### Bug Explanation

The [`ivy.bernoulli()`](https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/experimental/random.py#L257) function, which returns an array of probabilities or logits, doesn't work, it always throws error messages, so @akshatvishu and I did a few experiments.

Expectation: should return a tensor of probabilities, of the given the shape, each element being a sample drawn from a Bernoulli distribution, making it a tensor of 1s and 0s.

I did a couple of small experiments to see what is the issue with it:

### Tests in Native Ivy

```python
>>> ivy.bernoulli(probs=0.5, shape=(3, 3))
IvyBackendException: numpy: bernoulli: bernoulli() takes 1 positional argument but 2 positional arguments (and 5 keyword-only arguments) were given
>>> ivy.bernoulli(logits=0.5, shape=(3, 3))
IvyBackendException: numpy: bernoulli: bernoulli() missing 1 required positional argument: 'probs'
>>> ivy.bernoulli(probs=0.5)
IvyBackendException: numpy: bernoulli: bernoulli() takes 1 positional argument but 2 positional arguments (and 5 keyword-only arguments) were given
>>> ivy.bernoulli(logits=0.5)
IvyBackendException: numpy: bernoulli: bernoulli() missing 1 required positional argument: 'probs'
```

@akshatvishu did most of the heavy lifting: 

His quote:

> Yup, really hard to get the function to work, and even when the function runs for a respective back-end, I can't get same results as the original back-end implementation gives

### JAX Bernoulli

```python
>>> import ivy.functional.backends.jax as jb
>>> import jax
>>> key = jax.random.PRNGKey(0)
>>> shape = (3, 3)  # Example shape
>>> probs = jnp.array([0.5]) # Ivy Jax one dosen't work with float values for probs
>>> re = jax.random.bernoulli(key=key, p=probs, shape=shape)
>>> ivy.set_backend("jax")
>>> result = jb.bernoulli(seed=0, probs=probs, shape=shape)
>>> result
[ True]    # Ivy Jax Backend
>>> re
[[ True, False, True]   # Original Jax 
 [ True, False, True]
 [ True, True, False]]
```

for me `jb.bernoulli(seed=0, probs=probs, shape=shape)` didn't even run.

>Jax error seems to result from [this](https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/jax/experimental/random.py#L102)

```python
def bernoulli(
    probs: Union[float, JaxArray],
    *,
    logits: Optional[Union[float, JaxArray]] = None,
    shape: Optional[Union[ivy.NativeArray, Sequence[int]]] = None,
    device: Optional[jaxlib.xla_extension.Device] = None,
    seed: Optional[int] = None,
) -> JaxArray:
    if seed:
        rng_input = jax.random.PRNGKey(seed)
    else:
        RNG_, rng_input = jax.random.split(_getRNG())
        _setRNG(RNG_)
    if logits is not None:
        probs = jax.nn.softmax(logits, axis=-1)
    if not _check_shapes_broadcastable(shape, probs.shape):
        shape = probs.shape
    return to_device(jax.random.bernoulli(rng_input, probs, shape=shape), device)
```
> See even when we are specifying the `shape` we're always doing `shape = probs.shape` hence for a single value `probs = jnp.array[0.5]` we get a single result even tho we specify `shape = (3,3)` 

> It should be something like this

```python
if shape is None:
    if not _check_shapes_broadcastable(shape, probs.shape):
      shape = probs.shape
else:
  shape = shape
```

### Steps to Reproduce Bug

try any one of my experiments or any of @akshatvishu given in the code above,

for example: 

```python
>>> import ivy
>>> ivy.bernoulli(probs=0.5, shape=(3, 3))
IvyBackendException: numpy: bernoulli: bernoulli() takes 1 positional argument but 2 positional arguments (and 5 keyword-only arguments) were given
```
expected:
```python
>>> import ivy
>>> ivy.bernoulli(probs=0.5, shape=(3, 3))

[[1, 0, 0],
[0, 1, 1],
[1, 0, 1]]
```

### Environment

Windows 10 using PyCharm with a conda env.

### Ivy Version

v1.1.9

### Backend

- [ ] NumPy
- [ ] TensorFlow
- [ ] PyTorch
- [ ] JAX

### Device

CPU

Sam-Armstrong:
Close #15475

### DIFF
--- a/ivy/functional/backends/jax/experimental/random.py
+++ b/ivy/functional/backends/jax/experimental/random.py
@@ -112,7 +112,11 @@ def bernoulli(
         RNG_, rng_input = jax.random.split(_getRNG())
         _setRNG(RNG_)
     if logits is not None:
-        probs = jax.nn.softmax(logits, axis=-1)
-    if not _check_shapes_broadcastable(shape, probs.shape):
-        shape = probs.shape
+        if type(logits) not in (float, int):
+            probs = jax.nn.softmax(logits, axis=-1)
+        elif probs is None:
+            probs = logits
+    if type(probs) not in (float, int):
+        if not _check_shapes_broadcastable(shape, probs.shape):
+            shape = probs.shape
     return to_device(jax.random.bernoulli(rng_input, probs, shape=shape), device)
--- a/ivy/functional/backends/numpy/experimental/random.py
+++ b/ivy/functional/backends/numpy/experimental/random.py
@@ -93,7 +93,11 @@ def bernoulli(
     if seed is not None:
         np.random.seed(seed)
     if logits is not None:
-        probs = np.asarray(ivy.softmax(logits), dtype=dtype)
-    if not _check_shapes_broadcastable(shape, probs.shape):
-        shape = probs.shape
+        if type(logits) not in (float, int):
+            probs = np.asarray(ivy.softmax(logits), dtype=dtype)
+        elif probs is None:
+            probs = logits
+    if type(probs) not in (float, int):
+        if not _check_shapes_broadcastable(shape, probs.shape):
+            shape = probs.shape
     return np.asarray(np.random.binomial(1, p=probs, size=shape), dtype=dtype)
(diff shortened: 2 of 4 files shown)


ivy-seed:

Issue #15919, opened 2023-05-26
iterable


rum1887:

rum1887:
Iterable jax #15923

rum1887:
hi @hirwa-nshuti , can you review this? 
thanks!

hirwa-nshuti:
You can start working on this no issue

rum1887:
> hi @hirwa-nshuti , can you review this?
> thanks!

yeah, I have a pr

rum1887:
completed Iterable jax #15923

PR #15923, opened 2023-05-26
Iterable jax

rum1887:


rum1887:
iterable #15919

ivy/functional/frontends/jax/numpy/creation.py
```diff
@@ -243,3 +243,13 @@ def compress(condition, a, *, axis=None, out=None):
         )
     arr = arr[: condition_arr.shape[0]]
     return ivy.moveaxis(arr[condition_arr], 0, axis)
+
+
+@to_ivy_arrays_and_back
+def iterable(x):
```
hirwa-nshuti, line 249: Can we here assert if the given object has `__iter__` method this can be easily handled using `hasattr` function

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_creation.py
```diff
@@ -923,3 +923,28 @@ def test_jax_numpy_compress(
         a=arr[0],
         axis=ax,
     )
+    
+    
+@handle_frontend_test(
+    fn_tree="jax.numpy.iterable",
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes(), shared_dtype=True
```
hirwa-nshuti, line 931: We are generating only one array no need for `shared_dtype` and inside the `get_dtypes`  function you will have to specify the kind of dtypes to generate

hirwa-nshuti (changes requested):
Hey @rum1887 
Thanks for the PR would you check the comments below?

ivy/functional/frontends/jax/numpy/creation.py
rum1887: yeah, done!

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_creation.py
rum1887: done!

ivy/functional/frontends/jax/numpy/creation.py
```diff
@@ -243,3 +243,13 @@ def compress(condition, a, *, axis=None, out=None):
         )
     arr = arr[: condition_arr.shape[0]]
     return ivy.moveaxis(arr[condition_arr], 0, axis)
+
+
+@to_ivy_arrays_and_back
```
hirwa-nshuti, line 248: We only need to convert input to `ivy.Array` and as the returned values are Booleans we will used `input_to_ivy_array` decorator
This should work
```python
@inputs_to_ivy_arrays
def iterable(y):
    return hasattr(y, "__iter__") and y.ndim > 0
```

hirwa-nshuti (changes requested):
Hey
check the comment below and update the signature of the function as it is mentioned the value to be `y` not `x` from the documentation. https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.iterable.html

ivy/functional/frontends/jax/numpy/creation.py
rum1887: Hi, really appreciate your feedback. Will practice pushing cleaner code like yours

rum1887:
> Hey
> check the comment below and update the signature of the function as it is mentioned the value to be `y` not `x` from the documentation. https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.iterable.html

yeah, I have fixed it :)

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_numpy_creation.py
```diff
@@ -923,3 +923,28 @@ def test_jax_numpy_compress(
...
+    on_device,
+):
+    dtype, x = dtype_and_x
+    helpers.test_frontend_function(
+        input_dtypes=dtype,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        x=x[0],
```
hirwa-nshuti, line 948: This is supposed to be `y` as in the function signature
```suggestion
        y=x[0],
```

ivy/functional/frontends/jax/numpy/creation.py
```diff
@@ -243,3 +243,9 @@ def compress(condition, a, *, axis=None, out=None):
         )
     arr = arr[: condition_arr.shape[0]]
     return ivy.moveaxis(arr[condition_arr], 0, axis)
+
+
+@inputs_to_ivy_arrays
+def iterable(y):
+    return hasattr(y, "__iter__") or y.ndim > 0
```
hirwa-nshuti, line 250: You'll use `and` because `0-D` arrays are not iterable.

```python
import numpy as np
print(np.iterable(2)) # False
```

ivy/functional/frontends/jax/numpy/creation.py
```diff
@@ -243,3 +243,9 @@ def compress(condition, a, *, axis=None, out=None):
         )
     arr = arr[: condition_arr.shape[0]]
     return ivy.moveaxis(arr[condition_arr], 0, axis)
+
+
+@inputs_to_ivy_arrays
```
hirwa-nshuti, line 248: You will need to import this function wrapper in the top imports

hirwa-nshuti (changes requested):
Hey @rum1887 

So far so good😊
Check some comments I left below🙂

ivy/functional/frontends/jax/numpy/creation.py
rum1887: The hasattr for a 0-D array will return a False and thereby the iterable function will return a false.
https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.iterable.html says either a sequence or has an iterator method.

ivy/functional/frontends/jax/numpy/creation.py
rum1887: > You will need to import this function wrapper in the top imports

yes, added. Thanks

ivy/functional/frontends/jax/numpy/creation.py
hirwa-nshuti: It should be `and` because we convert the input to `ivy.Array` instance and the `ivy.Array` class has `__iter__` method which will always be true even for `0-D` arrays

rum1887: Oh, so there's no requirement to assert if the object has an iter method? Since all input objects are converted to an instance of ivy.array and will have an iter method

hirwa-nshuti: The reason to have this assertion is that they're objects that are scalars for example if we remove `hasattr` assertion the function will fail for scalar inputs because we will be checking the `ndim` of scalar inputs as the [`inputs_to_ivy_arrays`](https://github.com/unifyai/ivy/blob/e3797979e4298ce6b0cfb16bba7fa29ac9605965/ivy/functional/frontends/numpy/func_wrapper.py#LL326C1-L326C1) wrapper doesn't convert scalars to `ivy.Array` instances.

rum1887: we want the function to return False for scalar inputs and it does that

hirwa-nshuti: I explained why we need that `hasattr` assertion now that the tests are passing and I'm about to merge the PR.
Unless there is any other suggestion you want to add😊

rum1887: Okay, thank you very much for being patient and answering all my queries :)
Do you have suggestions on what issue to pick next? And in general to have a better understanding of the codebase.

hirwa-nshuti: 🤞ery task is an opportunity to explore the codebase I am sure you'll learn a lot with any item you pick🙂
All the best

PR #15961, opened 2023-05-27
added lu_solve function for solving systems of linear equations

moezAhmad:


moezAhmad:
[#15943](https://github.com/unifyai/ivy/issues/15943)

PR #16006, opened 2023-05-29
Update test_linalg.py( added matmul #15927)

ShreyanSircar:
In the provided code, I made modifications to add a new function called matmul to perform matrix multiplication. Here's a breakdown of the changes:

#Function matmul:

->I defined a new function called matmul that takes two matrices as input, mat1 and mat2.  ->Inside the matmul function, I first checked if the number of columns in mat1 is equal to the number of rows in mat2. If they are not equal, I raised a ValueError to indicate that matrix multiplication is not possible. ->Then, I created an empty result matrix, result, with the same number of rows as mat1 and the same number of columns as mat2.  ->Next, I used nested loops to iterate through the rows of mat1 and the columns of mat2 to perform the multiplication. ->Within the nested loops, I calculated the dot product of the current row in mat1 and the current column in mat2 using another loop.  ->The dot product was calculated by multiplying the corresponding elements in the row and column and summing them up.  ->Finally, I assigned the resulting dot product to the corresponding element in the result matrix.  ->Once the multiplication was complete, I returned the result matrix.

#Testing the matmul function:

->After defining the matmul function, I tested it by creating two sample matrices, matrix1 and matrix2, with appropriate dimensions.
 ->I then called the matmul function with these matrices as arguments and stored the result in the result_matrix variable. 
->Finally, I printed the result_matrix to verify the correctness of the matrix multiplication.

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_linalg.py
+++ b/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_linalg.py
@@ -869,3 +869,51 @@ def test_tensorflow_diag(
         v=x[0],
         k=k,
     )
+
+# matmul
+@handle_frontend_test(
+    fn_tree="tensorflow.matmul",
+    dtype_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("valid"),
+        shape=(3, 3),
+        num_arrays=2,
+        shared_dtype=True,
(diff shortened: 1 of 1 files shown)


paulaehab:
Great work @ShreyanSircar🙂, I found that tf.linalg.matmul have no test function, but next time please implement the function and it's test in one PR.
I will merge it.

Issue #16043, opened 2023-05-30
 asin


MuhammadNizamani:
Going to added this function   asin in paddle.tensor.tensor.Tenso.

MuhammadNizamani:
This function is added now this issue is solved

PR #16040, opened 2023-05-30
EHN: ACOSH added in paddle.tensor.tensor.Tensor class  from acosh #16007form list of function  #15115

MuhammadNizamani:
I have added acosh in Tensor class. Closed #16007

Issue #15115, opened 2023-05-09
Add `paddle.Tensor` Instance Methods to PaddlePaddle Frontend


hmahmood24:
Add [Tensor Instance Methods](https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/Tensor_en.html) to PaddlePaddle Frontend

> Please keep in mind that the proper way to link an issue to this list is to comment "- [ ] #issue_number" while the issue's title only includes the name of the function you've chosen.

Please note that the following instance methods are supposed to be added to the `paddle.tensor.tensor.Tensor` class.

- [ ] abs
- [ ] acos
- [ ] acosh
- [ ] add
- [ ] add_
- [ ] add_n
- [ ] addmm
- [ ] all
- [ ] allclose
- [ ] amax
- [ ] amin
- [ ] angle
- [ ] any
- [ ] argmax
- [ ] argmin
- [ ] argsort
- [ ] as_complex
- [ ] as_real
- [ ] asin
- [ ] asinh
- [ ] astype
- [ ] atan
- [ ] atanh
- [ ] backward
- [ ] bincount
- [ ] bitwise_and
- [ ] bitwise_not
- [ ] bitwise_or
- [ ] bitwise_xor
- [ ] bmm
- [ ] broadcast_shape
- [ ] broadcast_tensors
- [ ] broadcast_to
- [ ] bucketize
- [ ] cast
- [ ] ceil
- [ ] ceil_
- [ ] cholesky
- [ ] cholesky_solve
- [ ] chunk
- [ ] clear_grad
- [ ] clip
- [ ] clip_
- [ ] concat
- [ ] cond
- [ ] conj
- [ ] corrcoef
- [ ] cos
- [ ] cosh
- [ ] count_nonzero
- [ ] cov
- [ ] cross
- [ ] cumprod
- [ ] cumsum
- [ ] deg2rad
- [ ] diagonal
- [ ] diff
- [ ] digamma
- [ ] dist
- [ ] divide
- [ ] dot
- [ ] eig
- [ ] eigvals
- [ ] eigvalsh
- [ ] equal
- [ ] equal_all
- [ ] erf
- [ ] erfinv
- [ ] erfinv_
- [ ] exp
- [ ] exp_
- [ ] expand
- [ ] expand_as
- [ ] exponential_
- [ ] fill_
- [ ] fill_diagonal_
- [ ] fill_diagonal_tensor
- [ ] fill_diagonal_tensor_
- [ ] flatten
[...]

PR #16046, opened 2023-05-30
EHN: asin added in paddle.tensor.tensor.Tensor class, from issue asin #16043  and form list of function #15115

MuhammadNizamani:
I have added asin function it might conflict with PR #16040


soma2000-lang:
@MuhammadNizamani I think it would be unsupported dtypes.

MuhammadNizamani:
@soma2000-lang I was bit confuse about it so I checked file `paddle.tensor.tensor.math.py `  it has same method and it was using supported so I followed it because it was marge in main branch. Can you explain why it should unsupported in `paddle.tensor.tensor.Tensor`  but not in `paddle.tensor.tensor.Tensor`? 
Thanks for review.

soma2000-lang:
@MuhammadNizamani thanks for pointing out ,I will look into this,there might be some mistake. I think can you modify that to unsupported dtypes ?

MuhammadNizamani:
@soma2000-lang  I will modify this to unsupported dtypes. just so know sqrt, log_softmax and asin these are the method that uses supported in `paddle.tensor.tensor.math.py`.

soma2000-lang:
@MuhammadNizamani thanks for pointing out .After digging into this I found out that it will be supported only.I will be merging this.
Thanks for contributing😊

PR #16049, opened 2023-05-30
unsorted_segment_prod 

### LINKED ISSUE #16020: unsorted_segment_prod

tiago-falves:
Close #16020 

Here is a first implementation of the unsorted_segment_prod, I need a little help to understand why the tests are failing!

### DIFF
--- a/ivy/functional/frontends/paddle/tensor/creation.py
+++ b/ivy/functional/frontends/paddle/tensor/creation.py
@@ -66,3 +66,8 @@ def arange(start, end=None, step=1, dtype=None, name=None):
 @to_ivy_arrays_and_back
 def empty(shape, dtype=None):
     return ivy.empty(shape=shape, dtype=dtype)
+
+
+@to_ivy_arrays_and_back
+def eye(num_rows, num_columns=None, dtype=None, name=None):
+    return ivy.eye(num_rows, num_columns, dtype=dtype)
--- a/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_creation.py
+++ b/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_creation.py
@@ -302,3 +302,33 @@ def test_paddle_empty(
         shape=shape,
         dtype=dtype[0],
     )
+
+
+# eye
+@handle_frontend_test(
+    fn_tree="paddle.eye",
+    num_rows=helpers.ints(min_value=3, max_value=10),
+    num_columns=st.none() | helpers.ints(min_value=3, max_value=10),
+    dtypes=helpers.get_dtypes("valid", full=False),
+    test_with_out=st.just(False),
+)
+def test_paddle_eye(
+    *,
+    num_rows,
+    num_columns,
+    dtypes,
+    on_device,
+    fn_tree,
+    test_flags,
+    frontend,
+):
+    helpers.test_frontend_function(
+        input_dtypes=dtypes,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        num_rows=num_rows,
+        num_columns=num_columns,
+        dtype=dtypes[0],
+    )


ivy/functional/frontends/tensorflow/math.py
```diff
@@ -419,6 +419,21 @@ def unsorted_segment_sqrt_n(
     return x
 
 
+@to_ivy_arrays_and_back
+def unsorted_segment_prod(
+    data, segment_ids, num_segments, name="unsorted_segment_prod"
```
AnnaTz, line 424: Better to leave `name=None` as in the original tensorflow function.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
```diff
@@ -1382,6 +1382,36 @@ def test_tensorflow_unsorted_segment_sqrt_n(
     )
 
 
+# unsorted_segment_prod
+@handle_frontend_test(
+    fn_tree="tensorflow.math.unsorted_segment_prod",
+    data=helpers.array_values(dtype=ivy.int32, shape=(5, 6), min_value=1, max_value=9),
```
AnnaTz, line 1388: We should be testing for all supported cases, so it would be best to not limit the `shape` and to sample the `dtype` from all "valid" dtypes. 
I would suggest having a look at how other tests generate the input arrays.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
```diff
@@ -1382,6 +1382,36 @@ def test_tensorflow_unsorted_segment_sqrt_n(
...
+    segment_ids=helpers.array_values(
+        dtype=ivy.int32, shape=(5,), min_value=0, max_value=4
```
AnnaTz, line 1390: The ids could be either `int32` or `int64`, and we should test for all supported cases without hardcoding the `shape`. 
To make sure the size and values of the `segment_ids` match the `data` array, you will probably need to generate these in a `st.composite` helper function.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
```diff
@@ -1382,6 +1382,36 @@ def test_tensorflow_unsorted_segment_sqrt_n(
...
+    *,
+    data,
+    segment_ids,
+    frontend,
+    test_flags,
+    fn_tree,
+    on_device,
+):
+    helpers.test_frontend_function(
+        input_dtypes=[ivy.float32, ivy.int32],
```
AnnaTz, line 1404: `input_dtypes` should not be hardcoded. You should be passing the `data` and `segment_ids` dtypes from `@handle_frontend_test`.

ivy/functional/frontends/tensorflow/math.py
```diff
@@ -419,6 +419,21 @@ def unsorted_segment_sqrt_n(
...
+):
+    ivy.utils.assertions.check_equal(list(segment_ids.shape), [list(data.shape)[0]])
+    x = ivy.ones(tuple([num_segments] + list(data.shape[1:])))
+    count = ivy.ones((num_segments,))
+    for i in range(segment_ids.shape[0]):
+        x[segment_ids[i]] = x[segment_ids[i]] * data[i]
```
AnnaTz, line 430: The error indicates that the shape of `x[segment_ids[i]]` doesn't match the shape of `x[segment_ids[i]] * data[i]`.
I would suggest using the debugger to find out what's wrong.

ivy/functional/frontends/tensorflow/math.py
```diff
@@ -419,6 +419,21 @@ def unsorted_segment_sqrt_n(
...
+        count[segment_ids[i]] += 1
+    for j in range(num_segments):
+        x[j] = ivy.pow(x[j], 1.0 / count[j])
+    return x
```
AnnaTz, line 434: Also, I'm not sure why you are using `count`. It seems like you are averaging, but I think the function is supposed to calculate just the product.

AnnaTz (changes requested):

PR #16050, opened 2023-05-30
Numpy Around function correction

### LINKED ISSUE #15484: rounding.test_numpy_around

NicholusMagak:
This tries to fix #15484 which is related to #15292

### DIFF
--- a/ivy/utils/backend/handler.py
+++ b/ivy/utils/backend/handler.py
@@ -458,6 +458,11 @@ def set_backend(backend: str, dynamic: bool = False):
         backend_stack.append(backend)
         set_backend_to_specific_version(backend)
         _set_backend_as_ivy(ivy_original_dict, ivy, backend)
+        # following snippet is required to update the ivy.functional namespace with
+        # backend-specific functions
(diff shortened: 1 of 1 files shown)


NicholusMagak:
@Infrared1029 Please review my PR. Thank you.

NicholusMagak:
@Infrared1029 In the event you have some time, please review this PR.

Infrared1029:
Hey @NicholusMagak, the tests appear to be failing according the the CI, can you make sure they are passing so i can have a closer look? sorry for the late reply!

Infrared1029:
Sorry for the late reply, probably setting the test to only generate integers and floats should work.
`helpers.dtypes("float_and_integer")`, some backends do not support bools for the around function.

NicholusMagak:
@Infrared1029 Thank you for the reply. I assume you're busy hence I'm not usually worried about late replies. I made the changes you recommender and ended up having new errors. Reverted back to the way the file were and I could not get the errors to go down. Any other possible solution?

NicholusMagak:
Is it potentially because I am using a docker python image 3.8.10?

PR #16055, opened 2023-05-30
Update ivy.jac to support functions with multiple args

somaia02:


somaia02:
So before passing any multi-arg function, we should first wrap it in a function that takes a single argument right? 

This worked with `execute_with_gradients` but not `jac` because of the conversion to native arrays and back. 

I use the wrapper `to_native_and_back` instead of `ivy.to_native` and `ivy.to_ivy` because they don't iterate over lists.
I updated the code. Can you please take a look?
@VedPatwardhan

VedPatwardhan:
> So before passing any multi-arg function, we should first wrap it in a function that takes a single argument right?
> 
> This worked with `execute_with_gradients` but not `jac` because of the conversion to native arrays and back.
> 
> I use the wrapper `to_native_and_back` instead of `ivy.to_native` and `ivy.to_ivy` because they don't iterate over lists. I updated the code. Can you please take a look? @VedPatwardhan

I think we also do the conversion in `execute_with_gradients` which we can replicate for the `jac` function if that's the case. Feel free to use any of the applicable helpers from `execute_with_gradients` for the `jac` implementation.

Also, the `to_native` function should support lists if we pass `nested=True` given that we're using [`nested_map`](https://github.com/unifyai/ivy/blob/fdb1bdc8387504d067fa27e4fadfcd6de69a5f4e/ivy/data_classes/array/conversions.py#L148). The same is also applicable to `to_ivy`.

Thanks @somaia02 🙂

somaia02:
Thank you for being patient with me. This is a summary of the changes I made. I tried to compare with `execute_with_gradient` when possible.

1. Added nested=True to convert args list to native and back
2. Instead of wrapping the input function, I added `*` to unpack `xs` because torch and Paddle will attempt to unpack the `xs` if it's not a tensor. We don't need to do this in `execute_with_gradient` because we use torch.autograd.grad which doesn't call the function internally but torch.jacobian does.
3. Tensorflow jacobian doesn't accept multiple outputs at once, so I map over them using ivy.nested_map. We do a similar thing in tensorflow backend in  `execute_with_gradient`
4. Jax jacobian needs to explicitly specify which arguments to compute w.r.t. I used `argnums` to select all args. In  `execute_with_gradient` we don't need to do this because the function is wrapped so there is only one argument.
5. Since functions with multiple args can't be wrapped because of torch & paddle. I updated the test with new cases.

Thank you for reading through this. If I'm missing something or there is a better approach to solve any of these issues please tell me.
@VedPatwardhan

VedPatwardhan:
> Thank you for being patient with me. This is a summary of the changes I made. I tried to compare with `execute_with_gradient` when possible.
> 
> 1. Added nested=True to convert args list to native and back
> 2. Instead of wrapping the input function, I added `*` to unpack `xs` because torch and Paddle will attempt to unpack the `xs` if it's not a tensor. We don't need to do this in `execute_with_gradient` because we use torch.autograd.grad which doesn't call the function internally but torch.jacobian does.
> 3. Tensorflow jacobian doesn't accept multiple outputs at once, so I map over them using ivy.nested_map. We do a similar thing in tensorflow backend in  `execute_with_gradient`
> 4. Jax jacobian needs to explicitly specify which arguments to compute w.r.t. I used `argnums` to select all args. In  `execute_with_gradient` we don't need to do this because the function is wrapped so there is only one argument.
> 5. Since functions with multiple args can't be wrapped because of torch & paddle. I updated the test with new cases.
> 
> Thank you for reading through this. If I'm missing something or there is a better approach to solve any of these issues please tell me. @VedPatwardhan

Yep, sounds good!
1. awesome
2. I'd say for consistency with `execute_with_gradients` let's have `xs` as a single argument rather than a "spread" one (i.e. without `*`)
3. awesome, yeah that's a great approach
4. I think we might as well go with the wrapped approach in this function too. Just trying to think of the function signature as close as possible to `execute_with_gradients` to make it intuitive for users 😅
5. Oh okay, would be happy to learn more about such functions, as ideally we should try to implement a superset across the backends.

Thanks for diving into this @somaia02 🙂

somaia02:
I removed argnums in jax and all the *'s. But now we're back to having the problem with torch and paddle. This is an example of the problem:
```python
import torch
x1=torch.tensor([1.])
x2=torch.tensor([1.])
fn = lambda xs: xs[0]+xs[1]

# Try list
xs=[x1,x2]
torch.autograd.functional.jacobian(fn, xs)
# The inputs given to jacobian must be either a Tensor or a tuple of Tensors

# Try tuple
xs=(x1,x2)
torch.autograd.functional.jacobian(fn, xs)
# <lambda>() takes 1 positional argument but 2 were given
```
The jacobian function calls `fn` internally. If `xs` is not a tensor, it tries to unpack it. The same happens with paddle. The only solution I could think of is to call jacobian multiple times, one for each argument. Something like this:

```python
jacobian = []
for arg in args:
    other_args=exclude_arg(arg, args)
    jac_fn = lambda x: fn(x, other_args)
    jacobian += [torch.jacobian(jac_fn, arg)]
```
But this looks inefficient I'd rather go with the earlier unpacking approach. Is there a better way to get around this? @VedPatwardhan

VedPatwardhan:
Actually the reason behind suggesting not using `*` is because it would make `xs` more general, because now it could be a `list` but might as well be a `dict` or `Container`. It would be a bit inefficient, but probably we can tradeoff some of the efficiency to be able to support all the structures that `jax` natively supports in `jacfwd` and `jacrev` for implementing a superset. Thanks for looking into this @somaia02 🙂

somaia02:
I fixed the issue with torch and paddle. Can you please review? @VedPatwardhan

VedPatwardhan:
The changes look good to me! There's one last think left to do with this function, so @Ishticode recently worked on the `grad` function to allow 2 `grad` functions to be chained together to compute a higher order gradient, similar to what `jax.grad` natively supports. `jax.jacrev` and `jax.jacfwd` also has a similar feature and it would be good to have it in `ivy.jac`. I'll request a review from @Ishticode to suggest further steps to get chaining working for this function.

Thanks a ton for the effort you've made uptil now @somaia02 💪, we can get the PR merged once chaining is working 🙂

Ishticode:
I think best to leave chaining to a separate PR to keep things simpler. If you agree @VedPatwardhan then please merge the PR and @somaia02 can contact me for details and work on chaining.

VedPatwardhan:
Awesome, sounds good! Let's get the PR merged in that case @nassimberrada. Thanks @somaia02 @Ishticode @nassimberrada 🙂

PR #16068, opened 2023-05-30
added partition and its testing to numpy frontend.

### LINKED ISSUE #15870: partition

antoniobisegna:
I started working on the IDE but as soon as I run (any) testing I get: 
```
ImportError while loading conftest '/ivy/ivy_tests/test_ivy/conftest.py'.
ivy_tests/test_ivy/__init__.py:1: in <module>
    from .. import config
E   ImportError: cannot import name 'config' from 'ivy_tests' (/ivy/ivy_tests/__init__.py)

```

closes #15870


antoniobisegna:
By the way, I don't know how but I closed the other pull request about this task. I hope there is no problem in opening this new one.

antoniobisegna:
Hi @V-G-spec, I am working on the testing and I was wondering which helpers.array_helpers I could use. I would need to generate the same number of arrays  `a` and integers (indexes) `kth` and an axis `axis`. I am using `array_indices_axis` but I get the error _ValueError: too many values to unpack (expected 4)_ as in picture below. This is probably because I get two ints at the end of the tuple (in the screen you can see it in the Falsifying example) and I would need only one. I don't know how to set it to just one random axis and not equal to the size of the arrays/indexes. Thanks
Screenshot 2023-06-01 alle 12 04 19

antoniobisegna:
Hi, @V-G-spec @ivy-leaves I have been working on the testing of the frontend function but I got stuck now because there is no backend attribute corresponding to the one I created in frontend. Could you review my work and also give me some directions? Thanks

antoniobisegna:
Hi @ivy-leaves  @V-G-spec, could you review my pull request? Thanks

P.s. 

- I still get an error when running the intelligent-test-1, it takes a lot of time and then it suddenly closes with `Error: The operation was canceled.`. I also get intelligent-test-15 failing, but that is connected to another testing function that I am not working on.
- When running the test on the terminal I get 
`` AssertionError:  the results from backend numpy and ground truth framework numpy do not match
           [0 0 1]!=[0 1 0] ``
which does not say much since both results are correct, given the definition of the numpy function.

antoniobisegna:
Hi @ivy-leaves @V-G-spec, could you review my pull request? Thanks

V-G-spec:
Hi @antoniobisegna, really sorry about this massive delay. I was quite busy with some internal deadlines. I checked the failing test, and it looks like it's not from your changes. Everything else looks good!

antoniobisegna:
@V-G-spec, thanks!

PR #16101, opened 2023-05-31
Add automatic device shifting for arrays and `soft_device` global variable to control the behaviour

bipinKrishnan:
The PR is related to the task defined here:
https://trello.com/c/aHJLa4jD

### DIFF
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -795,8 +795,7 @@ def handle_device_shifting_for_arrays(fn: Callable) -> Callable:
                 *args, **kwargs
             )
         else:
-            inputs = list(args)
-            inputs.extend(kwargs.values())
+            inputs = args + tuple(kwargs.values())
             devices = set(ivy.dev(x) for x in inputs if isinstance(x, ivy.Array))
             if len(devices) == 1:
                 array_device = next(iter(devices))
@@ -806,9 +805,9 @@ def handle_device_shifting_for_arrays(fn: Callable) -> Callable:
                     )
             elif len(devices) > 1:
                 raise ivy.utils.exceptions.IvyException(
-                    "Expected all input arrays to be on the same device ",
-                    "but found atleast two devices - {}".format(devices),
-                    "Set `ivy.set_soft_device_mode(True)` to handle this problem.",
+                    "Expected all input arrays to be on the same device, "
+                    f"but found atleast two devices - {devices}, "
+                    "set `ivy.set_soft_device_mode(True)` to handle this problem."
                 )
         return fn(*args, **kwargs)
 
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -790,9 +790,26 @@ def handle_device_shifting_for_arrays(fn: Callable) -> Callable:
         -------
             The return of the function.
         """
-        args, kwargs = ivy.current_backend().handle_soft_device_variable(
-            *args, **kwargs
-        )
+        if ivy.get_soft_device_mode():
+            args, kwargs = ivy.current_backend().handle_soft_device_variable(
+                *args, **kwargs
+            )
+        else:
+            inputs = list(args)
+            inputs.extend(kwargs.values())
+            devices = set(ivy.dev(x) for x in inputs if isinstance(x, ivy.Array))
+            if len(devices) == 1:
+                array_device = next(iter(devices))
+                with ivy.DefaultDevice(array_device):
+                    args, kwargs = ivy.current_backend().handle_soft_device_variable(
+                        *args, **kwargs
+                    )
+            elif len(devices) > 1:
+                raise ivy.utils.exceptions.IvyException(
+                    "Expected all input arrays to be on the same device ",
+                    "but found atleast two devices - {}".format(devices),
+                    "Set `ivy.set_soft_device_mode(True)` to handle this problem.",
+                )
         return fn(*args, **kwargs)
 
     _handle_device_shifting_for_arrays.handle_device_shifting_for_arrays = True
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -31,7 +31,7 @@ FN_DECORATORS = [
     "with_unsupported_dtypes",
     "handle_nans",
     "handle_mixed_function",
-    "auto_shift_device",
+    "handle_device_shifting_for_arrays",
 ]
 
 
@@ -772,20 +772,34 @@ def infer_device(fn: Callable) -> Callable:
     return _infer_device
 
 
-def auto_shift_device(fn: Callable) -> Callable:
-    def _auto_shift_device(*args, **kwargs):
-        if ivy.functional.device.soft_device_placement and ivy.get_array_mode():
+def handle_device_shifting_for_arrays(fn: Callable) -> Callable:
+    @functools.wraps(fn)
+    def _handle_device_shifting_for_arrays(*args, **kwargs):
+        """
+        Move all array inputs of the function to `ivy.default_device()`.
+
+        Parameters
+        ----------
+        args
+            The arguments to be passed to the function.
+
+        kwargs
+            The keyword arguments to be passed to the function.
+
+        Returns
+        -------
+            The return of the function.
+        """
+        if ivy.get_soft_device_mode() and ivy.get_array_mode():
             device = ivy.default_device()
-            args = (
-                arg.to_device(device) if type(arg) == ivy.Array else arg for arg in args
+            args, kwargs = ivy.nested_map(
+                [args, kwargs],
+                lambda x: (x.to_device(device) if isinstance(x, ivy.Array) else x),
             )
-            kwargs = {
-                k: v.to_device(device) if type(v) == ivy.Array else v
-                for k, v in kwargs.items()
-            }
         return fn(*args, **kwargs)
 
-    return _auto_shift_device
+    _handle_device_shifting_for_arrays.handle_device_shifting_for_arrays = True
+    return _handle_device_shifting_for_arrays
 
 
 # Inplace Update Handling #
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -31,6 +31,7 @@ FN_DECORATORS = [
     "with_unsupported_dtypes",
     "handle_nans",
     "handle_mixed_function",
+    "auto_shift_device",
 ]
 
 
@@ -769,6 +770,22 @@ def infer_device(fn: Callable) -> Callable:
 
     _infer_device.infer_device = True
     return _infer_device
+
+
+def auto_shift_device(fn: Callable) -> Callable:
+    def _auto_shift_device(*args, **kwargs):
+        if ivy.functional.device.soft_device_placement and ivy.get_array_mode():
+            device = ivy.default_device()
+            args = (
+                arg.to_device(device) if type(arg) == ivy.Array else arg for arg in args
+            )
+            kwargs = {
+                k: v.to_device(device) if type(v) == ivy.Array else v
+                for k, v in kwargs.items()
+            }
+        return fn(*args, **kwargs)
+
+    return _auto_shift_device
 
 
 # Inplace Update Handling #
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -790,12 +790,9 @@ def handle_device_shifting_for_arrays(fn: Callable) -> Callable:
         -------
             The return of the function.
         """
-        if ivy.get_soft_device_mode() and ivy.get_array_mode():
-            device = ivy.default_device()
-            args, kwargs = ivy.nested_map(
-                [args, kwargs],
-                lambda x: (x.to_device(device) if isinstance(x, ivy.Array) else x),
-            )
+        args, kwargs = ivy.current_backend().handle_soft_device_variable(
+            *args, **kwargs
+        )
         return fn(*args, **kwargs)
 
     _handle_device_shifting_for_arrays.handle_device_shifting_for_arrays = True
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -13,6 +13,7 @@ import inspect
 # for wrapping (sequence matters)
 FN_DECORATORS = [
     "infer_device",
+    "handle_device_shifting_for_arrays",
     "infer_dtype",
     "handle_array_function",
     "integer_arrays_to_float",
@@ -31,7 +32,6 @@ FN_DECORATORS = [
     "with_unsupported_dtypes",
     "handle_nans",
     "handle_mixed_function",
-    "handle_device_shifting_for_arrays",
 ]
 
 
--- a/ivy/functional/backends/jax/device.py
+++ b/ivy/functional/backends/jax/device.py
@@ -124,12 +124,12 @@ def tpu_is_available() -> bool:
 
 
 def handle_soft_device_variable(*args, **kwargs):
-    default_device = ivy.default_device()
+    default_device = as_native_dev(ivy.default_device())
     args, kwargs = ivy.nested_map(
         [args, kwargs],
         lambda x: (
-            jax.device_put(x, as_native_dev(default_device))
-            if (isinstance(x, JaxArray) and ivy.dev(x) != default_device)
+            jax.device_put(x, default_device)
+            if (isinstance(x, jax.Array) and x.device != default_device)
             else x
         ),
     )
--- a/ivy/functional/backends/jax/device.py
+++ b/ivy/functional/backends/jax/device.py
@@ -124,28 +124,15 @@ def tpu_is_available() -> bool:
 
 
 def handle_soft_device_variable(*args, **kwargs):
-    if ivy.get_soft_device_mode():
-        default_device = ivy.default_device()
-        args, kwargs = ivy.nested_map(
-            [args, kwargs],
-            lambda x: (
-                jax.device_put(x, as_native_dev(default_device))
-                if (isinstance(x, JaxArray) and ivy.dev(x) != default_device)
-                else x
-            ),
-        )
-    else:
-        inputs = list(args)
-        inputs.extend(kwargs.values())
-        devices = set(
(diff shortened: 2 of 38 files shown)


ivy/func_wrapper.py
```diff
@@ -31,6 +31,7 @@
     "with_unsupported_dtypes",
     "handle_nans",
     "handle_mixed_function",
+    "auto_shift_device",
```
VedPatwardhan, line 34: let's rename this to `handle_device_shifting`, just for consistency

ivy/func_wrapper.py
```diff
@@ -771,6 +772,22 @@ def _infer_device(*args, device=None, **kwargs):
     return _infer_device
 
 
+def auto_shift_device(fn: Callable) -> Callable:
+    def _auto_shift_device(*args, **kwargs):
+        if ivy.functional.device.soft_device_placement and ivy.get_array_mode():
```
VedPatwardhan, line 777: we should be making use of something like `ivy.get_soft_device_mode` instead of `ivy.functional.device.soft_device_placement`

ivy/functional/backends/torch/layers.py
```diff
@@ -372,6 +372,7 @@ def conv3d_transpose(
     return res
 
 
+@auto_shift_device
```
VedPatwardhan, line 375: I think the decorator should be applied to the `ivy` function rather than the backend impementations, as the behaviour should be consistent across all backends

ivy/functional/ivy/device.py
```diff
@@ -42,7 +42,7 @@
 dev_handles = dict()
 split_factors = dict()
 max_chunk_sizes = dict()
-
+soft_device_placement = True
```
VedPatwardhan, line 45: Let's have this as a stack `soft_device_mode_stack = list()` with getters and setters similar to the other stacks (e.g. `array_mode`, `shape_array_mode`, etc.)

VedPatwardhan (changes requested):
Hey @bipinKrishnan, thanks for making the changes. I've requested a few more, let's see what @sherry30 thinks 🙂

ivy/func_wrapper.py
```diff
@@ -771,6 +772,22 @@ def _infer_device(*args, device=None, **kwargs):
...
+            args = (
+                arg.to_device(device) if type(arg) == ivy.Array else arg for arg in args
+            )
+            kwargs = {
+                k: v.to_device(device) if type(v) == ivy.Array else v
+                for k, v in kwargs.items()
+            }
+        return fn(*args, **kwargs)
+
+    return _auto_shift_device
```
sherry30, line 788: should add `_auto_shift_device.auto_shift_device = True` above this

ivy/func_wrapper.py
```diff
@@ -771,6 +772,22 @@ def _infer_device(*args, device=None, **kwargs):
     return _infer_device
 
 
+def auto_shift_device(fn: Callable) -> Callable:
+    def _auto_shift_device(*args, **kwargs):
+        if ivy.functional.device.soft_device_placement and ivy.get_array_mode():
+            device = ivy.default_device()
+            args = (
```
sherry30, line 779: I think it would be best to do this with `ivy.nested_map` in case an arbitrary list of arrays is passed.

sherry30 (changes requested):
I've asked for a few changes.
the rest of the problems are mentioned by Ved. 

thanks

ivy/func_wrapper.py
sherry30: Maybe lets rename it to `handle_device_shifting_for_arrays` to make it clear that it only changes device for arrays and not for the op being executed.

ivy/functional/backends/torch/layers.py
bipinKrishnan: Hi @VedPatwardhan @sherry30, I am using the following example to test the functionality(hope this is the expected scenario):
```python
import ivy

ivy.set_backend("torch")
ivy.set_soft_device_placement(True)

x = ivy.random_uniform(shape=(1, 1, 1, 2), device="gpu:0")
f = ivy.random_uniform(shape=(2, 1, 2, 2), device="cpu:0")

c = ivy.linear(x, f)
```
On calling `ivy.linear`, the execution moves to torch backend, so I thought the decorator is to be added there also. Initially my approach was same as Ved mentioned, but it wasn't moving the arrays to gpu as expected, and it worked only when the decorator is present in the backend function as well. Let me know your thoughts on this.

bipinKrishnan: Adding `_auto_shift_device.auto_shift_device = True` to the wrapper solved this issue :smile:

ivy/func_wrapper.py
bipinKrishnan: Thanks Sherry, this single line of code saved a lot of debugging hours for me😂

bipinKrishnan:
When the backend is tensorflow, the device shifting happens even if we set `ivy.set_soft_device_mode(False)`, might be because tensorflow already handles the device placement on its own. My thought is that we may have to explicity set the value of `tf.config.set_soft_device_placement` inside setter and unsetter functions when the backend is tensorflow. Any feedback on this would be great :smile:

sherry30:
Actually, after going through the docs again.
`tf.config.set_soft_device_placement` is not for allowing shifting of tensors between devices, but only op shifting between devices. For instance when an invalid device is given by the user, tensorflow will execute the op on the default device instead. There are some other conditions as well for it given in the docs.

The following code works fine:
```
import tensorflow as tf
tf.config.set_soft_device_placement(False)

with tf.device("cpu"):
  x = tf.random.uniform((1, 1, 1, 2))

y = tf.math.abs([[x, x]])
print(y)
print(y.device)
```
I think automatic tensor shifting is always on in tensorflow and no way to turn it off. So I guess we can just skip it in the case of tensorlfow to save some overhead.

VedPatwardhan (commented):
That's interesting. I think we can go with what @sherry30 is suggesting if there isn't another way to force the operation on a `cpu`. Let's maybe sync up next week and discuss this? Thanks @bipinKrishnan and @sherry30 🙂

sherry30:
> In that case, do you think we need to show some kind of a warning or raise an error when the user explicitly sets the soft device mode to False with either tf or numpy backend?

So in the case of tensorflow, we should raise an error manually if the `soft_device_placement` global variable is `False` since tensorflow doesn't handle that.
and in case of numpy, we should raise an error if the `soft_device_placement` variable is set to `True` and the default device is anything other than cpu.
I think the cleanest way to do this is to create a mixed function in device.py file called something like `hanlde_soft_device_variable`. tensorflow and numpy can have their own implementations (in which they raise error where needed and skip the conversion for tensorflow etc.) in the backend and jax and torch can use the ivy implementation where you can shift arrays between devices. and then we can call this function in the decorator.
Let me know what you think of this approach @VedPatwardhan.

bipinKrishnan:
> > In that case, do you think we need to show some kind of a warning or raise an error when the user explicitly sets the soft device mode to False with either tf or numpy backend?
> 
> So in the case of tensorflow, we should raise an error manually if the `soft_device_placement` global variable is `False` since tensorflow doesn't handle that. and in case of numpy, we should raise an error if the `soft_device_placement` variable is set to `True` and the default device is anything other than cpu.
> 
> I think the cleanest way to do this is to create a mixed function in device.py file called something like `hanlde_soft_device_variable`. tensorflow and numpy can have their own implementations in the backend (in which they raise error where needed and skip the conversion for tensorflow etc.) and jax and torch can use the ivy implementation where you can shift arrays between devices. and then we can call this function in the decorator. Let me know what you think of this approach @VedPatwardhan.

I was about to implement the functionality but I have some second thoughts in my mind, will this degrade user experience in any way? There may be cases when the user may have to use numpy in multiple places along with torch code, so the user has to set and unset soft variable each time a numpy code comes in(if the user requires the soft device feature). Is it okay we show some warning like "falling back to cpu" instead of raising an error? @sherry30 @VedPatwardhan

VedPatwardhan (changes requested):
Hey @bipinKrishnan, just suggested a small change. Also, is this decorator specific to layers, if not we might want to add it for the other functions which accept arrays. Also, let's see what @sherry30 thinks. Thanks 🙂

bipinKrishnan:
> Hey @bipinKrishnan, just suggested a small change. Also, is this decorator specific to layers, if not we might want to add it for the other functions which accept arrays. Also, let's see what @sherry30 thinks. Thanks slightly_smiling_face

Nop. I just added it to layers just to run the test example I have :smile:

VedPatwardhan (changes requested):
Just one last change, thanks @bipinKrishnan 🙂

sherry30 (changes requested):
requested 1 change for tensorflow mixed funciton.

We'll also have to add a mixed function in the case of jax. Jax has a concept of committing arrays on devices. You can read more about it [here](https://jax.readthedocs.io/en/latest/faq.html#faq-data-placement). But basically, Jax initializes arrays on the default device and they're considered uncommitted but in case we explicitly mention on which device we want the array on with `jax.device_put()` it becomes committed. The op is executed on the device on which the committed array resides. If there are 2 committed arrays on different devices, then it raises an error. But it is possible for 2 arrays to be on different devices and be uncommitted, in which case the op is executed on the default device.

So because of this weird nature of Jax, we'll have to add a mix function for this as well. we will have to shift all the arrays on the default device if `ivy.get_soft_device_mode()` is `True` like we do normally in the ivy implementation of the mixed function, otherwise we will have to check if the arrays are on different devices from teh default device and raise an error if they are (similar to tensorflow implementation of the mixed function).

We will also have to add a mixed function for paddle paddle but that will be the same as jax one. paddle seems to execute the op on gpu if any of the given arrays were on gpu regardless of what the default device was set. otherwise it will execute it on cpu if all the provided arrays were on cpu even if the default device was set to gpu (This is from what I've tested, it doesn't explicitly mention anything in the docs). So we can deal with this exactly the same as Jax.

Let me know if you have any question or any suggestions.
Thanks

ivy/func_wrapper.py
```diff
@@ -31,6 +31,7 @@
     "with_unsupported_dtypes",
     "handle_nans",
     "handle_mixed_function",
+    "handle_device_shifting_for_arrays",
```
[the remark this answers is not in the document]
bipinKrishnan: So, the above list decides the order of executing the decorator for each function? Does that mean I only need to update the decorator position in the list and keep the rest of the code unchanged?

ivy/functional/backends/tensorflow/device.py
```diff
@@ -99,6 +99,15 @@ def tpu_is_available() -> bool:
         return False
 
 
+def handle_soft_device_variable(*args, **kwargs):
+    if not ivy.get_soft_device_mode():
+        raise ivy.utils.exceptions.IvyBackendException(
```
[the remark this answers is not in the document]
bipinKrishnan: Does a simple matching of default device to tf device will work or should I add extra checks for multi GPU cases like default='gpu:0' and tf_tensor= 'gpu:1'?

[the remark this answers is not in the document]
sherry30: you can just do a simple comparison of default_device with `ivy.as_ivy_dev(x.device)`, that should be enough.

ivy/func_wrapper.py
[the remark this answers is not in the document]
VedPatwardhan: So this list decides the sequence by which the function would be wrapped when a backend is set. When using ivy without setting a backend, it uses the sequence specified above each function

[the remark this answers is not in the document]
bipinKrishnan: Since the decorator location is to be moved for all ivy functions, I will just confirm what I understood - `handle_device_shifting_for_arrays` should be below `infer_device` in the list and the decorators placed on ivy functions should be in the same order as in the `FN_DECORATORS` list, right?

VedPatwardhan:
Also, there's a couple more changes I'd like to request on this PR before we merge,
1. The `default_device` function seems to follow the [`tensorflow`](https://github.com/unifyai/ivy/blob/40c3f381043d1c470fb4f04a0a5fd380a8a95130/ivy/functional/ivy/device.py#L760) approach by default when `ivy.default_device_stack` is empty, we should also condition this based on whether the `soft_device` mode was set.
2. Let's update the decorator so that if the `soft_device` mode wasn't set, then we'd just throw an error before entering the backend implementations if the arrays being used in the same operation have different devices. I know this would require quite a few changes, and we might want to add these in a different PR (which is also fine), adding such a check would also ensure that when `soft_device` is disabled, we follow the `torch` behaviour. We could probably also rename the decorator to describe this functionality as well.

Sorry about not having this in the task initially as this is something that's come out of a discussion with @MahmoudAshraf97 today, but probably for the best to get the device handling done at once (albeit through different PRs if that's what you guys prefer), given that the release is around the corner. 

Thanks @bipinKrishnan @sherry30 and feel free to share your thoughts/suggestions/improvements 🙂

sherry30:
I'm not sure if I understood the 1st point completely. 
How should the `soft_device` mode affect `default_device`? Shouldn't the `default_device` be set to the best option available by default, which was suggested by Dan.


About the 2nd point, we're already doing that. if the arrays are on different devices when `soft_device` is disabled then we will throw an error (How torch deals with this.) . We will set it to `False` by default. The one difference between our approach and torch's approach is, torch executes the op based on what the input tensors are on, and we execute it based on the `default_device` and we expect all the arrays to be on that device.

ivy/functional/ivy/device.py
```diff
@@ -267,10 +280,78 @@ def print_all_ivy_arrays_on_dev(
...
+    >>> ivy.get_soft_device_mode()
+    True
+
+    >>> ivy.set_soft_device_mode(False)
+    >>> ivy.get_soft_device_mode()
+    False
+    """
+    global soft_device_mode_stack
+    if not soft_device_mode_stack:
+        return True
```
sherry30, line 346: should return `False` by default.

sherry30 (changes requested):
request to change default `soft_device` mode to be False

VedPatwardhan:
> I'm not sure if I understood the 1st point completely. How should the `soft_device` mode affect `default_device`? Shouldn't the `default_device` be set to the best option available by default, which was suggested by Dan.
> 
> About the 2nd point, we're already doing that. if the arrays are on different devices when `soft_device` is disabled then we will throw an error (How torch deals with this.) . We will set it to `False` by default. The one difference between our approach and torch's approach is, torch executes the op based on what the input tensors are on, and we execute it based on the `default_device` and we expect all the arrays to be on that device.

1. Yeah I wasn't completely sure about what the best approach was, but my rationale was that if we're automatically moving the arrays to a `gpu` when available with a `soft_device` mode, then it would make sense to return `gpu` from `ivy.default_device`. However, if we're not skewed to a `gpu` when `soft_device` is disabled, we might also not be skewed to it when we call `ivy.default_device`. But I think if Dan's suggested the current approach of setting it to the best option, then let's continue with that. Thanks for clarifying.
2. Ah makes sense, my second point was actually based on the premise that we'd return `cpu` as the default device when the `soft_device` mode is disabled. But then if the `default_device` is a `gpu` and we pass two `cpu` arrays to an `ivy` function, wouldn't that throw an error as well? Cuz if passing two arrays on the same device to a function is a valid operation, then we might want to allow that.

Thanks for clarifying on both the points @sherry30, it would be great if we could clarify this policy in the deep dive 🙂

VedPatwardhan:
> About the 2nd point, its true that it will throw an error if 2 arrays are on cpu and the `default_device` is `gpu` and `soft_device` is disabled. For this I thought the user can use the context manager like `with ivy.DefaultDevice("cpu"):` and execute any op there on the `cpu` with arrays initialized on `cpu`. But if you think that's too limiting then we can remove this dependency from `default_device` and then the device on which the op will be executed on will be inferred from the input arrays just like torch and not from `default_device`. Let me know which approach should we go with. @VedPatwardhan

I think it would be great if we could allow users to perform operations on the `cpu`, even if the `default_device` was a `gpu`. Because a `default_device` is essentially a fallback option when a `device` isn't passed explicitly, so if we're allowing users to create `cpu` arrays by passing `device='cpu'` even when the `default_device` is a `gpu`, we might as well also allow operations when input arrays are on a `cpu` by considering it as intended behaviour. Hope that makes sense @sherry30 🙂

sherry30 (changes requested):
Yup, makes sense.

In that case, the only thing we need to change are the mixed functions when `soft_device` is set to `False`. You would have to check if all the arrays are on the same device instead of comparing it to `default_device`. @bipinKrishnan

vedpatwardhan (approved):
The changes look good to me, however given that the CI hasn't run any of the intelligent tests, it would be a bit risky to merge this PR, given that it's such a major change. Hey @RashulChutani is there any issues in the workflow triggering pipeline? Thanks 🙂

RashulChutani:
Hi @Ved, Let me check this out.

RashulChutani:
Somehow the Workflows are not triggered on this PR, while they are still triggering for others. @bipinKrishnan, Would you mind rebasing on top of master? You may need to create a new PR. Thanks.

sherry30 (changes requested):
Lets change the name of the decorator to just `handle_device_shifting` since originally we were only changing the device of arrays and the op would've been executed on the `default_device` but now the device on which op is executed on can also be changed by this decorator (in case `soft_device` is disabled, we can execute the op on the device all the arrays are on).

ivy/functional/backends/jax/device.py
```diff
@@ -123,6 +123,19 @@ def tpu_is_available() -> bool:
     return _dev_is_available("tpu")
 
 
+def handle_soft_device_variable(*args, **kwargs):
+    default_device = as_native_dev(ivy.default_device())
```
sherry30, line 127: You can pass `as_native=True` to `ivy.default_device` which does the same thing as the above line but just looks a bit cleaner I think.

bipinKrishnan:
In that case I will add the changes @sherry30 mentioned and create a new PR. Thanks :slightly_smiling_face:

sherry30 (changes requested):
a small change requested

PR #16134, opened 2023-06-01
adding max pool 3d in paddle layers

abdulasiraj:


ivy/functional/backends/paddle/experimental/layers.py
```diff
@@ -69,7 +69,39 @@ def max_pool3d(
...
+        [
+            pad_w // 2,
+            pad_w - pad_w // 2,
+            pad_h // 2,
+            pad_h - pad_h // 2,
+            pad_d // 2,
+            pad_d - pad_d // 2,
+        ],
+        value=float("-inf"),
+        data_format="NDHWC",
```
MahmoudAshraf97, line 97: why is the data format constant here? I know it's handled in L80:81 but I don't get the reason

ivy/functional/backends/paddle/experimental/layers.py
```diff
@@ -103,20 +135,49 @@ def avg_pool2d(
...
+        ],
+        value=float("-inf"),
+        data_format="NCDHW",
+    )
+    if padding != "VALID" and padding != "SAME":
+        raise ValueError(f'Invalid padding arg {padding}\nMust be one of: "VALID" or "SAME"')
+    res = paddle.nn.functional.max_pool3d(x, kernel_size=kernel, stride=strides, padding=0)
+    if data_format == "NDHWC":
+        res = paddle.transpose(res, perm=[0, 2, 3, 4, 1])
+    return res
```
MahmoudAshraf97, line 180: I think you messed with avgpool3d please check and revert if necessary

MahmoudAshraf97 (changes requested):

ivy/functional/backends/paddle/experimental/layers.py
abdulasiraj: Because paddle's pad has defualt data_format=`NCHW`, which doesn't support 5 dims Tesnor. Also,In other backends `NDHWC` is defualt or may be they handle themsolves but they don't have data_format argument in their pad api.

PR #16156, opened 2023-06-01
Move handle_mixed_function decorator after handle_nestable 

hello-fri-end:


hello-fri-end:
Hey @VedPatwardhan! I moved the `handle_mixed_function` decorator after `handle_nestable` and all the tests are passing as expected. I believe that moving it up any further would create unnecessary overhead when the mixed function defers to the compositional implementation as all the decorators would be run again starting from the  `handle_exceptions` decorator. Happy to hear your thoughts about this :smile:

ivy/func_wrapper.py
```diff
@@ -1000,18 +1000,23 @@ def _wrap_function(
             if (hasattr(original, attr) and not hasattr(to_wrap, attr)) or (
                 mixed_fn
                 and (
-                    attr == "inputs_to_native_arrays" or attr == "outputs_to_ivy_arrays"
+                    attr == "inputs_to_native_arrays"
```
VedPatwardhan, line 1003: Can we make this function a bit cleaner? Seems like we're making a lot of checks and there's scope to improve readability. Thanks!

VedPatwardhan (changes requested):
Hey @hello-fri-end, just requested a change, I don't think it would add a lot of inefficiencies moving the decorator upward, given that compositional functions don't need to have `handle_nestable`, `to_native_arrays_and_back` and others, but I think it would impact correctness if we move it upward so we shouldn't. 
Thanks for looking into this 🙂

hello-fri-end:
> Hey @hello-fri-end, just requested a change, I don't think it would add a lot of inefficiencies moving the decorator upward, given that compositional functions don't need to have `handle_nestable`, `to_native_arrays_and_back` and others, but I think it would impact correctness if we move it upward so we shouldn't. Thanks for looking into this slightly_smiling_face

Hey @VedPatwardhan! I did not realize that the compositional function doesn't need the `handle_nestable` decorator. Would it create a problem if it's present? If yes, how would we remove it inside the `_wrap_function`?

ivy/func_wrapper.py
hello-fri-end: Updated it. Please let me know if there's anything else that should be changed :+1: :)

hello-fri-end:
ivy-gardner

hello-fri-end:
ivy-gardener

hello-fri-end:
@VedPatwardhan Merging this one. Will create a new PR for adding the attribute as discussed on SoWork.

PR #15376, opened 2023-05-12
foldl

louazz:
Added tensorflow foldl function to the frontend #14972

PR #16186, opened 2023-06-02
Adding `tf.foldl` to ivy frontends along with tests

### LINKED ISSUE #16179: foldl

FourFortyFour:
close #16179


FourFortyFour:
I proceeded with the implementation via `functools.reduce` and not `ivy.while_loop` mainly due to the comment left on #15376. So I assume the implementation wanted is via `reduce`. The test fails for paddlepaddle for reasons that are currently unclear to me returning: 
```python
FAILED ivy/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py::test_tensorflow_foldl[cpu-ivy.functional.backends.paddle-False-False] - RuntimeError: (NotFound) The kernel with key (CPU, NCHW, int16) of kernel `slice` is not registered.
``` 
while the tesnorflow tests fail with the following error summary: 
```python
FAILED ivy/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py::test_tensorflow_foldl[cpu-ivy.functional.backends.tensorflow-False-False] - hypothesis.errors.Flaky: Hypothesis test_tensorflow_foldl(fn=lambda a, b: a + b, elems_and_shape=([-281440616972288...
```

ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -80,6 +81,23 @@ def fill(dims, value, name=None):
     return ivy.full(dims, value)
 
 
+@to_ivy_arrays_and_back
+def foldl(fn, elems, initializer=None):
```
xoiga123, line 85: Please implement the exact same signature of [foldl](https://www.tensorflow.org/api_docs/python/tf/compat/v1/foldl)

ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -80,6 +81,23 @@ def fill(dims, value, name=None):
...
+    try:
+        iter(elems)
+    except TypeError:
+        raise TypeError("elems must be an iterable object")
```
xoiga123, line 89: Nit: maybe use `ivy.utils.exceptions.IvyException` :thinking:?

ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -80,6 +81,23 @@ def fill(dims, value, name=None):
...
+
+    arr = ivy.array(elems)
```
xoiga123, line 91: We don't need this as `@to_ivy_arrays_and_back` already did the conversion and `elems` should already be an `ivy.Array`

ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -80,6 +81,23 @@ def fill(dims, value, name=None):
...
+
+    if initializer is None and len(arr) > 0:
```
xoiga123, line 93: Can we use something other than `len(arr)`? I tested with an `ivy.Array` containing a `tf.Variable` and this failed.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py
```diff
@@ -265,6 +265,58 @@ def test_tensorflow_eye(
     )
 
 
+@st.composite
+def elems_and_shape(draw):
```
xoiga123, line 269: Is this helper necessary? Can we get a 1-D array of floats by utilizing already existing helpers?

xoiga123 (changes requested):
Thank you for contributing to Ivy :hugs: Looks great, however I have some feedback and questions.

> This method also allows multi-arity elems and output of fn. If elems is a (possibly nested) list or tuple of tensors, then each of these tensors must have a matching first (unpack) dimension. The signature of fn may match the structure of elems.

Is this accounted for in the implementation? If it is, please modify the test to also account for this.

Thank you!

xoiga123:
> I proceeded with the implementation via `functools.reduce` and not `ivy.while_loop` mainly due to the comment left on #15376. So I assume the implementation wanted is via `reduce`. The test fails for paddlepaddle for reasons that are currently unclear to me returning:
> 
> ```python
> FAILED ivy/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py::test_tensorflow_foldl[cpu-ivy.functional.backends.paddle-False-False] - RuntimeError: (NotFound) The kernel with key (CPU, NCHW, int16) of kernel `slice` is not registered.
> ```

Regarding this, I've identified the bug in the paddle backend and contacted the guys in charge of it. Consider your tests all passed :+1:

ivy/functional/frontends/tensorflow/general_functions.py
FourFortyFour: That's what I went with originally but saw that no where else in the file was an ivy.exception used or thrown, will change it as requested.

FourFortyFour:
Hello again, hope you're doing well! I've resolved most of the logistical changes you suggested (such as the function signature, use of `len(arr)`, exceptions, etc.). Regarding the use of the test function helper `elems_reshaped()`, I am not entirely sure how I can achieve the same behavior (mainly ensuring enough elems are generated with accordance to the shape) inside the test function itself or within the `handle_frontend_test` decorator. It is probably doable but I am not sure how to do it as of now, I will look towards trying to make that change and get back to you. As I see it currently I would need to replace the `draw()` functionality with `st.sampled()`, so to answer the question it does not seem to be necessary upon further inspection. 

Now, about the behavior of the current `foldl` implementation; it follows the behavior the foldl functionality based off of a rudimentary exploration in some notebooks, now that I sort of have a base for the overall implementation I can explore the details and potential edge cases further. One major difference is that the tensorflow implementation unpacks the input array on axis 0, which does not occur in the current implementation. The test function generates cases for inputs ranging from dimensions 1..10 on 4 different axes, specifically this line `shape = draw(st.lists(st.integers(min_value=1, max_value=10),min_size=1, max_size=4))`. So I believe it is accounted for in the tests specifically. A difference I observe currently is following: 
```python
In [156]: x = np.random.randn(3,3)

In [157]: f.reduce(lambda a,b: a+b, x)
Out[157]: array([ 1.37230861, -1.53313156,  0.12723355])

In [158]: tf.foldl(lambda a,b: a +b,x)
Out[158]: <tf.Tensor: shape=(3,), dtype=float64, numpy=array([ 1.37230861, -1.53313156,  0.12723355])>

In [159]: f.reduce(lambda a,b: a+b, [x])                                            
Out[159]: array([[ 0.50695112,  0.64469077,  0.75755285], 
          [ 1.49600689, -2.23531653, -0.13876601],
          [-0.6306494 ,  0.0574942 , -0.49155329]])                                    
In [160]: tf.foldl(lambda a,b: a +b,[x])
------------------------ ValueError Traceback (most recent call last) 
...
ValueError: The two structures don't have the same nested structure. 
First structure: type=list str=[TensorSpec(shape=(), dtype=tf.int32, name=None), [TensorSpec(shape=(), dtype=tf.int32, name=None), [TensorSpec(shape=(3,), dtype=tf.float64, name=None)]]]

Second structure: type=list str=[1, [<tf.Tensor: shape=(), dtype=int32, numpy=2>, [<tf.Tensor: shape=(3,), dtype=float64, numpy=array([0.50695112, 0.64469077, 0.75755285])>, <tf.Tensor: shape=(3,), dtype=float64, numpy=array([ 1.49600689, -2.23531653, -0.13876601])>]]]

More specifically: The two structures don't have the same number of elements. First structure: type=list str=[TensorSpec(shape=(3,), dtype=tf.float64, name=None)]. Second structure: type=list str=[<tf.Tensor: shape=(3,), dtype=float64, numpy=array([0.50695112, 0.64469077, 0.75755285])>, <tf.Tensor: shape=(3,), dtype=float64, numpy=array([ 1.49600689, -2.23531653, -0.13876601])>]
Entire first structure:
[., [., [.]]]
Entire second structure:
[., [., [., .]]]
```
since the `functools.reduce` implementation does not unstack the input array it just returns the input since its technically applying the binary add function to a single element list (which contains a (3,3) array). I believe there are more differences between this implementation and `tf.foldl` since we are also not creating an `ivy.Array` for each possibly nested array in the input, I am still exploring the full implications of this (aka figuring out what is going on xD). Thank you for your suggestions and help! Any advice on how to move forward or what else you may need is greatly appreciated and valued.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py
FourFortyFour: I resolved one of your comments, specifically the one about having the same signature of `tf.foldl()`. I am unsure if that was ok or not and could not revert it. :)

FourFortyFour:
Hello once more, hope your days are well! Are there any specific modifications you suggest I get done to get the PR into a potentially mergeable state?

ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -83,6 +84,35 @@ def fill(dims, value, name=None):
...
+    initializer=None,
+    parallel_iterations=10,
+    back_prop=True,
+    swap_memory=False,
+    name=None,
+):
+    try:
+        ivy.utils.assertions.check_isinstance(elems, ivy.Array)
+    except ivy.utils.exceptions.IvyException:
+        ivy.utils.exceptions.IvyException("elems must be an iterable object")
```
xoiga123, line 100: Maybe changing this into 1 line:
```python
ivy.utils.assertions.check_isinstance(elems, ivy.Array, "elems must be an iterable object")
```

ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -83,6 +84,35 @@ def fill(dims, value, name=None):
...
+    if not callable(fn):
+        ivy.utils.exceptions.IvyException(f"{fn.__name__} must be a callable function")
```
xoiga123, line 102: And this into 1 line:
```python
ivy.utils.assertions.check_true(callable(fn), f"{fn.__name__} must be a callable function")
```

ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -83,6 +84,35 @@ def fill(dims, value, name=None):
...
+    if initializer is None and elems.shape[0] > 0:
+        result = functools.reduce(fn, elems)
+        if isinstance(result, (int, float)):
+            return ivy.array([result])
+        else:
+            return result
+    elif initializer is not None:
+        return functools.reduce(fn, elems, initializer)
+    else:
+        return elems
```
xoiga123, line 113: Maybe we can use `ivy.reduce` for the whole implementation after checking for valid input shape? It already used `functools.reduce` inside and got better handling of scalars, instead of trying to reinvent the wheel here :pray:

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py
```diff
@@ -267,6 +267,58 @@ def test_tensorflow_eye(
...
+    )
+    elems_size = np.prod(shape).item()
+    elems = draw(
+        st.lists(
+            st.floats(min_value=-1000, max_value=1000),
+            min_size=elems_size,
+            max_size=elems_size,
+        )
+    )
+    return np.reshape(elems, shape)
```
xoiga123, line 283: I'm confused as to what this helper even does. From the lines, you are trying to:
- get a shape, i.e. `(1,2,3)`
- get the total size, i.e. `6`
- create a fixed list of floats of size equal to `total size`, i.e. `(1.0,2.0,3.0,4.0,5.0,6.0)`
- reshape it back to original shape, i.e. `[[[1., 2., 3.], [4., 5., 6.]]]`

All of this is entirely possible with `helpers.dtype_and_values` like how it's being used in other tests. Try these arguments
```python
max_value=1000,
max_dim_size=10,
max_num_dims=4,
```

By doing it this way, you can get generate `dtype` as well, and get rid of 
```python
dtype=helpers.get_dtypes("float", full=False),
```

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py
```diff
@@ -267,6 +267,58 @@ def test_tensorflow_eye(
...
+            lambda a, b: a + b,
+            lambda a, b: a - b,
+            lambda a, b: a * b,
+        ],
+    ),
+    initializer=st.one_of(st.none(), st.floats(min_value=-1000, max_value=1000)),
+    dtype=helpers.get_dtypes("float", full=False),
+    elems_reshaped=elems_reshaped(),
+)
+def test_tensorflow_foldl(
```
xoiga123, line 300: Try to test for other kwargs aside from `name` as well

xoiga123 (changes requested):
Hey @FourFortyFour! Sorry for the late reply. Hope you're doing well too :hugs:

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py
FourFortyFour: I'll check my understanding of the various helpers available and follow through with your suggestion ASAP.

FourFortyFour: Yeah you're understanding of what I'm trying to do is correct, but I think I failed to consider `helpers.dtype_and_values` and what it can do. Like you stated previously, the helper is unnecessary.

FourFortyFour: I resolved the issue with how I wrote the test function according to your suggestions/guidance :)

ivy/functional/frontends/tensorflow/general_functions.py
FourFortyFour: Shifting the implementation to use `ivy.reduce()` is a bit convoluted. First off, the initializer argument is not optional. Regarding the axes arg I believe it should just be kept at the default. Given the current implementation as you see it, replacing all the calls to `functools.reduce` with `ivy.reduce`, the test case for numpy passes, however the other test cases fail with a rather long stack trace. I'll share parts of the TensorFlow test stack trace below: 
```python
args = (ivy.array([]), ivy.array(-1), <function <lambda> at 0x7fa68c099d30>), kwargs = {} #these are the inputs to the test case
operand = ivy.array([]), init_value = ivy.array(-1), computation = <function <lambda> at 0x7fa68c099d30>

    @handle_exceptions
    @handle_nestable
    @inputs_to_ivy_arrays
    def reduce(
        operand: Union[ivy.Array, ivy.NativeArray],
        init_value: Union[int, float],
        computation: Callable,
        /,
        *,
        axes: Union[int, Sequence[int]] = 0,
        keepdims: bool = False,
    ) -> ivy.Array:
        """
        Reduces the input array's dimensions by applying a function along one or more axes.

        Parameters
        ----------
        operand
            The array to act on.
        init_value
            The value with which to start the reduction.
        computation
            The reduction function.
        axes
            The dimensions along which the reduction is performed.
        keepdims
            If this is set to True, the axes which are reduced are left in the result as
            dimensions with size one.

        Returns
        -------
        ret
            The reduced array.

        Examples
        --------
        >>> x = ivy.array([[1, 2, 3], [4, 5, 6]])
        >>> ivy.reduce(x, 0, ivy.add, 0)
        ivy.array([6, 15])

        >>> x = ivy.array([[1, 2, 3], [4, 5, 6]])
        >>> ivy.reduce(x, 0, ivy.add, 1)
        ivy.array([5, 7, 9])
        """
        axes = (axes,) if isinstance(axes, int) else axes
        axes = [a + operand.ndim if a < 0 else a for a in axes]
        axes = sorted(axes, reverse=True)
        init_value = ivy.array(init_value)
        op_dtype = operand.dtype
>       computation = _correct_ivy_callable(computation)

ivy/ivy/functional/ivy/experimental/general.py:72:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

func = <function <lambda> at 0x7fa68c099d30>

    def _correct_ivy_callable(func):
        # get the current backend of the given ivy callable
        if ivy.nested_any(
            func,
            lambda x: hasattr(x, "__module__")
            and x.__module__.startswith("ivy")
            and not x.__module__.startswith("ivy.functional.frontends"),
        ):
>           return ivy.__dict__[func.__name__]
E           KeyError: '<lambda>'

ivy/ivy/functional/ivy/experimental/general.py:19: KeyError

During handling of the above exception, another exception occurred:

    @handle_frontend_test(
>       fn_tree="tensorflow.foldl",
        fn=st.sampled_from(
            [
                lambda a, b: a + b,
                lambda a, b: a - b,
                lambda a, b: a * b,
            ],
        ),
        initializer=st.one_of(st.none(), st.floats(min_value=-1000, max_value=1000)),
        dtype_and_values=helpers.dtype_and_values(
            available_dtypes=helpers.get_dtypes("float", full=False),
            min_value=-1000,
            max_value=1000,
            max_dim_size=10,
            max_num_dims=4,
            min_dim_size=1,
        ),
    )

ivy/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py:272:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
ivy/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py:301: in test_tensorflow_foldl
ivy/ivy_tests/test_ivy/helpers/function_testing.py:608: in test_frontend_function
    ret = get_frontend_ret(
ivy/ivy_tests/test_ivy/helpers/function_testing.py:1774: in get_frontend_ret
    ret = frontend_fn(*args, **kwargs)
ivy/ivy/functional/frontends/tensorflow/func_wrapper.py:135: in _outputs_to_frontend_arrays_tf
    ret = fn(*args, **kwargs)
ivy/ivy/functional/frontends/tensorflow/func_wrapper.py:109: in _inputs_to_ivy_arrays_tf
    return fn(*ivy_args, **ivy_kwargs)
ivy/ivy/functional/frontends/tensorflow/general_functions.py:107: in foldl
    return ivy.reduce(elems[1:], elems[0], fn)

E           ivy.utils.exceptions.IvyBackendException: tensorflow: reduce:   File "/ivy/ivy/ivy/utils/exceptions.py", line 210, in _handle_exceptions
E               return fn(*args, **kwargs)
E             File "/ivy/ivy/ivy/func_wrapper.py", line 972, in _handle_nestable
E               return fn(*args, **kwargs)
E             File "/ivy/ivy/ivy/func_wrapper.py", line 482, in _inputs_to_ivy_arrays
E               return fn(*ivy_args, **ivy_kwargs)
E             File "/ivy/ivy/ivy/functional/ivy/experimental/general.py", line 72, in reduce
E               computation = _correct_ivy_callable(computation)
E             File "/ivy/ivy/ivy/functional/ivy/experimental/general.py", line 19, in _correct_ivy_callable
E               return ivy.__dict__[func.__name__]
```
I will continue to look into using `ivy.reduce()` but as of now using `functools.reduce` works fine for the functions being tested for. Perhaps trying to mirror the `tf.foldl` implementation more accurately and use a `ivy.while_loop` would be simpler.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py
FourFortyFour: A side note on a remark previously made by you regarding having all the same parameters to the function as `tf.foldl`; I removed the `back_prop` arg as it is deprecated, and will be removed in coming version of tensorflow. Instead using `ivy.stop_gradient` would be the way to go.

xoiga123: > I am not sure what you meant by this, I tried looking for examples of how certain test functions or other function implementations deal with the name argument but wasn't able to arrive at a clearer picture.

Yeah don't test for `name`, but please include `parallel_iterations` and `swap_memory` even if they don't do anything for now. We may update the frontend to make use of them in the future and forget that the test didn't account for them.

ivy/functional/frontends/tensorflow/general_functions.py
xoiga123: Ah the use of `ivy.atleast_1d` is nice. Yea it looks pretty simple now, I just thought maybe `ivy.reduce` could be used, but good job :+1:

FourFortyFour: Got it, however this does lead to a discrepancy between ivy's implementation and tensorflow's, as tensorflow will not accept a scalar while ours will.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_general_functions.py
FourFortyFour: Consider it done, will update asap 🤝.

ivy/functional/frontends/tensorflow/general_functions.py
xoiga123: @FourFortyFour Eh what? So what was the point of `atleast_1d`?

FourFortyFour: I will change it right now, it was because I confused `min_dim_size` and `min_num_dims` in the `dtype_and_values` helper (your question made me reevaluate everything 😅), so when the test function would run, testing `tf.foldl()` would always fail in the `tf.TensorShape`'s `__getitem__` method. It would sometimes be passed scalars, so I added the conversion `ivy.atleast_1d()` to get around that... I thought there was something weird going. Will push fixes right now, thank you for revealing this.

FourFortyFour:
Hello once more! Updated the test function to include the unused function parameters (`parallel_iterations, swap_memory`,and `name`). Also removed unnecessary dimension and shape conversions with `ivy.atleast_1d()` in the foldl implementation. However, the PaddlePaddle test fails due to a shape `AssertionError`. Examining the error further it always occurs with the following test parameters: 
```
AssertionError: returned shape = (), ground-truth returned shape = (1,)
E               Falsifying example: test_tensorflow_foldl(
E                   frontend='tensorflow',
E                   fn=lambda a, b: a + b,
E                   initializer=None,
E                   dtype_and_values=(['float64'], [array([-1.])]),
E                   parallel_iterations=10,
E                   swap_memory=False,
E                   name=None,
E                   fn_tree='ivy.functional.frontends.tensorflow.foldl',
E                   test_flags=FrontendFunctionTestFlags(
E                       num_positional_args=0,
E                       with_out=False,
E                       inplace=False,
E                       as_variable=[False],
E                       native_arrays=[False],
E                       generate_frontend_arrays=False,
E                   ),
E               )
E
```
As you can see its an array with one element in it (`[-1.]`), examining what `ivy.foldl` and `tf.foldl` return in such a case shows that they are behaving similarly.
- `ivy.foldl()`:
```python
In [10]: np.shape(foldl(lambda x,y:x+y, [-1.])) == ()
Out[10]: True

In [11]: foldl(lambda x,y:x+y, np.array([-1.]))
Out[11]: -1.0

In [12]: np.isscalar(foldl(lambda x,y:x+y, np.array([-1.])))
Out[12]: True

In [14]: foldl(lambda x,y:x+y, [-1.])
Out[14]: -1.0
``` 

- `tf.foldl()`:
```python
In [22]: np.shape(tf.foldl(lambda x,y: x+y, np.array([-1.])))
Out[22]: TensorShape([])

In [23]: np.shape(tf.foldl(lambda x,y: x+y, np.array([-1.]))) == ()
Out[23]: True

In [24]: tf.foldl(lambda x,y: x+y, np.array([-1.]))
Out[24]: <tf.Tensor: shape=(), dtype=float64, numpy=-1.0>
                                                                                                                                                                                                                   In [25]: np.isscalar(tf.foldl(lambda x,y: x+y, np.array([-1.])))
Out[25]: False
```
Some key differences to observe:
- `tf.foldl` wont accept a normal python list 
- `ivy.foldl` will return an actual scalar while `tf.foldl` will return a Scalar Tensor (aka an `EagerTensor` with no ndims attribute) 

How those differences could potentially be causing or affecting the PaddlePaddle test to fail the shape assertion are unclear to me currently, I will continue exploring and try to figure it out. 

:)

xoiga123:
@FourFortyFour Seems like there's another bug with paddle indexing, I'll look into it

ivy/functional/frontends/tensorflow/general_functions.py
```diff
@@ -83,6 +84,37 @@ def fill(dims, value, name=None):
...
+    name=None,
+):
+    ivy.utils.assertions.check_isinstance(
+        elems, (list, ivy.Array), "elems must be an iterable object"
+    )
+    ivy.utils.assertions.check_true(
+        callable(fn), f"{fn.__name__} must be a callable function"
+    )
+    if len(ivy.shape(elems)) == 0 or ivy.get_num_dims(elems) == 0:
+        ivy.utils.exceptions.IvyValueError(
```
xoiga123, line 103: `raise` please :black_flag:

xoiga123 (changes requested):

ivy/functional/frontends/tensorflow/general_functions.py
FourFortyFour: Done.

xoiga123 (approved):
LGTM, we got the green flag to merge even though paddle's `get_item` is failing. Thank you for contributing :star:

FourFortyFour:
Glad I could help contribute, and thank you so much for your time @xoiga123 I truly appreciate it!

Issue #10070, opened 2023-01-25
Fix pickling.test_pickle_to_string


ivy-leaves:
|   |   |
|---|---|
|jax|<a href="https://github.com/unifyai/ivy/actions/runs/5559169389/jobs/10155003945"><img src=https://img.shields.io/badge/-success-success></a>
|numpy|<a href="https://github.com/unifyai/ivy/actions/runs/5559169389/jobs/10155003945"><img src=https://img.shields.io/badge/-success-success></a>
|tensorflow|<a href="https://github.com/unifyai/ivy/actions/runs/5559169389/jobs/10155003945"><img src=https://img.shields.io/badge/-success-success></a>
|torch|<a href="https://github.com/unifyai/ivy/actions/runs/5559169389/jobs/10155003945"><img src=https://img.shields.io/badge/-success-success></a>
|paddle|<a href="https://github.com/unifyai/ivy/actions/runs/5559169389/jobs/10155003945"><img src=https://img.shields.io/badge/-success-success></a>

PR #16187, opened 2023-06-02
fixing misc pickling test

abdulasiraj:
closes

- [ ]  #10070
- [ ] #10069

### DIFF
--- a/ivy_tests/test_ivy/test_misc/test_pickling.py
+++ b/ivy_tests/test_ivy/test_misc/test_pickling.py
@@ -1,6 +1,7 @@
 import pickle
 import numpy as np
 import os
+import paddle
 
 from hypothesis import given, assume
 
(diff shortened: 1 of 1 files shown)


ivy_tests/test_ivy/test_misc/test_pickling.py
```diff
@@ -47,8 +47,9 @@ def test_pickle_to_string(dtype_and_x, on_device):
 )
 def test_pickle_to_and_from_disk(dtype_and_x, on_device):
     input_dtype, x = dtype_and_x
-    assume("bfloat16" not in input_dtype)
+    assume(["bfloat16"] not in input_dtype)
```
VedPatwardhan, line 50: I think we can revert this change, given that `input_dtype` is a list of str?

VedPatwardhan (changes requested):
Hey @abdulasiraj, thanks for making the changes, could you please only do the `to_numpy` conversion when testing the `paddle` backend? Also, it would be helpful if you could add a comment above that conversion saying that paddle tensors can't be pickled directly and linking [this](https://github.com/PaddlePaddle/Paddle/issues/41107) issue? Thanks 🙂

VedPatwardhan (approved):
lgtm! Feel free to get the PR merged. Thanks @abdulasiraj @zaeemansari70 🙂

PR #16193, opened 2023-06-02
Add dependency installer script for making the package installation process in doc-builder more generic

bipinKrishnan:
Currently, the doc-builder installs the packages required for each project by going through if-else statements as there are slight changes in the commands used for each project. Adding install_dependencies.sh script will allow the doc builder to generalise this process by having to just look for this script in the project directory.

The PR is related to the below task:
https://trello.com/c/VG4SYszL

### DIFF
--- a/install_dependencies.sh
+++ b/install_dependencies.sh
@@ -1,0 +1,10 @@
+pip install torch
+ver=$(pip show torch | grep Version | cut -d ' ' -f2)
+pip install torch-scatter -f https://data.pyg.org/whl/torch-$ver+cpu.html
+pip install -r requirements/requirements.txt
+if [[ $(arch) == 'arm64' ]]; then
+      pip install -r requirements/optional_m1_1.txt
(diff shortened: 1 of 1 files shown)


install_dependencies.sh
```diff
@@ -0,0 +1,10 @@
+pip install torch
+ver=$(pip show torch | grep Version | cut -d ' ' -f2)
+pip install torch-scatter -f https://data.pyg.org/whl/torch-$ver+cpu.html
```
KareemMAX, line 3: Not sure if that's still needed, can you check that?

KareemMAX (approved):
LGTM!

If they passes manual test with the doc builder feel free to merge. I've just added a comment on the torch installation

install_dependencies.sh
bipinKrishnan: The doc building failed the last time I tested without these lines, will re-check it and make the changes if the build is successful this time.

PR #16218, opened 2023-06-03
Fix compiler demos

Sam-Armstrong:
@sherry30 
Hey Sherry,

These two lines that you added the other day (https://github.com/unifyai/ivy/pull/14306) are causing a couple of the compiler demos to fail with this error:

ivy.utils.exceptions.IvyBackendException: numpy: broadcast_to: TypeError: descriptor '__getitem__' for 'list' objects doesn't apply to a 'tuple' object

I've removed these lines here which fixes the tests. Is this ok with you?


Sam-Armstrong:
For context, you can see the failures on the demo checks of https://github.com/unifyai/graph-compiler/pull/173

sherry30:
Hi Sam,

Sorry for the inconvenience, I added these to unify the exceptions. It ensures that the same type of Exception error is raised for the same type of error(`x.shape` not being broadcastable to `shape` in this case) regardless of the backend.
So I guess this is an error we can fix in the backend instead of removing the assertion 🤔.
Can you give me a minimal example on which this error occurs, I can't see the failures since I don't have access to the graph compiler repo 😅.

Also while looking into this, I noticed a few other inconsistencies with this assertion in the `assertion.py` file. Is it fine if I push some changes to those in this PR? (They might fix this error but I can't say for certain since I'm not sure what's causing the error yet) If you think its not appropriate to do it here, I can push it separately to the master.

Thanks 😄

Sam-Armstrong:
Hi Sherry,

While trying to create the minimal example I actually found the root cause of the problem 😅. I've put the assertions back in alongside the fix, let me know if you think it looks ok. And yeah feel free to add more changes to this PR

Sam-Armstrong:
@sherry30 We've got a fix merged into the graph compiler for this problem now, so I've removed all my changes in this PR. So feel free to either merge the changes you made here, or close the PR. Thanks :)

PR #16225, opened 2023-06-04
added nanmin, nanmax experimental features

g-1f:
#3856 extending experimental api, including nanmin and nanmax function for numpy, jax, paddle, tensorflow and pytorch backends.


g-1f:
this lint fixing does not end lol.

PR #16239, opened 2023-06-04
Fixes inplace_update function

usama-baloch:


YushaArif99:
@sherry30, I think you'd be in a better position to review this PR given that you were the author of the trello card corresponding to this PR 🙂

VedPatwardhan:
Given that @sherry30's authored the task, I'll probably leave this with you guys. Feel free to let me know if there's anything I can help with, thanks 🙂

ivy/functional/backends/torch/general.py
```diff
@@ -297,10 +297,13 @@ def inplace_update(
         if keep_input_dtype:
             val = ivy.astype(val, x.dtype)
         (x_native, val_native), _ = ivy.args_to_native(x, val)
+        x_native.data = val_native
```
sherry30, line 300: You don't have to add this line, the lines below basically in place updates the native array

ivy/functional/backends/jax/general.py
```diff
@@ -208,8 +208,8 @@ def inplace_update(
     keep_input_dtype: bool = False,
 ) -> ivy.Array:
     if ivy.is_array(x) and ivy.is_array(val):
-        if ensure_in_backend:
-            raise ivy.utils.exceptions.IvyException(
+        if ensure_in_backend or ivy.is_native_array(x):
```
sherry30, line 211: Lets add the check like `if ensure_in_backend or (ivy.is_native_array(x) and not ivy.get_array_mode())`

I guess it would be better to raise an error only if tensor was passed when array_mode was False, otherwise, we will be returning an ivy array anyway so I guess its fine in that case not to raise an error.
and same case with tensorflow

ivy/func_wrapper.py
```diff
@@ -454,6 +454,14 @@ def _inputs_to_ivy_arrays(*args, **kwargs):
         -------
             The return of the function, with ivy arrays passed in the arguments.
         """
+        if not ivy.get_array_mode():
+            x = ivy.nested_map(args, lambda x: ivy.is_native_array(x))
```
sherry30, line 458: Lets not add checks for native arrays, I guess its safe to assume that the user will pass native array when array_mode is False, since we halt all conversions of arrays in that case anyway. Its not expected to work with ivy arrays.

sherry30 (changes requested):
Requested a few changes.

Can you also change the docs of `ivy.set_array_mode` and add that it also stops the conversion of native arrays to ivy arrays when set to False.

Thanks 😄

ivy/func_wrapper.py
VedPatwardhan: Yep absolutely, we're free to assume that it's a native array 👍

ivy/functional/backends/jax/general.py
```diff
@@ -208,7 +208,7 @@ def inplace_update(
     keep_input_dtype: bool = False,
 ) -> ivy.Array:
     if ivy.is_array(x) and ivy.is_array(val):
-        if ensure_in_backend:
+        if ensure_in_backend or (ivy.is_native_array(x) and not ivy.get_array_mode()):
```
sherry30, line 211: One thing I realized, we don't have to check for `ivy.get_array_mode()` since `x` can only be native if this mode was disabled 😅.
So we can remove this check  here and in other backends as well.

sherry30 (changes requested):
1 change requested.
Can you also make these changes in numpy and paddle paddle backend.
The rest seems to be working fine.

ivy/functional/backends/jax/general.py
usama-baloch: Yes, you are right, I pushed the new changes. I hope, Now this will work fine 🌝

sherry30 (approved):
lgtm!

Feel free to merge it.

PR #16256, opened 2023-06-05
Make dtype_and_values take dtype without passing available_dtypes as well

somaia02:
When dtype_and_values is called with dtype only and not available_dtypes I get an error no function to prune. I don't know how to reproduce this outside the test file I'm working on but this is an example of the problem:
```
# This works
helpers.dtype_and_values(dtype=["float64"], available_dtypes=("float64"))

# This doesn't
helpers.dtype_and_values(dtype=["float64"])
```
If I pass dtype, I shouldn't have to pass available_dtypes as well.


CatB1t (approved):
lgtm! let's just make sure no new failures are introduced, otherwise it's up to @nassimberrada to get it merged 👍🏻. Thanks @somaia02!

PR #16259, opened 2023-06-05
Implement `handle_versioned_function` decorator to lazily dispatch the function to a version-specific implementation

hmahmood24:
This PR implements a new decorator to lazily dispatch the function to a version-specific implementation only when the function is actually called by the user rather than at import-time, as previously implemented.

The previous policy was implemented s.t. for the frontends, we iterated over the frontend's `__dict__` at import-time to change the mapping based on the framework version installed on the user's machine by importing the native framework (`paddle`, `torch`, `tensorflow` etc.). This caused the frontend imports to slow down since we not only imported a native framework at import-time but we also iterated over the entirety of the frontend’s `__dict__` in order to find the version-specific functions. A small point to note that currently a similar policy is applied in the backends as well but this PR focuses on refactoring the frontends based on my discussion with @RickSanchezStoic on Discord and also since the frontends directly affect the transpiler.

The updated policy shall now dictate that the dispatching takes place inside a decorator called `handle_versioned_function` which gets applied to the function we want to provide version support for. As an example, currently there are two functions in the torch frontend that support versioning: `ones_like_v_0p3p0_to_0p3p1` and `ones_like_v_0p4p0_and_above`

We will add another function with the actual function name i.e. `ones_like` with a generic signature but with no logic inside (this is the policy that should be followed moving forward). The decorator will then be applied to `ones_like` and will handle the dispatching at run-time based on the versioning logic which was previously being handled at import-time. 

This should significantly reduce the time it takes to import the frontends since now we only dispatch when that function is explicitly called and we don't need to iterate over the entirety of the frontend's `__dict__` in order to make the dispatch. I've also added a test to demonstrate the usability and working of the decorator. I'll update the relevant sections in the docs to also reflect this new policy once this PR is merged. Although this policy is designed for the frontends, I've deliberately tried to keep it generic in the sense that if we want to apply this policy to the backends as well, we would just need to add some logic under the `TODO` inside the decorator (kind of similar to what's been added for the frontend block) if we want to apply this decorator to the backend functions as well.

[Relevant Trello Task](https://trello.com/c/uwQed2On/1965-implement-a-handleversionedfunction-decorator-for-lazy-version-specific-dispatch)

### DIFF
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -1,12 +1,13 @@
 import contextlib
 import ivy
+import importlib
 import functools
 import logging
 import weakref
 import warnings
 import copy as python_copy
 from types import FunctionType
-from typing import Callable
+from typing import List, Callable
 import inspect
 
 
@@ -30,6 +31,7 @@ FN_DECORATORS = [
     "handle_exceptions",
     "handle_nans",
     "handle_mixed_function",
+    "handle_versioned_function",
 ]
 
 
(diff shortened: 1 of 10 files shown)


VedPatwardhan (commented):
Probably the review request wasn't intended, but being my OCD self, can we try and move the frontend-specific wrapper to a frontend-specific file 😂😅 ? Would keep things a bit modular, thanks @hmahmood24 🙂

VedPatwardhan:
> > Probably the review request wasn't intended, but can we try and move the frontend-specific wrapper to a frontend-specific file 😂😅 ? Would keep things a bit modular, thanks @hmahmood24 🙂
> 
> @VedPatwardhan You were a code owner which automatically triggered the review request. Also, aren't all function wrappers (including frontend-specific ones like `to_ivy_arrays_and_back` included in the `ivy/func_wrapper.py` file? I'm not sure where else to add the code for the decorator in the frontends 🤔 Also regarding the discussion in Discord [here](https://discord.com/channels/799879767196958751/1028707571743334420/1114142107129626697), were you able to form an opinion on what's the best approach for handling this moving forward?

Oh sorry, completely missed out on that convo, I'll go through it tomorrow morning and get back. Thanks for the helpful nudge @hmahmood24 🙂

PR #16271, opened 2023-06-05
sorted promotion tables

tomatillos:


VedPatwardhan:
Actually, there was another PR merged a few hours ago which removed duplicates from the promotion tables, could you please sort that table instead? Thanks and apologies @tomatillos 🙂

PR #16294, opened 2023-06-05
Added Jax Numpy round_ to Jax Frontend

lfcalmeida:


lfcalmeida:
The option to link the issue is not available to me in the Development section. #16279

sherry30 (changes requested):
Hi,

Sorry for the late review, this PR somehow slipped under my radar 😅.
Its weird that I'm not getting the option to link the issue as well 🤔, I'll close the issue myself once the PR gets merged in that case.

`round_` and `around` are an alias of `round`, so you can just add it like `round_ = around` and also the same for `round` function.
As for the testing, you can just add these as `aliases` in the `handle_frotnend_test` decorator of the `around` test function.

Thanks 😄

PR #16295, opened 2023-06-05
Throwing error on invalid devices

Fayad-Alman:
See Trello Card: https://trello.com/c/JINna7Je/1925-throwing-error-on-trying-to-use-invalid-devices-for-a-particular-backend

Currently, the following code would default to a cpu with the numpy backend:

import ivy
x = ivy.random_uniform(shape=(1, 1, 1, 2), device="gpu:0")
f = ivy.random_uniform(shape=(1, 1, 2, 2), device="cpu:0")
print(x.device, f.device)
Output: cpu cpu

The ideal behaviour should be throwing an error because gpu:0 is an invalid device with the numpy backend. Rather than hardcoding for numpy, we should instead be implementing this in a way that whenever a device is passed which is in the invalid devices for a particular backend, we should be throwing an error.


VedPatwardhan:
Hey @Fayad-Alman, thanks for looking into this task! Feel free to request a review when you're done making the changes 🙂

PR #16339, opened 2023-06-06
Fixes the unintended behavior created by ivy.Array.__getattr__

kgmann:
Fixed this https://trello.com/c/yLipjrid. 
The issue arises because the `to_ivy` function that's used expects arrays or iterables (e.g. `ivy.Array,` or `ivy.NativeArray`) but native framework methods like torch's `permute` get passed to it. I fixed this by adding a check for callables and converting the callables' output to ivy arrays.

### DIFF
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -28,7 +28,7 @@ from .set import _ArrayWithSet
 from .sorting import _ArrayWithSorting
 from .statistical import _ArrayWithStatistical
 from .utility import _ArrayWithUtility
-from ivy.func_wrapper import handle_view_indexing, outputs_to_ivy_arrays
+from ivy.func_wrapper import handle_view_indexing
 from .experimental import (
     _ArrayWithSearchingExperimental,
     _ArrayWithActivationsExperimental,
@@ -425,13 +425,7 @@ class Array(
         return super().__getattribute__(item)
 
     def __getattr__(self, item):
-        try:
-            attr = self._data.__getattribute__(item)
-        except AttributeError:
-            attr = self._data.__getattr__(item)
-        if callable(attr):
-            return outputs_to_ivy_arrays(attr)
-        return to_ivy(attr)
+        raise AttributeError(f"'ivy.Array' object has no attribute '{item}'")
 
     @handle_view_indexing
     def __getitem__(self, query):
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -28,7 +28,7 @@ from .set import _ArrayWithSet
 from .sorting import _ArrayWithSorting
 from .statistical import _ArrayWithStatistical
 from .utility import _ArrayWithUtility
-from ivy.func_wrapper import handle_view_indexing
+from ivy.func_wrapper import handle_view_indexing, outputs_to_ivy_arrays
 from .experimental import (
     _ArrayWithSearchingExperimental,
     _ArrayWithActivationsExperimental,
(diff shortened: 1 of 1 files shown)


vaatsalya123:
I'll let you take this one @VedPatwardhan if you don't mind , Since it's related more to the framework side and you would know how to better asses this since you wrote the orginal card for this ,  Thanks Ved  and ofcourse thanks @kgmann !

ivy/data_classes/array/array.py
```diff
@@ -390,6 +390,8 @@ def __getattr__(self, item):
             attr = self._data.__getattribute__(item)
         except AttributeError:
             attr = self._data.__getattr__(item)
+        if callable(attr):
```
VedPatwardhan, line 393: I think the ideal behaviour should be not allowing the use of `permute` in the first place, if it wasn't present in `ivy`. @Infrared1029 what do you think?

VedPatwardhan (commented):
I think @Infrared1029 is the author of this task, so I'll just request his review. Thanks for making the changes @kgmann, and apologies for the delay 😅

ivy/data_classes/array/array.py
Infrared1029: I agree, we should be adhering to our own API only. Ideally, this should fail (calling `permute`) without breaking the repo because I'm pretty sure this unintended behavior is sprinkled across the repo in some (could be many tbh) places.

kgmann: @vedpatwardhan @Infrared1029 that's what I thought too. But as you said, the '_feature_' seem to be used intentionally and unintentionally in the codebase, so it might break some stuff if we just naively prevent such method calls. So @Infrared1029 should we keep this quick fix for now and create another issue to look into this later ?

Infrared1029: hmm, I think if we keep this fix (allowing framework-specific methods to be called and returning an ivy.Array) it might reinforce the usage of such behavior more because the output is a valid ivy Array, my first time noticing this behavior is when I mistakenly used `ivy.permute` and noticed a `torch.Tensor` coming out.

thoughts @vedpatwardhan ?

vedpatwardhan: Hey @Infrared1029, yeah I think it would be much more clear to the user if they were never able to call framework-specific functions which didn't exist in `ivy` (e.g. `ivy.permute`) rather than it failing silently and them having to debug through a model with a bunch of functions to find that this function is deferring to a framework-specific function because of `getattr`. 
I agree with your point @kgmann, it might break some stuff but I think such usages should ideally break anyway. Ofc we wouldn't merge before fixing the issues arising from it. We can look through the PR workflows and fix all such failures if that's okay? 
Thanks both 🙂

ivy-seed:

ivy/data_classes/array/array.py
kgmann: Sure @vedpatwardhan @Infrared1029 , and sorry for the delay. I just made some changes now. I'm not that familiar with the intelligent tests in the PR workflows yet, but I will look into them later and see if they are any new failures associated to the changes.

ivy-seed:

ivy-seed:

ivy-seed:

kgmann:
Hey @Infrared1029 ,  @vedpatwardhan . This slipped under the radars a few times already. I inspected some of the failing tests and none of them is caused by the changes in this PR. They all have to do with the testing pipeline itself (missing module) and I'll notify the testing team about it. Also, once it's merged it should be pretty straightforward to spot any tests that would fail because of it anyway.

vedpatwardhan (commented):
Hey @kgmann, actually it seems like a few tests are failing as a result of these changes, e.g. `scatter_nd`. Probably given that we didn't have an idea of this issue with `__getattr__` until very recently, seems like quite a few functions are implicitly making use of the incorrect `__getattr__` behaviour as @Infrared1029 suspected. 
Let's see what @Infrared1029 thinks about all this, but I certainly believe we should dive into why these tests are failing because of the changes and make necessary fixes for the same before merging.
Thanks @kgmann 😄

PR #16362, opened 2023-06-07
Handling decorators for mixed functions

hello-fri-end:
https://trello.com/c/gI9oE52O/1991-handling-decorators-for-mixed-functions


hello-fri-end:
Hey @VedPatwardhan, I have made the changes as we discussed and everything seems to work as expected. However, just wanted to bring the following point to your notice: Since we moved the `handle_mixed_function` decorator before `handle_nestable`, the compositional implementation will always have this decorator as a side effect just because `handle_mixed_function` will always be executed after `handle_nestable`. I can't think of a way to avoid this. Please let me know if this is a problem.

hello-fri-end:
ivy-gardener

VedPatwardhan (changes requested):
> Hey @VedPatwardhan, I have made the changes as we discussed and everything seems to work as expected. However, just wanted to bring the following point to your notice: Since we moved the `handle_mixed_function` decorator before `handle_nestable`, the compositional implementation will always have this decorator as a side effect just because `handle_mixed_function` will always be executed after `handle_nestable`. I can't think of a way to avoid this. Please let me know if this is a problem.

Hey @hello-fri-end, just requested a small change, the rest seems good to go. We can probably run a few tests to be sure before merging of course.
Good point, but we could probably just add `handle_nestable` to `to_add` and remove it from the decorators if it isn't applicable to the compositional implementation right? ig if we do that the `handle_nestable` decorator wouldn't be used for the compositional implementation.
Thanks 🙂

hello-fri-end:
> Hey @hello-fri-end, just requested a small change, the rest seems good to go.

Don't see any changes requested :thinking: :sweat_smile: 

> Good point, but we could probably just add `handle_nestable` to `to_add` and remove it from the decorators if it isn't applicable to the compositional implementation right? ig if we do that the `handle_nestable` decorator wouldn't be used for the compositional implementation. Thanks slightly_smiling_face

The problem is that until we reach the `handle_mixed_function` decorator we're running the primary implementation which means `handle_exceptions` and `handle_nestable` have already been executed. Therefore, the compositional implementation will indirectly always have these two decorators. Hope that makes it clear :sweat_smile:

VedPatwardhan:
> Don't see any changes requested 🤔 😅

Sorry about that, probably I forgot to click the button lol. I just wanted to suggest having a different name for the `handle_backend_wrappers` attribute, which preferably has the word `mixed` to avoid confusion. We can remove the `handle_` if the attribute name gets too long 🙂

VedPatwardhan:
> The problem is that until we reach the `handle_mixed_function` decorator we're running the backend function which means `handle_exceptions` and `handle_nestable` have already been executed. Therefore, the compositional implementation will indirectly always have these two decorators. Hope that makes it clear 😅

Ah right makes sense!

VedPatwardhan:
I think the only way to avoid calling the decorator twice would be to figure out a way to make the nestable call when we call `condition` in the `handle_mixed_function`, which would be a bit redundant. Happy to know if you have any other ideas, thanks @hello-fri-end 🙂

hello-fri-end:
> I think the only way to avoid calling the decorator twice would be to figure out a way to make the nestable call when we call `condition` in the `handle_mixed_function`, which would be a bit redundant. Happy to know if you have any other ideas, thanks @hello-fri-end slightly_smiling_face

Hey @VedPatwardhan, I spent some time considering whether having the `handle_nestable` decorator in compositional implementations would impact the performance and I realized that it's quite the opposite and it's actually more efficient to include it since it would mean that we traverse each leaf of the container only once as opposed to traversing them multiple times i.e all the functions used by the compositional implementation will traverse each leaf once if the decorator is not added. And I agree handling the nestable call inside the `handle_mixed_function` decorator would be a bit redundant as it's equivalent to having the `handle_nestable` decorator added before. What we can do is remove the `handle_nestable` decorator from the compositional implementations of partial mixed functions inside the `_wrap_function` so that it doesn't get called twice. We can keep the rest as it is. Happy to hear your thoughts on this :)

VedPatwardhan:
> Hey @VedPatwardhan, I spent some time considering whether having the `handle_nestable` decorator in compositional implementations would impact the performance and I realized that it's quite the opposite and it's actually more efficient to include it since it would mean that we traverse each leaf of the container only once as opposed to traversing them multiple times i.e all the functions used by the compositional implementation will traverse each leaf once if the decorator is not added. And I agree handling the nestable call inside the `handle_mixed_function` decorator would be a bit redundant as it's equivalent to having the `handle_nestable` decorator added before. What we can do is remove the `handle_nestable` decorator from the compositional implementations of partial mixed functions inside the `_wrap_function` so that it doesn't get called twice. We can keep the rest as it is. Happy to hear your thoughts on this :)

I'm a bit hesitant to add that kind of handling given that we're generalizing similar occurrences with `to_add` and `to_skip`, but probably there's no other choice. Let's just go ahead and remove all wrappers which come before `handle_mixed_function` in `FN_DECORATORS` to avoid iterating through them twice. Thanks @hello-fri-end 🙂

hello-fri-end:
> > Hey @VedPatwardhan, I spent some time considering whether having the `handle_nestable` decorator in compositional implementations would impact the performance and I realized that it's quite the opposite and it's actually more efficient to include it since it would mean that we traverse each leaf of the container only once as opposed to traversing them multiple times i.e all the functions used by the compositional implementation will traverse each leaf once if the decorator is not added. And I agree handling the nestable call inside the `handle_mixed_function` decorator would be a bit redundant as it's equivalent to having the `handle_nestable` decorator added before. What we can do is remove the `handle_nestable` decorator from the compositional implementations of partial mixed functions inside the `_wrap_function` so that it doesn't get called twice. We can keep the rest as it is. Happy to hear your thoughts on this :)
> 
> I'm a bit hesitant to add that kind of handling given that we're generalizing similar occurrences with `to_add` and `to_skip`, but probably there's no other choice. Let's just go ahead and remove all wrappers which come before `handle_mixed_function` in `FN_DECORATORS` to avoid iterating through them twice. Thanks @hello-fri-end slightly_smiling_face

To be honest, I feel the same about it - for once, the `wrap_function` was looking so clean and readable :smiling_face_with_tear:   but doesn't look like we have any other option here but to have this exception for partial mixed functions. For now, I have updated it to remove `handle_nestable` and `handle_exceptions` if both are present, and otherwise only `handle_exceptions` will be removed (`handle_nans` is not used anywhere in the codebase, I wonder why we even have it :thinking: )

ivy/func_wrapper.py
```diff
@@ -825,7 +825,6 @@ def _infer_device(*args, device=None, **kwargs):
 
 def handle_out_argument(fn: Callable) -> Callable:
     handle_out_in_backend = hasattr(fn, "support_native_out")
-    is_compos_fn = "ivy.functional.ivy" in fn.__module__
```
VedPatwardhan, line 828: We'd still need the `is_compos_fn` attribute right? If it's the compositional implementation, we should just pass the arguments as they are, without the need of an explicit `inplace_update` 🙂

ivy/func_wrapper.py
```diff
@@ -1010,43 +1009,55 @@ def _wrap_function(
...
+                    continue
+                if attr in add_wrappers:
+                    to_wrap = getattr(ivy, attr)(to_wrap)
+
+        # we should remove the all the decorators
+        # after handle_mixed_fuction in FN_DECORATORS
+        # from the compos function because these will
+        # be run from the primary implementation.
+        if partial_mixed:
+            array_spec = to_wrap.compos.__dict__["array_spec"]
```
VedPatwardhan, line 1053: is there an issue with `array_spec`? Would be great to know more about this

ivy/func_wrapper.py
```diff
@@ -1010,43 +1009,55 @@ def _wrap_function(
...
+            if hasattr(to_wrap.compos, "handle_exceptions") and hasattr(
```
VedPatwardhan, line 1054: let's instead generalize this handling to all decorators which come before `handle_mixed_function` in `FN_DECORATORS`, as it'll be easier to maintain even if `FN_DECORATORS` has a couple more decorators below `handle_mixed_function` sometime in the future

VedPatwardhan (changes requested):
Hey @hello-fri-end, yeah not sure why `handle_nans` isn't being used, probably worth diving into the commit history a bit regarding this. Just requested a few changes, thanks for looking into this task and apologies for the delayed response 😅

ivy/func_wrapper.py
hello-fri-end: So, the idea is that the compositional implementation will not have any decorators that are only needed for the primary implementation ( e.g `handle_out_argument`) because we'll be adding these from the `add_wrappers` tuple inside the `_wrap_function`. And since, `handle_out_argument` will never be called with respect to the compositional implementation, there's no need for the `is_compos_fn` attribute.

ivy/func_wrapper.py
hello-fri-end: It's applied after wrapping with `handle_exceptions` and therefore gets removed when we remove the wrapper. So, we need to store it and reapply it after removing the wrappers.

ivy/func_wrapper.py
VedPatwardhan: Ah makes sense. Thanks for clarifying!

ivy/func_wrapper.py
VedPatwardhan: Got it!

ivy/func_wrapper.py
hello-fri-end: Done!

ivy/func_wrapper.py
```diff
@@ -1023,43 +1022,52 @@ def _wrap_function(
...
+                if attr in add_wrappers:
+                    to_wrap = getattr(ivy, attr)(to_wrap)
+
+        # we should remove the all the decorators
+        # after handle_mixed_fuction in FN_DECORATORS
+        # from the compos function because these will
+        # be run from the primary implementation.
+        if partial_mixed:
+            array_spec = to_wrap.compos.__dict__["array_spec"]
+            for attr in FN_DECORATORS[-1:14:-1]:
```
vedpatwardhan, line 1067: sorry to be annoying, but we shouldn't be hardcoding with `14` either, it should instead be whatever index `handle_mixed_function` corresponds to 😅

vedpatwardhan (changes requested):
One last change, let's look into the test results and get the docs updated with the example after this one. Thanks @hello-fri-end 🙂

ivy/func_wrapper.py
hello-fri-end: Oh, I'm sorry I didn't realize this. Thanks a lot for pointing this out. Have the updated the function now :smile:

hello-fri-end:
ivy-gardener

hello-fri-end:
> One last change, let's look into the test results and get the docs updated with the example after this one. Thanks @hello-fri-end slightly_smiling_face

Thanks for the reviews @vedpatwardhan :rocket:  All done here  - I have already updated the relevant sections of the deep-dive. Please let me know if I have missed anything :)

vedpatwardhan (approved):
lgtm! Most of the intelligent tests are passing, which should imply that there's certainly not something fundamentally breaking the code. Let's get the PR merged and be on the lookout for any issues, thanks @hello-fri-end 🙂

PR #16388, opened 2023-06-07
better way of casting dtype for paddle.subtract function

Killua7362:
Paddle.subtract supports int64,int32,float64,float32. in case other data types are used.


MahmoudAshraf97:
paddle subtract supports more dtypes than the 4 mentioned, I included the actual non supported dtypes regardless of what is mentioned in the documentation

Killua7362:
But in the docs it was mentioned that the inputs support these 4 data types. I might be mistaken. Sorry for the trouble

PR #16395, opened 2023-06-07
Add dropout2d and refactor dropout into primary and partial mixed implementations

hmahmood24:
- Add dropout2d to Ivy Experimental API and related backends
- Add dropout to torch backend as a partial mixed function and to tensorflow backend as a primary function
- Include testing with out for all dropout functions and fix existing dropout tests

### DIFF
--- a/ivy/functional/backends/tensorflow/experimental/layers.py
+++ b/ivy/functional/backends/tensorflow/experimental/layers.py
@@ -576,6 +576,24 @@ def fft(
     return ret
 
 
+def dropout(
+    x: Union[tf.Tensor, tf.Variable],
+    prob: float,
+    /,
+    *,
+    scale: bool = True,
+    dtype: tf.DType = None,
+    training: bool = True,
+    seed: Optional[int] = None,
+    noise_shape: Optional[Sequence[int]] = None,
+    out: Optional[Union[tf.Tensor, tf.Variable]] = None,
+) -> Union[tf.Tensor, tf.Variable]:
+    x = ivy.astype(x, dtype) if dtype else x
+    res = tf.nn.dropout(x, prob, noise_shape=noise_shape, seed=seed) if training else x
+    res = tf.multiply(res, (1.0 - prob)) if not scale else res
+    return res if not ivy.exists(out) else ivy.inplace_update(out, res)
+
+
 def dropout1d(
     x: Union[tf.Tensor, tf.Variable],
     prob: float,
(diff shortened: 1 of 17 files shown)


ivy/functional/backends/tensorflow/experimental/layers.py
```diff
@@ -576,6 +576,24 @@ def fft(
...
+    dtype: tf.DType = None,
+    training: bool = True,
+    seed: Optional[int] = None,
+    noise_shape: Optional[Sequence[int]] = None,
+    out: Optional[Union[tf.Tensor, tf.Variable]] = None,
+) -> Union[tf.Tensor, tf.Variable]:
+    x = ivy.astype(x, dtype) if dtype else x
+    res = tf.nn.dropout(x, prob, noise_shape=noise_shape, seed=seed) if training else x
+    res = tf.multiply(res, (1.0 - prob)) if not scale else res
+    return res if not ivy.exists(out) else ivy.inplace_update(out, res)
```
sherry30, line 594: You don't have to do inplace_update here. it doesn't work with native tf Eager Tensors. You can instead add `handle_out_argument` on the ivy function and it should work fine.
Same should be changed with other `inplace_updates` in the backends.

sherry30 (changes requested):
just 1 change requested.
I think we can add a backend implementation of `dropout` in numpy as well. Let me know what you think. 
The rest looks good.

ivy/functional/backends/tensorflow/experimental/layers.py
hmahmood24: @sherry30 That's a good point. I totally forgot that decorator existed haha. Added it where it was necessary. 

Regarding `numpy`, I didn't add the implementation to any backend which doesn't support the generic implementation of `dropout` natively which `numpy` doesn't. For such backends, the already existing composition implementation in `ivy` should work fine as it was previously imo. What do you think?

sherry30: I thought of adding dropout to numpy backend since you added dropout2d to it as well. It will probably be faster than ivy's compositional implementation also resulting in less nodes in the transpiled graphs. But we can add it sometime later, its not super important I guess.

hmahmood24: Yeah it's not super important for the time being. Of course at some point we will have to revisit this and add the implementations of the generic `dropout` to all backends and remove the compositional implementation in `ivy` but for now I think we can make do with `torch` and `tensorflow`

PR #16512, opened 2023-06-10
Added Embedding Stateful class and the associated test function.

### LINKED ISSUE #16496: embedding

baberabb:
closes #16496
Added Embedding Stateful class and the associated test function. Had to add some decorators to ivy.experimental.layer.embedding to sort out some type promotion issues.

The backend/frontend functions are all in the experimental stage right now (as is the ivy function), so I'm not sure if creating the class is a bit premature. The functions (backend as well as ivy) also don't implement all the parameters such as padding. I thought about adding them while I'm at it but not sure if it's appropriate here or in a new PR?

PS: Made a small edit in the last class' (AvgPool3D) docstring as the pre-commit hook wouldn't let me commit (line too long).

### DIFF
--- a/ivy/functional/backends/jax/random.py
+++ b/ivy/functional/backends/jax/random.py
@@ -179,6 +179,8 @@ def shuffle(
     seed: Optional[int] = None,
     out: Optional[JaxArray] = None,
 ) -> JaxArray:
+    if x.shape == ():
+        return x
     if seed:
         rng_input = jax.random.PRNGKey(seed)
     else:
--- a/ivy/functional/frontends/tensorflow/general_functions.py
+++ b/ivy/functional/frontends/tensorflow/general_functions.py
@@ -52,9 +52,10 @@ def clip_by_global_norm(t_list, clip_norm, use_norm=None):
     ], global_norm
 
 
+@with_supported_dtypes({"2.12.0 and below": ("float", "complex")}, "tensorflow")
 @to_ivy_arrays_and_back
 def clip_by_norm(t, clip_norm, axes=None):
-    t = ivy.array(t)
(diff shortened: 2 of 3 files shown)


Daniel4078:
The experimental module of Ivy currently provides a common part of the features present in the similar (special or vastly different between frameworks) methods. So the frontend part is only limited to the Ivy as it will be used by the actural frontend function outside of the experimental module compositionally to achieve the specific behavior each framework chose to implement for their version of the function while you need to implement this common behavior with each framework in the backends. So you can add the extra parameters in here as it will make the future usage of the experimental Ivy function easier.

baberabb:
Hi @Daniel4078. Thanks for the helpful suggestions. So I tried implementing the new parameters but might have gone a bit overboard: Started off by just amending the main embedding function to add the new params, but to pass the tests, needed to alter all usages if it. Let me know what you think. 
Most of the tests are failing, I think, because Tensor flow is used as the ground truth and I didn't implement the new params on it.

Daniel4078:
> Hi @Daniel4078. Thanks for the helpful suggestions. So I tried implementing the new parameters but might have gone a bit overboard: Started off by just amending the main embedding function to add the new params, but to pass the tests, needed to alter all usages if it. Let me know what you think. Most of the tests are failing, I think, because Tensor flow is used as the ground truth and I didn't implement the new params on it.

Its fine, I suppose you can then just do the basics.

Daniel4078 (changes requested):
You current pull request includes back a lot of old problems in the repo as you look at the many extra files differ from the master branch. You should try create a clean copy of the master branch and add your necessary changes again there to remove the pollution of old fixed changes. Thanks.

PR #16514, opened 2023-06-11
Jax frontend histogram function and test

### LINKED ISSUE #16509: histogram

mam15:
Close #16509


mam15:
Hi @sherry30 ! I tried running my tests locally, but I keep getting the error `KeyError: 'cpu'`, in the test I created as well as other tests from the Jax frontend. Do you have any idea why this could be happening? Apparently it happened in github checks as well

sherry30 (changes requested):
Hi,

I'm not sure why you're getting that error, maybe some environment issue, make sure your setup is correct, also make sure you're updated with the master.
I got a different error when I ran the test, I've requested a change for that error.

Thanks 😄

mam15:
@sherry30 I found out what was causing the `KeyError: 'cpu'` error. Someone on Ivy's discord said that the correct jax version to use is 0.4.10 and I was using 0.4.12. As of now, jax's version is unpinned in the requirements, maybe it would be good to open a new issue and find out why 0.4.12 breaks the tests? I'm sure Ivy wants to support the latest versions of jax.

After that I was able to run tests locally and fix the code, but I ran into other issues:

Every backend function for histogram calculation has a "TODO: weird error when returning bins: return ret, bins_out". If we don't return the bins, the tests fail. Not sure what this cryptic "weird error" is about, but if I add the bins to the return statement it works. This happens when testing numpy, jax and tensorflow backends.

The torch test fails and says that bins must be `int` type, not `Tensor`. Not sure why that happens since in torch's documentation bins can be either `int` or `Tensor`.

The paddle test fails with an even stranger message, I have no idea what it means.

If you could point me in the right direction I would appreciate it

sherry30 (commented):
I think the Jax version is being worked on currently.
About the issue with histogram, we should return `bins` since that's part of the function. You can add it back in. I'm not sure why torch fails. Can you add any changes you've made to the backend function to this PR, I would like to take a deeper look and debug why its failing.

Thanks

sherry30:
Hi,
Sorry for the super late review.
Yea we can flag this as an issue in the backend and some other engineer can take a look at it. You can go ahead and prioritize some other frontend function for now.

Thanks

PR #16560, opened 2023-06-12
slicing ivy.Shape now return ivy Shape instance

Aarsh2001:
Problem - Slicing ivy.Shape object returns a native shape instead. Here’s a code snippet for reproducing this bug-:

```
ivy.set_backend("torch")
x = ivy.random_uniform(shape=(1, 2, 3))
print(type(x.shape)) # <class 'ivy.Shape'>
print(type(x.shape[:-1])) # <class 'torch.Size'>
```
Slicing shapes now returns ivy.Shape instances

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -327,7 +327,11 @@ class Shape:
         return super().__getattribute__(item)
 
     def __getitem__(self, key):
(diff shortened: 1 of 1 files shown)


VedPatwardhan (approved):
lgtm!

PR #16561, opened 2023-06-12
add paddle.tensor.math.log2

HenkaSyvert:


ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_math.py
```diff
@@ -617,3 +617,29 @@ def test_paddle_remainder(
         x=x[0],
         y=x[1],
     )
+
+
+# log2
+@handle_frontend_test(
+    fn_tree="paddle.log2",
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
```
AnnaTz, line 626: You should be drawing helpers.get_dtypes("valid") here. The test handler will find the supported dtypes taking into account the with_supported_dtypes decorator you added to the function. Other than that, great PR!

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_math.py
AnnaTz: 1. There have been some issues with our decorators which are currently being resolved. In this case you can ignore this error.  If the test was passing for the float dtypes (meaning the function's results are correct), we can still merge this.
2. We have open issues to go through all tests and update them. New commits should follow the "valid" approach. This is to make sure the functions have got the correct supported dtypes decorator.

AnnaTz (approved):
Great, thanks!

PR #16565, opened 2023-06-12
added `aliases`  for handle frontend method decorator

Aarsh2001:


ivy_tests/test_ivy/helpers/testing_helpers.py
```diff
@@ -707,6 +712,11 @@ def test_wrapper(test_fn):
                 "init_flags": init_flags,
                 "method_flags": method_flags,
                 "frontend_method_data": st.just(frontend_helper_data),
+                "init_tree": (
```
CatB1t, line 715: This will not work, `init_tree` and `method_tree` should be adjusted accordingly inside of `frontend_helper_data`.

CatB1t (changes requested):

ivy_tests/test_ivy/helpers/testing_helpers.py
Aarsh2001: Thought so

PR #16585, opened 2023-06-12
added check of _is_variable

usama-baloch:
updated the behaviour of all torch backend functions with support_native_out=True, using the _is_variable in handle_out_argument decorator.


ivy/func_wrapper.py
```diff
@@ -847,6 +848,10 @@ def _handle_out_argument(*args, out=None, **kwargs):
         """
         if out is None or is_compos_fn:
             return fn(*args, out=out, **kwargs)
+
+        if _is_variable(out):
```
VedPatwardhan, line 852: probably we can do a `handle_out_in_backend = handle_out_in_backend and _is_variable(out)` instead as that's a bit cleaner

VedPatwardhan (changes requested):
Hey @usama-baloch, thanks for making the changes, just one minor change requested 🙂

VedPatwardhan (changes requested):
Hey @usama-baloch, one last change and we're good to merge 🙂

ivy/func_wrapper.py
```diff
@@ -8,6 +8,7 @@
 from types import FunctionType
 from typing import Callable
 import inspect
+from ivy.functional.ivy.gradients import _is_variable
```
VedPatwardhan, line 11: Let's remove this import as it leads to a circular import error

ivy/func_wrapper.py
```diff
@@ -847,6 +848,7 @@ def _handle_out_argument(*args, out=None, **kwargs):
         """
         if out is None or is_compos_fn:
             return fn(*args, out=out, **kwargs)
+        handle_out_in_backend = handle_out_in_backend and not _is_variable(out)
```
VedPatwardhan, line 851: Given that we're not able to import `_is_variable` in this file, let's just move back to the way you were implementing it,
```
        if ivy.gradients._is_variable(out):
            handle_out_in_backend = False
```

VedPatwardhan (changes requested):
Hey @usama-baloch, just requested a couple of changes, we're ready to merge once they're done. Thanks 🙂

ivy/func_wrapper.py
usama-baloch: Sure 👍

PR #16629, opened 2023-06-13
updated docs, added examples, fixed a bug with casting modes

RickSanchezStoic:
Updated docs, added examples, and fixed a bug with casting modes not working well for cross casting

### DIFF
--- a/docs/overview/deep_dive/data_types.rst
+++ b/docs/overview/deep_dive/data_types.rst
@@ -388,6 +388,51 @@ An example of the latter is :func:`ivy.abs` with a tensorflow backend:
         else:
             return tf.abs(x)
 
+
+
+
+The :code: `[un]supported_dtypes_and_devices` decorator can be used for more specific cases where a certain
+set of dtypes is not supported by a certain device.
+
+.. code-block:: python
+    @with_unsupported_device_and_dtypes({"2.4.2 and below": {"cpu": ("int8", "int16", "uint8")}}, backend_version)
+    def gcd(
+        x1: Union[paddle.Tensor, int, list, tuple],
+        x2: Union[paddle.Tensor, float, list, tuple],
+        /,
+        *,
+        out: Optional[paddle.Tensor] = None,
+    ) -> paddle.Tensor:
+        x1, x2 = promote_types_of_inputs(x1, x2)
+        return paddle.gcd(x1, x2)
+
+
+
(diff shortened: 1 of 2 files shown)


docs/overview/deep_dive/data_types.rst
```diff
@@ -467,33 +512,168 @@ Data Type Casting Modes
 -----------------------
 
 As discussed earlier, many backend functions have a set of unsupported dtypes which are otherwise supported by the
-backend itself. This raises a question that whether we should support these dtypes by casting them to some other but close dtype.
-This is where we have various dtype casting modes so as to give the users an option to automatically cast unsupported dtype operations to a supported and a nearly same dtype.
+backend itself. This raises a question that whether we should support these dtypes by casting them to some other but close dtype. We avoid manually casting unsupported dtypes
+at all costs as this could be seen as undesirable behavior to some of users. This is where we have various dtype casting modes so as to give the users an option to automatically cast unsupported dtype operations to a supported and a nearly same dtype.
```
VedPatwardhan, line 516: let's remove `at all costs` from here, as it sounds a bit strong 😅

docs/overview/deep_dive/data_types.rst
```diff
@@ -467,33 +512,168 @@ Data Type Casting Modes
...
+    def expm1(x: torch.Tensor, /, *, out: Optional[torch.Tensor] = None) -> torch.Tensor:
+        x = _cast_for_unary_op(x)
+        return torch.expm1(x, out=out)
+
+The function :code:`expm1` has :code:`float16` as one of the unsupported dtypes, for the version :code:`2.0.1` which
+is being used for execution at the time of writing this. We will see how cating modes handles this.
+
+.. code-block:: python
+
+    import ivy
```
VedPatwardhan, line 555: I think we should avoid using `try-except` in example code-blocks, let's instead add a comment to the right of the function call, like, 
```
ret = ivy.expm1(ivy.array([1], dtype='float16'))    # raises Exception
ivy.upcast_data_types()
ret = ivy.expm1(ivy.array([1], dtype='float16'))    # doesn't raise an exception
```
Same applies to the other examples as well. We wouldn't need the code block for the output of each example either, as a result

VedPatwardhan (changes requested):
Hey @RickSanchezStoic, thanks for updating the docs, just requested a few changes 🙂

PR #16676, opened 2023-06-14
fixes bug with casting modes where unsupported_devices_and_dtypes was…

RickSanchezStoic:
ixes bug with casting modes where unsupported_devices_and_dtypes was being ignored in some places.


RickSanchezStoic:
Uses the default device for now. Might need an update when more precise device detecting functions are built.

PR #16692, opened 2023-06-14
Fix optional.txt and optional_gpu.txt requirements to fix torch-scatter causing install failure

fabriziocacicia:
By having `torch-scatter` alongside `torch` in the same requirements file, the `pip install` fails because the `torch-scatter` need `torch` to be already installed.

The separation is already in place for the `optional_m1_1.txt` and `optional_m1_2.txt` and it's also needed for `optional.txt` and `optional_gpu.txt` files.

I also added `optional_2.txt` in the new `install_dependencies.sh` script.

To add some further evidence, I'm attaching the logs of the failure:

```shell
Collecting torch
  Downloading torch-2.0.1-cp38-none-macosx_10_9_x86_64.whl (143.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 143.1/143.1 MB 6.9 MB/s eta 0:00:00
Collecting torch-scatter
  Downloading torch_scatter-2.1.1.tar.gz (107 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 107.6/107.6 kB 13.2 MB/s eta 0:00:00
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [6 lines of output]
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/private/var/folders/0q/jh77wwcx1cs989k89kv79mqr0000gn/T/pip-install-dh1d7ri4/torch-scatter_4f55b654d6284f4db2105882832dfe5a/setup.py", line 8, in <module>
          import torch
      ModuleNotFoundError: No module named 'torch'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```


ivy-seed:

vedpatwardhan (commented):
Hey @fabriziocacicia, I was just wondering which approach would be preferable, adding a new `.txt` file or downloading from their [wheel](https://github.com/pyg-team/pytorch_geometric/issues/861#issuecomment-1108349211) before installing `torch` which would get it to work in the same file. 
Happy to know your thoughts, thanks @fabriziocacicia 🙂

fabriziocacicia:
> 

Let me try with wheels to see how it looks like

fabriziocacicia:
> Hey @fabriziocacicia, I was just wondering which approach would be preferable, adding a new `.txt` file or downloading from their [wheel](https://github.com/pyg-team/pytorch_geometric/issues/861#issuecomment-1108349211) before installing `torch` which would get it to work in the same file. Happy to know your thoughts, thanks @fabriziocacicia 🙂

Let me try with wheels to see how it looks like

fabriziocacicia:
@vedpatwardhan if you meant to update the `optional.txt` file like this
```
...
-f https://download.pytorch.org/whl/cpu/torch_stable.html
torch # unpinned, we test for latest version now
torch-scatter
...
```
it still fails, I just tried.

But from the console output, I think the line with `-f` is not picked.


```shell
Collecting tensorflow-probability (from -r requirements/optional.txt (line 14))
  Downloading tensorflow_probability-0.20.1-py2.py3-none-any.whl (6.9 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 6.9/6.9 MB 8.8 MB/s eta 0:00:00
Collecting torch (from -r requirements/optional.txt (line 15))
  Downloading torch-2.0.1-cp310-none-macosx_10_9_x86_64.whl (143.4 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 143.4/143.4 MB 7.6 MB/s eta 0:00:00
Collecting torch-scatter (from -r requirements/optional.txt (line 16))
  Downloading torch_scatter-2.1.1.tar.gz (107 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 107.6/107.6 kB 15.3 MB/s eta 0:00:00
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [6 lines of output]
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/private/var/folders/0q/jh77wwcx1cs989k89kv79mqr0000gn/T/pip-install-eukh0nis/torch-scatter_73902a3dc06a4964bcccb6dbabe6e49c/setup.py", line 8, in <module>
          import torch
      ModuleNotFoundError: No module named 'torch'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```

vedpatwardhan:
> @vedpatwardhan if you meant to update the `optional.txt` file like this
> 
> ```
> ...
> -f https://download.pytorch.org/whl/cpu/torch_stable.html
> torch # unpinned, we test for latest version now
> torch-scatter
> ...
> ```
> 
> it still fails, I just tried.
> 
> But from the console output, I think the line with `-f` is not picked.
> 
> ```shell
> Collecting tensorflow-probability (from -r requirements/optional.txt (line 14))
>   Downloading tensorflow_probability-0.20.1-py2.py3-none-any.whl (6.9 MB)
>      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 6.9/6.9 MB 8.8 MB/s eta 0:00:00
> Collecting torch (from -r requirements/optional.txt (line 15))
>   Downloading torch-2.0.1-cp310-none-macosx_10_9_x86_64.whl (143.4 MB)
>      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 143.4/143.4 MB 7.6 MB/s eta 0:00:00
> Collecting torch-scatter (from -r requirements/optional.txt (line 16))
>   Downloading torch_scatter-2.1.1.tar.gz (107 kB)
>      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 107.6/107.6 kB 15.3 MB/s eta 0:00:00
>   Preparing metadata (setup.py) ... error
>   error: subprocess-exited-with-error
>   
>   × python setup.py egg_info did not run successfully.
>   │ exit code: 1
>   ╰─> [6 lines of output]
>       Traceback (most recent call last):
>         File "<string>", line 2, in <module>
>         File "<pip-setuptools-caller>", line 34, in <module>
>         File "/private/var/folders/0q/jh77wwcx1cs989k89kv79mqr0000gn/T/pip-install-eukh0nis/torch-scatter_73902a3dc06a4964bcccb6dbabe6e49c/setup.py", line 8, in <module>
>           import torch
>       ModuleNotFoundError: No module named 'torch'
>       [end of output]
>   
>   note: This error originates from a subprocess, and is likely not a problem with pip.
> error: metadata-generation-failed
> 
> × Encountered error while generating package metadata.
> ╰─> See above for output.
> 
> note: This is an issue with the package mentioned above, not pip.
> hint: See above for details.
> ```

Yeah it seems like `-f` is not being picked, I was just thinking of something similar to `pip install torch torch-scatter -f https://data.pyg.org/whl/torch-2.0.1+cpu.html` where we don't need to specify the version eventually and it would install both libraries in a single command, coz it would be a bit overkill to have 2 files just for being able to install one library. The `m1_1.txt` and `m1_2.txt` was also a temporary fix at the time, but I think there's a possibility of getting this done in a single file. 
Thanks @fabriziocacicia, let me know if you any assistance 🙂

fabriziocacicia:
@vedpatwardhan Do you think we have to stick with a `pip install` command. How do you see to use the shell?

If we can tell the user to launch a shell script to install the optional dependencies I am thinking to something like:

```shell
while read -r dep; do
  IFS_TEMP=$IFS
  IFS='#'
  read -a strarr <<< "$dep"
  dep="${strarr[0]}"
  pip install "$dep"
  IFS=IFS_TEMP
done < ./requirements/optional.txt
```

This way we can keep a single `optional.txt` file and still avoiding the issue with `torch` and `torch-scatter` or any other one that could appear in the future

vedpatwardhan:
> @vedpatwardhan Do you think we have to stick with a `pip install` command. How do you see to use the shell?
> 
> If we can tell the user to launch a shell script to install the optional dependencies I am thinking to something like:
> 
> ```shell
> while read -r dep; do
>   IFS_TEMP=$IFS
>   IFS='#'
>   read -a strarr <<< "$dep"
>   dep="${strarr[0]}"
>   pip install "$dep"
>   IFS=IFS_TEMP
> done < ./requirements/optional.txt
> ```
> 
> This way we can keep a single `optional.txt` file and still avoiding the issue with `torch` and `torch-scatter` or any other one that could appear in the future

I think the `txt` file provides more control to the user than a script, so I prefer it that way. But if we can't get it done through a single file eventually (and the `-f` way doesn't work), we should probably go forward with the 2 file option 😅
(I know that's what you were doing in the first place, but given that the issue I linked had a positive response to the `-f` way, there's a chance we can get it to work too. But no worries if not!)

fabriziocacicia:
@vedpatwardhan I made it works writing this:

```
-f https://download.pytorch.org/whl/cpu/torch_stable.html
-f https://data.pyg.org/whl/torch-2.0.1%2Bcpu.html
torch==2.0.1 # unpinned, we test for latest version now
torch-scatter==2.1.1 # unpinned, mod_name=torch_scatter
```
I think that the only way to make it work is to specify explicitly the versions of both torch and torch-scatter. And obviously these have to be maintained and updated regularly.

vedpatwardhan:
> @vedpatwardhan I made it works writing this:
> 
> ```
> -f https://download.pytorch.org/whl/cpu/torch_stable.html
> -f https://data.pyg.org/whl/torch-2.0.1%2Bcpu.html
> torch==2.0.1 # unpinned, we test for latest version now
> torch-scatter==2.1.1 # unpinned, mod_name=torch_scatter
> ```
> 
> I think that the only way to make it work is to specify explicitly the versions of both torch and torch-scatter. And obviously these have to be maintained and updated regularly.

Ah that's unfortunate! Although I think you'd only need to specify the version for the `-f https://data.pyg.org/whl/torch-2.0.1%2Bcpu.html` and not for the last 2 lines, we shouldn't be having versions at all in `optional.txt`. Let's just revert back to the 2 file approach in that case. Thanks @fabriziocacicia 🙂

fabriziocacicia:
> Ah that's unfortunate! Although I think you'd only need to specify the version for the `-f https://data.pyg.org/whl/torch-2.0.1%2Bcpu.html` and not for the last 2 lines, we shouldn't be having versions at all in `optional.txt`. Let's just revert back to the 2 file approach in that case. Thanks @fabriziocacicia 🙂

@vedpatwardhan do you think that neither an automated solution that keeps all the dependencies version always pointing to the latest version isn't acceptable?
Do you think that we must stick to leaving the unpinned instead of pinned to the latest?

vedpatwardhan:
> > Ah that's unfortunate! Although I think you'd only need to specify the version for the `-f https://data.pyg.org/whl/torch-2.0.1%2Bcpu.html` and not for the last 2 lines, we shouldn't be having versions at all in `optional.txt`. Let's just revert back to the 2 file approach in that case. Thanks @fabriziocacicia 🙂
> 
> @vedpatwardhan do you think that neither an automated solution that keeps all the dependencies version always pointing to the latest version isn't acceptable? Do you think that we must stick to leaving the unpinned instead of pinned to the latest?

Yeah that's kinda what I'm not completely sure about, because just for the sake of unpinning, we might end up with an extra file for a single library which would also not be good ig. What do you think @fabriziocacicia ? Thanks 🙂

fabriziocacicia:
@vedpatwardhan What do you think about using a tool like [pur](https://github.com/alanhamlett/pip-update-requirements) to keep the versions pinned to the latest one?

vedpatwardhan:
> @vedpatwardhan What do you think about using a tool like [pur](https://github.com/alanhamlett/pip-update-requirements) to keep the versions pinned to the latest one?

Yeah that's a good point actually, we could just run an automated workflow everyday which would update the versions in the `requirements` files if a new version was found. And then we could actually also help us track when a new version was released through a commit on our repo, which would be even faster than realizing it after a test failure. Let's discuss this again in a couple of days. Thanks @fabriziocacicia 🙂

fabriziocacicia:
I'm ok with the 2-file solution, I will update the Setting Up docs page in this PR too and will merge it after.
Regarding the usage of a single file in combination with `pur`, why do you think is not feasible? Do you want to avoid the usage of pinned versions @vedpatwardhan ?

vedpatwardhan:
Actually, I don't think `pur` would be able to correctly update the `-f` line in `optional.txt` even if we end up going by the single-file approach 😅

fabriziocacicia:
> Actually, I don't think `pur` would be able to correctly update the `-f` line in `optional.txt` even if we end up going by the single-file approach 😅 So even with `pur`, we'd end up having 2 files

Of course `pur` wouldn't automatically update the line with the `-f`, but we can easily update it inside the same automated workflow which would call `pur` itself.

I played a bit with `pur` locally and, for example, the following python line:
```python
updates =  update_requirements(input_file='requirements/requirements.txt', dry_run=True)
```
returns a dict.
The `updates` dict has one item per dependency.
Screenshot 2023-07-11 at 11 36 11
Then each item contains some useful informations:
Screenshot 2023-07-11 at 11 37 03
And with this we could easily run the following check
```python
    if updates['psutil'][0]['updated']:
        print(updates['psutil'][0]['latest'].public)
```
the above code prints 
```
5.9.5
```

So, inside the automated workflow we could make `pur` automatically update the versions for the pinned dependencies, and with a script we could extract the version number of the dependencies we are interested in and use it to update the line with `-f`.

P.S. the `update_requirements` method from `pur` has the `dry_run` flag to check for updates without updating and it also has the `only` argument to specify a list of dependencies to check for, like:
```python
update_requirements(input_file='requirements/optional.txt', dry_run=True, only=['torch'])
```
This will only check for `torch` updates and we could extract its version number with the snippet above.

Any thoughts on this approach @vedpatwardhan ?

vedpatwardhan:
> > Actually, I don't think `pur` would be able to correctly update the `-f` line in `optional.txt` even if we end up going by the single-file approach 😅 So even with `pur`, we'd end up having 2 files
> 
> Of course `pur` wouldn't automatically update the line with the `-f`, but we can easily update it inside the same automated workflow which would call `pur` itself.
> 
> I played a bit with `pur` locally and, for example, the following python line:
> 
> ```python
> updates =  update_requirements(input_file='requirements/requirements.txt', dry_run=True)
> ```
> 
> returns a dict. The `updates` dict has one item per dependency. Screenshot 2023-07-11 at 11 36 11 Then each item contains some useful informations: Screenshot 2023-07-11 at 11 37 03 And with this we could easily run the following check
> 
> ```python
>     if updates['psutil'][0]['updated']:
>         print(updates['psutil'][0]['latest'].public)
> ```
> 
> the above code prints
> 
> ```
> 5.9.5
> ```
> 
> So, inside the automated workflow we could make `pur` automatically update the versions for the pinned dependencies, and with a script we could extract the version number of the dependencies we are interested in and use it to update the line with `-f`.
> 
> P.S. the `update_requirements` method from `pur` has the `dry_run` flag to check for updates without updating and it also has the `only` argument to specify a list of dependencies to check for, like:
> 
> ```python
> update_requirements(input_file='requirements/optional.txt', dry_run=True, only=['torch'])
> ```
> 
> This will only checks for `torch` updates and we could extract its version number with the snippet above.
> 
> Any thoughts on this approach @vedpatwardhan ?

This sounds great actually! And of course, `pur` would only be required in the CI workflow that performs the update which is even better! 
Could you please implement a minimal version of how this would look like? Also, given that this falls in the realm of CI, I'll request @RashulChutani's review. Thanks @fabriziocacicia 🙂

fabriziocacicia:
@vedpatwardhan I have working script and workflow that do what we need.

You can look at the changes to see what the script looks like.
And you also can see from the commits that the workflow actually works. I manually pinned `torch` and `torch-scatter` to previous versions and the script successfully update both the version and the lines with '-f'

fabriziocacicia:
I ran everything on a branch on my fork obviously

fabriziocacicia:
Just a small issue on the `torch-scatter` line. I'm going to fix it shortly

fabriziocacicia:
Actually is not an issue in the script updated itself. it is more a different behaviour of the result in executing `pur`.

Starting from
```
torch==2.0.0
torch-scatter==2.1.0
```
in my local machine (a Macbook) executing the script has this result:
```
torch==2.0.1
torch-scatter==2.1.1
```

Instead, running it inside the container of the github actions resulted in what it can be seen in the commit.

I have to investigate this behaviour

vedpatwardhan:
Awesome! Also, is there a way we could prevent an update from happening for a particular library as well? For instance, if a library is released and it breaks everything, the `pur` workflow updates the `optional.txt` file to the new version. Then realizing that there's a major problem which would need a decent amount of time to fix, we temporarily pin it back to the old version, ig the `pur` workflow would update the version the next time it's triggered again right? 
Thanks for diving into this @fabriziocacicia, great to see that we're reaching to a solution on this front 😄

fabriziocacicia:
> Awesome! Also, is there a way we could prevent an update from happening for a particular library as well? For instance, if a library is released and it breaks everything, the `pur` workflow updates the `optional.txt` file to the new version. Then realizing that there's a major problem which would need a decent amount of time to fix, we temporarily pin it back to the old version, ig the `pur` workflow would update the version the next time it's triggered again right? 
> 
Of course @vedpatwardhan , `pur` has the possibility to exclude certain packages from updating.

RickSanchezStoic:
Hey @fabriziocacicia running image creation on the latest changes (after the bot bug fix in JSON file), the image creation fails because of the same torch and torch-scatter issue as before. As we discussed before, when it comes to docker, we have to take care of the path as well as mappings that the changes to the optional.tx won't alone do.

You can revert the changes to the Dockerfile as well as mappings, as it relies on the optional.txt only for other requirements and not for frameworks at all. Once that is done, we can then see if the changes to the optional.txt conflict with docker.

fabriziocacicia:
@RickSanchezStoic It's strange, I had tested the creation of the codespaces several times. 

Anyway, I reverted both the changes.

ivy-seed:

RickSanchezStoic:
Hey @fabriziocacicia just tested out the cpu and gpu images and everything looks good sofar. Feel free to merge this.

RickSanchezStoic (approved):

vedpatwardhan:
I'd say in that case, the only thing remaining is to get the workflow you've added scheduled to run once a day, and also making the commit message by the bot more specific about what library was updated. Thanks @fabriziocacicia @RickSanchezStoic 😄

fabriziocacicia:
@vedpatwardhan I think the workflow may need some changes. Both commit message and commit user may need an update.
Also I don't know if this is the right choice of steps for this task.
Maybe a look by @RashulChutani would be better.

ivy-seed:

ivy-seed:

vedpatwardhan:
@RashulChutani could you please review this when you get back? This PR has been waiting for a review since 11th August 😅
Thanks!

vedpatwardhan:
> Looks good to me. Do we want the workflow to run repeatedly? If yes, we can add a schedule section in the workflow to run it like a corn job.

Exactly, so we'd have a cron job that would check if any of the libraries in the `requirements.txt` or `optional.txt` files had a new release and then update those files. Would be great if you could update the workflow accordingly @RashulChutani and if you could test things out @RickSanchezStoic to get this over the line. Thanks both 😄

RickSanchezStoic:
as of now building this dockerfile is failing, can you update your repo once to see if the issue gets resolved

PR #16703, opened 2023-06-14
add 'axis' kwarg to torch.mean frontend

Sam-Armstrong:
Despite the fact it is not mentioned in the torch docs, torch.mean can take the kwarg 'axis' and use it in the same way as 'dim'. e.g. torch.mean(x, axis=-1). I've added the capability to handle this kwarg, so it doesn't throw `TypeError: mean() got an unexpected keyword argument 'axis'`.

(this problem was identified by attempting to transpile ImageGPTModel - https://github.com/huggingface/transformers/blob/v4.30.0/src/transformers/models/imagegpt/modeling_imagegpt.py#L171)

### DIFF
--- a/ivy/functional/frontends/torch/reduction_ops.py
+++ b/ivy/functional/frontends/torch/reduction_ops.py
@@ -53,7 +53,9 @@ def sum(input, dim=None, keepdim=False, *, dtype=None, out=None):
 
 
 @to_ivy_arrays_and_back
-def mean(input, dim=None, keepdim=False, *, out=None):
+def mean(input, dim=None, axis=None, keepdim=False, *, out=None):
+    if dim is None:
+        dim = axis
     return ivy.mean(input, axis=dim, keepdims=keepdim, out=out)
 
 


ivy/functional/frontends/torch/reduction_ops.py
```diff
@@ -53,7 +53,9 @@ def sum(input, dim=None, keepdim=False, *, dtype=None, out=None):
 
 
 @to_ivy_arrays_and_back
-def mean(input, dim=None, keepdim=False, *, out=None):
+def mean(input, dim=None, axis=None, keepdim=False, *, out=None):
+    if dim is None:
+        dim = axis
     return ivy.mean(input, axis=dim, keepdims=keepdim, out=out)
```
AnnaTz, line 59: What happens if both `dim` and `axis` are given? Is there an assertion in the original torch function?

AnnaTz (changes requested):

AnnaTz:
> I'm not sure if this is the case for any torch functions other than mean, maybe this is something to look into

I think it would be too time-consuming to manually check every function to find other ones that have the same behavior. I guess we'll just keep this in mind and fix specific cases as they come up during transpilation.

ivy/functional/frontends/torch/reduction_ops.py
Sam-Armstrong: @AnnaTz I *think* that torch does this with function overloading in C++. If you try to call `torch.mean(x, dim=-1, axis=-1)` then that doesn't work in native torch. So I think this change would be fine.

AnnaTz (approved):
Looks good! Feel free to merge 🙂

PR #16718, opened 2023-06-14
fix reduce_window

Sam-Armstrong:
It looks like there are some problems with the window_strides and base_dilation args in reduce_window. The minimal example below fails, and I am also getting problems transpiling models which call reduce_window. I've come up with a quick solution here which fixes both of these cases, but I'm sure there is a better way of doing it - so let me know what you think.

Min example:
```
import ivy

if __name__ == '__main__':
    ivy.set_backend('torch')
    x = ivy.random_normal(shape=(4, 4))
    ivy.reduce_window(x, 0, ivy.add, (2, 2))
```

### DIFF
--- a/ivy/functional/ivy/experimental/layers.py
+++ b/ivy/functional/ivy/experimental/layers.py
@@ -2146,7 +2146,7 @@ def reduce_window(
     *,
     window_strides: Union[int, Sequence[int]] = 1,
     padding: Union[str, int, Sequence[Tuple[int, int]]] = "VALID",
-    base_dilation: Union[int, Sequence[int]] = 1,
+    base_dilation: Union[int, Sequence[int]] = None,
     window_dilation: Union[int, Sequence[int]] = 1,
 ) -> ivy.Array:
     """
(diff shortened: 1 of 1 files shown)


AnnaTz:
You said there are two cases that are failing and are both fixed by your changes, but you only provide one example. What is the second case?

ivy/functional/ivy/experimental/layers.py
```diff
@@ -2189,6 +2189,8 @@ def reduce_window(
     # ToDo: add support for window_dilation
     computation = _correct_ivy_callable(computation)
     op, dims, strides = operand, window_dimensions, window_strides
+    if isinstance(strides, int):
+        strides = tuple([strides] * len(dims))
     init_value = _cast_init(init_value, op.dtype)
```
AnnaTz, line 2194: Actually the same should be done for all the arguments: `window_dimensions`, `window_strides`, `padding`, `base_dilation` and `window_dilation`, since they could all be a single integer. 
And instead of using `len(dims)` I would suggest using `len(op.shape)` in case `dims` is also an integer.

AnnaTz (changes requested):

Sam-Armstrong:
@AnnaTz I've made those changes. The second example I was referring to was transpiling the huggingface vision model 'ConditionalDetrModel' (or 'FocalNetModel') to jax. Can send over my script I'm using if you wanted to have a look.

Sam-Armstrong:
@AnnaTz good suggestion, I've done that. Just checking that my implementation is correct, because I haven't used ivy.map before 😅

PR #16730, opened 2023-06-14
Fix bfloat16 placement issue in tensorflow abs

Dsantra92:


hirwa-nshuti (changes requested):
Not sure if the CI run being broken is being caused by the commits in this PR😁
Try to update this branch with master so the CI should run perfectly🙂
Let's see what @VedPatwardhan thinks of this

ivy/functional/backends/tensorflow/searching.py
```diff
@@ -111,7 +111,11 @@ def where(
     out: Optional[Union[tf.Tensor, tf.Variable]] = None,
 ) -> Union[tf.Tensor, tf.Variable]:
     x1, x2 = ivy.promote_types_of_inputs(x1, x2)
-    return tf.cast(tf.experimental.numpy.where(condition, x1, x2), x1.dtype)
+    dev = ivy.dev(x1, as_native=False)
+    ret = tf.cast(tf.experimental.numpy.where(condition, x1, x2), x1.dtype)
+    with tf.device(dev):
```
vedpatwardhan, line 116: Would we need to have this kind of logic for all functions? Probably cleaner to just use the device-shifting decorator in [this](https://github.com/unifyai/ivy/pull/16101) PR

vedpatwardhan (commented):
Just a doubt out of curiosity, apart from that I'll just request @sherry30's review as it's device-related. Thanks @Dsantra92 and @hirwa-nshuti 🙂

ivy/functional/backends/tensorflow/searching.py
Dsantra92: Yes, definitely a better way of handling it. Will wait for it to merge.

ivy-seed:

sherry30 (commented):
After the merge of `handle_device_shifting` decorator, this seems to work fine because the function gets called in the `tf.device` context manager now.

PR #16731, opened 2023-06-14
add missing @to_ivy_arrays_and_back to torch frontend activations

Sam-Armstrong:
Identified while transpiling MobileNetV2Model from torch to jax, which uses hardtanh and failed due to there not being a @to_ivy_arrays_and_back decorator (throwing the error `less requires ndarray or scalar arguments, got <class 'ivy.functional.frontends.torch.tensor.Tensor'> at position 0`). Noticed that some of the other functions here also do not have the decorator, so I've added those as well.


mattbarrett98 (commented):
looks good

Issue #16357, opened 2023-06-07
numpy  rfft2


rajveer43:
#1532

AnnaTz:
> @AnnaTz
> 
> need some clarification,
> 
> I was merging the updates of master branch of unifyivy in to my working branch but there where conflicts that It was showing that if you merge the branch then your all commits within your working branch will be deleted and I selected that option. now the PR is closed automatically due to that.. what should I do? create a new PR for my task?

I think you could have just reopened the PR, but opening a new one is okay as well.

PR #16739, opened 2023-06-15
Numpy rfft2

rajveer43:
- [x] #16357

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_numpy/test_fft/test_discrete_fourier_transform.py
+++ b/ivy_tests/test_ivy/test_frontends/test_numpy/test_fft/test_discrete_fourier_transform.py
@@ -194,6 +194,63 @@ def test_numpy_rfftfreq(n, sample_rate, frontend, test_flags, fn_tree, on_device
 @handle_frontend_test(
     fn_tree="numpy.fft.rfft2",
     dtype_input_axis=helpers.dtype_values_axis(
+        available_dtypes=helpers.get_dtypes("valid"),
+        min_num_dims=2,
+        max_num_dims=2,
+        force_tuple_axis=True,
+    ),
+    norm=st.sampled_from(["backward", "ortho", "forward"]),
+    s=st.lists(st.integers(), min_size=2, max_size=2),
+)
+def test_numpy_rfft2(
+    dtype_input_axis, norm, s, frontend, test_flags, fn_tree, on_device
+):
+    input_dtype, x, axis = dtype_input_axis
+    helpers.test_frontend_function(
+        input_dtypes=input_dtype,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        test_values=True,
+        x=x,
+        s=s,
+        axis=axis,
+        norm=norm,
+    )
+
+# @handle_frontend_test(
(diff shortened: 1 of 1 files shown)


ivy_tests/test_ivy/test_frontends/test_numpy/test_fft/test_discrete_fourier_transform.py
```diff
@@ -189,3 +189,33 @@ def test_numpy_rfftfreq(n, sample_rate, frontend, test_flags, fn_tree, on_device
...
     )
+
+
+@handle_frontend_test(
+    fn_tree="numpy.fft.rfft2",
+    dtype_input_axis=helpers.dtype_values_axis(
+        available_dtypes=helpers.get_dtypes("float_and_complex"),
+        shape=(2,2),
+        min_num_dims=0,
+        max_num_dims=2,
```
AnnaTz, line 200: Since the array has to be 2D you will need to set `min_num_dims=2` as well. And no need to hardcode the `shape`. The `shape` could be of any 2D size, not just `(2,2)`.

ivy_tests/test_ivy/test_frontends/test_numpy/test_fft/test_discrete_fourier_transform.py
```diff
@@ -189,3 +189,33 @@ def test_numpy_rfftfreq(n, sample_rate, frontend, test_flags, fn_tree, on_device
         n=n,
         d=d,
     )
+
+
+@handle_frontend_test(
+    fn_tree="numpy.fft.rfft2",
+    dtype_input_axis=helpers.dtype_values_axis(
+        available_dtypes=helpers.get_dtypes("float_and_complex"),
```
AnnaTz, line 197: You should use `helpers.get_dtypes("valid")` here and update the supported dtypes decorator of the function if it's only supposed to work for complex numbers.

AnnaTz (changes requested):

ivy/functional/frontends/numpy/fft/discrete_fourier_transform.py
```diff
@@ -125,3 +124,16 @@ def rfftfreq(n, d=1.0):
     N = n // 2 + 1
     results = ivy.arange(0, N, dtype=int)
     return results * val
+
+
+@to_ivy_arrays_and_back
+@with_supported_dtypes({"1.24.3 and above": ("float16",)}, "numpy")
```
AnnaTz, line 130: Have you tried running this locally? I don't think 'float16` should be the only supported dtype.

ivy/functional/ivy/random.py
```diff
@@ -46,7 +46,7 @@ def _check_bounds_and_get_shape(low, high, shape):
         return ivy.shape(low)
     if isinstance(high, valid_types):
         return ivy.shape(high)
-    return ivy.Shape(())
+    return ()
```
AnnaTz, line 49: Also, in order to merge the PR you'll have to remove changes in files that are not related to the issue you are solving, i.e.  only `discrete_fourier_transform.py` and `test_discrete_fourier_transform.py` should be affected.

AnnaTz (changes requested):

AnnaTz:
Changes in the idea and configuration files are hard to get rid of. I would suggest you make a new branch and open a new PR making sure you don't commit those this time.

Issue #15050, opened 2023-05-08
Add `paddle.nn.functional.activation` NN Activation Functions to PaddlePaddle Frontend


hmahmood24:
Add [NN Activation Functions](https://www.paddlepaddle.org.cn/documentation/docs/en/api/index_en.html) to PaddlePaddle Frontend

> Please keep in mind that the proper way to link an issue to this list is to comment "- [ ] #issue_number" while the issue's title only includes the name of the function you've chosen.

- [ ] celu
- [ ] elu
- [ ] elu_
- [ ] gelu
- [ ] hardshrink
- [ ] hardtanh
- [ ] hardsigmoid
- [ ] hardswish
- [ ] leaky_relu
- [ ] log_sigmoid
- [ ] maxout
- [ ] prelu
- [ ] relu
- [ ] relu_
- [ ] relu6
- [ ] selu
- [ ] sigmoid
- [ ] silu
- [ ] softmax
- [ ] softmax_
- [ ] softplus
- [ ] softshrink
- [ ] softsign
- [ ] swish
- [ ] mish
- [ ] #15042
- [ ] tanh_
- [ ] tanhshrink
- [ ] thresholded_relu
- [ ] log_softmax
- [ ] glu
- [ ] gumbel_softmax
- [ ] rrelu

sarakarati:
- [ ] #15565

KSANTHOSH200:
- [ ] #15794

maher-mohsen:
"- [ softmax] #15050

KSANTHOSH200:
- [ ] #15893

KSANTHOSH200:
- [ ] #15894

sarakarati:
Softplus 15896

sarakarati:
"-softplus #15896"

sarakarati:
"-[softplus] #15896"

sarakarati:
softplus #15894

KSANTHOSH200:
- [ ] #15897

KSANTHOSH200:
- [ ] #15899

KSANTHOSH200:
- [ ] #15893

KSANTHOSH200:
- [ ] #15900

KSANTHOSH200:
- [ ] #15901

KSANTHOSH200:
- [ ] #15893

maher-mohsen:
![log_softmax] #15909

maher-mohsen:
#15916

[...]

PR #17031, opened 2023-06-16
Add activations to paddle.nn.functional.activation

HaiderSultanArc:


HaiderSultanArc:
Added elu, leaky_relu, relu, sigmoid, silu, softmax, softplus, swish, mish, tanhshrink and thresholded_relu activation functions to paddle.nn.functional.activation
Issue: #15050

PR #17067, opened 2023-06-17
Add jax.lax.batch_matmul with tests

### LINKED ISSUE #17065: batch_matmul

fr30:
Close #17065

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_operators.py
+++ b/ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_operators.py
@@ -1745,17 +1745,16 @@ def test_jax_lax_dot(
 
 @st.composite
 def _get_dtype_inputs_for_batch_matmul(draw):
-    dim_size = draw(helpers.ints(min_value=1, max_value=4))
-    batch_size = draw(helpers.ints(min_value=1, max_value=3))
+    dim_size_left = draw(helpers.ints(min_value=1, max_value=4))
+    dim_size_right = draw(helpers.ints(min_value=1, max_value=4))
+    batch_size = draw(helpers.ints(min_value=0, max_value=3))
     dtype = draw(helpers.get_dtypes("numeric", index=1, full=False))
-    if dim_size == 1:
+
+    if batch_size == 0:
         lhs = draw(
             helpers.array_values(
                 dtype=dtype[0],
-                shape=(
-                    dim_size,
-                    dim_size,
-                ),
+                shape=(dim_size_left, dim_size_right),
                 min_value=2,
                 max_value=5,
             )
@@ -1763,10 +1762,7 @@ def _get_dtype_inputs_for_batch_matmul(draw):
         rhs = draw(
             helpers.array_values(
                 dtype=dtype[0],
-                shape=(
-                    dim_size,
-                    dim_size,
-                ),
+                shape=(dim_size_right, dim_size_left),
                 min_value=2,
                 max_value=5,
             )
@@ -1775,7 +1771,7 @@ def _get_dtype_inputs_for_batch_matmul(draw):
         lhs = draw(
             helpers.array_values(
                 dtype=dtype[0],
-                shape=(batch_size, dim_size, dim_size),
+                shape=(batch_size, dim_size_left, dim_size_right),
                 min_value=2,
                 max_value=5,
             )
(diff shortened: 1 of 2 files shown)


ivy-leaves:

Thank you :hugs:

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_operators.py
```diff
@@ -1743,6 +1743,82 @@ def test_jax_lax_dot(
...
+        rhs = draw(
+            helpers.array_values(
+                dtype=dtype[0],
+                shape=(batch_size, dim_size, dim_size),
+                min_value=2,
+                max_value=5,
+            )
+        )
+    return dtype, lhs, rhs
+
```
AnnaTz, line 1792: I think the way you are using `dim_size` makes it so `lhs` and `rhs` are always stacked square arrays. We should be testing for all possible shape cases.

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_operators.py
```diff
@@ -1743,6 +1743,82 @@ def test_jax_lax_dot(
     )
 
 
+@st.composite
+def _get_dtype_inputs_for_batch_matmul(draw):
+    dim_size = draw(helpers.ints(min_value=1, max_value=4))
+    batch_size = draw(helpers.ints(min_value=1, max_value=3))
+    dtype = draw(helpers.get_dtypes("numeric", index=1, full=False))
```
AnnaTz, line 1750: Also, please use all "valid" dtypes here instead of "numeric", and add `@with_supported_dtypes` to the function to specify that it only supports numeric inputs.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_operators.py
```diff
@@ -1775,19 +1771,20 @@ def _get_dtype_inputs_for_batch_matmul(draw):
...
         rhs = draw(
             helpers.array_values(
                 dtype=dtype[0],
-                shape=(batch_size, dim_size, dim_size),
+                shape=(batch_size, dim_size_right, dim_size_left),
                 min_value=2,
                 max_value=5,
             )
         )
+
```
AnnaTz, line 1787: The function actually works with shape of any size, not just 3D. 
Example:
```
import numpy as np
import jax.lax as lax

A = np.zeros((6, 2, 3, 5, 4))
B = np.zeros((6, 2, 3, 4, 5))

C = lax.batch_matmul(A, B)

print(C.shape)  # (6, 2, 3, 5, 5)
```
Basically, this function can be applied on any two matrices that have either the exact same shape or same shape with the last two dimension sizes swapped.
Instead of sampling the sizes first and passing them to `shape`, I would suggest you first sample `lhs` without specifying a `shape` and then using its randomly generated shape to create a shape for `rhs`. Hope this makes sense!

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_operators.py
```diff
@@ -1743,6 +1744,61 @@ def test_jax_lax_dot(
...
+    dtype = draw(helpers.get_dtypes("valid", index=1, full=False))
+    lhs = draw(
+        helpers.array_values(
+            dtype=dtype[0],
+            shape=lhs_shape,
+            min_value=2,
+            max_value=5,
+        )
+    )
+    rhs = draw(
```
AnnaTz, line 1762: You are overcomplicating things a bit. You can just use the `dtype_and_values` helper which does not require manually sampling a dtype and a shape beforehand. This way you could just do:
```
   dtype, lhs = draw(
        helpers.dtype_and_values(
            min_value=2,
            max_value=5,
        )
    )
   lhs_shape = lhs.shape
```

ivy_tests/test_ivy/test_frontends/test_jax/test_jax_lax_operators.py
```diff
@@ -1743,6 +1744,61 @@ def test_jax_lax_dot(
...
+):
+    input_dtypes, lhs, rhs = dtypes_and_xs
+    helpers.test_frontend_function(
+        input_dtypes=input_dtypes,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        rtol=1e-2,
+        atol=1e-2,
```
AnnaTz, line 1795: Does the test fail if you don't set these tolerance values?

AnnaTz (changes requested):

PR #17085, opened 2023-06-17
Multiversion devcontainer

RickSanchezStoic:
adds devcontainer support for multiversion docker image.

The requirement mappings json file is used to specify all the dependencies that each framework has, for certain dependencies we need to specify which version of this dependency will go with a given version of the framework, such as jaxlib. We do that by specifying the` {main framework:{dependency:{version of the main framework : version of the dependency} }}`that goes well with it.


.devcontainer/build_multiversion/devcontainer.json
```diff
@@ -0,0 +1,58 @@
...
+
+//update the fw arg to add as many versions to add
+{
+	"name": "Ivy Development Environment for Multiverison(build)",
+
+	"build": {
+		"dockerfile": "../../docker/DockerfileMultiversion",
+		"context": "../..",
+      "args": {
+    "fw": ["numpy/1.24.2  tensorflow/2.11.0 tensorflow/2.12.0 jax/0.4.10 jax/0.4.8"]
```
RickSanchezStoic, line 12: To add more versions of a framework just update this list

PR #17142, opened 2023-06-18
Added a script for running docstring tests locaally

hirwa-nshuti:
Hey @vedpatwardhan 

This PR adds the script for testing docstring locally of a given function and backends.


ivy-seed:

vedpatwardhan (changes requested):
Hey @hirwa-nshuti, I think it would be great if we just create a dummy function wrapped with `pytest.mark.parametrize` (or an equivalent) in `test_docstrings.py` which would then simply defer to `check_docstring_examples_run` for the input function instead of creating a new file to run docstring tests locally. It would also be more modular and be easily detected by Pytest Test Discovery and we wouldn't need a separate shell script.
I know there might be a few refactors to be made to `test_docstrings.py` (and how the `fn` argument is passed to the test function that runs the docstring tests for a single function), but probably that's a cleaner way to go about it.
Happy to know your thoughts, thanks @hirwa-nshuti 🙂

hirwa-nshuti:
> Hey @hirwa-nshuti, I think it would be great if we just create a dummy function wrapped with `pytest.mark.parametrize` (or an equivalent) in `test_docstrings.py` which would then simply defer to `check_docstring_examples_run` for the input function instead of creating a new file to run docstring tests locally. It would also be more modular and be easily detected by Pytest Test Discovery and we wouldn't need a separate shell script. I know there might be a few refactors to be made to `test_docstrings.py` (and how the `fn` argument is passed to the test function that runs the docstring tests for a single function), but probably that's a cleaner way to go about it. Happy to know your thoughts, thanks @hirwa-nshuti 🙂

Yeah I agree with this 😊, I can start working on that approach it won't be that much complicated😁

PR #17416, opened 2023-06-20
EHN: adding bitwise funtion in paddle tensor.py

### LINKED ISSUE #17414: bitwise_xor
Adding  bitwise_xor in paddle.tesnor.tesnor.Tesnor from this Todo list  #15115

MuhammadNizamani:
Adding bitwise_xor, Closes #17414


KareemMAX:
add_frontend_checklist

ivy/functional/frontends/paddle/tensor/tensor.py
```diff
@@ -180,3 +180,11 @@ def all(self, axis=None, keepdim=False, dtype=None, name=None):
     @with_unsupported_dtypes({"2.4.2 and below": ("float16", "bfloat16")}, "paddle")
     def sort(self, axis=-1, descending=False, name=None):
         return ivy.sort(self._ivy_array, axis=axis, descending=descending)
+
+    @with_supported_dtypes(
+        {"2.4.2 and below": ("bool", "uint8", "int8", "int16", "int32", "int64")},
+        "paddle",
+    )
+    def bitwise_xor(self, y, out=None, name=None):
+        y_ivy = _to_ivy_array(y)
```
KareemMAX, line 189: Is there is a specific reason why we use `_to_ivy_array`? I remember that we used `@to_ivy_arrays_and_back()` for this conversion but I can't find it in the rest of the file so I'm checking why it isn't there

KareemMAX (changes requested):
Great job @MuhammadNizamani 🚀🚀

I've made a small comment, maybe I might be mistaken about the `to_ivy_and_back` thing so I'm also asking if there is any internal changes that I'm not aware about. Please look at my comment I'd like to hear your thoughts

ivy/functional/frontends/paddle/tensor/tensor.py
MuhammadNizamani: @KareemMAX  we are implementing method in Tensor class here and we take our first input using `self ` and by using `self `  where can convert our input into ivy array so no need of `@to_ivy_arrays_and_back()`. 
I am using `_to_ivy_array`   because if i use `@to_ivy_arrays_and_back()` then it will became redundant with this `self._ivy_array`

KareemMAX: Yup, I got mixed up here

But I'm still curious that in other functions it's calling `paddle_frontend` like in `tensor.multiply()` https://github.com/unifyai/ivy/pull/17416/files#diff-4b9f6f2107068e5624a6f5d66bae5dae58c467a34f8c5aad65df3478ce4a5abbR187

KareemMAX (changes requested):

ivy/functional/frontends/paddle/tensor/tensor.py
MuhammadNizamani:
```
@with_unsupported_dtypes({"2.4.2 and below": ("float16", "bfloat16")}, "paddle")
    def multiply(self, y, name=None):
        return paddle_frontend.multiply(self, y)
```

The implementation appears to be incorrect based on my understanding. According to my understanding, `paddle.tensor.Tensor `should receive functions from the backend, rather than the frontend. However, the above function is using `paddle.tensor.tensor.math.py `to access the backend.

MuhammadNizamani:
@KareemMAX  sir, when you are going to merge this PR ?

ivy/functional/frontends/paddle/tensor/tensor.py
KareemMAX: `paddle_frontend` is the ivy implementation of paddle frontend. We are not calling any paddle functions if that is what do you mean.

I've asked @hmahmood24 and he said the following:
>  since instance methods operate on the instance of the frontend class (`paddle.Tensor` in this case) which have a `_ivy_array` attribute that holds the `ivy.Array` already) and there are only 2 ways we go about writing these instance methods:
>
> Calling `ivy.<func>` directly
> Calling a frontend function of the same functionality `paddle_frontend.<func>`

That's why I think this implementation is wrong and `multiply()` is have the correct implementation

KareemMAX:
@MuhammadNizamani once we fix the current issue your PR will be good to merge after running tests. Please refer to the latest comment I made on the unresolved conversation

MuhammadNizamani:
@KareemMAX the multiply method was not implemented by me so what should I do should I change it? And I have run local tests on the bitwise_xor  function it all passes.

ivy/functional/frontends/paddle/tensor/tensor.py
MuhammadNizamani: @KareemMAX  done

KareemMAX:
LGTM! merging now

Thanks for your work @MuhammadNizamani 🚀🚀🚀

KareemMAX (approved):

PR #17448, opened 2023-06-20
multi_head_attention_forward torch frontend function

Sam-Armstrong:
A frontend implementation for torch.nn.functional.multi_head_attention_forward (which is the functional called by torch.nn.MultiHeadAttention). This is required for transpiling at least one of the HuggingFace models. 

This frontend implementation captures all the behaviour of the original torch function aside add_zero_attn, which I was having trouble with - so have left a todo comment. I've got a lot of other stuff to get done with high priority, so figured we are best to get this merged with most of the correct functionality, and I can revisit if we run into issues. 

Also the test is a bit complex, so apologies about that, but I believe it covers all the important interactions. And all cases are passing 🙂

Here are links to the torch sourcecode that this is loosely based off:
https://github.com/pytorch/pytorch/blob/main/torch/nn/functional.py
https://github.com/pytorch/pytorch/blob/3e42854caa6bfd4cd2694b7da0b13ddd9695ca58/torch/csrc/api/include/torch/nn/functional/activation.h


Sam-Armstrong:
Close https://github.com/unifyai/ivy/issues/17447

ivy-leaves:

Thank you :hugs:

ivy_tests/test_ivy/test_frontends/test_torch/test_non_linear_activation_functions.py
```diff
@@ -1078,3 +1079,245 @@ def test_torch_softplus(
...
+        key_padding_mask,
+        is_causal,
+    )
+
+
+# multi_head_attention_forward
+@handle_frontend_test(
+    fn_tree="torch.nn.functional.multi_head_attention_forward",
+    dtype_mha_args=mha_forward_args(
+        dtypes=helpers.get_dtypes("float"),
```
AnnaTz, line 1246: The best approach would be to set this to `helpers.get_dtypes("valid")` and add `@with_supported_dtypes` to the function.

ivy_tests/test_ivy/test_frontends/test_torch/test_non_linear_activation_functions.py
```diff
@@ -1078,3 +1079,245 @@ def test_torch_softplus(
...
+    ),
+    add_zero_attn=st.just(False),
+    dropout_p=st.just(0.),
+    training=st.just(False),
+    need_weights=st.booleans(),
```
AnnaTz, line 1251: Shouldn't this be tested for `training=True` as well? and also for more values of `dropout`?

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_non_linear_activation_functions.py
Sam-Armstrong: @AnnaTz wouldn't having a dropout other than 0 make the outcome non-deterministic, which would cause the tests to fail?

AnnaTz: Yeah but I think you should still test for different values and in case the dropout is not zero you can just check that the shape and dtype of the return is correct, like how is done [here](https://github.com/unifyai/ivy/blob/c92849be48c5db478218200667a100404ce43248/ivy_tests/test_ivy/test_functional/test_nn/test_layers.py#L148)

Sam-Armstrong: @AnnaTz ok that makes sense, I've made a change that should do that. Let me know what you think. Also I had to fix this bug in ivy.dropout to get this to work, so this will have to be merged first if you could take a quick look https://github.com/unifyai/ivy/pull/17516. Thanks 🙂

Sam-Armstrong:
@AnnaTz are we good to merge this now?

AnnaTz (approved):
If you ran the test locally and it passes, then feel free to merge!

PR #17516, opened 2023-06-21
fix ivy.dropout returning native arrays/tensors

Sam-Armstrong:
Fixes problem where ivy.dropout can return things like tf.EagerTensor instead of ivy.Array.

Min example:
```
import ivy

if __name__ == '__main__':
    ivy.set_backend("tensorflow")

    x = ivy.random_normal(shape=(5, 5, 5))
    x = ivy.dropout(x, 0.1)
    print(type(x))
    x = x.reshape((25, 5))
```

### DIFF
--- a/ivy/functional/ivy/layers.py
+++ b/ivy/functional/ivy/layers.py
@@ -9,6 +9,7 @@ from ivy.utils.backend import current_backend
 from ivy.func_wrapper import (
     handle_array_function,
     inputs_to_ivy_arrays,
+    outputs_to_ivy_arrays,
     to_native_arrays_and_back,
     inputs_to_native_shapes,
     handle_out_argument,
@@ -184,6 +185,7 @@ linear.mixed_backend_wrappers = {
 @handle_array_like_without_promotion
 @handle_out_argument
 @inputs_to_ivy_arrays
+@outputs_to_ivy_arrays
 @handle_array_function
 def dropout(
     x: Union[ivy.Array, ivy.NativeArray],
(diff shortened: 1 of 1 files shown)


ivy-leaves:

Thank you :hugs:

AnnaTz:
Why was it returning native arrays? It is a composition of ivy functions so this shouldn't be happening 🤔

Sam-Armstrong:
@AnnaTz I've honestly got no idea what the problem is here, all the inner ivy functions seem to be working fine and when I create an identical sample dropout function in a test script it also returns ivy arrays rather than native. So I guess there must be some other problem within ivy that's causing this, but I'm no expert on what it could be 😅. What do you think we should do here? This solution does solve the problem ^

AnnaTz:
> @AnnaTz I've honestly got no idea what the problem is here, all the inner ivy functions seem to be working fine and when I create an identical sample dropout function in a test script it also returns ivy arrays rather than native. So I guess there must be some other problem within ivy that's causing this, but I'm no expert on what it could be 😅. What do you think we should do here? This solution does solve the problem ^

I think I figured it out. `dropout` is a mixed function so it should have a`mixed_backend_wrappers` attribute to define its decorators as in [interpolate](https://github.com/unifyai/ivy/blob/3ad4ba6e4de67aec619449c0e3dbdfa42adbf25b/ivy/functional/ivy/experimental/layers.py#L1918) for example.

AnnaTz (approved):

PR #17556, opened 2023-06-22
Throwing error message for unsupported dtypes using ivy and not relying on the underlying backend

RickSanchezStoic:
1. This can be disabled by setting calling `ivy.throw_error_for_unsupported_dtypes(False)`


ivy/__init__.py
```diff
@@ -1063,6 +1063,7 @@ def current_sub_backends():
 upcast_dtypes = False
 crosscast_dtypes = False
 cast_dtypes = lambda: downcast_dtypes and upcast_dtypes and crosscast_dtypes
+throw_error_for_dtypes = True
```
vedpatwardhan, line 1066: let's get this working similar to any other global mode, with a setter, an unsetter and a stack. Also could you please rename it to like `dtype_error_mode` or something?

vedpatwardhan (changes requested):
Hey @RickSanchezStoic, just requested a minor change, I think it would be great to deal with casting modes in a similar manner if possible. Thanks 🙂

CatB1t (approved):
LGTM!

PR #17568, opened 2023-06-22
removed lambdas

VinciGit00:


VinciGit00:
now i removed all the lambdas

PR #17610, opened 2023-06-23
cond #17520

### LINKED ISSUE #17520: cond
cond  paddle functio

ndurumo254:
close #17520


ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
```diff
@@ -606,3 +606,55 @@ def test_paddle_matrix_power(
...
+                    lambda x: x[0] == x[1]
+                ),
+                allow_inf=False,
+                allow_nan=False,
+            )
+        )
+        if np.linalg.det(matrix[1][0]) != 0:
+            break
+
+    return matrix[0], matrix[1]
```
hirwa-nshuti, line 638: Why returning two matrices here as we only need one Tensor for the function

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
```diff
@@ -606,3 +606,55 @@ def test_paddle_matrix_power(
...
+)
+def test_paddle_cond(*, dtype_and_x, p, on_device, fn_tree, frontend, test_flags):
+    dtype, x = dtype_and_x
+    helpers.test_frontend_function(
+        input_dtypes=dtype,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        test_values=False,
```
hirwa-nshuti, line 655: We should ensure that we perfome the value testing to be sure that the function is unified accross the backends

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
```diff
@@ -606,3 +606,55 @@ def test_paddle_matrix_power(
...
+# cond
+@st.composite
+def _get_dtype_and_matrix_non_singular(draw):
+    [None, "fro", "nuc", np.inf, -np.inf, 1, -1, 2, -2]
+    while True:
+        matrix = draw(
+            helpers.dtype_and_values(
+                available_dtypes=(
+                    ivy.float64,
+                    ivy.double,
```
hirwa-nshuti, line 620: `ivy.double` is an alias of `float63` I think we will only need to specified exact supported dtypes for the function.
Refer to the Docs here https://www.paddlepaddle.org.cn/documentation/docs/en/2.5rc1/api/paddle/linalg/cond_en.html#cond

ivy/functional/frontends/paddle/tensor/linalg.py
```diff
@@ -144,3 +144,9 @@ def bmm(x, y, transpose_x=False, transpose_y=False, name=None):
 @to_ivy_arrays_and_back
 def matrix_power(x, n, name=None):
     return ivy.matrix_power(x, n)
+
+
+@with_unsupported_dtypes({"2.4.2 and below": ("float16", "bfloat16")}, "paddle")
```
hirwa-nshuti, line 149: We will specifie the supported dtypes as indicated in docs, refer to the link added in the comments below

hirwa-nshuti (changes requested):
Hey @ndurumo254 

Check the comments below for the changes needed😊
Thanks

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
ndurumo254: > Why returning two matrices here as we only need one Tensor for the function

I think we need to return the dtypes of the input matrix and the input matrix itself. we need to calculate the condition number of the matrix for the dtypes  therefore returning one matrix for the dtype and the other matrix for the matrix itself. What do you think about this?

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
ndurumo254: > We should ensure that we perfome the value testing to be sure that the function is unified accross the backends

i also think the helpers.test_frontends functions will compare the result of the paddle.tensor.linalg.cond to the expected result. the test will pass only if the compared results are matching,.. am however not sure what you mean by value testing, may be you can give me more insiht on that

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
hirwa-nshuti: Got it here, but you can rename one to be dtype as it is confusing seems like it's two matrices being generated

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
ndurumo254: > The thing is we can't compare values when the `test_values=False` so we need to remove it and check for disparity between backend to see if we can make this function stable across all the backends

From https://www.paddlepaddle.org.cn/documentation/docs/en/2.5rc1/api/paddle/linalg/cond_en.html#cond I understand that paddle.tensor.linalg.cond will return a scalar value. when I set the test_values = True, (default value) the value will return more than one element. actually, in such a situation only the test will pass, I however tend to think this happens since the value of the tensor is not close to the condition number of the input matrix. i would to get your view on this.

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
hirwa-nshuti: Also, it would be better to not explicitly call ivy when testing frontend functions🙂

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
hirwa-nshuti: When testing the frontend function, as we are comparing the function output across all the backends with the target framework, we must ensure that the frontend function returns the same value as the framework and as it is indicated in the framework docs. This should be achieved by performing the value test, that's why I said we can't make it False in the case of this function, with some amount of appreciable tolerance value.

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
ndurumo254: > Also, it would be better to not explicitly call ivy when testing frontend functio

This makes sense

ivy/functional/frontends/paddle/tensor/linalg.py
ndurumo254: > We will specifie the supported dtypes as indicated in docs, refer to the link added in the comments below

Yeah sure, i didn't take keen interest on this

ivy/functional/frontends/paddle/tensor/linalg.py
```diff
@@ -144,3 +144,9 @@ def bmm(x, y, transpose_x=False, transpose_y=False, name=None):
 @to_ivy_arrays_and_back
 def matrix_power(x, n, name=None):
     return ivy.matrix_power(x, n)
+
+
+@with_unsupported_dtypes({"2.4.2 and below": ("float32", "float64")}, "paddle")
```
hirwa-nshuti, line 149: This should be specified as supported 😅
Using `with_supported_dtypes` decorator

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
```diff
@@ -606,3 +606,52 @@ def test_paddle_matrix_power(
...
+
+@handle_frontend_test(
+    fn_tree="paddle.tensor.linalg.cond",
+    dtype_and_x=_get_dtype_and_matrix_non_singular(dtypes=["float32", "float64"]),
+    p=st.sampled_from([None, "fro", "nuc", np.inf, -np.inf, 1, -1, 2, -2]),
+    test_with_out=st.just(False),
+)
+def test_paddle_cond(*, dtype_and_x, p, on_device, fn_tree, frontend, test_flags):
+    dtype, x = dtype_and_x
+
```
hirwa-nshuti, line 645: There is a helper function for checking whether the matrix is stable or not you can use it here to test on stable matrices
https://github.com/unifyai/ivy/blob/2a5c36b062d83b0b7dac86d73949646a5f8f1396/ivy_tests/test_ivy/helpers/hypothesis_helpers/general_helpers.py#L12

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
```diff
@@ -606,3 +606,52 @@ def test_paddle_matrix_power(
...
+    helpers.test_frontend_function(
+        input_dtypes=dtype,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        test_values=True,
+        x=x[0],
+        rtol=0,
```
hirwa-nshuti, line 654: This tol is too strict try using `1e-1`

hirwa-nshuti (changes requested):
Hey @ndurumo254 

Just a few changes to go😀
Check the comments

ivy/functional/frontends/paddle/tensor/linalg.py
ndurumo254: > This should be specified as supported 😅 Using `with_supported_dtypes` decorator

i dont know why am over looking this line

hirwa-nshuti: This decorator is for filtering unsupported dtypes of that function and from the docs this function supports only `float32` and `float64`

ndurumo254:
Yeah, I have seen, i hope it is solved now

ivy/functional/frontends/paddle/tensor/linalg.py
```diff
@@ -144,3 +144,10 @@ def bmm(x, y, transpose_x=False, transpose_y=False, name=None):
 @to_ivy_arrays_and_back
 def matrix_power(x, n, name=None):
     return ivy.matrix_power(x, n)
+
+
+# cond
+@with_supported_dtypes({"2.4.2 and below": ("float32", "float64")}, "paddle")
```
hirwa-nshuti, line 150: The paddle version is `2.5.0`
```suggestion
@with_supported_dtypes({"2.5.0 and below": ("float32", "float64")}, "paddle")
```

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
```diff
@@ -606,3 +606,52 @@ def test_paddle_matrix_power(
...
+                min_dim_size=1,
+                max_dim_size=5,
+                shape=st.tuples(st.integers(1, 5), st.integers(1, 5)).filter(
+                    lambda x: x[0] == x[1]
+                ),
+                allow_inf=False,
+                allow_nan=False,
+            )
+        )
+        if np.linalg.det(matrix[1][0]) != 0 and matrix_is_stable(matrix[1][0]):
```
hirwa-nshuti, line 631: The matrix_is_stable can be used in body of the function call using `matrix_is_stable(x[0])`

ivy/functional/frontends/paddle/tensor/linalg.py
ndurumo254: > The paddle version is `2.5.0`

how are we coming up with the version?

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
ndurumo254: > matrix_is_stable(x[0])

how can this be done, not sure how to move about it since all tests are failing after i made some changes

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
```diff
@@ -606,3 +606,55 @@ def test_paddle_matrix_power(
...
+@handle_frontend_test(
+    fn_tree="paddle.tensor.linalg.cond",
+    dtype_and_x=_get_dtype_and_matrix_non_singular(dtypes=["float32", "float64"]),
+    p=st.sampled_from([None, "fro", "nuc", np.inf, -np.inf, 1, -1, 2, -2]),
+    test_with_out=st.just(False),
+)
+def test_paddle_cond(*, dtype_and_x, p, on_device, fn_tree, frontend, test_flags):
+    dtype, x = dtype_and_x
+
+    if not matrix_is_stable(x[0]):
```
hirwa-nshuti, line 646: can we use assume here to filter stable martrix😂 just a final update I guess

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
hirwa-nshuti: You need to use assume to filter only when matrix is stable `assume(matrix_is_stable(x[0]))`

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
ndurumo254: > can we use assume here to filter stable martrix😂 just a final update I guess

as a decorator?

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
```diff
@@ -606,3 +607,54 @@ def test_paddle_matrix_power(
...
+                min_dim_size=1,
+                max_dim_size=5,
+                shape=st.tuples(st.integers(1, 5), st.integers(1, 5)).filter(
+                    lambda x: x[0] == x[1]
+                ),
+                allow_inf=False,
+                allow_nan=False,
+            )
+        )
+        if np.linalg.det(matrix[1][0]) != 0 and matrix_is_stable(matrix[1][0]):
```
hirwa-nshuti, line 632:
```suggestion
        if np.linalg.det(matrix[1][0]) != 0:
```

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
```diff
@@ -1,12 +1,13 @@
 # global
 import ivy
 from hypothesis import strategies as st
```
hirwa-nshuti, line 3:
```suggestion
from hypothesis import strategies as st, assume
```

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py
```diff
@@ -1,12 +1,13 @@
...
+from hypothesis import assume
```
hirwa-nshuti, line 4:
```suggestion
```

PR #17623, opened 2023-06-23
Add automatic device shifting decorator and soft device handling functions

bipinKrishnan:
Adds automatic device shifting functionality like in tensorflow and a soft device variable to control this behaviour.

The PR is related to the task here:
https://trello.com/c/TDEyHdDe


ivy-leaves:

Thank you :hugs:

ivy/functional/ivy/device.py
```diff
@@ -91,6 +92,7 @@ def __enter__(self):
         "cpu"
         """
         ivy.set_default_device(self._dev)
+        ivy.set_soft_device_mode(True)
```
vedpatwardhan, line 95: Probably you've mentioned this already, but not sure why we're enabling the `soft_device_mode` inside `default_device` if the default value of `soft_device_mode` seems to be [`False`](https://github.com/unifyai/ivy/pull/17623/files#diff-ca7db18115e5b52f90e231d41fae2418e12b8598b43cb7bb23ab9448d4d13af8R342)

ivy/functional/ivy/device.py
```diff
@@ -267,6 +283,66 @@ def print_all_ivy_arrays_on_dev(
...
+    >>> ivy.get_soft_device_mode()
+    True
+    """
+    global soft_device_mode_stack
+    if soft_device_mode_stack:
+        soft_device_mode_stack.pop(-1)
+
+
+@handle_exceptions
+def get_soft_device_mode() -> bool:
```
vedpatwardhan, line 329: Given that we've now replaced all getter methods for modes by [properties](https://github.com/unifyai/ivy/commit/f23a27b5128b713fc3e317de32f591fe05724928), we'd need to incorporate that in this mode too

vedpatwardhan (changes requested):
Hey @bipinKrishnan, thanks for making all the changes so far 💪
Could you please also add a test for `ivy.handle_soft_device_variable` inside `test_device.py`? Would be great to also test if the soft device mode is working properly.
The tests seem to pass which is great 🚀

ivy/functional/ivy/device.py
bipinKrishnan: The change was made so as to use `ivy.DefaultDevice` in the same way as `tf.device` context manager. Every operation in the context should be performed in the same device provided to the context manager. For this to happen we set soft device mode to `True` and unset it(in `__exit__`) :slightly_smiling_face:

vedpatwardhan: Oh okay, makes sense! Thanks for clarifying 👍

vedpatwardhan (changes requested):
Hey @bipinKrishnan, just ran through the PR a second time, there's a few final touches which would make it the most elegant solution 🔥
Also, could you please add the explanation of the policy in the docs referring the doc as well? Thanks and apologies for the changes coming in at every review 😅

ivy/functional/ivy/device.py
```diff
@@ -126,12 +128,26 @@ def __exit__(
...
         ivy.unset_default_device()
+        ivy.unset_soft_device_mode()
         if self and (exc_type is not None):
             print(exc_tb)
             raise exc_val
         return self
 
 
+def handle_soft_device_variable(*args, **kwargs):
+    default_device = ivy.default_device()
```
sherry30, line 139: This function expects native arrays and will return native arrays.
For this, You should set `array_mode` to false after entering this function then unset it before returning.

ivy/functional/ivy/device.py
```diff
@@ -126,12 +128,26 @@ def __exit__(
...
+    args, kwargs = ivy.nested_map(
+        [args, kwargs],
+        lambda x: (
+            x.to_device(default_device)
+            if (isinstance(x, ivy.Array) and ivy.dev(x) != default_device)
```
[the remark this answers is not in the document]
sherry30: should use `ivy.is_native_array` since it expects native arrays

ivy/functional/ivy/device.py
```diff
@@ -126,12 +128,26 @@ def __exit__(
...
             raise exc_val
         return self
 
 
+def handle_soft_device_variable(*args, **kwargs):
+    default_device = ivy.default_device()
+    args, kwargs = ivy.nested_map(
+        [args, kwargs],
+        lambda x: (
+            x.to_device(default_device)
```
sherry30, line 143: use `ivy.to_device(x, default_device)`

sherry30 (changes requested):
I was playing around with the soft_device mode and noticed it wasn't working fine in some cases especially for torch.
I've pointed out the required changes.
I think you can also remove the jax and paddle `handle_soft_device_mode` function since they seem to be exactly same as the mixed function.

sherry30 (changes requested):
Can you remove this decorator from all the mix and compositional functions.
For mix functions, you can add it in the `to_add` key of `mixed_backend_wrappers` attribute of the function.
You can check the list of functions [here](https://docs.google.com/spreadsheets/d/1DlpTXX5xG9e046F5OZ_K-U_5pALphlYkX_A9Gt_MYV4/edit#gid=0)

vedpatwardhan (commented):
Thanks for making the changes, could you please fix the issue @sherry30 mentioned with the `paddle` default device and finally add an explanation for all this in the deep dive? 
We can get the PR merged once that's done 👍👏

vedpatwardhan:
Hey @bipinKrishnan, could you please pull from master and resolve the conflicts to see if the fix pushed by @MahmoudAshraf97 resolves the paddle gpu issue? Also it would be great if you could run a few gpu tests locally, just to confirm that we aren't breaking anything (given that there aren't any gpu tests on PRs at the moment 😅)? We could get the PR merged today if everything works fine, thanks 🙂

ivy-gardener

vedpatwardhan:
ivy-gardener

bipinKrishnan:
As discussed, I'm working on a separate PR for the device shifting deep dive @vedpatwardhan

vedpatwardhan (approved):
lgtm! Feel free to merge the PR if the `tf.device` issue regarding inputs being on the `cpu` with a `gpu` default_device when `soft_device_mode=False` is resolved. Thanks @bipinKrishnan @sherry30 @MahmoudAshraf97 🚀🔥

PR #17633, opened 2023-06-23
Add ```paddle.tensor.manipulation.tensordot``` Manipulation Functions to Paddle Frontend

### LINKED ISSUE #17207: tensordot

benkimz:
Close #17207

### DIFF
--- a/ivy/functional/frontends/torch/tensor.py
+++ b/ivy/functional/frontends/torch/tensor.py
@@ -750,7 +750,7 @@ class Tensor:
         return torch_frontend.flatten(self, start_dim, end_dim)
 
     @with_unsupported_dtypes({"2.0.1 and below": ("float16",)}, "torch")
-    def cumsum(self, dim, dtype):
+    def cumsum(self, dim, *, dtype=None):
         return torch_frontend.cumsum(self, dim, dtype=dtype)
 
     @with_unsupported_dtypes({"2.0.1 and below": ("float16",)}, "torch")


sherry30 (changes requested):
Actually 1 thing I missed 😅 
`tensordot` also takes `axes` as an argument.
You can check it out [here](https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/tensordot_en.html)
Can you also add this and also test this argument.

Thanks

benkimz:
> Actually 1 thing I missed 😅 `tensordot` also takes `axes` as an argument. You can check it out [here](https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/tensordot_en.html) Can you also add this and also test this argument.
> 
> Thanks

I have included the axes argument. 
Thank you! 😊

sherry30 (changes requested):
Hi,

Thanks for the changes 😄.
One thing that's causing the tests to fail is that `heleprs.dtype_and_values` doesn't produce a value for `axis`, you will have to use `helpers.dtype_values_axis` to generate an axis along with an array. You can check out any other function which has an axis argument for reference how to use this helper function. Hope this makes sense.

Thanks

benkimz:
> Hi,
> 
> Thanks for the changes 😄. One thing that's causing the tests to fail is that `heleprs.dtype_and_values` doesn't produce a value for `axis`, you will have to use `helpers.dtype_values_axis` to generate an axis along with an array. You can check out any other function which has an axis argument for reference how to use this helper function. Hope this makes sense.
> 
> Thanks

All the tests are now passed. 
Thank you!

PR #17652, opened 2023-06-23
Add conv1D to paddle backend

### LINKED ISSUE #17649: conv1d

melch-inno:
close #17649


ivy/functional/backends/paddle/layers.py
```diff
@@ -77,18 +77,52 @@ def _pad_before_conv(x, filters, strides, padding, dims, dilations, data_format)
...
+    {"2.5.0 and below": {"cpu": ("float16",)}},
+    backend_version,
+)
 def conv1d(
     x: paddle.Tensor,
     filters: paddle.Tensor,
-    strides: Union[int, Tuple[int]],
-    padding: Union[str, int, Sequence[Tuple[int, int]]],
+    strides: Union[int, Tuple[int]] = 1,
+    padding: Union[str, int, Sequence[Tuple[int, int]]] = 0,
```
MahmoudAshraf97, line 88: please don't change anything in the function signature and keep it consistent across the framework

MahmoudAshraf97 (changes requested):
I see that you have special handling for different data formats although the conv function takes a data format argument, is there a reason for that?

ivy/functional/backends/paddle/layers.py
melch-inno: Got it 👌

melch-inno:
> I see that you have special handling for different data formats although the conv function takes a data format argument, is there a reason for that?

Hi @MahmoudAshraf97 , the hypothesis testing for paddle conv1d does not much the gound-truth "jax".

ivy/functional/backends/paddle/layers.py
```diff
@@ -88,7 +92,29 @@ def conv1d(
...
-    raise IvyNotImplementedException()
+    if data_format == "NCW":
+        x = paddle.transpose(x, perm=[0, 2, 1])
+
+    df = "NCL"
+    filters = paddle.transpose(filters, perm=[2, 1, 0])
+    x_pad = _pad_before_conv(
+        x, filters, strides, padding, dims=1, dilations=dilations, data_format=df
+    )
+    padding = "VALID"
```
MahmoudAshraf97, line 103: this is not correct, the function takes in a `padding` argument and you should accept it as is or manipulate it to achieve the same result

MahmoudAshraf97 (changes requested):
as in my previous review, data format should be handled in the function, no need for the transpose, and if the test is failing that is not related

ivy/functional/backends/paddle/layers.py
melch-inno: Hi @MahmoudAshraf97 , thanks for the comments:

The transposition needs to be done to match the shape gound-truth as it's done similarly in torch, tf, and numpy. The padding="Valid" is passed because the x input is already padded in the function `__pad_before_conv`, so there is no need to send the raw inputs; we can utilize the "Valid" option.

As for the hypotheses test, it passes all previous frameworks done already except paddle, and it's not good to change it for only paddle.

My first thought-out implementation was the code below:
```python

   strides = [strides] if isinstance(strides, int) else strides
   dilations = [dilations] if isinstance(dilations, int) else dilations
   df = "NCL" if data_format == "NCW" else "NLC"
   
    if isinstance(padding, int):
        padding = [(padding, padding)]
        
    x = _pad_before_conv(
        x, filters, strides, padding, dims=1, dilations=dilations, data_format=df
    )

    return paddle.nn.functional.conv1d(
        x=x,
        weight=filters,
        data_format=df,
        stride=strides,
        padding=padding,
        dilation=dilations,
    )
```
The `_pad_before_conv()` manipulates the padding, which could be used instead of the "valid" but it does not change anything.

will it be okay to chat further on Discord?

MahmoudAshraf97: I see your point, sure feel free to continue the discussion there, my ID is xmrfake

melch-inno:
Pending...

I have had a meeting with Mahmoud; currently waiting on Mahmoud to provide feedback from the internal team.

ivy-seed:

melch-inno:
[Repoted issue relating this ticket](https://github.com/PaddlePaddle/Paddle/issues/55596)

melch-inno:
Hi @MahmoudAshraf97, this task can be discussed internally because it can't pass the already written hypotheses and there is a bug on paddle side. link to reported issue can be found above

PR #17675, opened 2023-06-24
added paddle.median frontend function with its test

### LINKED ISSUE #16639: median

Mghrabi:
Close #16639 in #15115

### DIFF
--- a/ivy/functional/frontends/paddle/tensor/stat.py
+++ b/ivy/functional/frontends/paddle/tensor/stat.py
@@ -1,6 +1,6 @@
 # global
 import ivy
-from ivy.func_wrapper import with_unsupported_dtypes
+from ivy.func_wrapper import with_unsupported_dtypes, with_supported_dtypes
 from ivy.functional.frontends.paddle.func_wrapper import (
     to_ivy_arrays_and_back,
 )
@@ -31,3 +31,17 @@ def nanquantile(a, q, axis=None, keepdims=False, interpolation="linear", out=Non
     return ivy.nanquantile(
         a, q, axis=axis, keepdims=keepdims, interpolation=interpolation, out=out
     )
+
+
+@with_supported_dtypes(
+    {"2.5.0 and below": ("bool", "float16", "float32", "float64", "int32", "int64")},
+    "paddle",
+)
+@to_ivy_arrays_and_back
+def median(x, axis=None, keepdim=False, name=None):
+    x = (
+        ivy.astype(x, ivy.float64)
+        if ivy.dtype(x) == "float64"
+        else ivy.astype(x, ivy.float32)
+    )
+    return ivy.median(x, axis=axis, keepdims=keepdim)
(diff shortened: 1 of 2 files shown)


ivy-leaves:

Thank you :hugs:

Mghrabi:
Review please @DragosStoican, thanks

DragosStoican:
Hey @Mghrabi , can you make sure that the only modifications are to the function and the test for the function? There are a lot of changes in array_api_testing that shouldn't be there! Please remove those changes and then I'll quickly have another look! Thanks for your contribution!

Mghrabi:
Yeah, my bad

Mghrabi:
Done @DragosStoican

Mghrabi:
@DragosStoican When I made the changes the first time I directly checked if the PR introduced any new failures in GitHub actions and it didn't, and I thought that to be enough, but I turned out to be wrong :( , thanks for pointing that out.

Anyway, I made the required changes and everything should be fine from the frontend side, the only issue is with backends (except Tensorflow) since the test fails with (Returning the wrong data type), which is completely a backend issue. When I fixed them locally all tests passed for `test_paddle_median` which verifies the frontend is implemented correctly.

For example with the case of `numpy` for `median` func in `ivy/functional/backends/numpy/experimental/statistical.py` if we changed the following section 
``` 
...
    if input.dtype in [np.uint64, np.int64, np.float64]:
        return ret.astype(np.float64)
    elif input.dtype in [np.float16]:
        return ret.astype(input.dtype)
    else:
        return ret.astype(np.float32)
```
to:
```
...
 if input.dtype in [np.float64]:
        return ret.astype(np.float64)
 else:
        return ret.astype(np.float32)
```
this will pass the test for `numpy` backend, it simply makes numpy median function returns `float32` for all input data types except for `float64` which should be returned as it is according to docs [here](https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/median_en.html), This will make the test (`test_paddle_median`) to pass for `numpy` backend as mentioned, and the other tests will pass as well if the required changes made for each backend. 

Sorry for this long comment but this should verify the frontend function & its test are implemented as needed.

Note: the change I made for the `numpy` backend median function is not real fix for the returned type issue, it just fixes it for a single case (`paddle.median`).

Mghrabi:
Please let me know if something is not correct, thanks in advance

DragosStoican:
Hi @Mghrabi ! Sorry for the slow response I was away for the last few days!

I understand what you mean with the failing tests. However, with the changes you suggested for `numpy.meadian`, you would be introducing failing tests in other frontends. I tested your implementation and it causes a new failure in the `jax` frontend. In this scenario changing the backend function is not the correct thing to do, unless you notice a clear mistake (which I don't think is the case here).

Since `ivy.median` doesn't function in the correct way for the paddle frontend, I would suggest you add some logic in the frontend function to deal with the input dtypes such that the returned values are of the correct dtype. I think with a few type checks and casts inside the frontend function you can get `paddle.median` to work as expected and get all the tests passing!

Let me know how it goes!

Mghrabi:
@DragosStoican take your time, thanks for your valuable time.
I made the changes and all tests are passing and squashed the commits into one

btw when I made the changes for the backend, I was aware that they will make other frontends fail, the changes were just for demonstration purposes that the backend could be the source of the problem and they weren't the solution at all that's why I left a **Note** above. Thanks again for your time

PR #17698, opened 2023-06-24
fix and improve ivy.avg_pool2d tests

Sam-Armstrong:


ivy-leaves:

Thank you :hugs:

ivy_tests/test_ivy/test_functional/test_experimental/test_nn/test_layers.py
```diff
@@ -181,6 +182,11 @@ def test_avg_pool2d(
     fn_name,
 ):
     dtype, x, kernel, stride, pad = x_k_s_p
+
+    data_format = random.choice(["NCHW", "NHWC"])
```
sherry30, line 186: you shouldn't use random to generate any of the arguments, that might give you Flaky errors. all the arguments should be generated through hypothesis.

sherry30 (changes requested):
Hi Sam,

To generate `data_format`, you can pass `data_format = st.sampled_from(["channel_first", "channel_last"])` to `helpers.arrays_for_pooling`. This helper function has 1 problem tho, it doesn't return the `data_format` which would be randomly generated by this helper function. 

Can you fix this small issue. You will then also need to change the `data_format` from general `channel_first` or `channel_last` format to `NHWC` or `HCHW` in the test function. 
Hope that makes sense.

Thanks

Sam-Armstrong:
@sherry30 thanks, that's a good point. I've updated the test to randomly select a data format using hypothesis instead, do you think this looks ok?

sherry30 (approved):
It would be better to generate this through the helper function `arrays_for_pooling()` since it'll take care of the shapes of input and will be reusable in other pooling functions as well.
But this works for now, I'll take a look at it later when I try to improve this helper function.

thanks!

PR #17799, opened 2023-06-26
fix torch frontend avg_pool2d & tests

Sam-Armstrong:
Add ceil_mode, count_include_pad, divisor_override usage to avg_pool2d torch frontend. Also fixed the tests so they all pass 🙂

Test fix depends on https://github.com/unifyai/ivy/pull/17698 being merged.

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
+++ b/ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
@@ -5,6 +5,20 @@ from hypothesis import strategies as st
 import ivy
 import ivy_tests.test_ivy.helpers as helpers
 from ivy_tests.test_ivy.helpers import handle_frontend_test
+import math
+
+
+def is_same_padding(padding, stride, kernel_size, input_shape):
+    output_shape = tuple([(input_shape[i] + 2 * padding[i] - kernel_size[i]) // stride[i] + 1 for i in range(len(padding))])
+    return all([output_shape[i] == math.ceil(input_shape[i] / stride[i]) for i in range(len(padding))])
+
+
+def calculate_same_padding(kernel_size, stride, shape):
+    padding = tuple([max(0, math.ceil(((shape[i] - 1) * stride[i] + kernel_size[i] - shape[i]) / 2)) for i in range(len(kernel_size))])
+    if all([kernel_size[i] / 2 >= padding[i] for i in range(len(kernel_size))]):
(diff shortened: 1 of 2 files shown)


ivy-leaves:

Thank you :hugs:

ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
```diff
@@ -66,31 +80,31 @@ def test_torch_avg_pool1d(
         min_side=1,
         max_side=4,
     ),
```
hmahmood24, line 82: Afair, the `arrays_for_pooling` now has two arguments `explicit_or_str_padding` and `only_explicit_padding`. Tweaking their values should give you the explicit padding output instead of a string padding output. If that works then you wouldn't have to generate `str` paddings and convert them to explicit ones yourself, kind of how it's implemented in `avg_pool3d`. Can you maybe try that out?

hmahmood24 (changes requested):
Hey @Sam-Armstrong, I worked on this quite a few months ago and my PR at that time got merged with the incomplete signature since the `ivy.avg_pool2d` itself had some refactor going on. Anyways, left a comment for you. Let me know what you think about that?

ivy_tests/test_ivy/test_frontends/test_torch/test_pooling_functions.py
Sam-Armstrong: @hmahmood24 I actually did try that out, but the problem is this generates padding values which are not necessarily 'same' or 'valid' paddings. In that case our torch frontend will default to applying a valid padding, but the native torch function will continue applying the padding as defined - so there will be a shape mismatch. I guess this is a flaw with our ivy functions expecting a string padding. But if we're fine with this we can just keep the testing as I've done it and it will only check padding values which actually are 'same' or 'valid'.

hmahmood24: @Sam-Armstrong Right that makes sense. Thanks for the clarification. Although, given that `paddle` supports `str` or `numerical` padding, while `tensorflow` only supports `str` padding and `torch` supports `numeraical` one, I think it's fair to add this functionality to `ivy` itself such that `ivy` supports padding in all of these formats since we strive to implement the superset behaviour of all the other frameworks. Feel free to create a Trello card for this and/or notify the Framework team on Discord of the matter so someone can work on this, if not yourself. Thanks Sam 😃

hmahmood24 (approved):

PR #17811, opened 2023-06-26
Fix building haiku and flax modules with no params

xoiga123:
```python
import transformers

model = transformers.models.resnet.modeling_flax_resnet.Identity()
args = (jnp.ones((1, 10, 10)),)
params_v = model.init(jax.random.PRNGKey(0), *args)
ivy.transpile(model, to="torch", params_v=params_v, args=args)
```
Traceback:
```python
  File "ivy/stateful/converters.py", line 304, in _build
    _, param0 = next(param_iterator)

StopIteration: 
```

### DIFF
--- a/ivy/stateful/converters.py
+++ b/ivy/stateful/converters.py
@@ -170,7 +170,7 @@ class ModuleConverters:
                 params_dict = _hk_flat_map_to_dict(params_hk)
                 self._hk_params = ivy.Container(params_dict, dynamic_backend=False)
                 param_iterator = self._hk_params.cont_to_iterator()
-                _, param0 = next(param_iterator)
+                _, param0 = next(param_iterator, ["_", 0])
                 if hasattr(param0, "device"):
                     self._dev = ivy.as_ivy_dev(param0.device())
                 else:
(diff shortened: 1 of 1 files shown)


xoiga123:
@juliagsy Sorry if I troubled you :sweat_smile: I requested your review because you last worked on this

juliagsy:
Hey @xoiga123 ! No worries! Happy to review, but just to make sure, may I know what's the context of the issue?

xoiga123:
@juliagsy I assume that there are some layers/models that don't need any params (`Identity`) and this codeblock here in `_build` checks for a param to get the device that the model is on `self._dev`. So I set a default value for `next()` as `0` if params is empty, and let `ivy.dev(0)` make use of the default device set inside ivy. We can alternatively do something like
```python
if len(self._fx_params) == 0:
    self._dev = ivy.dev(0)
```
but this is cleaner I guess.

juliagsy:
@xoiga123 ahh I see, makes sense! Thanks for explaining! Looks good to me! Feel free to merge as long as all tests including related gc tests (if any) pass! Thanks again!

xoiga123:
@juliagsy Just checked gc tests: `test_demos`, `test_module` and `test_transpilation` all passed. I'll go ahead and merge. Thank youuu :smile:

PR #17833, opened 2023-06-26
added `unsorted_segment_min` to ivy functional API

### LINKED ISSUE #17437: unsorted_segment_min

akshatvishu:
Close #17437


zhumakhan:
Hi, thanks for your PR! Please use `with_unsupported_dtypes` decorator instead of validating data types inside the function.

akshatvishu:
> Hi, thanks for your PR! Please use `with_unsupported_dtypes` decorator instead of validating data types inside the function.

Can you tell me how to use for a particular `argument` ?  Because `num_segments` and `segment_ids` can be only of `int32`/`int64`  regardless of the backend while the `data` can be `float` , `int` and in some backends `unsigned_integers`.

zhumakhan:
Yeah, you are right, decorators check all inputs. Maybe you can try something like this: https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/general.py#L3163. If you look at its backend implementations they used `ivy.utils.assertions.check_gather_input_valid(params, indices, axis, batch_dims)` to check inputs

akshatvishu:
> Yeah, you are right, decorators check all inputs. Maybe you can try something like this: https://github.com/unifyai/ivy/blob/master/ivy/functional/ivy/general.py#L3163. If you look at its backend implementations they used (e.g. in torch backend) `ivy.utils.assertions.check_gather_input_valid(params, indices, axis, batch_dims)` to check inputs

Thank you for your insightful suggestion, @zhumakhan !!  I found it very useful and have incorporated a similar approach into all `backend` implementations. I value your feedback, so please don't hesitate to share more suggestions or highlight any other areas where you think improvements could be made.😁

zhumakhan:
LGTM!

akshatvishu:
> LGTM!

Thanks Zuma for the review!

PR #17890, opened 2023-06-27
Added dropout1d to Paddle Backend Experimental Layers 

yunsemr:


yunsemr:
This is not for an open issue, I wrote this backend function for the frontend issue: #17808 since the frontend function was failing for paddle backend during local test.

ivy/functional/backends/paddle/experimental/layers.py
```diff
@@ -257,7 +257,17 @@ def dropout1d(
...
+        is_batched = len(x.shape) == 3
+        if data_format == "NCW":
+            perm = (0, 2, 1) if is_batched else (1, 0)
+            x = paddle.transpose(x, perm)
+        res = paddle.nn.functional.dropout(x, prob)
+        if data_format == "NCW":
+            res = paddle.transpose(res, perm)
+    else:
+        res = x
+    return res
```
MahmoudAshraf97, line 270: these are suggested changes, no need to check for batched input and creating a permutation since we are swapping the last two axes which can be done using `swapaxes` or `moveaxis` in our case
also paddle dropout has a `training` argument which removes the need for the if condition

```suggestion
    if data_format == "NCW":
        x = paddle.moveaxis(x, -2,-1)
    res = paddle.nn.functional.dropout(x, p=prob,training=training)
    if data_format == "NCW":
        res = paddle.moveaxis(res, -2,-1)
    return res
```

MahmoudAshraf97 (changes requested):

yunsemr:
I committed the changes you suggested. Thanks for the information.

MahmoudAshraf97 (approved):

PR #17930, opened 2023-06-27
Fix color support during debugging

sladyn98:
In this pull request, we have introduced conditional coloration for JSON keys in the debug console output. This addresses the issue where hex color codes were appearing in the console output when viewed through debuggers like PyCharm or VSCode.

The core of the change involves checking for color support in the output console before applying any coloration to the keys. Specifically, we are now only coloring the keys if the output console supports color, i.e., if `os.isatty(sys.stdout.fileno()) `is True and the system is not Windows (os.name != "nt").

Also the check os.name != "nt" is done to verify if the current operating system is not Windows. This is important because the Windows command prompt does not natively support ANSI escape sequences (used for color and other terminal effects), so os.isatty(sys.stdout.fileno()) could return True on Windows, even though it doesn't fully support these sequences.

### DIFF
--- a/ivy/data_classes/container/base.py
+++ b/ivy/data_classes/container/base.py
@@ -9,6 +9,7 @@ import copy
 import termcolor
 import numpy as np
 import json
+import os
 
 from ivy.utils.exceptions import IvyBackendException, IvyException
 
--- a/ivy/data_classes/container/base.py
+++ b/ivy/data_classes/container/base.py
@@ -3903,6 +3903,8 @@ class ContainerBase(dict, abc.ABC):
             # color keys
             json_dumped_str_split = json_dumped_str.split('":')
             split_size = len(json_dumped_str_split)
+            # Checks if console supports color
+            has_color_support = os.isatty(sys.stdout.fileno()) and os.name != "nt"
             json_dumped_str = '":'.join(
                 [
                     (
@@ -3915,6 +3917,8 @@ class ContainerBase(dict, abc.ABC):
                                     ),
                                     self._default_key_color,
                                 )
+                                if has_color_support
+                                else Container.cont_trim_key(sub_str.split(' "')[-1], self._key_length_limit)
                             ]
                         )
                         if i < split_size - 1


sladyn98:
CC @vedpatwardhan

vedpatwardhan:
I'll request @sherry30's review on this one, as he's been working on the same issue recently. Thanks @sladyn98 🙂

sherry30 (commented):
Have you tested this out on vs code debug console?
I'm not sure if this will work. termcolor already has a check for `os.stdout.isatty()` [here](https://github.com/termcolor/termcolor/blob/dfa56ca2754fb01d0c887a6948e46d3cfa0b6cc5/src/termcolor/termcolor.py#L124), if this is False, it doesn't even try to add colors.
The problem with adding support of this to vs code debug console is that it actually supports ANSI escape sequences. 
You can try to execute the following line in it. 
`print("\033[1;32m This text is Bright Green  \n")` it'll print fine.
You can also try this line `os.environ['TERM']` and it'll print `xterm-256color`.
 My guess is termcolor adds something else that is not supported in the debug console. I'm not sure, still looking into it.

Thanks

Aarsh2001:
Hi @sladyn98 @vedpatwardhan @sherry30, I wanted to update you on an issue we encountered. It seems that manually setting the paths in the PyCharm config to include specific directories, such as `/opt/fw/tensorflow`, resulted in a duplicate installation of the `termcolor` package inside the TensorFlow directory. This caused problems with color rendering. However, you'll see that when you remove `/opt/bin/tensorflow` from the path  the issue is  resolved. 

I'll reach out to @RickSanchezStoic to address this and make the necessary updates to the Dockerfile. All we need to do is remove the duplicate `termcolor` installation from the directory. If you have any questions, please feel free to reach out to me directly via DM.

sladyn98:
@sherry30 Could you tell me some steps on how to reproduce this I am not sure what to attach the debugger to and how to run it. Thanks

Aarsh2001:
Just fixed the issue, we can close this PR now, thanks 🙂

sladyn98:
Thanks

sherry30:
Yuo can just create a new file and write simple code and debug it.

```
import ivy

x = ivy.Container({"x": 1})
print(x)
```

You can put the breaker on `print(x)` and test it out in the debug console.
if you type simply `x` in the debug console and press enter, it will print hex values instead colored output. but if you type `print(x)` instead, it'll print normally with colors. This works fine in pycharm's debug console.
I'm trying to see if there's a way to detect if ANSI escape sequences will be processed or not and disable color but no luck yet.

PR #17931, opened 2023-06-27
Automate squashing unifyai/docs into 1 commit

sladyn98:
This pull request introduces a new GitHub Actions workflow that will automatically squash all commits in the repository into a single commit every week at `00:00` every Sunday. This addition is aimed at reducing clutter and improving the readability of ivys repository's commit history. The key to mitigating race conditions in this workflow is the `git pull origin` master step. By pulling the latest changes right before squashing the commits, the workflow ensures that it is working with the most up-to-date version of the repository. This reduces the chance of losing updates due to race conditions.

Another possible solution that I could think of: 

**Implement a locking mechanism**: To prevent other CI jobs from committing while the squash operation is underway, we could use a simple locking mechanism. This can be a file in the repository that signals when the squash operation is ongoing. Other CI/CD processes should be configured to check for this lock before performing any operations.

Let me know your thoughts and feedback

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,9 +7,36 @@ on:
     tags:
       - v*
   workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * 0'  # Runs at 00:00 every Sunday
 
 jobs:
   update-docs:
     name: Update docs
     uses: unifyai/workflows/.github/workflows/docs.yml@master
     secrets: inherit
+
+  squash:
+    if: github.event_name == 'schedule'
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+      with:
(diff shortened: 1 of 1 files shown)


KareemMAX (changes requested):
This PR is destructive and shouldn't be merged, it was meant to https://github.com/unifyai/docs and not this repo

sladyn98:
@KareemMAX So this PR would be a viable solution for the docs, or do we want to go with a different approach ?

KareemMAX:
@sladyn98 yeah I guess so, I was actually searching for this PR so I can borrow the code you wrote 😅. I guess we can accept for now the tradeoff which is there is a small chance of missing few changes. As they will be rebuilt by future commits anyway.

Our only concern is having a race condition occurring when making a release. We do `workflow_dispatch` as a contingency for that exact issue. So I think I'll be porting your code over as is.

PR #17984, opened 2023-06-28
Updated invalid link in ivy_tests.rst

Zaituny:


Zaituny:
I was not sure whether the references should be from a specific commit or from the master branch, so I went with the master branch

PR #17999, opened 2023-06-28
Update nn.py with conv_transpose function (#17990)

ViktorTurchenko:
#17990 Created a conv_transpose function. I don't understand how many dimensions is it supposed to have (like 1 in Conv1D). So I added an n to pass this argument into the function. Please explain further the functionality that is supposed to be implemented as I could find any alike function in the TensorFlow documentation. Thanks!

### DIFF
--- a/ivy/functional/frontends/tensorflow/nn.py
+++ b/ivy/functional/frontends/tensorflow/nn.py
@@ -159,9 +159,9 @@ def conv_transpose(
 ):
     N = len(input)-2
     dilations = 1 if dilations is None else dilations
+    filters = filters.swapaxes(-2, -1)
     if N==1:
         strides, dilations = _reduce_strides_dilations(N, strides, dilations)
-        filters = filters.swapaxes(-2, -1)
         return ivy.conv1d_transpose(
             input,
             filters,
@@ -173,7 +173,6 @@ def conv_transpose(
         )
     elif N==2:
         strides, dilations = _reduce_strides_dilations(N, strides, dilations)
-        filters = filters.swapaxes(-2, -1)
         return ivy.conv2d_transpose(
             input,
             filters,
@@ -185,7 +184,6 @@ def conv_transpose(
         )
     elif N==3:
         strides, dilations = _reduce_strides_dilations(N, strides, dilations)
-        filters = filters.swapaxes(-2, -1)
         return ivy.conv3d_transpose(
             input,
             filters,
@@ -196,7 +194,7 @@ def conv_transpose(
             dilations=dilations,
         )
     else:
-        ivy.utils.exceptions.IvyException()
+        raise ivy.utils.exceptions.IvyException()
 
 @with_unsupported_dtypes({"2.12.0 and below": ("bfloat16",)}, "tensorflow")
 @to_ivy_arrays_and_back
--- a/ivy/functional/frontends/tensorflow/nn.py
+++ b/ivy/functional/frontends/tensorflow/nn.py
@@ -147,6 +147,33 @@ def conv3d_transpose(
 
 @with_unsupported_dtypes({"2.12.0 and below": ("bfloat16",)}, "tensorflow")
 @to_ivy_arrays_and_back
+def conv_transpose(
+    n,
+    input,
+    filters,
+    output_shape,
+    strides,
+    padding="SAME",
+    data_format="NDHWC",
+    dilations=None,
+    name=None,
+):
+    dilations = 1 if dilations is None else dilations
+    strides, dilations = _reduce_strides_dilations(n, strides, dilations)
+    filters = filters.swapaxes(-2, -1)
+    return ivy.conv_transpose(
+        n,
+        input,
+        filters,
+        strides,
+        padding,
+        output_shape=output_shape,
+        data_format=data_format,
+        dilations=dilations,
+    )
+
+@with_unsupported_dtypes({"2.12.0 and below": ("bfloat16",)}, "tensorflow")
+@to_ivy_arrays_and_back
 def depthwise_conv2d(
     input,
     filter,
--- a/ivy/functional/frontends/tensorflow/nn.py
+++ b/ivy/functional/frontends/tensorflow/nn.py
@@ -157,9 +157,10 @@ def conv_transpose(
     dilations=None,
     name=None,
 ):
+    N = len(input)-2
     dilations = 1 if dilations is None else dilations
-    if data_format=="NWC":
-        strides, dilations = _reduce_strides_dilations(1, strides, dilations)
+    if N==1:
+        strides, dilations = _reduce_strides_dilations(N, strides, dilations)
         filters = filters.swapaxes(-2, -1)
         return ivy.conv1d_transpose(
             input,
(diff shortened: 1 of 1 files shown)


ivy-leaves:

Thank you :hugs:

xoiga123:
Hi @ViktorTurchenko! Thank you for contributing to Ivy :hugs: 2 little things though:
- Please edit the PR **description** to correctly link an issue, you edited the **title** of the PR.
- In any case, a frontend must exactly match the framework's function signature. So you can't pass in a `n` argument here. This `conv_transpose` would call `conv1d_transpose`, `conv2d_transpose` or `conv3d_transpose` based solely on the data shape of the inputs passed in. You would have to infer `input`, `filters`, `data_format` or `dilations` to get the correct `N` that is specified in the function's documentation.

ViktorTurchenko:
I have a question 3 more questions:
1. If we are deciding which function to use with the `data_format` argument may I move it so it is a compulsory argument? 
2. If no then what default value should I use (most users will get errors if they don't change and their data is not matching the format needed)?
3. Should I check the 3-rd condition with a `elif` statement and for everything else raise some kind of an error?
Thanks for your time!

ViktorTurchenko:
I changed the `if-else` to a required argument and added the exception at the end. Sorry I haven't read the documentation enough to understand the required formats 😃

ivy/functional/frontends/tensorflow/nn.py
```diff
@@ -145,6 +145,59 @@ def conv3d_transpose(
...
+    input,
+    filters,
+    output_shape,
+    strides,
+    padding="SAME",
+    data_format="NDHWC",
+    dilations=None,
+    name=None,
+):
+    N = len(input)-2
```
xoiga123, line 160: This seems wrong:
```python
>>> arr = tf.constant([[1]])
>>> arr.shape
TensorShape([1, 1])
>>> len(arr)
1
```
`arr` is 2 dimensional here.

ivy/functional/frontends/tensorflow/nn.py
```diff
@@ -145,6 +145,59 @@ def conv3d_transpose(
...
+            input,
+            filters,
+            strides,
+            padding,
+            output_shape=output_shape,
+            data_format=data_format,
+            dilations=dilations,
+        )
+    else:
+        ivy.utils.exceptions.IvyException()
```
xoiga123, line 199: Please `raise Exception` :black_flag:

ivy/functional/frontends/tensorflow/nn.py
```diff
@@ -145,6 +145,59 @@ def conv3d_transpose(
...
+        strides, dilations = _reduce_strides_dilations(N, strides, dilations)
+        filters = filters.swapaxes(-2, -1)
+        return ivy.conv3d_transpose(
+            input,
+            filters,
+            strides,
+            padding,
+            output_shape=output_shape,
+            data_format=data_format,
+            dilations=dilations,
```
xoiga123, line 196: There is some repetition here, can we shorten this somehow?

xoiga123 (changes requested):
As this is a Frontend API, please implement an accompanying test as described in the [Frontend API Guide](https://unify.ai/docs/ivy/overview/contributing/open_tasks.html#frontend-apis). Thank you :hugs:

ivy/functional/frontends/tensorflow/nn.py
ViktorTurchenko: I think the `len()` is the right way to do it because you put an array inside an empty array (basically a double array). The double array in this case doesn't make any point for us. In the function, we are trying to find how many dimensions of the first order are there, and in this case, there is only one dimension of the first order and not 2. If we are trying to figure out the "NDHWC", we are trying to figure out the dimension of the first order and we don't care if it is "NDHWC" or 
"[N]DHWC" or "[[N]]D[H]WC". We just need to know that there are 5 first-order dimensions. The other advantage of `len()` is that some people might still use a list data type, which would crash when using `shape()` on it. If the argument is restricted to arrays only, then Yes, you are right and we could use the `shape()` function.

ViktorTurchenko: Now the `shape()` is also not so perfect, if we are trying to input something like this 
`arr = tf.constant([[[1, 2], [1, 2], [1, 2], [2, 2]]])`
`print(arr.shape)`
`print(len(arr))`
the output is 
`(1, 4, 2)`
`1`
and if you would index on the first element of the tuple to figure out the dimension you would get the wrong thing as you are expecting to count the first order dimentions.

ViktorTurchenko: On the other hand if to remove one dimension and put
arr = tf.constant([[1, 2], [1, 2], [1, 2], [2, 2]])
print(arr.shape)
print(len(arr))
the output is
(4, 2)
4
which perfectly works with `len` too

ViktorTurchenko:
`@handle_frontend_test(`
    `fn_tree="tensorflow.nn.conv_transpose",`
    `x_f_d_df=_x_and_filters(`
        `dtypes=helpers.get_dtypes("float", full=False),`
        `data_format=st.sampled_from(["NDHWC"]),`
        `padding=st.sampled_from(["SAME"]),`
        `type="3d",`
        `transpose=True,`
    `),`
    `test_with_out=st.just(False),`
`)`

For the frontend test part what `type=` argument should be inserted?

ivy/functional/frontends/tensorflow/nn.py
xoiga123: I don't think `N` is literally the first `N` dimension in `NC` or `NWC` or `NHWC`.
> input | An N+2 dimensional Tensor of shape [batch_size] + input_spatial_shape + [in_channels] if data_format does not start with "NC" (default), or [batch_size, in_channels] + input_spatial_shape if data_format starts with "NC"

The `N` you are describing is `batch_size`, and the `N` that is used throughout the docs (`N+2 dimensional Tensor`) is the number of dimensions of `input` minus 2. You can try this example:
```python
>>> arr = tf.random.uniform([1,3,224,224])
>>> arr.shape
TensorShape([1, 3, 224, 224])
# a single RGB image of size 224x224
# batch_size - channel - height - width
# so N is 2, and the correct format is either `NHWC` or `NCHW`
>>> len(arr)
1
# if N is 1, what is the correct format?
```

xoiga123:
> For the frontend test part what `type=` argument should be inserted?

You're going to have to sample from `1d`, `2d` or `3d` as `conv_transpose` is all 3 of them.

ivy/functional/frontends/tensorflow/nn.py
ViktorTurchenko: Alright, so can I use the `shape` function? (then it will not work for lists). That's the reason I wanted to use `len()`. If I can use `shape` function I can just do `len(arr.shape)-2` which would give me the right number of dimensions. Another solution would be to transform all lists in numpy arrays before taking the `shape`.

xoiga123: Does `tf.nn.conv_transpose` accept lists as input?

ivy-seed:

ivy-seed:

PR #18003, opened 2023-06-28
scipy.fft2

### LINKED ISSUE #18002: fft2

mobley-trent:
Closes #18002

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_scipy/test_fft/test_scipy_fft.py
+++ b/ivy_tests/test_ivy/test_frontends/test_scipy/test_fft/test_scipy_fft.py
@@ -1,5 +1,5 @@
 # # TODO: uncomment after frontend is not required
-# global
+# # global
 # from hypothesis import strategies as st
 
 # # local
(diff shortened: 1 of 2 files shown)


mobley-trent:
@AnnaTz @hmahmood24 Is this the correct format ?

ivy_tests/test_ivy/test_frontends/test_scipy/test_fft/test_scipy_fft.py
```diff
@@ -106,6 +106,36 @@
...
+#             small_abs_safety_factor=2.5,
+#             safety_factor_scale="log",
+#         )
+#     )
+#     s = (
+#         draw(st.integers(min_fft2_points, 256)),
+#         draw(st.integers(min_fft2_points, 256)),
+#     )
+#     dim = draw(st.sampled_from([(0, 1), (-1, -2), (1, 0)]))
+#     norm = draw(st.sampled_from(["backward", "forward", "ortho"]))
```
AnnaTz, line 135: The general format is alright but the test is not general enough because you are hardcoding `dim`. I would suggest you use `dtype_values_axis` to get both `x` and `dim`. In this case you won't need to generate `x_dim` and `dtype` separately either, you can set them using `dtype_values_axis`'s arguments.

AnnaTz (changes requested):

PR #18023, opened 2023-06-29
EHN: Adding bmm method in paddle tesnor.py

### LINKED ISSUE #18021: bmm
adding bmm in paddle tesnor.py

MuhammadNizamani:
Closed #18021


MuhammadNizamani:
@YushaArif99 can you review this PR?

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_tensor.py
```diff
@@ -2586,3 +2586,39 @@ def test_paddle_instance_sgn(
...
+
+
+# bmm
+@handle_frontend_method(
+    class_tree=CLASS_TREE,
+    init_tree="paddle.to_tensor",
+    method_name="bmm",
+    dtypes_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
+        shape=(3, 3, 3),
```
YushaArif99, line 2598: This seems a little too specific. We should ideally be testing on a subset of the 3D space of inputs here. 
So the input shapes should be `(b, m, k)`  and `(b, k, n)` where b,k,n could be any integers. 
You could try creating a composite helper to achieve this. Or perhaps something similar to this already exists (not sure though aha 😅 )

YushaArif99 (changes requested):
Hi @MuhammadNizamani 

Apologies for the late review. Have been extremely busy with work recently and so haven't had the time to review PRs 😓 
Could you kindly address the abovementioned point and re-request a review once that's done 🙂

MuhammadNizamani:
> Hi @MuhammadNizamani
> 
> Apologies for the late review. Have been extremely busy with work recently and so haven't had the time to review PRs sweat Could you kindly address the abovementioned point and re-request a review once that's done slightly_smiling_face

ok I will

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_tensor.py
MuhammadNizamani: I checked ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_linalg.py and in this file test bmm also tested that way. Check it please.

MuhammadNizamani:
@YushaArif99  i don't think shape should be dynamic because when I change the shape test got fail cause was matmul function.

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_paddle_tensor.py
YushaArif99: Yeah, I checked it, and it seems like a slip up most probably from the PR reviewer. Ideally, we should be testing for all corner cases rather than testing for only a subset just to get the tests passing.

YushaArif99:
> @YushaArif99 i don't think shape should be dynamic because when I change the shape test got fail cause was matmul function.

I noticed the change you made in [this commit](https://github.com/unifyai/ivy/pull/18023/commits/31593cd69d08df737b0712b1b943d9179b6ae3b1) and I'm a little confused on what you're trying to accomplish here: 
```python
shape=st.tuples(st.integers(3, 3), st.integers(3, 3), st.integers(3, 3)).filter(
            lambda x: x[0] == x[1]
        ),
```
What you need to do is generate 3-dimensional inputs such that their shapes are `(b,n,k)` and `(b,k,m)` so in a way you need to generate 4 integers `(b,n,k.m)` and you can then draw random arrays by specifying the shapes to be those mentioned previously. Hope that makes sense 🙂

PR #18034, opened 2023-06-29
Update ivy.Array__deepcopy__ to avoid non-leaf tensor errors

AnnaTz:
After [this](https://github.com/unifyai/ivy/commit/2ec1da189b0d81b44d4a0826ea74f4130fbd361c) commit, frontend tests started getting errors like [this](https://discord.com/channels/799879767196958751/1028707571743334420/1122582044724170863). The errors are avoided with this updated implementation of `ivy.Array__deepcopy__`.

### DIFF
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -1144,10 +1144,12 @@ class Array(
                 jax_array = ivy.array(np_array)
                 return to_ivy(jax_array)
             return to_ivy(copy.deepcopy(self._data))
-        except:
-            array_copy = ivy.copy_array(self._data)
-            memodict[id(self)] = array_copy
-            return to_ivy(array_copy)
+        except RuntimeError:
+            from ivy.functional.ivy.gradients import _is_variable
+            # paddle and torch don't support the deepcopy protocol on non-leaf tensors
+            if _is_variable(self):
+                return to_ivy(copy.deepcopy(ivy.stop_gradient(self)._data))
+            return to_ivy(copy.deepcopy(self._data))
 
     def __len__(self):
         if not len(self._data.shape):
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -1134,20 +1134,9 @@ class Array(
         return ivy.bitwise_right_shift(self._data, other)
 
     def __deepcopy__(self, memodict={}):
-        try:
-            return to_ivy(self._data.__deepcopy__(memodict))
-        except AttributeError:
-            # ToDo: try and find more elegant solution to jax inability to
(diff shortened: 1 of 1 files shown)


ivy/data_classes/array/array.py
```diff
@@ -1134,20 +1133,9 @@ def __irshift__(self, other):
         return ivy.bitwise_right_shift(self._data, other)
 
     def __deepcopy__(self, memodict={}):
-        try:
```
vedpatwardhan, line 1137: I don't think we should be removing this, ideally we should always defer to the native framework dunder method (`__deepcopy__` in this case) as the default option, and should handle any framework-specific special cases in the `except` blocks.

vedpatwardhan (commented):
Hey @AnnaTz, I wasn't able to find use of `self.is_leaf` in the codebase, but I think this is related to the `torch.Tensor.is_leaf` property in the frontends? It would be great if you could share the full stack trace of the issue. Thanks 🙂

ivy/data_classes/array/array.py
AnnaTz: Yeah I had a feeling you weren't going to like that 😅

ivy/data_classes/array/array.py
```diff
@@ -1144,10 +1144,10 @@ def __deepcopy__(self, memodict={}):
                 jax_array = ivy.array(np_array)
                 return to_ivy(jax_array)
             return to_ivy(copy.deepcopy(self._data))
-        except RuntimeError:
```
vedpatwardhan, line 1147: I think we can update this `except` block to something like this instead,
```
        except RuntimeError:
            from ivy.functional.ivy.gradients import _is_variable
            if ivy.current_backend_str() == "paddle":
                return to_ivy(copy.deepcopy(self._data.numpy()))
            elif _is_variable(self):
                return to_ivy(copy.deepcopy(ivy.stop_gradient(self)._data))
            return to_ivy(copy.deepcopy(self._data))
```
Could you please also add some commit above the `elif` stating the leaf tensors issue? Thanks!

vedpatwardhan (changes requested):
Hey @AnnaTz, thanks for making the changes, so for future reference this is a problem in `ivy` that can be reproduced with the following example,
```
import copy
import torch
import ivy


ivy.set_backend('torch')
x = torch.tensor([1.], requires_grad=True)
x = ivy.array(x)
x = x + 1
print(x)
print(copy.deepcopy(x))
```
The issue here is that `x` is a leaf-tensor after performing an operation on it, and `torch` doesn't support the deepcopy protocol on leaf tensors. 
Just requested a change, and we should be good to merge the PR. Thanks @AnnaTz 🙂

PR #18051, opened 2023-06-29
added rfftn function at backend

### LINKED ISSUE #18035: rfftn
propose adding the RFFTN function to Ivy's Experimental API., I have been working on the RFFT2 function for the NumPy frontend for over a month now, and I've come across some complexities while iterating over multiple axes within the RFFT2 function.

Considering this, I would like to suggest exploring the implementation of the NumPy RFFT2 function in Ivy's backend using existing NumPy functions. By adding the RFFTN function to the backend, we can simplify the RFFT2 implementation, as RFFT2 can be considered a subset of RFFTN. We will need to handle this difference in the backend logic.

I understand that adding around 30 lines of logic to the frontend goes against Ivy's frontend design policy. However, by incorporating the RFFTN function in the backend, we can maintain consistency with other backends and simplify the frontend implementation of RFFT2.

rajveer43:
Close #18035


Fayad-Alman:
Hi @rajveer43, for this function you'll have to add the backend implementation to the layers.py file of every backend framework. Once that's done, add the function to the ivy array API (layers.py in the array folder) and container class as well (layers.py in the container folder).

After all of this is complete, the final step is creating a test for this function. Go to ivy_tests/test_ivy/test_functional/test_experimental/test_nn/test_layers.py and add the test.

Let me know if you have any questions!

ivy/functional/ivy/experimental/layers.py
```diff
@@ -2480,3 +2480,67 @@ def ifftn(
...
+@handle_out_argument
+@inputs_to_ivy_arrays
+def rfftn(
+    x: Union[ivy.Array, ivy.NativeArray],
+    /,
+    *,
+    s: Optional[Sequence[int]] = None,
+    axes: Optional[Sequence[int]] = None,
+    norm: Optional[str] = None,
+) -> ivy.Array:
```
Fayad-Alman, line 2496: I believe there should be an out argument here

Fayad-Alman (changes requested):
Hi @rajveer43, apologies for the long wait! I've been looking at this PR for a few days now, still trying to figure out the source of the errors popping up. For now, here are a few changes you could make:

ivy/data_classes/container/experimental/layers.py
```diff
@@ -2023,3 +2023,43 @@ def ifftn(
...
         )
+
+    def rfftn(
+        self: Union[ivy.Array, ivy.NativeArray, ivy.Container],
+        /,
+        *,
+        axes: Optional[Union[int, Tuple[int]]] = None,
+        s: Optional[Sequence[int]] = None,
+        norm: str = "backward",
+        out: Optional[ivy.Array] = None,
```
Fayad-Alman, line 2034: The order of parameters should be the same for each rfftn function

ivy/functional/backends/jax/experimental/layers.py
```diff
@@ -799,3 +799,35 @@ def embedding(
...
+
+
+def rfftn2(
+    x: JaxArray,
+    *,
+    s: Sequence[int] = None,
+    dim: Sequence[int] = (-2, -1),
+    norm: str = "backward",
+    out: Optional[JaxArray] = None,
+) -> JaxArray:
```
Fayad-Alman, line 811: Is there a reason this function is named rfftn2?

Fayad-Alman: Also, there seems to be a new parameter 'dim' rather than the expected parameter 'axes'

ivy/functional/backends/numpy/experimental/layers.py
```diff
@@ -922,3 +922,35 @@ def embedding(
...
+
+
+def rfftn(
+    x: np.ndarray,
+    *,
+    s: Sequence[int] = None,
+    dim: Sequence[int] = (-2, -1),
+    norm: str = "backward",
+    out: Optional[np.ndarray] = None,
+) -> np.ndarray:
```
Fayad-Alman, line 934: There seems to be a new parameter 'dim' rather than the expected parameter 'axes'

Fayad-Alman (changes requested):
@rajveer43 For the container of the function, you need to add a static_rfftn as well. Please see the file for examples and the general format.

ivy/functional/backends/paddle/experimental/layers.py
```diff
@@ -365,3 +366,16 @@ def ifftn(
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
     return paddle.fft.ifftn(x, s, axes, norm)
+
+
+@with_unsupported_dtypes({"0.4.13 and below": ("float32", "complex")}, backend_version)
+def rfftn(
+    x: paddle.Tensor,
+    s: Optional[Union[int, Tuple[int]]] = None,
+    axes: Optional[Union[int, Tuple[int]]] = None,
```
Fayad-Alman, line 375: I believe the unsupported dtypes are float16 and complex

Fayad-Alman (changes requested):
Hi @rajveer43, apologies once again for the delayed response. 
I haven't looked much into the tensorflow implementation of rfftn (I'll update you if I find a potential solution), but I do think the paddle backend has one last issue.

Fayad-Alman (approved):
LGTM! Thanks for all your work and effort!

PR #18053, opened 2023-06-29
fix jitting jax backend split

mattbarrett98:
Currently the jax backend `split` fails to jit compile, due to the `num_or_size_splits` arg being converted into a jax array. This causes this argument to be traced by jax's jit, causing it to fail as the jit expects this argument to be a constant. So just fixed this by using numpy instead.

I think it could be a good idea to add jax jitting to the jax backend tests maybe @vedpatwardhan


mattbarrett98:
Will just push for now to save time, let me know if any issues

PR #18115, opened 2023-06-30
fix jax expand for shapes with more dims than input

Sam-Armstrong:
Fix a bug where the jax backend expand function cannot handle shapes which have more dims than the input, and one of the final dims is negative.

Min example:
```
import ivy

ivy.set_backend('torch')
x = ivy.random_normal(shape=(4))
x = x.expand((1, -1))
print(x.shape)

ivy.set_backend('jax')
x = ivy.random_normal(shape=(4))
x = x.expand((1, -1))
print(x.shape)
```


hirwa-nshuti (approved):
Looks good to me😊
feel free to merge

PR #18131, opened 2023-07-01
added real instance method to paddle

### LINKED ISSUE #18129: real

AkshayVenkataraman:
Close #18129


ivy/functional/frontends/paddle/tensor/tensor.py
```diff
@@ -494,3 +494,10 @@ def conj(self, name=None):
     )
     def logical_not(self, out=None, name=None):
         return ivy.logical_not(self.ivy_array)
+
+    @with_supported_dtypes(
+        {"2.5.0 and below": ("complex64", "complex128")},
+        "paddle",
+    )
+    def real(self, name=None):
+        return ivy.real(self._ivy_array)
```
jkeane508, line 503: If we are passing paddle tensors as arguments, its unlikely they will have an `_ivy_array` attribute. Which is why we are seeing test failures.

jkeane508 (changes requested):

PR #18158, opened 2023-07-02
fix docstring tests

sondertg:


sondertg:
There was only one main problem with docs that is multiline codes which should executed as a single line.
Example below.
```
        >>> x1 = ivy.Container(a=ivy.array([1., 2., 3.]),\
        ...                         b=ivy.array([1., 2., 3.]))
        >>> x2 = ivy.Container(a=ivy.array([1., 2., 3.0003]),\
        ...                         b=ivy.array([1.0006, 2., 3.]))
        >>> y = x1.allclose(x2, rtol=1e-3)
        >>> print(y)
```
`\` and `...` indicates continuation however parser didn't remove those ```...```. I fixed this by adding two lines to `trim` function.

Besides from that, got an error from `container.einsum`. Problem seems same but whatever I tried, didnt solve it so I just created a lint error. ( char limit exceeded)

Last minor problem was related to `numpy` backend. From dictionary, it draws a `Ivy.GlobalsDict` object and send it to do docs checkers which throws a TypeError. Solved it too by adding another condition.

Ishticode (changes requested):
Hi @sondertg 
Thank you very much for diving deep into this. One concern I have is the update to the trim logic is perhaps removing more than expected or otherwise.
If you go to a docstring and make the results totally inaccurate the test still passes. It means it has stopped detecting incorrect results. Like if you go to the docstring for ivy. abs in ivy/functional/ivy/elementwise.py and mess up some result. The docstring test should be detecting that and not passing but it still passes.

Could you kindly take a look into that. I can look into it too. Lmk if something doesn't make sense. Thanks again and I hope we will be able to resolve this in this PR. 🙂

sondertg:
Hi again, didn't think about that way. Yes , when I changed docs for random fn it didn't throw an error. I just thought it calculates and generate correct result for non-failing tests when I started to work on. I'll investigate it further.

ivy-leaves:

Thank you :hugs:

sondertg:
Hi @Ishticode , I just checked all docs for container methods and related failures. Below examples require checking backend or functions implementations. I solved some failures that related to `test_docstrings.py` functions or logic. I also solved some easy ones such as missing paranthesis or wrong usage of parameters etc. Rest (added ones to skip list or conditional skip lists) require fixes to original backend functions or ivy code. This means problem is not in the docs but the function itself as far as I see. 

Explanation to reasons, 
Wrong example: Unknown reason, when running doc example by using ivy, it doesn't work.

Container doesn't work: x.<function> should work but it throws "<function> require [# of] positional argument/s" which means container doesn't work as supposed to.


Name                    Reason                Error(if matters)

- Container.conv2d_transpose: Wrong example

- Container.conv3d_transpose: Wrong example

- Container.einops_reduce: Wrong example

- Container.l2_normalize: Wrong example, container doesn't work

- Container.lp_normalize: Wrong example, container doesn't work

- Container.max_pool2d: Wrong example, OverflowError: cannot convert float infinity to integer

- Container.pinv: Wrong example(2nd one), NameError: name 'a' is not defined

- Container.relu6: Wrong example, container doesn't work.


Below examples are just to show why I didn't debug for methods fails at most two backends. Usually they are related to internal structure of those backend frameworks. 

- Container.scatter_flat: jax doesn't support


- Container.unique_values: removed an example because of an inf loop(
x = ivy.Container(a=[1, 2, 3], b=[2, 2, 3], c=[4, 4, 4])
y = ivy.Container()
y = x.unique_values(out=y)
print(y))

- Container.mean: fails for torch due to below example:
x = ivy.Container(a=ivy.array([-1, 0, 1]), b=ivy.array([1.1, 0.2, 1.4]))
x.mean(out=x)
print(x)

(torch throws this error when mean equals to zero)

the error: could not infer output dtype. Input dtype must be either a floating point or complex dtype. Got: Long

- Container.prod: fails for numpy and torch:
ex 1:
x = ivy.Container(a=ivy.array([-1, 0, 1]), b=ivy.array([1.1, 0.2, 1.4]))
x.prod(out=x)
print(x)
ex 2:
x = ivy.Container(a=ivy.array([[0., 1., 2.], [3., 4., 5.]]),
                  b=ivy.array([[3., 4., 5.], [6., 7., 8.]]))
x.prod(axis=0, out=x)
print(x)

('numpy: prod: output parameter for reduction operation multiply has the wrong number of dimensions: Found 1 but expected 0',)


torch failure example:
x = ivy.Container(a=ivy.array([-1, 0, 1]), b=ivy.array([1.1, 0.2, 1.4]))
x.prod(out=x)
print(x)

after execution b generates triple 0 and 0.30800003 while it should generate a single value


- Container.scaled_dot_product_attention:
'numpy: scaled_dot_product_attention: numpy: nested_map: numpy: default_dtype: numpy: is_complex_dtype: numpy: as_ivy_dtype: Cannot convert to ivy dtype. print(result) is not supported by NumPy backend.'
'tensorflow: scaled_dot_product_attention: tensorflow: nested_map: tensorflow: default_dtype: tensorflow: is_complex_dtype: tensorflow: as_ivy_dtype: Cannot convert to ivy dtype. print(result) is not supported by TensorFlow backend.'
etc


- Container.abs:
this generates very close values to correct results and it passes for two backends
```py
import ivy
x = ivy.Container(a=ivy.array([-1.6, 2.6, -3.5]),
                  b=ivy.array([4.5, -5.3, -2.3]))
y = x.abs()
print(y)
```
output:
```
{
    a: ivy.array([1.60000002, 2.5999999, 3.5]),
    b: ivy.array([4.5, 5.30000019, 2.29999995])
}
```

for tensorflow backend it throws below error from `exec()`:
"tensorflow: abs: tensorflow: where: Attempt to convert a value (None) with an unsupported type (<class 'NoneType'>) to a Tensor."

ivy_tests/test_docstrings.py
```diff
@@ -285,6 +338,21 @@ def test_docstrings(backend):
         "dropout",
         "dropout1d",
         "dropout3",
+        # generate different results due to randomization
+        "svd",
+        # wrong examples/implementations
+        "conv2d_transpose",
```
Ishticode, line 344: wrong examples should be fixed or removed. The tests should not be skipped for these.

ivy/data_classes/container/experimental/activations.py
```diff
@@ -354,11 +354,9 @@ def static_relu6(
...
         Examples
         --------
-        >>> x = {
-                    a: ivy.array([-3., -2., -1., 0., 1., 2., 3., 4., 5.]),
-                    b: ivy.array([1., 2., 3., 4., 5., 6., 7., 8., 9.])
-                }
-        >>> y = ivy.Container.static_relu6(x)
+        >>> x = ivy.Container(a = ivy.array([-3., -2., -1., 0., 1., 2., 3., 4., 5.]),
+        ...                   b = ivy.array([1., 2., 3., 4., 5., 6., 7., 8., 9.]))
+        >>> y = x.static_relu6()
```
Ishticode, line 359: the correct way to call static methods is like `ivy.Container.static_relu6(x)` as these are not really like instance methods.

Ishticode (changes requested):
Hi @sondertg amazing work and thank you for following up wit the details.
I have made two requests. could you make these changes and then I will run a local test to see no unexpected behaviour is generated.

ivy_tests/test_docstrings.py
sondertg: Just cleared those, at next PR I'll solve array method docs with ivy functional docs.

ivy/data_classes/container/creation.py
```diff
@@ -1583,7 +1583,7 @@ def frombuffer(
         ...     a = b'\x00\x00\x00\x00\x00\x00\xf0?',
         ...     b = b'\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@'
         ... )
-        >>> y = ivy.Container.static_frombuffer(x)
+        >>> y = ivy.Container.frombuffer(x, dtype=ivy.float64)
```
Ishticode, line 1586: as this is instance method example, it should be `y = x.frombuffer(dtype=ivy.float64)`
I am not sure why the docstring test is missing this error out when I ran it locally from this branch. :thinking:

Ishticode (changes requested):

ivy/data_classes/container/creation.py
sondertg: docstring test can't check if container or array method used correctly. I tried to correct as long as I noticed.

sondertg:
> This is a super amount of work and a lot of effort. A great deal of improvement in docstrings.
> 
> I am a bit concerned about the logic updates and too many skips. Although a lot of it is understandble to be fixed incrementally and removed from skip lists.
> 
> Can I have @vedpatwardhan take a look at this. Especially the test docstring test logic part.
> 
> Thank you very much @sondertg. I will try to wrinkle out stuff quickly and get this merged over the weekend in no big issues. Thanks a lot again for the effort. Let's see

About the skip lists, few of them should be there because functions or methods not working properly. Some of them generates different results in different backends. (for example when you call supported devices, jax return a list ["cpu", "gpu"] and tensorflow ["gpu", "cpu"]). I can add an if statement to fix this  with few line of codes but for each exception, I need to add another logical check. This might cause test fuction grows a lot.
That words_to_remove list there because of this. `tf.tensor([1, 2])` is same as `torch.tensor([1, 2])`. However, I am not even sure that we need such a remove list cause very few function require such checks.

I didn't make any major logical changes. First problem was multiline execution. I realized buffer throws a common syntax error due to unremoved `...`.  Solved that with a simple loop. 

Carried skip lists to doc checker function because the dictionary sometimes just throws function object but no name. This cause weird behaviours. ( When I add ` if k != "some_function": continue `,  I was running tests faster. For array and container methods, this worked well but for functionals, some dictionary elements are just objects but not contain their names as values. So either we should use `fn_name` caller at the end of  main function or better solution is just to move skip list.

Also, there were some comments inside the doc examples. Those were causing our function not able to find end of an example or result. While `with` was a stopping point word, I added `#` and `instance` too. This solved few cases.

I also added `abs_tol` to value cheking part because again backends have small differences.

ivy_tests/test_docstrings.py
```diff
@@ -70,6 +69,105 @@ def check_docstring_examples_run(
...
+         "dropout1d",
+         "dropout2d",
+         "dropout3d",
+         "total_mem_on_dev",
+         "supports_inplace_updates",
+         "get",
+         "deserialize",
+         "set_split_factor",
+    ]
+    # # the temp skip list consists of functions which have an issue with their
```
vedpatwardhan, line 117: minor change, additional `#` to be removed

ivy_tests/test_docstrings.py
```diff
@@ -131,6 +232,13 @@ def check_docstring_examples_run(
         if ">>> print(" in line:
             is_multiline_executable = False
 
+    # remove "..." for multilines
+    for i, v in enumerate(executable_lines):
+        executable_lines[i] = v.replace("...", "")
```
vedpatwardhan, line 237: this is already been done at line [216](https://github.com/unifyai/ivy/pull/18158/files#diff-48ffb3a283811c82d46de1000e0dbf654f102f8fbb23f5dcf9c974e21964bc90R216) right? Probably there's some bug in that which could be why the replace isn't working

ivy_tests/test_docstrings.py
```diff
@@ -131,6 +232,13 @@ def check_docstring_examples_run(
...
+        # backend which doesn't support inplace_update throws an exception
+        executable_lines[i] = v.replace(
```
vedpatwardhan, line 239: I don't think we should be having a `ensure_in_backend=True` example in the `inplace_updates` function, given that it's an expected inconsistency, or we should be able to skip the docstring tests for `jax` and `tensorflow` if we want to keep that example.

ivy_tests/test_docstrings.py
```diff
@@ -172,15 +280,31 @@ def check_docstring_examples_run(
...
     numeric_pattern = re.compile(
         r"""
-            [\{\}\(\)\[\]]|\w+:
+            [\{\}\(\)\[\]\<>]|\w+:
         """,
         re.VERBOSE,
     )
+
+    # word list to remove from execution output
+    words_to_remove = [
```
vedpatwardhan, line 292: not sure why we'd have an example that generates a non-`ivy` result (I mean `ivy.NativeArray`, `ivy.NativeShape`, etc. like `torch.Size`, `tf.Tensor`). The docstring examples should ideally generate `ivy`-specific objects (`ivy.Array`, `ivy.Shape`, etc.). So ideally there shouldn't be anything in the list `words_to_remove` except `ivy.array` and `ivy.Shape`, happy to know if there are any corner cases.

ivy_tests/test_docstrings.py
```diff
@@ -222,75 +347,25 @@ def test_docstrings(backend):
...
-        "det",
-        "cumprod",
-        "where",
-        "sinc",
-        "grad",
-    ]
 
     # skip list for array and container docstrings
     skip_arr_cont = [
+        # generates different results due to randomization
```
vedpatwardhan, line 353: It would be great if this list was also defined in the same place as the other 2 lists, just to make it easier to find

vedpatwardhan (changes requested):
Hey @sondertg, thanks for putting in the effort for all these changes 💪
Just had a few doubts about the changes to `test_docstrings.py`, happy to know what you think as well @Ishticode 😄

ivy_tests/test_docstrings.py
sondertg: Below ones fails when I remove that list and only trim "ivy.shape" and "ivy.array".: 
vmap
slogdet
svdvals

I believe this there are few more hidden cases. Good thing about having a list like that save "slogdet" test with a single element in the list. Some of them can be easily fixed by adding one word to that list.

```
output for  vmap  on run:  torch.Size([3,5,2]) 
output in docs : (3,5,2) 
 torch.Size3  !=  3 

output for  slogdet  on run:  slogdet(sign=ivy.array(1.),logabsdet=ivy.array(1.60943794)),slogdet(sign=ivy.array(-1.),logabsdet=ivy.array(1.0986129)),[{a:ivy.array(-1.),b:ivy.array(-1.)},{a:ivy.array(0.69314718),b:ivy.array(1.09861231)}] 
output in docs : slogdet(sign=ivy.array(1.),logabsdet=ivy.array(1.60943794)),slogdet(sign=ivy.array(-1.),logabsdet=ivy.array(1.098611)),[{a:ivy.array(-1.),b:ivy.array(-1.)},{a:ivy.array(0.69314718),b:ivy.array(1.09861231)}] 
 logabsdet=1.0986129  !=  logabsdet=1.098611 

output for  svdvals  on run:  ivy.Shape(2,),ivy.Shape(2,),ivy.array([0.,0.]),torch.Size([4,3]),ivy.array([10.25994301,1.16403711,0.61529785]),ivy.array([10.25994301,1.16403687,0.61529785]),ivy.array([0.39777088,0.15413129,0.1029852]),{a:ivy.array([9.01383495,0.86647373]),b:ivy.array([15.77865314,5.55970573,4.16857529,0.86412716])},ivy.array([13.37566662,3.88478065]),{a:ivy.array([12.95925522,4.64447212,2.54687858]),b:ivy.array([23.16135025,10.35037804,4.31025028,1.35769403])} 
output in docs : ivy.Shape(2,),ivy.Shape(2,),ivy.array([0.,0.]),(4,3),ivy.array([10.3,1.16,0.615]),ivy.array([10.3,1.16,0.615]),ivy.array([0.39777088,0.15413129,0.1029852]),{a:ivy.array([9.01383495,0.86647356]),b:ivy.array([15.7786541,5.55970621,4.16857576,0.86412698])},ivy.array([13.37566757,3.88477993]),{a:ivy.array([12.95925522,4.6444726,2.54687881]),b:ivy.array([23.16134834,10.35037804,4.31025076,1.35769391])} 
 torch.Size4  !=  4 
```

ivy_tests/test_docstrings.py
sondertg: this fails but when `ensure_in_backend` not explicitly set to `True` it is fine.(since default is `False`. Should I add it to skip list?
```py
ivy.set_backend("jax")
x = ivy.array([1, 2, 3])
y = ivy.array([0])
x.inplace_update(y, ensure_in_backend=True)
print(x)
```

ivy_tests/test_docstrings.py
sondertg: ```            executable_lines[-1] += line.split("...")[1][1:]```
This line removes first three dots but keep rest. I didn't change that part because it correctly extract multiline code that should be executed as single line.

sondertg:
I forgot to mention but below notes are just mine to summarize and see common problems. I didn't clear and organize them but it might be useful for skip list functions. There might be some cases that I fixed or misunderstand, so better to check if explanation still holds.
```
E             Container.quantile
only torch fails
torch: quantile: 'NoneType' object is not iterable


E             Array.cummax:
Torch return 'torch.return_types.cummax(values=ivy.array([1,2,5,5,5]),indices=ivy.array([0,1,2,2,2])),ivy.array([[0,0],[0,0],[0,0]])'
which all other backends just return numeric values, this cause failure even results are correct
added to conditional skip list

Array.l1_normalize (container and functional too)
jax numpy ivy and tensorflow torch generates different results


ivy.set_backend("tensorflow")
x = ivy.array([[1., 2.], [3., 4.]])
y = x.l1_normalize(axis=1)
print(y)


ivy.set_backend("numpy")
x = ivy.array([[1., 2.], [3., 4.]])
y = x.l1_normalize(axis=1)
print(y)


rad2deg

tensorflow require floats to run succesfully   


E             Array.inplace_update
 
 below code works fine with different outputs for different backends

import ivy

ivy.set_default_device("cpu")
ivy.set_backend("jax")

x = ivy.array([1, 2, 3])
y = ivy.array([0])
x.inplace_update(y)
print(x)

jax, numpy, tensroflow output: ivy.array([0])
torch output: ivy.array([0, 0, 0])

also:

import ivy
ivy.set_backend("torch")
x = ivy.Container(a=ivy.array([5, 6]), b=ivy.array([7, 8]))
y = ivy.Container(a=ivy.array([1]), b=ivy.array([2]))
x.inplace_update(y)
print(x)

torch output:
{
    a: ivy.array([1, 1]),
    b: ivy.array([2, 2])
}

other backends output: 
{
    a: ivy.array([1]),
    b: ivy.array([2])
}

E             Array.is_array

below example not working (without exclusive arg, it is fine)
x = ivy.native_array([9.1, -8.3, 2.8, 3.0])
print(x.is_array(exclusive=True))


E             Array.is_ivy_array
not working
x = ivy.native_array([9.1, -8.3])
ret = x.is_ivy_array(exclusive=True)
print(ret)

E             Array.is_native_array

x = ivy.native_array([9.1, -8.3])
ret = x.is_native_array(exclusive=True)
print(ret)


E             Array.lgamma

x = ivy.array([1., 2., 3.])
y = x.lgamma()
print(y)

jax result:  ivy.array([4.76837158e-07, 0.00000000e+00, 6.93147421e-01])

very close to doc result, added atol to fix it




E             Container.std
torch: std: std and var only support floating point and complex dtypes
torch error goes away if all numbers are float on examples

tensorflow:
ivy.set_backend("jax")

x = ivy.Container(a=ivy.array([[2, 1]]), b=ivy.array([[2, -2]]))
y = x.std(axis=1, keepdims=True)
print(y)

ivy.set_backend("tensorflow")

x = ivy.Container(a=ivy.array([[2, 1]]), b=ivy.array([[2, -2]]))
y = x.std(axis=1, keepdims=True)
print(y)

above two generates different results (jax, ivy and numpy results are consistent)

if passed floats, it is fine.


E             Array.where
below example not work, tried as method

x1 = ivy.array([[1.1, 2, -3.6], [5, 4, 3.1]])
x2 = ivy.Container(a=ivy.array([0, 7, 2]),b=ivy.array([3, 8, 5]))
res = ivy.where((x1.b < x2.b), x1, x2)
print(res)

as method
x1 = ivy.array([[1.1, 2, -3.6], [5, 4, 3.1]])
x2 = ivy.Container(a=ivy.array([0, 7, 2]),b=ivy.array([3, 8, 5]))
condition = x1.b < x2.b
res = x1.where(condition, x2)
print(res)

besides not all docs has mixed(container+array) example


torch: generates different results compared to jax, numpy, tensorflow and ivy for below example

x1 = ivy.Container(a=ivy.array([3, 1, 5]), b=ivy.array([2, 4, 6]))
x2 = ivy.Container(a=ivy.array([0, 7, 2]), b=ivy.array([3, 8, 5]))
condition = x1.a > x2.a
res = x1.where(condition, x2)
print(res)


E             max_unpool1d


indices parameter in the example not defined and corrected to below, still not working

x = ivy.arange(0, 24.).reshape((2, 3, 4))
pool_result, indices = ivy.max_pool1d(x, 2, 2, 'SAME')
unpool_result = ivy.max_unpool1d(pool_result, indices, 2, 2, 'SAME')
print(unpool_result)

E             pool
implementation is wrong, all examples fails

E             dct
jax fails due to unadjustable tolerances but results are correct, added conditional  skip list
tensorflow throws below error with the example:


ivy.set_backend("tensorflow")


x = ivy.array([8, 16, 24, 32, 40, 48, 56, 64])
y = ivy.dct(x, type=2, n=None, norm='ortho')
print(y)

tensorflow: idct: Cannot convert 0.0 to EagerTensor of dtype int32

E             idct

jax and tensorflow fails, same reasons with "dct"



E             rfftn

none of the backends work but if backend is not explicitly set function works
rfftn: 'NoneType' object is not iterable



E             put_along_axis

function itself not working with numpy put_along_axis doc example and ivy.put_along_axis doc example.

numpy doc example:
a = ivy.array([[10, 30, 20], [60, 40, 50]])
ai = ivy.argmax(a, axis=1, keepdims=True)
ivy.put_along_axis(a, ai, 99, 1)
print(a)

tried to pass 99 as ivy array too.


ivy doc example:
arr = ivy.array([[4, 3, 5], [1, 2, 1]])
indices = ivy.array([[0, 1, 1], [2, 0, 0]])
values = ivy.array([[9, 8, 7], [6, 5, 4]])
ivy.put_along_axis(arr, indices, values, 1, mode='clip')
print(arr)

all the input arrays must have same number of dimensions, but the array at index 0 has 2 dimension(s) and the array at index 1 has 3 dimension(s)


E             histogram
histogram supposed to return a tuple but it returns an array
tensorflow error:
IvyBackendException('tensorflow: histogram: cannot compute Less as input #1(zero-based) was expected to be a int32 tensor but is a float tensor [Op:Less] name: ')

example to gen this error:
ivy.set_backend("tensorflow")


x = ivy.array([0, 1, 2])
y = ivy.array([0., 0.5, 1., 1.5, 2.])
z = ivy.histogram(x, bins=y)
print(z)

torch container, array and functional not working

E             cov
below example not works

With :class:`ivy.Array` input with both fweights and aweights given,
and rowVar set to False:
>>> x = ivy.array([[1,2,3],
...                [4,5,6]])
>>> fw = ivy.array([1,3])
>>> aw = ivy.array([ 1.5, 4 ])
>>> y = x[0].cov(x[1], fweights=fw, aweights=aw, rowVar=False)
>>> print(y)
ivy.array([[ 1.22727273,  1.22727273, 1.22727273 ],
...        [ 1.22727273,  1.22727273, 1.22727273 ],
...        [ 1.22727273,  1.22727273, 1.22727273 ]])




E             linspace
only jax fails and with device="gpu", tensorflow and torch not working

jax example
ivy.set_backend("jax")
x = ivy.linspace(1, 10, 4, dtype = int)
print(x)

also 

ivy.set_backend("jax")
x = ivy.linspace(1, 10, 4, dtype ="int32")
print(x)

fails too.

E             native_array

tests are passing but native array representation generate different results for tensorflow and torch backend

E             logspace
jax fails with below example(like linspace), probably inplace updates is the reeason
print(ivy.logspace(1, 2, 4, dtype= int))

torch generates different result for below example:

x = ivy.array([1, 2])
y = ivy.array([4])
print(ivy.logspace(x, y, 4))

ivy.set_backend("torch")

x = ivy.array([1, 2])
y = ivy.array([4])
print(ivy.logspace(x, y, 4))

   
E             result_type

function not working as expected, all examples are wrong



E             set_split_factor
examples are working properly but different backends generate different results since set_backend also used in examples

E             function_unsupported_devices

numpy doesn't support ones with gpu unlike others






E             maximum

 torch: maximum: !needs_dynamic_casting<func_t>::check(iter) INTERNAL ASSERT FAILED at "../aten/src/ATen/native/cpu/Loops.h":310, please report a bug to PyTorch. 

ivy.set_backend("torch")
x = ivy.array([1, 5, 9, 8, 3, 7])
y = ivy.array([[9], [3], [2]])
z = ivy.zeros((3, 6))
ivy.maximum(x, y, out=z)
print(z)

E             minimum
torch: minimum: !needs_dynamic_casting<func_t>::check(iter) INTERNAL ASSERT FAILED at "../aten/src/ATen/native/cpu/Loops.h":310, please report a bug to PyTorch. 
x = ivy.array([1, 5, 9, 8, 3, 7])
y = ivy.array([[9], [3], [2]])
z = ivy.zeros((3, 6))
ivy.minimum(x, y, out=z)
print(z)

E             deg2rad

tensorflow fails

AttributeError: EagerTensor object has no attribute 'astype'. 
        If you are looking for numpy-related methods, please run the following:
        from tensorflow.python.ops.numpy_ops import np_config
        np_config.enable_numpy_behavior()
      . Did you mean: 'dtype'?

ivy.set_backend("tensorflow")
x=ivy.array([0,90,180,270,360])
y=ivy.deg2rad(x)
print(y)


E             try_else_none
z = ivy.try_else_none(ivy.add,x, y) this line below example prints exception adress
<traceback object at 0x7f24f24daf00>

x = ivy.array([1, 2, 3])
y = 'hemant'
z = ivy.try_else_none(ivy.add,x, y)
print(z)



E             cache_fn


E             einops_reduce
unknown tensor type, all examples wrong


E             set_item

jax 
Usinganon-tuplesequenceformultidimensionalindexingisnotallowed;use`arr[array(seq)]`insteadof`arr[seq]`.Seehttps://github.com/google/jax/issues/4564formoreinformation.
numpy 
shape mismatch



E             get_num_dims
bad examples

E             vmap
torch generates 
torch.Size([3,5,2]) 
rest result
(3, 5, 2)

E             value_and_grad
torch fails
torch: nested_map: element 0 of tensors does not require grad and does not have a grad_fn

example

x = ivy.array([[4.6, 2.1, 5], [2.8, 1.3, 6.2]]
func = lambda x: ivy.mean(ivy.square(x))
grad_fn = ivy.value_and_grad(func)
value_grad = grad_fn(x)
print(value_grad)


E             scaled_dot_product_attention
exec throws error "cannot convert to ivy dtype"
q = ivy.Container(a=ivy.array([[[0.2, 1.], [2.7, 3.], [4.4, 5.6]]]),
                  b=ivy.array([[[1.2, 1.], [2.2, 3.], [4.4, 5.6]]]))
k = ivy.Container(a=ivy.array([[[4.2, 1.], [2.2, 3.3],[4.4, 5.6]]]),
                  b=ivy.array([[[3.2, 1.], [2.2, 3.6], [4.0, 5.6]]]))
v = ivy.Container(a=ivy.array([[[5.2, 1.], [2.1, 3.],[4.4, 5.6]]]),
                  b=ivy.array([[[0.2, 1.], [2.2, 3.],[4.4, 5.6]]]))
mask = ivy.Container(a=ivy.array([[[1.0, 1.0, 1.0],
                            [1.0, 1.0, 1.0],
                            [1.0, 1.0,1.0]]]),
              b=ivy.array([[[1.0, 1.0, 1.0],
                            [1.0, 1.0, 1.0],
                            [1.0, 1.0,1.0]]]))
result = q.scaled_dot_product_attention(k,
                                        v,
                                        1,
                                        mask=mask)
print(result)

examples are fine if I run them

E             conv1d_transpose
tensorflow fails because "dilations" should be equal or less than 1 if device is cpu.
x = ivy.full((1, 6, 1), 2.7)
a = ivy.random_normal(mean=0, std=1, shape=[3, 1, 1])
b = ivy.random_normal(mean=0, std=1, shape=[3, 1, 1])
filters = ivy.Container(a=a, b=b)
y = ivy.conv1d_transpose(x, filters, 1, 'VALID', dilations=2)
print(y.shape)


E             conv2d_transpose
tensorflow fails because "dilations" should be equal or less than 1 if device is cpu.
x = ivy.full((1, 6, 6, 1), 2.7)
a = ivy.random_normal(mean=0, std=1, shape=[3, 3, 1, 1])
b = ivy.random_normal(mean=0, std=1, shape=[3, 3, 1, 1])
filters = ivy.Container(a=a, b=b)
y = ivy.conv2d_transpose(x, filters, 1, 'VALID', dilations=2)
print(y.shape)




E             conv3d_transpose

ivy and numpy fails
numpy: conv3d: Unable to allocate 123. GiB for an array with shape (33126506496, 1) and data type float32

x = ivy.random_normal(mean=0, std=1, shape=[1, 7, 256, 256, 64])
filters = ivy.random_normal(mean=0, std=1, shape=[3, 3, 3, 64, 32])
y = ivy.conv3d_transpose(x, filters, [1, 1, 1], 'VALID')
print(y.shape)


E             eigvalsh
jax version require high tolerance
x = ivy.native_array([[[1., 1., 2.], [1., 2., 1.], [1., 1., 2]]])
y = ivy.eigvalsh(x)
print(y)

ivy.set_backend("jax")
x = ivy.native_array([[[1., 1., 2.], [1., 2., 1.], [1., 1., 2]]])
y = ivy.eigvalsh(x)
print(y)

torch also generate different results for below example
ivy.set_backend("torch")

x = ivy.Container(a=ivy.array([[[1.,2.,3.],[2.,4.,5.],[3.,5.,6.]]]),
                     b=ivy.array([[[1.,1.,2.],[1.,2.,1.],[2.,1.,1.]]]),
                     c=ivy.array([[[2.,2.,2.],[2.,3.,3.],[2.,3.,3.]]]))
y = ivy.eigvalsh(x)
print(y)



E             vector_norm
fails in torch for below example:
x = ivy.array([1,2,3,4], dtype = ivy.float16)
z = ivy.empty(shape = 1)
y = ivy.vector_norm(x, ord = 0, out = z)
print(y)

probably because of dtype not supported



E             nonzero
container methods not working with below kind example. (dimension number difference )

x = ivy.Container(a=ivy.array([0,1,2,3,0]), b=ivy.array([[1,1], [0,0]]))
y = ivy.nonzero(x)
print(y)


E             searchsorted
exec() fails dtype error

E             mean
torch fails when array contains int only 
```

ivy_tests/test_docstrings.py
vedpatwardhan: I think given that this inconsistency is something we've already communicated in the [docs](https://unify.ai/docs/ivy/overview/deep_dive/inplace_updates.html#inplace-updates) I think we could probably avoid using examples with `ensure_in_backend=True` in the docstrings 😄

ivy_tests/test_docstrings.py
vedpatwardhan: I think we should be fixing the implementations of `vmap` and `svdvals` to ensure that native arrays and shapes aren't returned and we return `ivy` arrays and shapes instead (probably a fix to the function wrappers for the 2 functions). As for `slogdet`, we could either pick a simpler example in the docstring or update the tolerance for testing that function rather than skipping the tests for these functions entirely. 
Feel free to point out other functions where such issues exist, we should fix all of them on a case-by-case basis 😄

vedpatwardhan:
@Ishticode could you please verify the explanations which @sondertg shared? Thanks 😄

ivy_tests/test_docstrings.py
sondertg: Problem is not tolerance, since there are strings on there, numeric checks fails. When I just trim "logabsdet=" part, it passes tests. For fixing implementations, aren't those out of the scope of this PR? Maybe after merging this as it is, I can go one by one for failures require deeper investigation as another task or please guide about which wrappers to look at.

ivy_tests/test_docstrings.py
sondertg: Just removed that part.

vedpatwardhan (approved):
The changes made to `test_docstrings.py` are good to merge. Thanks @sondertg 😄

Ishticode (approved):

PR #18164, opened 2023-07-02
Stateful changes

RickSanchezStoic:
Fixes bug with the ivy.Module not propagating weights to child modules upon initialization, and fixed a bug with container methods set at key chains and overwrite at key chains, not working correctly for inplace = True


ivy/data_classes/container/base.py
```diff
@@ -2728,7 +2728,7 @@ def cont_set_at_key_chains(self, target_dict, return_dict=None, inplace=False):
                 return_dict[k] = self.cont_set_at_key_chains(v, return_dict[k], inplace)
             else:
                 return_dict[k] = v
-        return ivy.Container(return_dict, **self._config)
+        return return_dict if inplace else ivy.Container(return_dict, **self._config)
```
vedpatwardhan, line 2731: I don't think this line makes a difference in the behaviour of the `inplace` argument, as the `return_dict` is always a `Container`. 

I'm not sure why we need the `ivy.Container(return_dict, **self.config)` in the first place, as we're retaining the config even when we do at line [`2725`](https://github.com/unifyai/ivy/pull/18164/files#diff-d15a6b35d36e632e8493df8bdfbcf9a027cb4599dccac4f41ca077788892fdc2R2725), and that's the only possible use of calling the `ivy.Container.__init__` with `return_dict` being a container (i.e. if the `config` of `return_dict` wasn't correct after calling `cont_copy` but that isn't the case). 
@saeedashrraf any ideas?

In the meantime, could you please revert this change @RickSanchezStoic 😅?

ivy/data_classes/container/base.py
```diff
@@ -2771,7 +2771,7 @@ def cont_overwrite_at_key_chains(
                 )
             else:
                 return_dict[k] = v
-        return ivy.Container(return_dict, **self._config)
+        return return_dict if inplace else ivy.Container(return_dict, **self._config)
```
vedpatwardhan, line 2774: Could you please also revert this change for the same reason?

ivy/stateful/module.py
```diff
@@ -593,7 +593,9 @@ def build(
                 ivy.Container.cont_assert_identical_structure(
                     [created_n_found, v_from_constructor]
                 )
-                self.v = v_from_constructor
+                # propagate v_from_constructor to child variables
+                created_n_found.cont_set_at_key_chains(v_from_constructor, inplace=True)
```
vedpatwardhan, line 597: I think we can instead use `cont_inplace_update` here instead because we know that the structure of `v_from_constructor` and `created_n_found` are the same till we reach this point, and hence we don't need the handling used to update specified keychains from the `target_dict` in the `created_n_found` container (as all key_chains need to be updated).

vedpatwardhan (changes requested):
Hey @RickSanchezStoic, thanks for diving into this issue 💪
Just requested a few changes, but I think the issue is still not resolved even after making these changes, because we still have 2 copies of the variables container at this [line](https://github.com/unifyai/ivy/pull/18164/files#diff-9d3cd8360ac6368fb7cfebdd75ef04230b2963749af122e8515e44ea6ddf3971R578), meaning we would still end up consuming double the memory when input weights are passed by users. 
We'll need to explore other ways to deal with this 😅

vedpatwardhan:
Also, let's add the changes to the `Dockerfile` in a different PR/commit for clarity 🙂

RickSanchezStoic:
> Hey @RickSanchezStoic, thanks for diving into this issue 💪 Just requested a few changes, but I think the issue is still not resolved even after making these changes, because we still have 2 copies of the variables container at this [line](https://github.com/unifyai/ivy/pull/18164/files#diff-9d3cd8360ac6368fb7cfebdd75ef04230b2963749af122e8515e44ea6ddf3971R578), meaning we would still end up consuming double the memory when input weights are passed by users. We'll need to explore other ways to deal with this 😅

That's intended behavior. This is what the original design was for `with_partial_v`, where we update the internal variables to match the variables passed to the constructor. For the parent class to gain information about the child classes, it has to be initialized, initializing them would cause weights to be generated ( at least temporarily in this case), these weights are then updated to match the `v` passed to the constructor. 

The original issue wasn't that the internal weights are being generated for the module say we do `Net()`. The original issue was that these weights are retained even after we pass the weights using `v`. This is not the case after this change. Hope this makes sense @vedpatwardhan

ivy/data_classes/container/base.py
RickSanchezStoic: It does make a difference because even for `inplace=True` a new container is returned with the **config added. Earlier, I too just added a `return return_dict` but since wasn't sure about the config thing so I omitted it. This indeed is a bug since it doesn't update the code inplace if `inplace=True`. Let me know if you have more doubts @vedpatwardhan

ivy/stateful/module.py
RickSanchezStoic: Just went through its implementation, actually, `cont_set_at_key_chains` or `cont_overwrite_at_keychains` would be the fastest and most appropriate for this case, given its implementation. Since we already confirm the structure of both of these containers to be the same, these functions are designed to quickly update the values, these functions were designed for this specific reason.

vedpatwardhan:
> > Hey @RickSanchezStoic, thanks for diving into this issue 💪 Just requested a few changes, but I think the issue is still not resolved even after making these changes, because we still have 2 copies of the variables container at this [line](https://github.com/unifyai/ivy/pull/18164/files#diff-9d3cd8360ac6368fb7cfebdd75ef04230b2963749af122e8515e44ea6ddf3971R578), meaning we would still end up consuming double the memory when input weights are passed by users. We'll need to explore other ways to deal with this 😅
> 
> That's intended behavior. This is what the original design was for `with_partial_v`, where we update the internal variables to match the variables passed to the constructor. For the parent class to gain information about the child classes, it has to be initialized, initializing them would cause weights to be generated ( at least temporarily in this case), these weights are then updated to match the `v` passed to the constructor.
> 
> The original issue wasn't that the internal weights are being generated for the module say we do `Net()`. The original issue was that these weights are retained even after we pass the weights using `v`. This is not the case after this change. Hope this makes sense @vedpatwardhan

Hey @RickSanchezStoic, thanks for the quick response. Yeah it should get deleted eventually, but if at any particular instance we have 2 copies of variables, the peak memory consumption would be double what it should be, and we'd therefore run out of memory for large models. I think unless we fix the change to avoid random initialization, this won't be fixed 🙂

ivy/data_classes/container/base.py
RickSanchezStoic: That's what we are doing right? We need to pass the values then to `self.v`  like `self.v=created_n_found` , so better do this in one line. @vedpatwardhan

saeedashrraf: I think just to ensure that the function always returns an `ivy.Container`, we should maybe change [2731](https://github.com/unifyai/ivy/pull/18164/files#diff-d15a6b35d36e632e8493df8bdfbcf9a027cb4599dccac4f41ca077788892fdc2R2731)  to be `return return_dict if inplace and isinstance(return_dict, ivy.Container) else ivy.Container(return_dict, **self._config)` . 

Let me know what you think

saeedashrraf: Actually, no need to check for `inplace`, we can just change it to be `return return_dict if isinstance(return_dict, ivy.Container) else ivy.Container(return_dict, **self._config)`

RickSanchezStoic: > Actually, no need to check for `inplace`, we can just change it to be `return return_dict if isinstance(return_dict, ivy.Container) else ivy.Container(return_dict, **self._config)`

This sounds good, what do you say @vedpatwardhan

vedpatwardhan: > Actually, no need to check for `inplace`, we can just change it to be `return return_dict if isinstance(return_dict, ivy.Container) else ivy.Container(return_dict, **self._config)`

Yeah this sounds perfect, let's update `set_at_key_chains` and other relevant instance methods with this handling @RickSanchezStoic, thanks @saeedashrraf 🙂

ivy-seed:

ivy-seed:

PR #18220, opened 2023-07-03
add asarray fix

aparajith21:


vedpatwardhan (commented):
Hey @AnnaTz, could you please take a look? Given that the `current_backend` call in `asarray` was updated because it consumed a lot of time, probably we should think about the same factor before adding it back. Thanks @aparajith21 and @AnnaTz 🙂

vedpatwardhan:
> > Hey @AnnaTz, could you please take a look? Given that the `current_backend` call in `asarray` was updated because it consumed a lot of time, probably we should think about the same factor before adding it back. Thanks @aparajith21 and @AnnaTz 🙂
> 
> There was a use case of `jax.lax.reduce_window` that took forever to run because the implementation includes converting a very long list to an array and `_determine_backend_from_args` was going through all items in the list one by one.

Ah right, I think we can then update `current_backend` so that it doesn't convert a `list` to an `ivy.Array`, because that doesn't really add any value to `current_backend`. @aparajith21 feel free to start working on those changes if you'd like to, certainly not top of the list to fix at the moment. Thanks for the helpful explanation @AnnaTz 🙂

AnnaTz:
> > > Hey @AnnaTz, could you please take a look? Given that the `current_backend` call in `asarray` was updated because it consumed a lot of time, probably we should think about the same factor before adding it back. Thanks @aparajith21 and @AnnaTz 🙂
> > 
> > 
> > There was a use case of `jax.lax.reduce_window` that took forever to run because the implementation includes converting a very long list to an array and `_determine_backend_from_args` was going through all items in the list one by one.
> 
> Ah right, I think we can then update `current_backend` so that it doesn't convert a `list` to an `ivy.Array`, because that doesn't really add any value to `current_backend`. @aparajith21 feel free to start working on those changes if you'd like to, certainly not super high priority at the moment. Thanks for the helpful explanation @AnnaTz 🙂

I think I wasn't very clear 😅. The implementation of `jax.lax.reduce_window` includes calling `ivy.asarray` on a very long list. Then `_determine_backend_from_args` goes through all items in each argument which means it will try to iterate through that very long list. I'm not sure we can avoid that.

vedpatwardhan:
> I think I wasn't very clear 😅. The implementation of `jax.lax.reduce_window` includes calling `ivy.asarray` on a very long list. Then `_determine_backend_from_args` goes through all items in each argument which means it will try to iterate through that list. I'm not sure we can avoid that.

Yeah but ig the `list` being passed from [`reduce_window`](https://github.com/unifyai/ivy/blob/09d2d3a98b2b1100748619826fa3d1e4b562f495/ivy/functional/ivy/experimental/layers.py#L2227) to `asarray` is a `list` of `Number` rather than `ivy.[Native]Array` (i.e. ig none of the elements in the `list` are `ivy.[Native]Array`). So basically in your case, the logic in `current_backend` wouldn't infer a backend anyway (even after iterating the full list), which is why I was thinking we speed up the process by not calling [`_determine_backend_from_args`](https://github.com/unifyai/ivy/blob/09d2d3a98b2b1100748619826fa3d1e4b562f495/ivy/utils/backend/handler.py#L112) for every element, but only after we've checked that that element is an `ivy.[Native]Array`. Using `filter` and stuff, ig we can considerably improve the time consumed 😅
Happy to know your thoughts, thanks @AnnaTz @aparajith21 🙂

fabriziocacicia:
Could this https://github.com/unifyai/ivy/pull/15772 have changed anything?

AnnaTz:
> Could this #15772 have changed anything?

No, just checked and the problem is still there. E.g. `ivy.asarray([0]*10**6)` takes forever.

ivy-seed:

ivy-seed:

PR #18247, opened 2023-07-03
Feature: Check if default device exists

sladyn98:
Test if the incoming device object or string representation points to a real device not an arbitrary reference.


vedpatwardhan (commented):
Hey @sladyn98! 
Could you please look into the causes of [this](https://github.com/unifyai/ivy/actions/runs/5504513962/jobs/10031231898?pr=18247#step:3:112596) error in the CI? Also pinging @abdulasiraj as he's working on fixing a similar problem.
Thanks 🙂

sladyn98:
@vedpatwardhan I am not able to quite understand why the test is still failing. Some help would be good here : 0

ivy/functional/backends/jax/device.py
```diff
@@ -75,13 +75,22 @@ def _to_device(x, device=None):
...
-    return ivy.Device(p + ":" + str(dev_id))
+    else:
+        p, dev_id = (device.platform, device.id)
+        if p == "cpu":
+            device_object = ivy.Device(p)
+        else:
+            device_object = ivy.Device(p + ":" + str(dev_id))
+    
+    # Verify if the device exists on the host
+    available_devices = jax.local_devices()
```
vedpatwardhan, line 89: Actually, the issue here is that `jax.local_devices()` returns `jax` native devices (e.g. `CpuDevice(id=0)` while we're trying to compare it against an `ivy.Device` which was created on line [86](https://github.com/unifyai/ivy/pull/18247/files#diff-2b01bb1c1257260fd336443a518ea5713550f12a81ba35ff6ff08414f7e6f3c8R86) (and which subclasses `str`). As a result, the result of `str(device_object) not in available_devices` is always `False` leading to [this](https://github.com/unifyai/ivy/actions/runs/5535639911/jobs/10106420561?pr=18247#step:3:31) error.
Basically we'd need to convert the `available_devices` list to instead have elements of the form `f'{device.platform}:{device.id}` if `device.platform` isn't a `cpu` and use `f'{device.platform}'` otherwise,
```
available_devices = [f'{device.platform}:{device.id}' if device.platform != 'cpu' else device.platform for device in available_devices]
```
Once we've done that, we'd be able to correctly do a `str(device_object) not in available_devices` 
Thanks @sladyn98 😄👍

vedpatwardhan (changes requested):

abdulasiraj:
Hi @vedpatwardhan 
I've completed the device checking, along with handling the fall back. can you mind taking another look?
Thanks!

abdulasiraj:
> I think there's some confusion regarding the task itself, because now that I try,
> 
> ```
> 
> import torch
> 
> 
> dev = torch.device('cuda:142')
> x = torch.tensor([1., 2., 3.]).to(dev)
> print(dev)
> ```
> 
> This throws an error. @Dsantra92 could you please elaborate on this issue and provide a test example that is expected to work with all backends? Thanks @sladyn98 @abdulasiraj 😄

Makes sense, I also asked him on discord for example.

Dsantra92:
Hi @abdulasiraj and @vedpatwardhan looks the example is no longer reproducible. But this is still a thing with TensorFlow:
```python
with tf.device("/GPU:3"):
    x = tf.constant([1., 2, 3])
    x.device
```
I think this example more sense at this point. I will update the card accordingly. Given that torch doesn't support this anymore with new version (I am still figuring out where this changed) it's best we do this arbitrary device handling to torch as well.

vedpatwardhan:
> Hi @abdulasiraj and @vedpatwardhan looks the example is no longer reproducible. But this is still a thing with TensorFlow:
> 
> ```python
> with tf.device("/GPU:3"):
>     x = tf.constant([1., 2, 3])
>     x.device
> ```
> 
> I think this example more sense at this point. I will update the card accordingly. Given that torch doesn't support this anymore with new version (I am still figuring out where this changed) it's best we do this arbitrary device handling to torch as well.

Yeah I agree, we should handle this in the `torch` backend as well as a part of superset behaviour. Feel free to make a start on this already @sladyn98 @abdulasiraj, thanks 😄

sladyn98:
@vedpatwardhan Thanks, should we merge this PR first or get that fixed first. I am not sure exactly where in the code I need to make those changes so @abdulasiraj if we could collaborate that would be great.

vedpatwardhan:
> @vedpatwardhan Thanks, should we merge this PR first or get that fixed first. I am not sure exactly where in the code I need to make those changes so @abdulasiraj if we could collaborate that would be great.

I'd say given that both the backends are to be changed to achieve the same objective (i.e. fallback when an invalid device is passed) we could probably aim to get both done in this PR. Thanks 😄

sladyn98:
@vedpatwardhan Sure that makes sense could you point to me in the code where where this change needs to be done I am a bit new to the codebase still

abdulasiraj:
> @vedpatwardhan Sure that makes sense could you point to me in the code where where this change needs to be done I am a bit new to the codebase still

I'm working on it. no worries

sladyn98:
Thanks

Ishticode:
any updates on this @abdulasiraj @vedpatwardhan

vedpatwardhan:
I don't think there has been any progress on this yet, could you please take up these changes and help get this over the line @Ishticode? No rush, feel free to do that whenever you're free, thanks 😄

ivy-seed:

ivy-seed:

PR #18273, opened 2023-07-04
add torch.masked_fill frontend

Sam-Armstrong:
Add frontend function for torch.masked_fill. This function is rarely used and very poorly documented (as the instance method torch.Tensor.masked_fill is usually used instead), but it is required for transpiling some models, such as Longformer. 

Also fixed torch 'where' backend function so it can handle non-bool conditions the same as other frameworks (functional tests still pass).


AnnaTz:
If the test passes, feel free to merge 🙂

PR #18335, opened 2023-07-05
added sgn instance method for paddle frontend

### LINKED ISSUE #18239: sgn

mosesdaudu001:
close #18239


tomatillos:
Hi Moses, I think we should be testing for complex types as well as floats.

PR #18336, opened 2023-07-05
do not throw exception when checking int-inf compatibility

Sam-Armstrong:
Changed the check_fill_value_and_dtype_are_compatible assertion to not throw for inf fill_value and int dtype. This was causing LongformerModel to fail transpilation, due to function calls like this:

https://github.com/huggingface/transformers/blob/v4.30.0/src/transformers/models/longformer/modeling_longformer.py#L810


vedpatwardhan (approved):
lgtm! Feel free to merge, thanks @Sam-Armstrong, great catch!

PR #18343, opened 2023-07-05
added sgn instance method for paddle frontend and also corrected the …

### LINKED ISSUE #18239: sgn

mosesdaudu001:
close #18239


tomatillos:
Does ivy.sign not cover the complex case too? It would be cleaner to just use that

PR #18345, opened 2023-07-05
Fix paddle.cumprod by differently handling the cases of int and float unsupported output dtypes

fabriziocacicia:


fabriziocacicia:
Small note: the tests for the function were passing even before this changes. I made them because this issue was causing the `torch` frontend function failing with the `paddle` backend.
Further investigation make me think that the backend test for the `cumprod` function are not exhaustive. I'm working on another PR that make the tests more exhaustive and they are in fact failing now.

ivy-seed:

PR #18355, opened 2023-07-05
frontend/tensorflow/linalg.py eigvals function

jlmaurora233:
Completed the eigvals function and its corresponding test function. Delayed by `ImportError` in `ivy_tests`

### DIFF
--- a/ivy/functional/frontends/tensorflow/linalg.py
+++ b/ivy/functional/frontends/tensorflow/linalg.py
@@ -31,6 +31,13 @@ def eigh(tensor, name=None):
 
 
 @to_ivy_arrays_and_back
+def eigvals(tensor, tol=None):
+    x = tensor.eig()
+    eigval = x['eigenvalues']
+    return ivy.astype(ivy.matrix_rank(eigval, atol=tol), ivy.complex64)
+
+
+@to_ivy_arrays_and_back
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/tensorflow/linalg.py
```diff
@@ -30,6 +30,13 @@ def eigh(tensor, name=None):
     return ivy.eigh(tensor)
 
 
+@to_ivy_arrays_and_back
+def eigvals(tensor, tol=None):
+    x = tensor.eig()
```
hirwa-nshuti, line 35: Can we use `ivy.eighvals` as we have the function implemented on the functional API

hirwa-nshuti (changes requested):
Hey @jlmaurora233 

Thanks for the PR😊
There some changes requested in the comments below,
also add the issue number of the subtask to the first comment of the PR to be linked with the PR, you can do this by `close #issue_number`

Thanks

ivy-seed:

ivy/functional/frontends/tensorflow/linalg.py
jlmaurora233: Hi, I believe that we can do that now. The reason I didn't do that is probably because when I worked on this function, only ivy.eig is available in the documentation. Should I change it?

PR #18357, opened 2023-07-05
adding size attribute to ndarray

### LINKED ISSUE #18354: [Feature Request]: size attribute of numpy ndarray
**Is your feature request related to a problem? Please describe.**
size attribute needs to be added to ndarray class of numpy frontend

ShreyanshBardia:
closes #18354


sladyn98:
Can you add a test for this ?

ShreyanshBardia:
I think there's an issue with size attribute's logic in ivy.array
```
import ivy

ivy.array(1).size 
## 0
```
Is this the expected behaviour? Is 0 dimensional array supposed to be of size 0 and not 1? If yes then this deviates from the numpy standard. Below is the corresponding code from the array class

```
    def size(self) -> Optional[int]:
        """Number of elements in the array."""
        if self._size is None:
            self._size = (
                functools.reduce(mul, self._data.shape)
                if len(self._data.shape) > 0
                else 0
            )
        return self._size 

    def __len__(self):
        if not len(self._data.shape):
            return 0
        try:
            return len(self._data)
        except TypeError:
            return self._data.shape[0]
```

ivy-seed:

ShreyanshBardia:
Hi @AnnaTz, can you take a look at this.

sagewhocodes:
@ShreyanshBardia iirc, 0 dimensional arrays to have the shape 0 is the standard across all the frameworks.

ShreyanshBardia:
@sagewhocodes I think Numpy differs

```
>>> import numpy as np
>>> np.array(5).size
1
>>> np.array(5).ndim
0

```

sagewhocodes:
I see, I just tested it and the way `numpy` and  `jax`  define `x.size` as  the number of elements in the array whereas torch defines `x.size`  as an alias to `x.shape`. ivy follows torch in this regard.  Since you are implementing a frontend task, it should ideally mimic the behaviour of the frontend. In that case your implementation looks good to me. 
Maybe @AnnaTz can  confirm if it's the ideal behaviour.

References: 
SO Answer: https://stackoverflow.com/a/44805136
github issue : https://github.com/pytorch/pytorch/issues/5544

ShreyanshBardia:
I am not sure if ivy follows torch, because the behaviour seems to mimic numpy only. As you mentioned numpy returns an int instead of shape and ivy seems to do the same

```
>>> arr_2d = [[1,2],[3,4]]
>>> import torch, ivy
>>> ivy.array(arr_2d).size
4
>>> torch.Tensor(arr_2d).size()
torch.Size([2, 2])
```

sagewhocodes:
@ShreyanshBardia I actually meant ivy follows torch for the `.size()` function

sagewhocodes:
@ShreyanshBardia the highlighted issue has been fixed in #19680 by Ved to align with `np.ndarray.size` and `torch.numel`. I think your frontend logic would get much simpler now. Again, thank you for flagging this issue.

ShreyanshBardia:
@sagewhocodes updated it

sagewhocodes:
@ShreyanshBardia the test is failing, I think you haven't included changes from #19680 . Could you please do that and make sure the test passes?

ShreyanshBardia:
@sagewhocodes  Apologies for that, I hope this is fine now.

sagewhocodes:
LGTM!

PR #18471, opened 2023-07-06
make stack traces configurable

abdulasiraj:
closes https://trello.com/c/VXW7SxxO


abdulasiraj:
This PR is not ready for review.

ivy/utils/exceptions.py
```diff
@@ -4,79 +4,48 @@
...
-
-sys.excepthook = _custom_exception_handle
-
+    show_wrappers = ivy.show_func_wrapper_trace_mode
+    while 1:
+        if not tb.tb_next:
+            break
+        frame = tb.tb_next.tb_frame
+        if trace_mode == "ivy":
+            if "ivy/functional/ivy/" in frame.f_code.co_filename:
```
vedpatwardhan, line 30: can we instead use `os.path.join('ivy', 'functional', 'ivy')` instead? Different OS have different delimiters (`/`, `\`) and this would be more general.
Also given that we're using the paths in later checks in this function, we can probably also refactor it to compute `os.path.join` once and use it everywhere.

ivy/utils/exceptions.py
```diff
@@ -4,79 +4,48 @@
...
+                tb = tb.tb_next
+            else:
+                tb.tb_next = tb.tb_next.tb_next
+        elif trace_mode == "frontend":
+            if "ivy/functional/frontends/" in frame.f_code.co_filename or "ivy/functional/ivy/" in frame.f_code.co_filename:
```
vedpatwardhan, line 35: Also, we can refactor the logic of setting the `tb.tb_next` attribute as a helper which all 2 checks would use, and the only think that would vary is the path to be checked which would be an argument to that helper

vedpatwardhan (changes requested):
Hey @abdulasiraj, thanks for diving into this! Just requested a couple of changes 🙂

vedpatwardhan:
Also, it would be great if `configure_stack_trace` was made a private function, I don't think users would need to interact with it directly if they set the correct mode, thanks!

abdulasiraj:
I've refactor everything you want. They were all best practices and great suggestions.
Thanks @vedpatwardhan

vedpatwardhan (approved):
lgtm! There's quite a few failures in the CI for this PR, but the failures don't seem to be related to the changes made., let's see what @juliagsy thinks. I'll still request @Aarsh2001's review as some of the array api tests check the type of exception thrown and stuff, so would be great to be sure that the PR doesn't cause failures in those tests. Thanks @abdulasiraj 🙂

juliagsy (approved):
lgtm! thanks!

PR #18948, opened 2023-07-07
Add __iter__ method to Sequential

tomatillos:
Small quality of life improvement to allow you to iterate over a sequential container e.g. for loops. Also updated the docstring for the forward pass as it was previously the docstring for Linear I think.


vedpatwardhan (approved):
lgtm! Feel free to merge @tomatillos, this should make the `Sequential` module much more usable 🚀

PR #18983, opened 2023-07-08
Trello

ndurumo254:
close https://trello.com/c/Tog2x79Q/1773-update-the-unsupporteddtypes-documentation


AnnaTz:
Hi @ndurumo254, thanks for the PR but I don't think that is the way the `unsupported_dtypes` attribute is supposed to be used. I think it would be best to leave this trello task to someone with more expertise on ivy framework related issues.

ndurumo254:
> Hi @ndurumo254, thanks for the PR but I don't think that is the way the `unsupported_dtypes` attribute is supposed to be used. I think it would be best to leave this trello task to someone with more expertise on ivy framework related issues.
sure

PR #19148, opened 2023-07-10
fix(frontend-testing): Simplifies the conversion of frontend_ret to frontend_ret_np_flat and fixes it in case the return is a list.

AnnaTz:


ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -768,15 +768,12 @@ def arrays_to_numpy(x):
         frontend_fw = importlib.import_module(module_name)
         frontend_ret = frontend_fw.__dict__[fn_name](*args_frontend, **kwargs_frontend)
 
-        if ivy.isscalar(frontend_ret):
-            frontend_ret_np_flat = [np.asarray(frontend_ret)]
+        if frontend == "tensorflow" and isinstance(frontend_ret, tf.TensorShape):
+            frontend_ret_np_flat = [np.asarray(frontend_ret, dtype=np.int32)]
```
CatB1t, line 772: Why is this not the case anymore for scalar values?

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -1492,6 +1489,10 @@ def test_frontend_method(
         shallow=False,
     )
 
+    create_frontend_array = importlib.import_module(
```
CatB1t, line 1492: Why is creation frontend function needed here?

CatB1t (commented):
Hello @AnnaTz, just few questions regarding the details of those additions. Thanks!

ivy_tests/test_ivy/helpers/function_testing.py
AnnaTz: The scalar case is being handled through the flatten_frontend_to_np call.

ivy_tests/test_ivy/helpers/function_testing.py
AnnaTz: Because it needs to be passed to `flatten_frontend_to_np` below.

CatB1t (commented):
Hello @AnnaTz , could you kindly fix the merge conflicts and re-request a review? I'll have a look right after!

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -813,13 +813,9 @@ def arrays_to_numpy(x):
     if frontend_config.isscalar(frontend_ret):
         frontend_ret_np_flat = [frontend_config.to_numpy(frontend_ret)]
     else:
-        # tuplify the frontend return
-        if not isinstance(frontend_ret, tuple):
-            frontend_ret = (frontend_ret,)
-        frontend_ret_idxs = ivy.nested_argwhere(
-            frontend_ret, frontend_config.is_native_array
+        frontend_ret_flat = flatten_frontend(
+            ret=ret, backend=backend_to_test, frontend_array_fn=create_frontend_array
```
CatB1t, line 817: I think we shouldn't rely on using `create_frontend_array` as not all frontends have it, also we should modify flatten_frontend to use the frontend config in that case. similarly for `test_frontend_method`.

CatB1t (changes requested):

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -1828,8 +1815,11 @@ def flatten(*, backend: str, ret):
     return ret_flat
 
 
-def flatten_frontend(*, ret, backend: str, frontend_array_fn=None):
+def flatten_frontend(*, ret, backend: str, frontend_array_fn=None, frontend_config=None):
```
CatB1t, line 1818: Is the `frontend_config` required?

CatB1t (changes requested):

ivy_tests/test_ivy/helpers/function_testing.py
AnnaTz: It's to handle scalars in case there is no `frontend_array_fn`.

AnnaTz: No actually. I have removed it.

PR #19154, opened 2023-07-10
Onnx frontend

mosesdaudu001:
**onnx frontend implementation.** 

Created 5 operators in onnx/tensor.py
Tested the 5 operators in ivy_tests/test_ivy/test_frontends/test_onnx/test_tensor.py

Created 5 methods in onnx/elementwise.py
Tested and commented out the 5 methods in  ivy_tests/test_ivy/test_frontends/test_onnx/test_elementwise.py
Wrote 5 unconventional tests that passed.

### DIFF
--- a/ivy/functional/frontends/onnx/func_wrapper.py
+++ b/ivy/functional/frontends/onnx/func_wrapper.py
@@ -1,0 +1,91 @@
+import functools
+from typing import Callable
+
+import ivy
+import ivy.functional.frontends.onnx as onnx_frontend
+
+
+def _from_ivy_array_to_onnx_frontend_tensor(x, nested=False, include_derived=None):
+    if nested:
+        return ivy.nested_map(
+            x, _from_ivy_array_to_onnx_frontend_tensor, include_derived, shallow=False
+        )
+    elif isinstance(x, ivy.Array) or ivy.is_native_array(x):
+        a = onnx_frontend.Tensor(x)
+        return a
+    return x
+
+
+def _ivy_array_to_onnx(x):
+    if isinstance(x, ivy.Array) or ivy.is_native_array(x):
+        return onnx_frontend.Tensor(x)
+    return x
+
+
+def _onnx_frontend_array_to_ivy(x):
+    if hasattr(x, "ivy_array"):
+        return x.ivy_array
+    return x
+
+
+def _native_to_ivy_array(x):
+    if isinstance(x, ivy.NativeArray):
+        return ivy.array(x)
+    return x
+
+
+def _to_ivy_array(x):
+    return _onnx_frontend_array_to_ivy(_native_to_ivy_array(x))
+
+
+def inputs_to_ivy_arrays(fn: Callable) -> Callable:
+    @functools.wraps(fn)
+    def _inputs_to_ivy_arrays_onnx(*args, **kwargs):
+        """
+        Convert `Tensor` into `ivy.Array` instances.
+
+        Convert all `Tensor` instances in both the positional and
+        keyword arguments into `ivy.Array` instances, and then calls the
+        function with the updated arguments.
+        """
+        # convert all arrays in the inputs to ivy.Array instances
+        new_args = ivy.nested_map(
+            args, _to_ivy_array, include_derived={tuple: True}, shallow=False
+        )
+        new_kwargs = ivy.nested_map(
+            kwargs, _to_ivy_array, include_derived={tuple: True}, shallow=False
+        )
+        return fn(*new_args, **new_kwargs)
+
+    return _inputs_to_ivy_arrays_onnx
+
+
+def outputs_to_frontend_arrays(fn: Callable) -> Callable:
+    @functools.wraps(fn)
+    def _outputs_to_frontend_arrays_onnx(*args, **kwargs):
+        """
+        Convert `ivy.Array` into `Tensor` instances.
+
+        Call the function, and then converts all `ivy.Array` instances
+        returned by the function into `Tensor` instances.
+        """
+        # call unmodified function
+        ret = fn(*args, **kwargs)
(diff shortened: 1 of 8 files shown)


ivy/functional/frontends/onnx/func_wrapper.py
```diff
@@ -0,0 +1,91 @@
...
+
+def _native_to_ivy_array(x):
+    if isinstance(x, ivy.NativeArray):
+        return ivy.array(x)
+    return x
+
+
+def _to_ivy_array(x):
+    return _onnx_frontend_array_to_ivy(_native_to_ivy_array(x))
+
```
AnnaTz, line 39: I think defining ` _onnx_frontend_array_to_ivy` and `_native_to_ivy_array` just to use them here is a bit of an overkill 😅

ivy/functional/frontends/onnx/func_wrapper.py
```diff
@@ -0,0 +1,91 @@
...
+    @functools.wraps(fn)
+    def _outputs_to_frontend_arrays_onnx(*args, **kwargs):
+        """
+        Convert `ivy.Array` into `Tensor` instances.
+
+        Call the function, and then converts all `ivy.Array` instances
+        returned by the function into `Tensor` instances.
+        """
+        # call unmodified function
+        ret = fn(*args, **kwargs)
```
AnnaTz, line 73: The rest of the frontends need some handling here with regards to the default dtypes. Was this not needed for onnx?

ivy_tests/test_ivy/test_frontends/test_onnx/test_elementwise.py
```diff
@@ -0,0 +1,227 @@
...
+#     helpers.test_frontend_function(
+#         input_dtypes=input_dtype,
+#         frontend=frontend,
+#         test_flags=test_flags,
+#         fn_tree=fn_tree,
+#         on_device=on_device,
+#         input=x[0],
+#     )
+
+
```
AnnaTz, line 150: Could you try testing with `@given` like you did in `test_tensor.py`? 
If it works we should do the same for all the new frontends now that I think about it 😅

ivy_tests/test_ivy/test_frontends/test_onnx/test_elementwise.py
```diff
@@ -0,0 +1,227 @@
...
+    onnx_abs = onnx.abs(x)
+    torch_abs = torch.abs(x)
+    jax_abs = jax.lax.abs(x)
+    paddle_abs = paddle.abs(x)
+    tf_abs = tf.abs(x)
+
+    all_list = [onnx_abs, jax_abs, paddle_abs, tf_abs]
+
+    assert torch.allclose(torch_abs, all_list)
```
AnnaTz, line 162: There is no point in comparing the results of different frontends. What we want is to check that the onnx frontend's returns setting different backends is the same as the return of the original onnx function.

AnnaTz (changes requested):

ivy/functional/frontends/onnx/func_wrapper.py
mosesdaudu001: I was not sure about that as well, because the onnx frontend should be supporting whatever dtypes the framework it's currently working with supports

AnnaTz: Right, makes sense. Let's leave this as is then and we'll come back if we notice any failures while testing.

ivy_tests/test_ivy/test_frontends/test_onnx/test_elementwise.py
```diff
@@ -0,0 +1,227 @@
...
+
```
AnnaTz, line 163: Also, instead of using `torch.allclose` I would suggest using [`value_test`](https://github.com/unifyai/ivy/blob/c1c24051fdc5856ad14e222ca20a9f30dc653b1b/ivy_tests/test_ivy/helpers/assertions.py#L81). This checks not only for values, but for shapes and dtypes as well.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_onnx/test_elementwise.py
mosesdaudu001: Hi @AnnaTz

So i have tried using the `helpers.value_test`, but the issue with it is that you can only compare one framework with another framework. So that means I will only be able to compare the onnx operator result with another framework like torch and not all the frameworks.

The next commit I will make shows that using the value_test for the onnx.abs method actually works, but it can only be tested with one framework.

AnnaTz: We have to check all backends one after another anyway to test for all cases, so calling `helpers.value_test` should work.

mosesdaudu001: So this new commit makes use of the `@given` and the `helpers.value_test` method, but the helper function takes in 2 arrays and also a ground truth backend.

The first array is obviously ONNX, while the second is PyTorch and the ground truth backend I used was also PyTorch

AnnaTz: When testing a frontend function we should be comparing the return of the function to that of the original function. 
So here we want to import our onnx frontend and the original onnx and compare between them. There is no reason to use torch at all.
Also the `ground_truth_backend` argument is only used to print correct error messages. In the frontend testing case we always pass the name of the frontend, in this case "onnx".

danielmunioz:
Merging this now to move ahead with the reformatting task as discussed last week
Thanks for all the Good work @mosesdaudu001!

PR #19272, opened 2023-07-11
Fix-Todo: Add docstring to unify function

sladyn98:
This commit adds a comprehensive docstring to the 'unify' function. The 'unify' function is designed to compile and combine multiple callable objects into a single one, with optional additional behaviors such as including custom source code and numpy support. The newly added docstring provides detailed descriptions of the parameters, their types, and the return type of the function.

### DIFF
--- a/ivy/compiler/compiler.py
+++ b/ivy/compiler/compiler.py
@@ -159,7 +159,6 @@ def transpile(
     )
 
 
-# TODO: include docstring
 def unify(
     *objs: Callable,
     source: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


sladyn98:
I think we can merge this PR.

Infrared1029:
I guess someone on the graph compiler team would be more suited to review the docstrings cc: @mattbarrett98

ivy/compiler/compiler.py
```diff
@@ -168,6 +167,40 @@ def unify(
     with_numpy: bool = False,
     **transpile_kwargs,
 ) -> Callable:
+    """
+    Compiles and unifies multiple callable objects into a single callable. This is
```
mattbarrett98, line 171: this doesn't sound right. if multiple callables are passed in it will unify each of them separately, returning multiple callables. `unify` is equivalent to transpiling to `ivy`, meaning it's used to convert some native code (eg torch code) to an equivalent graph of ivy functions

ivy/compiler/compiler.py
```diff
@@ -168,6 +167,40 @@ def unify(
...
+    primarily used to create a single function that encapsulates the behavior of
+    multiple functions.
+
+    Parameters
+    ----------
+    *objs : Callable
+        The callable objects (e.g., functions) to unify.
+
+    source : str, optional
```
mattbarrett98, line 180: I guess this is written by copilot? 😅 `source` specifies what the source framework of the callable is. if you're unifying torch code, you need `source="torch"`

ivy/compiler/compiler.py
```diff
@@ -168,6 +167,40 @@ def unify(
...
+        The source code string to be included in the unified function. This can be
+        useful if the unified function needs to include some custom logic not
+        encapsulated by the input callable objects.
+
+    args : tuple, optional
+        A tuple of arguments to be passed to the unified function.
```
mattbarrett98, line 186: the args and kwargs are there to be passed into the original callable, as example inputs to the original callable for the tracing

ivy/compiler/compiler.py
```diff
@@ -168,6 +167,40 @@ def unify(
...
+
+    kwargs : dict, optional
+        A dictionary of keyword arguments to be passed to the unified function.
+
+    with_numpy : bool, optional
```
mattbarrett98, line 191: with_numpy indicates whether any use of numpy should be included in the graph. so if the source is torch but with_numpy is True, then you are also enabling numpy functions (eg `np.add`) to be in the graph, of course as well as any torch functions which are in the graph

mattbarrett98 (commented):
I imagine there are already docstrings in the gc repo which describe a lot of this already

sladyn98:
IMO this can be done by someone with more torch knowledge

PR #19273, opened 2023-07-11
Adding `hermitian` parameter to `ivy.pinv`

mobley-trent:
This PR adds the `hermitian` parameter to `ivy.linalg.pinv`. If the input is a hermitian matrix and the flag is set to `True`, a more efficient calculation is performed on the matrix.

### DIFF
--- a/ivy/functional/backends/jax/linear_algebra.py
+++ b/ivy/functional/backends/jax/linear_algebra.py
@@ -274,10 +274,13 @@ def pinv(
     /,
     *,
     rtol: Optional[Union[float, Tuple[float]]] = None,
+    hermitian: Optional[bool] = False,
     out: Optional[JaxArray] = None,
 ) -> JaxArray:
     if rtol is None:
         ret = jnp.linalg.pinv(x)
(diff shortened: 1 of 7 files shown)


mobley-trent:
The tensorflow backend doesn't support the `hermitian` flag so I had to implement it compositionally using the same computation used in paddle. I also added a testing strategy that draws a random hermitian matrix if it draws `True` for the hermitian flag.

MahmoudAshraf97 (changes requested):
Hello Edd, the current design of `pinv` introduces incosistencies when `rtol=None` because it uses the default for each framework which is not the same. I suggest setting `rtol` to zero manually when it's equal to None, this will also enable us to return the result directly and pass all the argument without having if conditions.
as for the tf implementation it can be much simpler if you used their internal implementation and used the if condition in the singular calculation part only instead of having two implementations

mobley-trent:
Hey @MahmoudAshraf97 Tensorflow does not support the hermitian flag natively, thats why I added the support with lower level ops.

PR #19275, opened 2023-07-11
Torch Frontend linalg refactor

DragosStoican:


ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -21,87 +20,55 @@
...
 # vector_norm
 @handle_frontend_test(
     fn_tree="torch.linalg.vector_norm",
     dtype_values_axis=helpers.dtype_values_axis(
-        available_dtypes=helpers.get_dtypes("numeric"),
+        available_dtypes=helpers.get_dtypes("valid"),
         valid_axis=True,
-        min_value=-1e04,
-        max_value=1e04,
+        force_int_axis=True,
```
AnnaTz, line 51: I think `vector_norm` supports tuple `dim` as well.

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -1325,33 +1223,31 @@ def test_torch_characteristic_equation_solver(
...
-        lambda x: "float16" not in x[0]
-        and "bfloat16" not in x[0]
-        and np.linalg.cond(x[1][0][:, :-1]) < 1 / sys.float_info.epsilon
-        and np.linalg.det(x[1][0][:, :-1]) != 0
-        and np.linalg.cond(x[1][0][:, -1].reshape(-1, 1)) < 1 / sys.float_info.epsilon
+        lambda x: np.linalg.cond(x[1][0][:, :-1]) < 1 / sys.float_info.epsilon
+                  and np.linalg.det(x[1][0][:, :-1]) != 0
+                  and np.linalg.cond(x[1][0][:, -1].reshape(-1, 1)) < 1 / sys.float_info.epsilon
     ),
-    check=st.booleans(),
```
AnnaTz, line 1334: Even though there is no `check` argument, there is a `left` argument to be tested.

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -1273,50 +1172,49 @@ def test_torch_multi_dot(
...
+        input_dtypes=[input_dtype[0], input_dtype[0]],
         frontend=frontend,
+        test_flags=test_flags,
         fn_tree=fn_tree,
-        test_values=True,
-        tensors=x,
+        on_device=on_device,
+        A=input,
+        B=other,
+        check_errors=check,
```
AnnaTz, line 1213: I think there should be a `left` argument here as well.

AnnaTz (changes requested):

MahmoudAshraf97 (commented):
Hi @DragosStoican , I see that you added several unsupported dtypes decorators for functions that support these dtypes, is there any reason for that?

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
```diff
@@ -530,46 +467,46 @@ def test_torch_matrix_exp(
...
         on_device=on_device,
-        # rtol=1e-04,
-        # atol=1e-04,
+        rtol=1e-03,
+        atol=1e-03,
         input=x[0],
         ord=ord,
         dim=axis,
         keepdim=keepdim,
-        dtype=dtype[0],
```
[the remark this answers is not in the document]
DragosStoican: Ah yes correct!

This was a complicated case. Because the "dtype" cannot be a dtype with less bits the the input dtype, and it can only be a float dtype for float inputs. So the logic behind generating values for this argument when testing for all valid dtypes is not trivial. I'll revert the changes and update this test in another PR, where I will have to implement that logic in a st.composite function

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
DragosStoican: left is not implemented at all. I will open another PR for implementing left in both solve and solve_ex

DragosStoican:
Hi @MahmoudAshraf97 . If you are referring to the unsupported dtypes decorators in the paddle backend, I added them because I was getting the error `RuntimeError: (NotFound) The kernel with key (CPU, Undefined(AnyLayout), complex64) of kernel `qr` is not registered. Selected wrong DataType `complex64`. Paddle support following DataTypes: float64, float32` , which made me think those functions don't support complex dtypes at all. I looked a little bit more into it and it's only `complex64` (and `complex128` in some case) that's causing the error. I updated the decorators to be more inclusive. Hope you're happy with it now! Let me know what you think!

ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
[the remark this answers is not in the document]
AnnaTz: I've seen this problem before. Currently we have `helpers.get_castable_dtype` but it doesn't cover such cases.
It would be great if you could make the strategy general enough to handle [this](https://github.com/unifyai/ivy/blob/183173a29a1362ba5a8dc13f4828761c9026c2b2/ivy_tests/test_ivy/test_frontends/test_torch/test_reduction_ops.py#L885) case as well!

AnnaTz (approved):
Fine by me, let's wait to get the okay by @MahmoudAshraf97 as well.

MahmoudAshraf97 (approved):
Hello, the decorators used now were correct, but there is a bugh that prevents the expansion of `complex` to `"complex64","complex128"` hence the need for explicitly stating them
other than that LGTM

PR #19285, opened 2023-07-12
Update linear_algebra.py

yluoc:
update info and -info form float to string in ivy/data_classes/array/linear_algebra.py/vector_norm.

### DIFF
--- a/ivy/data_classes/array/linear_algebra.py
+++ b/ivy/data_classes/array/linear_algebra.py
@@ -916,7 +916,7 @@ class _ArrayWithLinearAlgebra(abc.ABC):
         *,
         axis: Optional[Union[int, Sequence[int]]] = None,
         keepdims: bool = False,
-        ord: Union[int, float, Literal[inf, -inf]] = 2,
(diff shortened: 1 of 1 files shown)


yluoc:
Type arguments for "Literal" must be None, a literal value (int, bool, str, or bytes), or an enum value. So I simply change the red lighted info and -info form float to String

vedpatwardhan (commented):
Hey @Ishticode, thanks for requesting the review abd @yluoc thanks for making the changes! I think the reason we've added the type-hint in this way is because the same is followed in the [Array API Standard](https://data-apis.org/array-api/latest/extensions/generated/array_api.linalg.vector_norm.html#extensions-generated-array-api-linalg-vector-norm--page-root). So I think we should first create an issue on their [repository](https://github.com/data-apis/array-api-tests) (if not done already) and based on their response, we can move forward with making the change to the type-hint here. Thanks 😄

ivy-seed:

ivy-seed:

PR #19314, opened 2023-07-12
Fix torch.as_strided for non-contiguous tensors

AnnaTz:
So that this is True:
```
ivy.set_torch_backend()
size = [12, 1, 512, 64]
stride = [64, 196608, 768, 1]
x = torch.rand((12, 1, 64, 512)).transpose(-1, -2)
orig = torch.as_strided(x, size, stride).detach().numpy()
front = torch_frontend.as_strided(x, size, stride).ivy_array.data.detach().numpy()
print(np.allclose(orig, front))
```

### DIFF
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -134,16 +134,14 @@ class Array(
         _ArrayWithSortingExperimental.__init__(self),
         _ArrayWithStatisticalExperimental.__init__(self),
         _ArrayWithUtilityExperimental.__init__(self),
-        self._view_attributes(data),
         self._init(data, dynamic_backend)
+        self._view_attributes(data)
 
     def _init(self, data, dynamic_backend=None):
         if ivy.is_ivy_array(data):
             self._data = data.data
         elif ivy.is_native_array(data):
             self._data = data
-            if hasattr(data, '_base'):
-                self._base = data._base
         elif isinstance(data, np.ndarray):
             self._data = ivy.asarray(data)._data
         else:
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -134,14 +134,16 @@ class Array(
         _ArrayWithSortingExperimental.__init__(self),
         _ArrayWithStatisticalExperimental.__init__(self),
         _ArrayWithUtilityExperimental.__init__(self),
+        self._view_attributes(data),
         self._init(data, dynamic_backend)
-        self._view_attributes(data)
 
     def _init(self, data, dynamic_backend=None):
         if ivy.is_ivy_array(data):
             self._data = data.data
         elif ivy.is_native_array(data):
             self._data = data
+            if hasattr(data, '_base'):
+                self._base = data._base
         elif isinstance(data, np.ndarray):
             self._data = ivy.asarray(data)._data
         else:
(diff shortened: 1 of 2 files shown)


ivy-leaves:

Thank you :hugs:

vedpatwardhan (commented):
Hey @AnnaTz, looking through the current view implementation in the framework, I'm not sure if we can consistently deal with the case when the input `ivy.NativeArray` itself has a view, because it's not just about the `_base` attribute for the `ivy.Array`, we also need to populate the `_manipulation_stack` of the operations that were performed on that `ivy.NativeArray` before this view was created. This is because we don't deal with the memory layout of arrays on the framework level and this seems to be an inherent limitation of that constraint.
So I think unless we can populate the `_manipulation_stack`, `_view_refs` and stuff correctly, we can't support the behaviour in the framework where an input `ivy.NativeArray` was already a view 😅

AnnaTz:
ivy-gardener

AnnaTz:
> 

Thanks Ved! For now I will limit the changes to `torch.as_strided` just to make it work correctly for the demo. Hope that's fine 😅
I will have to dive into view handling in the frontends as well soon enough.

AnnaTz:
> Hey @AnnaTz, looking through the current view implementation in the framework, I'm not sure if we can consistently deal with the case when the input `ivy.NativeArray` itself has a view, because it's not just about the `_base` attribute for the `ivy.Array`, we also need to populate the `_manipulation_stack` of the operations that were performed on that `ivy.NativeArray` before this view was created, which is then useful to ensure that all views are updated when an inplace update is performed. This is because we don't deal with the memory layout of arrays on the framework level and this seems to be an inherent limitation of that constraint. So I think unless we can populate the `_manipulation_stack`, `_view_refs` and stuff correctly by retrieving an equivalent from `torch.Tensor` and `np.ndarray` (which is unlikely), I can't really think of a way to support the behaviour in the framework where an input `ivy.NativeArray` was already a view 😅 But happy to hear suggestions if you think there's a way to get this to work, I'll also dive a bit into this soon. Thanks 😄

Thanks Ved! For now I will limit the changes to torch.as_strided just to make it work correctly for the demo. Hope that's fine 😅
I will have to dive into view handling in the frontends as well soon enough.

AnnaTz:
> Hey @AnnaTz, looking through the current view implementation in the framework, I'm not sure if we can consistently deal with the case when the input `ivy.NativeArray` itself has a view, because it's not just about the `_base` attribute for the `ivy.Array`, we also need to populate the `_manipulation_stack` of the operations that were performed on that `ivy.NativeArray` before this view was created, which is then useful to ensure that all views are updated when an inplace update is performed. This is because we don't deal with the memory layout of arrays on the framework level and this seems to be an inherent limitation of that constraint. So I think unless we can populate the `_manipulation_stack`, `_view_refs` and stuff correctly by retrieving an equivalent from `torch.Tensor` and `np.ndarray` (which is unlikely), I can't really think of a way to support the behaviour in the framework where an input `ivy.NativeArray` was already a view 😅 But happy to hear suggestions if you think there's a way to get this to work, I'll also dive a bit into this soon. Thanks 😄

Thanks Ved! For now I will limit the changes to `torch.as_strided` just to make it work correctly for the demo. Hope that's fine 😅
I will have to dive into view handling in the frontends as well soon enough.

AnnaTz:
@Sam-Armstrong Could you let me know if the current changes work for you? Thanks!

PR #19318, opened 2023-07-12
Enable torch_frontend.Tensor supports sparse tensor

themaigod:
Notice: Current ivy.SparseArray version is not able to fit most sparse tensor behaviors in Pytorch. It will be a huge job to support sparse tensor support, especially when ivy.SparseArray is not inherit from ivy.Array.

Current limitations:
1. ivy.SparseArray does not support device

TODO:
1. Support more functions of sparse tensor in pytorch
2. Check current functions is able for ivy.SparseArray


ivy-leaves:

Thank you :hugs:

AnnaTz:
ivy-gardener

ivy/functional/frontends/torch/tensor.py
```diff
@@ -18,6 +18,16 @@
...
     def __init__(self, array, device=None, _init_overload=False):
+
+        # ivy.SparseArray does not have a device attribute
+        # TODO: Update this when ivy.SparseArray has a device attribute
+        if ivy.is_ivy_sparse_array(array):
+            self.ivy_array = array
+            return
+        elif ivy.is_native_sparse_array(array):
+            self.ivy_array = ivy.SparseArray(array)
+            return
```
AnnaTz, line 29: I don't think we should be adding extra logic here. To add `torch.sparse_coo_tensor`  to our frontend, you will need to create a `sparse_coo_tensor` function (in `creation_ops.py`) that will return an `ivy.SparseArray` object. 

The required arguments are already supported; `indices`, `values`, and `size` of `sparse_coo_tensor` correspond to `coo_indices`, `data`, and `dense_shape` of `SparseArray` respectively.
Other than that, I think the `dtype` and `device` arguments can be handled calling the necessary dtype and device handling ivy functions on the created `SparseArray` object.

AnnaTz: I do think however that it might be beneficial to make `ivy.SparseArray` inherit from `ivy.Array`.
Also I think we should be able to create sparse arrays in ivy by calling the `ivy.array` creation function instead of using the `ivy.SparseArray` initialization directly.
Currently calling `ivy.array` given a sparse native array results in a non-sparse array, as shown below:
```
import ivy
import torch

# Create a sparse torch tensor
i = [[0, 1, 1],
         [2, 0, 2]]
v = [3, 4, 5]
s = torch.sparse_coo_tensor(i, v, (2, 3))

ivy.set_torch_backend()
x = ivy.array(s)

ivy.is_ivy_sparse_array(x)   # False
ivy.is_native_sparse_array(x)   # True
```
This would be up to @vedpatwardhan to decide.

AnnaTz (changes requested):

ivy/functional/frontends/torch/tensor.py
```diff
@@ -1478,13 +1490,32 @@ def cholesky(self, upper=False):
...
-            and isinstance(reps[0], Iterable)
+                isinstance(reps, Iterable)
+                and len(reps) == 1
+                and isinstance(reps[0], Iterable)
         ):
             reps = reps[0]
         return torch_frontend.tile(self, reps)
 
+    @property
+    def layout(self):
```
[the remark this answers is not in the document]
themaigod: > `torch.layout` should be a class not a property, and I'm not sure we need it in order to add `torch.sparse_coo_tensor`.

However, pytorch currently supports multiple sparse types, and the only way to justify which it use is by layout property

ivy/functional/frontends/torch/tensor.py
themaigod: > I do think however that it might be beneficial to make `ivy.SparseArray` inherit from `ivy.Array`. Also I think we should be able to create sparse arrays in ivy by calling the `ivy.array` creation function instead of using the `ivy.SparseArray` initialization directly. Currently calling `ivy.array` given a sparse native array results in a non-sparse array, as shown below:
> 
> ```
> import ivy
> import torch
> 
> # Create a sparse torch tensor
> i = [[0, 1, 1],
>          [2, 0, 2]]
> v = [3, 4, 5]
> s = torch.sparse_coo_tensor(i, v, (2, 3))
> 
> ivy.set_torch_backend()
> x = ivy.array(s)
> 
> ivy.is_ivy_sparse_array(x)   # False
> ivy.is_native_sparse_array(x)   # True
> ```
> 
> This would be up to @vedpatwardhan to decide.

Yes, it is quite wired that SparseArray is not a kind of Array. But it is up to how different frameworks achieve this. In pytorch, there is not a class for sparse tensor, only Tensor. So it can be benficial from it.

themaigod: > I don't think we should be adding extra logic here. To add `torch.sparse_coo_tensor` to our frontend, you will need to create a `sparse_coo_tensor` function (in `creation_ops.py`) that will return an `ivy.SparseArray` object.
> 
> The required arguments are already supported; `indices`, `values`, and `size` of `sparse_coo_tensor` correspond to `coo_indices`, `data`, and `dense_shape` of `SparseArray` respectively. Other than that, I think the `dtype` and `device` arguments can be handled calling the necessary dtype and device handling ivy functions on the created `SparseArray` object.

For `torch.sparse_coo_tensor` , it should return a `Tensor` shown in `tensor.py`. Like I have mentioned, pytorch can use sparse tensor, but it is still a `torch.Tensor`. If it is a `ivy.SparseArray` , the shown behaviors are totally different.

themaigod: > > I don't think we should be adding extra logic here. To add `torch.sparse_coo_tensor` to our frontend, you will need to create a `sparse_coo_tensor` function (in `creation_ops.py`) that will return an `ivy.SparseArray` object.
> > The required arguments are already supported; `indices`, `values`, and `size` of `sparse_coo_tensor` correspond to `coo_indices`, `data`, and `dense_shape` of `SparseArray` respectively. Other than that, I think the `dtype` and `device` arguments can be handled calling the necessary dtype and device handling ivy functions on the created `SparseArray` object.
> 
> For `torch.sparse_coo_tensor` , it should return a `Tensor` shown in `tensor.py`. Like I have mentioned, pytorch can use sparse tensor, but it is still a `torch.Tensor`. If it is a `ivy.SparseArray` , the shown behaviors are totally different.

Here is an example that I think we should support:

```
import ivy.frontends.torch as torch

ivy.set_backend('torch')

# Create a sparse torch tensor
i = [[0, 1, 1],
         [2, 0, 2]]
v = [3, 4, 5]
s = torch.sparse_coo_tensor(i, v, (2, 3))
isinstance(s, torch.Tensor)
```

themaigod:
[Enable torch_frontend.Tensor supports sparse tensor] #19484 

Sorry, I fork someone personal sub fork. So I delete the my fork and refork. Here is the new commit.

ivy/functional/frontends/torch/tensor.py
AnnaTz: The thing is all frontend implementations use ivy functions which return ivy arrays, and then through their decorator `@to_ivy_arrays_and_back` the ivy arrays get converted to frontend torch tensors. So I think the function `sparse_coo_tensor` should create and return an `ivy.SparseArray` which would then get converted to a torch tensor inside [`outputs_to_frontend_arrays`](https://github.com/unifyai/ivy/blob/c1c24051fdc5856ad14e222ca20a9f30dc653b1b/ivy/functional/frontends/torch/func_wrapper.py#L63). We will need to update this accordingly but first we need to decide if `ivy.SparseArray` will become an `ivy.Array` @vedpatwardhan.

ivy/functional/frontends/torch/tensor.py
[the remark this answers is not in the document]
AnnaTz: I don't really understand what you mean. I think `sparse_coo_tensor` can be added to our frontend without `layout`.

ivy/functional/frontends/torch/tensor.py
vedpatwardhan: Yeah that's a good point actually! Given that `torch.sparse_coo_tensor` returns a `torch.Tensor`, `ivy.SparseArray` sub-classing `ivy.Array` makes sense to me. However, I suppose this is something which might've been thought about before when this class was implemented and there could be some rationale behind that not being the case at the moment. @juliagsy any ideas? Thanks 😄

themaigod: Great! Let's see what is the next step. I will do the modification if we will accept this change (inherit from Array).

juliagsy: > Yeah that's a good point actually! Given that `torch.sparse_coo_tensor` returns a `torch.Tensor`, `ivy.SparseArray` sub-classing `ivy.Array` makes sense to me. However, I suppose this is something which might've been thought about before when this class was implemented and there could be some rationale behind that not being the case at the moment. @juliagsy any ideas? Thanks 😄

It's been some time and I can't seem to remember why `ivy.Array` wasn't inherited 😅 But I agree that it's more relevant to inherit from `ivy.Array` since it's kinda a sub array type. The change sounds good to me! 
Thanks all and thanks for flagging @themaigod !

vedpatwardhan: Thanks for clarifying @juliagsy, we'll get someone to work on making `ivy.SparseArray` subclass `ivy.Array` from the framework side soon, is this a high-priority frontend function @AnnaTz? Thanks 😄

AnnaTz: No it's not high-priority @vedpatwardhan

vedpatwardhan: Got it! Thanks for clarifying @AnnaTz 😄

PR #19336, opened 2023-07-13
Fix `IvyWithGlobalProps` for `ivy.with_backend`

CatB1t:
The internal `ivy.with_backend` object wasn't getting updated due to updating `sys.modules`, which local ivy backends doesn't exist in, it only exists in Ivy cache.

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -1439,4 +1439,13 @@ class IvyWithGlobalProps(sys.modules[__name__].__class__):
         self.__dict__[name] = value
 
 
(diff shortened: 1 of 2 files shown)


vedpatwardhan (approved):
lgtm! Just a suggestion, we probably should have some tests which would get triggered which ensure that the changes made work both with `with_backend` and the global `ivy`, so that such failures are captured when we make similar changes. Thanks @CatB1t 😄

CatB1t:
@vedpatwardhan Great Idea, I'll have a think about it, thanks Ved!

PR #19343, opened 2023-07-13
Stateful changes 2

RickSanchezStoic:
Modifies ivy.Module to freeze weight initialization and delay it till required to handle duplicate weight generation and peak memory utilization. Modifies container_identical methods to have `build_callable` and `asssign_and_assert` to build weights while comparing and assigning immediately.


vedpatwardhan (changes requested):
Hey @RickSanchezStoic, I tested this code on the example we were using before, and the weights still seem to be getting initialized.
```
import ivy

ivy.set_backend('torch')


class Net(ivy.Module):
    def __init__(self, N, v=None):
        self.N = N

        super().__init__(
            v=v,
        )

    def _build(self, *args, **kwargs):
        self.lin1 = ivy.Linear(self.N, self.N, with_bias=False)
        self.lin2 = ivy.Linear(self.N, self.N, with_bias=False)

    def _forward(self, x):
        return self.lin2(self.lin1(x))


N = 1
v = ivy.Container(
    {"lin1": {"w": ivy.ones(shape=(N, N))}, "lin2": {"w": ivy.ones(shape=(N, N))}}
)

test_net = Net(N=N, v=v)
``` 
Given that the only purpose of the `__new__` method is to assign a static variable, I think we could assign `Module._init_var` it in the `__init__` itself. In fact, we could treat the `Module._init_var` static variable in the exact same way as we treated the global `var_initialization_mode` which we had tried before. That way we wouldn't need to make any changes to the container (except for the `return_dict` line which @saeedashrraf mentioned [here](https://github.com/unifyai/ivy/pull/18164#discussion_r1250803939)) and also no need to update `__getattribute__` and the logic to build a container of callables. 
Let me know what you think, happy to clarify further. Thanks for diving into this @RickSanchezStoic 💪🔥

RickSanchezStoic:
> Hey @RickSanchezStoic, I tested this code on the example we were using before, and the weights still seem to be getting initialized.
> 
> ```
> import ivy
> 
> ivy.set_backend('torch')
> 
> 
> class Net(ivy.Module):
>     def __init__(self, N, v=None):
>         self.N = N
> 
>         super().__init__(
>             v=v,
>         )
> 
>     def _build(self, *args, **kwargs):
>         self.lin1 = ivy.Linear(self.N, self.N, with_bias=False)
>         self.lin2 = ivy.Linear(self.N, self.N, with_bias=False)
> 
>     def _forward(self, x):
>         return self.lin2(self.lin1(x))
> 
> 
> N = 1
> v = ivy.Container(
>     {"lin1": {"w": ivy.ones(shape=(N, N))}, "lin2": {"w": ivy.ones(shape=(N, N))}}
> )
> 
> test_net = Net(N=N, v=v)
> ```
> 
> Given that the only purpose of the `__new__` method is to assign a static variable, I think we could assign `Module._init_var` it in the `__init__` itself. In fact, we could treat the `Module._init_var` static variable in the exact same way as we treated the global `var_initialization_mode` which we had tried before. That way we wouldn't need to make any changes to the container (except for the `return_dict` line which @saeedashrraf mentioned [here](https://github.com/unifyai/ivy/pull/18164#discussion_r1250803939)) and also no need to update `__getattribute__` and the logic to build a container of callables. Let me know what you think, happy to clarify further. Thanks for diving into this @RickSanchezStoic 💪🔥

Hey Ved, as we discussed before, weight generation will happen for leaf nodes, as we build compare and assign at the leaf nodes (contrary to building the entire structure and then comparing) , hence if peak memory utilization in ideal case is `m` (where no weights are generated at all), then in this case it will be `m`+`p` where `p` is the memory consumed for generating a leaf node. You can check this out by running this code, by running twice, once removing `v=v` . 


```
import ivy

ivy.set_backend('torch')

class Net(ivy.Module):
    def __init__(self, N, v=None):
        self.N = N

        super().__init__(
            v=v,
        )

    def _build(self, *args, **kwargs):
        self.lin1 = ivy.Linear(self.N, self.N, with_bias=False)
        self.lin2 = ivy.Linear(self.N, self.N, with_bias=False)

    def _forward(self, x):
        return self.lin2(self.lin1(x))


N = 1
v = ivy.Container(
    {"lin1": {"w": ivy.ones(shape=(N, N))}, "lin2": {"w": ivy.ones(shape=(N, N))}}
)
import resource
test_net = Net(N=N)
print(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss/1024)
```
For cases where there is just one node , this makes no difference, but when model grows in size, this becomes more and more optimal. I ran these tests for different models with layers increasing in size and it became more and more optimized compared to the building the entire structure and then comparing method.

vedpatwardhan:
Ah yeah it seems to be leaf by leaf now that I test it, not sure why it looked the other way, sorry about that!
Actually what I meant by the static variable is that `init_var` can also be updated at the first line of the `__init__` and given that it's a static variable, it would propagate through the child modules anyway. The main thing being used specific to `__new__` in the implementation seems to be `cls.__module__` which is equivalent to `self.__module__` anyway. And given that the `__new__` method is called before we enter `__init__` and we're always calling `super().__new__(cls)` regardless of whether or not it's a stateful class, it's equivalent to having the same logic in the `__init__` instead.
```
    def __init__(<params>):
        # check the module of the class
        # if it's stateful, it's internal
        # we leave this untouched
        if "stateful" in self.__module__:
            # we are not assigning it a variable
            pass
        else:
            # first check if a var is already assigned
            # this would mean it is a nested custom class
            if not hasattr(Module, "_init_var"):
                # if not , create it and add
                Module._init_var = [self.__class__]
            else:
                Module._init_var.append(self.__class__)
        <everything else already in the __init__>
```
If you try out this code, this will also have the same behaviour as with the `__new__` method. Which is why I was thinking that with the static variable approach you're using (`_init_var` being that static variable to the `ivy.Module` class) it simplifies a lot of stuff, we could also bypass the updates to the container without the need to construct containers of callables and stuff as I mentioned above.
Happy to know your thoughts, thanks @RickSanchezStoic 😄 💪

RickSanchezStoic:
> Ah yeah it seems to be leaf by leaf now that I test it, not sure why it looked the other way, sorry about that! Actually what I meant by the static variable is that `init_var` can also be updated at the first line of the `__init__` and given that it's a static variable, it would propagate through the child modules anyway. The main thing being used specific to `__new__` in the implementation seems to be `cls.__module__` which is equivalent to `self.__module__` anyway. And given that the `__new__` method is called before we enter `__init__` and we're always calling `super().__new__(cls)` regardless of whether or not it's a stateful class, it's equivalent to having the same logic in the `__init__` instead.
> 
> ```
>     def __init__(<params>):
>         # check the module of the class
>         # if it's stateful, it's internal
>         # we leave this untouched
>         if "stateful" in self.__module__:
>             # we are not assigning it a variable
>             pass
>         else:
>             # first check if a var is already assigned
>             # this would mean it is a nested custom class
>             if not hasattr(Module, "_init_var"):
>                 # if not , create it and add
>                 Module._init_var = [self.__class__]
>             else:
>                 Module._init_var.append(self.__class__)
>         <everything else already in the __init__>
> ```
> 
> If you try out this code, this will also have the same behaviour as with the `__new__` method. Which is why I was thinking that with the static variable approach you're using (`_init_var` being that static variable to the `ivy.Module` class) it simplifies a lot of stuff, we could also bypass the updates to the container without the need to construct containers of callables and stuff as I mentioned above. Happy to know your thoughts, thanks @RickSanchezStoic 😄 💪


Yeah. but you have to consider the scenario where you are defining the variables before calling the Module.__init__ , since we already are inside a custom class's init. we can't do anything to stop weight initialization there, the `__new__` method approach helps us add this feature too, allowing a user to define their variables anywhere within their model.

vedpatwardhan (approved):
> Yeah. but you have to consider the scenario where you are defining the variables before calling the Module.**init** , since we already are inside a custom class's init. we can't do anything to stop weight initialization there, the `__new__` method approach helps us add this feature too, allowing a user to define their variables anywhere within their model.

Yeah you're right, my bad! Let's get this merged in that case, could you please update the docstring of `cont_identical`  method (and the other updated container methods) mentioning this use case for the additional arguments and updating the `return `ivy.Container(return_dict)` line as @saeedashrraf mentioned in the other PR?

Probably in the future we can consider moving `super().__init__()` to the top of the `__init__` to deal with such cases when child modules were defined in the `__init__`, but given that this PR fixes a very important issue, let's get it merged. Thanks for all the hard work @RickSanchezStoic 😄

Issue #18973, opened 2023-07-07
avg_pool1d


so-dipe:

so-dipe:
HI, I can't run the tests because I keep getting this error. I am working with a GitHub codespace and I'm positive I set everything right...

`ImportError while loading conftest '/workspaces/ivy/ivy_tests/test_ivy/conftest.py'.
ivy_tests/test_ivy/__init__.py:1: in <module>
    from .. import config
E   ImportError: cannot import name 'config' from 'ivy_tests' (/workspaces/ivy/ivy_tests/__init__.py)`

diksha-shrivastava13:
Hi @so-dipe! 

I stumbled across this issue and just a quick tip, it would be hard for anyone to follow along your branch and codespace if you don't have a PR linked. In case you haven't yet resolved the issue mentioned, try "pip3 install jax jaxlib" on your codespace terminal, I think that might be the problem.

Also, join the discord channel (https://discord.com/invite/sXyFF8tDtm), you'd be able to ask for help more efficiently.
Have a good day!

PR #19361, opened 2023-07-13
avg_pool1d #18973

so-dipe:
Close https://github.com/unifyai/ivy/issues/18973


ivy_tests/test_ivy/test_frontends/test_paddle/test_nn/test_functional/test_pooling.py
```diff
@@ -1,3 +1,34 @@
...
+        input_dtypes=input_dtype,
+        frontend=frontend,
+        test_flags=test_flags,
+        on_device=on_device,
+        fn_tree=fn_tree,
+        x=x[0],
+        kernel_size=kernel_size,
+        stride=stride,
+        padding=padding,
+        exclusive=False,
```
Daniel4078, line 32: You should try to generate a boolean value for exclusive and pass it in here instead of just setting it to False to cover all the situations, the same for ceil_mode.

ivy/functional/backends/torch/experimental/layers.py
```diff
@@ -272,6 +272,9 @@ def avg_pool1d(
 
     if data_format == "NWC":
         x = x.permute(0, 2, 1)
+    elif data_format == "NCL":
```
Daniel4078, line 275: why not just use an "or" clause to comebine these two cases as the treatment is the same in both cases right?

Daniel4078 (changes requested):
Great work, but I have some suggestions for you, please check the comments below. Thanks.

so-dipe:
@Daniel4078 I made the requested changes but you should note that ceil_mode is affected when it's value is set to True and the padding is 'VALID', so the tests for this function in torch and tensorflow do not generate boolean but only use False.

Daniel4078 (changes requested):
Understand, you can just use false in that case. Your tests are passing fine, but why not improve the other backend implementations with my suggestion of combining the two clauses together and remove the comments as they serve no purposes now? 
Also, even though it may not be your problem to solve and you do not need to deal with the problem, but I found some strange problem with other functions using the avg_pool1d function, please have a look here(https://github.com/unifyai/ivy/actions/runs/5564070210/jobs/10163803930?pr=19361):
```
E   AssertionError: returned shape = (1, 2, 1), ground-truth returned shape = (1, 1, 2)
E   Falsifying example: test_tensorflow_avg_pool1d(
E       on_device='cpu',
E       x_k_s_p_df=(['float16'],
E        [array([[[-1., -1.]]], dtype=float16)],
E        (1,),
E        (1,),
E        'VALID'),
E           with_out=False,
E           inplace=False,
E       ),
E       fn_tree='ivy.functional.frontends.tensorflow.nn.avg_pool1d',
E   )
```

Daniel4078 (approved):
Looks good to me!

PR #19472, opened 2023-07-16
Integrate SoftPlus Function into Paddle Frontend

BouchanaHicham:
Requesting Team Decision on Softplus Functions

- Created my own Softplus functions and corresponding tests
- Similar Softplus functions were found and commented out
- Both implementations have been thoroughly tested and passed successfully
- I have verified that all lint checks pass successfully, and the code is now clean and consistent. ✅

To proceed, the team needs to decide which implementation of Softplus functions to use:

- The existing commented-out functions
- My newly created functions

Please review both implementations and provide feedback on which option to proceed with. Additionally, if there are any changes or additions required in either implementation, please specify them for further refinement.

- Close https://github.com/unifyai/ivy/issues/19296

### DIFF
--- a/ivy/functional/frontends/paddle/nn/functional/activation.py
+++ b/ivy/functional/frontends/paddle/nn/functional/activation.py
@@ -211,12 +211,6 @@ def mish(x, name=None):
     return ivy.mish(x)
 
 
-@with_supported_dtypes({"2.5.0 and below": ("float32", "float64")}, "paddle")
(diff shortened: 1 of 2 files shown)


D0m-inic:
Hi, which commented functions are you talking about? I don't think you've committed them. Also, i've noticed that you've changed tests for functions *silu* and *leaky_relu*, these don't need to be changed so you can uncommit them 😄 .

PR #19491, opened 2023-07-16
interpolate function

### LINKED ISSUE #19489: interpolate

ausafchaps:
close #19489


ivy_tests/test_ivy/test_frontends/test_paddle/test_nn/test_functional/test_common.py
```diff
@@ -192,6 +192,61 @@ def test_paddle_zeropad2d(
...
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        x=x[0],
+        size=size,
+        scale_factor=scale_factor,
+        mode=mode,
+        align_corners=align_corners,
+    )
+
```
sladyn98, line 248: Arguments size and scale_factor: For interpolation, you generally use either size (the output size) or scale_factor (the multiplier for the spatial size). It's unusual to use both at the same time because they serve similar purposes. The function should probably handle cases where one or the other is None.


Tests for different dimensions: min_num_dims=4 and max_num_dims=4 imply that the function will only test 4D tensors. While it's common to have 4D tensors in convolutional neural networks (CNNs), testing for 3D or 5D tensors could make the test more robust, especially if you're working with different types of data or architectures.

Tests for different interpolation modes with align_corners: The test generates cases with all the interpolation modes combined with align_corners=True and align_corners=False. However, the align_corners option is not supported for all interpolation modes in some libraries. In PyTorch, for instance, it's not supported for 'nearest' or 'area'. If this is also the case for PaddlePaddle, you should adjust the test to only use align_corners with the modes that support it.

ausafchaps: - size (list|tuple|Tensor|None) – Output shape of image resize layer, the shape is (out_w, ) when input is a 3-D Tensor, the shape is (out_h, out_w) when input is a 4-D Tensor and is (out_d, out_h, out_w) when input is a 5-D Tensor. Default: None. If a list/tuple, each element can be an integer or a Tensor of shape: [1] or []. If a Tensor, its dimensions size should be a 1.

- scale_factor (float|Tensor|list|tuple|None) – The multiplier for the input height or width. At least one of size or scale_factor must be set. And size has a higher priority than scale_factor.Has to match input size if it is either a list or a tuple or a Tensor.If a list/tuple, each element can be an integer or a Tensor of shape: [1] or []. Default: None.

Argument size and scale_factor: Both size and scale_factor are used in interpolate function of paddlepaddle. 
https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/nn/functional/interpolate_en.html#interpolate

ausafchaps: align_corners: I only used the ones that are supported by PaddlePaddle

sladyn98 (approved):

sladyn98 (approved):

ausafchaps:
Do you require any more changes? If not then can you please merge my PR. Thank you.

sladyn98:
@ausafchaps The tests seem to be failing, could you check the reason

PR #19500, opened 2023-07-17
Implement apply_over_axes function

### LINKED ISSUE #19498: apply_over_axes

aazam-gh:
There is some work to be done with for the test of this function.
In order to input ndarray I am using the expand dims fucntion in the test, but I get the following error 
`TypeError: expand_dims requires ndarray or scalar arguments, got <class 'ivy.data_classes.array.array.Array'> at position 0`

Any tips?


aazam-gh:
Close #19498

PR #19519, opened 2023-07-17
Tensorflow igamma function

### LINKED ISSUE #19516: igamma

ChasJC23:
Close #19516

### DIFF
--- a/ivy/functional/frontends/tensorflow/math.py
+++ b/ivy/functional/frontends/tensorflow/math.py
@@ -711,3 +711,8 @@ def real(input, name=None):
 @to_ivy_arrays_and_back
 def atanh(x, name="atanh"):
     return ivy.atanh(x)
+
+
+@to_ivy_arrays_and_back
+def igamma(a, x, name=None):
+    return ivy.igamma(a, x=x)


AryanSharma21 (commented):
While adding front end functions, we should also add the corresponding test cases.
To add the test case for this function, you can add into 
/workspaces/ivy/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py

ChasJC23:
Of course, my bad. I've since added an implementation, however it seems to trip up at the asymptotes on the negative integers at the moment (eg. `a=-1.0, x=-1.0`). Is there some way of addressing / expecting this in the test?

AryanSharma21:
As per my understanding, the lgama function computes the log values of  factorial of (passed value -1 )
for ex lgama(5) = log(factorial(5-1))
so if we pass negative values, then the factorial can not be determined. so to address this problem maybe you should try restricting the test case to positive value. 

dont forget to look into documentation and do correct me if I am wrong

AryanSharma21 (commented):
As per my knowledge , the igamma implementation in the tensorflow library is based on standard definition of "incomplete gamma function" which indicates the vales of both x and a should be positive.
However other libraries may have an extended implementations which can cover negative values as well using euler reflection formula

so in this case you can try restricting the test cases to positive vales of x and a . 

Feel free to correct me if am wrong

ChasJC23:
yep, that sounds consistent with what I'm seeing in the implementation for `ivy.igamma`. I haven't personally seen a strict mathematical reason as to why negatives are a no-no, maybe I need a calculus refresher, but I think there's enough justification to restrict the domain. I'll just do that now, thanks very much!

AryanSharma21 (commented):
The test 1 in the workflow looks like the test cases are not passing.
Can you tell if the test cases are running locally ?

AryanSharma21 (commented):
atanh test passes in local environment might be config issue , otherwise the PR seems suifficient any updates for the test case 1 ?

AryanSharma21:
Try to run the test once in local environment by setting the device to cpu 
with tf.device('<appropriate cpu>') 

you can look into 
https://stackoverflow.com/questions/37660312/how-to-run-tensorflow-on-cpu
for reference
please provide the screen shot of test once you run it

AryanSharma21:
Try to run the test once in local environment by setting the device to cpu
with tf.device('')

you can look into
https://stackoverflow.com/questions/37660312/how-to-run-tensorflow-on-cpu
for reference
please provide the screen shot of test once you run it

ChasJC23:
I'm not sure what's happening here, is there anything I need to change? As far as I can tell, the error cases are not related to the code I've written, so what's the problem? Ignoring the `Could not find device for node` errors as your comment claiming how I can resolve that in my local tests was deleted, here are the errors I've found:

Tests failing for `numpy` backend are:
- a result of [this](https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/numpy/experimental/statistical.py#L402) cast detaching the output type from the input types;
- numerical inaccuracy for larger inputs presumably from the static "magic number" found [here](https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/numpy/experimental/statistical.py#L399).

Tests failing for `torch` backend are:
- numerical inaccuracy for larger inputs due to difference in the underlying frameworks' implementations.

Tests failing for `paddle` backend are:
- a result of [this](https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/paddle/experimental/statistical.py#L391) cast detaching the output type from the input types;
- being unable to flatten [here](https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/paddle/experimental/statistical.py#L377) if input type is `float16`;
- `NaN` getting into the mix for large inputs presumably due to `Infinity` showing up [here](https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/paddle/experimental/statistical.py#L386-L388).
- numerical inaccuracy for larger inputs presumably from the static "magic numbers" found [here](https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/paddle/experimental/statistical.py#L384-L385)

Apart from perhaps the `float16` one for `paddle` or the one about differences between the underlying backends, none of this has got to do with the wrapper I wrote. What's going on?

AryanSharma21:
I am requesting a review from @AnnaTz  here, Please let us understand what is happening here

ChasJC23:
Thank you, apologies for my impatience.

> please provide the screen shot of test once you run it

Here's roughly what I get when running the test locally. Other failing tests I've mentioned previously only happen occasionally, when the `tensorflow.python.framework.errors_impl.NotFoundError` error doesn't happen.

[Test Results — pytest for test_math test_tensorflow_igamma](https://github.com/unifyai/ivy/assets/45319866/027d69cd-ad98-4cda-93e0-35c987207853)

ivy/functional/frontends/tensorflow/math.py
```diff
@@ -711,3 +711,8 @@ def real(input, name=None):
 @to_ivy_arrays_and_back
 def atanh(x, name="atanh"):
     return ivy.atanh(x)
+
+
```
AnnaTz, line 715: Hi both, `ivy.gamma` had a few issues regarding dtype handling, which I have now resolved. Please merge the latest master and try running the frontend again. 
Also, kindly add the supported dtypes decorator here and use `available_dtypes=helpers.get_dtypes("valid")` in the test. This is a matter of good practice to make sure all our functions are well-defined. Thanks!

AnnaTz (changes requested):
Hi both, I ran the test locally and the errors I am getting are dtype related. The documentation clearly states that only `bfloat16`, `float16`, `float32` and `float64` are supported

ivy/functional/frontends/tensorflow/math.py
ChasJC23: Thanks very much!

If you don't mind me asking, what do you mean by the changes to the decorators? I'm assuming switching `get_dtypes("float")` to `get_dtypes("valid")` will refer to the types provided by the supported dtypes decorator or something else trustworthy, but I'm not sure which decorator "supported dtypes" refers to.

AnnaTz: `@with_supported_dtypes` (or `@with_supported_device_and_dtypes` in case the supported dtypes are different across devices)

ChasJC23: Thanks very much! I've since added those changes, although I might have fumbled my use of git somewhat.

AnnaTz (commented):
Hi @ChasJC23, the code looks good overall but I see there are some errors in the numpy, torch and paddle backends. Please have a look at those. If they are minor arithmetic differences you could try setting the relative tolerance or limiting the value range of the arrays in the test (as it's done in `test_igamma`).

AryanSharma21:
What I see in the test case is that in case of numpy it is actually the case of precision. it says :
E   AssertionError:  the results from backend numpy and ground truth framework tensorflow do not match
E    1.0000112056732178!=1.0 


But for paddle the result is very different. 

E   AssertionError:  the results from backend paddle and ground truth framework tensorflow do not match
E    inf!=0.8427007929497151

Daniel4078:
> What I see in the test case is that in case of numpy it is actually the case of precision. it says : E AssertionError: the results from backend numpy and ground truth framework tensorflow do not match E 1.0000112056732178!=1.0
> 
> But for paddle the result is very different.
> 
> E AssertionError: the results from backend paddle and ground truth framework tensorflow do not match E inf!=0.8427007929497151

The first one can be solve by setting a larger tolerance(default is 10^-7) in the test function, the second one seems to be a problem in the backend implementation for the paddle backend.

AnnaTz:
If you copy the test settings of `test_igamma` the test passes.

ChasJC23:
Thanks for the help! I was a bit cautious about fiddling with the tolerances to make the tests pass; that just feels like a bad idea generally. I appreciate the observation of the config used in `test_igamma`, it makes me more at ease knowing the config in use isn't a bunch of numbers I've pulled out of thin air.

AnnaTz (approved):
Looks good! Thanks!

PR #19531, opened 2023-07-17
Fix documentation for Ivy.to_native() function

### LINKED ISSUE #19530: [Documentation]: Ivy.to_native() function expects ivy.Array object, but accepts torch.Tensor object in documentation
### Path

 https://unify.ai/docs/ivy/overview/design/building_blocks.html

### Section

Building Blocks

### Actual Behavior

The documentation for the Ivy.to_native() function states that it expects to be passed an Ivy.Array object. However, the code snippet in the documentation passes a Torch.Tensor object to the function. This results in an error, as the Ivy.to_native() function cannot convert Torch.Tensor objects to native objects.

### Expected Behavior

The documentation should be updated to state that the Ivy.to_native() function only works with Ivy.Array objects. Additionally, the code snippet in the documentation should be updated to pass an Ivy.Array object to the function.

### Screenshots

Screenshot (56)

shephinphilip:
Title: Fix documentation for Ivy.to_native() function

Issue Number: close #19530


.github/workflows/test-ivy-nn.yml
```diff
@@ -73,16 +73,16 @@ jobs:
           mkdir -p examples
         continue-on-error: true
 
-       - name: Run Functional-NN Tests
```
shephinphilip, line 76: indendation error , wherever moved , raising the same error

shephinphilip (commented):
indendation error , wherever moved , raising the same error

Ookamice:
@shephinphilip 
Hiya, I'm closing your PR and issue for now as I see no problems with the documentation.
```python
import torch
t = torch.tensor([23,42, -1])
type(ivy.to_native(ivy.sum(t)))
```
I believe your concern is this section of example code where you mentioned we are passing a torch.Tensor into ivy.to_naive(). However, because we applied ivy.sum, we would get an ivy.Array out meaning there is nothing strictly wrong with this usage.

Feel free to reopen your issue and PR if there's something I have missed! 🚀

PR #19564, opened 2023-07-18
trapz

### LINKED ISSUE #19376: trapz

dash96:
close #19376


Ookamice:
@dash96
I think there may be some non-transparent casting within the ivy.trapz function that for certain backends may cast it to a different dtype before returning.  A type-cast to the original dtype before returning the result should solve the error.

PS: For `available_dtypes=helpers.get_dtypes("float"), min_num_dims=1, max_num_dims=1` , please use `"valid"` instead of `"float"`

dash96:
I tried a unit test for the current ivy.trapz() and np.trapz(). It is failing for every inputs I gave so far. I tried fixing it but kept running into new errors after every fix. That is when I realised it was better to build a new function instead. 

However, I couldn't solve this one error "AssertionError: returned dtype = float16, ground-truth returned dtype = float64"

Ookamice:
@dash96 
My original thought would be that you could still use the `ivy.trapz()` function but just store the original dtype before passing it through the ivy function and then casting to it before returning.

However, I am concerned that the current `ivy.trapz()` is failing its unit tets. I'll see if it is working on my end but I would highly recommend using `ivy.trapz()` rather than reimplementing it through a compositional approach.

Ookamice:
Hiya, I just ran some tests on `ivy.trapz()` on github codespaces with `python -m pytest -v "ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py::test_trapz"` and it seems to be functioning correctly. I would advise following the approach i set out about storing the input dtype, using `ivy.trapz()` and then casting it back before returning.

dash96:
Please let me know if I am missing anything here,

1) I created a .py file and imported numpy and ivy
2) assigned an array y=[0, 1, 2, 3] and x=[2,4,6,8]
3) now tried ivy.trapz(y,x) and numpy.trapz(y,x) 

**Problem** 1): encountering the error that ivy.trapz() can only take one *args and zero **kwargs
--> Due to this what ever I do with x is not reflected in ivy.trapz() 
But then to dig deeper I gave "x=[1,2,3,4]" and for a different example "x=[ ] and dx = 1"and those are the only two cases for which values for numpy and ivy match. 
**conclution:** For ivy.trapz() the values for x and dx are fixed intrinsically
**Solution**: Knowing how trapezoidal rule works. Tried different arrays for x with linear spread of values within the array and fixed the function by multiplying dx with ivy.trapz() where dx=x[1]-x[0] 

But the above solution failed to work for non-linear x. The fix then was x = ivy.sum(ivy.diff(x)) after few more fixes.

I found my self back on square one"AssertionError: returned dtype = float16, ground-truth returned dtype = float64"

Ookamice:
@dash96 
Hiya, I think this is because Ivy tries to follow the array API standard so if you see the argument signature of [ivy.trapz](https://unify.ai/docs/ivy/docs/functional/ivy/elementwise/ivy.functional.ivy.elementwise.trapz.html), x is a keyword only variable where as y is a positional only variable. (In general, variables before `\` are positional only and those after `*` are keyword only). This differs from [numpy.trapz](https://numpy.org/doc/stable/reference/generated/numpy.trapz.html) which gives you a lot more flexibility.

If you dig deeper into the ivy code, `ivy.trapz` should be implemented simiarily to `current_backend.trapz()` and then in the numpy-backend for example, `np_backend.trapz()` would then call `numpy.trapz` so if your backend is set to numpy, it should behave similarily/identically with the exception that some variables are forced to be positional/keyword.

ivy-seed:

PR #19565, opened 2023-07-18
Fix ivy and torch strides for non-contiguous arrays

AnnaTz:
```
import torch
from ivy.functional.frontends import torch as torch_frontend
import ivy
import numpy as np

x = np.random.rand(12, 1, 64, 512)
size = [12, 1, 512, 64]
stride = [64, 196608, 768, 1]

for b in ['numpy', 'torch', 'tensorflow', 'jax', 'paddle']:
    ivy.set_backend(b)

    trans_x = torch_frontend.tensor(x).transpose(-2, -1)
    front = torch_frontend.as_strided(trans_x, size, stride).ivy_array.to_numpy()

    trans_x = torch.tensor(x).transpose(-2, -1)
    orig = torch.as_strided(trans_x, size, stride).detach().numpy()

    print(np.equal(orig, front).max())
```
```
from ivy.functional.frontends import torch as torch_frontend
import ivy

gt = torch.randn((2, 5, 5)).permute(1, 2, 0).stride()

for b in ["numpy", "jax", "torch", "tensorflow", "paddle"]:
    ivy.set_backend(b)

    x = torch_frontend.randn((2, 5, 5)).permute(1, 2, 0)
    
    print(np.equal(x.stride(), gt).max())
```


ivy-leaves:

Thank you :hugs:

vedpatwardhan (changes requested):
Hey @AnnaTz, thanks for diving into this issue! I think even if we revert all changes to `to_numpy` (backends and ivy), and update the `strides` function compositional implementation to be the following instead,
```
    # if x is an ivy array with a base,
    # convert it to a numpy array with the same base:
    if ivy.is_ivy_array(x) and x.base is not None:
        x = x.base
    return ivy.to_numpy(x).strides
```
it seems to work consistently with the examples you've shared 😅
Thanks 😄

AnnaTz:
> Hey @AnnaTz, thanks for diving into this issue! I think even if we revert all changes to `to_numpy` (backends and ivy), and update the `strides` function compositional implementation to be the following instead,
> 
> ```
>     # if x is an ivy array with a base,
>     # convert it to a numpy array with the same base:
>     if ivy.is_ivy_array(x) and x.base is not None:
>         x = x.base
>     return ivy.to_numpy(x).strides
> ```
> 
> it seems to work consistently with the examples you've shared 😅 Thanks 😄

No, actually that wouldn't be correct.
`np.ones((2,5,5), dtype=np.float32).strides` is `(100,20,4)`
`np.ones((2,5,5), dtype=np.float32).transpose((1, 2, 0)).base.strides` is `(100,20,4)`
but `np.ones((2,5,5), dtype=np.float32).transpose((1, 2, 0)).strides` is `(20,4,100)`

So what you suggested gives the same result for all backends, but not the correct one. 
I will update the PR to show this.

vedpatwardhan (commented):
> No, actually that wouldn't be correct. `np.ones((2,5,5), dtype=np.float32).strides` is `(100,20,4)` `np.ones((2,5,5), dtype=np.float32).transpose((1, 2, 0)).base.strides` is `(100,20,4)` but `np.ones((2,5,5), dtype=np.float32).transpose((1, 2, 0)).strides` is `(20,4,100)`
> 
> So what you suggested gives the same result for all backends, but not the correct one. I will update the PR to show this.

Yep you're right, it's consistent but not correct! Thanks for reverting the changes to `to_numpy` and moving the logic to `strides`, certainly helps avoid running unnecessary logic when `ivy.strides` isn't being used. 
I think we can also just call `ivy.to_numpy` instead of trying to infer the backend, because if the array `x` belongs to a different backend than the one set currently, then the function isn't expected to work correctly anyway.
Also, we shouldn't set the global backend in the function, because it changes the ids of all `ivy` functions globally (as they are wrapped again) which would break compilation. We can instead use `with_backend`. Even though we shouldn't be using any backend setting function inside an `ivy` function, we'd have to make an exception for this function, as there doesn't seem to be any other feasible way to compute strides correctly.
So making the code a bit more compact, we can probably implement `strides` as,
```
    if ivy.is_native_array(x) or (ivy.is_ivy_array(x) and x.base is None):
        return ivy.to_numpy(x).strides
    # if x is an ivy array with a base,
    # convert it to a numpy array with the same base:
    ret = ivy.to_numpy(x.base)
    ivy_numpy = ivy.with_backend('numpy')
    for fn, args, kwargs, index in x._manipulation_stack:
        ret = ivy_numpy.__dict__[fn](ret, *args, **kwargs)
        ret = ret[index] if ivy.exists(index) else ret
    return ret.to_native().strides
```
Also, I think it would be great to add this to the deep dive inside the "Views" section, pointing why we had to defer to `numpy` for computing `strides`, as this is an exception where we are using a specific backend inside a function.
Happy to know what your thoughts, thanks @AnnaTz 😄

AnnaTz:
> > No, actually that wouldn't be correct. `np.ones((2,5,5), dtype=np.float32).strides` is `(100,20,4)` `np.ones((2,5,5), dtype=np.float32).transpose((1, 2, 0)).base.strides` is `(100,20,4)` but `np.ones((2,5,5), dtype=np.float32).transpose((1, 2, 0)).strides` is `(20,4,100)`
> > So what you suggested gives the same result for all backends, but not the correct one. I will update the PR to show this.
> 
> Yep you're right, it's consistent but not correct! Thanks for reverting the changes to `to_numpy` and moving the logic to `strides`, certainly helps avoid running unnecessary logic when `ivy.strides` isn't being used. I think we can also just call `ivy.to_numpy` instead of trying to infer the backend, because if the array `x` belongs to a different backend than the one set currently, then the function isn't expected to work correctly anyway. Also, we shouldn't set the global backend in the function, because it changes the ids of all `ivy` functions globally (as they are wrapped again) which would break compilation. We can instead use `with_backend`. Even though we shouldn't be using any backend setting function inside an `ivy` function, we'd have to make an exception for this function, as there doesn't seem to be any other feasible way to compute strides correctly. So making the code a bit more compact, we can probably implement `strides` as,
> 
> ```
>     if ivy.is_native_array(x) or (ivy.is_ivy_array(x) and x.base is None):
>         return ivy.to_numpy(x).strides
>     # if x is an ivy array with a base,
>     # convert it to a numpy array with the same base:
>     ret = ivy.to_numpy(x.base)
>     ivy_numpy = ivy.with_backend('numpy')
>     for fn, args, kwargs, index in x._manipulation_stack:
>         ret = ivy_numpy.__dict__[fn](ret, *args, **kwargs)
>         ret = ret[index] if ivy.exists(index) else ret
>     return ret.to_native().strides
> ```
> 
> Also, I think it would be great to add this to the deep dive inside the "Views" section, pointing why we had to defer to `numpy` for computing `strides`, as this is an exception where we are using a specific backend inside a function. Happy to know what your thoughts, thanks @AnnaTz 😄

Thanks Ved! I didn't know about `ivy.with_backend`, that's exactly what we needed here 😄

AnnaTz:
> Also, I think it would be great to add this to the deep dive inside the "Views" section, pointing why we had to defer to numpy for computing strides, as this is an exception where we are using a specific backend inside a function.

Not really sure what to write about this 😅 Would you mind doing that? I think you've got a better understanding of ivy views anyways.

vedpatwardhan:
> Not really sure what to write about this 😅 Would you mind doing that @vedpatwardhan? I think you've got a better understanding of ivy views anyways.

Just added a `note` to `inplace_updates` in the deep dive, feel free to merge the PR based on what @Sam-Armstrong thinks. Thanks @AnnaTz 😄

PR #19683, opened 2023-07-20
Implemented testing for copy

KevinUli:


AnnaTz (changes requested):
Hi @KevinUli, this looks good overall but you will need to merge master and resolve the conflicts. There's been a refactor in function_testing.py regarding the backend handling so make sure to keep that in mind. Thanks!

ivy_tests/test_ivy/helpers/test_parameter_flags.py
```diff
@@ -43,6 +43,7 @@ def _compile_strategy(draw):  # TODO remove later when paddle is supported
 BuiltInplaceStrategy = st.just(False)
 BuiltGradientStrategy = _gradient_strategy()
 BuiltWithOutStrategy = st.booleans()
+BuiltWithCopyStrategy = st.just(False)
```
CatB1t, line 46: Why does this have a default of False?

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -309,6 +314,18 @@ def test_function(
         target_fn, *args, test_compile=test_flags.test_compile, **kwargs
     )
 
+    if test_flags.with_copy:
+        array_fn = ivy.is_array
```
CatB1t, line 318: We shouldn't use global Ivy here.

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -309,6 +314,18 @@ def test_function(
...
+        if "copy" in list(inspect.signature(target_fn).parameters.keys()):
```
CatB1t, line 319: We shouldn't do this kind of search in arguments for each example, but instead this logic should be moved into the decorator where this is done only once.

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -667,6 +685,29 @@ def test_frontend_function(
                 else:
                     assert ret.data is out.data
             assert ret is out
+    elif test_flags.with_copy:
+        assert not isinstance(ret, tuple)
+
+        if test_flags.generate_frontend_arrays:
+            assert _is_frontend_array(ret)
+        else:
+            assert ivy.is_array(ret)
```
CatB1t, line 694: Why is is this assertion specific to `test_flags.with_copy`?

CatB1t (changes requested):

ivy_tests/test_ivy/helpers/test_parameter_flags.py
KevinUli: I assumed only one inplace update can be True at a time. Since with_out defaults to both True and False, I had with_copy defaults as False so that it doesn't interfere

ivy_tests/test_ivy/helpers/function_testing.py
KevinUli: This assertion was only available in `test_flags.inplace` I didn't know whether I should refactor it or add the same assertion to `test_flags.with_copy`.

ivy_tests/test_ivy/helpers/function_testing.py
KevinUli: Alright. I'll have it changed to `ivy_backend`

ivy_tests/test_ivy/helpers/function_testing.py
KevinUli: By decorator, do you mean the frontend `inputs_to_ivy_arrays` decorator?

AnnaTz:
Hey @sherry30, could you review this? We have dragged this for 2 months now 😅 And I would say it's an important feature missing from the pipeline.

sherry30 (changes requested):
The changes mostly looks good.
Can you add this flag to the functions which require it.

ivy-seed:

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -264,6 +264,22 @@ def target_fn(instance, *args, **kwargs):
...
+                *args, array_fn=array_fn, **kwargs
+            )
+            ret_, ret_np_flat_ = get_ret_and_flattened_np_array(
+                fw,
+                target_fn,
+                *args,
+                test_compile=test_flags.test_compile,
+                **kwargs,
+            )
+            assert not np.may_share_memory(first_array, ret_)
```
AnnaTz, line 281: I think we need to assert that `ret_` is not a `tuple` here as is done in the frontend case below.

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -872,6 +896,31 @@ def test_frontend_function(
                 ):
                     assert ret.ivy_array.data is out.ivy_array.data
                 assert ret is out
+        elif test_flags.with_copy:
+            assert not isinstance(ret, tuple)
+
+            if test_flags.generate_frontend_arrays:
+                assert _is_frontend_array(ret)
+            else:
+                assert ivy_backend.is_array(ret)
```
AnnaTz, line 905: Arrays returned by frontend functions are always frontend arrays. This is being asserted [here](https://github.com/unifyai/ivy/blob/bae0fdfe24630f466ea5a474b8576d4ca717f502/ivy_tests/test_ivy/helpers/function_testing.py#L844). So this `if` statement can be replaced by a simple `assert _is_frontend_array(ret)`.

AnnaTz (changes requested):
Hi @KevinUli, apologies for this PR taking so long to merge. I will review it so we can finish it asap.

ivy_tests/test_ivy/helpers/function_testing.py
AnnaTz: Also, since you are asserting that `ret` is an array, I don't think you need to assert it is not a tuple as well.

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -872,6 +896,31 @@ def test_frontend_function(
...
+                copy_kwargs["copy"] = True
+            first_array = ivy_backend.func_wrapper._get_first_array(
+                *copy_args, array_fn=array_fn, **copy_kwargs
+            )
+            ret_ = get_frontend_ret(
+                backend_to_test, frontend_fn, *copy_args, **copy_kwargs
+            )
+            if _is_frontend_array(first_array):
+                first_array = first_array.ivy_array
+            if _is_frontend_array(ret_):
```
AnnaTz, line 921: This `if` is not necessary either, since we already asserted that the return is a frontend array.

AnnaTz (changes requested):

AnnaTz (approved):
Looks good, thanks!

PR #19694, opened 2023-07-20
EHN: adding logspace in paddle creation.py

### LINKED ISSUE #19689: logspace

MuhammadNizamani:
Closed #19689


Sam-Armstrong (approved):
@MuhammadNizamani thanks for making the PR! I've updated the test to use float values for start and stop, as I feel this is more complete. With this change the test with the backend paddle is failing, but this is caused by ivy.linspace and is also occurring with the torch linspace and logspace frontend functions. So I think we're fine to get this merged. Great work! 😊

PR #19706, opened 2023-07-20
fixing ivy can_cast 

### LINKED ISSUE #19704: [Bug]: ivy.can_cast
### Bug Explanation

 Function always returns True.

### Steps to Reproduce Bug

```
>>> import numpy as np
>>> import ivy
>>> ivy.can_cast(ivy.float64,ivy.int32)
True
>>> np.can_cast(np.float64,np.int32)
False
```

### Environment

Colab

### Ivy Version

v1.1.9

### Backend

- [ ] NumPy
- [ ] TensorFlow
- [ ] PyTorch
- [ ] JAX

### Device

_No response_

ShreyanshBardia:
closes #19704

### DIFF
--- a/ivy/functional/ivy/data_type.py
+++ b/ivy/functional/ivy/data_type.py
@@ -559,8 +559,8 @@ def can_cast(
     if isinstance(from_, (ivy.Array, ivy.NativeArray)):
         from_ = from_.dtype
     try:
-        ivy.promote_types(from_, to)
-        return True
+        dtype = ivy.promote_types(from_, to)
+        return dtype == to
     except KeyError:
         return False
 


ShreyanshBardia:
@sagewhocodes can you check this? 
The test cases that are failing are I think the one that deviates from array_api standard. Ivy seems to follow numpy standard of promotion that allows bool to be casted to int8.

ShreyanshBardia:
@MahmoudAshraf97 the way that numpy has solved this is by not wrapping the np.can_cast function directly but rewriting it under array_api. Here's the relevant PR https://github.com/numpy/numpy/pull/20883

ShreyanshBardia:
Hi @MahmoudAshraf97, can you take a look

PR #19707, opened 2023-07-20
Made shape class to inherit the Sequence class for running array API …

hirwa-nshuti:
Hey @Aarsh2001 

I was running the array API tests and they were failing because even our `ivy.Shape` class can be iterated over, it isn't of `Sequence` type so I made our `Shape` class inherit from the `Sequence` class to make array API tests running collectively😊

Any input on this I can still work around this corner to start straight away with the Array API tests skipping task given that they'll be running now😆

And this can be considered to be a superset as only TensoFlow shapes aren't instances of `Sequence` class.

```python
import numpy as np
import torch
import tensorflow as tf
import paddle
import jax
import ivy
from collections.abc import Sequence

x_np = np.array([[1, 2, 3], [4, 5, 6]])
x_torch = torch.tensor([[1, 2, 3], [4, 5, 6]])
x_tf = tf.constant([[1, 2, 3], [4, 5, 6]])
x_paddle = paddle.to_tensor([[1, 2, 3], [4, 5, 6]])
x_jax = jax.numpy.array([[1, 2, 3], [4, 5, 6]])
x_ivy = ivy.array([[1, 2, 3], [4, 5, 6]]) # After this fix

def is_sequence(x):
    return isinstance(x, Sequence)
for i in [x_np, x_torch, x_tf, x_paddle, x_jax, x_ivy]:
    print(is_sequence(i.shape))

# True
# True
# False
# True
# True
# True

```

Thanks


Aarsh2001 (approved):
Yeah sounds good, @vedpatwardhan any issues if we do this?

vedpatwardhan:
Hey @Aarsh2001, I don't think there will be any issues as a result of inheriting `Sequence`, as long as we correctly override all relevant methods of the `Sequence` class in `ivy.Shape`. Thanks for diving into this @hirwa-nshuti 😄

PR #19724, opened 2023-07-20
Jax devicearray prod

### LINKED ISSUE #19417: prod

VictorOdede:
Close #19417


ivy_tests/test_ivy/test_frontends/test_jax/test_devicearray.py
```diff
@@ -510,6 +510,45 @@ def test_jax_devicearray_nonzero(
     )
 
 
+@handle_frontend_method(
+    class_tree=CLASS_TREE,
+    init_tree="jax.numpy.array",
+    method_name="prod",
+    dtype_x_axis=helpers.dtype_values_axis(
+        available_dtypes=["int64"],
```
hirwa-nshuti, line 518: I can't fully understand why this is the only supported dtype

hirwa-nshuti (changes requested):
Hey @VictorOdede 

Thanks for the PR, all looks great just a few changes are needed

ivy_tests/test_ivy/test_frontends/test_jax/test_devicearray.py
VictorOdede: Hey @hirwa-nshuti , thanks for the review. I've also tried testing with other valid dtypes but the ground-truth is always returning `int64` causing the tests to fail with an `AssertionError` since returned dtype is different from ground truth.  I haven't encountered this problem before, is there a way to solve this issue?

hirwa-nshuti: Then if there is a special promotion being done should be handled in the frontend function not on the testing side to ensure that we cover all the corner cases

VictorOdede:
Hey @hirwa-nshuti , I've added type promotion for all `int` and `uint` dtypes. Now all tests are passing except for `Paddle` which sometimes fails because it doesn't support `uint64`

ivy/functional/frontends/jax/devicearray.py
```diff
@@ -123,6 +123,22 @@ def nonzero(self, *, size=None, fill_value=None):
             fill_value=fill_value,
         )
 
+    def prod(self, axis=None):
+        arr = self
```
hirwa-nshuti, line 127: From my understandings is that the promotion being done here should be done on the `jax.numpy.product` function as we can promote all `unsigned integers` to `uint64` by a simple `ivy.is_uint_dtype` and `ivy.is_int_dtype` functions.

Here we will simply need to call the frontend function once we ensured that it is working correctly🙂

hirwa-nshuti (changes requested):
Hey
Just a few suggested changes🙂
Keep up the good work😊

ivy/functional/frontends/jax/devicearray.py
VictorOdede: Thanks for the great pointers. Will look into this asap.

VictorOdede:
Hey @hirwa-nshuti , the tests are passing for the most part but sometimes it fails for higher precision dtypes like this: 
jax_error
Is this the expected behaviour?

hirwa-nshuti (changes requested):
For precision errors handling you can use the torelance values mentioned in the helper function 

https://github.com/unifyai/ivy/blob/356e6e5654970ee4b6f7cb9b41b3b572436e1262/ivy_tests/test_ivy/helpers/function_testing.py#L485C4-L485C4

VictorOdede:
> For precision errors handling you can use the torelance values mentioned in the helper function
> 
> https://github.com/unifyai/ivy/blob/356e6e5654970ee4b6f7cb9b41b3b572436e1262/ivy_tests/test_ivy/helpers/function_testing.py#L485C4-L485C4

Thanks! Will look into this.

ivy/functional/frontends/jax/numpy/mathematical_functions.py
```diff
@@ -695,6 +695,13 @@ def product(
     promote_integers=True,
     out=None,
 ):
+    if ivy.is_uint_dtype(a.dtype):
```
hirwa-nshuti, line 698: These are only supported when the `promote_integers` is `True`

hirwa-nshuti: We can redefine prod function as
```python
@to_ivy_arrays_and_back
def product(
    a,
    *,
    axis=None,
    dtype=None,
    keepdims=False,
    initial=None,
    where=None,
    promote_integers=True,
    out=None,
):
    if ivy.is_array(where):
        a = ivy.where(where, a, ivy.default(out, ivy.ones_like(a)), out=out)
    if promote_integers:
        if ivy.is_uint_dtype(a.dtype):
            dtype = "uint64"
        elif ivy.is_int_dtype(a.dtype):
            dtype = "int64"
    if initial is not None:
        if axis is not None:
            s = ivy.to_list(ivy.shape(a, as_array=True))
            s[axis] = 1
            header = ivy.full(ivy.Shape(tuple(s)), initial)
            a = ivy.concat([header, a], axis=axis)
        else:
            a[0] *= initial
    return ivy.prod(a, axis=axis, dtype=dtype, keepdims=keepdims, out=out)
    ```

hirwa-nshuti (changes requested):
Hey,
Great work so far I have added the suggestions in the comments below
Thanks

VictorOdede:
> Hey, Great work so far I have added the suggestions in the comments below Thanks

Thanks, will work on implementing this

hirwa-nshuti (approved):
LGTM!
Going to merge

PR #19820, opened 2023-07-21
ELU added to Ivy Stateful Api

tejeshbhalla:
#19747


tejeshbhalla:
@vedpatwardhan  in the sub_backends elaboration it was failing due to it opening .ds_store on my macbook as it is not a directory and that was the fix i made

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -22,6 +22,8 @@
 ):
     if backend.startswith("__"):
         continue
+    if backend.endswith(".DS_Store"):
+        continue
```
CatB1t, line 26: We don't need new line as a check, also it's better to create a seperate issue and PR to get this fixed, @Infrared1029 can provide more insight on a better solution than hardcoding `DS_STORE` into there.

CatB1t (changes requested):

ivy/utils/backend/sub_backend_handler.py
tejeshbhalla: Sure

vedpatwardhan (commented):
I completely agree with what @CatB1t said, let's see what @Infrared1029 thinks. Thanks @tejeshbhalla 😄

tejeshbhalla:
@Mr-Niraj-Kulkarni could you explain whats wrong with the tests ?

Mr-Niraj-Kulkarni:
> @Mr-Niraj-Kulkarni could you explain what's wrong with the tests?
Everything looks good to me according to the code.

Mr-Niraj-Kulkarni (approved):

PR #19907, opened 2023-07-23
Add adjust_brightness to transforms.py

### LINKED ISSUE #19906: adjust_brightness
- [ ] #15046

alec-kr:
Closes #19906.

This PR is my draft attempt to begin fixing the aforementioned issue. I have added a call to ```paddle.vision.transforms.adjust_brightness(img, brightness_factor)``` in the new function.

However, I have not added any tests as yet, since I am unsure where these should be written/placed. I am looking forward for some guidance/assistance with getting this function and PR merged, so that I can get a better understanding of the project.


juliagsy (changes requested):
Hello! thanks for contributing! As this is a frontend function, you should link it to the equivalent ivy function. However, as the `adjust_brightness` function is not yet available in the ivy api, you should either implement this frontend function compositionally or propose the function to be added to the ivy api first. More info here: https://unify.ai/docs/ivy/overview/contributing/open_tasks.html#frontend-apis
Thanks!

juliagsy:
Hey @alec-kr ! Apologies for the delayed response! The file you're putting it in is correct, but the impleementation shouldn't involve the native paddle function, and should be using functions from the ivy api. Hope this makes sense!

alec-kr:
Thank you for the tip @juliagsy. I am having some difficulty figuring out what ivy function can replace 
```paddle.vision.transforms.adjust_brightness```, since I am calling this from ```ivy.functional.frontends.paddle```.

Can you kindly provide some guidance?

juliagsy:
Hey! So forexample you can ref the native implementation: https://github.com/PaddlePaddle/Paddle/blob/fc177d2553ecfb19ab350b4c892c4b571b8b2652/python/paddle/vision/transforms/functional_pil.py#L316
You might need to dig a bit deeper into their implementation or functions involved behind-the-scenes to get the matching results, thanks!

alec-kr:
Okay thanks for the tip. I've just made a few changes to remove the calls to the native paddle function.

juliagsy:
Hey! Thanks for the changes! PIL functions should not be present in the frontend as well. For example, if their adjust brightness function involves adding and multiplying factor, you'll need to do it compositionally with `ivy.add`, `ivy.matmul`, etc. Hope this makes sense, thanks!

alec-kr (commented):
Refactored the brightness adjustment function to use array operations from the ivy library.

PR #20208, opened 2023-07-24
Added torch front end instance for bmm

### LINKED ISSUE #20116: bmm

BebehCodes:
Implemented the torch front end instance for bmm and also its unit test.

When I tried to run the unit test (which is located in test_tensor.py) I always came across the following error for each unit test in that module:
image

I noticed that in torch backend creation.py and torch front end creation_ops.py there wasn't a ones_like function implemented (only a couple with names such as ones_like_v_0p1p12_to_0p2p0), so I implemented them in order to be able to run my tests. 

Is this a general problem that anyone else has come across? 
Should I create a different PR with these additions or add them here?

EDIT: After creating the PR I noticed that this error persists during the Github actions checks too, should I include my solution here?

Close #20116


BebehCodes:
The key error was fixed after discussing about it on discord

BebehCodes:
Hello @Ishticode , could you have a look at the pull request? Thank you!

Ishticode (approved):
@BebehCodes 
Thanks for the PR. I have tested locally and the tests pass. Thank you very much. i will get this merged. Thanks

PR #20267, opened 2023-07-24
added new exception class and altered array class for the exception

mosesdaudu001:
Added new exception class and modified `ivy.Array.__getattr__` to raise exception when backend is changed except when `dynamic = True`


ivy/data_classes/array/array.py
```diff
@@ -393,6 +394,10 @@ def __getattribute__(self, item):
         return super().__getattribute__(item)
 
     def __getattr__(self, item):
+        if ivy.current_backend() != _determine_backend_from_args(self._data):
+            raise ivy.utils.exceptions.InvalidBackendException(
+                "This operation is only allowed when dynamic_backend = True"
```
vedpatwardhan, line 399: I think this message could end up being a bit misleading, as not using dynamic backend setting is one of the reasons that could lead to this issue but there's other possible reasons as well. ig an appropriate message would be something along the lines of `The current backend is {ivy.current_backend} while the ivy.Array used is based on the {array_backend} backend. Please set dynamic_backend=True for the array if you want to convert it to the target backend`

vedpatwardhan (changes requested):
Hey @mosesdaudu001, thanks for diving into this, just requested a small change 😄

ivy/data_classes/array/array.py
mosesdaudu001: I also agree on that as well @vedpatwardhan 
I changed the error message in the latest commit

vedpatwardhan (changes requested):
Hey @mosesdaudu001, the changes made seem to accurately capture the case when `__getattr__` gets called, but there seem to be cases when `__getattr__` doesn't get called which aren't getting checked. For e.g. in the example below,
```
import ivy


ivy.set_backend('torch')
x = ivy.array([1., 2., 3.])
query = ivy.array([0, 1])
val = ivy.array([10, 10])
ivy.set_backend('jax')
x[query] = val
```
The above defers to `ivy.Array.__setitem__` which doesn't defer to `__getattr__` and as a result this case isn't captured.
Then there are other cases where we call `ivy` functions which doesn't have wrappers that implcitly use `__getattr__` and adding this logic to `__getattribute__` instead wouldn't be ideal, because we still want to allow `__repr__` and `data` and a few such properties even if the backend isn't the same as that of the array.
We'll probably need to go by the function wrapper approach to deal with all this, so that we can ensure that every `ivy` function receives arguments compatible to the set backend. This will probably be the cleanest (though relatively tedious) approach to solve this problem. Happy to know what you think @mosesdaudu001 @CatB1t, thanks 😄

CatB1t:
> Hey @mosesdaudu001, the changes made seem to accurately capture the case when `__getattr__` gets called, but there seem to be cases when `__getattr__` doesn't get called which aren't getting checked. For e.g. in the example below,
> 
> ```
> import ivy
> 
> 
> ivy.set_backend('torch')
> x = ivy.array([1., 2., 3.])
> query = ivy.array([0, 1])
> val = ivy.array([10, 10])
> ivy.set_backend('jax')
> x[query] = val
> ```
> 
> The above defers to `ivy.Array.__setitem__` which doesn't defer to `__getattr__` and as a result this case isn't captured. Then there are other cases where we call `ivy` functions which doesn't have wrappers that implcitly use `__getattr__` and adding this logic to `__getattribute__` instead wouldn't be ideal, because we still want to allow `__repr__` and `data` and a few such properties even if the backend isn't the same as that of the array. We'll probably need to go by the function wrapper approach to deal with all this, so that we can ensure that every `ivy` function receives arguments compatible to the set backend. This will probably be the cleanest (though relatively tedious) approach to solve this problem. Happy to know what you think @mosesdaudu001 @CatB1t, thanks smile

Totally agree with you! ideally this should only happen when we try to change the underlying tensor or do operation on that tensor, read-only properties should work fine with  any bakcend.

ivy/func_wrapper.py
```diff
@@ -556,6 +557,26 @@ def _outputs_to_ivy_arrays(*args, **kwargs):
     return _outputs_to_ivy_arrays
 
 
+def handle_backend_invalid(fn: Callable) -> Callable:
+    @functools.wraps(fn)
+    def _handle_backend_invalid(*args, **kwargs):
+        from ivy.utils.backend.handler import _determine_backend_from_args
```
CatB1t, line 563: do we need to import that lazily? This will add a bit of an overhead when calling the function.

ivy/func_wrapper.py
```diff
@@ -556,6 +557,26 @@ def _outputs_to_ivy_arrays(*args, **kwargs):
...
+
+        for arg in args:
+            if ivy.current_backend() != _determine_backend_from_args(arg):
```
CatB1t, line 566: I think greedly searching through all arguments may not be the best thing here, can we use the type hints (If exist) as a guided search?

ivy/func_wrapper.py
```diff
@@ -556,6 +557,26 @@ def _outputs_to_ivy_arrays(*args, **kwargs):
...
+                raise ivy.utils.exceptions.InvalidBackendException(
+                    f"The current backend is {ivy.current_backend()} while the"
+                    " ivy.Array used is based on the"
+                    f" {_determine_backend_from_args(arg)} backend."
```
CatB1t, line 570:
```suggestion
                    f"Operation not allowed. Array was instantiated with backend {_determine_backend_from_args(arg)}."
                    " But current backend is {ivy.current_backend_str()}."
```

CatB1t, line 570: We can cache the result from `_determine_backend_from_args` instead of calling it once more.

CatB1t (changes requested):

Issue #9998, opened 2023-01-25
Fix Functional.core tests


DragosStoican:
[See dashboards here](https://github.com/unifyai/ivy/tree/dashboard/test_dashboards/Functional%20API/Core)

PR #20573, opened 2023-07-24
Finally added jax.numpy.apply_along_axis

### LINKED ISSUE #19726: apply_along_axis
Ivy Implementation of ``jax.numpy.apply_along_axis()``

yashjha123:
close #19726


yashjha123:
Consider checking the original jax's implementation of the [apply_along_axis](https://jax.readthedocs.io/en/latest/_modules/jax/_src/numpy/lax_numpy.html#apply_along_axis ) function.

D0m-inic:
Make sure to add the `backend_to_test` parameter to your tests. Your test is also failing for some cases - make sure to account for possible edge cases (inputted array or any parameters being 0).

yashjha123:
Hey, I have made those changes. Take a took

D0m-inic:
Hi, could you take a look at the tests failing, they seem to be caused by a wrong selected DataType. Also, you can remove the conditionals checking if `axis` is a tuple or None within the test, since you're using a composite you know that the axis will be valid.

As well as this, you can remove the type hints within the function (the `:Callable` and `:int`),  for frontend functions we typically refrain from them to keep our style consistent 😄

yashjha123:
Hey @D0m-inic,, Thanks for the review. I've made all the changes you suggested. The PR is now passing all the tests.

yashjha123:
Hey @D0m-inic ,
After the tests started to fail for the paddle backend, I tried to dug deeper into the issue. After thorough debugging, I discovered slightly different results for the paddle concat operation, leading to issues with the vmap function and thus my code.
There's an automated issue on this already: #11262 #9998  under the ``ToDo_internal`` tag. I can't progress until these get resolved, Can I resolve this issue as an ``external`` member?

MahmoudAshraf97:
Hi @yashjha123 , I don't think it's wise to revert these changes without actually knowing two things first:
1. whether the call to `vmap` you made in the example is actually correct
2. if the reverting of the commit will still keep the functional tests passing
I'm not an expert with `vmap` tbh but I'd encourage  more investigation into the root cause of the issue

Ishticode:
close this as apply_along_axis is already implementd in jax front
https://github.com/unifyai/ivy/blob/ea402610b7e2cdf77eccf2460e7b6be42e71c2b0/ivy/functional/frontends/jax/numpy/statistical.py#L395

PR #20818, opened 2023-07-25
Fixes to allow typesets in unsupported dtypes and fixed `einops_reduce`

vedpatwardhan:
1. Moved the typesets logic from the `_dtype_device_wrapper_creator` to `_get_dtypes` to ensure that this can be used for functions like `einops_reduce`, which use the `unsupported_dtypes` attribute. 
2. Replaced the `inputs_to_ivy_arrays` decorator in `einops_reduce` by `inputs_to_native_arrays` as `einops.reduce` accepts native arrays.
3. Specified the `unsupported_dtypes` for tensorflow and paddle for `einops_reduce` to get the tests passing


ivy/functional/ivy/data_type.py
```diff
@@ -203,13 +203,26 @@ def _get_dtypes(fn, complement=True):
...
+    }
+
     for key, merge_fn, base in basic:
         if hasattr(fn, key):
             v = getattr(fn, key)
             # only einops allowed to be a dictionary
             if isinstance(v, dict):
                 v = v.get(ivy.current_backend_str(), base)
-
+            for i, d in enumerate(v):
```
RickSanchezStoic, line 223: This is causing a minor bug, where the existing dictionary seems to be getting overwritten. Try replicating the exact operation that was being performed back in the func_wrapper and maybe that fixes it.

ivy/func_wrapper.py
```diff
@@ -1260,36 +1260,6 @@ def _dtype_device_wrapper_creator(attrib, t):
...
-        typesets = {
-            "valid": ivy.valid_dtypes,
-            "numeric": ivy.valid_numeric_dtypes,
-            "float": ivy.valid_float_dtypes,
-            "integer": ivy.valid_int_dtypes,
-            "unsigned": ivy.valid_uint_dtypes,
-            "complex": ivy.valid_complex_dtypes,
-        }
-
-        for key, value in version_dict.items():
```
RickSanchezStoic, line 1272: Try doing the exact operation back in the data_types and it should fix the bug.

RickSanchezStoic (changes requested):
So if you go to jax's choleksy function, and fetch its unsupported_dtypes it will only fetch the complex ones, while ignoring the floats, I believe this is happening because the dictionary is getting overwritten. Try replicating the exact dictionary operation happening back in func_wrappers and it might fix it, though I still have a little doubt about moving the operation entirely there, let's hope this alone fixes it.

ivy/functional/ivy/data_type.py
vedpatwardhan: the bug was basically with the way we were doing the `x[:i] + y + x[i+1:]`, not sure why the same line was working on the `func_wrapper`, but anyway just pushed a fix for that and should be resolved now

ivy/func_wrapper.py
vedpatwardhan: done

vedpatwardhan:
> So if you go to jax's choleksy function, and fetch its unsupported_dtypes it will only fetch the complex ones, while ignoring the floats, I believe this is happening because the dictionary is getting overwritten. Try replicating the exact dictionary operation happening back in func_wrappers and it might fix it, though I still have a little doubt about moving the operation entirely there, let's hope this alone fixes it.

Thanks for reviewing @RickSanchezStoic, just pushed a fix for this and `cholesky` should be working now. Actually we'd need to move the operation to this function if we want to allow typesets for functions like `einops_reduce`. Please take a look and let me know if there's anything else to be done, thanks 😄

vedpatwardhan:
just found another bug, would get that fixed now

vedpatwardhan:
done

RickSanchezStoic:
Haven't done deep looking, but looks good to me otherwise, will see later if any issues arise. You can merge this, thanks!

RickSanchezStoic (approved):

vedpatwardhan:
awesome, thanks @RickSanchezStoic 😄

PR #20824, opened 2023-07-25
Fix bug in Array.__repr__ with implicit backend

jshepherd01:
Before this, when trying to run Ivy with an interactive terminal like iPython, we'd see this behaviour:
```py
In [1]: import ivy
In [2]: ivy.abs(-5)
Out[2]: ... KeyError: 'none'
In [3]: ivy.set_backend('numpy')
Out[3]: <module 'ivy.functional.backends.numpy' from '/home/joe/dev/ivy/ivy/functional/backends/numpy/__init__.py'>
In [4]: ivy.abs(-5)
Out[4]: ivy.array(5)
```
i.e. the functions wouldn't work with an implicit backend. Turns out, this wasn't related to the functions at all, but was a bug in the Array.\_\_repr\_\_ method, since it was expecting the default value of `Ivy.backend` to be `''` rather than `'none'`.

### DIFF
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -367,7 +367,7 @@ class Array(
                 self._post_repr = ")"
         sig_fig = ivy.array_significant_figures
         dec_vals = ivy.array_decimal_values
-        if self.backend == "" or ivy.is_local():
+        if self.backend in ["", "none"] or ivy.is_local():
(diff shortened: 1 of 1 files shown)


jshepherd01:
#20884 provides a much better fix

jshepherd01:
Reopening temporarily so we can see the tests again

PR #20850, opened 2023-07-26
Complex activations

jshepherd01:
Adds functionality for complex numbers in activation functions, using a new `handle_complex_input` decorator. Currently only applied to `leaky_relu` as a proof of concept.


vedpatwardhan (commented):
Hey @jshepherd01, thanks for diving into this, I'll request @sherry30's review in this PR as we discussed on discord 😄

ivy/functional/ivy/activations.py
```diff
@@ -77,13 +79,35 @@ def gelu(
...
+
 @handle_exceptions
 @handle_nestable
 @handle_array_like_without_promotion
 @handle_out_argument
 @to_native_arrays_and_back
 @handle_array_function
+@handle_complex_input
 @handle_device_shifting
+@add_attributes(jax_like=_leaky_relu_jax_like)
```
sherry30, line 110: Do we really need this decorator? 🤔 

Cant we just do `leaky_relu.jax_like = _leaky_relu_jax_like` below this function.

ivy/functional/ivy/activations.py
```diff
@@ -77,13 +79,35 @@ def gelu(
...
+    )
+
+
 @handle_exceptions
 @handle_nestable
 @handle_array_like_without_promotion
 @handle_out_argument
 @to_native_arrays_and_back
 @handle_array_function
+@handle_complex_input
```
sherry30, line 108: Is there a reason why this decorator is above `handle_device_shifting`? 
If you put this decorator below it then you won't have to put `handle_device_shifting` decorator on `_leaky_relu_jax_like`

sherry30 (commented):
Confused about 2 things, pointed them out on specific lines.

Are the changes in the promote_dtype function also related to these changes?

ivy/functional/ivy/activations.py
jshepherd01: I tried that first, it took me a while to come up with the decorator as a workaround. `functools.wraps` copies attributes *up* the stack of decorators, but not down, so if you do the assignment after all of the decorators then only the top-level one is able to see it. The assignment has to run after the function definition but before the complex number handling decorator, hence the new decorator here

ivy/functional/ivy/activations.py
jshepherd01: Honestly, it was mostly because I couldn't exactly tell what `handle_device_shifting` does but it just felt more low-level than this one so I left it there. I was planning to ask someone where exactly in the stack this function is best placed, but I wanted to get it working well enough first

jshepherd01:
@sherry30 I've answered your questions on the particular lines above. About the promote_dtype function: yes, it was previously failing when one of the inputs was a complex scalar which was breaking a line in my code (the `1j * angle_inp` part) so I decided to write up a fix for that as well. I made sure I tested it as fully as I could while I was there

ivy/functional/ivy/activations.py
sherry30: I don't fully follow, I tried doing `leaky_relu.jax_like = _leaky_relu_jax_like` after the function implementation and removing the `add_attributes` decorator completely. It works fine.

Or perhaps you're saying there could be some other issues with this approach? 🤔

sherry30 (commented):
1 question.

Sounds good about the dtype part 👍

jshepherd01:
Now I look at it, the thing about `@add_attributes` is really odd. I'm sure that when I first tried it, just applying the attribute afterwards didn't work, but now that I test it again it seems to work fine. Really not sure what happened there.

ivy/functional/ivy/activations.py
```diff
@@ -103,6 +129,9 @@ def leaky_relu(
     out
         optional output array, for writing the result to. It must have a shape that the
         inputs broadcast to.
+    complex_mode
```
vedpatwardhan, line 132: Even though this argument is being handled completely by the wrapper, having this as an argument to the function itself makes things much more intuitive 😄

vedpatwardhan (changes requested):
Hey @jshepherd01 @mosesdaudu001, thanks for diving into this 💪
Is this issue specific to `leaky_relu`? Because if that's the case, having a function wrapper for a single function might be a bit overkill as the function wrappers are aimed towards more generally applicable issues. We could simply put the logic in the new function wrapper  into a private helper in `ivy/functional/ivy/activations.py` and use it in the backends. 
Happy to know what you think @jshepherd01 @mosesdaudu001 @sherry30 😄

sherry30:
Agreed with @vedpatwardhan,
I was under the assumption that this was for all activation functions. If that's the case then it makes sense to add this decorator.
Otherwise if its only for 1 function, then we can handle this in a private function

Ishticode:
@vedpatwardhan @sherry30 
This PR is meant as an initial setup. It will be required to handle complex types for most of the activation functions. That right @jshepherd01

jshepherd01:
That's right. Since every activation function (and probably most of the elementwise functions, since they're often used as activations) will need this functionality, I designed it to be nice and easy to add it to other functions. A decorator makes the most sense for that in my opinion.

ivy/functional/ivy/activations.py
jshepherd01: Do you mean I should add it to the parameters inside the `def leaky_relu( ... )` part? I was on the fence about that a little bit, but if its more intuitive with than without then it makes sense to add it. Thanks for the feedback :)

ivy_tests/test_ivy/test_frontends/test_jax/test_nn/test_non_linear_activations.py
```diff
@@ -133,7 +133,8 @@ def test_jax_silu(
 @handle_frontend_test(
     fn_tree="jax.nn.leaky_relu",
     dtype_and_x=helpers.dtype_and_values(
-        available_dtypes=helpers.get_dtypes("float_and_integer"),
+        available_dtypes=helpers.get_dtypes("float_and_integer")
```
sherry30, line 136: Can't you just do `helpers.get_dtypes("numeric")`?
that contains complex dtypes as well.

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
```diff
@@ -31,10 +31,12 @@ def test_relu(*, dtype_and_x, test_flags, backend_fw, fn_name, on_device):
 
 
 # leaky_relu
+@reproduce_failure("6.72.0", b"AXicY2QAAUYGBjiNzEYHAADUAAY=")
```
sherry30, line 34: debug code

sherry30 (changes requested):
pointed out 1 thing
The rest looks good to me

ivy/functional/ivy/activations.py
vedpatwardhan: Yep, I think that would be more intuitive as the user clearly knows that the functions accepts a `complex_mode`, thanks for making the changes @jshepherd01 😄

ivy_tests/test_ivy/test_frontends/test_jax/test_nn/test_non_linear_activations.py
jshepherd01: That's a good point, updating it and the other tests now. Thanks for pointing it out

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
```diff
@@ -34,7 +34,8 @@ def test_relu(*, dtype_and_x, test_flags, backend_fw, fn_name, on_device):
 @handle_test(
     fn_tree="functional.ivy.leaky_relu",
     dtype_and_x=helpers.dtype_and_values(
-        available_dtypes=helpers.get_dtypes("float", full=False, key="leaky_relu"),
+        available_dtypes=helpers.get_dtypes("float", full=False, key="leaky_relu")
```
sherry30, line 37: I think Its better to do `numeric` here and adding `bfloat16` as unsupported dtype to whichever function doesn't support it

sherry30 (commented):
pointed out 1 thing

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
jshepherd01: This should be fixed by the new commit. The problem I was having was actually the opposite: some functions had `"complex"` listed as an unsupported dtype, which meant the test handler was never generating complex numbers to test with unless `prune_function` was false (which then introduced `bfloat16`s).

I didn't realise the unsupported types of one backend could affect the tests of a different backend, so I didn't notice until I went through them all to check just now. Thanks for pointing me in the right direction :)

jshepherd01: To be clear, I switched it to `float_and_complex` rather than `numeric` since the original was `float` and it seems that a lot of the backends don't support integers anyway

Ishticode (approved):
lgtm.

Ishticode:
although I would add adding this new wrapper under [miscellaneous](https://unify.ai/docs/ivy/overview/deep_dive/function_wrapping.html#miscellaneous-wrappers) section or whatever you think appropriate along with any other relevant details you migh consider.
Although feel free to do in a separate PR so as to merge this and see any downsides.
@jshepherd01

PR #20932, opened 2023-07-28
Aml 20867 paddle frontend var

### LINKED ISSUE #20867: var
Adding var which is a Statistical Function to PaddlePaddle Frontend

Aml-Hassan-Abd-El-hamid:
close #20867


Aml-Hassan-Abd-El-hamid:
Hi @ZoeCD ,

Can you please review the PR? 
I'd appreciate your feedback on this, And, if anything needs changing, please let me know.

Thank you.

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_stat.py
```diff
@@ -131,3 +131,34 @@ def test_paddle_median(
...
+    *,
+    unbiased,
+    dtype_and_x,
+    keepdim,
+    fn_tree,
+    frontend,
+    backend_fw,
+    test_flags,
+):
+    input_dtype, x, axis = dtype_and_x
```
hirwa-nshuti, line 153: The issue is here when unpacking values this helper function returns tuple of 4 variables but you're unpacking only 3
```suggestion
    input_dtype, x, axis, _ = dtype_and_x
```

ivy/functional/frontends/paddle/tensor/stat.py
```diff
@@ -45,3 +45,13 @@ def median(x, axis=None, keepdim=False, name=None):
         else ivy.astype(x, ivy.float32)
     )
     return ivy.median(x, axis=axis, keepdims=keepdim)
+
+
+@with_supported_dtypes({"2.5.0 and below": ("float16", "float32", "float64")}, "paddle")
```
hirwa-nshuti, line 50: The version value should be `2.5.1`

hirwa-nshuti, line 50: The version value should be `2.5.1`

hirwa-nshuti (changes requested):
Hey @Aml-Hassan-Abd-El-hamid 

Sure no problem I took a look and pointed out the cause of failures hope yyou'll update your PR to get the tests passing 😊

Aml-Hassan-Abd-El-hamid:
Thank you very much for your help @hirwa-nshuti , I did the requested changes but I still have failing tests -they are now more than before- so is that ok or do I need to change more things?

ZoeCD:
> I tried my best but It kept failing in run_tests (1),run_tests (58) and run_tests (75)  so can you please take a look and tell me what changes I need to do? 

Hello! You don't need to worry about failing tests that are not related to your PR, so you can ignore run_tests (58) and run_tests (75). Regarding your failing tests, I get the following error for all backends:
````
RuntimeError: (NotFound) The kernel with key (CPU, Undefined(AnyLayout), float16) of kernel `mean` is not registered. Selected wrong DataType `float16`. Paddle support following DataTypes: float64, complex128, float32, complex64, bool.
[Hint: Expected kernel_iter == iter->second.end() && kernel_key.backend() == Backend::CPU != true, but received kernel_iter == iter->second.end() && kernel_key.backend() == Backend::CPU:1 == true:1.] (at /paddle/paddle/phi/core/kernel_factory.cc:199)
````

Honestly, I've never seen that before. @hirwa-nshuti do you know what might be happening?

**Comment by samthakur587** [comment]
hii @ZoeCD @hirwa-nshuti  i am sam . i am also working on var function to paddle.Tensor frontend method (#18995). and i am also getting this error. and i found that the error is coming because the backend of the var function is calling the _std function (mean ) and the mean doesn't sport the float16 dtype as it is only sport the float32 and float64 dtype.

and i changed the dtype after but still it can't work. i don't know how but when i try the following code it passes all the test.

```python 
#paddle.tensor.stat.py
@with_unsupported_dtypes({"2.5.1 and below": ("float16", "bfloat16")}, "paddle")
@to_ivy_arrays_and_back
def var(x, axis=None, unbiased=True, keepdim=False, *, out=None):
    return ivy.var(x, axis=axis, correction=int(unbiased), keepdims=keepdim, out=out)
```
```python
# var
@handle_frontend_test(
    fn_tree="paddle.var",
    dtype_and_x=_statistical_dtype_values(
        function="var",
        min_value=-1e04,
        max_value=1e04,
    ),
    keepdims=st.booleans(),
)
def test_paddle_var(
    *,
    dtype_and_x,
    keepdims,
    on_device,
    fn_tree,
    backend_fw,
    frontend,
    test_flags,
):
    input_dtype, x, axis, correction = dtype_and_x
        input_dtypes=input_dtype,
        frontend=frontend,
        test_flags=test_flags,
        fn_tree=fn_tree,
        on_device=on_device,
        backend_to_test=backend_fw,
        x=x[0],
        axis=axis,
        unbiased=bool(correction),
        keepdim=keepdims,
    )
```

this is the result 

image

samthakur587:
i think there must a problem with the paddle backend implementation of var function. because i worked on var method on jax frontend there also all the test is passing except the paddle backend. 

can you please check if the code that i provided is correct implementation then you can merge it so i can  call this var in my paddle.Tensor directly and it should work there also.

thank you

hirwa-nshuti (changes requested):
Basically the `paddle.var` function doesn't support `float16` even it was mentioned here in the docs https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/var_en.html#var

```python
import paddle
x = paddle.to_tensor([-1., -1.], dtype="float16")
print(paddle.var(x, axis=None, unbiased=True, keepdim=False, name=None))
```
will throw runtime error for unsupported dtype
Removing `float16` from supported dtypes will fix the issue.
Also resolve the merge conflicts

hirwa-nshuti (approved):
LGTM!

Thanks for contributing😊

PR #20960, opened 2023-07-28
Default Args PR

Infrared1029:
The strategy used for setting the arguments is the following:
1) Check the established frameworks we all know (torch, tf etc.) and see what kind of defaults are set, and consider those.
2) If a parameter has a specific known value do not set it to None, rather set it to its value, good examples of this are the cases where axis is defaulted to `None` even though it is `-1` e.g `softmax`, there are cases however where `None` would make sense for axis, the cases where `None` means the array will be flattened first and whatever operation will be performed.

There were a few issues that I've noticed while going through the functional API tho, there might be more issues as i was not looking with the intention of finding these obviously
1) Docstrings that miss some arguments (e.g linl_alg.py cross function)
2) Some arguments appear to be completely wrong naming wise and possibly implementation wise( losses.py cross_entropy has an epsilon argument that is supposed to be `label_smoothing` according to its docstrings, and im skeptical about its implementation)
3) Functions that im skeptical about its implementation, like `multinomial` in random.py, the implementation/usage of the numpy's multinomial is more aligned with what id expect from a multinomial function.


ivy-leaves:

Thank you :hugs:

Infrared1029:
> If you are working on an open task, please **edit** the PR description to link to the issue you've created.
> 
> For more information, please check [ToDo List Issues Guide](https://unify.ai/docs/ivy/overview/contributing/the_basics.html#todo-list-issues).
> 
> Thank you 🤗

shush

ivy/functional/backends/jax/activations.py
```diff
@@ -37,7 +37,7 @@ def sigmoid(x: JaxArray, /, *, out: Optional[JaxArray] = None) -> JaxArray:
 
 
 def softmax(
-    x: JaxArray, /, *, axis: Optional[int] = None, out: Optional[JaxArray] = None
+    x: JaxArray, /, *, axis: Optional[int] = -1, out: Optional[JaxArray] = None
```
vedpatwardhan, line 40: `Optional` should be used when the default value of an argument is `None`, we can remove it from here. Similar changes apply to all other functions

ivy/functional/backends/jax/activations.py
```diff
@@ -37,7 +37,7 @@ def sigmoid(x: JaxArray, /, *, out: Optional[JaxArray] = None) -> JaxArray:
...
 ) -> JaxArray:
     if axis is None:
```
vedpatwardhan, line 42: Also, this line becomes redundant in that case, assuming we aren't accepting `None` as the `axis` from the user. Similar changes apply to all other functions

ivy/functional/backends/jax/creation.py
```diff
@@ -319,9 +319,9 @@ def one_hot(
     depth: int,
     /,
     *,
-    on_value: Optional[Number] = None,
```
vedpatwardhan, line 322: Given that this is inspired from `tf.one_hot`, not sure if we should change the default value in this case, but given that they've mentioned in their docs that they defer to `1` and `0` when they're `None`, probably this change is more sensible, as `axis=-1` is more intuitive and is also something `jax.nn.one_hot` does

ivy/functional/backends/jax/layers.py
```diff
@@ -55,8 +55,8 @@ def _get_tranpose_padding(
 def conv1d(
     x: JaxArray,
     filters: JaxArray,
-    strides: Union[int, Tuple[int]],
-    padding: Union[str, int, Sequence[Tuple[int, int]]],
+    strides: Union[int, Tuple[int]] = 1,
```
vedpatwardhan, line 58: I think if we're changing the default value of strides for `conv1d` based on `torch.nn.functional.conv2d`, we should follow the same for the other convolution functions as well, as you've already done with the other backends of course!

vedpatwardhan (changes requested):
Hey @Infrared1029, thanks for looking into this, just requested a few changes/made a few suggestions. Let's see what @MahmoudAshraf97 thinks as well 😄

ivy/functional/backends/jax/activations.py
Infrared1029: Woops, makes sense!

ivy/functional/backends/jax/activations.py
Infrared1029: That is correct, on it.

ivy/functional/backends/jax/creation.py
Infrared1029: i think having explicit values makes more sense indeed even if the signature is sto-, i mean, taken, from tensorflow

ivy/functional/backends/jax/layers.py
Infrared1029: i thought the other convolution funcs were addressed in this pr did i miss any?

vedpatwardhan: Yep, just the `jax` backend ones, thanks!

Infrared1029:
This should be it i hope? once this is merged i can jump into the pos and keyword args task

vedpatwardhan (changes requested):
There seem to be quite a few test failures in the CI, could you please update your branch to master to confirm if the tests are failing due to the changed function signatures or an inherent problem in the tests was fixed? Thanks @Infrared1029 😄

Infrared1029:
Given that this PR now removes the `None` default value from args like `axis` im assuming some frontend code might start failing because they are probably passing `axis` directly without a check (because `None` was a valid option)

@vedpatwardhan

ivy/functional/backends/jax/creation.py
vedpatwardhan: sounds good, thanks!

vedpatwardhan (approved):
lgtm overall, regarding the possible frontend failures @AnnaTz could you please take a look if the changes in this PR lead to any failures in the frontends? Thanks @Infrared1029!

ivy/functional/backends/jax/activations.py
```diff
@@ -85,11 +77,7 @@ def softsign(x: JaxArray, /, *, out: Optional[JaxArray] = None) -> JaxArray:
     return jax.nn.soft_sign(x)
 
 
-def log_softmax(
-    x: JaxArray, /, *, axis: Optional[int] = None, out: Optional[JaxArray] = None
-):
-    if axis is None:
-        axis = -1
+def log_softmax(x: JaxArray, /, *, axis: int = -1, out: Optional[JaxArray] = None):
     return jax.nn.log_softmax(x, axis)
```
AnnaTz, line 83: The changes to the `axis` of `ivy.softmax` seem to be causing failures in `test_tensorflow_softmax`, `test_tensorflow_Softmax` and `test_tensorflow_sparse_categorical_crossentropy`.
The changes to the `axis` of `ivy.log_softmax` also seem to be causing failures in `test_tensorflow_LogSoftmax`.

ivy/functional/backends/jax/activations.py
```diff
@@ -45,36 +45,28 @@ def sigmoid(x: JaxArray, /, *, out: Optional[JaxArray] = None) -> JaxArray:
 
 
 def softmax(
-    x: JaxArray, /, *, axis: Optional[int] = None, out: Optional[JaxArray] = None
+    x: JaxArray, /, *, axis: int = -1, out: Optional[JaxArray] = None
 ) -> JaxArray:
-    if axis is None:
-        axis = -1
     return jax.nn.softmax(x, axis)
```
AnnaTz, line 50: But first I would suggest fixing the jax backend of `ivy.softmax`. I can see `test_softmax` is also failing there and this might be related to some frontend failures.

AnnaTz (changes requested):

ivy-seed:

PR #20973, opened 2023-07-28
Fix multiple issues with ivy.jac

somaia02:
Update ivy.jac to:
1. Convert x_in to ivy_array since the function passed to ivy.jac expects the input to be ivy arrays and not native array
2. Accept functions that return named tuples
3. use jacrev instead of jacfwd since the results are different in some cases
4. For paddle: 
   * Return ivy.array instead of a paddle.jacobian object
   * Return the jacobian with the right shape


ivy-leaves:

Thank you :hugs:

ivy/functional/backends/jax/gradients.py
```diff
@@ -145,10 +145,15 @@ def stop_gradient(
...
 def jac(func: Callable):
-    grad_fn = lambda x_in: ivy.to_native(func(x_in), nested=True)
+    grad_fn = lambda x_in: ivy.to_native(
+        func(ivy.to_ivy(x_in, nested=True)),
+        nested=True,
+        include_derived={tuple: True},
+    )
     callback_fn = lambda x_in: ivy.to_ivy(
-        jax.jacfwd(grad_fn)((ivy.to_native(x_in, nested=True))),
+        jax.jacrev(grad_fn)((ivy.to_native(x_in, nested=True))),
```
vedpatwardhan, line 154: I think there's some cases where `jacrev` is consistent with the other backends and others where `jacfwd` is more consistent. I had also created an [issue](https://github.com/google/jax/issues/14149) for this, but doesn't seem to have been responded to.

ivy/functional/backends/jax/gradients.py
```diff
@@ -145,10 +145,15 @@ def stop_gradient(
...
         nested=True,
+        include_derived={tuple: True},
```
vedpatwardhan, line 156: Good catch! `include_derived=True` would be even more generalized in that case, so that we cover other derived types like `ivy.Container`. Same applies to the other backends

vedpatwardhan (changes requested):
Hey @somaia02, just requested a couple of changes, I'll also request @Ishticode's review. Thanks for looking into this 😄

ivy/functional/backends/jax/gradients.py
somaia02: It's a problem with torch.jacfwd as well. I only tried with the `eigh` function though. It doesn't return nan but returns different values in each case.

somaia02:
Regarding the issue with computing higher-order gradients. I tried this code and it works well. Are there any other cases that fail?
```python
x = ivy.array(1.)
fn = lambda x: ivy.jac(ivy.sin)(x)
j = ivy.jac(fn)(x)
print(j) # ivy.array(-0.84147096)
```

ivy/functional/backends/jax/gradients.py
vedpatwardhan: oh okay got it! Thanks for clarifying, we can replace `jacrev` by `jacfwd` if it's more consistent for the cases we're testing 👍

vedpatwardhan:
> Regarding the issue with computing higher-order gradients. I tried this code and it works well. Are there any other cases that fail?
> 
> ```python
> x = ivy.array(1.)
> fn = lambda x: ivy.jac(ivy.sin)(x)
> j = ivy.jac(fn)(x)
> print(j) # ivy.array(-0.84147096)
> ```

I can't really think of a specific case, but I'll leave it with @Ishticode to test it out 😄

somaia02:
Hello @Ishticode Can you please review this? Thank you.

Ishticode (approved):
Hi @somaia02 
Thanks for the great work here.
the chaining above works with ivy in accordance with jax in following code
```python
import jax
import jax.numpy as jnp
x = jnp.array(1.)
fn = lambda x: jax.jacrev(jnp.sin)(x)
j = (jax.jacrev(fn))(x)
print(j)
``` 

and any other similar combinations.

Although I am not sure we are testing for higher orders unless i missed something. Do you want to create another PR to add the additional tests @somaia02 and we merge this (which is easier) or continue on.

somaia02:
Okay. I will merge this and make another PR with the additional tests. Thank you.

PR #20974, opened 2023-07-29
Update prune_nest_at_indices to make sure indices are in the right order

somaia02:
`prune_nest_at_indices` doesn't work properly when indices are not in the right order. Here is an example of a case that fails.
```python
xs = [1, 1]
ivy.prune_nest_at_indices(xs, [[0],[1]])
```
Error message: `list assignment index out of range`


somaia02:
Hello @sondertg . Can you please review this? Thank you.

sondertg (approved):
Hi @somaia02, thanks for your contribution. LGTM!

PR #21002, opened 2023-07-29
torch complex

### LINKED ISSUE #16974: complex

ken-kay:
Close #16974


hmahmood24 (changes requested):
Hey @ken-kay, thanks for the PR! However, in order to proceed forward, there a couple of points we need to work on.

1. First off, this PR is supposed to add just a frontend function and its test and so adding functions to any other APIs is not recommended/acceptable.
2. Even if you were to add a function to `ivy`, you would need to add it to all the backends and the data_classes submodule as well as its test in the same PR.

I would suggest you look at the implementation of `paddle.complex` in the `PaddlePaddle` frontend in `ivy` and try to implement your function along the same lines. Kindly ask for a re-review once you're done. Thanks! :)

ken-kay:
Hey @hmahmood24

It appears that this issue was closed because it has already been implemented but thank you for the advice.
I'll close this branch as the topic has already been closed.

PR #21024, opened 2023-07-30
 interpolate mindspore

### LINKED ISSUE #21048: interpolate
- [interpolate] #21024

sushmanthreddy:
close #21048


sushmanthreddy:
@jshepherd01 can u plss review the pull request?
and pls let me knew for any changes

ivy_tests/test_ivy/test_frontends/test_mindspore/test_ops/test_function/test_mindspore_nn_func.py
```diff
@@ -77,3 +77,51 @@
...
+#         max_value=5,
+#         min_dim_size=4,
+#         shape=(
+#             st.integers(min_value=2, max_value=10),
+#             4,
+#             st.integers(min_value=12, max_value=64),
+#             st.integers(min_value=12, max_value=64),
+#         ),
+#     ),
+#     size=st.lists(st.integers(min_value=1, max_value=10), min_size=3, max_size=3) | None,
```
jshepherd01, line 98: The size argument can also accept `int` and `tuple[int]` types, which should also be tested here. `scale_factor`, similarly, supports `float` and `tuple[float]`.

Also, how do you enforce that only one of them can be `None`?

jshepherd01 (changes requested):
Looks pretty good, and I know writing code you can't test is difficult at the best of times so you did well in that respect. Just a few points to pay attention to in terms of the docs (I'm going off the ones [here](https://mindspore.cn/docs/en/master/api_python/ops/mindspore.ops.interpolate.html) but I'm not an expert on mindspore so I might be wrong)

ivy/functional/frontends/mindspore/ops/function/nn_func.py
```diff
@@ -36,3 +36,11 @@ def selu(input_x):
 @to_ivy_arrays_and_back
 def softsign(x):
     return ivy.divide(x, ivy.add(1, ivy.abs(x)))
+
+@with_supported_dtypes(
+    {"2.0.0 and below": ("int8", "int16", "int32", "int64", "float16", "float32", "float64")},
+    "mindspore",
+)
+@to_ivy_arrays_and_back
+def interpolate(input, size=None, scale_factor=None, mode='nearest', align_corners=False):
```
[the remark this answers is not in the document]
sushmanthreddy: sorry it got out of sight , I will add the recomputer_scale_factor

ivy_tests/test_ivy/test_frontends/test_mindspore/test_ops/test_function/test_mindspore_nn_func.py
sushmanthreddy: It appears that both `size` and `scale_factor` are allowed to be None. The `| None` part in both lines makes the respective argument optional, meaning it can take either the specified values or `None`. There is no enforcement that only one of them can be `None`.

To clarify, here's what the code is doing for each argument:

1. `size=st.lists(st.integers(min_value=1, max_value=10), min_size=3, max_size=3) | None,`

   - The `size` argument is allowed to be a list of three integers with values between 1 and 10 (inclusive) and must have a minimum size of 3 and a maximum size of 3.
   - Alternatively, it can be set to `None`, making it an optional argument.

2. `scale_factor=st.lists(st.floats(min_value=0.1, max_value=2.0), min_size=3, max_size=3) | None,`

   - The `scale_factor` argument is allowed to be a list of three floats with values between 0.1 and 2.0 (inclusive) and must have a minimum size of 3 and a maximum size of 3.
   - Alternatively, it can be set to `None`, making it an optional argument.

There is no enforcement that only one of them can be `None`. Both `size` and `scale_factor` can be provided or set to `None` independently. They are independent optional arguments.

jshepherd01: Sorry for the misunderstanding. The restriction I was talking about is from the mindspore docs I linked above:
>  One and only one of size and scale_factor can be set to None.

ivy_tests/test_ivy/test_frontends/test_mindspore/test_ops/test_function/test_mindspore_nn_func.py
```diff
@@ -77,3 +77,51 @@
...
+#     scale_factor=st.lists(st.floats(min_value=0.1, max_value=2.0), min_size=3, max_size=3) | None,
+#     mode=st.sampled_from(['nearest', 'linear', 'bilinear', 'bicubic', 'trilinear']),
```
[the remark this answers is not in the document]
sushmanthreddy: sorry it got out of my look, I have just checked the docs and will update them.

sushmanthreddy:
@jshepherd01 
here is the summary of updates  from first commit and second commit .

The main difference between the two code snippets is how the `size` and `scale_factor` arguments are generated and handled during the test execution:

1. Commit  1 (Original code with Hypothesis):
   - The `size` and `scale_factor` arguments are generated using Hypothesis strategies directly in the `@handle_frontend_test` decorator.
   - The Hypothesis strategies define the possible values for `size` and `scale_factor`, including valid and invalid cases.
   - The test function receives `size` and `scale_factor` as separate parameters and uses them directly during the test execution.

2. commit  2 (Modified code with `@given` decorator and if-else statements):
   - The `size_or_scale_factor` argument is generated using a Hypothesis strategy with `st.one_of`.
   - The Hypothesis strategy defines four possibilities for `size_or_scale_factor`: an integer, a tuple of integers, a list of integers, or `None`.
   - The test function receives `size_or_scale_factor` as a single parameter and then extracts `size` and `scale_factor` based on the type of `size_or_scale_factor`.
   - An if-else statement is used to ensure that one and only one of `size` and `scale_factor` is `None`, as required by the MindSpore documentation.

In summary, the main difference lies in how the test data is generated and how the `size` and `scale_factor` arguments are handled in the test function. The modified code (Code Snippet 2) is more flexible and adheres to the documented behavior of the `interpolate` function by ensuring that only one of `size` and `scale_factor` is `None` in each test case.

ivy_tests/test_ivy/test_frontends/test_mindspore/test_ops/test_function/test_mindspore_nn_func.py
```diff
@@ -77,3 +79,67 @@
...
+#     align_corners,
+#     recompute_scale_factor,
+#     on_device,
+#     fn_tree,
+#     frontend,
+#     test_flags,
+# ):
+#     dtype, x = dtype_and_x
+
+#     # Extract size and scale_factor from size_or_scale_factor argument
```
jshepherd01, line 119: This can still result in both of them being `None` though, right? And it doesn't use the integer or float option.

If I were you, I'd generate `size = st.integers(...) | st.tuples(...) | st.lists(...)`, and `scale_factor` similarly, as well as a boolean `size_or_scale_factor`. You can then do something like:
```py
if not recompute_scale_factor:
    if size_or_scale_factor:
        size = None
    else:
        scale_factor = None
```

ivy_tests/test_ivy/test_frontends/test_mindspore/test_ops/test_function/test_mindspore_nn_func.py
```diff
@@ -77,3 +79,67 @@
 #         on_device=on_device,
 #         x=x[0],
 #     )
+
+# interpolate
+# @given(size_or_scale_factor=st.one_of(st.integers(min_value=1, max_value=10), st.tuples(st.integers(min_value=1, max_value=10)), st.lists(st.integers(min_value=1, max_value=10), min_size=3, max_size=3), st.none()))
```
jshepherd01, line 84: We should generally add these strategies inside `handle_frontend_test` rather than in a `given` decorator.

Also, if you're using this for the scale factor then it should be able to generate floats, rather than just integers (I think they should be separate though).

jshepherd01 (changes requested):
I think there's still some issues with the `size` and `scale_factor` thing

sushmanthreddy:
@jshepherd01 are these changes ,can u pls review it?

ivy_tests/test_ivy/test_frontends/test_mindspore/test_ops/test_function/test_mindspore_nn_func.py
```diff
@@ -1,5 +1,7 @@
 # global
 # from hypothesis import strategies as st
+# from hypothesis import given
```
jshepherd01, line 3: If you're no longer using this import you should remove it

jshepherd01 (commented):
Looks like there's still some small issues that need to be resolved before we can merge it

jshepherd01:
Throwing a `ValueError` would still cause the test to fail, even though its the fault of the test rather than the function itself. Far better if you just prevent hypothesis from generating both values as `None` in the first place, which you can do by following the code I suggested earlier.

Also, your removed import still leaves a newline in your diff, which you should probably remove to keep your PR nice and clean. You can see exactly which lines you've changed on the "Files changed" tab, and you should try to only have the changes in that be deliberate ones.

sushmanthreddy:
@jshepherd01 removed the raisevalu error and removed the empty line also..

jshepherd01:
That doesn't really solve the problem that both `size` and `scale_factor` can still both be None at the same time, and this is an invalid input to the function. You should modify the strategies in your `@handle_frontend_test` decorator such that the two of them will never both be None.

I don't see any reason why the solution I suggested before wouldn't work, but I assume you have some reason not to use it so I can suggest an alternative: generate a single property called `size_and_scale_factor` which is something along the lines of
```py
st.one_of(
    st.tuples(
        st.one_of(
            st.integers(min_value=1, max_value=10),
            st.tuples(st.integers(min_value=1, max_value=10)),
            st.lists(st.integers(min_value=1, max_value=10), min_size=3, max_size=3),
        ),
        st.just(None),
    ),
    st.tuples(
        st.just(None),
        st.one_of(
            st.floats(min_value=0.1, max_value=2.0),
            st.tuples(st.floats(min_value=0.1, max_value=2.0)),
            st.lists(st.floats(min_value=0.1, max_value=2.0), min_size=3, max_size=3),
        ),
    ),
    st.tuples(
        st.one_of(
            st.integers(min_value=1, max_value=10),
            st.tuples(st.integers(min_value=1, max_value=10)),
            st.lists(st.integers(min_value=1, max_value=10), min_size=3, max_size=3),
        ),
        st.one_of(
            st.floats(min_value=0.1, max_value=2.0),
            st.tuples(st.floats(min_value=0.1, max_value=2.0)),
            st.lists(st.floats(min_value=0.1, max_value=2.0), min_size=3, max_size=3),
        ),
    ),
)
```
so that it contains tuples of either `(size, None)`, `(None, scale_factor)`, or `(size, scale_factor)`, and then you can unpack them and use the values as needed. For better code clarity, you might want to extract the size and scale factor strategies (the `one_of`s) into their own functions using `st.composite`

sushmanthreddy:
> Throwing a `ValueError` would still cause the test to fail, even though its the fault of the test rather than the function itself. Far better if you just prevent hypothesis from generating both values as `None` in the first place, which you can do by following the code I suggested earlier.
> 
> Also, your removed import still leaves a newline in your diff, which you should probably remove to keep your PR nice and clean. You can see exactly which lines you've changed on the "Files changed" tab, and you should try to only have the changes in that be deliberate ones.

so what should I do to pass the testcases other than putting valueError. My code doesn't work only for case where size is None and scale_factor is None.Other than adding valueError,should I return the string error?

sushmanthreddy:
@jshepherd01 I understood the error and have created three functions outside the test functions, which work on inputs for the function and have three tests here `none,scale_factor` , `size,None` and `size,scale_factor` and these are feed to the 
tests function .

jshepherd01:
I don't think you can unpack strategies that return tuples in that way. I think you need to put `size_and_scale_factor = size_and_scale_factor_strategy(),` inside your `@handle_frontend_test(` decorator, and add the unpacking inside the function body (like `size, scale_factor = size_and_scale_factor`). Look at `dtype_and_x` for an example.

You also should really be naming private functions like that with underscores at the start, like
```py
def _size_strategy():

def _scale_factor_strategy():

def _size_and_scale_factor_strategy():
```
so that they're not confused with actual test functions.

sushmanthreddy:
@jshepherd01 made the changes...as mentioned..

ivy_tests/test_ivy/test_frontends/test_mindspore/test_ops/test_function/test_mindspore_nn_func.py
```diff
@@ -77,3 +77,91 @@
...
+#             "bicubic",
+#             "trilinear",
+#             "area",
+#             "nearest-exact",
+#         ]
+#     ),
+#     align_corners=st.booleans(),
+#     recompute_scale_factor=st.booleans(),
+#     size=size_and_scale_factor[0],
+#     scale_factor=size_and_scale_factor[1],
```
jshepherd01, line 135: `size_and_scale_factor` is currently undefined at this point in the code. You need to remove these two lines (134 and 135) and replace them with the code that's currently on line 151

ivy_tests/test_ivy/test_frontends/test_mindspore/test_ops/test_function/test_mindspore_nn_func.py
```diff
@@ -77,3 +77,51 @@
...
 #     )
+
+# interpolate
+# @handle_frontend_test(
+#     fn_tree="mindspore.ops.function.nn_func.interpolate",
+#     dtype_and_x=helpers.dtype_and_values(
+#         available_dtypes=helpers.get_dtypes("valid"),
+#         num_arrays=1,
+#         shared_dtype=True,
+#         min_value=2,
```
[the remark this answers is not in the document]
jshepherd01: I'd appreciate an answer to this question as well, if it's not too much trouble

jshepherd01 (changes requested):

sushmanthreddy:
i didn't got the question u have mentioned .....can u pls elaborate it more??

ivy_tests/test_ivy/test_frontends/test_mindspore/test_ops/test_function/test_mindspore_nn_func.py
```diff
@@ -77,3 +77,89 @@
...
+#     mode,
+#     align_corners,
+#     recompute_scale_factor,
+#     on_device,
+#     fn_tree,
+#     frontend,
+#     test_flags,
+# ):
+#     dtype, x = dtype_and_x
+#     size_and_scale_factor = _size_and_scale_factor_strategy()
```
jshepherd01, line 149: I don't think you need to call this function twice

jshepherd01:
> @jshepherd01 i didn't got the question u have mentioned .....can u pls elaborate it more??

Sorry, I didn't realise GitHub would display it like this. I was replying to the question I asked in the first review, about the reason behind those size and value restrictions

sushmanthreddy:
> > @jshepherd01 i didn't got the question u have mentioned .....can u pls elaborate it more??
> 
> Sorry, I didn't realise GitHub would display it like this. I was replying to the question I asked in the first review, about the reason behind those size and value restrictions

@jshepherd01 i have replied over there..can u pls give look

jshepherd01:
I don't see your answer anywhere, could you point me to it?

sushmanthreddy:
> I don't see your answer anywhere, could you point me to it?

https://github.com/unifyai/ivy/pull/21024#discussion_r1280392501

jshepherd01:
That's the question, could you please point me to the answer?

sushmanthreddy:
> That's the question, could you please point me to the answer?

@jshepherd01 what was question was really about? 
Can u pls rephrase it here? 
If question was about the commented code... Ivy doesn't support mindspore yet...further to not make any complications, we have commented out the whole code.. 
And other tasks in this issues was also got commented 
If it's about min_value = 2 I have took it random... 
If u have anything in mind for implementation.. We will try to implement it... For min_value = 2

And other code in mindspore  ivy/mindspore/there also some min_value are there... U can give look on them.. Pls let me knew, why that code has some other min_value = 2 pr other values... 

I just saw them and took same value... Here... In my code...

sushmanthreddy:
> That's the question, could you please point me to the answer?

@jshepherd01 what was question was really about? 
Can u pls rephrase it here? 
If question was about the commented code... Ivy doesn't support mindspore yet...further to not make any complications, we have commented out the whole code.. 
And other tasks in this issues was also got commented 
If it's about min_value = 2 I have took it random... 
If u have anything in mind for implementation.. We will try to implement it... For min_value = 2

And other code in mindspore  ivy/mindspore/there also some min_value are there... U can give look on them.. Pls let me knew, why that code has some other min_value = 2 pr other values... 

I just saw them and took same value... Here... In my code...

sushmanthreddy:
@jshepherd01 In the code provided, `min_value` and `max_value` are parameters used in Hypothesis strategies to define the range of values that can be generated when generating random test data.

Here's a breakdown of what `min_value` and `max_value` do in this context:

1. `min_value`: This parameter specifies the minimum value that can be generated for the test data. For example, if `min_value` is set to 1, then the generated values will not be lower than 1.

2. `max_value`: This parameter specifies the maximum value that can be generated for the test data. If `max_value` is set to 10, then the generated values will not exceed 10.

In the context of  code, these parameters are used to define the range of values that can be generated for integers and floats when generating test data for testing the `interpolate` function from the MindSpore library. For instance, in `_size_strategy()` and `_scale_factor_strategy()`, these parameters are used to define the range of integers and floats respectively.

These parameters help in creating a variety of test cases that cover different possible scenarios and edge cases, ensuring that the tested code is robust and handles different input ranges correctly.

sushmanthreddy:
@jshepherd01 any update?

sushmanthreddy:
@jshepherd01 sorry ,now i understood the question .I think u are correct ,we need to test them entire range of parameters . i have limited  them   range of values generated during testing can make the test suite run faster and consume fewer resources.i think u are. right regarding that , I should have not restricted the values...I have changed the code for it..

so , I have changed it and have mistakenly added the shape factor also which is not needed at all.
I have removed it ...after referring the mindspore documentation


anyways I think we need to remove the shape factor also

ivy_tests/test_ivy/test_frontends/test_mindspore/test_ops/test_function/test_mindspore_nn_func.py
```diff
@@ -79,6 +79,83 @@
...
+#         st.tuples(st.integers()),
+#         st.lists(st.integers(), min_size=3, max_size=3),
+#     )
+
+# def _scale_factor_strategy():
+#     return st.one_of(
+#         st.floats(),
+#         st.tuples(st.floats()),
+#         st.lists(st.floats(), min_size=3, max_size=3),
+#     )
```
jshepherd01, line 94: Removing the bounds from the `dtype_and_x` part is good but I think here they were quite useful. In `x`, below, bounding the values doesn't really change much, like the function can still run with unbounded values so it's probably best to leave them without bounds, but here I think you need them.

For example, if either of these come up with a negative value, I think the function can't handle that, and also if they come up with very large values the function can easily run out of memory (because it'll try to create a tensor with `3e+38` entries)

jshepherd01 (changes requested):

PR #21059, opened 2023-07-31
Hamming window

### LINKED ISSUE #21058: hamming_window

walexi:
Close #21058


JPGoodale:
Hi @walexi, a couple of things need to be added such as setting the `backend_to_test` argument to your test function as well as making sure that datatypes which are not available in every framework, such as `bfloat16` are not generated in the tests.

JPGoodale:
Hey @walexi, you should still include `float32` and `float64` as supported types, as when you run your current code it will give: ```E  hypothesis.errors.InvalidArgument: Cannot sample from a length-zero sequence.``` due to `dtype=helpers.get_dtypes("float"),` having nothing to select from. I went ahead and added those types back in to see if things pass and it looks like there are still some mismatched results, they are very close however, so this can most likely be fixed by setting higher `atol` and `rtol` values.

JPGoodale:
Hi @walexi, just switch the order of your args back to periodic coming before alpha and beta in your function and test and you should be good to go.

PR #20979, opened 2023-07-29
fix automatic sub_backend detection

Infrared1029:

PR #21084, opened 2023-07-31
Skip all folders with invalid naming in sub_back_handler backends dirs check

### LINKED ISSUE #21082: [Bug] Sub backend handler doesn't ignore hidden folder and treat them as backend folder
In MacOS it's common that a directory has a `.DS_Store` directory inside it.

The `sub_backend_handler` scans the `ivy/functional/backends` folder looking for all the frontends directories but if the `.DS_Store` is presents it treats it like the others and this cause issues.

For example, trying to execute a test could fail due to this issue:
```
/opt/miniconda/envs/multienv/bin/python3 /opt/.pycharm_helpers/pycharm/_jb_pytest_runner.py --target test_linalg.py::test_vector_norm -- -N 400 --jb-show-summary 
Testing started at 10:36 ...
/opt/.pycharm_helpers/pycharm/_jb_pytest_runner.py:8: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  from pkg_resources import iter_entry_points
Launching pytest with arguments -N 400 test_linalg.py::test_vector_norm in /opt/project/ivy_tests/test_ivy/test_functional/test_core

ImportError while loading conftest '/opt/project/ivy_tests/test_ivy/conftest.py'.
../../conftest.py:7: in <module>
    import ivy_tests.test_ivy.helpers.test_parameter_flags as pf
../../helpers/__init__.py:2: in <module>
    from . import hypothesis_helpers
../../helpers/hypothesis_helpers/__init__.py:1: in <module>
    from . import general_helpers
../../helpers/hypothesis_helpers/general_helpers.py:8: in <module>
    import ivy
../../../../ivy/__init__.py:13: in <module>
    import ivy.utils.backend.handler
../../../../ivy/utils/__init__.py:1: in <module>
    from . import backend
../../../../ivy/utils/backend/__init__.py:1: in <module>
    from . import handler
../../../../ivy/utils/backend/handler.py:14: in <module>
    from ivy.utils.backend.sub_backend_handler import _clear_current_sub_backends
../../../../ivy/utils/backend/sub_backend_handler.py:31: in <module>
    for sub_backend in os.listdir(sub_backends_dir):
E   NotADirectoryError: [Errno 20] Not a directory: '/opt/project/ivy/functional/backends/.DS_Store/sub_backends'

Process finished with exit code 4

Empty suite
```

fabriziocacicia:
Fix #21082

### DIFF
--- a/ivy/utils/backend/sub_backend_handler.py
+++ b/ivy/utils/backend/sub_backend_handler.py
@@ -19,7 +19,7 @@ for backend in os.listdir(
         _backends_subpackage_path.replace(".", os.path.sep),
     )
 ):
-    if backend.startswith("__") or backend.startswith("."):
+    if not backend[0].isidentifier():
         continue
 
     sub_backends_dir = os.path.join(
@@ -29,7 +29,7 @@ for backend in os.listdir(
         "sub_backends",
     )
     for sub_backend in os.listdir(sub_backends_dir):
-        if sub_backend.startswith("__"):
+        if not backend[0].isidentifier():
             continue
         _sub_backend_dict[sub_backend] = (
             f"{_backends_subpackage_path}.{backend}.sub_backends.{sub_backend}"
--- a/ivy/utils/backend/sub_backend_handler.py
+++ b/ivy/utils/backend/sub_backend_handler.py
@@ -19,7 +19,7 @@ for backend in os.listdir(
         _backends_subpackage_path.replace(".", os.path.sep),
     )
 ):
-    if backend.startswith("__"):
+    if backend.startswith("__") or backend.startswith("."):
         continue
 
     sub_backends_dir = os.path.join(


fabriziocacicia:
@vedpatwardhan it solves it and it seems a cleaner solution. But I think `isalpha` is the wrong check to use since there could be some framework that could have a '_' in its name which `isalpha` will skip (for example `best_ml_framework`).
So, I updated my solution using `isidentifier`. Or if you want, you can ask the author of the other PR to update his. (I don't mind, don't worry) 😄

vedpatwardhan:
> @vedpatwardhan it solves it and it seems a cleaner solution. But I think `isalpha` is the wrong check to use since there could be some framework that could have a '_' in its name which `isalpha` will skip (for example `best_ml_framework`). So, I updated my solution using `isidentifier`. Or if you want, you can ask the author of the other PR to update his. (I don't mind, don't worry) 😄

Awesome! But even in case of `best_ml_framework`, `backend[0].is_alpha()` would still be `True` right? ig we should go with the other PR, given that it it was created earlier, thanks for understanding 😅

fabriziocacicia:
> Awesome! But even in case of `best_ml_framework`, `backend[0].is_alpha()` would still be `True` right? ig we should go with the other PR, given that it it was created earlier, thanks for understanding 😅

@vedpatwardhan nope. `"best_ml_framework".isalpha()` returns `False` since '_' is not an alphanumeric character.
I don't mind to have this PR merged, don't worry about it. I understand. But I still think that `isidentifier` is the right method to use 😄

fabriziocacicia:
Closed since #20979 is solving the same issue

vedpatwardhan:
> > Awesome! But even in case of `best_ml_framework`, `backend[0].is_alpha()` would still be `True` right? ig we should go with the other PR, given that it it was created earlier, thanks for understanding 😅
> 
> @vedpatwardhan nope. `"best_ml_framework".isalpha()` returns `False` since '_' is not an alphanumeric character. I don't mind to have this PR merged, don't worry about it. I understand.
> 
> But I still think that `isidentifier` is the right method to use 😄

actually I was referring to `"best_ml_framework[0].isalpha()` which is `"b".isalpha()` instead, this helps avoid selecting private modules and stuff because they start with a `_`. Thanks for looking into this though @fabriziocacicia 😄

vedpatwardhan:
Depends on what `backend` [here](https://github.com/unifyai/ivy/pull/20979/files#diff-e307abfa858606b3d696c4bd920e37920c0d7ec93532241dd5ee896cfb01c9baR22) is actually, you could be right about that 👍

fabriziocacicia:
> actually I was referring to `best_ml_framework[0].isalpha()` 

You are right 😅 my bad. I was ignoring the `[0]` part 😆

PR #21091, opened 2023-07-31
refactor einsum_helper for tf backend and to generate output_eq correctly taking common dims between the operands

### LINKED ISSUE #21090: refactor `einsum_helper`
At present , the `einsum_helper`  fails for tensorflow backend with the error message:
```bash 
Expected dimension 1 at axis 0 of the input shaped [2] but got dimension 2 
```

### Changes:
- Decide which dimensions are common and update `eq_2` accordingly; previous this was done randomly.
- Made `dtype=float64` because numpy experience overflow error with `dtype=float32`while other frameworks works fine.
-  Shorten the value generation because numpy will throw `nan` while other framework will throw `inf`

akshatvishu:
Close #21090

### DIFF
--- a/ivy_tests/test_ivy/helpers/hypothesis_helpers/array_helpers.py
+++ b/ivy_tests/test_ivy/helpers/hypothesis_helpers/array_helpers.py
@@ -2042,6 +2042,7 @@ def get_second_solve_matrix(draw):
 
 @st.composite
 def einsum_helper(draw):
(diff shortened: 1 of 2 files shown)


Ishticode (approved):
This looks great. Ive ran tests locally. Seems to generate data right. made a minor commit update u can see. I will merge this now. Thanks @akshatvishu

PR #21110, opened 2023-07-31
multivariate_normal func

### LINKED ISSUE #20928: multivariate_normal

TemiOjo25:
close #20928


ivy-leaves:

Thank you :hugs:

TemiOjo25:
@vaithak can you pls review and let me know what it is I need to fix. Thank you.

TemiOjo25:
> Hi @TemiOjo25, the test is failing. Please try running the tests locally as the test is marked @pytest.xfail (refer this comment for the reason: [#19551 (comment)](https://github.com/unifyai/ivy/pull/19551#issuecomment-1641775783)).
> 
> The test seems to be failing because covariance matrix requires atleast 2 dimension in cholesky function. I think changing the parameters of the test should fix this.

@vaithak I was able to  change the parameters of the test, However I am getting a new error-> 'float' has not object attribute 'shape'. How do I tackle this because I have no idea  what could be causing this error?

vaithak:
> @vaithak I was able to change the parameters of the test, However I am getting a new error-> 'float' has not object attribute 'shape'. How do I tackle this because I have no idea what could be causing this error?

This is because you are sampling the mean as a float value, whereas it should be a vector of some shape (as it is multivariate distribution). Similarly, for the covariance matrix, instead of hardcoding it, we should sample it as a 2d matrix (an example of how it is done in a test for Cholesky method [here](https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_linalg.py#L430)).

One thing to make sure of is that shape of the mean vector should be compatible with that of the covariance matrix. So, it would be better to have a function `get_mean_cov_vector` in some helper file (like done in `array_helpers.py`).

TemiOjo25:
> > @vaithak I was able to change the parameters of the test, However I am getting a new error-> 'float' has not object attribute 'shape'. How do I tackle this because I have no idea what could be causing this error?
> 
> This is because you are sampling the mean as a float value, whereas it should be a vector of some shape (as it is multivariate distribution). Similarly, for the covariance matrix, instead of hardcoding it, we should sample it as a 2d matrix (an example of how it is done in a test for Cholesky method [here](https://github.com/unifyai/ivy/blob/master/ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_linalg.py#L430)).
> 
> One thing to make sure of is that shape of the mean vector should be compatible with that of the covariance matrix. So, it would be better to have a function `get_mean_cov_vector` in some helper file (like done in `array_helpers.py`).

@vaithak Thanks for your response. Can you check the modification I made to the cov and let me know if that is the right way of sampling it because I still feel like something is off. For now, I included the get_mean_cov_vector in the random_test file for convenience but is the function alright? The current error which is the local variable 'ret' referenced before the assignment is weird because I do not see ret being referenced before the assignment in the test function. I apologize for any inconvenience because it feels like I'm going in circles. Thank you.

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
```diff
@@ -1417,3 +1419,99 @@ def call():
...
+    dtype,
+    method,
+    frontend,
+    backend_fw,
+    test_flags,
+    fn_tree,
+):
+    input_dtype, key= dtype_key
+    x = cov
+    x = np.asarray(x[0], dtype=dtype[0])
```
vaithak, line 1486: `cov` contains both dtype and the matrix, that's why you are getting the error of cannot convert string to float.

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
```diff
@@ -1417,3 +1419,99 @@ def call():
...
+        mean_size = draw(st.integers(min_value=1, max_value=mean_dim_size))
+        mean_values = draw(st.lists(st.floats(min_value=-10, max_value=10), min_size=mean_size, max_size=mean_size))
+        return np.array(mean_values, dtype=dtype)
+
+    mean = mean_strategy()
+
+    return mean
+
+# Sample the mean
+mean = get_mean_cov_vector(shape=(4,), dtype=np.float64)
```
vaithak, line 1440: why is it required to be global?

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
```diff
@@ -1417,3 +1419,99 @@ def call():
...
+    shape=helpers.get_shape(
+        min_num_dims=1, max_num_dims=6, min_dim_size=1, max_dim_size=6
+    ),
+    dtype=helpers.get_dtypes("float", full=False),
+    mean=st.lists(st.floats(min_value=-10, max_value=10), min_size=1, max_size=10).map(np.array),
+    cov=helpers.dtype_and_values(
+        available_dtypes=("float32","float64"),
+        min_value=0,
+        max_value=10,
+        shape=helpers.ints(min_value=2, max_value=5).map(lambda x: tuple([x, x])),
```
vaithak, line 1462: You need to ensure that both mean and `cov` vectors have compatible shapes. For ex. in your example run, mean single valued vector, whereas covariance matrix is 2x2, which is wrong.

vaithak (changes requested):

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
```diff
@@ -1417,3 +1419,99 @@ def call():
...
+    ).filter(
+        lambda x: "float16" not in x[0]
+        and "bfloat16" not in x[0]
```
vaithak, line 1465: given that our available dtypes list don't contain these, I don't think these ``dtype filters are required.

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
TemiOjo25: That was a mistake, I intended to call it within the scope of the test. therefore:
` mean=get_mean_cov_vector(shape=(4,), dtype=np.float64),
    cov=helpers.dtype_and_values(
        available_dtypes=("float32","float64"),
        min_value=0,
        max_value=10,
        shape=helpers.ints(min_value=2, max_value=5).map(lambda x: tuple([x, x])),`

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
TemiOjo25: Does this mean that both the mean shape and cov shape needs to have the same structure but not necessarily the same values? In that case, should I sample the mean in the same manner I am sampling the cov?
```
@handle_frontend_test(
    fn_tree="jax.random.multivariate_normal",
    dtype_key=helpers.dtype_and_values(
        available_dtypes=["uint32"],
        min_value=0,
        max_value=2000,
        min_num_dims=1,
        max_num_dims=1,
        min_dim_size=2,
        max_dim_size=2,
    ),
    shape=helpers.get_shape(
        min_num_dims=1, max_num_dims=6, min_dim_size=1, max_dim_size=6
    ),
    dtype=helpers.get_dtypes("float", full=False),
    mean=helpers.dtype_and_values(
        available_dtypes=("float32","float64"),
        min_value=0,
        max_value=10,
        shape=helpers.ints(min_value=2, max_value=5).map(lambda x: tuple([x, x])),
    ),
    cov=helpers.dtype_and_values(
        available_dtypes=("float32","float64"),
        min_value=0,
        max_value=10,
        shape=helpers.ints(min_value=2, max_value=5).map(lambda x: tuple([x, x])),
    ),
    method=st.sampled_from(["cholesky", "eigh", "svd"]),
)
```

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
TemiOjo25:
```
    input_dtype, key= dtype_key
    print("pass the troubles")
    cov_dtype, cov_matrix = cov
    x = np.asarray(cov_matrix[0], dtype=cov_dtype[0])
    # positive definite
    x = np.matmul(x.T, x) + np.identity(x.shape[0]) * 1e-3
```
Now I am extracting the cov matrix and its corresponding dtypes, but I followed the same logic from the cholesky test function in making the matrix positive definite. However, I still get an error saying matrix is not positive definite and will the postive definite for each matrix differ depending on the method?

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
TemiOjo25:
```
cov=helpers.dtype_and_values(
        available_dtypes=("float32","float64"),
        min_value=0,
        max_value=10,
        shape=helpers.ints(min_value=2, max_value=5).map(lambda x: tuple([x, x])),
    ),
```

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
vaithak: Please read the documentation of the corresponding function to understand the relation between mean and covariance's shape: https://jax.readthedocs.io/en/latest/_autosummary/jax.random.multivariate_normal.html

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
vaithak: Maybe the non-dtype filters present in the cholesky method can help, did you try those out?

TemiOjo25:
@vaithak I was able to make some changes and I believe this is on the right track. However I am currently ecountering an issue with the dtype_key. When I use "float32", "float64" it gives me the current error:
```
FAILED ivy_tests/test_ivy/test_frontends/test_jax/test_random.py::test_jax_multivariate_normal[cpu-numpy-False-False] - TypeError: JAX encountered invalid PRNG key data: expected key_data.dtype = uint32; got dtype=float32
FAILED ivy_tests/test_ivy/test_frontends/test_jax/test_random.py::test_jax_multivariate_normal[cpu-jax-False-False] - TypeError: JAX encountered invalid PRNG key data: expected key_data.dtype = uint32; got dtype=float32
FAILED ivy_tests/test_ivy/test_frontends/test_jax/test_random.py::test_jax_multivariate_normal[cpu-tensorflow-False-False] - TypeError: JAX encountered invalid PRNG key data: expected key_data.dtype = uint32; got dtype=float32
FAILED ivy_tests/test_ivy/test_frontends/test_jax/test_random.py::test_jax_multivariate_normal[cpu-torch-False-False] - TypeError: JAX encountered invalid PRNG key data: expected key_data.dtype = uint32; got dtype=float32
FAILED ivy_tests/test_ivy/test_frontends/test_jax/test_random.py::test_jax_multivariate_normal[cpu-paddle-False-False] - TypeError: can only concatenate tuple (not "list") to tuple
```
, and to avoid it I use uint32. However, whenever I switch to uint32 it affects the positive definite of the cov matrix and outputs this error:
```
FAILED ivy_tests/test_ivy/test_frontends/test_jax/test_random.py::test_jax_multivariate_normal[cpu-numpy-False-False] - ivy.utils.exceptions.IvyValueError: numpy: cholesky: Matrix is not positive definite
FAILED ivy_tests/test_ivy/test_frontends/test_jax/test_random.py::test_jax_multivariate_normal[cpu-jax-False-False] - ValueError: normal parameter shapes must be broadcast-compatible with shape argument, and the result of broadcasting the shapes must equal the shape argument, but got result (2,) for shape argument (1,).
FAILED ivy_tests/test_ivy/test_frontends/test_jax/test_random.py::test_jax_multivariate_normal[cpu-tensorflow-False-False] - ivy.utils.exceptions.IvyBackendException: tensorflow: cholesky: Value for attr 'T' of uint32 is not in the list of allowed values: double, float, half, complex64, complex128
FAILED ivy_tests/test_ivy/test_frontends/test_jax/test_random.py::test_jax_multivariate_normal[cpu-torch-False-False] - ivy.utils.exceptions.IvyBackendException: torch: _handle_nestable_dtype_info_wrapper: torch: as_native_dtype: Cannot convert to PyTorch dtype. uint32 is not supported by PyTorch.
FAILED ivy_tests/test_ivy/test_frontends/test_jax/test_random.py::test_jax_multivariate_normal[cpu-paddle-False-False] - ivy.utils.exceptions.IvyBackendException: paddle: is_float_dtype: paddle: as_ivy_dtype: Cannot convert to ivy dtype. uint32 is not supported by Paddle backend.
```

TemiOjo25:
@vaithak I also tried changing the supported dtypes in the random file for the multivariate func but that gives me a keyerror: cpu.

TemiOjo25:
@vaithak I keep getting an invalid PRNG key data error.

vaithak:
> @vaithak I keep getting an invalid PRNG key data error.

You can use `uint32` for the key parameter, for `cov` and `mean` testing with `float` dtype makes more sense.

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
```diff
@@ -1417,3 +1419,104 @@ def call():
...
+    if mean_batch_shape != cov_matrix_shape[:-2]:
+        common_batch_shape = ivy.broadcast_shapes(mean_batch_shape, cov_matrix_shape[:-2])
+        cov_matrix_shape = common_batch_shape + cov_matrix_shape[-2:]
+
+    return mean, cov, cov_matrix_shape
+
+@handle_frontend_test(
+    fn_tree="jax.random.multivariate_normal",
+    dtype_key=helpers.dtype_and_values(
+        available_dtypes=["float32", "float64"],
```
vaithak, line 1461: for `key` parameter, you should use `uint32` dtype as done for testing other functions in this file.

TemiOjo25: @vaithak I changed it to uint32

TemiOjo25: @vaithak Now I'm getting an error about Cholesky not being positive definite. I believe if I can figure that out I should be able to complete this function. Do you have any idea on how I could tackle this? I also feel like due to fact that I changed the key type it affected the way the positive definite of the matrix is being calculated. Wouldn't that be an expected failure because the dtypes aren't compatible?

vaithak: For dtype fixing, you can look into how `input_dtypes` is passed in case of different dtype for key and other parameters. It is done in some other test cases in the same file, by concatenating the list of dtypes.

TemiOjo25:
@vaithak 1 of the test is passing but the remaining tests give a uint not supported error?

TemiOjo25:
@vaithak I keep getting a matrix is not positive definite for the numpy test but this error does not occur with the jax test as it passes. This is a bit confusing. Is there a specific way to make the matrix positive definite because all the resources I've checked do not really help. Thank you!

TemiOjo25 (commented):
fixed

TemiOjo25:
@vaithak I was able to fix issue with regards to positive definite calculation. I was able to pass two of the tests but for the rest it has to do with uint32 not being supported. However, for the tensorflow test, I do think it can be tackled but I do not know what I need to do add uint32 to the list of allowed types ->` ivy.utils.exceptions.IvyBackendException: tensorflow: cholesky: Value for attr 'T' of uint32 is not in the list of allowed values: double, float, half, complex64, complex128`

TemiOjo25:
@vaithak pls do any of these failures have to do with my code because it I had no issue yesterday. Can you pls help?

vaithak:
lgtm. Apologies for the late response, was unwell for a couple of days.
The failing test for uint32 key dtype is a known issue.
Thanks 👍🏼

TemiOjo25:
@vaithak I hope you're feeling better. Thank you for the response and I'm so happy and Thank you!!!

PR #21132, opened 2023-08-01
added Invalid backend exception class and decorator to relevant functions

mosesdaudu001:


ivy/func_wrapper.py
```diff
@@ -294,6 +295,43 @@ def _check_in_nested_sequence(sequence, value=None, _type=None):
...
 
 
+# Invalid Backend Handling #
+# ---------------#
+
+
+def handle_backend_invalid(fn: Callable) -> Callable:
+    @functools.wraps(fn)
+    def _handle_backend_invalid(*args, **kwargs):
+        if args:
```
vedpatwardhan, line 305: I think a relatively general approach that also captures `ivy.Array` and `ivy.NativeArray` instances in `ivy.Container` arguments would be something along the lines of,
```
array_indices = ivy.nested_argwhere([args, kwargs], ivy.is_array)
array_vals = ivy.multi_index_nest([args, kwargs], array_indices)

def func(x):
    target_backend = ivy.utils.backend.handler._determine_backend_from_args(a)
    return target_backend is None and ivy.current_backend() != target_backend

if ivy.nested_any([array_vals], func):
    raise ivy.utils.exceptions.InvalidBackendException(
        "Operation not allowed. Array was instantiated with backend"
        f" {target_backend}. But current backend is"
        f" {ivy.current_backend_str()}. Please set dynamic=True"
        " for the array if you want to convert it to the target"
        " backend"
    )

return fn(*args, **kwargs)
```
Probably you're doing this already, but we'd skip all functions which don't accept `ivy.Array` in their arguments, let me know if the example you shared fails after making these changes, thanks @mosesdaudu001, let's see what @CatB1t thinks 😄

vedpatwardhan: Just a correction here as there's a couple of minor issues with the block I shared above, the code-block should instead look something like,
```
        array_indices = ivy.nested_argwhere([args, kwargs], lambda x: isinstance(x, ivy.Array))
        array_vals = ivy.multi_index_nest([args, kwargs], array_indices)

        def func(x):
            target_backend = ivy.utils.backend.handler._determine_backend_from_args(x)
            if target_backend is not None and ivy.current_backend() != target_backend:
                raise ivy.utils.exceptions.InvalidBackendException(
                    f"Operation not allowed. Array was instantiated with backend"
                    f" {target_backend.backend}. But current backend is"
                    f" {ivy.current_backend_str()}. Please set dynamic=True"
                    " for the array if you want to convert it to the target"
                    " backend"
                )
            return x

        ivy.nested_map(array_vals, func, include_derived=True)
        return fn(*args, **kwargs)
```

ivy/func_wrapper.py
```diff
@@ -294,6 +295,38 @@ def _check_in_nested_sequence(sequence, value=None, _type=None):
             )
 
 
+# Invalid Backend Handling #
```
vedpatwardhan, line 298: Can we move this wrapper section after the currently existing sections?

ivy/func_wrapper.py
```diff
@@ -294,6 +295,38 @@ def _check_in_nested_sequence(sequence, value=None, _type=None):
...
+# ---------------#
```
vedpatwardhan, line 299: A minor formatting fix, the number of `-` should match the header above it

ivy/func_wrapper.py
```diff
@@ -294,6 +295,38 @@ def _check_in_nested_sequence(sequence, value=None, _type=None):
...
+
+
+def handle_backend_invalid(fn: Callable) -> Callable:
+    @functools.wraps(fn)
```
vedpatwardhan, line 303: Could you please add a docstring for this wrapper

vedpatwardhan (changes requested):
Hey @mosesdaudu001, just requested a few changes, thanks 😄

vedpatwardhan (changes requested):
Hey @mosesdaudu001, we wouldn't need to add this decorator to compositional functions actually, could you please remove the decorator from all compositional implementations (`cross_entropy` for e.g.)? Ofc, as a result for mixed functions you'd need to add this decorator to the `to_add` key in the `mixed_backend_wrappers` attribute. 
Rest looks great to merge, the wrapper is also quite generalized now, thanks @mosesdaudu001 😄

ivy/func_wrapper.py
```diff
@@ -1508,6 +1509,53 @@ def _handle_complex_input(
...
+            If not, it raises an InvalidBackendException
+        """
+        array_indices = ivy.nested_argwhere(
+            [args, kwargs], lambda x: isinstance(x, ivy.Array)
+        )
+        array_vals = ivy.multi_index_nest([args, kwargs], array_indices)
+
+        def func(x):
+            target_backend = ivy.utils.backend.handler._determine_backend_from_args(x)
+            if target_backend is not None and ivy.current_backend() != target_backend:
```
vedpatwardhan, line 1542: let's replace the condition here by `if target_backend is not None and ivy.current_backend_str() != target_backend.backend:` as it's failing all the tests in the CI given that the backend objects are different instances even if the backend is the same. Thanks @mosesdaudu001 😄

vedpatwardhan (changes requested):

vedpatwardhan:
ivy-gardener

vedpatwardhan (approved):
lgtm! There are a few failures in the [CI](https://github.com/unifyai/ivy/actions/runs/5797661602/job/15714221938?pr=21132#step:4:12), but none of them seem to be due to the changes made in the PR.
Feel free to merge, thanks @mosesdaudu001 😄

PR #21133, opened 2023-08-01
Update numpy version to 1.25.2 on all the backend related files

fabriziocacicia:
There is a new numpy minor release and it's causing all the numpy tests to fail.

I updated all the occurrences of the old `1.25.1` numpy version in the backend files with the new `1.25.2` version number


RickSanchezStoic:
@fabriziocacicia since the tests work for backends with these changes, could you do this for the frontends too? Thanks!

RickSanchezStoic (approved):

PR #21135, opened 2023-08-01
refactoring handle_device_shifting to support function without input …

abdulasiraj:
…array


ivy/func_wrapper.py
```diff
@@ -812,6 +812,14 @@ def _handle_device_shifting(*args, **kwargs):
         -------
             The return of the function.
         """
+        if "device" in kwargs and kwargs["device"] is not None:
+            out = fn(*args, **kwargs)
```
vedpatwardhan, line 816: let's rename this to `ret` to avoid confusion with the `out` argument
Secondly, I think this implementation would simply bypass the device shifting logic when we have a `device` argument in `kwargs`, our handling would be different based on whether soft device was enabled or not. 
I'll request a review from @sherry30, thanks @abdulasiraj 😄

sherry30:
@vedpatwardhan So the PR is pretty much done,
I've removed all instances of explicit device shifting in the backend implementations.
A few things:
1. I'm not sure if we can remove it [here](https://github.com/unifyai/ivy/blob/e75c53220c533ce233c88584fabb2f739a211ba3/ivy/functional/backends/tensorflow/gradients.py#L25), if this function is always called in a function which has `handle_device_shifting` decorator then it should be safe to remove. Let me know if that's the case.
2. In some cases you can't control on which device will the tensor be initialized on in torch backend. Take `torch.normal` for instance, it doesn't take `device` argument and its possible to not pass any tensors to it in the input. in this case it will always be initialized on the `cpu` and then we have to transfer it to the desired device with `.to(device)`. This is the case even if we use device context manager in torch `with torch.device("cuda:0")` or even if we set the default device `torch.set_default_device("cuda:0")` in either cases it still initializes on `cpu`. I was thinking of creating an issue on the torch repo since this is something that should be controllable. I think in the meantime in this case we can use `.to(device)` since there is no work around, the other solution I can think of is that we check in the output device in the `handle_soft_device_variable` function but that would be really ugly. Let me know if you can think of something else or if its ok to let the torch backend have `.to(device)` for now where we can't control on what device the output will be initialized on.
```
import torch
torch.set_default_device("cuda:0")
with torch.device("cuda:0"):
  x = torch.normal(5.0, 5.0, (1, 2))
print(x.device) # cpu
```

sherry30:
ivy-gardener

ivy/func_wrapper.py
```diff
@@ -812,15 +812,25 @@ def _handle_device_shifting(*args, **kwargs):
         -------
             The return of the function.
         """
+        dev = None
+        if "device" in kwargs and kwargs["device"] is not None:
+            dev = ivy.as_native_dev(kwargs["device"])
         if ivy.soft_device_mode:
-            return ivy.handle_soft_device_variable(*args, fn=fn, **kwargs)
+            return ivy.handle_soft_device_variable(*args, fn=fn, dst_dev=dev, **kwargs)
```
vedpatwardhan, line 819: let's rename `dst_dev` to `device_shifting_dev` for code clarity. Same applies to all other occurrences of `dst_dev`

ivy/functional/backends/torch/device.py
```diff
@@ -87,6 +87,14 @@ def tpu_is_available() -> bool:
     return False
 
 
+def handle_soft_device_variable(*args, fn, dst_dev=None, **kwargs):
+    args, kwargs, dst_dev = ivy.shift_native_arrays_on_def_dev(
```
vedpatwardhan, line 91: `def_dev` doesn't really convey what the function does, let's rename this to `shift_native_arrays_on_default_device`.

ivy/functional/ivy/device.py
```diff
@@ -163,6 +154,20 @@ def _get_nvml_gpu_handle(device: Union[ivy.Device, ivy.NativeDevice], /) -> int:
     return handle
 
 
+def shift_native_arrays_on_def_dev(*args, dst_dev=None, **kwargs):
```
vedpatwardhan, line 157: let's make this function private if this is a helper

ivy/functional/ivy/device.py
```diff
@@ -163,6 +154,20 @@ def _get_nvml_gpu_handle(device: Union[ivy.Device, ivy.NativeDevice], /) -> int:
...
+    with ivy.ArrayMode(False):
+        default_device = ivy.default_device(dst_dev, as_native=True)
+        args, kwargs = ivy.nested_map(
+            [args, kwargs],
+            lambda x: (
+                ivy.to_device(x, default_device)
+                if (ivy.is_native_array(x) and ivy.dev(x) != default_device)
```
vedpatwardhan, line 164: Also, not completely sure why we need to explicitly check for native arrays, are ivy arrays being handled in the decorator itself? If that's the case we could probably also check for native arrays in the decorator

vedpatwardhan (changes requested):
Hey @sherry30 @abdulasiraj,
A bit obvious but just to confirm, we've tested that the behaviour remains the same for all backends even after removing the manual device logic right? 
Also, quite a lot of intelligent tests are failing, could you please confirm if none of the failures are related to the changes? 
Thanks 😄

ivy/functional/ivy/device.py
sherry30: I'm not sure what you mean. we're checking for native arrays here because we're passing `args` and `kwargs` to `nested_map` so there can be other input arguments in there.

sherry30:
Yup, I've already checked that it behaves the same after the manual logic removed, I've also asked @abdulasiraj to double check it. I'll test everything once again before merging.
As far as I've gone through the failures, they're for different functions failing and unrelated to these changes

ivy/functional/ivy/device.py
vedpatwardhan: Oh ignore that, I probably misunderstood what this line was doing, given that it's directly used in a backend implementation it doesn't need to check for `ivy` arrays using `is_ivy_array` or anything like that. Thanks for clarifying!

vedpatwardhan (approved):
lgtm! I think the failures are related to some issues with `function_unsupported_devices_and_dtypes` while don't occur locally. For e.g. [`test_linear`](https://github.com/unifyai/ivy/actions/runs/5834554594/job/15826266165?pr=21135#step:3:53676).
I'd suggest pulling the latest code from `main` and seeing if it affects the number of failures in the CI before merging.
Thanks @sherry30 @abdulasiraj 😄

abdulasiraj:
> lgtm! I think the failures are related to some issues with `function_unsupported_devices_and_dtypes` while don't occur locally. For e.g. [`test_linear`](https://github.com/unifyai/ivy/actions/runs/5834554594/job/15826266165?pr=21135#step:3:53676). I'd suggest pulling the latest code from `main` and seeing if it affects the number of failures in the CI before merging. Thanks @sherry30 @abdulasiraj 😄

I've tested most of the functions which include `handle_device_function` as decorator on the latest branch. sometime tests are failed but due to following two reasons:
1 - Due to dtypes inference which Ved mentioned earlier.
2 - Due to paddle backend (only paddle test failed).

I believe `handle_device_shifting` decorator is working very fine. On my side, I think It's good to go but @sherry30 mentioned he'll check again before merging. 
Thanks!

PR #21157, opened 2023-08-01
Embedding and Identity layers

kgmann:
Addresses this https://trello.com/c/hgF0F9Wu.


ivy-leaves:

Thank you :hugs:

ivy/functional/ivy/experimental/layers.py
```diff
@@ -1235,23 +1235,12 @@ def embedding(
     ivy.utils.assertions.check_equal(
```
vedpatwardhan, line 1235: Given that this assertion is defined in the `ivy` implementation, this will only be used when no backend is set, we'd need to use an assertion like this in the backend implementations as well if we'd like to check for the same whena. backend has been set.

ivy/stateful/initializers.py
```diff
@@ -414,7 +415,7 @@ def create_variables(
             ivy.random_normal(
                 mean=self._mean,
                 std=self._stddev,
-                shape=self._shape,
+                shape=self._shape if not var_shape else var_shape,
```
vedpatwardhan, line 418: Just to be consistent with the other initializers, we could actually just use `var_shape` and not have `shape` in the constructor at all. Or otherwise we could also add `var_shape` to the constructors of all the other intiializers, happy to know what you think 😄

ivy/stateful/layers.py
```diff
@@ -2098,40 +2098,106 @@ def _forward(self, x):
...
+class Identity(Module):
+    def __init__(self, dtype=None):
+        """
+        Identity layer. The layer is argument insensitive and returns the input argument
+        as output when called.
+
+        It's typically used as a placeholder when no operation is to be
+        performed. It doesn't have any learnable parameter.
+        """
+        Module.__init__(self, device=None, v=None, dtype=dtype)
```
vedpatwardhan, line 2185: I think this is the same as not passing any arguments to `Module.__init__(self)` right? `None` is the default value for `device` and `v` anyway and `dtype` isn't used in `_forward`😅

vedpatwardhan (changes requested):
Hey @kgmann, just requested a few changes, thanks 😄

ivy/stateful/initializers.py
kgmann: I think the first option is simpler and is just fine. I did this for backward compatibility to not break any code in the `models` repo that might be using the initializer but I just checked now and there's none so it's fine.

vedpatwardhan: Yeah that's a good point!
@hello-fri-end given that you had added this [initializer](https://github.com/unifyai/ivy/commit/e688e9832ed1947f47b4a0000d6f4640b7a80b81), could you please confirm if there's some other part of this repository or other repository where it's being used? If not, we can change it of course 😄

hello-fri-end: Hey @kgmann @vedpatwardhan ! I had actually added this initializer for the YOLOv7 model but ended up using something else in its place. I agree that it makes more sense not to have the `shape` argument in the constructor. Feel free to make this change :+1:

ivy_tests/test_ivy/test_stateful/test_initializers.py
kgmann, line 1: Quick note on this @vedpatwardhan, the test for the `RandNormal` initializer is currently failing locally. I inspected the issue, and it seems to come from `helpers.floats` and `helpers.ints` and to confirm it I ran the test for the `Constant` initializer and it's failing too because of those functions. 
I have an idea of what's causing it (tweaked the functions and the test worked just fine). But I was wondering if I should inspect the root  cause as part of this PR or just check and fix it in a new PR considering that it's affecting other tests that are using the functions as well. If you're for the latter option feel free  to merge this if there's no more changes, and I'll inspect the issue in another PR.

vedpatwardhan: Hey @kgmann, thanks for clarifying! I think we should merge the PR if the test failures have nothing to do with the changes you've made (which seems to be the case to me as well). Ofc we should look into this in a different PR as well as you mentioned 😄

vedpatwardhan (approved):
lgtm! There seem to be a few failures in the [CI](https://github.com/unifyai/ivy/actions/runs/5800077890/job/15721887961?pr=21157#step:4:12), but those seem unrelated to the changes made. Feel free to merge, thanks @kgmann 😄

PR #21175, opened 2023-08-02
fmin added backend/numpy/experiment/experimentwise

### LINKED ISSUE #21063: fmin

kashi476074:
close #21063


Aarsh2001:
Hey @kashi476074, this function is already there in the functional API, so i don't think we require an implementation in the experimental API !

PR #21181, opened 2023-08-02
Update func_wrapper.py

abdulasiraj:
As discussed on discord

### DIFF
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -821,8 +821,7 @@ def handle_device_shifting(fn: Callable) -> Callable:
         unique_devices = set(devices)
         # check if arrays are on the same device
         if len(unique_devices) == 1:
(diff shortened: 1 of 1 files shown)


sherry30 (changes requested):
This is not what I meant,
we still have to use `ivy.handle_soft_device_variable` because this function basically ensures that the backend framework executes it on the `default_device`. We need this for tensorflow, and in the future we might also need it for some other framework which cannot simply be handled by making sure that the input arrays area all on the same desired device.
We can skip this for torch and jax because we can control the device of execution by making sure that all the input arrays are on the same desired device.

what we can do is modify the compositional `ivy.handle_soft_device_variable` function to skip the `nested_map` part if `ivy.soft_device` is disabled

Sorry for not making this clear earlier

Thanks

PR #21198, opened 2023-08-02
Added nanprod to ivy experimental API

### LINKED ISSUE #21196: nanprod

BebehCodes:
Close #21196

### DIFF
--- a/ivy/data_classes/array/experimental/statistical.py
+++ b/ivy/data_classes/array/experimental/statistical.py
@@ -180,6 +180,66 @@ class _ArrayWithStatisticalExperimental(abc.ABC):
             self._data, axis=axis, keepdims=keepdims, dtype=dtype, out=out
         )
 
+    def nanprod(
+        self: ivy.Array,
+        /,
+        *,
+        axis: Optional[Union[Tuple[int], int]] = None,
+        dtype: Optional[Union[ivy.Dtype, ivy.NativeDtype]] = None,
+        out: Optional[ivy.Array] = None,
+        keepdims: Optional[bool] = False,
+        initial: Optional[Union[int, float, complex]] = None,
+        where: Optional[ivy.Array] = None,
+    ) -> ivy.Array:
+        """
+        ivy.Array instance method variant of ivy.nanprod. This method simply wraps the
+        function, and so the docstring for ivy.prod also applies to this method with
+        minimal changes.
+
+        Parameters
+        ----------
+        self
+            Input array.
+        axis
+            Axis or axes along which the product is computed.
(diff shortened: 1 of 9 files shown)


BebehCodes:
Hey @nassimberrada , sorry for the delayed response, I didn't have access to my workspace. I believe I took care of all your comments. Have a look and tell me what you think. Thank you!

BebehCodes:
Hey @nassimberrada , this error seems to also occur (very rarely) with all other backends but I can't really reproduce it. Even when I re-run the failed tests with the test case decorator, the tests pass and I get the error from the hypothesis that it couldn't reproduce it. I'm not sure if there's some casting or rounding that happens internally that causes this. 

Regarding the `to_numpy` method, I made some changes because when returning bfloat16, I noticed that in some cases the function turned the paddle tensor into a uint16 array and I was getting failed tests due to the type difference.

BebehCodes:
Hey @nassimberrada , thanks for your feedback. I think paddle is fixed now.

Regarding `to_numpy`, in some cases I noticed that a paddle bfloat16 tensor was converted to a uint16 array and my test cases would fail due to different returned types.

Have a look and tell me what you think!

BebehCodes:
Hey @nassimberrada , I've tried handling it by either doing casting when this occurs while testing, etc, but the thing is that the wrong casting that happens from `to_numpy` also changes the final results due to the different types, so I can handle the different dtypes but I can't seem to be able to do something about the numerical results. I noticed that something similar happens to torch's `to_numpy` where they also handle the `bfloat16` type exclusively, similarly to how I did it for paddle. Is it possible that this could be the correct solution?
Thank you!

BebehCodes:
Hey @nassimberrada , sorry for the confusion, I reverted the changes of `to_numpy` and pushed them, so that's why you are seeing this error. This value difference seems to happen because some weird overflowing or something happens because the paddle tensor is cast to uint16 instead of keeping its bfloat16 type. With the changes in `to_numpy` I didn't come across something similar and seemed to fix this issue as well.

The changes I made in the function helpers were in order to handle the error since I reverted `to_numpy` back to its original implementation, but they didn't seem to work.

Yes you are right about `initial`, I will add the strategy!

Generally, I haven't figured out a way to bypass the wrong casting that happens by paddle's `to_numpy` other than the changes I made previously regarding the handling of bfloat16.

BebehCodes:
Hey @nassimberrada . Thanks a lot for your suggestion! I added the supported dtypes decorator and the tests are passing for all backends. Also, I added the strategy for the initial argument as well.
Have a look and tell me what you think! 
Thanks!

PR #21246, opened 2023-08-03
Stateful changes 2

RickSanchezStoic:
Added buffer variables support for ivy Module. Methods such as _set_buffers,_get_buffers,_register_buffers,_buffer_tracker, _create_buffers added. Also _call was modified to account for the 'buffers' passed during the call.


ivy/data_classes/container/base.py
```diff
@@ -2845,7 +2845,7 @@ def cont_overwrite_at_key_chains(
                 )
             else:
                 return_dict[k] = v
-        return ivy.Container(return_dict, **self._config)
+        return return_dict
```
vedpatwardhan, line 2848: let's use [this](https://github.com/unifyai/ivy/pull/18164#discussion_r1250803939) return statement instead

ivy/stateful/module.py
```diff
@@ -144,6 +144,9 @@ def __init__(
         self._target = None
         self._lazy_compiled = False
         self._dynamic_backend = dynamic_backend
+        if hasattr(self, "_create_buffers"):
```
vedpatwardhan, line 147: I'm not sure why we need to have a tracker for buffers, I think `module.create_buffer` could just add something to the `self._buffers` and then we can have a read-only property `buffers` to get the value. What I mean is we don't need to explicitly allow a overridable `_create_buffers` function analogous to `_create_variables`, as all custom usages should always flow through `self.create_buffer` which could be called anywhere in the module (in the `__init__`, `_build`, `__call__`, `_forward`, etc.) and whenever we come across a call like that, we add it to the `self._buffers` dict. 
I think this will simplify the approach quite a bit, happy to know what you think. Thanks @RickSanchezStoic 😄

vedpatwardhan (changes requested):

ivy/stateful/module.py
RickSanchezStoic: The `_create_buffers` is just analogous to `_create_variables` , where users can define variables specifically meant to be buffer variables, unlike registering it on spot like torch (support for which is also added) , thought would be a good addon

ivy/data_classes/container/base.py
RickSanchezStoic: But then, in the case of `inplace=True` it won't return `self` , it will return a copy. Better alternative: ` return return_dict if inplace else ivy.Container(return_dict, **self._config)`

RickSanchezStoic: the whole `**self._config` logic applies only to `inplace=False` I guess

ivy/stateful/module.py
vedpatwardhan: I think given that the other frameworks follow the approach of creating buffers on spot ([torch](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_buffer) and [haiku](https://dm-haiku.readthedocs.io/en/latest/api.html#get-state)), I think our approach should also mirror theirs. I'm not sure if we'd add a lot of value by providing an overridable method for users to explicitly register buffers, the on-spot way would provide quite a lot of flexibility.
Just as a side note, we'd of course need to also store the buffers created in child modules in the buffer_dict or state_dict of the module where those child modules are initialized, just like we've done with variables.
Happy to know what you think, thanks @RickSanchezStoic 😄

ivy/stateful/module.py
```diff
@@ -428,6 +447,76 @@ def found_dup_callback(x, kc):
...
+                    f"{buffer} hasn't been defined for the given Module structure"
+                )
+            else:
+                setattr(self, buffer, buffers[buffer])
+                if hasattr(self, "buffers"):
+                    self.buffers.update({buffer})
+                else:
+                    setattr(self, "buffers", {buffer})
+
+    def _register_buffers(self, var_name, value):
```
RickSanchezStoic, line 516: make this public, and remove variables.

ivy/stateful/module.py
```diff
@@ -144,6 +145,9 @@ def __init__(
         self._target = None
         self._lazy_compiled = False
         self._dynamic_backend = dynamic_backend
+        if hasattr(self, "_create_buffers"):
```
vedpatwardhan, line 148: Seems like we're still using `_create_buffers` and `_buffer_tracker`, could you please remove it as we'd only need to allow users to register buffers "on-spot"?

ivy/stateful/module.py
```diff
@@ -171,16 +180,19 @@ def __init__(
...
                 Module._init_var.pop()
 
             # do a final check if _init_var  becomes empty, then delete it all together
-            del Module._init_var
+            if not Module._init_var:
+                del Module._init_var
 
             return
-        self.build(*args, dynamic_backend=dynamic_backend, **kwargs)
+        self.build(*args, dynamic_backend=dynamic_backend, buffers=buffers**kwargs)
```
vedpatwardhan, line 195: syntax error, `,` to be added

ivy/stateful/module.py
```diff
@@ -428,6 +447,76 @@ def found_dup_callback(x, kc):
...
+            getattr(self, "buffers", {}).update(
+                set(final_object_snapshot.keys()).difference(
+                    set(initial_object_snapshot.keys())
+                )
+            )
+
+        wrapper.buffers_tracked = True
+        return wrapper
+
+    def _get_buffers(self):
```
vedpatwardhan, line 467: I'm not sure why we need this method actually, given that we are already using `_find_buffers` while building, won't `self.buffers` be a container or arrays anyway?

ivy/stateful/module.py
vedpatwardhan: could you please rename `register_buffers` to `register_buffer`, `set_buffers` to `set_buffer` and others and make the necessary ones public?

vedpatwardhan (changes requested):
Hey @RickSanchezStoic, just made a few suggestions, seems like we need to clean it up a bit. Thanks 😄

ivy/stateful/module.py
```diff
@@ -448,6 +491,17 @@ def _create_variables(self, *, device=None, dtype=None):
         """
         return {}
 
+    def _create_buffers(self):
```
vedpatwardhan, line 494: Probably this is just left over from the previous changes, can we remove this method too? Users would be able to create buffers at any point by just calling `self.register_buffer`.
Also, could you please message on the PR once you think it's ready for review? Just to avoid me reviewing the PR if it isn't supposed to be reviewed. Thanks for digging into this @RickSanchezStoic 😄💪

vedpatwardhan (changes requested):

ivy/stateful/module.py
RickSanchezStoic: yep

ivy/data_classes/container/base.py
vedpatwardhan: @saeedashrraf what do you think?

ivy/stateful/module.py
vedpatwardhan: .

ivy/stateful/module.py
```diff
@@ -646,8 +699,9 @@ def build(
             True for successfully built a module.
         """
         self._dev = ivy.default(device, self._dev)
+        # build buffers if any
+        self._create_buffers()
```
vedpatwardhan, line 703: Given that `ivy.Module._create_buffers` has been removed, we should also remove this

ivy/stateful/module.py
```diff
@@ -798,9 +862,27 @@ def __getattribute__(self, name):
                 self._build_and_return_v(
                     self._args, dynamic_backend=self._dynamic_backend, **self._kwargs
                 )
-
+        if name == "buffers":
```
vedpatwardhan, line 865: This `if` block is redundant right?

ivy/stateful/module.py
```diff
@@ -798,9 +862,27 @@ def __getattribute__(self, name):
...
+            if name in self.buffers:
+                return self.buffers[name]
         return super().__getattribute__(name)
 
+    def __setattr__(self, name, value):
+        if hasattr(self, "buffers"):
+            if name in self.buffers:
+                self.buffers[name] = value
+                return
+
```
vedpatwardhan, line 877: extra blank line

vedpatwardhan (changes requested):
Hey @RickSanchezStoic, just requested a final few changes, thanks 😄

ivy/data_classes/container/base.py
vedpatwardhan: I don't think the config logic would apply to `inplace=False` either, given that `self.cont_copy` also creates a container with the same `config`. Probably we can use `return return_dict` as you mentioned

ivy/stateful/module.py
vedpatwardhan: this throws an error right?

vedpatwardhan (changes requested):
Final couple of changes left from the previous review, thanks @RickSanchezStoic 😄

vedpatwardhan (approved):

PR #21247, opened 2023-08-03
Torch/Paddle adaptive_max_pool2d

### LINKED ISSUE #21224: torch.nn.functional.adaptive_max_pool2d

jshepherd01:
Closes #21224

Adds `adaptive_max_pool2d` to the torch frontend, and also to Ivy's functional API (under the experimental folder). The only other frontend (as far as I could tell from a quick Google) to have an implementation of `adaptive_max_pool2d` is paddle, so I added it to that frontend as well, along with those two backends.

### DIFF
--- a/ivy/data_classes/array/experimental/layers.py
+++ b/ivy/data_classes/array/experimental/layers.py
@@ -832,6 +832,33 @@ class _ArrayWithLayersExperimental(abc.ABC):
             output_size,
         )
 
+    def adaptive_max_pool2d(
+        self: ivy.Array,
+        output_size: Union[Sequence[int], int],
+    ) -> ivy.Array:
+        """
+        Apply a 2D adaptive maximum pooling over an input signal composed of several
+        input planes.
+
+        Parameters
+        ----------
+        self
+            Input array. Must have shape (N, C, H_in, W_in) or (C, H_in, W_in) where N
+            is the batch dimension, C is the feature dimension, and H_in and W_in are
+            the 2 spatial dimensions.
+        output_size
+            Spatial output size.
+
+        Returns
+        -------
+            The result of the pooling operation. Will have shape (N, C, S_0, S_1) or
+            (C, S_0, S_1), where S = `output_size`
+        """
+        return ivy.adaptive_max_pool2d(
+            self._data,
+            output_size,
+        )
+
     def reduce_window(
         self: ivy.Array,
         init_value: Union[int, float],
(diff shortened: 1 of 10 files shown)


Daniel4078:
Well, you should implement the backend of this function for all the other frameworks in the experimental module as now all the relevant tests are failing left and right. In addition, I suppose adaptive_max_pool2d should not be placed in the experimental module in the first place and you should either implement it compositionally using existing ivy functions or ask ivy members to create the missing ivy.adaptive_max_pool2d function with complete backend support for you to use.

jshepherd01:
@Daniel4078 What happened to prompt this comment? This PR is over a month old.

To address your concerns:

> Well, you should implement the backend of this function for all the other frameworks in the experimental module as now all the relevant tests are failing left and right.

I implemented it as a mixed function so that it works for all backends. At the time that I merged this, the tests for it were passing both locally and in the CI.

> In addition, I suppose adaptive_max_pool2d should not be placed in the experimental module in the first place

On what basis? My understanding was that new functions always go into `experimental` first and then get moved into core later on, once they've been there for a while and are known to be stable.

> and you should either implement it compositionally using existing ivy functions or ask ivy members to create the missing ivy.adaptive_max_pool2d function with complete backend support for you to use.

I *am* an ivy member, so I created the missing function myself

Daniel4078:
> @Daniel4078 What happened to prompt this comment? This PR is over a month old.
> 
> To address your concerns:
> 
> > Well, you should implement the backend of this function for all the other frameworks in the experimental module as now all the relevant tests are failing left and right.
> 
> I implemented it as a mixed function so that it works for all backends. At the time that I merged this, the tests for it were passing both locally and in the CI.
> 
> > In addition, I suppose adaptive_max_pool2d should not be placed in the experimental module in the first place
> 
> On what basis? My understanding was that new functions always go into `experimental` first and then get moved into core later on, once they've been there for a while and are known to be stable.
> 
> > and you should either implement it compositionally using existing ivy functions or ask ivy members to create the missing ivy.adaptive_max_pool2d function with complete backend support for you to use.
> 
> I _am_ an ivy member, so I created the missing function myself

Oh, thank you for your notice, then can you please look that the failing tests of your function as you probably know what went wrong better than me. like adaptive_max_pool2d was failuring for multiple errors like AssertionError:  the results from backend jax and ground truth framework torch do not match: [[[0.]]]!=[[[1.]]] for the one in experimental while there is another test with similar name torch_adaptive_max_pool2d suffers from AssertionError: returned dtype = float32, ground-truth returned dtype = bfloat16 in jax backend and RuntimeError: "adaptive_max_pool2d" not implemented for 'Half' for jax backend at the same time

PR #21264, opened 2023-08-03
For double sided maxwell

### LINKED ISSUE #19483: double_sided_maxwell

stalemate1:
close #19483

### DIFF
--- a/ivy/functional/frontends/jax/random.py
+++ b/ivy/functional/frontends/jax/random.py
@@ -399,7 +399,8 @@ def poisson(key, lam, shape=None, dtype=None):
 )
 def rademacher(key, shape, dtype="int64"):
     seed = _get_seed(key)
-    a = ivy.random_uniform(seed=seed, shape=shape, dtype=dtype)
+    a_1 = ivy.ones(shape=shape, dtype=dtype)
+    a = ivy.multiply(a_1, 0.5)
     b = ivy.bernoulli(a, shape=shape, dtype="float32", seed=seed)
     b = ivy.astype(b, dtype)
     return 2 * b - 1
--- a/ivy/functional/frontends/jax/random.py
+++ b/ivy/functional/frontends/jax/random.py
@@ -399,12 +399,10 @@ def poisson(key, lam, shape=None, dtype=None):
 )
 def rademacher(key, shape, dtype="int64"):
     seed = _get_seed(key)
-    a_1 = ivy.ones(shape=shape, dtype=dtype)
-    a = ivy.multiply(a_1, 0.5)
-    b = ivy.bernoulli(a, shape=shape, dtype="float32", seed=seed)
+    prob = ivy.full(shape, 0.5, dtype="float32")
+    b = ivy.bernoulli(prob, shape=shape, dtype="float32", seed=seed)
     b = ivy.astype(b, dtype)
     return 2 * b - 1
-
 
 @handle_jax_dtype
 @to_ivy_arrays_and_back
--- a/ivy/functional/frontends/jax/random.py
+++ b/ivy/functional/frontends/jax/random.py
@@ -272,6 +272,28 @@ def maxwell(key, shape=None, dtype="float64"):
 @with_unsupported_dtypes(
     {
         "0.4.14 and below": (
+            "uint32"
+        )
+    },
+    "jax",
+)
+def double_sided_maxwell(key, loc, scale, shape=(), dtype="float64"):
+    params_shapes = ivy.broadcast_shapes(ivy.shape(loc), ivy.shape(scale))
+    if not shape:
+        shape = params_shapes
+
+    shape = shape + params_shapes
+    maxwell_rvs = maxwell(key, shape=shape, dtype=dtype)
+    random_sign = rademacher(key, shape=shape, dtype=dtype)
+
+    return random_sign * maxwell_rvs * scale+loc
+
+
+@handle_jax_dtype
+@to_ivy_arrays_and_back
+@with_unsupported_dtypes(
+    {
+        "0.4.14 and below": (
             "float16",
             "bfloat16",
         )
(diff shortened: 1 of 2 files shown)


zaeemansari70 (changes requested):
Hi @stalemate1 

`uint32` is not supported by `torch` and `paddle`, kindly add those in the `unsupported_dtypes` decorator.
Also, do make sure that the tests are being passing when you request a review.

Feel free to comment on the PR if you need to ask any question.

Thanks!

stalemate1:
Hello @zaeemansari70!
I had push this function last week after making sure it runs for all the frameworks except for torch and paddle. When I had committed this PR there were only 5 failures in the CI tests. When I went through the `display_test_results` there were new failures introduced but non of them were related to the function I was working on. 

Today, it says more than 90 files are failing for the same function. Could you please help me understand where I might be going wrong.

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
```diff
@@ -1355,6 +1355,68 @@ def call():
         assert u.shape == v.shape
 
 
+@pytest.mark.xfail
+@handle_frontend_test(
+    fn_tree="jax.random.double_sided_maxwell",
+    dtype_key=helpers.dtype_and_values(
+        available_dtypes=["uint32"],
```
zaeemansari70, line 1362: You are specifying the dtype here, this should be `valid` dtype and the dtypes that should be skipped should be added in the decorator above.

zaeemansari70 (changes requested):
Hi @stalemate1 
You are passing in a `uint32` dtype which is neither supported by `torch` nor `paddle`.
Kindly fix this :slightly_smiling_face: 
Also do not worry about the CI tests, I'll test your PR locally before merging :+1:

stalemate1:
hi @zaeemansari70 I did try different things including using `helper.get_dtypes("valid")` and hard coding`    if backend_fw == torch:
        key.dtype = torch.int64`. But nothing is working. Could you please help me understand how I can fix this.

ivy/functional/frontends/jax/random.py
```diff
@@ -328,6 +328,31 @@ def maxwell(key, shape=None, dtype="float64"):
     return x
 
 
+@handle_jax_dtype
+@to_ivy_arrays_and_back
+@with_unsupported_dtypes(
```
zaeemansari70, line 333: You have to specify the unsupported dtypes here

ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
zaeemansari70: and pass in `valid` here, so that the decorator skips the dtypes which are not to be tested on. Rather than us trying to restrict those in the tests.

zaeemansari70 (changes requested):
Addresses your comments :slightly_smiling_face: 

Let me know if you have some questions, thanks! :slightly_smiling_face:

Ookamice:
@stalemate1 
Hiya, just a quick word of advice but you should always review the additions and subtractions your commits make in the "files changed" tab. Right now, it seems like there is a lot of weird changes that are not relevant with the core issue of maxwell and you should try to resolve them before pursuing further. (This can either be done manually or alternatively you can copy your work so far into a temporary text file, then force fetch from upstream and then paste back your work). This will save a lot of time later on when you need to fix all the conflicts between your PR and the main stream.

stalemate1:
sorry @Ookamice I did not notice that. I have corrected those files now.

Ookamice:
@stalemate1
Hiya, thanks for making the changes. I don't think the changes you've made to rademacher is valid and should be rolled back to what it was before. (0.5 is chosen means that benoulli has an equal probability of being 0 or 1 and after scaling with 2*b-1, the range becomes -1 to 1)

PS: weirdly rademacher doesn't pass the tests on my end for some reason so I'm not sure why it is merged. Perhaps this may be the cause as to why pytorch test for double_sided_maxwell not working?

stalemate1:
@Ookamice but when I change it back to 0.5 rademacher is producing only one element, thus it is not working with double_sided_maxwell. Hence, this time I created an array of 0.5's so that it is compatible with any shape. I hope this is okay.

Ookamice:
@stalemate1 
I think using ivy.array([0.5]) is an oversight by the original creator of the function. I think it would be simpler to just use a scaler `0.5` although I see the motivation behind your workaround. Try using a scaler and if that doesn't work then I think your workaround is the best bet! 😸

stalemate1:
@Ookamice Could this be merged now?

Ookamice:
@stalemate1
Hiya, just doing some rudimentary tests on your commit Firstly, I had a look at rademacher which everything worked fine (minor refactor made).

Secondly, I had a look at maxwell and plotted your output distribution compared to jax's native implementation. I plotted the output of both and it seems like there's a descrepency that should be fixed:
image

PS: The code used to test this is:
```python
import ivy
import jax

from ivy.functional.frontends.jax.random import _get_seed

def maxwell_current(key, shape=None, dtype="float64"):
    seed = _get_seed(key)
    # generate uniform random numbers between 0 and 1
    z = ivy.random_uniform(seed=seed, shape=shape, dtype=dtype)
    # applying inverse transform sampling
    x = (z**2) * ivy.exp(-(z**2) / 2)
    return x

n = 1000000
import matplotlib.pyplot as plt
x = maxwell_current((123,123), (n,))
z = jax.random.maxwell(jax.numpy.array([123,123], dtype='uint32'), shape=(n,))


num_bins = int(ivy.sqrt(n))
n0, bins0, patches0 = plt.hist(x, bins=num_bins, histtype='step', color='blue')
n, bins, patches = plt.hist(z, bins=num_bins, histtype='step', color='red')
plt.show()
```

Ookamice:
@stalemate1 
Hiya, the tests aren't that accurate for random functions due to its random nature which is why it usually just involves a sanity check by the reviewer and then a shape/dtype check. This is why it is important to plot the distribution to check if the shape matches for large samples of N. I'll have a look at if the code you've posted will provide the correct distribution on Monday :)

Ookamice:
@dash96 
with the maxwell bug fixed, this is also ready now 🚀 🚀 🚀 
Keep up the good work 🥳

PR #21269, opened 2023-08-04
fix(framework): refactoring as_ivy_dev and as_native_dev in all backends

abdulasiraj:
will also close, #18247

### DIFF
--- a/ivy/functional/frontends/paddle/tensor/tensor.py
+++ b/ivy/functional/frontends/paddle/tensor/tensor.py
@@ -874,3 +874,14 @@ class Tensor:
     )
     def frac(self, name=None):
         return paddle_frontend.frac(self._ivy_array)
+
+    @with_unsupported_dtypes({"2.5.1 and below": ("float16", "bfloat16")}, "paddle")
+    def gather(self, y, name=None):
+        return paddle_frontend.gather(self, y)
+
+    @with_unsupported_dtypes(
+        {"2.5.1 and below": ("float16", "uint8", "int8", "bool")}, "paddle"
+    )
+    def gather_(self, y, name=None):
+        res = self.gather(self, y)
+        return ivy.inplace_update(self, res)
--- a/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_tensor.py
+++ b/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_tensor.py
@@ -2377,6 +2377,39 @@ def test_paddle_tensor_frac(
     )
 
 
+# gather
+@handle_frontend_method(
+    class_tree=CLASS_TREE,
+    init_tree="paddle.to_tensor",
+    method_name="gather",
+    dtypes_and_x=helpers.dtype_and_values(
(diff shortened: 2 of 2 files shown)


ivy-leaves:

Thank you :hugs:

vedpatwardhan (changes requested):
Hey @abdulasiraj, thanks for looking into this task, I think we could just use `as_native_dev` inside `as_ivy_dev` to avoid duplicating the fallback logic. 
Happy to know your thoughts 😄

abdulasiraj:
> Hey @abdulasiraj, thanks for looking into this task, I think we could just use `as_native_dev` inside `as_ivy_dev` to avoid duplicating the fallback logic.
> Happy to know your thoughts 😄

Hey @vedpatwardhan,
Thanks for your comment. Actually, I tried this but `as_ivy_dev` and `as_native_dev` has a little bit different logic. Also, both return different devices. so, `as_native_dev` can't be used inside the `as_ivy_dev`.

vedpatwardhan (commented):
Thanks for clarifying @abdulasiraj 
I think what we should do in that case is get the common logic between these 2 functions (`as_ivy_dev` and `as_native_dev`) as a private helper which both of these functions would make use of, happy to know what you think 😄

abdulasiraj:
> Thanks for clarifying @abdulasiraj I think what we should do in that case is get the common logic between these 2 functions (`as_ivy_dev` and `as_native_dev`) as a private helper which both of these functions would make use of, happy to know what you think 😄

Yep, It'll be a good solution. let me try!
Thanks!

abdulasiraj:
Hi @vedpatwardhan,
I created the private functions where possible but It's not possible for most of the time!

vedpatwardhan (commented):
> Hi @vedpatwardhan, I created the private functions where possible but It's not possible for most of the time!

Actually in your PR, in the `backends/jax/device.py` for e.g., except for the `ivy.Device` calls in `as_ivy_dev`, rest of the code is the same as that in `as_native_dev`. We could avoid this code duplication through a single helper that executes this common functionality 😄

sladyn98 (commented):
The creation of ivy.Device object is repeated multiple times. This can be minimized.

abdulasiraj:
Hi @vedpatwardhan,

It's possible to add helper for torch and jax but for paddle it's not yet possible cuz paddle missing some functions. I've raised issue on paddlepaddle repo. 
https://github.com/PaddlePaddle/Paddle/issues/55977

ivy-seed:

vedpatwardhan:
Hey @abdulasiraj, are you creating a different PR for this issue?

vedpatwardhan:
> > Hey @abdulasiraj, are you creating a different PR for this issue?
> 
> Hey @vedpatwardhan, I was working on some other important tasks on testing side and bot marked this PR as staled so I consider this to close. Due you think we should work on this or do you have someone on framework to work on this?. let me know your thoughts. Thanks :)

Got it! If you're busy with something else, @RickSanchezStoic could you please look into getting this over the line? I think we're done for the most part just a final few changes remaining. Thanks 😄

abdulasiraj:
I'll create new PR for this

PR #21286, opened 2023-08-04
refactor paddle.sqrt to behave w.r.t. data api conventions

### LINKED ISSUE #21231: [Bug]: Inconsistent Behavior of paddle.sqrt in Ivy Backend with Non-Floating Dtypes Support
### Bug Explanation

As the [paddle.sqrt](https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/sqrt_en.html#sqrt) only supports floating dtypes (`float16`, `float32`, `float64`) natively, the current backend implementation of it in `ivy` is modified to add support for integer, unsigned, bool, and complex dtypes. However, this current implementation behavior does not match with any other framework that supports these dtypes and the value outputted seems to be wrong too due to casting it back to their original dtype .

Thus, I want to know which framework's behavior we should try to emulate with it, and upon reaching this resolution, I can raise a pull-request for it.

### Steps to Reproduce Bug

```
import numpy as np
import jax.numpy as jnp
import torch
import ivy


dtypes = [
    np.int8,
    np.int16,
    np.int32,
    np.int64,
    np.uint8,
    np.float16,
    np.complex64,
    np.complex128,
]

for dtype in dtypes:
    x = np.array([2], dtype=dtype)
    y = np.sqrt(x)

    print(f"Output:{y}, Input dtype: {x.dtype}, Output dtype: {y.dtype}")

dtypes = [
    torch.int8,
    torch.int16,
    torch.int32,
    torch.int64,
    torch.uint8,
    torch.complex64,
    torch.complex128,
    torch.bool,
]

for dtype in dtypes:
    x = torch.tensor([2], dtype=dtype)
    y = torch.sqrt(x)

    print(f"Output:{y},Input dtype: {x.dtype}, Output dtype: {y.dtype}")


dtypes = [
    jnp.int8,
    jnp.int16,
    jnp.int32,
    jnp.int64,
    jnp.uint8,
    jnp.float16,
    jnp.complex64,
    jnp.complex128,
    jnp.bool_,
]

for dtype in dtypes:
    x = jnp.array([2], dtype=dtype)
    y = jnp.sqrt(x)

    print(f"Output:{y},Input dtype: {x.dtype}, Output dtype: {y.dtype}")


dtypes = [
    ivy.int8,
    ivy.int16,
    ivy.int32,
    ivy.int64,
    ivy.uint8,
    ivy.complex64,
    ivy.complex128,
    ivy.bool,
]

for dtype in dtypes:
    ivy.set_backend("paddle")
    x = ivy.array([2], dtype=dtype)
    y = ivy.sqrt(x)
    print(f"Output:{y},Input dtype: {x.dtype}, Output dtype: {y.dtype}")
    
    
    """
# --- Numpy  --- #
Output:[1.414], Input dtype: int8, Output dtype: float16
Output:[1.4142135], Input dtype: int16, Output dtype: float32
Output:[1.41421356], Input dtype: int32, Output dtype: float64
Output:[1.41421356], Input dtype: int64, Output dtype: float64
Output:[1.414], Input dtype: uint8, Output dtype: float16
Output:[1.414], Input dtype: float16, Output dtype: float16
Output:[1.4142135+0.j], Input dtype: complex64, Output dtype: complex64
Output:[1.41421356+0.j], Input dtype: complex128, Output dtype: complex128

# --- Torch  --- #
Output:tensor([1.4142]),Input dtype: torch.int8, Output dtype: torch.float32
Output:tensor([1.4142]),Input dtype: torch.int16, Output dtype: torch.float32
Output:tensor([1.4142]),Input dtype: torch.int32, Output dtype: torch.float32
Output:tensor([1.4142]),Input dtype: torch.int64, Output dtype: torch.float32
Output:tensor([1.4142]),Input dtype: torch.uint8, Output dtype: torch.float32
Output:tensor([1.4142+0.j]),Input dtype: torch.complex64, Output dtype: torch.complex64
Output:tensor([1.4142+0.j], dtype=torch.complex128),Input dtype: torch.complex128, Output dtype: torch.complex128
Output:tensor([1.]),Input dtype: torch.bool, Output dtype: torch.float32

# --- Jax  --- #
Output:[1.4142135],Input dtype: int8, Output dtype: float32
Output:[1.4142135],Input dtype: int16, Output dtype: float32
Output:[1.4142135],Input dtype: int32, Output dtype: float32
Output:[1.41421356],Input dtype: int64, Output dtype: float64
Output:[1.4142135],Input dtype: uint8, Output dtype: float32
Output:[1.414],Input dtype: float16, Output dtype: float16
Output:[1.4142135+0.j],Input dtype: complex64, Output dtype: complex64
Output:[1.41421356+0.j],Input dtype: complex128, Output dtype: complex128
Output:[1.],Input dtype: bool, Output dtype: float32

# --- Paddle --- #
Output:ivy.array([1]),Input dtype: int8, Output dtype: int8
Output:ivy.array([1]),Input dtype: int16, Output dtype: int16
Output:ivy.array([1]),Input dtype: int32, Output dtype: int32
Output:ivy.array([1]),Input dtype: int64, Output dtype: int64
Output:ivy.array([1]),Input dtype: uint8, Output dtype: uint8
Output:ivy.array([1.4142135+0.j]),Input dtype: complex64, Output dtype: complex64
Output:ivy.array([1.41421356+0.j]),Input dtype: complex128, Output dtype: complex128
Output:ivy.array([ True]),Input dtype: bool, Output dtype: bool
    
    """

```

### Environment

Linux + vscode + docker

### Ivy Version

1.1.9

### Backend

- [ ] NumPy
- [ ] TensorFlow
- [ ] PyTorch
- [ ] JAX
- [x] Paddle

### Device

CPU

akshatvishu:
Close #21231


sagewhocodes:
Hello @akshatvishu , could you please link the related open task/todo list for this issue? 
Thanks!

Pchatain:
Hey @akshatvishu , thanks for the detailed issue and follow up pull request. @sagewhocodes it seems like this potentially isn't a TODO list task? Here's the related issue that @akshatvishu opened: https://github.com/unifyai/ivy/issues/21231 

@akshatvishu I dug into the CI errors, and I noticed that bfloat16 isn't supported. Maybe we don't need to support this, but since torch supports bfloat16, I think we should try to add it. Let me know if paddle doesn't support bfloat16. To be specific, I would expect bfloat16 to stay a bfloat16 after doing sqrt since this is what pytorch does, and usually if people are using bfloat16 they are doing so for performance reasons and would want to keep the dtype the same size.

akshatvishu:
> Hey @akshatvishu , thanks for the detailed issue and follow up pull request. @sagewhocodes it seems like this potentially isn't a TODO list task? Here's the related issue that @akshatvishu opened: #21231
> 
> @akshatvishu I dug into the CI errors, and I noticed that bfloat16 isn't supported. Maybe we don't need to support this, but since torch supports bfloat16, I think we should try to add it. Let me know if paddle doesn't support bfloat16. To be specific, I would expect bfloat16 to stay a bfloat16 after doing sqrt since this is what pytorch does, and usually if people are using bfloat16 they are doing so for performance reasons and would want to keep the dtype the same size.

Sure,we can add in an addition `elif` condition for `bfloat16`. Also, I was looking at other backend `sqrt` implementation and it seems like we've not modified our tensorflow backend to support integer and unsigned dtypes, Thus, let me know if we would like to add this functionality to our `TF` backend too or not .

- Current, `sqrt` implementation at our backend -> [code](https://github.com/unifyai/ivy/blob/master/ivy/functional/backends/tensorflow/elementwise.py#L700)
- `tf.sqrt` official documentation -> [link](https://www.tensorflow.org/api_docs/python/tf/math/sqrt)

akshatvishu:
Hey!  @Pchatain , I was thinking that we can our code more readable with adding a helper function like below determine the casting logic instead of a dictionary map, let me know what are your thoughts on it :
```py
def _determine_dtype_cast(dtype: Type[paddle.Tensor]) -> Tuple[Optional[str], Optional[str]]:
    """Determine the appropriate casting dtype for sqrt operations.
    
    Returns:
        (intermediate_dtype, output_dtype)
    """
    common_types = {
        paddle.uint8, 
        paddle.int8, 
        paddle.int16, 
        paddle.int32, 
        paddle.float16, 
        paddle.bool
    }
    
    if dtype in common_types:
        return "float32", "float32"
    elif dtype == paddle.int64:
        return "float64", "float64"
    elif dtype == paddle.bfloat16:
        return "float32", "bfloat16"
    else:
        return None, None

```

sagewhocodes:
Thanks for the information @Pchatain 😄 , I'm assigning this issue to you since you have been assisting with in on #21231

Pchatain:
Hey @akshatvishu , Yes I think adding that helper function is more readable, and hopefully should see use in other functions that need to do this dtype promotion. Are you sure it doesn't exist elsewhere though? If not, yeah add it to this PR!

akshatvishu:
Hola @Pchatain 

I've carefully examined our existing utilities, and I noticed the `ivy.promote_types_of_inputs` function at the beginning of our file. However, my understanding is that this function would primarily be beneficial when combining two tensors in operations, ensuring they have a common dtype prior to performing the operation (such as addition or multiplication). It doesn't seem tailored for handling dtype casting for unary operations, specifically like` sqrt`. Thus, I've taken a different approach. Let me know your thoughts!

Pchatain:
Yeah I like this a lot!

Pchatain:
It looks like a few tests are failing for e.g. uint8, and for float16 being promoted to float32 upon return. I think as discussed above if the problem is with the test itself, either mark it as pytest skip in the case that paddle.sqrt doesn't support that dtype, or fix the test if the problem is with the test itself. After that, let me know if you are finished and I'll double check it and then merge it!

akshatvishu:
Hey @Pchatain  , 


I think `ivy` still not support `bfloat16` dtype for `paddle` and that's why we're seeing this unusual behavior for it in our test.

Here, is one example to showcase this:
```py
import ivy

ivy.set_backend("paddle")
x = ivy.array(-1, dtype=ivy.bfloat16)
print(f"x:{x}, x_dtype:{x.dtype}") # x:ivy.array(49024), x_dtype:bfloat16
x_ =ivy.sqrt(x)
print(f"x_:{x_}, x_dtype:{x_.dtype}") # x_:ivy.array(65472), x_dtype:bfloat16
```
Hence,  I filter it out while testing . (Do tell, if you want it to be handled any other way than this!!)


Also, for `paddle.sqrt`, the `float16` dtype is not yet implemented for it's `CPU` kernel but the official doc for it ->[link](https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/sqrt_en.html#sqrt) don't state it explicitly for some reason.I tested it locally and confirmed that it is indeed the case:

```py
import paddle
print(f"p:{paddle.sqrt(paddle.to_tensor([2.], dtype=paddle.float16))}")
"""
    print(f"p:{paddle.sqrt(paddle.to_tensor([2.], dtype=paddle.float16))}")
  File "/opt/fw/paddle/paddle/tensor/ops.py", line 920, in sqrt
    return _C_ops.sqrt(x)
RuntimeError: (NotFound) The kernel with key (CPU, Undefined(AnyLayout), float16) of kernel `sqrt` is not registered. Selected wrong DataType `float16`. Paddle support following DataTypes: float64, float32.
  [Hint: Expected kernel_iter == iter->second.end() && kernel_key.backend() == Backend::CPU != true, but received kernel_iter == iter->second.end() && kernel_key.backend() == Backend::CPU:1 == true:1.] (at /paddle/paddle/phi/core/kernel_factory.cc:199)
"""
```

Thus,  I ended up casting it to `float32` too for calculating the `sqrt` then returning it as `float16` with the `ToDo` note added to remove it once, they add the `CPU` kernel for it!


Do let me know what do you think of these changes!

Pchatain:
Yes that makes sense. Thanks for diving in to verify that ivy does not support the bfloat16 dtype for the paddle_backend, and your solution to filter it out during testing makes sense.

Let me know if it's done so I can merge it.

akshatvishu:
> Yes that makes sense. Thanks for diving in to verify that ivy does not support the bfloat16 dtype for the paddle_backend, and your solution to filter it out during testing makes sense.
> 
> Let me know if it's done so I can merge it.

I've included the suggest changes as mentioned above !!

akshatvishu:
Thanks for the review @Pchatain !!

PR #21291, opened 2023-08-04
paddle frontend cholesky_solve added

### LINKED ISSUE #21182: cholesky_solve

dhanush-2501:
Close #21182


dhanush-2501:
Hi @AnnaTz, can you pls check on this?

ivy/functional/frontends/paddle/tensor/linalg.py
```diff
@@ -122,6 +122,16 @@ def solve(x1, x2, name=None):
     return ivy.solve(x1, x2)
 
 
+# cholesky_solve
+@with_supported_dtypes({"2.5.1 and below": ("float32", "float64")}, "paddle")
+@to_ivy_arrays_and_back
+def cholesky_solve(y, x, /, *, upper=False, name=None):
```
AnnaTz, line 128: The original function puts `x` before `y`. We should do so as well; the signature should be exactly the same.

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_linalg.py
```diff
@@ -528,6 +528,87 @@ def test_paddle_solve(
...
+        )
+    )
+    shared_size = draw(
+        st.shared(helpers.ints(min_value=2, max_value=4), key="shared_size")
+    )
+    return input_dtype, draw(
+        helpers.array_values(
+            dtype=input_dtype, shape=tuple([shared_size, 1]), min_value=2, max_value=5
+        )
+    )
```
AnnaTz, line 576: You don't need to copy-paste other tests' helpers. It's cleaner to just import them from the file you found them in.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_linalg.py
dhanush-2501: Actually I tried that but that function was expecting a tensor. And the tests didn't pass. Yet still i would give a try again

ivy/functional/frontends/paddle/tensor/linalg.py
dhanush-2501: Actually that was a bug in their documentation. I found that, reported and fixed that to them as well. thats the reason for that confusion. link to the paddle paddle bug : https://github.com/PaddlePaddle/Paddle/issues/55998

AnnaTz (approved):
Looks good, thanks!

PR #21314, opened 2023-08-05
Fix flatten_frontend_to_np to use the right backend

somaia02:
My test hangs when I call flatten_frontend_to_np. This code below also shows the problem.
```python
import ivy

ivy.set_backend("torch")

x=torch.tensor(1.)
y = helpers.flatten_frontend_to_np(ret=x, backend="torch")
```


somaia02:
Hello @CatB1t What do you think about this? Is there something else that I could do in the test itself to avoid this error?

somaia02:
I came up with another solution. Might be better.

PR #21369, opened 2023-08-06
Add increment method to Paddle frontend

### LINKED ISSUE #21294: increment

FelixHuyghe:
Close #21294

### DIFF
--- a/ivy/functional/frontends/paddle/tensor/math.py
+++ b/ivy/functional/frontends/paddle/tensor/math.py
@@ -412,3 +412,17 @@ def any(x, axis=None, keepdim=False, name=None):
 @to_ivy_arrays_and_back
 def diff(x, n=1, axis=-1, prepend=None, append=None, name=None):
     return ivy.diff(x, n=n, axis=axis, prepend=prepend, append=append)
+
+
+@with_supported_dtypes(
+    {"2.5.0 and below": ("float32", "float64", "int32", "int64")}, "paddle"
+)
+@to_ivy_arrays_and_back
+def increment(x, value=1.0):
+    if x.size == 1:
+        return ivy.inplace_increment(x, value)
+    else:
+        raise ValueError(
+            "The number of elements in Input(X) should be 1.Now the number is"
+            f" {x.size}."
+        )
--- a/ivy/functional/frontends/paddle/tensor/math.py
+++ b/ivy/functional/frontends/paddle/tensor/math.py
@@ -4,8 +4,6 @@ import ivy
 from ivy.func_wrapper import with_unsupported_dtypes, with_supported_dtypes
 from ivy.functional.frontends.paddle.func_wrapper import to_ivy_arrays_and_back
 
-import paddle
-
 
 @with_unsupported_dtypes({"2.5.1 and below": ("float16", "bfloat16")}, "paddle")
 @to_ivy_arrays_and_back
@@ -423,7 +421,8 @@ def diff(x, n=1, axis=-1, prepend=None, append=None, name=None):
 @to_ivy_arrays_and_back
 def increment(x, value=1.0):
     if x.size == 1:
-        return ivy.inplace_increment(x, value)
+        return ivy.inplace_increment(ivy.copy_array(x), value)
     else:
-        # Call paddle to return the correct exception
-        return paddle.increment(paddle.to_tensor(ivy.to_native(x)), value)
+        return ValueError(
+            "(InvalidArgument) The number of elements in Input(X) should be 1."
+        )
--- a/ivy/functional/frontends/paddle/tensor/math.py
+++ b/ivy/functional/frontends/paddle/tensor/math.py
@@ -1,7 +1,10 @@
 # global
+
 import ivy
 from ivy.func_wrapper import with_unsupported_dtypes, with_supported_dtypes
 from ivy.functional.frontends.paddle.func_wrapper import to_ivy_arrays_and_back
+
+import paddle
 
 
 @with_unsupported_dtypes({"2.5.1 and below": ("float16", "bfloat16")}, "paddle")
@@ -415,14 +418,12 @@ def diff(x, n=1, axis=-1, prepend=None, append=None, name=None):
 
 
 @with_supported_dtypes(
-    {"2.5.0 and below": ("float32", "float64", "int32", "int64")}, "paddle"
+    {"2.5.1 and below": ("float32", "float64", "int32", "int64")}, "paddle"
 )
 @to_ivy_arrays_and_back
 def increment(x, value=1.0):
     if x.size == 1:
         return ivy.inplace_increment(x, value)
     else:
-        raise ValueError(
-            "The number of elements in Input(X) should be 1.Now the number is"
-            f" {x.size}."
-        )
+        # Call paddle to return the correct exception
+        return paddle.increment(paddle.to_tensor(ivy.to_native(x)), value)
--- a/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
+++ b/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
@@ -951,7 +951,7 @@ def test_paddle_lgamma(
         test_flags=test_flags,
         fn_tree=fn_tree,
         on_device=on_device,
-        atol=1E-4,
+        atol=1e-4,
         x=x[0],
     )
 
@@ -1863,4 +1863,25 @@ def test_paddle_diff(
         axis=axis,
         prepend=prepend[0],
         append=append[0],
+    )
+
+
+def test_paddle_increment(
+    *,
+    dtype_and_input_and_label,
+    frontend,
+    test_flags,
+    fn_tree,
+    backend_fw,
+):
+    input_dtypes, input_and_label = dtype_and_input_and_label
+    input, label = input_and_label
+    helpers.test_frontend_function(
+        input_dtypes=input_dtypes,
+        backend_to_test=backend_fw,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        input=input,
+        label=label,
     )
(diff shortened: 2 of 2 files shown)


FelixHuyghe:
Close #21294

ivy/functional/frontends/paddle/tensor/math.py
```diff
@@ -412,3 +412,17 @@ def any(x, axis=None, keepdim=False, name=None):
...
 def diff(x, n=1, axis=-1, prepend=None, append=None, name=None):
     return ivy.diff(x, n=n, axis=axis, prepend=prepend, append=append)
+
+
+@with_supported_dtypes(
+    {"2.5.0 and below": ("float32", "float64", "int32", "int64")}, "paddle"
+)
+@to_ivy_arrays_and_back
+def increment(x, value=1.0):
+    if x.size == 1:
```
hmahmood24, line 422: `x.size` returns a list. Maybe you forgot to take its length?

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
```diff
@@ -1864,3 +1864,24 @@ def test_paddle_diff(
         prepend=prepend[0],
         append=append[0],
     )
+
+
```
hmahmood24, line 1868: You didn't generate any input strategies for the test using the `@handle_frontend_test` decorator. Kindly add these and run the test using `pytest` to see if it passes with all backends or not

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
```diff
@@ -1864,3 +1864,24 @@ def test_paddle_diff(
...
+    input_dtypes, input_and_label = dtype_and_input_and_label
+    input, label = input_and_label
+    helpers.test_frontend_function(
+        input_dtypes=input_dtypes,
+        backend_to_test=backend_fw,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        input=input,
+        label=label,
```
hmahmood24, line 1886: The input parameters are `x` and `value`, not `input` and `label`. It seems to me as if you copied this test from somewhere but forgot to modify it for your own use case :)

hmahmood24 (changes requested):
Hey @FelixHuyghe, thanks for the PR! Requesting some changes on the PR. Thank you :)

ivy/functional/frontends/paddle/tensor/math.py
FelixHuyghe: I am a bit confused by this: if type in the terminal something like `a = ivy.array([3,4,5,6,7]` and then do `a.size` I just get 5, I want to check if the whole array contains exactly one element.

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
FelixHuyghe: Yes indeed sorry for that

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
FelixHuyghe: Thank you, I did not really understand how the testing worked indeed. In my latest commit I have added a `@handle_frontend_test` but my tests are not working. 
For example with input data `[0]` and value 1 I get the error that my solution is 1 while the groundtruth solution is 2.
After inspecting the testing method, I think that this is caused because the same array is used to test my implementation of the frontend function first and then the ground truth version later. The problem is that a side effect of the padddle.increment value is that the data that you give it is incremented itself. So when the frontend is called with x=`[0]` and value=1, the correct value of 1 is returned, but causes `x=[1]` which is why when the groundtruth value is calculated afterwards it adds 1 to 1 and says that the 1 != 2. I am a bit lost in how to properly test this method, so hope you can help me!

ivy/functional/frontends/paddle/tensor/math.py
```diff
@@ -412,3 +415,15 @@ def any(x, axis=None, keepdim=False, name=None):
...
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float32", "float64", "int32", "int64")}, "paddle"
+)
+@to_ivy_arrays_and_back
+def increment(x, value=1.0):
+    if x.size == 1:
+        return ivy.inplace_increment(x, value)
+    else:
+        # Call paddle to return the correct exception
+        return paddle.increment(paddle.to_tensor(ivy.to_native(x)), value)
```
hmahmood24, line 429: We are not supposed to import any of the native frameworks (`paddle`, `torch`, `tensorflow` etc.) in our frontends. The frontend functions are only supposed to be implemented as a composition of `ivy` functions only. You'll have to remove this and the `paddle` import up top

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
hmahmood24: @FelixHuyghe Try copying the array using something like `ivy.copy_array` before passing it to `ivy.inplace_increment`. See if that works for you and let me know :)

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
```diff
@@ -1864,3 +1866,37 @@ def test_paddle_diff(
...
+    dtype_and_x_and_value,
+    frontend,
+    test_flags,
+    fn_tree,
+    backend_fw,
+):
+    input_dtypes, x_and_value = dtype_and_x_and_value
+    value = x_and_value[1]
+    # Get a single element from the array to use as value parameter
+    value_first_element = value[0].item()
```
hmahmood24, line 1893: This will throw an error since only 1-element arrays are converted to python scalars by calling `.item()` otherwise you have to pass in an index to `.item(...)` to get this working properly

ivy/functional/frontends/paddle/tensor/math.py
hmahmood24: @FelixHuyghe Why is this `if` condition needed anyways?

hmahmood24 (changes requested):

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
FelixHuyghe: Oh makes sense, I changed this

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
FelixHuyghe: Yes that works! I copied the array before passing it to ivy.inplace_increment and now my tests pass! However, I think this implementation does not fully reflect the original paddle one, because if you pass an array to the paddle function, a side effect is that array is incremented "in place" and with my frontend implementation when the function returns the provided array is still the same.

ivy/functional/frontends/paddle/tensor/math.py
FelixHuyghe: The paddle function accepts only inputs with one element: 
`paddle.increment(x, value=1.0, name=None)
The API is usually used for control flow to increment the data of x by an amount value. Notice that the number of elements in x must be equal to 1.`
The problem is that ivy.increment_incplace accepts inputs with any shape, so I thought I check if the size is equal to one and if not I return an exception. However if I try to test this by setting the max_num_size parameter for x to something larger then 1, the tests fail and I get the exception I wrote myself. I was trying to test that when you provide an input array with a size larger thant 1 the function indeed raises an exception, just like the original Paddle function. But I don't think I am doing it in the correct way if you could help me :)

ivy/functional/frontends/paddle/tensor/math.py
FelixHuyghe: Yes I can understand, I removed it now. I commented above a bit about why I tried doing this :)

FelixHuyghe:
Hey, I got the tests (partially) working! I do however have two main remaining concern. I left comments on your old comments, so I think you will have to scroll up quite a bit to get to them :)

ivy/functional/frontends/paddle/tensor/math.py
hmahmood24: > The paddle function accepts only inputs with one element: `paddle.increment(x, value=1.0, name=None) The API is usually used for control flow to increment the data of x by an amount value. Notice that the number of elements in x must be equal to 1.` The problem is that `ivy.increment_incplace` accepts inputs with any shape, so I thought I check if the size is equal to one and if not I return an exception. However if I try to test this by setting the `max_num_size` parameter for x to something larger then 1, the tests fail and I get the exception I wrote myself. I was trying to test that when you provide an input array with a size larger than 1 the function indeed raises an exception, just like the original Paddle function. But I don't think I am doing it in the correct way, if you could help me :) I was also wondering if in this case the frontend function should return the exact same exception as when paddle.increment is called directly. This is why in my previous commit I tried calling paddle.increment directly from the frontend, to get the exact same exception, but I understand this is not the way to go.

@FelixHuyghe You don't need to test for values that you already know don't fall into the acceptable range of values for this function. The unit test is required to provide complete coverage of an acceptable range of values and so only values that the original function should run fine on should be generated with the unit tests if that makes sense. :)

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
hmahmood24: > Yes that works! I copied the array before passing it to ivy.inplace_increment and now my tests pass (when x is of shape [1] anyways)! However, I think this implementation does not fully reflect the original paddle one, because if you pass an array to the paddle function, a side effect is that array is incremented "in place" and with my frontend implementation when the function returns the provided array is still the same (because of the copy).

I see what you mean. I think the test for `ivy.inplace_increment` should give you an idea of how to test this frontend function because it is kind of similar to what's happening with `ivy.inplace_increment` I would think so.

hmahmood24 (changes requested):
Hey @FelixHuyghe, left some replies to your comments. Let me know if they help?

hmahmood24:
Closing this PR since it has been stale for more than 2 weeks now.

PR #21397, opened 2023-08-07
Remove redundant statement in `backends.numpy.experimental.statistical.cov`

joshiayush:
NumPy already handles the conversion of "fweights" input frequency weights vector into python "float" [see](https://github.com/numpy/numpy/blob/3b99af4b075af9fb44a869d3b4726f368fe77524/numpy/lib/function_base.py#L2703-L2704). While the backend function adds a statement to convert "fweights" input frequency vector into "np.int64" which will have no effect once the numpy function is called.


joshiayush:
@AnnaTz This PR needs review

AnnaTz (approved):
Looks good!

PR #21404, opened 2023-08-07
add jax.numpy.lexsort and corresponding tests

### LINKED ISSUE #21403: lexsort

illia-bab:
Close #21403


ivy-leaves:

Thank you :hugs:

ivy/functional/backends/paddle/experimental/sorting.py
```diff
@@ -19,6 +19,13 @@ def invert_permutation(
 def lexsort(
     keys: paddle.Tensor, /, *, axis: int = -1, out: Optional[paddle.Tensor] = None
 ) -> paddle.Tensor:
+    if keys.dtype in [
```
zaeemansari70, line 22: Why are we upcasting these? :slightly_smiling_face: 
Can't we handle this by adding the `unsupported_dtypes` decorators above this function?

zaeemansari70: According to our discussion as the dtypes suppported by this are `[int32, int64, float32, float64]`. Adding a logic here where we upcast to the closest possible `dtype` would make more sense I guess, what you can do is upcast to `float32` if its a float `dtype` and similarly cast to a closest `int` `dtype` if we get an unsupprted `int` dtype. We will actually be handling a broader set in this way. 
Hope that makes sense :+1:

zaeemansari70 (changes requested):
Just required a small change

zaeemansari70:
Also, @MahmoudAshraf97 can you take a last look at this as this is related to the paddle backend as also the dtypes handling in paddle backend specifically. 
Thanks!

MahmoudAshraf97 (changes requested):
Hello, the new change in the paddle backend doesn't ensure that the returned dtype is the same as the input, please make sure the tests for the functional function is passing in the paddle backend

PR #21406, opened 2023-08-07
Add manipulation function `block` to functional api

joshiayush:


ivy-leaves:

Thank you :hugs:

joshiayush:
@MuhammadSaeedBatikh Could you please review this one?

JPGoodale:
Hi @joshiayush, due to this being a new addition to the functional api, please add it to the `experimental/manipulation` directory rather than the main one, also whenever adding a new Ivy function it needs to be implemented as a method of the `Array` and `Container` dataclasses as well, both in their respective `experimental` folders. If you have any questions about this, feel free to ask!

joshiayush:
> Hi @joshiayush, due to this being a new addition to the functional api, please add it to the `experimental/manipulation` directory rather than the main one, also whenever adding a new Ivy function it needs to be implemented as a method of the `Array` and `Container` dataclasses as well, both in their respective `experimental` folders. If you have any questions about this, feel free to ask!

Working on it :+1:

joshiayush:
@JPGoodale I added the definition to both of the dataclasses. Do I have to make further changes or should I start with writing the tests?

JPGoodale:
Hey @joshiayush, your dataclass implementations look good so you can go ahead and write the tests. As for the other backends, if there's no native equivalent then it needs be implemented compositionally in Ivy, thereby making it a mixed function, `ivy.pad` is a good example of this as it also directly calls the native equivalent in `jax` and `numpy` but is implemented compositionally for the other backends here: https://github.com/unifyai/ivy/blob/main/ivy/functional/ivy/experimental/manipulation.py#L981.

joshiayush:
@JPGoodale I have a problem with the `test_numpy_block` function, when I run this function it raises:

```
TypeError: Field elements must be 2- or 3-tuples, got ''uint16''
```

Could you please provide a solution?

joshiayush:
@JPGoodale Right now the test I've implemented is made considering that the `numpy` frontend function `block` now supports the `out` argument but the original implementation still doesn't take more than one argument, because of this reason at the time of comparing the result with the associated framework function raises an error that `"numpy() function takes only 1 positional argument but two were given..."`

How should I go about fixing it?

joshiayush:
@JPGoodale Could you please help me with the tests. Ideally a value generating function generating arrays with multiple dimensions should work but in my case the test result from the ground truth backend and the ivy backend differs 🤔

I'm unable to spot the problem here. Could you please look at my test function and the value generating function?

JPGoodale:
@joshiayush it looks like you currently only have tests for the frontend functions, please add tests for the backends as well and get those passing before working with the frontends, it also seems that you don't have a torch or paddle backend implementation, every backend must be handled if adding a new function to the API. Please move your `jax` and `numpy` backend implementations to the`experimental` directory rather than the main one as well.

PR #21421, opened 2023-08-07
Add handle_complex_input to documentation

jshepherd01:
1. Add the new `@handle_complex_input` decorator to the "Function Wrapping" deep dive documentation
2. Modify the list of function wrappers so that it gets numbered automatically, rather than having to re-number it every time one gets added or moved, which will hopefully keep `git diff` sizes down
3. Add a link to the doc-builder repo to the "Building the Docs Pipeline" deep dive, so that other contributors in the future don't have to search Discord to find out where the `make_docs_without_docker` script is


KareemMAX (approved):
Hey @jshepherd01,
Building the docs doc LGTM, great work. I guess maybe @Ishticode might take a look into the function wrapper doc just to be sure.

Thanks 😄

Ishticode (changes requested):
Sorry for the delay.
I think its best to update the full wrappers order list from the list [`FN_DECORATORS`](https://github.com/unifyai/ivy/blob/4a6abad477120346d2bebb2eaa8859e8e6f27121/ivy/func_wrapper.py#L17). For example, docs still have `@handle_mixed_function` which doesn't exist anymore (and is perhaps adopted in `@handle_partial_mixed_function`).
Goood to merge after that. Thanks 😄

PR #21447, opened 2023-08-08
Fixed the failing tests for the acos method in PyTorch frontend.

he11owthere:
closes https://github.com/unifyai/ivy/issues/20773

image

### DIFF
--- a/ivy/functional/frontends/torch/tensor.py
+++ b/ivy/functional/frontends/torch/tensor.py
@@ -584,7 +584,7 @@ class Tensor:
         self.ivy_array = self.arctan2(other).ivy_array
         return self
 
-    @with_unsupported_dtypes({"2.0.1 and below": ("float16",)}, "torch")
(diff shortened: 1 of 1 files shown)


vedpatwardhan:
Hey @he11owthere, actually the function supports `bfloat16` but there's an issue in the `ivy.Array.__repr__` method which is currently being fixed. Could you please close this and the other PRs created for a similar reason? 
Thanks for helping us spot the issue, the `ivy.Array.__repr__` method is the root cause, all these functions should work as expected once that's fixed 😄

he11owthere:
@vedpatwardhan Alright I am closing it right away along with the other PRs which are inappropriate by checking if the data types are supported by the original PyTorch functions. I will try my best to see if there is any way to fix the function you mentioned above. Sorry if I caused any problem to the codebase and will try my best to be helpful. Thank you very much! :blush:

PR #21488, opened 2023-08-08
Feat: add percentile support for numpy frontend

### LINKED ISSUE #21426: <Percentile>
- [] #17469

kristinazk:
Closes #21426


kristinazk:
Hi, @sherry30 I made the necessary changes, could you please review them?

ivy/functional/frontends/numpy/statistics/order_statistics.py
```diff
@@ -48,6 +48,68 @@ def ptp(a, axis=None, out=None, keepdims=False):
...
+@handle_numpy_out
+def percentile(
+    a,
+    q,
+    axis=None,
+    out=None,
+    overwrite_input=False,
+    method="linear",
+    keepdims=False,
+    interpolation=None,
```
sherry30, line 60: This is a keyword-only arg according to the [docs](https://numpy.org/doc/stable/reference/generated/numpy.percentile.html)

ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics/test_order_statistics.py
```diff
@@ -42,6 +42,48 @@ def test_numpy_ptp(
...
+    fn_tree,
+    backend_fw,
+    on_device,
+    keep_dims,
+):
+    input_dtypes, values, axis = dtype_values_axis
+    if isinstance(axis, tuple):
+        axis = axis[0]
+
+    where, input_dtypes, test_flags = np_frontend_helpers.handle_where_and_array_bools(
```
sherry30, line 66: This function doesn't accept `where` argument so no need to do this.

ivy_tests/test_ivy/test_frontends/test_numpy/test_statistics/test_order_statistics.py
```diff
@@ -42,6 +42,48 @@ def test_numpy_ptp(
...
+        where=where,
+        input_dtype=input_dtypes,
+        test_flags=test_flags,
+    )
+
+    np_frontend_helpers.test_frontend_function(
```
sherry30, line 72: you should use `helpers.test_frontend_function` for this

sherry30 (changes requested):
Hi,
Requested a few changes

sherry30 (changes requested):
Its still failing with the same error
`ValueError: Percentiles must be in the range [0, 100]`
Can you try to fix it.
You can check these failures in the CI

kristinazk:
@sherry30 everything runs correctly when performing the tests with the same inputs manually, but somehow the tests fail to succeed and I need help resolving this issue.

kristinazk:
hi @sherry30 the error occurs because I am using the '_quantile_is_valid' function which was there previously, I don't think there should be anything wrong with that. However, I have some issues concerning the testing of the percentile function: I fail all the tests with the following errors: 
```
AssertionError: Frontend function returned non-frontend arrays: []
FAILED test_order_statistics.py::test_numpy_percentile[cpu-paddle-False-False] - ivy.utils.exceptions.IvyBackendException: paddle: divide: paddle: is_float_dtype: paddle: as_ivy_dtype: dtype('float64')

```

And I am not able to understand what's wrong with the code, as I have debugged it several times and compared my results with the ones numpy produces and everything was identical.

sherry30 (changes requested):
This was a bug we encountered last week, but has been fixed, can you update your branch, probably will be fixed after that,

Thanks

kristinazk:
I have fixed my branch, but I still fail the tests with errors mentioned earlier.

sherry30 (changes requested):
So I debugged the test a bit, and its because there are some instances where nothing gets returned. this is why the assertion was returning `False`, this shouldn't happen, there should always be something returned and if the input is invalid then it should raise an error and not return nothing

kristinazk:
Hi, I've made all the necessary changes (adding the exception handling functionality just like you described), however there are still some tests which do not pass. I observe a couple of reasons for that:
1. The tests fail if an exception is thrown based on `_quantile_is_valid` helper's output (in the tests run by me manually, the code works just fine and throws exceptions when needed).
2. The results of some other failed tests display the errors of the `nanpercentile` function (I don't understand why this is the case).

sherry30 (changes requested):
I still get the `ivy.utils.exceptions.IvyException: Percentiles must be in the range [0, 100]` error after running tests. Can you modify the helper function so it doesn't produce inputs that raise this exception. Feel free to add an argument to it so the other tests which use this helper function dont start failing.

kristinazk:
Hi, so I checked the `_quantile_is_valid` function and it seems like it's working just fine. I checked the failing tests, where the ` ivy.utils.exceptions.IvyException: Percentiles must be in the range [0, 100]` exception is being thrown, and everything is correct, given the input value of q:

1st case: a = ivy.array(-1.), q = ivy.array(-1.), axis = 0, out = None, overwrite_input = None, method = None, keepdims = False

2nd case:  a = -1.0, q = -1.0, axis = 0, out = None, overwrite_input = None, method = None, keepdims = False

The errors are supposed to be thrown and they are, in this case I don't understand why the test case is failing, the value of q is clearly out of range.

sherry30 (changes requested):
Hey,

So if the error is being thrown correctly, the test will still fail.
you shouldn't generate input which causes this exception to be thrown otherwise the test will fail.

sherry30:
Not sure about the error you're facing, but when I tested it, I got the following error.
`AssertionError: returned dtype = bool, ground-truth returned dtype = float64`
on the following input:
`dtype_values_axis=(['float32'], [array([0., 0.], dtype=float32)], 0)`

Not sure why is it producing bool, can you debug this?

kristinazk:
Hi, the problem was in the test input, and I fixed it. However, the error mentioned earlier (`AssertionError: returned shape = (), ground-truth returned shape = (2,)`) still persists. Additionally, now I get another errors like these: 
`FAILED test_order_statistics.py::test_numpy_percentile[cpu-tensorflow-False-False] - ivy.utils.exceptions.IvyBackendException: numpy: nested_map: numpy: nested_map: numpy: as_native_dtype: Cannot convert to numpy dtype. bfloat16 is not supported by NumPy.
FAILED test_order_statistics.py::test_numpy_percentile[cpu-torch-False-False] - ivy.utils.exceptions.IvyBackendException: torch: nested_map: torch: nested_map: torch: nested_map: torch: dtype: torch: nested_map: torch: nested_map: torch: nested_map: torch: to_device: Can...
`
Is this because of an incorrectly generated test input?

joaozenobio:
Hi @kristinazk ! Sorry for the delay on getting an answer for your PR. I have taken a look into your code and it seems your percentile function only supports as input axis equal to 0 and 1. It seems another function in the same file has the same behavior, which could be misleading you. Numpy percentile function works with any number of dimensions and the axis could be any integer. Therefore, you should add support to this functionality for your function to work. Any doubts you may have please pin me :smile:

joaozenobio:
Hi @kristinazk ! Sorry for the delay on getting an answer for your PR. Could you please inform the input the log of the errors you are facing? :smile:

Ishticode:
closing as it is labeled stale by ivy-seed. Feel free to reopen if you would like to work on this.

PR #21489, opened 2023-08-08
Check if all global modes are being used correctly as read-only properties

NripeshN:


ivy/__init__.py
```diff
@@ -1416,6 +1416,23 @@ def cast_data_types(val=True):
     "dynamic_backend",
     "precise_mode",
     "soft_device_mode",
+    "logging_mode",
+    "default_dtype",
+    "default_float_dtype",
+    "default_int_dtype",
+    "default_complex_dtype",
+    "default_uint_dtype",
+    "global_attr",
```
vedpatwardhan, line 1425: Could you please remove the ones added on and after this line? They don't really fall in the same category as the global modes. Thanks for looking into this @NripeshN 😄

vedpatwardhan (changes requested):

ivy/__init__.py
NripeshN: Thank you soo much @vedpatwardhan

PR #21490, opened 2023-08-08
Adding dtype specific tolerances

mobley-trent:
This PR adds the option to specify rtol and atol values according to the dtypes specified by the user. 

CC: [Trello](https://trello.com/c/IFpycQDJ/2310-adding-dtype-specific-tolerances-to-the-tests)


ivy_tests/test_ivy/helpers/assertions.py
```diff
@@ -148,6 +152,12 @@ def value_test(
             )
     else:
         for ret_np, ret_np_from_gt in zip(ret_np_flat, ret_np_from_gt_flat):
+            if specific_tolerance_dict is not None:
```
sherry30, line 155: above in the `if` statement we're only checking if `rolt` is given or not then we're getting `rtol` from global `TOLERANCE_DICT`. This should be modified so that it would first check if `specific_tolerance_dict` exists if it doesn't then it would check if `rtol` exists, if that doesn't then it would get it from the global `TOLERANCE_DICT`.

the priority should be in the following order:
`specific_tolerance_dict`
`rtol`
`TOLERANCE_DICT`

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
```diff
@@ -124,6 +125,8 @@ def test_max(*, dtype_and_x, keep_dims, test_flags, backend_fw, fn_name, on_devi
 )
 def test_mean(*, dtype_and_x, keep_dims, test_flags, backend_fw, fn_name, on_device):
     input_dtype, x, axis = dtype_and_x
+    tolerance_dict = defaultdict(lambda: 1e-2)
```
sherry30, line 128: Lets remove this since it adds unnecesary code, we can just pass `{"bfloat16": 1e-1}`
directly in the `tolerance_dict` argument below

sherry30 (changes requested):
Hi Edward,
your PR looks good for the most part.
A few things:
1. I don't think we need the defaultdict logic, we can just check if the given `dtype` exists in the `specific_tolerance_dict` (if its not None), if it doesn't then we can get the tolerance from the global `TOLERANCE_DICT`
2. Can you remove the changes in the array api test submodule.

mobley-trent:
> Hi Edward, your PR looks good for the most part. A few things:
> 
> 1. I don't think we need the defaultdict logic, we can just check if the given `dtype` exists in the `specific_tolerance_dict` (if its not None), if it doesn't then we can get the tolerance from the global `TOLERANCE_DICT`
> 2. Can you remove the changes in the array api test submodule.

@sherry30 I think using a `defaultdict` would be a good option in this case because it would allow us to set a single rtol for an arbitrary number of dtypes, as well as specifying the rtols for the rest of the dtypes explicitly; even for dtypes that are not in the global `TOLERANCE_DICT`.

ivy_tests/test_ivy/helpers/assertions.py
```diff
@@ -135,9 +138,23 @@ def value_test(
...
     )
     # value tests, iterating through each array in the flattened returns
-    if not rtol:
+    if specific_tolerance_dict is not None:
+        for ret_np, ret_np_from_gt in zip(ret_np_flat, ret_np_from_gt_flat):
+            dtype = str(ret_np_from_gt.dtype)
+            if specific_tolerance_dict.get(dtype) is not None:
+                rtol = specific_tolerance_dict.get(dtype)
+            else:
+                rtol = TOLERANCE_DICT.get(dtype, 1e-03)
```
sherry30, line 147: You can change this to `rtol = TOLERANCE_DICT.get(dtype, 1e-03) if rtol is None else rtol` since I would expect `rtol` of all the other dtypes to be the value I provided in the `rtol` argument and since thats in higher priority

sherry30 (changes requested):
In that case, you can make the global `TOLERANCE_DICT` `defaultdict`, because otherwise it would be annoying for developers to declare `defaultdict` in the frontend test function everytime.

ivy_tests/test_ivy/helpers/assertions.py
mobley-trent: @sherry30 I already added this logic when the rtol is `None` because if we pass in the tolerance dict we won't need the tolerance that is passed in by the user. It can be inferred from the global `TOLERANCE_DICT`

sherry30: Yes, but what I'm saying is the value passed by the user in the `rtol` argument is of higher priority than `TOLERANCE_DICT`. So we should infer from that instead if the specific dtype does not exist in `specific_tolerance_dict` and considering `rtol` was provided by the user.

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
```diff
@@ -132,6 +132,7 @@ def test_mean(*, dtype_and_x, keep_dims, test_flags, backend_fw, fn_name, on_dev
         on_device=on_device,
         rtol_=1e-1,
         atol_=1e-1,
+        tolerance_dict={"bfloat16": 1e-1},
```
sherry30, line 135: There's no point in adding this specific tolerance since already the rtol is 1e-1.

sherry30 (approved):
lgtm!

Can you also make changes to the docs describing this argument.
feel free to do that in a separate PR.

Thanks 😄

mobley-trent:
Merging right away 🚀

PR #21509, opened 2023-08-08
added take math function to paddle frontend

### LINKED ISSUE #21508: Take

Chosenundead15:
Close  #21508

### DIFF
--- a/ivy/functional/frontends/paddle/tensor/math.py
+++ b/ivy/functional/frontends/paddle/tensor/math.py
@@ -428,3 +428,17 @@ def diff(x, n=1, axis=-1, prepend=None, append=None, name=None):
 @to_ivy_arrays_and_back
 def mm(input, mat2, name=None):
     return ivy.matmul(input, mat2)
+
+
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float32", "float64", "int32", "int6")}, "paddle"
+)
+@to_ivy_arrays_and_back
+def take(
+    x,
+    index,
+    mode=None,
+    name=None,
+):
+    x = ivy.reshape(x, (-1,))
+    return ivy.gather(x, index, axis=0)
--- a/ivy/functional/frontends/paddle/tensor/math.py
+++ b/ivy/functional/frontends/paddle/tensor/math.py
@@ -437,8 +437,18 @@ def mm(input, mat2, name=None):
 def take(
     x,
     index,
-    mode=None,
+    mode="raise",
     name=None,
 ):
+    if mode not in ["raise", "wrap", "clip"]:
+        raise ValueError(
+            "'mode' in 'take' should be 'raise', 'wrap', 'clip', but received {}."
+            .format(mode)
+        )
     x = ivy.reshape(x, (-1,))
+    if mode == "clip":
+        index = ivy.clip(index, 0, x.shape[-1] - 1)
+    elif mode == "wrap":
+        index = ivy.where(index < 0, index % x.shape[-1], index)
+        index = ivy.where(index >= x.shape[-1], index % x.shape[-1], index)
     return ivy.gather(x, index, axis=0)
--- a/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
+++ b/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
@@ -1932,7 +1932,8 @@ def test_paddle_kron(
 @handle_frontend_test(
     fn_tree="paddle.take",
     dtype_and_x=helpers.array_indices_axis(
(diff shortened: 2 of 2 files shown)


ivy/functional/frontends/paddle/tensor/math.py
```diff
@@ -428,3 +428,17 @@ def diff(x, n=1, axis=-1, prepend=None, append=None, name=None):
...
+
+
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float32", "float64", "int32", "int6")}, "paddle"
+)
+@to_ivy_arrays_and_back
+def take(
+    x,
+    index,
+    mode=None,
```
jshepherd01, line 440: Since `mode` influences the maths of the function, I think it should be implemented here

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
```diff
@@ -1926,3 +1926,32 @@ def test_paddle_kron(
         x=x[0],
         y=x[1],
     )
+
+
+# take
+@handle_frontend_test(
+    fn_tree="paddle.take",
+    dtype_and_x=helpers.array_indices_axis(
+        array_dtypes=helpers.get_dtypes(), indices_dtypes=["int32", "int64"]
```
jshepherd01, line 1935: Probably best to explicitly set something inside `get_dtypes()`, even if its just `"valid"`, since its behaviour with no `draw` argument isn't defined in its docstring. Based on the unsupported dtypes on the function, `"float_and_integer"` might be a good fit.

jshepherd01 (changes requested):

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
Chosenundead15: corrected on new commit.

ivy/functional/frontends/paddle/tensor/math.py
Chosenundead15: added mode functionality and checks to make sure the mode is using is a correct one (good chance someone made a type writing the value of string, I have done that mysefl a lot). The only catch, is that appears that the `helpers.test_frontend_function` might be ignoring the mode use out of the box, I am currently uncertain if I have to fix it on my side or it falls on the helper function.

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
```diff
@@ -1926,3 +1926,33 @@ def test_paddle_kron(
...
+
+
+# take
+@handle_frontend_test(
+    fn_tree="paddle.take",
+    dtype_and_x=helpers.array_indices_axis(
+        array_dtypes=helpers.get_dtypes("float_and_integer"),
+        indices_dtypes=["int32", "int64"],
+    ),
+)
```
jshepherd01, line 1938: I think this should also generate a `mode` so you can test the clip and wrap options

jshepherd01 (changes requested):
Looks pretty good, just one thing on the tests

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
```diff
@@ -1926,3 +1926,44 @@ def test_paddle_kron(
...
+        )
+    )
+
+    return dtypes, xs, indices, mode
+
+
+# take
+@handle_frontend_test(
+    fn_tree="paddle.take", dtype_and_values=_test_paddle_take_helper()
+)
```
jshepherd01, line 1948: I think the `array_indices_axis` helper only generates indices that fit within the array, so the `mode` won't be used anyway. You could try generating `indices` as a separate list of integers, and clipping the values to the length of `xs` if `mode=="raise"`

I didn't explain it very well, what I mean is something like
```py
mode = draw(st.sampled_from(["raise", "clip", "wrap"]))

dtype, x = draw(helpers.dtype_and_values(...))

length = len(x[0].reshape((-1,)))
indices = st.lists(st.integers(
    min_value=(-length if mode=="raise" else None),
    max_value=(length - 1 if mode=="raise" else None)
))
```
That way if the mode is clip or wrap it will generate indices that need to be clipped or wrapped, but if mode is raise then it won't

jshepherd01 (changes requested):
That's pretty good but I think there's still one small bug

Chosenundead15:
There was an actual simple way to make the test use out of bounds indeces, `array_ndices_axis` had a flag to do that, Went to toy with it for a bit and print what actual values was using (also went to check and printed the resulting values of the operation) and appears to be working as intended. Inverting the flag to do it on `raise` results on error from out of bounds index, so i am extra sure from that.

jshepherd01 (approved):

PR #21514, opened 2023-08-09
Added project.toml

sladyn98:
Convert setup.py to project.toml


sladyn98:
@vedpatwardhan @KareemMAX I added this as a draft start do let me know how it looks

vedpatwardhan:
I think the `pyproject.toml` should be in the root project folder rather than the `ivy_tests`. I'd also suggest checking if the installation still continues to work with editable mode if we've removed the `setup.py` (i.e. using the `-e`) as it seemed as though having a `setup.py` was also important for using a package in editable mode (i.e. with the symbolic link from the local folder to `site-packages` to reflect changes in the local folder in the `site-packages` when we try to `pip install` from a local directory which is done with the `-e` flag). 
Happy to know what you think @KareemMAX, thanks @sladyn98 😄

KareemMAX:
Not much to add. There's still missing areas that are in `setup.py` which are not implemented yet, after they are implemented we can proceed

sladyn98:
I added the dependencies, could you let me know how to run it  ? I dont have a lot of experience with toml files.

vedpatwardhan:
@KareemMAX would have a better idea of the process, thanks!

project.toml
```diff
@@ -3,4 +3,32 @@ requires = [
     "setuptools>=42",
     "wheel"
 ]
-build-backend = "setuptools.build_meta"
\ No newline at end of file
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "ivy"
+version = "{__version__}"
```
KareemMAX, line 10: We import this value from `ivy/_version.py`. Does that heppen?

Ref: https://github.com/unifyai/ivy/blob/master/setup.py#L53

project.toml
```diff
@@ -3,4 +3,32 @@ requires = [
...
+description = "The unified machine learning framework, enabling framework-agnostic functions, layers and libraries."
+readme = "README.rst"
```
KareemMAX, line 12: Readme is now `README.md`, we also run custom code on the raw document to remove dark mode blocks 

Ref: https://github.com/unifyai/ivy/blob/master/setup.py#L28-L45

project.toml
```diff
@@ -3,4 +3,32 @@ requires = [
...
+authors = [{ name = "Unify", email = "hello@unify.ai" }]
+license = "Apache-2.0"
```
KareemMAX, line 14: Is there is a need for the dash? In `setup.py` it uses `Apache 2.0`

Ref: https://github.com/unifyai/ivy/blob/master/setup.py#L74

project.toml
```diff
@@ -3,4 +3,32 @@ requires = [
...
+psutil = "5.9.1"
+termcolor = "1.1.0"
+colorama = "0.4.5"
+packaging = "21.3"
+"nvidia-ml-py" = "<=11.495.46" 
+diskcache = "*"
+google-auth = "*"
+requests = "*"
+pyvis = "*"
+dill = "*"
```
KareemMAX, line 34: This way duplicate our code a bit, can we make it look like the original `setup.py`?

Ref: https://github.com/unifyai/ivy/blob/master/setup.py#L71

KareemMAX (changes requested):
The general structure looks good to me (although I'm not that deep into how `pyproject.toml` work). I've addressed some issues and then we can start testing.

My idea of testing things would be to use `pip`. Maybe we can test in 2 phases, locally using `pip install .` and then we can upload it to PyPI test registry

KareemMAX:
Also one small addition. Seems like one of our volunteers @NripeshN has made some duplicate efforts by mistake 😅. https://github.com/unifyai/ivy/pull/21577

As I see you both looking into this maybe you can join forces @sladyn98 @NripeshN? Feel free to contact privately and use each others code. Maybe decide to close one of those PRs and keep the other

KareemMAX (changes requested):
Hey @NripeshN,

Overall everything seem good to me, but you still having some TODOs that need to be resolved. After that we might start testing the behavior of it against PyPI test environment maybe.

PR #21539, opened 2023-08-09
Complex maths functions

jshepherd01:
Add complex number support to tanh (as an activation function as well as just a mathematical one, so it now has `@handle_complex_input` and the like) and exp. Modify pow for better numerical stability, and a slight change to gelu to resolve some type issues. Now also adds complex number support to all trig functions, including inverse and hyperbolic, in the API, as well as expanding test coverage to also check complex values.


ivy-leaves:

Thank you :hugs:

jshepherd01:
There remains a bug with `tan` on the jax backend, because for some reason jax thinks `tan(-1-45i) = 0` (it's actually `-i`)

ivy/functional/backends/paddle/activations.py
```diff
@@ -68,18 +68,13 @@ def gelu(
     approximate: bool = False,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
+    if x.dtype in [paddle.complex64, paddle.complex128]:
+        return (
+            0.5
+            * x
+            * (1 + paddle_backend.tanh(0.7978845608 * (x + 0.044715 * x * x * x)))
```
vedpatwardhan, line 75: I know you haven't added this and are just moving it, but ideally we shouldn't use any "magic" numbers in the implementation, probably we can define them as constants to indicate what they signify 😅

ivy/stateful/activations.py
```diff
@@ -275,24 +275,34 @@ def _forward(self, x):
...
+        Parameters
+        ----------
+        complex_mode
+            Specifies how to handle complex input.
+        """
+        self._complex_mode = complex_mode
         Module.__init__(self)
 
-    def _forward(self, x):
+    def _forward(self, x, complex_mode=None):
```
vedpatwardhan, line 290: let's remove the `complex_mode` parameter from `_forward` as we shouldn't need to pass a separate value for `complex_mode` for each forward pass. [This](https://github.com/unifyai/ivy/pull/21902) PR fixes it for the rest of the stateful classes

ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py
```diff
@@ -1489,7 +1521,9 @@ def test_tan(*, dtype_and_x, test_flags, backend_fw, fn_name, on_device):
 # tanh
 @handle_test(
     fn_tree="functional.ivy.tanh",
-    dtype_and_x=helpers.dtype_and_values(available_dtypes=helpers.get_dtypes("float")),
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float_and_complex")
+    ),
```
vedpatwardhan, line 1526: we should also generate the `complex_mode` argument in the tests

vedpatwardhan (changes requested):
I haven't gone through the implementation in a lot of detail, but just made a few suggestions. Let's see what @Ishticode thinks, I'll also request @MahmoudAshraf97's review for the changes to the `paddle` backend. Thanks for diving into this @jshepherd01 😄

ivy/functional/backends/paddle/activations.py
jshepherd01: Unfortunately they don't really signify anything, they're just the numbers that make the approximation work well. I've put a comment in giving the source of them though

ivy/stateful/activations.py
jshepherd01: Wasn't aware of that PR but that does make a lot more sense

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -921,11 +979,14 @@ def tan(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.T
     ]:
         ret_dtype = x.dtype
         return paddle.tan(x.astype("float32")).astype(ret_dtype)
+    if x.dtype in [paddle.complex64, paddle.complex128]:
```
MahmoudAshraf97, line 982:
```suggestion
    if paddle.is_complex(x):
```

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -683,6 +733,12 @@ def sinh(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.
     ]:
         ret_dtype = x.dtype
         return paddle.sinh(x.astype("float32")).astype(ret_dtype)
+    if x.dtype in [paddle.complex64, paddle.complex128]:
```
MahmoudAshraf97, line 736:
```suggestion
    if paddle.is_complex(x):
```

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -701,6 +757,11 @@ def square(
 ) -> paddle.Tensor:
     if x.dtype in [paddle.int32, paddle.int64, paddle.float32, paddle.float64]:
         return paddle.square(x)
+    if x.dtype in [paddle.complex64, paddle.complex128]:
```
MahmoudAshraf97, line 760:
```suggestion
    if paddle.is_complex(x):
```

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -976,6 +1040,11 @@ def log(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.T
 def exp(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
     if x.dtype in [paddle.int32, paddle.int64, paddle.float32, paddle.float64]:
         return paddle.exp(x)
+    if x.dtype in [paddle.complex64, paddle.complex128]:
+        return paddle.multiply(
+            paddle.exp(x.real()),
+            paddle.complex(paddle.cos(x.imag()), paddle.sin(x.imag())),
+        )
     return pow(math.e, x).astype(x.dtype)
```
MahmoudAshraf97, line 1048: I guess while we are at it
```suggestion
    return paddle_backend.pow(math.e, x).astype(x.dtype)
```

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -976,6 +1040,11 @@ def log(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.T
 def exp(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
     if x.dtype in [paddle.int32, paddle.int64, paddle.float32, paddle.float64]:
         return paddle.exp(x)
+    if x.dtype in [paddle.complex64, paddle.complex128]:
```
MahmoudAshraf97, line 1043:
```suggestion
    if paddle.is_complex(x):
```

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -1049,6 +1118,8 @@ def atanh(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle
     ]:
         ret_dtype = x.dtype
         return paddle.atanh(x.astype("float32")).astype(ret_dtype)
+    if x.dtype in [paddle.complex64, paddle.complex128]:
```
MahmoudAshraf97, line 1121:
```suggestion
    if paddle.is_complex(x):
```

MahmoudAshraf97 (changes requested):

ivy/data_classes/array/elementwise.py
```diff
@@ -2566,6 +2571,8 @@ def tanh(self: ivy.Array, *, out: Optional[ivy.Array] = None) -> ivy.Array:
         out
             optional output, for writing the result to. It must have a shape that the
             inputs broadcast to.
+        complex_mode
+            optional specifier for how to handle complex data types.
```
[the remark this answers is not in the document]
jshepherd01: There's an explanation in the base function and I was trying to keep the array method docstrings as minimal as possible, I can definitely add it if that's preferred though

ivy/functional/backends/paddle/activations.py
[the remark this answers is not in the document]
jshepherd01: The other case relies on `erf`, which none of the backends implement for complex numbers (the only implementation I was able to find anywhere was in scipy, and it's a very long and complicated function that falls back on taylor series expansions for different cases). There's also a wrapper on the ivy API level which sets approximate to True if the input is complex, so the other case should never be reachable with a complex input

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -717,20 +778,17 @@ def pow(
...
-            return result
         return paddle.pow(x1.astype("float32"), x2.astype("float32")).astype(ret_dtype)
+    if x1.dtype in [paddle.complex64, paddle.complex128]:
+        # https://math.stackexchange.com/questions/476968/complex-power-of-a-complex-number
+        r = paddle.abs(x1)
+        theta = paddle.angle(x1)
+        res_mag = paddle.pow(r, x2.real()) / paddle.exp(x2.imag() * theta)
+        res_ang = paddle.log(r) * x2.imag() + theta * x2.real()
+        result = res_mag * paddle.complex(paddle.cos(res_ang), paddle.sin(res_ang))
+        return result.astype(ret_dtype)
```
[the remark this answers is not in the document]
jshepherd01: The original was having issues with numerical stability because of the `exp(log(r))`, so I refactored it to remove that part. I also unindented it to bring it in line with how other functions check for complex numbers

ivy/functional/backends/paddle/elementwise.py
jshepherd01: I don't see why this is needed, since the case of complex `x` is already handled (and handled more efficiently than using `paddle_backend.pow(math.e, x)` since we use the special case of the base being real)

MahmoudAshraf97: this change isn't concerned with complex input, `pow` here refers to the paddle backend implementation of `pow` but if we use `pow` instead of `paddle_backend.pow` we are actually calling `ivy.pow` and hence passing through all the wrappers that we don't need

jshepherd01: I thought since `pow` is defined in this file, `pow` and `paddle_backend.pow` are the same function? Done either way, since its more clear with `paddle_backed.` if nothing else

ivy/data_classes/array/elementwise.py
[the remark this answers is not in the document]
vedpatwardhan: Yeah I agree with @MahmoudAshraf97, the parameter descriptions should ideally be the same as the functional one

ivy/func_wrapper.py
```diff
@@ -1486,16 +1486,21 @@ def _handle_complex_input(
         jax_like = fn.jax_like if hasattr(fn, "jax_like") else "entire"
```
vedpatwardhan, line 1486: a bit silly, but do we rename `handle_complex_input` to `handle_complex_inputs` 😅 ?

ivy/functional/backends/jax/elementwise.py
```diff
@@ -304,6 +304,10 @@ def log10(x: JaxArray, /, *, out: Optional[JaxArray] = None) -> JaxArray:
 
 
 def log1p(x: JaxArray, /, *, out: Optional[JaxArray] = None) -> JaxArray:
+    if x.dtype in ["complex64", "complex128"]:
+        # For certain complex numbers, jnp.log1p(z) != jnp.log(1+z)
```
vedpatwardhan, line 308: That's a bit strange, were there any particular category of examples which were leading to these issues (i.e. very small values, very large values, with a negative imaginary part, etc.)? Would be great to identify those and add a comment accordingly in the implementation and the tests

ivy/functional/ivy/elementwise.py
```diff
@@ -6369,11 +6370,13 @@ def tan(
 @to_native_arrays_and_back
 @handle_array_function
 @handle_device_shifting
+@handle_complex_input
 def tanh(
     x: Union[ivy.Array, ivy.NativeArray],
     /,
     *,
     out: Optional[ivy.Array] = None,
+    complex_mode: Literal["split", "magnitude", "jax"] = "jax",
```
vedpatwardhan, line 6379: Probably some of the previous changes might've been accidentally reverted, as there doesn't seem to be a `complex_mode` argument for `ivy.tanh` in the backends 😅

ivy/functional/ivy/elementwise.py
```diff
@@ -6445,6 +6448,9 @@ def tanh(
     out
         optional output, for writing the result to. It must have a shape that the inputs
         broadcast to.
+    complex_mode
+        optional specifier for how to handle complex data types. See
+        `ivy.func_wrapper.handle_complex_input` for more detail.
```
vedpatwardhan, line 6453: Would be great to enclose this with double back-ticks than single ones 😄

ivy/stateful/activations.py
```diff
@@ -91,7 +84,7 @@ def _forward(self, x, complex_mode=None):
         ret
             The outputs following the RELU activation *[batch_shape, d]*
         """
-        return ivy.relu(x, complex_mode=ivy.default(complex_mode, self._complex_mode))
+        return ivy.relu(x, self._complex_mode)
```
vedpatwardhan, line 87: just a minor change, but `complex_mode` is a keyword-only argument in `ivy.relu`

ivy/stateful/activations.py
```diff
@@ -292,7 +289,7 @@ def _forward(self, x):
          ret
             The outputs following the TANH activation *[batch_shape, d]*
         """
-        return ivy.tanh(x)
+        return ivy.tanh(x, self._complex_mode)
```
vedpatwardhan, line 292: same for `tanh` (keyword-only argument)

ivy_tests/test_ivy/test_frontends/test_jax/test_nn/test_non_linear_activations.py
```diff
@@ -191,7 +189,7 @@ def test_jax_gelu(
 ):
     input_dtype, x = dtype_and_x
     # As erf function doesn't support complex dtype
-    if "complex" in str(x[0].dtype):
+    if "complex" in str(input_dtype):
```
vedpatwardhan, line 192: worth confirming, but probably `input_dtype` is a `list[str]` rather than a `str`

vedpatwardhan (changes requested):
Just a few minor things here and there, thanks @jshepherd01 😄

ivy/data_classes/array/elementwise.py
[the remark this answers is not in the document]
jshepherd01: Fair enough. Would it be enough to add `See ``ivy.func_wrapper.handle_complex_input`` for more detail.` or should I put a description in directly?

ivy/func_wrapper.py
jshepherd01: It's only used for (and is only designed for) activation functions so I think the singular works better

ivy/functional/backends/jax/elementwise.py
jshepherd01: It is, but in the time since I made that comment I discovered that jax's complex number maths is buggy in general. It seems to cause problems when the values are too big or too small, but not anywhere near breaking floating point maths or anything, just stuff like stuff like `log1p(-1.-0.00110054j)` or `arccos(-35246690.-2.j)` give the wrong answers. Not sure what the best way to deal with it is since I'm pretty sure it's a bug on their side

ivy/functional/ivy/elementwise.py
jshepherd01: `complex_mode` is handled by a function wrapper rather than by the backends themselves

ivy_tests/test_ivy/test_frontends/test_jax/test_nn/test_non_linear_activations.py
jshepherd01: Yeah, you're right. `str(input_dtype)` does still work since it returns `"['complex128']"` (for example) but I guess it's more clear to do `input_dtype[0]` directly

ivy/data_classes/array/elementwise.py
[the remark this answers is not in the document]
vedpatwardhan: Yeah it's completely fine to add "See ``ivy.func_wrapper.handle_complex_input`` for more detail.". The only thing is the parameter descriptions for the `ivy.tanh` function should be the same as the parameter descriptions for the `ivy.Array.tanh` method 😄

ivy/func_wrapper.py
vedpatwardhan: yeah you're right, `handle_complex_input` sounds better for it's purpose

ivy/functional/backends/jax/elementwise.py
vedpatwardhan: got it! ig it would be great if you could create an issue on the jax repo pointing out to some of these bugs and then add that link to this comment, just for reference. Thanks!

ivy/functional/ivy/elementwise.py
vedpatwardhan: yep sure, but ideally the `ivy.tanh` function and the backend implementations should have the same function signature (not typehints of course, but the arguments to the function)

ivy_tests/test_ivy/test_frontends/test_jax/test_nn/test_non_linear_activations.py
vedpatwardhan: sounds good, thanks!

ivy/functional/ivy/elementwise.py
jshepherd01: Yep, you're right, I'll add that in

ivy/functional/backends/jax/elementwise.py
vedpatwardhan: thanks!

ivy/functional/backends/jax/activations.py
```diff
@@ -18,17 +18,25 @@ def gelu(
...
 
 
 def leaky_relu(
-    x: JaxArray, /, *, alpha: float = 0.2, out: Optional[JaxArray] = None
+    x: JaxArray,
+    /,
+    *,
+    alpha: float = 0.2,
+    out: Optional[JaxArray] = None,
+    complex_mode="jax",
```
vedpatwardhan, line 32: A minor thing, but `complex_mode` should come before `out`, ideally `out` should be the last argument of the function. Same applies to the other functions and backends

ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py
```diff
@@ -57,10 +57,11 @@ def test_abs(*, dtype_and_x, test_flags, backend_fw, fn_name, on_device):
 @handle_test(
     fn_tree="functional.ivy.acosh",
     dtype_and_x=helpers.dtype_and_values(
-        available_dtypes=helpers.get_dtypes("float"),
+        available_dtypes=helpers.get_dtypes("float_and_complex"),
         min_value=1,
-        large_abs_safety_factor=4,
-        small_abs_safety_factor=4,
+        large_abs_safety_factor=2.1,
```
vedpatwardhan, line 62: You've tested with a relatively large number of examples before reducing the safety factors right? Sometimes hypothesis isn't able to find a failing example until we run with like `100` max_examples

vedpatwardhan (changes requested):
Almost there, just a final couple of changes, thanks @jshepherd01 😄

ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py
jshepherd01: I also set the scale to `"log"` so the safety factors are actually bigger now (also tested just now with 100 examples per backend and everything passes except the gradient thing)

vedpatwardhan: perfect, thanks!

PR #21552, opened 2023-08-09
Add linalg.tensor_diag_part to tensorflow frontends

### LINKED ISSUE #21288: tensor_diag_part

awath-abdat:
Close #21288 Implemented the linalg tensor_diag_part which calculated the diagonal of the input tensor as described in the tensorflow docs here https://www.tensorflow.org/api_docs/python/tf/linalg/tensor_diag_part, I validated the array's shape then reshape it to a 2D array with the diagonal elements needed in place and then get its diagonal then reshape this diagonal back to the desired output shape.


hmahmood24:
Hey @awath-abdat, thanks for the PR. Everything looks good to me, Thanks for the contribution! :) 🚀

PR #21570, opened 2023-08-10
Refactor frontend testing config

CatB1t:
- Added `SupportedDtypes` and `SupportedDevices` dataclasses, currently they're only a place holder instead of multiple attributes, for the future reference it should compute all parameters from a single `valid_dtypes` and `valid_devices` lists respectively.
- Added `FrontendConfig` and `FrontendConfigWithBackend` classes, the first holds the same legacy information as before but will throw an exception if one is missing, the later is used for frontends that have a backend implemented, to reduce code redundancy.
- Refactored `numpy`, `tensorflow`, `torch`, `jax` and `paddle` configs to use `FrontendConfigWithBackend`
- Refactored `mxnet`, `onnx`, `scipy` and `pandas` frontend configs to use `FrontendConfig`


sherry30 (approved):
lgtm

PR #21573, opened 2023-08-10
feat: Dunder method patching to prioritise dunder methods of ivy.Array over native tensors

yopknopixx:
This PR is related to the Trello task: [https://trello.com/c/sFUmwaFV/1614-investigate-the-behaviour-of-dunder-methods-in-ivy-array-with-native-arrays](url)
___
### Task Description:
All Python operators, like +, ==, and in, rely on [Dunder methods](https://mathspp.com/blog/pydonts/dunder-methods) to implement their behavior. Most of these operations are implemented in both Native Tensor classes and in Ivy Array class, when a python operator is invoked, the dunder method of the left operand is called and if it’s not implemented then the method of the right operator is called. In any case where the method of the left operator doesn’t support the right operator (for example paddle methods only support paddle tensors or numpy arrays) or has a different behavior (for example a  *b !=  b*  a, assuming its a commutative operation and a is a native tensor and b is an ivy array that has slightly different logic for this operator). This renders all the right side dunder methods of ivy array useless in case it was used with another framework that has the left side method implemented regardless of whether it supports ivy arrays or not.

You should try to find a way to dynamically override dunder methods of native tensors to return NotImplemented so that the ivy dunder method is used instead


ivy/functional/backends/torch/__init__.py
```diff
@@ -16,6 +16,77 @@
 
 use = ivy.utils.backend.ContextManager(_module_in_memory)
 
+
```
vedpatwardhan, line 19: let's add a comment explaining the reason for doing this. Also, just to confirm the reverse operators work with backends other than `torch` right? Thanks for diving into this @yopknopixx 😄

vedpatwardhan (changes requested):

ivy/functional/backends/torch/__init__.py
yopknopixx: No problem @vedpatwardhan! I've added a short comment explaining what we're doing and why. Yes, it does indeed work with other frameworks. Torch and TF are done, I'll continue with Jax MXNet Numpy and Paddle shortly.

ivy-leaves:

Thank you :hugs:

ivy-leaves:

Thank you :hugs:

PR #21597, opened 2023-08-10
Cummax

### LINKED ISSUE #21215: cummax

natek-1:
close #21215


natek-1:
@danielmunioz  it seems like some of the test that are running are not on the file that were changed. Could you help me out with those.
Also I am not sure how to solve the issues:
'AssertionError: returned dtype = int64, ground-truth returned dtype = uint8'
seems a dtype mismatch but I am not sure why it occurs

danielmunioz:
Sure I'll take a look at that maybe something to do with the backends, could you please fix the merge issues please?

Also it seems that the Tensor implementation of [cummax](https://pytorch.org/docs/stable/generated/torch.Tensor.cummax.html) does not include the `out` argument, only `dim`, could you please remove it so that it looks like the original?

Issue #21187, opened 2023-08-02
ivy.deform_conv2d


themaigod:
**Why should this be implemented?**
As discussion in #20138, deformable convolution is a popular and common operation. Here is the reason why I think it should be added as an experimental api:
- 3+ of the native frameworks have this function (Pytorch (in torchvision), mindspore, paddlepaddle)
- it's an NN-related function and should be as efficient as possible (official implementations in native frameworks have advantage in speed)
- it's needed for a complex/long frontend function implementation

**Links to native framework implementations**
- [PyTorch (in torchvision)](https://pytorch.org/vision/stable/generated/torchvision.ops.deform_conv2d.html#torchvision.ops.deform_conv2d)
- [Paddlepaddle:paddle.vision.ops.deform_conv2d](https://www.paddlepaddle.org.cn/documentation/docs/zh/api/paddle/vision/ops/deform_conv2d_cn.html#paddle.vision.ops.deform_conv2d) and [paddle.static.nn.deform_conv2d](https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/static/nn/deform_conv2d_en.html). [Notice that deformable_conv is old API and connot be used.](https://www.paddlepaddle.org.cn/documentation/docs/zh/api/paddle/fluid/layers/deformable_conv_cn.html#paddle.fluid.layers.deformable_conv)
- [Mindspore](https://www.mindspore.cn/docs/en/r1.9/api_python/ops/mindspore.ops.deformable_conv2d.html#mindspore.ops.deformable_conv2d)
- [MXNet v2](https://mxnet.apache.org/versions/1.9.1/api/python/docs/api/ndarray/contrib/index.html?highlight=deform#mxnet.ndarray.contrib.ModulatedDeformableConvolution) [MXNet v1](https://mxnet.apache.org/versions/1.9.1/api/python/docs/api/ndarray/contrib/index.html?highlight=deform#mxnet.ndarray.contrib.DeformableConvolution)
- Though Tensorflow does not have official implementation, you can check multiple related PRs for that. It seems that the implementations were not merged due to some outer issues, such as feature freeze. [PR1](https://github.com/tensorflow/addons/pull/2196) [PR2](https://github.com/tensorflow/addons/pull/1129) [PR Sggestion](https://github.com/keras-team/keras-cv/issues/1140)

**Link to Third Party Implementations only by Python which are complex and low efficiency**
-[PyTorch](https://github.com/oeway/pytorch-deform-conv) 
[...]

PR #21729, opened 2023-08-11
Add experimental functional API: ivy.deform_conv2d

themaigod:
solve #21187

### DIFF
--- a/ivy/data_classes/array/experimental/layers.py
+++ b/ivy/data_classes/array/experimental/layers.py
@@ -1172,5 +1172,14 @@ class _ArrayWithLayersExperimental(abc.ABC):
         >>> print(result.shape)
         (4, 5, 8, 8)
         """
-        return ivy.deform_conv2d(self._data, offset, weight, bias=bias, stride=stride,
-                                 padding=padding, dilation=dilation, mask=mask, out=out)
+        return ivy.deform_conv2d(
+            self._data,
+            offset,
+            weight,
+            bias=bias,
+            stride=stride,
+            padding=padding,
+            dilation=dilation,
+            mask=mask,
+            out=out,
+        )
--- a/ivy/data_classes/array/experimental/layers.py
+++ b/ivy/data_classes/array/experimental/layers.py
@@ -1107,7 +1107,6 @@ class _ArrayWithLayersExperimental(abc.ABC):
         """
         return ivy.rfftn(self._data, s=s, axes=axes, norm=norm, out=out)
 
-
     def deform_conv2d(
         self: ivy.Array,
         offset: Union[ivy.Array, ivy.NativeArray],
--- a/ivy/data_classes/array/experimental/layers.py
+++ b/ivy/data_classes/array/experimental/layers.py
@@ -1106,3 +1106,72 @@ class _ArrayWithLayersExperimental(abc.ABC):
             The result of the RFFT operation.
         """
         return ivy.rfftn(self._data, s=s, axes=axes, norm=norm, out=out)
+
+
+    def deform_conv2d(
+        self: ivy.Array,
+        offset: Union[ivy.Array, ivy.NativeArray],
+        weight: Union[ivy.Array, ivy.NativeArray],
+        *,
+        bias: Optional[Union[ivy.Array, ivy.NativeArray]] = None,
+        stride: Union[int, Tuple[int, int]] = (1, 1),
+        padding: Union[int, Tuple[int, int]] = (0, 0),
+        dilation: Union[int, Tuple[int, int]] = (1, 1),
+        mask: Optional[Union[ivy.Array, ivy.NativeArray]] = None,
+        out: Optional[ivy.Array] = None,
+    ) -> ivy.Array:
+        """
+        Perform Deformable Convolution v2, if mask is not None, and Perform Deformable
+        Convolution, if mask is None. Deformable convolution, as described in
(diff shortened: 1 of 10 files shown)


ivy-leaves:

Thank you :hugs:

themaigod:
The format checking seems that it looks at which lines are not my edited (from 805 to 915).

themaigod:
The tests writing have some bugs shown in automated test result. I will do further modification soon.

themaigod:
Fixed the problem in test.

themaigod:
In this implementation, I am required to use the function from torchvision in pytorch backend.But standard ivy import is not included this. Can we import this, or I should delete this implementation in pytorch backend.

themaigod:
Hi, @danielmunioz can you review it? Or someone can do this. And I have a question about importing torchvision. Really appreciate it. Thanks.

danielmunioz:
Hey @themaigod! Thanks for contributing!

Yeah, the implementation looks good so far, just one thing, we are trying no to use extra packages in the backends and I noticed you are using `torchvision` on the torch backend, do you think it'd be possible to have it implemented in pure `pytorch`? I guess nailing that down would suffice to then implement the same algorithm on the other missing backends

themaigod:
> Hey @themaigod! Thanks for contributing!
> 
> Yeah, the implementation looks good so far, just one thing, we are trying no to use extra packages in the backends and I noticed you are using `torchvision` on the torch backend, do you think it'd be possible to have it implemented in pure `pytorch`? I guess nailing that down would suffice to then implement the same algorithm on the other missing backends

Hi, @danielmunioz .

I also check some personal implementations which use pure framerwork without additional C/C++ implementation. The readme and issuse for these respositories mention a heavy effeciency issue for this kind of implementation. Here is an [example](https://github.com/kastnerkyle/deform-conv/). It noticed that "This implementation is not efficient. In fact a forward pass with deformed convolution takes 260 ms, while regular convolution takes only 10 ms. Also, GPU average utilization is only around 10%." Maybe it is caused by too many reshape operations, For the author code of these two papers, they also release a C/C++ implementations for MXNet. Here is the [respository](https://github.com/msracver/Deformable-ConvNets).

So, I think it it is good if we implement it in pure framework for those do not have official implementation. As for Pytorch, it is better to use the implementation in torchvision which is a C/C++ implementation. People working on CV require to use this function effeciently. 

By the way, it is a little bit wired for me to import only torch, but not torch, torchvision and maybe torchaudio. I agree with the importance of no to use extra packages in the backends. However, torchvision is a part of pytorch, and people are familiar to work with it and pytorch together. Torchvision supports a lot of widely used operators in CV and always downloaded with torch as pytorch official web page recommended. 

I can update a version implemented like this [example](https://github.com/kastnerkyle/deform-conv/) for pytorch backend or directly by ivy.functional.API if you think it is helpful. I still hope we can have torchvision in the pytorch backend.

themaigod:
A small question for writing it in ivy.functional.API:
Can I use "try-except" structure, like:
```
try:
    return ivy.current_backend().deform_conv2d(
        x,
        offset,
        weight,
        bias=bias,
        stride=stride,
        padding=padding,
        dilation=dilation,
        mask=mask,
        out=out,
    )
except IvyNotImplementedException:
    # ivy.functional.API implementation
```
There are some doubts using "try-except". So I am not sure it is acceptable. Or we have other methods to avoid this?

danielmunioz:
Hey @themaigod! Yeah, for the regular backends a full pythonic implementation is OK, again we don't really need to force the users to have to import all extra dependencies when using a backend, if we'd do this we'd end up importing much more than just `torchvision` as there have been other cases where an adjacent library provides a better out of the box implementation, in these cases however the solution we have is to, again, have the full native implementation on the backend and after that we can add the more efficient one as a sub-backend for when eager performance is of great concern.

For the purposes of this task however we need to get the regular functionality implemented across all backend without any extra dependency!

And as of using `try-except` in a block of code like that, we typically don't need too because we are looking to have all implementations fully functioning across all backends every time we add a new function!

Hope that makes sense! And please feel free to ask any questions! :)

themaigod:
Hi, @danielmunioz. I am not fully understanding the part of "try-except" structure. 

Here is my consideration: If it has official implementation in backend framework, we just call the backend function, which is efficiency. If not, we can treat it as a mixed function. So it will be implemented by ivy functional API in (ivy/functional/ivy/experimental/layers.py) instead of implementing it in each backend file (i.e. ivy/functional/backends/jax/experimental/layers.py) which does not have official implementation. In these backend files, they will raise `IvyNotImplementedException`. That's why it requires a "try-except" structure.

In my understanding of your words, we need to implement it in each backend file by using pure framework for those which do not have official implementation. This kind of implementation follows the same idea.

I think this way is too heavy. By using ivy functional API, we can only implement it once.

Can you further clarify it?

danielmunioz:
Hey @themaigod!
Sorry for the delay it has been a couple of hectic weeks over here!

Answering the questions yes, doing a pure `ivy` compositional implementation is an option, however as you mentioned we'd be missing some of the pure backend implementations that are already optimized, I know it seems like a lot of work (and it is) but if you take a look at the other `experimental` functions, you'll see that lot's of those follow the same structure of implementing the full function on every backend while following a superset, we cannot just use a `try-except` structure because the objective of adding a new function to ivy is to fully support it across all backends, so for this case the option that we should follow is to fully implement the algorithm natively across all backends, using the supported functions when needed and without adding new dependencies! :)

Please don't hesitate to ask any questions! 
Thanks!

PR #21736, opened 2023-08-12
Aml 21114 jax frontend nanquantile

### LINKED ISSUE #21114: nanquantile

Aml-Hassan-Abd-El-hamid:
close  #21114


ivy/functional/frontends/jax/numpy/statistical.py
```diff
@@ -512,3 +514,29 @@ def quantile(
...
+):
+    if interpolation:
+        warnings.warn(
++            "interpolation is deprecated name for the method keyword argument since"
++            " version 1.22.0 in JAX"
++        )
++       method = interpolation  
+    return ivy.nanquantile(
+        a, q, axis=axis, keepdims=keepdims, interpolation=method, out=out
+    )
```
MahmoudAshraf97, line 542: this function was removed from the functional API just a while ago, so either implement it again or use a compositional implementation, perhaps a combination of `where` and `quantile`

MahmoudAshraf97 (changes requested):

ivy/functional/frontends/jax/numpy/statistical.py
Aml-Hassan-Abd-El-hamid: Hi @MahmoudAshraf97 
Thank you very much for your helpful review.
I thought I could use the nanquantile function from the ivy/experimental
https://github.com/geeythree/ivy/blob/9b8bb5c78b14070e7ee168de3a8c533fae3f8655/ivy/functional/ivy/experimental/statistical.py#L498
If that can't work, please let me know.

MahmoudAshraf97: as I mentioned, the function was removed, you are right, you could've used it if it wasn't removed so this is not your fault

ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_statistical.py
MahmoudAshraf97, line 1: please test the `overwrite_input` and `interpolation` arguments to testing

ivy/functional/frontends/jax/numpy/statistical.py
```diff
@@ -514,3 +516,29 @@ def quantile(
...
+    keepdims=False,
+    interpolation=None,
+):
+    if interpolation:
+        method = interpolation
+        warnings.warn(
+            "interpolation is deprecated name for the method keyword argument since"
+            " version 1.22.0 in JAX"
+        ) 
+    return ivy.nanquantile(
```
MahmoudAshraf97, line 542: don't forget to change this function, maybe you can use regular quantile after filtering the `NaN` values as a compositional implementation

MahmoudAshraf97 (changes requested):

ivy/functional/frontends/jax/numpy/statistical.py
Aml-Hassan-Abd-El-hamid: Hi @MahmoudAshraf97 
Thank you very much for clarifying that. 

I have one last question, Is there an IVY function that can filter Nans or should I just use Numpy?
If I used NumPy, it would probably look something like this: ```a = a[~numpy.isnan(a)]```

ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_statistical.py
Aml-Hassan-Abd-El-hamid: I added them, but I don't think that they are really that effective because overwrite_input isn't used in the function body at all -I didn't find its use in the docs so I couldn't add it- and interpolation isn't supposed to be used as it's deprecated according to Jax's docs, I set it to None so it doesn't rais a warning.

ivy/functional/frontends/jax/numpy/statistical.py
MahmoudAshraf97: that sounds plausible to me, try it and see if the tests pass

Issue #21801, opened 2023-08-13
log1p_


he11owthere:
To-do list: https://github.com/unifyai/ivy/issues/3612

PR #21802, opened 2023-08-13
Added log1p_ method in the PyTorch frontend

he11owthere:
closes https://github.com/unifyai/ivy/issues/21801

image

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py
+++ b/ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py
@@ -10604,6 +10604,40 @@ def test_torch_tensor_log1p(
     )
 
 
+# log1p_
+@handle_frontend_method(
+    class_tree=CLASS_TREE,
+    init_tree="torch.tensor",
+    method_name="log1p_",
+    dtype_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
+        max_value=1e37,
+    ),
+)
+def test_torch_tensor_log1p_(
+    dtype_x,
+    frontend,
+    frontend_method_data,
+    init_flags,
+    method_flags,
+    on_device,
+    backend_fw,
+):
+    input_dtype, x = dtype_x
+    helpers.test_frontend_method(
+        init_input_dtypes=input_dtype,
+        backend_to_test=backend_fw,
+        init_all_as_kwargs_np={"data": x[0]},
+        method_input_dtypes=input_dtype,
+        method_all_as_kwargs_np={},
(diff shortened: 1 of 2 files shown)


ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py
```diff
@@ -10604,6 +10604,40 @@ def test_torch_tensor_log1p(
     )
 
 
+# log1p_
+@handle_frontend_method(
+    class_tree=CLASS_TREE,
+    init_tree="torch.tensor",
+    method_name="log1p_",
+    dtype_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
```
AnnaTz, line 10613: Please change this to `helpers.get_dtypes("valid")`. Even though most tests don't do this, it's best practice to test for all "valid" dtypes. This way we can be sure that `@with_unsupported_dtypes` has been set correctly.

AnnaTz: Other than that, looks good!

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py
he11owthere: @AnnaTz, I wanted to share that I've encountered the following errors when setting it to "valid." The issue arises from the fact that when it's set to "valid," the input is tested with various data types, whereas PyTorch's `log1p_` method exclusively supports float data type. I even encountered the below errors while testing PyTorch's `log1p_` method itself for all data types other than "float."

image

I'm reaching out to inquire whether I should proceed with the proposed changes despite these challenges. Your guidance would be greatly appreciated. Thank you very much! :slight_smile:

AnnaTz: Since `log1p_` exclusively supports float tensors, we should still test for all "valid" dtypes but add `@with_supported_dtypes({"2.0.1 and below": ("float",)}, "torch")` to the function.
That's exactly the purpose of using `with_supported_dtypes`/`with_unsupported_dtypes` 🙂

he11owthere: @AnnaTz, I have completed the required modifications. I've included "float32" and "float64" in the `with_supported_dtypes` decorator since PyTorch does not support "float16" or "half". Thank you very much for your help! 😄

AnnaTz (approved):
Looks good, thanks!

PR #21812, opened 2023-08-13
Refactor torch frontend avg pooling functions

mohame54:
refactored avg pooling frontend functions and thier tests


ivy-leaves:

Thank you :hugs:

mohame54:
https://github.com/unifyai/ivy/pull/21814 
this a link to the other PR which is related to the refactoring task

ivy/stateful/module.py
```diff
@@ -5,7 +5,7 @@
 import os
 import abc
 import copy
-import dill
+#import dill
```
AnnaTz: Please remove this change. You shouldn't be updating files not related to the PR.

ivy/functional/frontends/torch/nn/functional/pooling_functions.py
```diff
@@ -62,16 +75,9 @@ def avg_pool2d(
...
-    elif input_rank == 4:
-        # NCHW
-        data_format = "NCHW"
-
+    if stride is None:
+        stride = kernel_size
+    data_format = "NCHW"
     kernel_size = _broadcast_pooling_helper(kernel_size, "2d", name="kernel_size")
     stride = _broadcast_pooling_helper(stride, "2d", name="stride")
     padding = _broadcast_pooling_helper(padding, "2d", name="padding")
```
AnnaTz, line 83: As I commented in the other PR, please check whether we really need to use `_broadcast_pooling_helper` in all these functions.

ivy/functional/frontends/torch/nn/functional/pooling_functions.py
```diff
@@ -25,10 +25,18 @@ def _broadcast_pooling_helper(x, pool_dims: str = "2d", name: str = "padding"):
         )
 
 
+@to_ivy_arrays_and_back
```
AnnaTz, line 28: Need to remove this duplicate.

ivy_tests/test_ivy/test_frontends/test_torch/test_nn/test_functional/test_pooling_functions.py
```diff
@@ -173,7 +171,12 @@ def test_torch_avg_pool3d(
 ):
     input_dtype, x, kernel_size, stride, padding = dtype_x_k_s
 
-    padding = [padding[i][0] for i in range(len(padding))]
+    if len(stride) == 1:
+        stride = [stride[0]] * 3
+    if padding == "SAME":
+        padding = calculate_same_padding(kernel_size, stride, x[0].shape[2:])
+    elif padding == "VALID":
+        padding = (0, 0, 0)
```
AnnaTz, line 179: Same comment as in the other PR. We can probably avoid handling `"VALID"` and `"SAME"` in the tests.

AnnaTz (changes requested):

ivy/functional/frontends/torch/nn/functional/pooling_functions.py
mohame54: yeah, indeed it's handled internally in ivy.avg_pool1d but we need to broadcast these attrs to determine which padding algorithm should we use.

AnnaTz: Right, I just noticed that the avg pooling ivy functions only support `str` padding. I will make a separate task to extend this. For now as you said we have to keep the broadcasting of the padding here. 
I think however that the broadcasting of `kernel_size` is not needed.

ivy_tests/test_ivy/test_frontends/test_torch/test_nn/test_functional/test_pooling_functions.py
```diff
@@ -173,8 +166,12 @@ def test_torch_avg_pool3d(
...
     input_dtype, x, kernel_size, stride, padding = dtype_x_k_s
 
-    padding = [padding[i][0] for i in range(len(padding))]
-
+    if len(stride) == 1:
+        stride = [stride[0]] * 3
+    x_shape = x[0].shape[2:]
+    padding = (pad[i] for i, pad in enumerate(padding))
+    if not sum(padding) == 0:
+        padding = calculate_same_padding(kernel_size, [stride[0]] * 3, x_shape)
```
AnnaTz, line 174: I don't think we need to process the padding returned by `arrays_for_pooling`. We have set `only_explicit_padding=True` so we are going to receive a tuple of integers as padding, which matches what the torch function expects.
Same for the rest of the tests above.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_nn/test_functional/test_pooling_functions.py
AnnaTz: This is not something we should be filtering in the tests. The tests should cover all cases supported by the ground truth function (in this case the original torch pooling function). It's up to our frontend implementation to make sure our function supports the same cases as well.

ivy_tests/test_ivy/test_frontends/test_torch/test_nn/test_functional/test_pooling_functions.py
AnnaTz: If that was needed because our frontend function currently fails for some of the cases that should be supported, you can leave this in the test but add a ToDo comment to remove that once the ivy functions have been updated to allow the frontend to handle all possible explicit padding cases.

mohame54: I broadcast the kernel_size param for the padding specification below it's essential for specifying the padding algorithm `SAME` or `VALID` until we update the current implementation of the pooling functions to accept also explicit padding we have to do so I think?

mohame54: @AnnaTz 
image
I hope my explanation make sense if not 😕 you can see from the above image if I take the padding from array for pooling directly without processing it will give me this error

mohame54: > If that was needed because our frontend function currently fails for some of the cases that should be supported, you can leave this in the test but add a ToDo comment to remove it once the ivy functions have been updated to allow the frontend to handle all possible explicit padding cases.
yeah, indeed it's related. sorry for the two comments above I wrote them cuz i didn't see your last comment

AnnaTz (approved):
Great, we can merge this then. Thanks!

PR #21820, opened 2023-08-13
Checking decorator order in `function_testing.py`

ZiadAmerr:


ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -75,6 +76,25 @@ def _find_instance_in_args(backend: str, args, array_indices, mask):
...
+def _get_decorators(func):
+    func_node = ast.parse(inspect.getsource(func)).body[0]
+    return [decorator.id for decorator in func_node.decorator_list]
+
+
+def _check_decorator_order(decorators: List[str]):
+    FN_DECORATORS = ivy.func_wrapper.FN_DECORATORS
+
+    # excludes the decorator if it is not in FN_DECORATORS -- this shouldn't happen --
+    for unknown in set(decorators).difference(set(FN_DECORATORS)):
```
ZiadAmerr, line 88:
```suggestion
    for unknown in set(decorators).difference(FN_DECORATORS):
```

sherry30 (changes requested):
I dont think you can get the decorators from the backend functions because of how they get wrapped in the backend handler. You can instead try doing `func.__dict__` and get its keys, that will contain all the wrappers (ignoring anything starting with `__` and `arrayspec`)

ZiadAmerr:
Hey @sherry30, we are working on a different approach in the `func_wrapper.py` itself to ensure accurate order by applying a set of decorators and that set is to be ordered in `func_wrapper.py`, so I guess adding the decorator order check would be redundant here. I will go ahead and close this PR.

PR #21559, opened 2023-08-09
Implement max_unpool2d

weronikazak:
Close #19761 
Close #21762

PR #21828, opened 2023-08-14
Revert "Implement max_unpool2d"

theRealBird:
Reverts unifyai/ivy#21559


VaruN-dev-dev:
Could you please provide further details on the reasons for the reversal? I'm considering implementing that change and having more context would be helpful.

PR #21833, opened 2023-08-14
add PyTorch pointwise gradient frontend + test

lukaM19:
https://github.com/unifyai/ivy/issues/21758


lukaM19:
Hi !

Although I might be wrong, as far as I understand form the display_test_results job, all the errors are in file which I have not edited at all. Any further guidance on whether I have to change anything will be greatly appriciated.

Best,
Luka.

lukaM19:
@nassimberrada  Hi, thank you so much for your feedback. I did some revisions and now the method passes most tests except when the backend is paddle, I was wondering if you could take a look at it. 

The error message is ` AttributeError: 'Tensor' object has no attribute 'to_list'. Did you mean: 'tolist'? `, I'm not sure this is something I can fix without editing one of the general ivy functional methods, which I'm not sure I can do.

Thank you for you help.

PR #21843, opened 2023-08-14
Added adjust_brightness function to paddle.vision.transforms

### LINKED ISSUE #21841: adjust_brightness

### LINKED ISSUE #22154: adjust_brightness

ozogxyz:
Close #22154


ozogxyz:
@rishabgit Hi Rishab, all tests are passing. I've only flagged `float32` and `float64` types as supported because `paddle.lerp` and `paddle.full_like` (used by `paddle.zeros_like`) currently do not support other data types.

ozogxyz:
@rishabgit Could we merge this? All tests are passing.  I've written `adjust_hue` and that was merged. After your review I'll move on to other functions in the module, I've got the images helper functions sorted out so I think most of them will be swiftly implemented. Also, maybe we could include all helpers from paddle in this file, not just the ones that are used in `adjust_hue` and `adjust_brightness` so that others working on the transforms module can also use them, there's no need for them to check channels, sizes etc within each function separately. I see an open PR #21984 also modifying helpers in this module. To avoid conflicts let's settle this down😄

ozogxyz:
@rishabgit wait, there might be a bug in `paddle`'s implementation. 

`F.adjust_brightness` and `functional_tensor.adjust_brightness` give different results. 

Their docs say a `brightness_factor` of `2.0` should give twice the image but the tensor implementation just gives back the image if `brightness_factor > 1.0`.

If the input is `np.array` it falls back to `F.cv2` and gives back `uint8`.
If the input is `paddle.Tensor` it falls back to `functional_tensor.adjust_brightness` and gives back whatever input `dtype` the input is.

Screenshot 2023-08-17 at 18 46 30

ozogxyz:
@rishabgit For now we can only support `paddle.Tensor` input. 

So if you revert the last commit [20da906](https://github.com/unifyai/ivy/pull/21843/commits/20da90646d5cb65561edda7181f0b59ddd74f1c1) it will work again. I opened an issue PaddlePaddle/Paddle#56410 on paddle repo for this, when they fix it we can add `max_value = 1.0 if ivy.check_float(img1) else 255.0` again😁

ozogxyz:
@rishabgit 

I rechecked `paddle` source code for `F_t.adjust_brightness`.

The only other usage of `_blend_images` function is for `adjust_contrast`. Before using it there they convert the image to `float32`. And `adjust_brightness` only accepts `float32` or `float64`. So that check seems a bit redundant for me.

`ivy.check_float(img1)` returns `False`, so we have two options for now:

either
```python
    max_value = 1.0 if ivy.dtype(img1) == "float32" or ivy.dtype(img1) == "float64" else 255.0
```
or

```python
    max_value = 1.0
```

either one passes the tests.

rishabgit:
Hi @ozogxyz, really appreciate the deep dive on this! :muscle: 

Not sure if I follow completely, tests seem to pass for me with the last [commit](https://github.com/unifyai/ivy/pull/21843/commits/20da90646d5cb65561edda7181f0b59ddd74f1c1) :thinking: 

Just to add more context, this is how I'm testing - 
```
pytest /workspaces/ivy/ivy_tests/test_ivy/test_frontends/test_paddle/test_vision/test_transforms.py::test_paddle_adjust_brightness
```
image

ozogxyz:
@rishabgit Sure, in short:

`ivy.check_float(img1)` returns `False` (ideally it shouldn't, its function signature states it accepts `Any` as a parameter?) so the `max_value` is always set to `255.0`, then the tests complain: 

`ret_np = array([[[2.]]]), ret_from_gt_np = array([[[1.]]])` because of some rounding stuff under the hood. But in line 7 you reassign `max_value` to `1.0` so your tests pass:)

If we want to retain future possible `int` compatibility we better use: 

```python
def _blend_images(img1, img2, ratio):
    max_value = 1.0 if ivy.dtype(img1) == "float32" or ivy.dtype(img1) == "float64" else 255.0
    return ivy.astype(
        ivy.lerp(img2, img1, float(ratio)).clip(0, max_value), ivy.dtype(img1)
    )
```

PR #21859, opened 2023-08-14
paddle.layers: add conv_general_transpose 

### LINKED ISSUE #21738: conv_general_transpose

suchunkang0822:
Close #21738

### DIFF
--- a/ivy/functional/backends/paddle/layers.py
+++ b/ivy/functional/backends/paddle/layers.py
@@ -435,4 +435,62 @@ def conv_general_transpose(
     bias: Optional[paddle.Tensor] = None,
     out: Optional[paddle.Tensor] = None,
 ):
-    raise IvyNotImplementedException()
+    if data_format == "channel_last":
+        x = x.transpose(x, (0, dims + 1, *range(1, dims + 1)))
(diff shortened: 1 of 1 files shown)


ivy/functional/backends/paddle/layers.py
```diff
@@ -435,4 +435,62 @@ def conv_general_transpose(
     bias: Optional[paddle.Tensor] = None,
     out: Optional[paddle.Tensor] = None,
 ):
-    raise IvyNotImplementedException()
+    if data_format == "channel_last":
```
ZoeCD, line 438: Thank you for your contribution! 🙌
Could you please add a unit test for the function?
Thanks!

ZoeCD (changes requested):

ivy/functional/backends/paddle/layers.py
suchunkang0822: Thanks for the update. If you don't mind, can you shed some light on where to write the unit test for this backend function? I'm scouring the doc but am struggling a little bit.

ZoeCD: Sorry, my mistake. We already have an Ivy test!

PR #21881, opened 2023-08-15
Autograd grad

somaia02:
For task: https://trello.com/c/6eTBz5Rs/1549-implement-torchautogradgrad-in-the-torch-frontend
Approach:
* Use decorator for torch frontend functions to store, in the output, the function and its inputs
* In autograd.grad wrap all functions from input to output in a single function
* Use ivy.jac to compute jacobians for the function
* Multiply jacobians with grad_outputs to get the final gradients

I'm facing a problem with Paddle backend that I can't figure out. When the input is used more than once in the computations, Paddle returns the gradients as if it was used only once. Here is an example.
```python
import ivy
ivy.set_backend("paddle")

x=torch.tensor(1., requires_grad=True)
y = 2*x
z = y+x
g=torch.autograd.grad(z,x)
print(g)
# Prints: 2
# Expected: 3
```

All the tests are failing now after I pulled the latest changes. But that's probably unrelated. Only paddle was failing before I pulled the changes. So I hope it's okay if we can focus first on the problem with paddle and deal with the tests later.


ivy-leaves:

Thank you :hugs:

vedpatwardhan (commented):
Hey @somaia02, thanks for diving into this, quite a complex aspect of the `torch` frontend. That's a great approach to support gradients similar to `torch`, probably one thing to do next is to modify the `__repr__` of the `torch.frontend.Tensor` to print the last called operation in`<>` similar to how it's done in `torch`, apart from the changes made to other functions in the frontends which seem to be unrelated to the PR, which @AnnaTz would be in a better position to comment on. Thanks!

somaia02:
The current `__repr__` returns this `return f'ivy.frontends.torch.Size([{", ".join(str(d) for d in self)}])'`
But when I use `print` I get `ivy.frontends.torch.Tensor(...)`
Does this mean `__repr__` doesn't get called when we print?

somaia02:
I removed the unrelated changes and modified the `__repr__` method. 

I kept one change that's unrelated in `function_testing`. It's just a temporary fix that I need to make the test pass. I made a separate PR here https://github.com/unifyai/ivy/pull/21314

ivy/functional/frontends/torch/func_wrapper.py
```diff
@@ -215,8 +255,6 @@ def outputs_to_frontend_arrays_torch(*args, **kwargs):
 
 
 def to_ivy_arrays_and_back(fn: Callable) -> Callable:
-    """Wrap `fn` so that input arrays are all converted to `ivy.Array` instances and
-    return arrays are all converted to `Tensor` instances."""
```
AnnaTz, line 219: Is there a specific reason you removed the docstrings? I think it's useful to have a brief explanation of what each decorator does.

AnnaTz: Other than that, everything looks good!

AnnaTz (changes requested):

ivy/functional/frontends/torch/func_wrapper.py
AnnaTz: What if you put both `"""` in separate lines?

somaia02: I skipped the pre-commit hook and finaly manged to commit. Thanks.

somaia02:
Hello @vedpatwardhan Tests are passing now. What should I do about Paddle? I tried really hard but I can't figure out what's the problem.

vedpatwardhan:
> Hello @vedpatwardhan Tests are passing now. What should I do about Paddle? I tried really hard but I can't figure out what's the problem.

Not sure about the solution for `paddle`, @MahmoudAshraf97 could you please take a look? Thanks 😄

vedpatwardhan:
ivy-gardener

MahmoudAshraf97:
Hi @vedpatwardhan and @somaia02 , after a quick investigation I fount that the problem is mostly not in `ivy.jac` implementation in the paddle backend because it returns the correct result if torch frontend is not involved
```python
import ivy
from ivy.functional.frontends.torch.autograd import (
    _get_wrapped_fn,
    to_frontend_array_and_back,
)
from ivy.functional.frontends.torch.func_wrapper import to_ivy_arrays_and_back
ivy.set_backend("paddle")


## minimal example after isolating unnecessary code in torch frontend
x = torch_frontend.tensor(1.0, requires_grad=True)
y = 2 * x
z = y + x
fn = to_ivy_arrays_and_back(ivy.jac(to_frontend_array_and_back(_get_wrapped_fn(z, x))))
print(fn(x))
# ivy.frontends.torch.Tensor(2.)


## same example in ivy functional code directly
x2 = ivy.array(1.0)
lambda_1 = lambda x: 2 * x
lambda_2 = lambda x: lambda_1(x) + x
fn_jac = ivy.jac(lambda_2)
print(fn_jac(x2))
# ivy.array(3.)
```

so my guess is that `_get_wrapped_fn` needs more investigation to see the real cause of failure in paddle backend, I quickly checked and the function returns the correct outputs for any input but returns the wrong gradients in paddle backend

somaia02:
Thanks a lot for looking into this @MahmoudAshraf97 The minimal example was very helpful. I was trying to make it more minimal and found out the issue is actually in the wrapper `to_frontend_array_and_back`. 

When we convert to torch_frontend if I  don't set `requires_grad=True` it outputs the correct results. But I need to set it to True to be able to calculate high order gradients. We will need to look into what happens when we set `requires_grad=True`

Thank you very very much for your help.

PR #21904, opened 2023-08-15
Decorator Factory Testing out

ZiadAmerr:


vedpatwardhan (commented):
Looks like a clean approach to reduce redundancy while also not affecting readability. Feel free to move forward with this approach for the other files, probably the only issue here would be duplication of decorator sequences if that sequence of decorators is already existing in the file and isn't noticed.
We could also try a more hierarchical approach to avoid this duplication in the sense that the decorators common to all functions from the larger group, then different sub-groups for the decorators that differ and so on.
Hope that makes sense, thanks @ZiadAmerr 😄

ZiadAmerr:
So should I keep that DECORATORS_PER_FUNC in func_wrapper.py and let it contain all the functions in ivy instead of one for each file?

vedpatwardhan:
I think one to each file would be better, as it would be easier to find the decorators applied to a particular function 😄

ZiadAmerr:
Ah yes, got you. Then can you explain what you mean by the hierarchal approach?

ZiadAmerr:
If we're talking about common decorators as to remove redundancy from the frozen sets, then I guess this couldn't happen in elementwise as all the sets have no common decorator

ZiadAmerr:
> We could also try a more hierarchical approach to avoid this duplication in the sense that the decorators common to all functions from the larger group, then different sub-groups for the decorators that differ and so on. Hope that makes sense, thanks @ZiadAmerr 😄

Oh got it, but in this case I will kind-of "flip" the data structure, it will be a group of decorators alongside a list of functions taking these decorators, and in the case of elementwise, it will be a decorator corresponding to a set of functions. I don't really know, but I think this will introduce a bit of duplication?

What I initially thought of is creating a list of decorators used in that particular file, and the just removing the unneeded decorators for a group of functions, it would have decreased the redundancy a bit instead of creating a list of decorators for each particular set of functions, but instead it will add some additional computational complexity to the file, making it less readable for less redundancy. Generally, it is a trade-off, and I think this structure would work best.

Another optimization was that I would go and make the dict's values sets instead of lists, so that I can recall them back later, add some decorators and remove some decorators without worrying about the order. And then right after the dict definition, I would run a `rearrange_decorators_of_map(DECORATORS_PER_FUNC)` that will convert them back to lists with the correct order. Lmk what you think about this :D

ivy/func_wrapper.py
```diff
@@ -302,6 +302,46 @@ def _check_in_nested_sequence(sequence, value=None, _type=None):
...
+    fn_name : str
+        The name of the function to get the decorators for
+    
+    decorator_lookup : dict
+        The lookup table of frozen sets of function names to decorators list
+    """
+    for funcs, decorators in decorator_lookup.items():
+        if fn_name in funcs:
+            return decorators
+    return []
```
ZiadAmerr, line 319:
```suggestion
    raise KeyError(f"Function `{fn_name}` not found in lookup table!")
```

Should we raise an error when the function name is not found in the lookup table?

vedpatwardhan:
> > We could also try a more hierarchical approach to avoid this duplication in the sense that the decorators common to all functions from the larger group, then different sub-groups for the decorators that differ and so on. Hope that makes sense, thanks @ZiadAmerr 😄
> 
> Oh got it, but in this case I will kind-of "flip" the data structure, it will be a group of decorators alongside a list of functions taking these decorators, and in the case of elementwise, it will be a decorator corresponding to a set of functions. I don't really know, but I think this will introduce a bit of duplication?
> 
> What I initially thought of is creating a list of decorators used in that particular file, and the just removing the unneeded decorators for a group of functions, it would have decreased the redundancy a bit instead of creating a list of decorators for each particular set of functions, but instead it will add some additional computational complexity to the file, making it less readable for less redundancy. Generally, it is a trade-off, and I think this structure would work best.
> 
> Another optimization was that I would go and make the dict's values sets instead of lists, so that I can recall them back later, add some decorators and remove some decorators without worrying about the order. And then right after the dict definition, I would run a `rearrange_decorators_of_map(DECORATORS_PER_FUNC)` that will convert them back to lists with the correct order. Lmk what you think about this :D

Yeah I think your current approach would be much cleaner than that, let's continue with this 👍

ivy/func_wrapper.py
vedpatwardhan: yep, along with a message asking them to add the function to that list, thanks!

Infrared1029:
Slightly inclined towards the simpler approach we discussed, but some interesting points were raised (contributors finding stuff easily etc) although i'd argue that we shouldn't be complicating our design  (adding the function name in two places whenever a function is added) to make it easier for incompetent contributions, as it is very simple to just look at similar functions in the file and infer the right list/set/tuple of decorators to use, regardless, great work Ziad, move on with whatever path you and Ved think makes more sense.

vedpatwardhan:
> Slightly inclined towards the simpler approach we discussed, but some interesting points were raised (contributors finding stuff easily etc) although i'd argue that we shouldn't be complicating our design (adding the function name in two places whenever a function is added) to make it easier for incompetent contributions, as it is very simple to just look at similar functions in the file and infer the right list/set/tuple of decorators to use, regardless, great work Ziad, move on with whatever path you and Ved think makes more sense.

Hey @Infrared1029, I completely agree that the approach you're suggesting is cleaner than the frozenset one @ZiadAmerr, but probably considering the advantage of knowing what decorators were being used by similar functions in the submodule, probably the approach @ZiadAmerr suggested is less likely to be messed up. Let's probably continue with this approach, and keep tabs on the likelihood of such errors and decide from there on. Thanks @Infrared1029 and @ZiadAmerr 😄

ivy/func_wrapper.py
```diff
@@ -597,17 +598,10 @@ def to_ivy_arrays_and_back(fn: Callable) -> Callable:
     instances and return arrays are all converted to `ivy.NativeArray`
     instances.
     """
-    return output_to_native_arrays(inputs_to_ivy_arrays(fn))
+    return outputs_to_native_arrays(inputs_to_ivy_arrays(fn))
 
 
 def to_native_arrays_and_back(fn: Callable) -> Callable:
```
ZiadAmerr, line 604:
```suggestion
def to_native_arrays_and_back(fn: Callable) -> Callable:
    """
    Make `fn` receive `ivy.NativeArray` and return `ivy.Array`.
    Wrap `fn` so that input arrays are all converted to
    `ivy.NativeArray` instances and return arrays are all converted to
    `ivy.Array` instances.
    """
```

ZiadAmerr:
Hey @vedpatwardhan, all `functional/ivy` files are done. Except for `general.py` because it raises a very strange error with `inspection.py`. We can discuss that as soon as we meet :)

vedpatwardhan (commented):
> Hey @vedpatwardhan, all `functional/ivy` files are done. Except for `general.py` because it raises a very strange error with `inspection.py`. We can discuss that as soon as we meet :)

Hey @ZiadAmerr, the changes made look good to go (I haven't really dived into it line-by-line though). But there's quite a few test failures in the CI, could you please check if they're related to the changes made? 
Sure, we can discuss `general.py` so that we can get that working and merge this PR as there would be merge conflicts to constantly deal with. 
Thanks for diving into this, let's see what @Infrared1029 thinks 😄

ivy-seed:

vedpatwardhan:
<some comment>

PR #21922, opened 2023-08-15
Hypergeometric test and function redo

### LINKED ISSUE #21387: hypergeometric

Svet0zar:
Close #21387

### DIFF
--- a/ivy/functional/frontends/numpy/random/functions.py
+++ b/ivy/functional/frontends/numpy/random/functions.py
@@ -1,4 +1,7 @@
 # local
+import random
+
+
 import ivy
 from ivy.functional.frontends.numpy.func_wrapper import (
     to_ivy_arrays_and_back,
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/numpy/random/functions.py
```diff
@@ -233,3 +236,27 @@ def triangular(left, mode, right, size=None):
...
+            ngood -= 1
+        elif nbad > 0:
+            nbad -= 1
+    return sum(u)
+
+
+@to_ivy_arrays_and_back
+@from_zero_dim_arrays_to_scalar
+def hypergeometric(ngood, nbad, nsample, size=None):
+    u = ivy.empty(size, dtype=int)
```
jkeane508, line 258: `size` argument is default None which is a required argument for `empty` which will always throw an error. As seen in the test logs [here](https://github.com/unifyai/ivy/actions/runs/5868789216/job/15912380408?pr=21922#step:3:247).

jkeane508 (changes requested):

ivy/functional/frontends/numpy/random/functions.py
Svet0zar: Whenever I reproduce any of the failures I get an error that states that the test passed

Svet0zar: I tested the test cases locally and they don't fail like they do in the test suite. I even compared it to the numpy.random.hypergeometric module and they produce the same results (having in mind that the numbers are different due to the randomness)

jkeane508: yea makes sense. The issue is a flaw in the logic. If you test call this function and don't specify a value for `size` then this would break because `None` is not a valid argument for `size` in `ivy.empty`.

Svet0zar: I dont get that issue in the new commits anymore

jkeane508: Are all tests passing for you locally?

Svet0zar: locally I dont get any errors, yes

Svet0zar: I just want to mention, I have this many commits because my local tests only run for numpy and don't check the other backends.

PR #21955, opened 2023-08-15
`torch.linalg.norm` frontend

### LINKED ISSUE #21954: torch.norm

mobley-trent:
Closes #21954


mobley-trent:
@paulaehab do you mind reviewing this PR. I saw that you worked on `torch.reduction_ops.norm` which is essentially a replica of this function,

MahmoudAshraf97 (changes requested):
I have a couple of questions, if the two functions are essentially the same why write a different test strategy instead of using the old one or extending it if needed?
also note that `torch.norm` is deprecated
the tests are failing probably due to wrong input from test strategy

mobley-trent:
@MahmoudAshraf97 I tried using the same strategy but it is failing for some reason. I'm getting an error that the combination of inputs being fed into the function is wrong yet both of these functions have the same input signatures. Also if `torch.norm` is deprecated, why was it implemented first instead of `torch.linalg.norm` ?

MahmoudAshraf97:
> @MahmoudAshraf97 I tried using the same strategy but it is failing for some reason. I'm getting an error that the combination of inputs being fed into the function is wrong yet both of these functions have the same input signatures. Also if `torch.norm` is deprecated, why was it implemented first instead of `torch.linalg.norm` ?

because we implement deprecated functions as a part of multiversion compatibility
I guess you should keep trying with the test strategy and let me know if you need help debugging it

PR #21967, opened 2023-08-15
refactor torch frontend lp pooling functions and their tests

mohame54:
refactored lp pooling torch frontend functions


AnnaTz (approved):
Great, thanks!

PR #21997, opened 2023-08-16
Refactor `ivy.clip` function

NripeshN:


NripeshN:
ivy-gardner

hello-fri-end (changes requested):
Hey @NripeshN ! Thanks for working on this issue. The way you have tried to fix this problem is not the correct way to go about it. Since `clip` is a primary function, you need to make the change in all the backend implementations of `clip`. For example, numpy already supports one of `x_min` or `x_max` to be `None`, so no update is needed there but in the jax backend it's implemented using `jax.where` so, you should first check that both values are not `None`at the same time, like:
```python
    if x_min is None and x_max is None:
        raise ValueError("At least one of the x_min or x_max must be provided")
```
And then clip the values like :

```python 
if x_max is not None:
x = jnp.where(x > x_max, x_max, x)

if x_min is not None:
return jnp.where(x < x_min, x_min, x)
```

You can use the same idea to update the other backend implementations as well if they don't natively support one of the arguments to be `None`. Finally, you should update the test of `ivy.clip` such that one of `x_min` or `x_max` can be `None`. 

Feel free to request another review when you have made the changes.

NripeshN:
ivy-gardener

ivy-leaves:

Thank you :hugs:

ivy/functional/backends/paddle/manipulation.py
```diff
@@ -446,13 +446,20 @@ def swapaxes(
...
     /,
     *,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    return paddle_backend.minimum(paddle_backend.maximum(x, x_min), x_max)
+    if x_min is None and x_max is None:
+        raise ValueError("At least one of the x_min or x_max must be provided")
+
+    if x_min is not None:
+        x = paddle.maximum(x, x_min)
```
hello-fri-end, line 459:
```suggestion
        x = paddle_backend.maximum(x, x_min)
```

ivy/functional/backends/paddle/manipulation.py
```diff
@@ -446,13 +446,20 @@ def swapaxes(
...
+    if x_max is not None:
+        x = paddle.minimum(x, x_max)
```
hello-fri-end, line 461:
```suggestion
        x = paddle_backend.minimum(x, x_max)
```

hello-fri-end (changes requested):
Could you please update the type hints in the numpy backend and update the test to allow one of `x_min` or `x_max` to be `None`?

hello-fri-end (changes requested):
Awesome work so far, would you please update the tests as well :)

ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py
```diff
@@ -340,13 +340,14 @@ def _basic_min_x_max(draw):
...
-    min_val = draw(helpers.array_values(dtype=dtype[0], shape=()))
-    max_val = draw(
-        helpers.array_values(dtype=dtype[0], shape=()).filter(lambda x: x > min_val)
-    )
+    if random.choice([True, False]):
+        min_val = draw(helpers.array_values(dtype=dtype[0], shape=()))
+        max_val = None
+    else:
+        min_val = None
+        max_val = draw(helpers.array_values(dtype=dtype[0], shape=()))
```
hello-fri-end, line 348: There's no need to use the `random` module for this. It's possible to generate `None` value using hypothesis in the following way:

```python
min_val = draw(st.one_of(st.just(None), helpers.array_values(dtype=dtype[0], shape=()))
max_val = draw(st.one_of(st.just(None), helpers.array_values(dtype=dtype[0], shape=()).filter(lambda x: x > min_val)))
```

hello-fri-end (changes requested):

hello-fri-end:
ivy-gardener

ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py
```diff
@@ -340,9 +339,16 @@ def _basic_min_x_max(draw):
...
+        st.one_of(st.just(None), helpers.array_values(dtype=dtype[0], shape=()))
+    )
     max_val = draw(
-        helpers.array_values(dtype=dtype[0], shape=()).filter(lambda x: x > min_val)
+        st.one_of(
+            st.just(None),
+            helpers.array_values(dtype=dtype[0], shape=()).filter(
+                lambda x: x > min_val
+            ),
+        )
```
ShreyanshBardia, line 351:
```suggestion
           if generate_max:
        scalar = min_val.tolist() if min_val is not None else None
        max_val = draw(
                helpers.array_values(dtype=dtype[0], shape=(),min_value=scalar)
            )
    if min_val is None and max_val is None:
        generate_max = draw(st.booleans())
        if generate_max:
            max_val = draw(helpers.array_values(dtype=dtype[0], shape=()))
        else:
            min_val = draw(helpers.array_values(dtype=dtype[0], shape=()))
```

ShreyanshBardia (commented):
This will make sure either one of them is not None

ivy/functional/backends/tensorflow/manipulation.py
```diff
@@ -336,27 +336,34 @@ def swapaxes(
...
+        if x_max is not None:
+            x_max = tf.cast(x_max, promoted_type)
+
     if tf.size(x) == 0:
         ret = x
     elif x.dtype == tf.bool:
         ret = tf.clip_by_value(tf.cast(x, tf.float16), x_min, x_max)
         ret = tf.cast(ret, x.dtype)
     else:
         ret = tf.clip_by_value(x, x_min, x_max)
```
ShreyanshBardia, line 365:
```suggestion
    if x_min is not None and x_max is not None:
        if tf.size(x) == 0:
            ret = x
        elif x.dtype == tf.bool:
            ret = tf.clip_by_value(tf.cast(x, tf.float16), x_min, x_max)
            ret = tf.cast(ret, x.dtype)
        else:
            ret = tf.clip_by_value(x, x_min, x_max)
    else:
        ret = tf.experimental.numpy.clip(x, x_min, x_max)
```

ShreyanshBardia (commented):
Using tf.experimental.numpy when one of the value is `None`, since `tf.clip_by_value` doesn't accept `None`

ShreyanshBardia:
Hi @hello-fri-end, tensorflow differs from all other frameworks when `x_min > x_max`, so should we standardise this edge case by considering numpy as the standard? Here's an example

```
import numpy as np
import tensorflow as tf

tf.clip_by_value(tf.constant([1,10,13]),12,11)

np.clip([1,10,13],12,11)
```

hello-fri-end:
> Hi @hello-fri-end, tensorflow differs from all other frameworks when `x_min > x_max`, so should we standardise this edge case by considering numpy as the standard, and override tensorflow's behaviour? Here's an example
> 
> ```
> import numpy as np
> import tensorflow as tf
> 
> tf.clip_by_value(tf.constant([1,10,13]),12,11)
> 
> np.clip([1,10,13],12,11)
> ```

Hey @ShreyanshBardia ! Thanks for pointing this out :rocket: It seems that `torch.clamp` also has the same behavior i.e it clips using the `min` value first and then by the `max` value:
```python
import numpy as np
import tensorflow as tf
import torch

print(tf.clip_by_value(tf.constant([1,10,13]),12,11))

print(np.clip([1,10,13],12,11))

print(torch.clamp(torch.tensor([1, 10, 13]), min=12, max=11))
```
```
>>>tf.Tensor([12 12 12], shape=(3,), dtype=int32)
>>>[11 11 11]
>>>tensor([11, 11, 11])
```

Would you please update our implementations to match this behavior? @NripeshN In the tensorflow backend implementation, we can use `tf.experimental.numpy.clip` for all the cases while for jax and paddle, where the function is implemented compositionally, we should clip using `x_min` first and then by `x_max`. You should also update the test where you'd need to remove the condition that `x_max` always needs to greater than `x_min`. Feel free to request another review when you are done. Thanks :)

ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py
```diff
@@ -340,9 +339,21 @@ def _basic_min_x_max(draw):
             available_dtypes=helpers.get_dtypes("numeric"),
         )
     )
-    min_val = draw(helpers.array_values(dtype=dtype[0], shape=()))
+    min_val,max_val = None, None
+    if generate_min:
+        min_val = draw( helpers.array_values(dtype=dtype[0], shape=()))
     max_val = draw(
```
ShreyanshBardia, line 345:
```suggestion
```

ShreyanshBardia: @NripeshN, can you please fix the indentation for `if generate_max`

ShreyanshBardia (commented):
Indentation error

ivy/functional/backends/tensorflow/manipulation.py
```diff
@@ -356,16 +356,16 @@ def clip(
...
+    if cond:
         if tf.size(x) == 0:
             ret = x
         elif x.dtype == tf.bool:
             ret = tf.clip_by_value(tf.cast(x, tf.float16), x_min, x_max)
             ret = tf.cast(ret, x.dtype)
         else:
-            ret = tf.clip_by_value(x, x_min, x_max)
+            ret = tf.experimental.numpy.clip(x, x_min, x_max)
     else:
```
hello-fri-end, line 367: I'm not sure if I follow why we need an else case here since `x_min is not None and x_max is not None and x_min > x_max` case (i.e `not cond`) can be handled by `tf.experimental.numpy.clip` as well, right? Maybe I'm  missing something here..

Also, it looks like the frontend tests are failing because of `cond` with the following error:
```python
E     ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

This is because `x_min` and `x_max` can also be arrays.

ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py
```diff
@@ -339,22 +339,18 @@ def _basic_min_x_max(draw):
             available_dtypes=helpers.get_dtypes("numeric"),
         )
     )
-    min_val,max_val = None, None
-    if generate_min:
-        min_val = draw( helpers.array_values(dtype=dtype[0], shape=()))
+    min_val = draw(
+        st.one_of(st.just(None), helpers.array_values(dtype=dtype[0], shape=()))
+    )
     max_val = draw(
```
hello-fri-end, line 345: Can we update `min_val` and `max_val` such that these can be arrays as well? They can be of any shape that is broadcastable to the shape of the input `x`. You should use the broadcast helper functions in the Array helpers section here --> :https://unify.ai/docs/ivy/docs/helpers/ivy_tests.test_ivy.helpers.hypothesis_helpers/ivy_tests.test_ivy.helpers.hypothesis_helpers.array_helpers.html#ivy_tests.test_ivy.helpers.hypothesis_helpers.array_helpers.mutually_broadcastable_shapes

hello-fri-end (changes requested):
Hey @NripeshN ! Awesome work so far, just some minor comments this time. Will be happy to merge once these are resolved :)

ivy/functional/backends/tensorflow/manipulation.py
ShreyanshBardia: Oh right, sorry missed the fact that they could be arrays too, I think `.any` should be able to resolve this. `tf.experimental.numpy.clip` also follows `tf.clip_by_value` behaviour when `x_min>x_max`,  which we discussed in the above comments. I think the only difference is numpy namespace and the fact that it can accept `None`. You can verify this from the following code.
```>>> print(tf.clip_by_value(tf.constant([1,10,13]),12,11))
tf.Tensor([12 12 12], shape=(3,), dtype=int32)
>>> print(tf.experimental.numpy.clip(tf.constant([1,10,13]),12,11))
tf.Tensor([12 12 12], shape=(3,), dtype=int32)
```

hello-fri-end: > Oh right, sorry missed the fact that they could be arrays too, I think `.any` should be able to resolve this. `tf.experimental.numpy.clip` also follows `tf.clip_by_value` behaviour when `x_min>x_max`, which we discussed in the above comments. I think the only difference is numpy namespace and the fact that it can accept `None`. You can verify this from the following code.
> 
> ```
> >>> print(tf.clip_by_value(tf.constant([1,10,13]),12,11))
> tf.Tensor([12 12 12], shape=(3,), dtype=int32)
> >>> print(tf.experimental.numpy.clip(tf.constant([1,10,13]),12,11))
> tf.Tensor([12 12 12], shape=(3,), dtype=int32)
> ```

Ah, that's interesting. I expected `tf.experimental.numpy.clip` to behave exactly the same as `np.clip`.

ivy/functional/backends/numpy/manipulation.py
```diff
@@ -261,13 +261,13 @@ def unstack(
...
-    x_min: Union[Number, np.ndarray],
-    x_max: Union[Number, np.ndarray],
+    x_min: Union[float, int, np.ndarray] = None,
+    x_max: Union[float, int, np.ndarray] = None,
     /,
     *,
     out: Optional[np.ndarray] = None,
 ) -> np.ndarray:
-    return np.asarray(np.clip(x, x_min, x_max, out=out), dtype=x.dtype)
+    return np.clip(x, x_min, x_max, out=out)
```
hello-fri-end, line 270:
```suggestion
    promoted_type = x.dtype
    if x_min is not None:
        promoted_type = ivy.promote_types(promoted_type, x_min.dtype)
    if x_max is not None:
        promoted_type = ivy.promote_types(promoted_type, x_max.dtype)

    return np.asarray(np.clip(x, x_min, x_max, out=out), dtype= promoted_type)
```

This should fix your failing numpy test.

hello-fri-end (changes requested):
Just left a comment that will fix your failing numpy test. Another thing to look into is the following error from the frontend tests:

```
    promoted_type = ivy.as_native_dtype(ivy.promote_types(x.dtype, x_min.dtype))
E   AttributeError: 'float' object has no attribute 'dtype'
```
Since `x_min` and `x_max` can also be `Numbers`, they won't necessarily have the `dtype` attribute. Would you please update all the implementations to check if this attribute is present first and then try to access it.

NripeshN:
ivy-gardener

NripeshN:
ivy-gardener

hello-fri-end (approved):
Looks like the failures in the CI are unrelated to this function now. Feel free to merge it - thanks @NripeshN @ShreyanshBardia :rocket:

PR #22007, opened 2023-08-16
Builtins frontend

illia-bab:


ivy/functional/frontends/builtins/built_in_functions.py
```diff
@@ -0,0 +1,37 @@
...
+    return ivy.all(iterable)
+
+
+@to_ivy_arrays_and_back
+@from_zero_dim_arrays_to_scalar
+def any(iterable):
+    return ivy.any(iterable)
+
+
+@from_zero_dim_arrays_to_scalar
```
AnnaTz, line 33: Doesn't this need `@to_ivy_arrays_and_back` as well?

ivy/functional/frontends/builtins/math/power_and_logarithmic_functions.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@from_zero_dim_arrays_to_scalar
+def log2(x):
+    return ivy.log2(x)
+
+
+@from_zero_dim_arrays_to_scalar
+def exp(x):
+    return ivy.exp(x)
```
AnnaTz, line 31: Likewise shouldn't we add `@to_ivy_arrays_and_back` in these functions as well?

ivy/functional/frontends/builtins/func_wrapper.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # convert native arrays and lists to ivy arrays
+    elif isinstance(x, ivy.NativeArray):
+        return ivy.array(x)
+
+    return x
+
+
+def _infer_return_array(x: Iterable) -> Callable:
+    # get module path
```
AnnaTz, line 22: Could you add a docstring that explains what this decorator does overall? It is not very obvious when it should be used.

AnnaTz (changes requested):

ivy/functional/frontends/builtins/built_in_functions.py
illia-bab: Natively, this function operates on scalars and returns scalars, so we're not expecting an array input and this is the reason why I haven't added `@to_ivy_arrays_and_back`

ivy/functional/frontends/builtins/math/power_and_logarithmic_functions.py
illia-bab: This function from `math` built-in module also accepts scalars and returns scalars

ivy/functional/frontends/builtins/func_wrapper.py
illia-bab: This function should always be used within `@to_ivy_arrays_and_back` decorator to infer the return array type. Let's consider the `jax` frontend. In this case, if decorator receives `jax.Array`, its frontend implementation or `list` the output will always be converted to a frontend implementation of `jax.Array`. However, in case of `builtins` the behaviour changes to: `almost_any_array -> ivy.Array -> almost_any_array`. In other words, we need to track what array to convert back to. So, if our built-in function recieves `np.ndarray` or its frontend implementation, the output will be converted to `np_frontend.ndarray` and so on. It also has a default behaviour. By default we convert the output to a frontend implementation of array which corresponds to current backend. This was done to cover the cases, when function accepts scalar arguments, but returns an array. For example, `range`. But yeah I will add a docstring.

ivy/functional/frontends/builtins/func_wrapper.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def _to_ivy_array(x):
+    # if x is any frontend array including frontend list
+    # then extract the wrapped ivy array
+    if hasattr(x, "ivy_array"):
+        return x.ivy_array
+
+    # convert native arrays and lists to ivy arrays
+    elif isinstance(x, ivy.NativeArray):
```
AnnaTz, line 15: If we want lists to get converted, we have to change this to `else`.

ivy/functional/frontends/builtins/func_wrapper.py
```diff
@@ -0,0 +1,101 @@
...
+def outputs_to_frontend_arrays(fn: Callable) -> Callable:
+    @functools.wraps(fn)
+    def _outputs_to_frontend_arrays(*args, **kwargs):
+        # update config for jax backend
+        if ivy.current_backend_str() == "jax":
+            import jax
+
+            jax.config.update("jax_enable_x64", True)
+
+        ret, frontend_array = fn(*args, **kwargs)
```
AnnaTz, line 73: The way `inputs_to_ivy_arrays` and `outputs_to_frontend_arrays` have been designed right now they can't be used individually. They would only work when used together in `to_ivy_arrays_and_back`. 
E.g. `ret, frontend_array = fn(*args, **kwargs)` would cause an error if `fn` was an unwrapped frontend function because there would not be a `frontend_array` returned.

ivy/functional/frontends/builtins/func_wrapper.py
```diff
@@ -0,0 +1,101 @@
...
+
+        if isinstance(ret, (ivy.Array, ivy.NativeArray)):
+            return frontend_array(ret)
+        return ret
+
+    return _outputs_to_frontend_arrays
```
AnnaTz, line 79: Actually I think that built-in functions will always produce outputs in standard Python data types (scalars, lists, etc.), even when applied to iterables containing some kind of framework array/tensor. 
If I'm right, then we would not need `outputs_to_frontend_arrays` at all. We would have to make sure each builtin frontend function returns the type it's supposed to inside its implementation. For example, `min` should return a list, so at the end we would do `ivy.min(args).to_list()`.

ivy/functional/frontends/builtins/func_wrapper.py
```diff
@@ -0,0 +1,101 @@
...
+
+def _infer_return_array(x: Iterable) -> Callable:
+    # get module path
+    module_path = x.__class__.__module__.split(".")
+
+    # if function's input is a scalar, list, or tuple
+    # convert to current backend's frontend array
+    if "builtins" in module_path:
+        cur_backend = ivy.current_backend_str()
+        module_str = cur_backend if len(cur_backend) != 0 else "numpy"
```
AnnaTz, line 29: Regardless of everything else, this would not be correct because we want to mimic the original function exactly and builtin functions never return an "array" asfaik.

AnnaTz (changes requested):

ivy/functional/frontends/builtins/func_wrapper.py
illia-bab: Yeah, I hust forgot to modify the comment.

ivy/functional/frontends/builtins/func_wrapper.py
illia-bab: I was experimenting with this. Here's the results:
```
# Example for NumPy
>>> a = np.array([1, 2, 3, 4])
>>> type(abs(a))
numpy.ndarray
>>> type(all(a))
bool
>>> type(min(a))
numpy.int32
>>> type(max(a))
numpy.int32
>>> type(sum(a))
numpy.int32

# Example for torch
>>> a = torch.Tensor([1, 2, 3, 4])
>>> type(abs(a))
torch.Tensor
>>> type(all(a))
bool
>>> type(min(a))
torch.Tensor
>>> type(max(a))
torch.Tensor
>>> type(sum(a))
torch.Tensor
```
`min`, `max`, `sum` functions return 0-d array in this case, but it's still a `torch.Tensor`

ivy/functional/frontends/builtins/func_wrapper.py
```diff
@@ -0,0 +1,101 @@
...
+
+        if isinstance(ret, (ivy.Array, ivy.NativeArray)):
+            return frontend_array(ret)
+        return ret
+
+    return _outputs_to_frontend_arrays
+
+
+def to_ivy_arrays_and_back(fn: Callable) -> Callable:
+    return outputs_to_frontend_arrays(inputs_to_ivy_arrays(fn))
```
AnnaTz, line 94: - Make another wrapper inside here like:
```
@functools.wraps(fn) 
def _to_ivy_arrays_and_back(*args, **kwargs):
```
- Move the `frontend_array = _infer_return_array(args[0]) if len(args) != 0 else None` to here.
- And then pass `frontend_array` to `outputs_to_frontend_arrays` by adding it as a new argument like `outputs_to_frontend_arrays(fn: Callable, frontend_array: str)`.

illia-bab: Thanks! I'll try this approach!

Issue #22035, opened 2023-08-16
reciprocal_


he11owthere:
To-do list: https://github.com/unifyai/ivy/issues/3612

PR #22036, opened 2023-08-16
Added reciprocal_ method in the PyTorch frontend

he11owthere:
closes https://github.com/unifyai/ivy/issues/22035

image

### DIFF
--- a/ivy/functional/frontends/torch/tensor.py
+++ b/ivy/functional/frontends/torch/tensor.py
@@ -1053,6 +1053,13 @@ class Tensor:
     def remainder(self, other, *, out=None):
         return torch_frontend.remainder(self, other, out=out)
 
+    @with_supported_dtypes(
+        {"2.0.1 and below": ("float16", "float32", "float64", "bfloat16")}, "torch"
+    )
+    def reciprocal_(self):
+        self.ivy_array = torch_frontend.reciprocal(self).ivy_array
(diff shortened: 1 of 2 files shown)


he11owthere:
@Ookamice please remove "bfloat16" from the `with_supported_dtypes` decorator before running the tests until the issue with `ivy.Array.repr` method has been fixed as mentioned by Ved in https://github.com/unifyai/ivy/pull/21447. Thank you! 🙂

Ookamice:
@he11owthere 
Hiya, thanks for the PR. From reading the PyTorch documentation, it appears that integer types should also be supported for [reciprocal](https://pytorch.org/docs/stable/generated/torch.reciprocal.html). Please could you add these to the function decorator

he11owthere:
@Ookamice, I've noticed that PyTorch's `reciprocal` method doesn't seem to support integers, contrary to what's stated in the documentation. When I provide integer inputs, I encounter an error message: `result type Float can't be cast to the desired output type Int.` If there's something I'm missing or if there's a recommended approach to handle this, I'd greatly appreciate your insights. Thank you for your time and assistance!

Ookamice:
@he11owthere 
No worries then, I suspect this is more to do with Ivy's wrapping where there is some implicit conversion of the output back to the input dtype. Hence why the error is about going from float to int.

PR #22054, opened 2023-08-17
Added nanmedian function to ReductionOps for Pytorch Frontend.

### LINKED ISSUE #21830: nanmedian

constanreedjohn:
Close #21830


tomatillos:
Hi, good work so far. I think the tests don't generate values including NaN, so the test isn't fully covering the behaviour of the function. I also suspect that the function won't have the intended behaviour, since I don't think account for ignoring the NaN values when taking the median? Still a great start, let me know if you need any guidance with this :)

constanreedjohn:
Hi thank you for your response. I agree with you that accounting for NaN values won't be as crucial as common function. About the ignoring Nan values, I've read that we can replace the NaN values with 0 or remove the NaN values from the input., what's your opinion on this?
During implementation, I've tried to use the backend ivy.nanmedian() for the function but the function failed the test so I had to implemented it from scratch. Can you suggest how I can modify the backend ivy.nanmedian() and called the function directly in the torch frontend nanmedian task?
Also how can I finished the task from this PR?

constanreedjohn:
Hi @tomatillos , please guide me what should I do next to finish the task and move on with other tasks.

tomatillos:
I think you should remove the NaN values from input, as described by https://pytorch.org/docs/stable/generated/torch.nanmedian.html

The other thing to do is to fix the test, so that it generates NaN values (at the moment it is only generating regular values, so we aren't fully testing the function on inputs it would be likely to receive)

constanreedjohn:
I'll keep you updated, thank you.

constanreedjohn:
Hi @tomatillos, I am currently stuck at the test cases, basically I run the test then print each of the input test case and focus on the final summary. But I am currently stuck at a loop hole where the shape between the groundtruth and the function are unmatch.

For example, test case raises unmatch error between gt shape = (1, ) and function shape = (), I fix this error but then the test case raises unmatch error between gt shape = () and function shape = (1, ) and over again. 

Can you please check my code and maybe debug where I am wrong?

**Information about the code:**
> nanmedian codes: /home/john/Unify/ivy/ivy/functional/frontends/torch/reduction_ops.py:195
> test nanmedian codes: /home/john/Unify/ivy/ivy_tests/test_ivy/test_frontends/test_torch/test_reduction_ops.py:440

tomatillos:
Hi John, I don't quite follow the problem you're having without explicit examples, but one place that I can see that the code will go wrong is on line 209 where you index: `input = input[~ivy.isnan(input)]`. The result of this operation will often lose the shape, since the result will be a non-square tensor. Therefore I'd recommend only using this line on each row, before you compute the median. You also need some logic where if every value is NaN, the result is NaN.

constanreedjohn:
Thanks, I will fix the problem.

PR #22057, opened 2023-08-17
Add apple silicon GPU Acceleration support("mps")

NripeshN:


NripeshN:
ivy-gardener

ivy-leaves:

Thank you :hugs:

vedpatwardhan (changes requested):
Hey @NripeshN, thanks for adding support for the torch backend, could you please also update the other backends and the corresponding __init__.py files. Thanks 😄

NripeshN:
> Hey @NripeshN, thanks for adding support for the torch backend, could you please also update the other backends and the corresponding **init**.py files. Thanks 😄

Hi @vedpatwardhan, jax and tenserflow do not require any changes in the backends to support apple silicon. Rather we have to change the installation process.
Refer: https://developer.apple.com/metal/jax/
           https://developer.apple.com/metal/tensorflow-plugin/

vedpatwardhan (commented):
> > Hey @NripeshN, thanks for adding support for the torch backend, could you please also update the other backends and the corresponding **init**.py files. Thanks 😄
> 
> Hi @vedpatwardhan, jax and tenserflow do not require any changes in the backends to support apple silicon. Rather we have to change the installation process. Refer: https://developer.apple.com/metal/jax/ https://developer.apple.com/metal/tensorflow-plugin/

If that's the case, then probably we can just update the `torch` backend `device` functions where the input passed to the function would still be `gpu:0` and then we'd use `cuda:0` or `mps:0` based on the one available.
This would also ensure that users don't need to explicitly pass `mps` to `device` functions, they'd just pass `gpu` like in the regular case.
Happy to know what you think, thanks @NripeshN 😄

NripeshN:
> If that's the case, then probably we can just update the `torch` backend `device` functions where the input passed to the function would still be `gpu:0` and then we'd use `cuda:0` or `mps:0` based on the one available. This would also ensure that users don't need to explicitly pass `mps` to `device` functions, they'd just pass `gpu` like in the regular case. Happy to know what you think, thanks @NripeshN 😄

I believe it's essential to maintain the option for users to select the MPS device if they want. From my understanding, older MacBooks can utilize CUDA when connected to an eGPU. Moreover, considering MPS is compatible with older Macs equipped with AMD GPUs, and if we're broadening support to include AMD, it seems important to allow users the freedom to opt for their preferred accelerator. What's your take on this, @vedpatwardhan? 😄

vedpatwardhan:
> > If that's the case, then probably we can just update the `torch` backend `device` functions where the input passed to the function would still be `gpu:0` and then we'd use `cuda:0` or `mps:0` based on the one available. This would also ensure that users don't need to explicitly pass `mps` to `device` functions, they'd just pass `gpu` like in the regular case. Happy to know what you think, thanks @NripeshN 😄
> 
> I believe it's essential to maintain the option for users to select the MPS device if they want. From my understanding, older MacBooks can utilize CUDA when connected to an eGPU. Moreover, considering MPS is compatible with older Macs equipped with AMD GPUs, and if we're broadening support to include AMD, it seems important to allow users the freedom to opt for their preferred accelerator. What's your take on this, @vedpatwardhan? 😄

Sure that's a good point, but then how would that work for tensorflow and jax? Because I thought with tensorflow and jax, the device name would still be `GPU` and it would use whichever GPU is available 😅
In case that's the case, I think that's a cleaner approach that helps avoid ambiguity of creating separate devices for nvidia, amd and apple gpus and instead the same device name should use whichever kind of gpu is available 😄

vedpatwardhan (changes requested):
Just pushed some fixes to `device.py` and `random.py`, could you please also pull the latest code from `main` into your branch? Given that there were some compiler test issues unrelated to your changes which have now been fixed, we'd get a better idea of whether there are any failures due to the changes made.
Thanks @NripeshN 😄

NripeshN:
> Just pushed some fixes to `device.py` and `random.py`, could you please also pull the latest code from `main` into your branch? Given that there were some compiler test issues unrelated to your changes which have now been fixed, we'd get a better idea of whether there are any failures due to the changes made. Thanks @NripeshN 😄

Hi @vedpatwardhan, I have pulled the latest code from `main` into my branch

vedpatwardhan:
ivy-gardener

vedpatwardhan:
ivy-gardener

vedpatwardhan (approved):
lgtm! Feel free to merge as none of the [failures](https://github.com/unifyai/ivy/actions/runs/5937532524/job/16100631573?pr=22057#step:4:12) seem to be related to the changes made. Thanks @NripeshN 😄

PR #22064, opened 2023-08-17
mode

Sarvesh-Kesharwani:
Closes: #22040


abdulasiraj (changes requested):
Please update the branch. I think `mode` implementation already exists.
Thanks!

PR #22077, opened 2023-08-17
Fixed typos in deep dive docs

lucasalavapena:
I also added the missing function name (`issubclass`) in the arrays code example, otherwise the check is a bit odd in my opinion.

### DIFF
--- a/docs/overview/deep_dive/ivy_tests.rst
+++ b/docs/overview/deep_dive/ivy_tests.rst
@@ -187,7 +187,7 @@ Suppose you need to generate a 1-D array or a scaler value, which also generate 
 
 we can then later use this strategy in any of our tests.
 
-Writing Hypothesis Test
+Writing Hypothesis Tests
 ^^^^^^^^^^^^^^^^^^^^^^^
 
 Writing Hypothesis tests are intuitive and simple, as an example, we've implemented our own :code:`add` function, which takes in 2 parameters :code:`x` and :code:`y`.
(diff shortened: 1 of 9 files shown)


docs/overview/deep_dive/ivy_tests.rst
```diff
@@ -354,7 +354,7 @@ The test then runs on the newly created arrays with specified data types.
 Why do we need helper functions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-It is usually the case that any ivy function should run seamlessly on ‘all the possible varieties, as well as the edge cases’ encountered by the following parameters -:
+It is usually the case that any ivy function should run seamlessly on `all the possible varieties, as well as the edge cases` encountered by the following parameters -:
```
KareemMAX, line 357: Why do you think that is better?

KareemMAX (changes requested):
Hey @lucasalavapena,
Great job, made some few comments

docs/overview/deep_dive/ivy_tests.rst
lucasalavapena: I thought that is what they were going for, and it was an accident that they did not italise it. What are your thoughts?

KareemMAX: I guess they were going for `'`, but they may have used a word processor that stylized the quotes. I like the stylized version more and see no need to remove it

KareemMAX (approved):

PR #22080, opened 2023-08-17
Moving gc modules to ivy

RickSanchezStoic:
Moves the to_$(framework)_module functions from gc to ivy's stateful converters


RickSanchezStoic:
@guillesanbri @mattbarrett98 This is the first part of moving these functions to ivy's repo, made the corresponding changes in the gc's ivy submodule along with a minor change in a private function and the tests pass, let me know what you guys think, I will be making another pr on the gc one highlighting the changes there.

vedpatwardhan (approved):
lgtm! Let's see what @guillesanbri and @mattbarrett98 think. Thanks @RickSanchezStoic 😄

mattbarrett98 (commented):
I think it's quite problematic copy pasting over the helpers like `_to_native`, `array_to_new_backend` etc duplicating them across the repos, can you try and refactor such that ideally it's only the functions like `to_torch_module` being moved?

RickSanchezStoic:
> I think it's quite problematic copy pasting over the helpers like `_to_native`, `array_to_new_backend` etc duplicating them across the repos, can you try and refactor such that ideally it's only the functions like `to_torch_module` being moved?

Might be difficult to do since these functiosn aren't accessible from within ivy. How does moving these functions from compiler repo to ivy repo entirely sound like? I don't see any hard compiler dependencies here and these functions might come handy in ivy too. cc: @vedpatwardhan

vedpatwardhan (commented):
> > I think it's quite problematic copy pasting over the helpers like `_to_native`, `array_to_new_backend` etc duplicating them across the repos, can you try and refactor such that ideally it's only the functions like `to_torch_module` being moved?
> 
> Might be difficult to do since these functiosn aren't accessible from within ivy. How does moving these functions from compiler repo to ivy repo entirely sound like? I don't see any hard compiler dependencies here and these functions might come handy in ivy too. cc: @vedpatwardhan

In fact, looking at the code again, `array_to_new_backend` and `nest_array_to_new_backend` seems to be redundant as we can implement the same thing by setting the `dynamic_backend` attribute of the container of weights correctly, except for the frontend array checks.
So I think even if we don't add these helpers for the converters, we can refactor it in a way that we just move the converters and it would work 😄

mattbarrett98:
> > > I think it's quite problematic copy pasting over the helpers like `_to_native`, `array_to_new_backend` etc duplicating them across the repos, can you try and refactor such that ideally it's only the functions like `to_torch_module` being moved?
> > 
> > 
> > Might be difficult to do since these functiosn aren't accessible from within ivy. How does moving these functions from compiler repo to ivy repo entirely sound like? I don't see any hard compiler dependencies here and these functions might come handy in ivy too. cc: @vedpatwardhan
> 
> Hey @RickSanchezStoic, looking at the code again `array_to_new_backend` and `nest_array_to_new_backend` seem to be redundant as we can implement the same thing by setting the `dynamic_backend` attribute of the container of weights correctly, except for the frontend array checks. So I think even if we don't add these helpers for the converters, we can refactor it in a way that we just move the converters and it would work 😄

Something to keep in mind with this- looks like the `dynamic_backend` works by using `numpy` as an intermediary for the conversion, which can sometimes lead to copies being made, increasing memory usage. @tomatillos has been working on using dlpack for conversion to ensure no copies are made

> > > I think it's quite problematic copy pasting over the helpers like `_to_native`, `array_to_new_backend` etc duplicating them across the repos, can you try and refactor such that ideally it's only the functions like `to_torch_module` being moved?
> > 
> > 
> > Might be difficult to do since these functiosn aren't accessible from within ivy. How does moving these functions from compiler repo to ivy repo entirely sound like? I don't see any hard compiler dependencies here and these functions might come handy in ivy too. cc: @vedpatwardhan
> 
> Hey @RickSanchezStoic, looking at the code again `array_to_new_backend` and `nest_array_to_new_backend` seem to be redundant as we can implement the same thing by setting the `dynamic_backend` attribute of the container of weights correctly, except for the frontend array checks. So I think even if we don't add these helpers for the converters, we can refactor it in a way that we just move the converters and it would work 😄

One thing to consider is that `dynamic_backend` seems to use `numpy` as an intermediary when doing the conversion which can lead to copies being made, using more memory. This might undo some of the work @tomatillos is doing currently.

I feel like maybe it's possible to do the conversion to the new backend somewhere in the gc code, as opposed to in the call methods of the transpiled modules

mattbarrett98:
> > > I think it's quite problematic copy pasting over the helpers like `_to_native`, `array_to_new_backend` etc duplicating them across the repos, can you try and refactor such that ideally it's only the functions like `to_torch_module` being moved?
> > 
> > 
> > Might be difficult to do since these functiosn aren't accessible from within ivy. How does moving these functions from compiler repo to ivy repo entirely sound like? I don't see any hard compiler dependencies here and these functions might come handy in ivy too. cc: @vedpatwardhan
> 
> Hey @RickSanchezStoic, looking at the code again `array_to_new_backend` and `nest_array_to_new_backend` seem to be redundant as we can implement the same thing by setting the `dynamic_backend` attribute of the container of weights correctly, except for the frontend array checks. So I think even if we don't add these helpers for the converters, we can refactor it in a way that we just move the converters and it would work 😄

One thing to consider is that dynamic_backend seems to use numpy as an intermediary when doing the conversion which can lead to copies being made, using more memory. This might undo some of the work @tomatillos is doing currently.

I feel like maybe it's possible to do the conversion to the new backend somewhere in the gc code, as opposed to in the call methods of the transpiled modules

vedpatwardhan:
> One thing to consider is that dynamic_backend seems to use numpy as an intermediary when doing the conversion which can lead to copies being made, using more memory. This might undo some of the work @tomatillos is doing currently.
> 
> I feel like maybe it's possible to do the conversion to the new backend somewhere in the gc code, as opposed to in the call methods of the transpiled modules

That's a great point, yeah we need to update the dynamic backend setting logic in the backend handler to instead operate on an array by array basis so that we create one copy at a time which goes out of scope before we attempt conversion of the second array. But as far as dynamic backend setting in containers goes, that's already on a [leaf-by-leaf](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/data_classes/container/base.py#L4172) basis wherein the numpy array created and the array in the older backend would go out of scope when we exit the [setter](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/data_classes/array/array.py#L186) for each array. As a result, I think we should still be able to correctly do that conversion without consuming a lot of space, but I'll also fix the dynamic backend logic in the [backend handler](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/utils/backend/handler.py#L431) to deal with this by trading off a bit of runtime efficiency. Thanks for flagging @mattbarrett98 😄

mattbarrett98:
> > One thing to consider is that dynamic_backend seems to use numpy as an intermediary when doing the conversion which can lead to copies being made, using more memory. This might undo some of the work @tomatillos is doing currently.
> > I feel like maybe it's possible to do the conversion to the new backend somewhere in the gc code, as opposed to in the call methods of the transpiled modules
> 
> That's a great point, yeah we need to update the dynamic backend setting logic in the backend handler to instead operate on an array by array basis so that we create one copy at a time which goes out of scope before we attempt conversion of the second array. But as far as dynamic backend setting in containers goes, that's already on a [leaf-by-leaf](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/data_classes/container/base.py#L4172) basis wherein the numpy array created and the array in the older backend would go out of scope when we exit the [setter](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/data_classes/array/array.py#L186) for each array. As a result, I think we should still be able to correctly do that conversion without consuming a lot of space, but I'll also fix the dynamic backend logic in the [backend handler](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/utils/backend/handler.py#L431) to deal with this by trading off a bit of runtime efficiency. Thanks for flagging @mattbarrett98 😄

these are native modules so I think actually the parameters are probably native arrays (and should be), so I guess dynamic_backend for ivy arrays may not be helpful here

mattbarrett98:
> > these are native modules so I think actually the parameters are probably native arrays (and should be), so I guess dynamic_backend for ivy arrays may not be helpful here
> 
> Actually I was referring to [these](https://github.com/unifyai/ivy/pull/22080/files#diff-058e64c94457a8d6db95f0d9d62938deea3c8563acb825e614614b3a2ddb7bc0R829) instances where `self._ivy_module` is an `ivy.Module` which ideally should have a `ivy.Container` of `ivy.Array` in it's leaves 😅 If that's the case, then we could just replace [these](https://github.com/unifyai/ivy/pull/22080/files#diff-058e64c94457a8d6db95f0d9d62938deea3c8563acb825e614614b3a2ddb7bc0R829-R834) lines by,
> 
> ```
> self._ivy_module.v.dynamic_backend = True
> self._ivy_module.v = self._ivy_module.v.cont_map(
>         lambda x, kc: _to_native(x, inplace=True)
> )
> ```
> 
> But if it instead contains native arrays then as you mentioned, we should just try and see if we can handle these conversions in the gc code itself rather than moving the helpers to `ivy` because I suppose that also provides more flexibility of updation of these helpers without updating `ivy`
> 
> Happy to know what you think @mattbarrett98 @RickSanchezStoic, thanks for all the insights uptil now 😄 👍

not sure but I imagine it's the case that the parameters of the ivy module in these cases will be native arrays, since they are going to be fed to a native graph, so they need to be native arrays. But yea I agree we should keep any of this graph compiler logic in the gc repo

vedpatwardhan:
> not sure but I imagine it's the case that the parameters of the ivy module in these cases will be native arrays, since they are going to be fed to a native graph, so they need to be native arrays. But yea I agree we should keep any of this graph compiler logic in the gc repo

But ideally an `ivy.Module` should always have `ivy.Array` instances in it's `v` right which can then be converted to native if we want to use them with a native module? probably that's why [this](https://github.com/unifyai/ivy/pull/22080/files#diff-058e64c94457a8d6db95f0d9d62938deea3c8563acb825e614614b3a2ddb7bc0R829) line explicitly converts the variables to native? Thanks @mattbarrett98 😄

RickSanchezStoic:
> > not sure but I imagine it's the case that the parameters of the ivy module in these cases will be native arrays, since they are going to be fed to a native graph, so they need to be native arrays. But yea I agree we should keep any of this graph compiler logic in the gc repo
> 
> But ideally an `ivy.Module` should always have `ivy.Array` instances in it's `v` right which can then be converted to native if we want to use them with a native module? probably that's why [this](https://github.com/unifyai/ivy/pull/22080/files#diff-058e64c94457a8d6db95f0d9d62938deea3c8563acb825e614614b3a2ddb7bc0R829) line explicitly converts the variables to native? Thanks @mattbarrett98 😄

Let me know what you think @mattbarrett98 , accessing these new_backend functions through compiler would be difficult in this context, hacky at least.

mattbarrett98 (commented):
I don't think anything has changed so not much to review.. we still need to try and remove the need for duplicating code across repos. I just tried removing any use of `array_to_new_backend` in `to_torch_module` etc and all of the module tests still passed so maybe we can simply remove its use, have you looked into this?

RickSanchezStoic:
@mattbarrett98 @vedpatwardhan, I removed the usage of `array_to_new_backend` and all tests passed, also updated the _to_native to include check for frontends,

ivy/stateful/converters.py
```diff
@@ -9,6 +9,9 @@
 
 from ivy.utils.backend import current_backend
 
+# helpers
+frontend_arrays = []
```
mattbarrett98, line 13: this can be removed I think

mattbarrett98 (commented):
for simplicity and a slight speed up the frontend checks can instead be done with `hasattr(x, "ivy_array")` and `hasattr(x, "ivy_shape")`. apart from this if all the gc tests are working then we have a winner! (did you check these tests by making changes locally in the ivy submodule of the gc?)

RickSanchezStoic:
> for simplicity and a slight speed up the frontend checks can instead be done with `hasattr(x, "ivy_array")` and `hasattr(x, "ivy_shape")`. apart from this if all the gc tests are working then we have a winner! (did you check these tests by making changes locally in the ivy submodule of the gc?)

I just copied the code over to the ivy repo. This causes one of the errors in graph compiler code, the prior one causes just assertion errors which we discussed can be skipped. So not sure. Let me know otherwise @mattbarrett98

ivy/data_classes/array/conversions.py
```diff
@@ -17,23 +17,42 @@
 # --------#
 
 
-def _to_native(x: Any, inplace: bool = False, to_ignore: tuple = ()) -> Any:
```
vedpatwardhan, line 20: let's add the `Any` typehints back as it doesn't affect the changes made

ivy/data_classes/array/conversions.py
```diff
@@ -17,23 +17,42 @@
...
-    elif type(x) is ivy.Shape:
+    elif isinstance(x, ivy.Shape):
         return x.shape
+    elif is_frontend_array(x):
+        return x.ivy_array.data
+    elif is_frontend_shape(x):
+        return x.ivy_shape.shape
     elif isinstance(x, ivy.Container):
         return x.cont_map(
-            lambda x_, _: _to_native(x_, inplace=inplace, to_ignore=to_ignore),
```
vedpatwardhan, line 31: Also, let's pass the `to_ignore` argument in the recursive call

vedpatwardhan (changes requested):
Just requested a couple of changes, thanks @RickSanchezStoic 😄

mattbarrett98 (commented):
I just made the same changes in the gc repo (to the frontend object detection fns) and works fine. Not too big of a deal though, so I'm happy with the changes 😁

RickSanchezStoic:
> I just made the same changes in the gc repo (to the frontend object detection fns) and works fine. Not too big of a deal though, so I'm happy with the changes 😁

Excellente! Will add these then.

ivy/data_classes/array/conversions.py
vedpatwardhan: This one

vedpatwardhan (commented):
Hey @RickSanchezStoic, could you please make the changes requested from the last review? Thanks 😄

ivy/data_classes/array/conversions.py
RickSanchezStoic: It 's already there right ?

ivy/data_classes/array/conversions.py
vedpatwardhan: thanks!

ivy/data_classes/array/conversions.py
vedpatwardhan: yep it is, thanks!

vedpatwardhan (approved):
lgtm! Certainly shouldn't break any existing code in to_native and as @mattbarrett98 has already tested the converters, we can get it merged once @guillesanbri approves. Thanks @RickSanchezStoic 😄

PR #22081, opened 2023-08-17
Fixed some incorrect code examples in the ivy_tests deep dive

lucasalavapena:
Note that without adding the new line after code-block:: python for `@handle_test`, the code snippet would not show up. I assume for the other two changes the code was simply not run, because it was not even valid python and in the other it printed 3 types even though the range loop when up to 2 (and range is not inclusive).


KareemMAX (approved):
LGTM! Merging now

PR #22103, opened 2023-08-17
Reformatting nested_map()

ronniepiku:
nested_map() function is reformatted to use def instead of lambda functions. This starts on line 1006. Also updates to function arguments and docstrings.


ivy-seed:

PR #22152, opened 2023-08-18
feat(frontends): added the std method to paddle frontend

### LINKED ISSUE #22151: ivy.functional.frontends.paddle.tensor.tensor in std method
### Bug Explanation

there a method called std in ivy.functional.frontends.paddle.tensor.tensor is only implemented without the test function. and it has wrong implementation.

```python 
@with_unsupported_dtypes({"2.4.2 and below": ("float32", "float64")}, "paddle")
    def std(self, axis=None, unbiased=True, keepdim=False, name=None):
        return paddle_frontend.Tensor(
            ivy.std(self._ivy_array, axis=axis, keepdims=keepdim)
        )

```
it has previous paddle version and also the wrong supported dtype.

### Steps to Reproduce Bug

go to the ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_tensor.py this file location and write the test function.

it is not passing any case.

### Environment

wsl2 , ubuntu20.04 , vscode 

### Ivy Version

latest 

### Backend

- [X] NumPy
- [X] TensorFlow
- [X] PyTorch
- [X] JAX

### Device

_No response_

samthakur587:
close #22151


samthakur587:
hey @mohame54 can you please check this PR.I am not getting any error in local by testing it.

samthakur587:
hey @hmahmood24 can you please review the PR. and let me know whether any changes required or not. So i can move to the next task. 🚀

ivy/functional/frontends/paddle/tensor/tensor.py
```diff
@@ -643,10 +643,12 @@ def atan(self, name=None):
...
         return ivy.atanh(self._ivy_array)
 
-    @with_unsupported_dtypes({"2.4.2 and below": ("float32", "float64")}, "paddle")
+    @with_supported_dtypes({"2.5.1 and below": ("float32", "float64")}, "paddle")
     def std(self, axis=None, unbiased=True, keepdim=False, name=None):
         return paddle_frontend.Tensor(
-            ivy.std(self._ivy_array, axis=axis, keepdims=keepdim)
+            ivy.std(
+                self._ivy_array, axis=axis, correction=int(unbiased), keepdims=keepdim
+            )
```
hmahmood24, line 651: There is a `std` function in `paddle_frontend` now. We should use that here instead of `ivy.std`. The resulting code would look something like:
```python
def std(self, axis=None, unbiased=True, keepdim=False, name=None):
        return paddle_frontend.std(self, axis=axis, unbiased=unbiased, keepdim=keepdim)
```

hmahmood24 (changes requested):
Hey @samthakur587, thanks a lot for the spot and this PR! I have suggested some minor changes. Let me know once you're done with them. Thanks! :)

ivy/functional/frontends/paddle/tensor/tensor.py
samthakur587: done , made the changes required i think now its ready to merge. 🚀 

thank you

hmahmood24 (changes requested):
@samthakur587 The tests are [failing](https://github.com/unifyai/ivy/actions/runs/6101602581/job/16558275942?pr=22152). I think because you're using a strategy from some other module to generate the inputs for your function which doesn't allow you to generate the `valid` dtypes that your function is valid for.

samthakur587:
hey @hmahmood24  i have tried to debug the ivy.std and  i changed the composite function __statistical_dtype_values the helper dtype to ('valid') then all the backend are not passing bcz i have seen that the std at the paddle stat fronted include the unit8 and same as tensorflow and numpy so they returned the error as arrestion with dtype .

i am not found the actual bug that why it taking all float16 even i taken the valid dtype. can you please check.

image

hmahmood24 (changes requested):
Hey @samthakur587, thanks for the update. Can you try removing the `with_supported_dtypes` decorator altogether and see if that works for you? Since this `std` method calls the `paddle_frontend.std` function which already has the `with_supported_dtypes` decorator applied to it [here](https://github.com/unifyai/ivy/blob/5985504d7cffabf7398f8307587d7dcc16e8e86d/ivy/functional/frontends/paddle/stat.py#L55-L58). Also, no need to change the ` _statistical_dtype_values` since that is being used in numerous places so we shouldn't really be modifying it

samthakur587:
I tried to check the ivy.std function. by running the tests at  pytest -v ivy_tests/test_ivy/test_functional/test_core/test_statistical.py::test_std. it also falling in all the backend so i added the unsupported_dtype("bfloat","float16") after that all the backend passed but the numpy is not passing even i have also applied the same decortor to numpy backend.

```python
@_scalar_output_to_0d_array
@with_unsupported_dtypes({"1.26.0 and below": ("bfloat16","float16")}, backend_version)
def std(
    x: np.ndarray,
    /,
    *,
    axis: Optional[Union[int, Sequence[int]]] = None,
    correction: Union[int, float] = 0.0,
    keepdims: bool = False,
) -> np.ndarray:
    return ivy.astype(np.std(x, axis=axis, ddof=correction, keepdims=keepdims, out=out),x.dtype,copy=False)


std.support_native_out = True
```

I don't konw but numpy still supporting the bfloat16 dtype. in the ivy.std also the numpy is not passing and returning the same error bfloat16 not supporting. here i have atteched the result.

image

image

samthakur587:
Hi! @hmahmood24 

Now its passing all the test with all the backend. and ready to merge 🚀 

image

hmahmood24 (changes requested):
Hey @samthakur587, looks good now. Can you just fix the [lint](https://github.com/unifyai/ivy/actions/runs/6247380546/job/16959840970?pr=22152) issues with the PR so we can go ahead and merge it? Thanks!

hmahmood24 (approved):
@samthakur587 Everything looks good to merge. Thanks! 🚀

PR #22204, opened 2023-08-19
Add sum ndarray numpy frontend

niixxaaa:
Close https://github.com/unifyai/ivy/issues/22203


niixxaaa:
ivy-gardener

NripeshN:
Hi @niixxaaa, Thanks for contributing but your code seems to be failing locally. Please fix the tests.

AnnaTz:
> Hi @niixxaaa, Thanks for contributing but your code seems to be failing locally. Please fix the tests.

Actually it seems that the corresponding frontend function `numpy_frontend.sum` is not working as expected, so the errors are not caused by the method's code the contributor is adding. I think we can merge this as is in this case.

PR #22235, opened 2023-08-20
Scipy tests

umairjavaid:
Fixed the scipy key error when getting the dtypes, also fixed the scipy func wrappings

### DIFF
--- a/ivy/functional/frontends/scipy/linalg/linalg.py
+++ b/ivy/functional/frontends/scipy/linalg/linalg.py
@@ -1,6 +1,6 @@
 # global
 import ivy
-from ivy.functional.frontends.scipy.func_wrapper import (
+from ivy.functional.frontends.numpy.func_wrapper import (
     to_ivy_arrays_and_back,
 )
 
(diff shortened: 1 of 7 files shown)


sherry30 (changes requested):
The changes in the get_dtypes strategy looks good.
Can you update your branch, the addition of the config file is not needed since the refactored version is already there on the upstream.
Its best if @AnnaTz reviews the changes in the scipy wrappers and test files.

thanks

ivy/functional/frontends/scipy/linalg/linalg.py
```diff
@@ -1,6 +1,6 @@
 # global
 import ivy
-from ivy.functional.frontends.scipy.func_wrapper import (
+from ivy.functional.frontends.numpy.func_wrapper import (
     to_ivy_arrays_and_back,
```
AnnaTz, line 4: It would be better practice if you imported the numpy function wrapper into the scipy `func_wrapper` file and used that in the scipy function files.

AnnaTz (changes requested):

PR #22277, opened 2023-08-21
update pr-lint-bot.yml to automatically lint on PR

NripeshN:


KareemMAX (changes requested):
As per our private conversation.

What do you think about making the commit message a variable, which is passed to the `pr-lint-bot` workflow. This way we can make sure that our code is consistant and not to worry about conflicts between repos.

KareemMAX (approved):
LGTM! We can merge once we verify that the workflow in `unifyai/workflows` is working correctly.

Issue #14945, opened 2023-05-05
Implement stateful classes for stateful layers 


zaeemansari70:
`Layers:`

- [ ] max_pool1d
- [ ] max_pool3d
- [ ] avg_pool1d 
- [ ] avg_pool3d  
- [ ] dct 
- [ ] fft 
- [ ] ifft 
- [ ] embedding 
- [ ] interpolate 
- [ ] adaptive_avg_pool1d
- [ ] adaptive_avg_pool2d
- [ ] transformer

sahilempire:
#15058

Ishticode:
- [ ] #15289

mobley-trent:
- #15380

Kalderon-Sheikhman:
Please is thing an opened task? How do I go about attempting to do a PR and work on it?

baberabb:
- [ ]  #15542

H4R5H1T-007:
I want to work on this issue, How can I start working on this? is there any documentation or blog I should look?

mohame54:
#max_pool1d

mohame54:
#16221

mohame54:
#16439

baberabb:
- [ ] #16496

mohame54:
#16532

baberabb:
- [ ] #16665

rubada:
#17431

jakubbober:
[ifft] #17528

so-dipe:
There's just one task left here unassigned. I want to take it, what are the guidelines? The initial comment doesn't say much

shubham1206agra:
- [ ] #17945

ChrisEduS:
#18019

MuhammadNizamani:
#18142

MuhammadNizamani:
@zaeemansari70 can you suggest me the docs for implement the class transformer?

HaiderSultanArc:
[embedding] #19528

Shakil7869:
- [ ] #20823

Aayomis:
#21494

DevBhuyan:
transformer #21810

Younggregs:
#22168

AzhanJaved:
interpolate #22679

arshPratap:
- [ ] #22791

Curiositysavesthecat:
-[ ] #22927

Curiositysavesthecat:
- [ ] #22927

[...]

PR #22288, opened 2023-08-21
Close #21810

### LINKED ISSUE #21810: transformer

DevBhuyan:
Closes https://github.com/unifyai/ivy/issues/21810
https://github.com/unifyai/ivy/issues/21810

From ToDo list: https://github.com/unifyai/ivy/issues/14945
https://github.com/unifyai/ivy/issues/14945

Update:

Added transformer layer in ~/ivy/ivy/stateful/layers.py
Added test test_transformer_layer in ~/ivy/ivy_tests/test_ivy/test_stateful/test_layers.py
Added custom composite strategy to generate data (ref. transformer_data())
This is my first PR. I tried to make it as appropriate as possible. Please let me know if there are any modifications that you'd suggest. Thank you!😊

### DIFF
--- a/ivy/stateful/layers.py
+++ b/ivy/stateful/layers.py
@@ -5,6 +5,11 @@ import ivy
 from ivy.func_wrapper import handle_nestable
 from ivy.stateful.initializers import GlorotUniform, Zeros
 from ivy.stateful.module import Module
+from ivy.stateful.norms import LayerNorm
+import copy
+from typing import Optional, Any, Union, Callable
+import warnings
+from ivy import random_uniform as xavier_uniform_
 
 # ToDo: update docstrings and typehints according to ivy\layers
 
@@ -147,7 +152,8 @@ class Dropout(Module):
         """
         self._prob = prob
         self._scale = scale
-        Module.__init__(self, device=None, v=None, dtype=dtype, training=training)
+        self.training = training
+        Module.__init__(self, device=None, v=None, dtype=dtype)
 
     def _create_variables(self, device, dtype=None):
         """
@@ -209,7 +215,6 @@ class MultiHeadAttention(Module):
         v=None,
         build_mode="on_init",
         dtype=None,
-        training=True,
     ):
         """
         Multi Head Attention layer.
@@ -258,8 +263,6 @@ class MultiHeadAttention(Module):
         dtype
             the desired data type of the internal variables to be created if not provided.
             Default is ``None``.
-        training
-            If True, dropout is used, otherwise dropout is not activated.
         """
         # proj
 
@@ -287,7 +290,6 @@ class MultiHeadAttention(Module):
             build_mode=build_mode,
             with_partial_v=True,
             dtype=dtype,
-            training=training,
         )
 
     def _create_variables(self, device, dtype=None):
@@ -374,6 +376,7 @@ class MultiHeadAttention(Module):
         is_causal=False,
         return_attention_weights=False,
         average_attention_weights=True,
+        training=False,
     ):
         """
         Perform forward pass of the MultiHeadAttention layer.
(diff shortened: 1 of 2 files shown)


rishabgit (commented):
Hi, thanks for looking into this @DevBhuyan 😄 

You'll have to rewrite the syntax a bit so that it is in line with general convention. I'll recommend checking out other classes in that file.

Not super sure if the implementation is correct - it should ideally be similar to Pytorch's Transformer module - https://pytorch.org/docs/stable/generated/torch.nn.Transformer.html#torch.nn.Transformer https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/transformer.py 
and not creating a model with positional encodings pre-coded like this official Pytorch example making use of Transformer module - https://github.com/pytorch/examples/blob/13009eff7a80ebcf6ae89ed217d5d176bd3e019d/word_language_model/model.py#L107

@vedpatwardhan - do you think the above makes sense or @DevBhuyan's implementation is fine as it is? 🤔

DevBhuyan:
Hi @rishabgit, Thanks for taking the time to review my PR. Yes, you're right, I need to change the syntax to fit with the other classes. My previous code was also just an attempt to write the layer as simply as possible (autoregressive, without a decoder), I guess I'll need to modify it to an encoder-decoder architecture (like that of PyTorch's).

I'll work on it and make a commit as soon as I have it ready. Thanks again for the suggestions:)

vedpatwardhan:
> Hi, thanks for looking into this @DevBhuyan 😄
> 
> You'll have to rewrite the syntax a bit so that it is in line with general convention. I'll recommend checking out other classes in that file.
> 
> Not super sure if the implementation is correct - it should ideally be similar to Pytorch's Transformer module - https://pytorch.org/docs/stable/generated/torch.nn.Transformer.html#torch.nn.Transformer https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/transformer.py and not creating a model with positional encodings pre-coded like this official Pytorch example making use of Transformer module - https://github.com/pytorch/examples/blob/13009eff7a80ebcf6ae89ed217d5d176bd3e019d/word_language_model/model.py#L107
> 
> @vedpatwardhan - do you think the above makes sense or @DevBhuyan's implementation is fine as it is? 🤔

Hey @rishabgit, your suggestion makes perfect sense, we should definitely try and align with the `torch.nn.Transformer` module as closely as possible. I'm not sure why we wouldn't want to create a model with pre-coded positional encodings, do you mean we directly do a `self.register_buffer` instead to ensure that it's a non-trainable state variable? Thanks @rishabgit @DevBhuyan 😄

DevBhuyan:
Hi @vedpatwardhan, I guess I was assuming a totally different direction, a decoder-only transformer. I agree with @vedpatwardhan and @rishabgit, and I'm rewriting it entirely to be in line with Pytorch's implementation and the other classes from `layers.py`. 

Please excuse the unforeseen closing and opening of this Pull Request:) I'm still new to contributing on GitHub, I accidentally tried to remove another branch and this happened.

DevBhuyan:
Hi @rishabgit, I have updated the Transformer class to fit in line with PyTorch's implementation as well as with the other classes in the `layers.py` file. The docstrings also have been edited to match with the other classes. 

Since I had used PyTorch's implementation as a starting point for the class(es), There were portions in PyTorch's implementation that required lower-level backend implementation to speed up processes ('FlashAttention'). I have not totally removed those lines, instead, I commented them out with `FIXME` tags. If required, these can be incorporated later into Ivy's backend itself to speed up processing.

Kindly let me know if there are any changes you'd suggest.

Thank you for your patience:)

DevBhuyan:
I guess I messed up this PR. I'll create a new fork of the repo and create a fresh PR

PR #22305, opened 2023-08-21
Add inplace update modes and generate a warning when setting a `jax` or `tensorflow` backend

Madjid-CH:
the PR is for this [task](https://trello.com/c/36EfB0LY)


vedpatwardhan (commented):
Let's move the warning logic to the `__init__.py` instead of `inplace_update` as discussed, feel free to request a review once you're done. Thanks 😄

ivy/functional/backends/tensorflow/__init__.py
```diff
@@ -148,6 +150,45 @@
 
 native_inplace_support = False
 
+backend = "tensorflow"
+
+
+def _handle_inplace_mode():
```
vedpatwardhan, line 156: Given that throwing the warning in the `__init__.py` is leading to code duplication, let's just generate the warning in the backend handler instead as you were doing before, with a warning generated from [`set_backend`](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/utils/backend/handler.py#L396), [`previous_backend`](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/utils/backend/handler.py#L526) and [`with_backend`](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/utils/backend/handler.py#L623) with this helper being moved to [`exceptions.py`](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/utils/exceptions.py).

ivy/functional/backends/tensorflow/__init__.py
```diff
@@ -148,6 +150,45 @@
...
+    if not hasattr(ivy, "inplace_mode") or ivy.inplace_mode not in [
+        "lenient",
+        "strict",
+    ]:
+        ivy.set_inplace_mode("lenient")
```
vedpatwardhan, line 161: I think the `inplace_mode_stack` should be empty by default, we can keep `"lenient"` as the default mode as explained in the comments below

ivy/functional/backends/tensorflow/__init__.py
```diff
@@ -148,6 +150,45 @@
...
+
+    if ivy.inplace_mode == "lenient":
+        message = (
```
vedpatwardhan, line 164: I think we can just throw a warning here if the `inplace_mode` is `lenient`, given that we're anyway throwing an exception if it was `strict`, and given that `strict` isn't the default, we can assume the user knows that to be the case. 
Also, lets probably shorten this a bit to,
```
The current backend: '{backend}' does not support inplace updates natively. 
Ivy would quietly create new arrays when using inplace updates with this 
backend, leading to memory overhead (same applies for views). If you want to 
control your memory management, consider doing ivy.set_inplace_mode('strict') 
which should raise an error whenever an inplace update is attempted with this 
backend. 
```

ivy/functional/backends/tensorflow/__init__.py
```diff
@@ -148,6 +150,45 @@
...
+    else:
+        message = (
+            f"The current backend: '{backend}' does not support "
+            "inplace updates natively for arrays but does for variables. "
+            "And since the current ivy.inplace_mode is set to 'strict', "
+            "Ivy will throw an error "
+            "whenever an inplace update is attempted."
+        )
+    warnings.warn(message)
+    warnings.warn(
```
vedpatwardhan, line 182: Let's include the information about views in the same warning as mentioned above

ivy/functional/backends/tensorflow/general.py
```diff
@@ -290,6 +291,16 @@ def inplace_update(
         return val
 
 
+def _check_inplace_mode():
```
vedpatwardhan, line 294: I think this error message can be integrated into line [247](https://github.com/unifyai/ivy/pull/22305/files#diff-08657e30ab249a75e71b4d0cb464494dc19ae11c64604c22d61017f988c809b2R247) by adding an additional condition to the `if` block, same applies to the other backends.
Also, we should create a new Exception class for this like `InplaceUpdateException` in `exceptions.py` and throw that instead.

ivy/functional/ivy/general.py
```diff
@@ -3263,6 +3264,53 @@ def inplace_update(
 inplace_update.unsupported_dtypes = {"torch": ("bfloat16",)}
```
vedpatwardhan, line 3266: We should set a default value to the mode like [here](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/functional/ivy/general.py#L379)
We should also be correctly passing the mode correctly to the [`ivy/__init__.py`](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/__init__.py) file [here](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/__init__.py#L937) and [here](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/__init__.py#L1416)

vedpatwardhan (changes requested):
Hey @Madjid-CH, just requested a few changes, thanks 😄

ivy/functional/backends/tensorflow/__init__.py
Madjid-CH: But this is the first time where the user encounter these warnings, so shouldn't we give him all information about inplace modes ?

Madjid-CH:
Hey @vedpatwardhan I noticed that `native_inplace_support = False` in numpy backend `__init__.py` [here](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/functional/backends/numpy/__init__.py#L133). Should it be `True` since numpy support inplace updates natively ?
Thanks

vedpatwardhan (commented):
> Hey @vedpatwardhan, I noticed that `native_inplace_support = False` in numpy backend `__init__.py` [here](https://github.com/unifyai/ivy/blob/8ffd9ffad6f715a3b0bdedb3bbafe52d0085d2e6/ivy/functional/backends/numpy/__init__.py#L133). Should it be `True` since numpy do support inplace updates natively ? another point is that after I modified [test_inplace_update](https://github.com/unifyai/ivy/blob/62d12f72ae60ee93b008dd96720ba5c8633a72ca/ivy_tests/test_ivy/test_functional/test_core/test_general.py#L1245C1-L1245C1) to account for the exception raising. The Jax backend start failing by raising IvyBackendException instead of InplaceUpdateException like the other backends. Please tell me if you have any idea why this is the case.
> 
> Thanks

Good catch! Yeah it should be `True`, thanks @Madjid-CH 😄

ivy/__init__.py
```diff
@@ -1414,7 +1415,7 @@ def cast_data_types(val=True):
     "warning_level",
     "nan_policy",
     "array_mode",
-    "nestable_mode",
+    "inplace_modenestable_mode",
```
vedpatwardhan, line 1418: probably an error introduced while fixing merge conflicts

ivy/functional/backends/jax/__init__.py
```diff
@@ -1,5 +1,7 @@
 # global
 import sys
+import warnings
```
vedpatwardhan, line 3: unused import

ivy/functional/backends/jax/__init__.py
```diff
@@ -178,9 +183,6 @@ def closest_valid_dtype(type=None, /, as_native=False):
     return ivy.as_ivy_dtype(type) if not as_native else ivy.as_native_dtype(type)
 
 
-backend = "jax"
```
vedpatwardhan, line 181: doesn't really make a difference, but let's revert this change

ivy/functional/backends/tensorflow/__init__.py
```diff
@@ -1,6 +1,8 @@
 # global
 import sys
 import logging
+import warnings
```
vedpatwardhan, line 4: unused import

ivy/functional/backends/tensorflow/__init__.py
```diff
@@ -157,9 +162,6 @@ def closest_valid_dtype(type=None, /, as_native=False):
     return ivy.as_ivy_dtype(type) if not as_native else ivy.as_native_dtype(type)
 
 
-backend = "tensorflow"
```
vedpatwardhan, line 160: let's revert this too

ivy/functional/ivy/general.py
vedpatwardhan: We'd also need to add a `unset_inplace_mode` method like the other modes

ivy/utils/assertions.py
```diff
@@ -354,17 +354,34 @@ def check_kernel_padding_size(kernel_size, padding_size):
             )
 
 
+def check_inplace_update_support(ensure_in_backend, x):
```
vedpatwardhan, line 357: let's also make this function private and move it to `exceptions.py` below `handle_inplace_mode`

ivy/utils/assertions.py
```diff
@@ -354,17 +354,34 @@ def check_kernel_padding_size(kernel_size, padding_size):
...
+
+
 def check_dev_correct_formatting(device):
     assert device[0:3] in ["gpu", "tpu", "cpu"]
     if device != "cpu":
         assert device[3] == ":"
         assert device[4:].isnumeric()
 
 
-# Jax Specific #
```
vedpatwardhan, line 364: let's revert this change as `check_dev_correct_formatting` isn't a jax-specific assertion

ivy/utils/backend/handler.py
```diff
@@ -584,8 +587,12 @@ def previous_backend():
                 ivy.__dict__[k] = v
             if k in ivy.functional.__dict__ and not k.startswith("__"):
                 ivy.functional.__dict__[k] = v
+
     if verbosity.level > 0:
         verbosity.cprint("backend stack: {}".format(backend_stack))
+
```
vedpatwardhan, line 593: let's remove the additional blank lines

ivy/utils/backend/handler.py
```diff
@@ -652,4 +662,7 @@ def with_backend(backend: str, cached: bool = True):
         compiled_backends[backend].append(ivy_pack)
     except KeyError:
         compiled_backends[backend] = [ivy_pack]
+
```
vedpatwardhan, line 665: here as well

ivy/utils/exceptions.py
```diff
@@ -1,3 +1,5 @@
+import warnings
+
```
vedpatwardhan, line 2: another blank line to remove

ivy/utils/exceptions.py
```diff
@@ -288,6 +290,11 @@ def __init__(self, message=""):
         super().__init__(message)
 
 
+class InplaceUpdateException(IvyException):
```
vedpatwardhan, line 293: let's move this exception class at the end of the other exception classes and before `handle_exceptions`

ivy/utils/exceptions.py
```diff
@@ -390,6 +397,25 @@ def _handle_exceptions(*args, **kwargs):
...
+        except InplaceUpdateException as e:
+            _configure_stack_trace(e.__traceback__)
+            raise ivy.utils.exceptions.InplaceUpdateException(
+                fn.__name__, str(e), include_backend=True
+            )
 
     _handle_exceptions.handle_exceptions = True
     return _handle_exceptions
+
+
```
vedpatwardhan, line 409: let's add a section named `Inplace Updates` above `handle_inplace_mode`

ivy/utils/exceptions.py
```diff
@@ -390,6 +397,25 @@ def _handle_exceptions(*args, **kwargs):
...
+def handle_inplace_mode():
```
vedpatwardhan, line 410: let's probably also make this private as users shouldn't really need to call this function

ivy_tests/test_ivy/test_functional/test_core/test_general.py
```diff
@@ -1239,6 +1241,25 @@ def test_inplace_update(
...
+    supports_variable_inplace_update = (
+        test_flags.as_variable and ivy_backend.inplace_variables_supported()
+    )
+    return supports_array_inplace_update or supports_variable_inplace_update
+
+
+@pytest.mark.parametrize("mode", ["lenient", "strict"])
+def test_set_inplace_mode(mode):
+    ivy.set_inplace_mode(mode)
+    assert ivy.inplace_mode == mode
```
vedpatwardhan, line 1262: we'd also need a test for `unset_inplace_mode` once it's added

vedpatwardhan (changes requested):
Hey @Madjid-CH, just requested a few changes, could you please also pull from `main`? There's quite a few failures in the CI so pulling would help determine if the failures are related to the changes made. Thanks 😄

ivy/utils/assertions.py
Madjid-CH: this function is used by the inplace_update funtions in tensorflow and jax backend so I can't make it private.
Unless you want me to inline it where I use it ?

ivy/functional/backends/jax/__init__.py
```diff
@@ -167,6 +167,9 @@
 
 native_inplace_support = False
 
+backend = "jax"
```
vedpatwardhan, line 170: let's revert this too, given that we have it already on line [`184`](https://github.com/unifyai/ivy/pull/22305/files#diff-d8dd39b7d69bddbcc5b46052df6d89bf2c571285ffbb13039750c155dae8b27eR184)

ivy/functional/backends/paddle/general.py
```diff
@@ -330,6 +331,8 @@ def inplace_update(
     ensure_in_backend: bool = False,
     keep_input_dtype: bool = False,
 ) -> ivy.Array:
+    _check_inplace_update_support()
```
vedpatwardhan, line 334: we need to pass `ensure_in_backend` and `x`

ivy/utils/exceptions.py
```diff
@@ -390,6 +396,45 @@ def _handle_exceptions(*args, **kwargs):
...
+            "inplace updates natively. Ivy would quietly create new arrays when "
+            "using inplace updates with this backend, leading to memory overhead "
+            "(same applies for views). If you want to control your memory "
+            "management, consider doing ivy.set_inplace_mode('strict') which "
+            "should raise an error whenever an inplace update is attempted "
+            "with this backend."
+        )
+
+
+def _check_inplace_update_support(ensure_in_backend, x):
```
vedpatwardhan, line 426: I think having `x` as the first argument and `ensure_in_backend` as the second argument would be more intuitive

vedpatwardhan (changes requested):
Just requested a few changes, also given that there's quite a few failures in the CI could you please pull from main into your branch to confirm if the failures are related to your changes? Thanks @Madjid-CH 😄

ivy/functional/backends/jax/general.py
```diff
@@ -93,7 +94,7 @@ def set_item(
     ret = x.at[query].set(val)
     if copy:
         return ret
-    return ivy.inplace_update(x, _to_device(ret))
```
vedpatwardhan, line 96: I think we should revert this change as it seems unrelated to our changes

ivy/functional/backends/jax/general.py
```diff
@@ -251,10 +252,8 @@ def inplace_update(
     keep_input_dtype: bool = False,
 ) -> ivy.Array:
     if ivy.is_array(x) and ivy.is_array(val):
-        if ensure_in_backend or ivy.is_native_array(x):
-            raise ivy.utils.exceptions.IvyException(
-                "JAX does not natively support inplace updates"
-            )
+        _check_inplace_update_support(x, ensure_in_backend)
+
```
vedpatwardhan, line 256: let's remove this blank line

ivy/functional/backends/paddle/general.py
```diff
@@ -330,6 +331,8 @@ def inplace_update(
     ensure_in_backend: bool = False,
     keep_input_dtype: bool = False,
 ) -> ivy.Array:
+    _check_inplace_update_support(x, ensure_in_backend)
+
```
vedpatwardhan, line 335: another blank line

ivy/functional/backends/tensorflow/general.py
```diff
@@ -241,10 +242,8 @@ def inplace_update(
     keep_input_dtype: bool = False,
 ) -> ivy.Array:
     if ivy.is_array(x) and ivy.is_array(val):
-        if ensure_in_backend or ivy.is_native_array(x):
-            raise ivy.utils.exceptions.IvyException(
-                "TensorFlow does not support inplace updates of the tf.Tensor"
-            )
+        _check_inplace_update_support(x, ensure_in_backend)
+
```
vedpatwardhan, line 246: another blank line

vedpatwardhan (changes requested):
Requested a final few changes, thanks @Madjid-CH 😄

ivy/functional/backends/jax/general.py
Madjid-CH: there's no reference anywhere to this function `_to_device ` so my IDE flag it as an error, that why I removed it. But I'll revert it

vedpatwardhan (approved):
lgtm! Feel free to merge, I don't think the failures in the CI are related to the changes made, thanks @Madjid-CH 😄

PR #22309, opened 2023-08-22
Added fft2 to Jax frontend 

### LINKED ISSUE #22162: fft2

AbdullahSabry:
Closes #22162

Currently the test doesn't work but I can't figure out why. I keep getting the error `ValueError: setting an array element with a sequence.` I even went as far as to try out the tests already made [here](https://github.com/unifyai/ivy/blob/8584bee085da7e3a2508d73d85a690f9e2068323/ivy_tests/test_ivy/test_functional/test_experimental/test_nn/test_layers.py#L1128). It still gave me the same error.


AbdullahSabry:
Hey @abdulasiraj

abdulasiraj (changes requested):
I think you're using some kinda private function, which are not allowed to use without API key.
```
SystemExit: Please sign up for free pilot access here [https://console.unify.ai/], in order to make use of ivy.compile, ivy.transpile and ivy.unify.
```
So, use the following flag:
`
--skip-compile-testing
`

abdulasiraj (changes requested):
same case like other PR

AbdullahSabry:
@abdulasiraj 

I believe it is ready to merge now along with #22536 to fix the tests.

abdulasiraj:
Once the PR for tensorflow backend is merged than this need to updated in order to get merged.
thanks!

ivy/functional/frontends/jax/numpy/fft.py
```diff
@@ -11,6 +11,15 @@ def fft(a, n=None, axis=-1, norm=None):
     return ivy.fft(a, axis, norm=norm, n=n)
 
 
+@to_ivy_arrays_and_back
+def fft2(a, s=None, axes=(-2, -1), norm=None):
+    if norm is None:
+        norm = "backward"
+    if ivy.dtype(a) == "complex128":
```
abdulasiraj, line 18: I think you can omit the check for dtype as function will only recieve complex types. you can directly do this `return ivy.array(ivy.fft2(a, s=s, dim=axes, norm=norm), dtype=ivy.dtype(a))`

AbdullahSabry: Yes seems right to just make it inclusive of these dtypes, Pushed changes all seems to be working well!

PR #22312, opened 2023-08-22
Update `ivy.Container` methods to prevent copies

leacornelio:


jieunboy0516:
Hi @vedpatwardhan , would you mind taking a look of this PR from @leacornelio please. As I am not familiar with the framework enough to decide these fundamental changes. Thanks 😅

ivy/data_classes/container/base.py
```diff
@@ -1569,7 +1569,7 @@ def _cont_at_key_chains_input_as_seq(self, key_chains, ignore_key_errors=False):
         return return_cont
 
     def _cont_at_key_chains_input_as_dict(
-        self, key_chains, current_chain="", ignore_key_errors=False
+        self, key_chains, current_chain="", ignore_key_errors=False, inplace=False
```
vedpatwardhan, line 1572: Actually we don't need to add the `inplace` argument to the methods which don't have it, we just need to avoid creating copies for the ones that have that argument. Could you please check if the methods that have the `inplace` argument have the issue with creating copies? There's a possibility that someone might've already fixed it on main, but just wanted to confirm if that's the case. Thanks @leacornelio 😄

ivy-seed:

PR #1234, opened 2022-05-29
reformat: unique_counts 

Ya2001:
fixes https://github.com/unifyai/ivy/issues/1232
Edited the unique_counts method to make it easier to read and also added some helpful comments and examples to help
new users understand the purpose and the use cases of this method.

PR #22324, opened 2023-08-22
docs(contributing): clarify ToDo list comments

jshepherd01:
Applicants seem to be confused sometimes about how to properly link their issue to the relevant ToDo list, so this commit adds an example to the docs to hopefully make it more clear


KareemMAX (changes requested):
Hey @jshepherd01,
Looks good to me, but I want you also to mention that you don't really need to add a checklist item. For example you can just comment `#12345`. And actually fun fact, you can comment the issue URL, as in `https://github.com/unifyai/ivy/issues/1234`. Which might be a bit faster for contributors (as it was fast to me).

Can you mention those as well in the doc while we are at it, other than that everything LGTM

KareemMAX (approved):
LGTM! Merging now

PR #22345, opened 2023-08-22
Replacing the remaining use of itertools.product in ivy.get_item/set_item

AnnaTz:


AnnaTz:
This works but takes more time than using `itertools.product`. Confirmed it for small and large arrays.
Made the PR in case the code is needed in the future.

PR #22354, opened 2023-08-22
Update functions that has `copy` argument

Madjid-CH:
this PR is for this [task](https://trello.com/c/cceeXvrr)


ivy/data_classes/array/creation.py
```diff
@@ -5,13 +5,15 @@
...
 import ivy
+from ivy import handle_view
 
 
 # Array API Standard #
 # -------------------#
 
 
 class _ArrayWithCreation(abc.ABC):
+    @handle_view
```
AnnaTz, line 16: @vedpatwardhan Is `ivy.asarray` an edge case? why have a `copy` argument if there is no `view` wrapper? Is `copy` handled somewhere else?

ivy/data_classes/array/experimental/manipulation.py
```diff
@@ -231,6 +229,12 @@ def rot90(
 
         Parameters
         ----------
+        copy
+            boolean indicating whether or not to copy the input array.
+            If True, the function must always copy.
+            If False, the function must never copy.
+            In case copy is False we avoid copying by returning
+             a view of the input array.
```
AnnaTz, line 237: Could you move this to under `self`? We want the docstrings to be in the same order as the arguments.

ivy/data_classes/array/manipulation.py
```diff
@@ -754,13 +747,12 @@ def unstack(
         ----------
         self
             Input array to unstack.
-        copy
+         copy
```
AnnaTz, line 750: This indent is probably a mistake I think.

AnnaTz (changes requested):

ivy/data_classes/array/creation.py
AnnaTz: @Madjid-CH actually please remove the changes from `creation` and `general` files. These are special cases. After that we are good to merge.

Madjid-CH:
Hey @AnnaTz,
I did the requested changes 🙂

AnnaTz (approved):

PR #22357, opened 2023-08-22
fixed bug in jax.numpy.where

### LINKED ISSUE #22356: [Bug]: Incomplete implementation of jax.numpy.where
### Bug Explanation

The function accepts only condition argument in the native function but the implementation of `ivy.frontends.jax.numpy` doesn't follow the same behaviour and only works when all three arguments are passed

### Steps to Reproduce Bug

```
import jax.numpy as jnp
jnp.where(jnp.array([True,False,True,True]))
#(DeviceArray([0, 2, 3], dtype=int32),)

import ivy.functional.frontends.jax.numpy as jnp
jnp.where(jnp.array([True,False,True,True]))
#raises error

```

### Environment

Ubuntu

### Ivy Version

Github

### Backend

- [ ] NumPy
- [ ] TensorFlow
- [ ] PyTorch
- [ ] JAX

### Device

_No response_

ShreyanshBardia:
closes #22356


NripeshN:
ivy-gardener

NripeshN:
LGTM!

PR #22415, opened 2023-08-23
Made softmax supports complex dtypes

mohame54:
Updated the following functions **max**, **exp** `only for paddle` and **softmax** for **tensorflow, torch and paddle**  backends like

### DIFF
--- a/docs/overview/contributing/open_tasks.rst
+++ b/docs/overview/contributing/open_tasks.rst
@@ -128,7 +128,7 @@ At some point, a member of our team will assess whether it should be added, and 
 
 After this, you then have two options for how to proceed:
 
-#. Try to implement the function as a composition of currently present ivy functions, as explained in the "Temporary Compositions" sub-section of the :ref:`Ivy Frontends` guide, and add the :code:`#ToDo` comment in the implementation as explained.
+#. Try to implement the function as a composition of currently present ivy functions, as explained in the :ref:`Short Frontend Implementations` sub-section of the :ref:`Ivy Frontends` guide, and add the :code:`#ToDo` comment in the implementation as explained.
    Once the PR is merged, your sub-task issue will then be closed as normal.
 #. Alternatively, if you do not want to try and implement the frontend function compositionally, or if this is not feasible, then you can simply choose another frontend function to work on.
    You could also choose to work on another open task entirely at this point if you wanted to.


ivy/functional/backends/paddle/elementwise.py
```diff
@@ -1008,6 +1008,13 @@ def log(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.T
 def exp(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
     if x.dtype in [paddle.int32, paddle.int64, paddle.float32, paddle.float64]:
         return paddle.exp(x)
+    elif paddle.is_complex(x):
+        imag = paddle.imag(x)
+        const = paddle.to_tensor(1j, dtype=x.dtype)
+        mag = paddle.cast(paddle.exp(paddle.real(x)), x.dtype)
+        real = paddle.cast(paddle.cos(imag), x.dtype)
+        imag = paddle.multiply(paddle.cast(paddle.sin(imag), x.dtype), const)
+        return paddle.multiply(paddle.add(real, imag), mag)
```
jshepherd01, line 1017: You can refactor this to simplify it quite a bit
```suggestion
    elif paddle.is_complex(x):
        return paddle.multiply(
            paddle.exp(x.real()),
            paddle.complex(paddle.cos(x.imag()), paddle.sin(x.imag())),
        )
```

ivy-leaves:

Thank you :hugs:

ivy/functional/backends/paddle/statistical.py
```diff
@@ -77,9 +77,15 @@ def max(
...
     ]:
         if paddle.is_complex(x):
-            real_part = paddle.amax(x.real(), axis=axis, keepdim=keepdims)
-            imag_part = paddle.amax(x.imag(), axis=axis, keepdim=keepdims)
-            ret = paddle.complex(real_part, imag_part)
+            real = x.real()
+            imag = x.imag()
+            const = paddle.to_tensor(1j, dtype=x.dtype)
+            real_max = paddle.max(real, axis=axis, keepdim=keepdims)
+            imag = paddle.where(real == real_max, imag, paddle.full_like(imag, -1e5))
```
jshepherd01, line 84: I'm not sure I like this method since the maximum imaginary part could still be less than `-1e5` (in `complex64` it does down to about `-1e38`). I also don't really like the idea of defining complex number behaviour for the `max` function since it's technically undefined mathematically.

It seems like the main reason to implement this here is to use it in softmax for better stability, but in that case the imaginary part doesn't really matter. Might be better to revert the change to this file and just put a simple `paddle.max(x.real(), ...)` in the `softmax` function

ivy/functional/backends/torch/statistical.py
```diff
@@ -52,6 +51,16 @@ def max(
...
+    elif "complex" in str(x.dtype):
+        real = torch.real(x)
+        imag = torch.imag(x)
+        const = torch.tensor(1j, dtype=x.dtype)
+        real_max = torch.max(real, dim=axis, keepdim=keepdims).values
+        imag = torch.where(real == real_max, imag, torch.finfo(imag.dtype).min)
+        # we consider the number with the biggest real and imag part
+        img_max = torch.max(imag, dim=axis, keepdim=keepdims).values
+        img_max = img_max.to(x.dtype)
+        return torch.add(real_max.to(x.dtype), torch.multiply(img_max, const))
```
jshepherd01, line 63: And with this one

jshepherd01 (changes requested):

ivy/functional/backends/paddle/statistical.py
mohame54: the imaginary part is important when we the real parts are equal suppose we have 
```python
# in this example the real parts are equal and we define the bigger between these two according to imaginary parts
x = np.array([1-1j, 1+1j])
```

ivy/functional/backends/paddle/elementwise.py
mohame54: okay on it.

ivy/functional/backends/paddle/statistical.py
mohame54: After some reverse engineering I discovered how does **numpy** and **jax** support complex **dtype** as you can see first we take the maximum according to the real part  but as I said before we could encounter something like mentioned above where the real parts are equal so the suggested solution works fine like **numpy** and **jax** and the -1e6 problem I couldn't find a way in paddle to get the minimum number of the **dtype** automatically so I put this number until I find a way to determine the minimum  number in a certain dtype.

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
```diff
@@ -116,16 +116,17 @@ def test_sigmoid(*, dtype_and_x, test_flags, backend_fw, fn_name, on_device):
 @handle_test(
     fn_tree="functional.ivy.softmax",
     dtype_and_x=helpers.dtype_and_values(
-        available_dtypes=helpers.get_dtypes("float"),
+        available_dtypes=helpers.get_dtypes("float_and_complex"),
         min_num_dims=1,
         large_abs_safety_factor=8,
-        small_abs_safety_factor=8,
+        small_abs_safety_factor=4,
```
jshepherd01, line 122: I see you reduced the safety factor, was there a reason for this?

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
```diff
@@ -116,16 +116,17 @@ def test_sigmoid(*, dtype_and_x, test_flags, backend_fw, fn_name, on_device):
...
         safety_factor_scale="log",
     ),
     axis=st.one_of(
         helpers.ints(min_value=-1, max_value=0),
         st.none(),
     ),
+    test_gradients=st.just(False),
```
jshepherd01, line 129: Was this to resolve the error paddle gives about gradients? If so, I think you need to remove it because that's a wider issue with ivy as a whole rather than this particular function

jshepherd01 (changes requested):

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
mohame54: I was trying to solve the errors that appeared and I tried this safety factor and it worked for me

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
mohame54: I changed it already I think I that line was added cuz softmax in torch doesn't support bfloat16 dtype which caused some erros but I  had removed it already.

jshepherd01 (changes requested):
All looks good to me, and seems to pass the tests. Could you please add the `@handle_complex_input` decorator to the functional API (and make the associated docstring/call signature changes) and then it'll be good to merge

mohame54:
okay on it

Issue #22465, opened 2023-08-24
subtract_


he11owthere:
To-do list: https://github.com/unifyai/ivy/issues/15115

PR #22467, opened 2023-08-24
Added subtract_ method to the Paddle frontend

he11owthere:
closes https://github.com/unifyai/ivy/issues/22465

image


he11owthere:
Please add 'bfloat16' in the `with_unsupported_dtypes` decorator before running the tests until the issue with `ivy.Array.repr` method has been fixed as mentioned by Ved in https://github.com/unifyai/ivy/pull/21447. Thank you! 🙂

jieunboy0516:
Hi @he11owthere ! Thanks for your contribution to Ivy! I had a look at your code and it looks good to me! Could you please put `"float32", "float64"` for the unsupported types for now? thanks :)

ivy/functional/frontends/paddle/tensor/tensor.py
```diff
@@ -167,6 +167,13 @@ def erf(self, name=None):
     def subtract(self, y, name=None):
         return paddle_frontend.Tensor(ivy.subtract(self._ivy_array, _to_ivy_array(y)))
 
+    @with_unsupported_dtypes(
+        {"2.5.1 and below": ("float16", "uint8", "int8", "bool")}, "paddle"
```
jieunboy0516, line 171: could you please just leave it as "float32", "float64" for now, thanks

he11owthere:
Hi there @jieunboy0516,
Given that the `subtract_` method carries out an in-place modification on the input tensor, it's necessary to include the following code:
`self.ivy_array = self.subtract(y).ivy_array`

You can observe a similar implementation in other frontends as well. For instance, here's the PyTorch version of `subtract_`:
```
def subtract_(self, other, *, alpha=1):
    self.ivy_array = self.sub(other, alpha=alpha).ivy_array
    return self
```
 
The approach you're suggesting would indeed work well if the method were named `subtract` instead of `subtract_`. 
Please feel free to correct me if I'm mistaken, drawing from my experiences with similar PRs. It seems that the `add_` and `float_` methods might not be implemented accurately in this case.
Your insights are much appreciated. Thank you! 🙂

ivy/functional/frontends/paddle/tensor/tensor.py
he11owthere: I have tested on my local machine by importing paddle and noticed that PyTorch's `subtract` (also `subtract_`) method supports every other data type including `["int16", "int32", "int64", "bfloat16", "float32", "float64", "complex64", "complex128"]` and does not support `["float16", "uint8", "int8", "bool"]` so that's why I have included them in `@with_unsupported_dtypes` decorator.
 
So I suggest to make the above method exactly similar to the PyTorch's `subtract_` method.
Thank you very much! 😊

jieunboy0516:
Hi @he11owthere , I think what you said was right and the `add_` is indeed not correct for the paddle front end. I will merge your `subtract_` and try to fix the `add_` ! Thanks again for your contribution!

jieunboy0516:
ivy-gardener

he11owthere:
@jieunboy0516 Thank you very much for your assistance! I'd like to inform you that I've already submitted a pull request addressing the fix for the `add_` method. You can find it here: https://github.com/unifyai/ivy/pull/22529. Your consideration is greatly appreciated! 🙂

PR #22469, opened 2023-08-24
adds eval mode to stateful module

RickSanchezStoic:
Adds a recursive eval function that iterates through modules to set the eval mode to the specified mode if `training` is an attribute of the said child modules or the module itself. Fixed BatchNoorm2D to maintain the internal state of training.

### DIFF
--- a/ivy/stateful/module.py
+++ b/ivy/stateful/module.py
@@ -817,6 +817,16 @@ class Module(ModuleHelpers, ModuleConverters, ModuleMeta):
         """Set the buffer at any place within the class."""
         self._set_buffers({var_name: value})
 
+    def eval(self, mode: bool = False):
+        # disables training mode for child modules
+        # if it exists
+        if hasattr(self, "training"):
+            self.training = mode
+        for module in self.v:
+            module = getattr(self, module, None)
+            if module:
+                module.eval(mode=mode)
+
     def __repr__(self):
         return object.__repr__(self)
 
--- a/ivy/stateful/norms.py
+++ b/ivy/stateful/norms.py
@@ -103,6 +103,7 @@ class BatchNorm2D(Module):
         device=None,
         v=None,
         dtype=None,
+        training=True,
     ):
         """
         Class for applying Layer Normalization over a mini-batch of inputs.
@@ -151,6 +152,7 @@ class BatchNorm2D(Module):
         self._bias_init = Zeros()
         self._running_mean_init = Zeros()
         self._running_var_init = Ones()
+        self.training = training
         Module.__init__(self, device=device, v=v, dtype=dtype)
 
     def _create_variables(self, device, dtype=None):
@@ -175,7 +177,7 @@ class BatchNorm2D(Module):
     def _forward(
         self,
         inputs,
-        training: bool = False,
+        training: bool = None,
     ):
         """
         Perform forward pass of the BatchNorm layer.
@@ -192,6 +194,8 @@ class BatchNorm2D(Module):
         ret
             The outputs following the batch normalization operation.
         """
+        orig_training = self.training
+        training = training if training is not None else self.training
         normalized, running_mean, running_var = ivy.batch_norm(
             inputs,
             self.v.running_mean,
@@ -207,4 +211,5 @@ class BatchNorm2D(Module):
             self.v.running_mean = running_mean
             self.v.running_var = running_var
 
+        self.training = orig_training
         return normalized


ivy/stateful/module.py
```diff
@@ -817,6 +817,16 @@ def register_buffer(self, var_name, value):
         """Set the buffer at any place within the class."""
         self._set_buffers({var_name: value})
 
+    def eval(self, mode: bool = False):
```
vedpatwardhan, line 820: Actually `eval` doesn't need to receive a `mode` argument, basically `model.train()` would set `training` to `True` and `model.eval()` would set `training` to `False` as used in `torch.nn.Module`. Also `self.training` should be `True` by default, so we should be setting that in the `__init__` itself

ivy/stateful/module.py
```diff
@@ -817,6 +817,16 @@ def register_buffer(self, var_name, value):
...
+        # disables training mode for child modules
+        # if it exists
+        if hasattr(self, "training"):
+            self.training = mode
+        for module in self.v:
```
vedpatwardhan, line 825: `self.v` is a `ivy.Container` with `ivy.Array` leaves right? Not sure why it would have an `ivy.Module` instance. Did you intend to use `self.__dict__` instead?

ivy/stateful/norms.py
```diff
@@ -151,6 +152,7 @@ def __init__(
         self._bias_init = Zeros()
         self._running_mean_init = Zeros()
         self._running_var_init = Ones()
+        self.training = training
```
vedpatwardhan, line 155: Actually we wouldn't need to set a `training` variable for `LayerNorm` or `Dropout` explicitly, all this should be handled directly through whether `self.training` was `True` for the parent module (which could of course also be this same module)

ivy/stateful/norms.py
```diff
@@ -175,7 +177,7 @@ def _create_variables(self, device, dtype=None):
     def _forward(
         self,
         inputs,
-        training: bool = False,
+        training: bool = None,
```
vedpatwardhan, line 180: let's remove this argument, I think we might just introduce confusion by adding an additional argument where `model.train()` and `model.eval()` should really be the only way to control the training mode passed. So provided we ensure that, only passing `self.training` on line [206](https://github.com/unifyai/ivy/pull/22469/files#diff-08ded1880c50b8c039f86533aab4b2b05be766fe26e9c7edc8d4655f4acb6149R206) without any additional handling should work

vedpatwardhan (changes requested):
Hey @RickSanchezStoic, thanks for looking into this, just requested a few changes 😄

ivy/stateful/module.py
RickSanchezStoic: I am iterating the keys, which are going to be Module instances. Iterating over `self.__dict__` will be suboptimal, as we need to just focus on the Module instances that are contributing to weight updates.

vedpatwardhan: yeah you're right, thanks for clarifying!

ivy/stateful/norms.py
```diff
@@ -192,6 +194,8 @@ def _forward(
         ret
             The outputs following the batch normalization operation.
         """
```
mattbarrett98, line 196: I agree with ved, I'm not seeing why this logic is necessary. training mode is now dictated by `self.training` so surely no need for `training` to also be an argument to the forward, and we just pass `self.training` to the functional batch norm directly?

mattbarrett98:
+ maybe add a test

ivy/stateful/module.py
vedpatwardhan: There's a minor bug with this logic where if a user passes `model.eval(True)`, it would set `training=True`

ivy/stateful/module.py
```diff
@@ -817,6 +818,18 @@ def register_buffer(self, var_name, value):
...
+        # disables training mode for child modules
+        self.training = mode
+        for module in self.v:
+            module = getattr(self, module, None)
+            if isinstance(module, ivy.Module):
+                module.eval()
+
+    def train(self, mode: bool = True):
+        # enables/disables training mode
+        self.eval(mode=mode)
```
vedpatwardhan, line 831: this should actually be `self.eval(mode = not mode)`

ivy/stateful/norms.py
vedpatwardhan: but how would this be different from doing a `model.train()` or `model.eval()` on a BatchNorm module. If we consider the `torch` design for e.g., [`torch.nn.functional.batch_norm`](https://pytorch.org/docs/stable/generated/torch.nn.functional.batch_norm.html) has a `training` parameter, but [`torch.nn.BatchNorm2d`](https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm2d.html#torch.nn.BatchNorm2d) doesn't have such a parameter, and it doesn't need it either given that there's `model.train()` and `model.eval()` 😅
Secondly it becomes hard to debug if this `training` parameter overwrites `self.training` because we already have `model.train()` and `model.eval()` to do that.
To your point of being able to control this to the module, we could accept `training` as an argument to the `ivy.Module.__init__` which would be `True` by default, and then once a `ivy.Module` is created, the only way to modify it would be by calling `model.train()` or `model.eval()` (for the parent module or any child module) 😄

vedpatwardhan (changes requested):

ivy/stateful/norms.py
RickSanchezStoic: In that case, we would need to get rid of the argument from the _forward method of this implementation.

ivy/stateful/norms.py
RickSanchezStoic: agreed, we can get rid of the argument from forward entirely.

Issue #22478, opened 2023-08-24
rsqrt_


he11owthere:
To-do list: https://github.com/unifyai/ivy/issues/15115

PR #22485, opened 2023-08-24
Added rsqrt_ method to the Paddle frontend

he11owthere:
closes https://github.com/unifyai/ivy/issues/22478

image


zhumakhan:
It seems tests are failing because package `dill` is missing. Implementation seems all good!

PR #22495, opened 2023-08-24
Modified floor_ method in the Paddle frontend

DRACrY0:
I've fixed the `floor_` method, which previously wasn't correctly performing the desired in-place modifications on the input array. This modification now ensures that the `floor_` method works as intended, directly modifying the input array inplace.

image


KareemMAX (approved):
Hey @DRACrY0,
LGTM, actually your code reveals a flaw in our testing pipeline which need to be fixed. I'll flag that internally

PR #22496, opened 2023-08-24
fixed the scipy func wrapping issues

umairjavaid:
-specific scipy function wrapping is not required because scipy expects a numpy array as input and outputs results using numpy array. we can simply use the scipy function wrapping here.


ivy/functional/frontends/scipy/linalg/linalg.py
```diff
@@ -1,6 +1,6 @@
 # global
 import ivy
-from ivy.functional.frontends.scipy.func_wrapper import (
+from ivy.functional.frontends.numpy.func_wrapper import (
     to_ivy_arrays_and_back,
 )
```
AnnaTz, line 6: It would be better practice if you imported the numpy function wrapper into the scipy func_wrapper file and used that in the scipy function files.

AnnaTz (changes requested):

ivy/functional/frontends/scipy/linalg/linalg.py
umairjavaid: fixed it

PR #22516, opened 2023-08-25
fix: adding to_dlpack method to support __dlpack__ and __dlpack_device__ in ivy.array

abdulasiraj:
There's problem with jax regarding this behavior. I've raised the issue on jax [repo](https://github.com/google/jax/issues/17285).


ivy-leaves:

Thank you :hugs:

vedpatwardhan (commented):
Given that there's an ongoing discussion on the discord server regarding this, I haven't suggested any changes here yet, let me know once you're done with the changes as per the conclusion of that discussion. Thanks @abdulasiraj 😄

hmahmood24:
I hope you don't mind if you can self-assign this @vedpatwardhan since you're the most suited for reviewing this? If not, no worries. Thanks! 😃

vedpatwardhan:
Sure, I'll self-assign it 😄

abdulasiraj:
hi @vedpatwardhan,
This PR is ready to review. I've fixed `to_dlpack`, `from_dlpack` methods as well. I've tested them properly. they are working accept for `numpy` backend which doesn't support PyCapsule objects. Have a look and let me know if there's anything. Thanks!

abdulasiraj:
just adding @tomatillos to make sure things are working fine!

abdulasiraj:
Right now, we can't make it work for numpy as numpy only supports `__dlpack__` dunder method in order to consume but tf and paddle don't have implementation for `__dlpack__`. so, we can only leverage the capsule method by excluding numpy. Hope you get my point @tomatillos ?
thanks!

ivy/data_classes/array/array.py
```diff
@@ -792,7 +792,7 @@ def __bool__(self):
         return self._data.__bool__()
 
     def __dlpack__(self, stream=None):
-        return self._data.__dlpack__()
+        return ivy.to_dlpack(self)
 
     def __dlpack_device__(self):
```
vedpatwardhan, line 797: have we tested if `__dlpack_device__` works correctly?

ivy/functional/backends/jax/creation.py
```diff
@@ -122,9 +122,15 @@ def eye(
...
 
 
+def to_dlpack(x, /, *, out: Optional[JaxArray] = None):
+    return jax.dlpack.to_dlpack(x)
+
+
 def from_dlpack(x, /, *, out: Optional[JaxArray] = None) -> JaxArray:
-    capsule = jax.dlpack.to_dlpack(x)
-    return jax.dlpack.from_dlpack(capsule)
+    # Right now, we can't support array because paddle and tf
```
vedpatwardhan, line 130: Have we created an issue on the `paddle` and `tensorflow` repositories regarding this? Would be great to link those issues here with a `ToDo` comment

ivy/functional/ivy/creation.py
```diff
@@ -1691,13 +1691,53 @@ def full(
     )
 
 
+@handle_exceptions
 @handle_backend_invalid
 @handle_nestable
-@handle_array_like_without_promotion
```
vedpatwardhan, line 1696: not sure why we've removed this decorator

ivy/functional/ivy/creation.py
```diff
@@ -1691,13 +1691,53 @@ def full(
...
+    in the standard.
+
+    Both the description and the type hints above assumes an array input for simplicity,
+    but this function is *nestable*, and therefore also accepts :class:`ivy.Container`
+    instances in place of any of the arguments.
+    """
+    return current_backend(x).to_dlpack(x, out=out)
+
+
+@handle_backend_invalid
```
vedpatwardhan, line 1740: we should also add all applicable decorators to `from_dlpack` if it accepts an `ivy.Array`

vedpatwardhan (changes requested):
Hey @abdulasiraj, just requested a few changes, thanks!

tomatillos:
Small point: the array api standard says that `from_dlpack` should take an object with a `__dlpack__` method, however in this implementation we need to pass capsules in some frameworks. (e.g. the test at the moment passes a capsule to the `from_dlpack` method, rather than an object with `__dlpack__` method). How strictly do we want to adhere to the api standard? If we stick with taking capsules, we could call it something different, or at least change the docstring. @vedpatwardhan 

I also wonder if the test could be improved by starting with an array in one framework, and then calling `from_dlpack` in another one, since this is likely the only reason these functions will be used. For the transpiler to be able to use this, we'd need to be able to convert between every pair of frameworks with these functions.

vedpatwardhan:
> Small point: the array api standard says that `from_dlpack` should take an object with a `__dlpack__` method, however in this implementation we need to pass capsules in some frameworks. (e.g. the test at the moment passes a capsule to the `from_dlpack` method, rather than an object with `__dlpack__` method). How strictly do we want to adhere to the api standard? If we stick with taking capsules, we could call it something different, or at least change the docstring. @vedpatwardhan
> 
> I also wonder if the test could be improved by starting with an array in one framework, and then calling `from_dlpack` in another one, since this is likely the only reason these functions will be used. For the transpiler to be able to use this, we'd need to be able to convert between every pair of frameworks with these functions.

Yeah that's a good point! I think we should create those issues on `paddle` and `tf` repos because we need the `__dlpack__` method for us to adhere to the standard, and let's continue with the same docstring as the standard for now, with a `ToDo` comment linking to those issues in the `paddle` and `tf` backend implementations. If the other backends support it anyway, we should also support array in `from_dlpack` for those backends saying that we'd be able to support them for `paddle` and `tf` once those issues are fixed on their repos 😄

Yeah completely agree with the improvement to the tests 👍

ivy/data_classes/array/array.py
abdulasiraj: No, as `tf` and `paddle` doesn't support these methods yet.

ivy/functional/ivy/creation.py
abdulasiraj: these decorators operate on array but on `PyCapsule` object as it's not array most decorators throw errors.

ivy/functional/ivy/creation.py
abdulasiraj: for `to_dlpack`, we can add these no problem as paramter is array.

abdulasiraj:
> 

makes sense. right now, we can just support for `to_dlpack` and `from_dlpack` but for aligning with array_api_standards we have to wait for `tf` and `paddle` to support these dunders. let me know the final implementation for now @tomatillos @vedpatwardhan .
thanks!

vedpatwardhan:
> > 
> 
> makes sense. right now, we can just support for `to_dlpack` and `from_dlpack` but for aligning with array_api_standards we have to wait for `tf` and `paddle` to support these dunders. let me know the final implementation for now @tomatillos @vedpatwardhan . thanks!

I think in the meantime, we can also have `to_dlpack` and `from_dlpack` aligned with the standard for the other backends right? ig we should the get those backends working according to the standard and leave `ToDo` comments with the `paddle` and `tf` backends linking the issues created on their repositories regarding supporting them. Thanks @abdulasiraj 😄

abdulasiraj:
Hi @vedpatwardhan,
thanks for your review but @tomatillos said we need all frameworks working which is only possible with legacy methods. let me know what you think. thanks!

vedpatwardhan:
> Hi @vedpatwardhan, thanks for your review but @tomatillos said we need all frameworks working which is only possible with legacy methods. let me know what you think. thanks!

So just to understand the point, the `tensorflow` and `paddle` backends don't have a `__dlpack__` dunder method right? But then does the legacy method you're referring to allow us to pass `ivy.Array` to `from_dlpack` and convert it to a capsule and then convert it to an `ivy.Array` of another backend? 
Also, what's the legacy method of conversion which you're referring to when it comes to `tensorflow` and `paddle`?
Because if that's not possible, we can't really say that the legacy method would get things working with `tensorflow` and `paddle`. And if it does, then we're already adhering with the standard except for the presence of the `__dunder__` method which we've already created issues for and can link for the time being, with the legacy method still working with `tensorflow` and `paddle` meaning it would be a working solution for @tomatillos in the time being across all backends.
Apologies if I'm missing something, thanks @abdulasiraj @tomatillos 😄

abdulasiraj:
so basically, `to_dlpack` and `from_dlpack` are not working across all backends before this PR. lagacy method is to use `to_dlpack` and `from_dlpack`. All framework supports these lagacy methods except numpy.  so, after this PR,

1 - `to_dlpack` and `from_dlpack` are working across all backends except numpy.
2 - dunders version is also working except in paddle and tf.

I'll add link to the issue for paddle and tf and we are good to merge this PR for now.  let me know what you think @vedpatwardhan 
thanks!

ivy/data_classes/array/array.py
```diff
@@ -792,9 +792,19 @@ def __bool__(self):
         return self._data.__bool__()
 
     def __dlpack__(self, stream=None):
-        return self._data.__dlpack__()
+        # Not completely supported yet as paddle and tf
```
vedpatwardhan, line 795: I think we can just add the comment once in `__dlpack__` as `__dlpack_device__` is right below it anyway. Rest looks good to merge, thanks @abdulasiraj 😄

PR #22539, opened 2023-08-25
Add torch frontend grid_sample and test function

### LINKED ISSUE #22307: grid_sample

Deepdive543443:
Close #22307


Deepdive543443:
Hi @Dsantra92. I just made this PR because I ran into some issues on testing this function. It yield error like this and I didn't spot error about the min and max values in testing function. The function only test the 2D grid_sample in bilinear and nearest sampling, in my own testing it already has output that identical as torch does. Looking forward to hear back from you.
image

Deepdive543443:
Hi @Dsantra92  and @paulaehab . Hope you just have a fantastic weekend! I think I need some support on testing the function. The ground truth values in screenshot 1 might not be correct. As shown in screenshot 2, I tried import the this values into the torch grid sample function and receive an output same as ret_np. I'm not sure which part of it was going wrong. If you have some free time, I would like to hear some advise from you.
image
image

Sam-Armstrong:
Hi @Deepdive543443, I've just been trying this out using the following minimal example:
```
import torch

if __name__ == '__main__':
    inp = torch.randn((3, 3, 4, 4))
    grid = torch.randint(0, 4, size=(3, 4, 4, 2), dtype=torch.float32)

    mode = 'bilinear'
    padding = 'reflection'

    result_orig = torch.nn.functional.grid_sample(inp.clone(), grid.clone(), mode=mode, padding_mode=padding, align_corners=False)
    print('native fn result shape', result_orig.shape)
    
    result_new = torch_frontend.nn.functional.grid_sample(inp.clone(), grid.clone(), mode=mode, padding_mode=padding, align_corners=False)
    print('frontend fn result shape', result_new.shape)
    
    print(result_orig[0])
    print(result_new[0])
    print(torch.allclose(result_orig, result_new, atol=1e-3))
```
It looks like there is a problem with the `grid[zeros_mask]` call in grid_sample_padding, from the exception thrown by this example. Do you know what the issue is at all?

Deepdive543443:
Hi @Sam-Armstrong The issues was happened because I used an ivy boolean array to mask the grid. It will throw an error when the grid input is a torch Tensor, not ivy array.
I updated the code to transfer the zeros_mask into native array in the last commit. It should fix the issue. But I need to remove the **@to_ivy_arrays_and_back** decorator, otherwise I cannot transfer the function output to native torch array and feed it into **torch.allclose()**. Not sure about the reason of this.

Sam-Armstrong:
Sorry, here's a better example that's working:
```
import torch
import numpy as np

if __name__ == '__main__':
    inp = torch.randn((3, 3, 4, 4))
    grid = torch.randint(0, 4, size=(3, 4, 4, 2), dtype=torch.float32)

    mode = 'bilinear'
    padding = 'reflection'

    result_orig = torch.nn.functional.grid_sample(inp.clone(), grid.clone(), mode=mode, padding_mode=padding, align_corners=False)
    print('native fn result shape', result_orig.shape)
    
    result_new = torch_frontend.nn.functional.grid_sample(inp.clone(), grid.clone(), mode=mode, padding_mode=padding, align_corners=False)
    print('frontend fn result shape', result_new.shape)

    print(np.allclose(result_orig, result_new.ivy_array.to_numpy(), atol=1e-3))
```
I don't understand why grid is a torch.Tensor at that point though, shouldn't it have been converted into an ivy array?

Deepdive543443:
Hi @Sam-Armstrong So surprise to have you around this early :) Do you think this is a bug on **@to_ivy_arrays_and_back** decorator? Or should I covert them to ivy array in manually?

ivy/functional/frontends/torch/nn/functional/vision_functions.py
```diff
@@ -372,3 +373,248 @@ def upsample_bilinear(input, size=None, scale_factor=None):
...
+    masks = []
+    for idx, border in enumerate(borders):
+        masks.append(ivy.bitwise_or(grid[..., idx] < -4, grid[..., idx] > border + 2))
+        borders[idx] += 2
+
+    zeros_mask = masks[0]
+    for i in range(1, len(borders)):
+        zeros_mask = ivy.bitwise_or(zeros_mask, masks[i])
+
+    zeros_mask = ivy.native_array(zeros_mask)
```
Sam-Armstrong, line 426:
```suggestion
```
We can remove this now

ivy/functional/frontends/torch/nn/functional/vision_functions.py
```diff
@@ -95,6 +95,7 @@ def affine_grid(theta, size, align_corners=False):
             return grid.view((N, D, H, W, 3))
 
 
+# @with_unsupported_dtypes({"2.0.1 and below": ("float16", "bfloat16")}, "torch")
```
Sam-Armstrong, line 98:
```suggestion
```

Sam-Armstrong (changes requested):
We can actually resolve that problem by fixing the minimal example to pass ivy arrays, so we can (and should) remove that ivy.native_array line. I've tried one of the test cases which is failing in pytest, and it seems to be passing fine in the minimal example 🤔 so I'll talk to someone from the testing team about that to see if they know what's going on. By the way, do you know why `Exception occured: list index out of range` is printed out several times when running this? Thanks!

Sam-Armstrong:
Corrected minimal example:
```
import torch
import numpy as np
import ivy

if __name__ == '__main__':
    ivy.set_backend('numpy')
    
    inp = torch.randn((3, 3, 4, 2))
    grid = torch.randint(0, 4, size=(3, 4, 4, 2))

    mode = 'nearest'
    padding = 'border'
    align_corners = True

    result_orig = torch.nn.functional.grid_sample(inp.clone(), grid.clone(), mode=mode, padding_mode=padding, align_corners=align_corners)
    print('native fn result shape', result_orig.shape)
    
    result_new = torch_frontend.nn.functional.grid_sample(ivy.array(inp.clone().detach().numpy()), ivy.array(grid.clone().detach().numpy()), mode=mode, padding_mode=padding, align_corners=align_corners)
    print('frontend fn result shape', result_new.shape)

    print(np.allclose(result_orig, result_new.ivy_array.to_numpy(), atol=1e-4))
```

Deepdive543443:
Afternoon @Sam-Armstrong I just added the request change. Would you like share more details about the test cases that gave you `Exception occured: list index out of range`? I didn't find how to reproduce this error. 
About the pytest part, it looks like it calls a Tensorflow ground truth function with behaviour that doesn't match with PyTorch. We might need to hear from testing team about this issue.
Another thing I want to note is I usually use randn() to generate grid because PyTorch expect most of the grid input in range between -1 and 1(some of them outside this range would be padding). Using randint should be fine but I found it makes the test fail when doing 3D sampling🤔I'm looking into it

Deepdive543443:
Hi @Sam-Armstrong Hope you are doing well this week. Did you get update from test team? From this week, I would need to move most of my effort to graduation. I might response slower than usual, but will try my best on get this PR done ;)

Sam-Armstrong:
I think @sherry30 is looking into this, not sure if there is any update? No worries at all, best of luck with your graduation work also @Deepdive543443 😁

sherry30:
Hi @Deepdive543443 @Sam-Armstrong ,
Sorry it took me a while to debug this but I think I've figured out the problem.
Your frontend implementation inplace updates `grid` array. So in our testing pipeline we first execute the frontend function, that changes the `grid` array, and then when the torch's  native function is called it produces different results because of the modified `grid` array.
Your function shouldn't inplace update the argument.

Deepdive543443:
Hi @sherry30 @Sam-Armstrong After adding copy and type casting I think the function could pass three of the test for now. The remain issues are:

- Tensorflow testing showing a `KeyError: 'cpu'` error. It seems happened some paddle testing before, but I didn't find how to deal with it in Tensorflow. 
- ivy.pad() seems no working when testing for paddle
- Pre-commit's auto-formatting would throw functions everywhere, so I haven't applied it yet

It looks like not just my implementation of grid_sample(), some others test function under `test_vision_functions.py` are not passing all the testing as well. Is that necessary to make sure all the testing could pass?
image
image

ivy/functional/frontends/torch/nn/functional/vision_functions.py
```diff
@@ -355,3 +355,272 @@ def upsample_bilinear(input, size=None, scale_factor=None):
...
+    if grid[zeros_mask].shape[0] > 0:
+        grid[zeros_mask] = ivy.array(borders)
+    return grid
+
+
+cubic_conv1 = lambda A, x: ((A + 2) * x - (A + 3)) * x * x + 1
+cubic_conv2 = lambda A, x: (((A * x) - (5 * A)) * x + (8 * A)) * x - (4 * A)
+
+
+@with_supported_dtypes({"2.0.1 and below": ("float32", )}, "torch")
```
Sam-Armstrong, line 416:
```suggestion
@with_supported_dtypes({"2.0.1 and below": ("float32",  "float64")}, "torch")
```
float64 is still a valid type. I've checked and the tests still pass with float64 included

Sam-Armstrong (changes requested):
Thanks @Deepdive543443, that's great! I don't think these test errors are too much of a big deal - the other backends clearly show that the function is working correctly, so I think we can address these failures later. I've left a couple of comments with things we need to change before we merge, please could you take a look? And thanks for the great work!

ivy/functional/frontends/torch/nn/functional/vision_functions.py
```diff
@@ -355,3 +355,272 @@ def upsample_bilinear(input, size=None, scale_factor=None):
...
+            grid_clone[..., 2] = ((grid_clone[..., 2] + 1) * d - 1) / 2
+
+        batch_coor = ivy.reshape(ivy.arange(n), (-1, 1))
+        batch_coor = ivy.repeat(batch_coor, to_d * to_h * to_w, axis=1)
+        batch_coor = ivy.reshape(batch_coor, (n, to_d, to_h, to_w))
+        padding = [(0, 0) for _ in range(2)] + [(3, 3) for _ in range(3)]
+        input_clone = ivy.pad(input_clone, padding, mode="constant", constant_values=0)
+
+        grid_clone = grid_sample_padding(grid_clone, padding_mode, align_corners, borders=[w, h, d])
+        # grid_clone += 3
```
Sam-Armstrong, line 548:
```suggestion
```

Deepdive543443:
Thank you @Sam-Armstrong and @sherry30 for supervising and guiding. I still not sure about if pre_commit is working correctly, so I placed all functions to the bottom of file after formatting. Hopefully it's good to go :)

Sam-Armstrong:
@Deepdive543443 I'll just check whether the tf and paddle test failures are ok with Sherry, then should be good to merge 😁

Sam-Armstrong:
Merged! Thanks for the great work you put into this PR @Deepdive543443!

Deepdive543443:
Thanks @Sam-Armstrong and @sherry30 Finally it is done. Learned a lot from this task :)

PR #22541, opened 2023-08-25
Added round_ method to the Paddle frontend

he11owthere:
closes https://github.com/unifyai/ivy/issues/22480

### DIFF
--- a/ivy/functional/frontends/paddle/tensor/tensor.py
+++ b/ivy/functional/frontends/paddle/tensor/tensor.py
@@ -223,6 +223,11 @@ class Tensor:
         self.ivy_array = self.floor().ivy_array
         return self
 
+    @with_supported_dtypes({"2.5.1 and below": ("float32", "float64")}, "paddle")
+    def round_(self, name=None):
+        self.ivy_array = paddle_frontend.round(self).ivy_array
+        return self
+
     @with_supported_dtypes(
         {"2.5.1 and below": ("float32", "float64", "int32", "int64")}, "paddle"
     )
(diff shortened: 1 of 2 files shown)


he11owthere:
I'm encountering the following error message across all the backends, indicating that the return value from my `round_` method (specifically, -0.0) doesn't match the returned value from the backend's `round_` method (i.e. -1.0) when testing with input = -0.5.

As per the array API standard, the expected behavior is to return the nearest integer value in such situations. I would greatly appreciate your guidance on this matter. Thank you very much for your assistance!

image

ZoeCD:
Hello, Bhushan! I see what you mean. Paddle paddle does not follow the standard 
[https://www.paddlepaddle.org.cn/documentation/docs/zh/api/paddle/round_cn.html](https://www.paddlepaddle.org.cn/documentation/docs/zh/api/paddle/round_cn.html). I'm tagging @AnnaTz since she is leading the frontends. She'll know what's best to do in this case!

AnnaTz:
> I'm encountering the following error message across all the backends, indicating that the return value from my `round_` method (specifically, -0.0) doesn't match the returned value from the backend's `round_` method (i.e. -1.0) when testing with input = -0.5.
> 
> As per the array API standard, the expected behavior is to return the nearest integer value in such situations. I would greatly appreciate your guidance on this matter. Thank you very much for your assistance!
> 
> image

It seems that `paddle.round` does not follow the array api standard. You can see in the official documentation's [example](https://www.paddlepaddle.org.cn/documentation/docs/zh/api/paddle/round_cn.html#round) that paddle indeed round -0.5 to -1.0.
A frontend function should mimic the original function's behavior, so your implementation should return the same results as paddle.

he11owthere:
Thanks, @AnnaTz, for your effort and guidance. I've made the necessary changes. 😊
@ZoeCD, I've resolved the merge conflicts, and the PR is ready for merging. 🙂

ZoeCD:
Thank you for the help @AnnaTz! 
And thank you for your PR @he11owthere!

PR #22552, opened 2023-08-25
Add eigvals to numpy frontend

### LINKED ISSUE #22266: eigvals

abdllahdev:
Close #22266


abdllahdev:
Note: the unit test fails when using torch as a backend even though the eigenvalues are the same. I use np.sort before testing the function but it seems that np.sort doesn't have any effect on the returned np.array when the backend is set to torch. I'm not sure why this happens. I tried to figure out where the problem might originate but I couldn't.

Here is a sample of the failed test
image

PR #22624, opened 2023-08-26
Paddlepaddle avg pool3d

### LINKED ISSUE #22597: avg_pool3d

KaylaMLe:
This is a subtask of #15055.

## Related Issue 

Close #22597


JPGoodale:
Hi @KaylaMLe, it looks like there is an issue with your test args as all tests are currently returning: 

"""
E   hypothesis.errors.InvalidArgument: test_paddle_avg_pool3d() got an unexpected keyword argument 'dtype_x_k_s', from dtype_x_k_s=arrays_for_pooling(min_dims=4, max_dims=4, min_side=2, max_side=4) in @given
"""

KaylaMLe:
Ah, okay. That's the test I tried to add, so that makes sense. Are those more specific error messages available in GitHub or only when you run the test locally? From my end, all I could tell were that some number of seemingly random tests were failing, and I couldn't find the error message

KaylaMLe:
I had been trying to set up local testing for a while, but I was having issues with getting Docker to run with PyCharm

JPGoodale:
@KaylaMLe those errors were from running locally. As for your Docker/PyCharm setup issues, if you check out the PyCharm channel on the Ivy discord, I believe that others have faced this issue as well and offered some solutions.

KaylaMLe:
Thanks. There's a pinned comment about adding package paths to the interpreter which may be causing the issue, but the UI I see in PyCharm is different from the screenshot. I'm currently asking about it in that channel.

JPGoodale:
Hi @KaylaMLe, any luck on getting the tests set up in PyCharm?

KaylaMLe:
Thanks for following up on this. Unfortunately, I still haven't been able to work out the testing issues. I'm also still unsure why my PR is failing. Someone else also got a PR in for this subtask before it was marked as in progress, and AFAIK their PR is also failing. It may be best at this point to close the PR given that I haven't been able to make progress for so long.

PR #22626, opened 2023-08-26
Mfccs from log mel spectrograms

### LINKED ISSUE #22464: mfccs_from_log_mel_spectrograms
tf.signal.mfccs_from_log_mel_spectrograms

luc-zago:
## Related Issue 

Close #22464 

## Checklist 

- [X] Added a function
- [X] Added tests
- [X] Followed the provided steps

### DIFF
--- a/ivy/functional/frontends/tensorflow/signal.py
+++ b/ivy/functional/frontends/tensorflow/signal.py
@@ -34,13 +34,10 @@ def kaiser_window(window_length, beta=12.0, dtype=ivy.float32, name=None):
     {"2.13.0 and below": ("float16", "float32", "float64", "bfloat16")},
     "tensorflow",
 )
-@handle_tf_dtype
 @to_ivy_arrays_and_back
-def mfccs_from_log_mel_spectrograms(
-    log_mel_spectrograms, dct_type=2, n_mfcc=13, name=None
-):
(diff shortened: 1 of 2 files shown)


hmahmood24 (changes requested):
Hey @luc-zago, thanks for the PR! However, in order to proceed forward, there a couple of points we need to work on. 

1. First off, this PR is supposed to add just a frontend function and its test and so adding functions to any other APIs is not recommended/acceptable. 
2. Secondly, you've added the `mfccs_from_log_mel_spectrograms` to just the `tensorflow` backend and so this function is non-existant in all the other backends which would cause the tests to fail with every other backend apart from `tensorflow`.

There are two ways you can go about now:

1. The proper procedure of adding a new function to `ivy` is by first creating a new `Missing Function Suggestion` issue on `ivy`. This issue is then reviewed by our team and if approved, this function would then need to be added not only to the experimental API in `ivy` but to all its backends along with its test as well. Once that PR is then merged, only then can you proceed with using that function from the frontends for your own PR. Till then this PR would have to be closed or halted.
2. Otherwise you can proceed to remove the implementations you added from the backend and the functional API and just implement the `mfccs_from_log_mel_spectrograms` in the `tensorflow` frontend using a composition of existing `ivy` functions in this PR. 

Let me know it this makes sense. Thanks! :)

hmahmood24:
Closing this PR since there has been no activity for more than a week.

PR #22644, opened 2023-08-27
refactor: fill_diagonal to allow array_like v

yopknopixx:
# PR Description 

`np.fill_diagonal` allows the fill value `val` to be an array like while `ivy.fill_diagonal` only supports `val` be to a `float` or an `int`. Your task is to update the `ivy.fill_diagonal` function to support the `val` to be an array-like as well.

## Related Issue 

This PR is related to the Trello task: https://trello.com/c/2iwvraV8/2600-refactor-ivyfilldiagonal-function
### Backend Progress:
- [x] Torch
- [x] Paddle
- [x] TensorFlow
- [x] Jax
- [x] Numpy


ivy-leaves:

Thank you :hugs:

hello-fri-end (changes requested):
Hey @yopknopixx, thanks for working on this problem and apologies for the super late response here. I looked at the `np.fill_diagonal` function and according to it's documentation it in-place updates the input tensor and the problem is that we can't do inplace updates in jax and tensorflow backend implementations since jax and tensorflow do not natively support inplace updates. However, in our jax and tensorflow backends we have implemented the function compositionally. So, it makes more sense to directly have a compositional implementation in ivy where in we can handle the inplace update by calling the `ivy.inplace_update` function. So, you would basically need to do the following:
1. Remove jax, tensorflow and paddle backend implementations ( even paddle has a compositional implementation)
2. Add a compositional implementation in ivy.
3. Update the tests to generate an array or a number for the `fill_value`

This would basically convert the function to a [mixed function](https://unify.ai/docs/ivy/overview/deep_dive/function_types.html#mixed-functions) . Feel free to directly dm me on discord or gchat incase you have any questions.

yopknopixx:
Hey @hello-fri-end apologies for the delayed response. I just pushed 1 and 2. I'd already completed 3. But, I'm facing issues with `paddle` and `tensorflow`.

1. For `paddle`, it seems there is a bug in the `scatter_nd` function that causes the error: `ivy.Shape(9,) must be equal to ivy.Shape()`.
2. As for `tensorflow`, I'm getting `hypothesis.errors.InvalidArgument: Cannot sample from a length-zero sequence.`

Could you please let me know how to proceed? Especially with the `tensorflow` backend? Meanwhile I'll try to resolve the paddle issue.

Also, I'm not sure I completely understand the motivation behind doing this. This won't have any performance benefits, right? (Ivy would still have to create new backend arrays)

ivy/functional/backends/torch/experimental/manipulation.py
```diff
@@ -435,7 +435,7 @@ def unique_consecutive(
 
 def fill_diagonal(
     a: torch.Tensor,
```
hello-fri-end, line 437: let's remove the torch implementation as well since there isn't any a native `fill_diagonal` function that we are leveraging here. Our compositional implementation in ivy would be just as efficient when the code is compiled.

ivy/functional/ivy/experimental/manipulation.py
```diff
@@ -2177,14 +2177,14 @@ def unique_consecutive(
 
 
 @handle_exceptions
-@handle_backend_invalid
 @handle_nestable
+@handle_partial_mixed_function
```
hello-fri-end, line 2181: No need for this since this is just a mixed function and not a partial mixed function.

hello-fri-end (changes requested):
Hey @yopknopixx, you seem to have removed a lot of test from `ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_manipulation.py`. Could you please undo that and resolve the comments I have left. Happy to review again once you are done :)

yopknopixx:
Hey, @hello-fri-end I've made the changes you've asked. But there is some problem running the test for the `tensorflow` backend. `hypothesis.errors.InvalidArgument: Cannot sample from a length-zero sequence.`
Could you please point me in the right direction on how to solve it?

hello-fri-end:
> Hey, @hello-fri-end I've made the changes you've asked. But there is some problem running the test for the `tensorflow` backend. `hypothesis.errors.InvalidArgument: Cannot sample from a length-zero sequence.` Which seems related to [this message](https://discord.com/channels/799879767196958751/982738078445760532/1106160632677535744) on Discord.

That happens since `ivy.function_supported_dtypes(ivy.fill_diagonal)` for Tensorflow backend returns `int32` and `int64` and therefore `helpers.get_dtypes("float")` would return an empty list which leads to this error. Changing `float` to `valid` should fix the problem I believe.

yopknopixx:
Thanks @hello-fri-end! That solved the problem. Is this ready to be merged?

hello-fri-end (approved):
Looks like there are no related failures on the CI! Feel free to merge this @yopknopixx. Great work here, thanks for the contribution! :rocket:

PR #22653, opened 2023-08-27
cleaned up usage of try-except block

mosesdaudu001:
# PR Description


mosesdaudu001:
This PR is related to this trello card [https://trello.com/c/duUpnVgt/2467-clean-up-the-try-except-usage-in-datatype-if-applicable](url)

vedpatwardhan (changes requested):
Hey @mosesdaudu001, thanks for looking into this, just requested a few changes. Also, regarding `promote_types`, we should try and ensure that the [sorting](https://github.com/unifyai/ivy/blob/45f1e1af0e9500258f8635eb08ff6620c6ef6472/ivy/functional/ivy/data_type.py#L2131) steps works correctly so that we don't need to have the [swapping](https://github.com/unifyai/ivy/blob/45f1e1af0e9500258f8635eb08ff6620c6ef6472/ivy/functional/ivy/data_type.py#L2143) logic in the `except` block 😄

ivy/functional/ivy/data_type.py
```diff
@@ -580,10 +580,10 @@ def can_cast(
...
     if isinstance(from_, (ivy.Array, ivy.NativeArray)):
         from_ = from_.dtype
-    try:
-        dtype = ivy.promote_types(from_, to)
-        return dtype == to
-    except KeyError:
+    dtype = ivy.promote_types(from_, to)
+    if dtype == to:
+        return True
+    else:
```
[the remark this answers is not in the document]
mosesdaudu001: Done

ivy/functional/ivy/data_type.py
```diff
@@ -969,11 +969,7 @@ def is_hashable_dtype(dtype_in: Union[ivy.Dtype, ivy.NativeDtype], /) -> bool:
     ret
         True if data type is hashable else False
     """
-    try:
-        hash(dtype_in)
-        return True
-    except TypeError:
-        return False
+    return hash(dtype_in)
```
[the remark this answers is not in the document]
mosesdaudu001: Done

ivy/functional/ivy/data_type.py
```diff
@@ -2549,7 +2545,8 @@ def is_native_dtype(dtype_in: Union[ivy.Dtype, ivy.NativeDtype], /) -> bool:
     >>> ivy.is_native_array(ivy.float64)
     False
     """
-    try:
-        return current_backend(None).is_native_dtype(dtype_in)
-    except ValueError:
+    backend = current_backend(None)
```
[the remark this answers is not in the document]
mosesdaudu001: Done

ivy/functional/ivy/data_type.py
```diff
@@ -2130,18 +2126,18 @@ def promote_types(
     query.sort(key=lambda x: str(x))
```
vedpatwardhan, line 2126: I actually meant here that the way we're sorting the `query` isn't perfectly aligned with the order of [dtype pairs](https://github.com/unifyai/ivy/blob/47359a574231bd476f54fea63ee4fd8105d68991/ivy/__init__.py#L1291) in the promotion table, so if we fix the way we're sorting here or update the order in the promotion table if it's wrong anywhere, then we wouldn't need the try except at all, we can just do `return _promote(query)` from the function directly

vedpatwardhan (changes requested):
Just one change, rest looks good to merge, thanks @mosesdaudu001 😄

ivy/functional/ivy/data_type.py
mosesdaudu001: Done

ivy/functional/ivy/data_type.py
```diff
@@ -580,11 +581,10 @@ def can_cast(
     """
     if isinstance(from_, (ivy.Array, ivy.NativeArray)):
         from_ = from_.dtype
-    try:
-        dtype = ivy.promote_types(from_, to)
-        return dtype == to
-    except KeyError:
-        return False
+    dtype = ivy.promote_types(from_, to)
+    if dtype == to:
```
vedpatwardhan, line 585: let's instead do a `return dtype == to` here, as I suppose the additional `return False` in the `except` isn't needed anymore

ivy/functional/ivy/data_type.py
```diff
@@ -2127,28 +2123,24 @@ def promote_types(
...
-        try:
-            ret = _promote(query)
-        except KeyError:
-            raise ivy.utils.exceptions.IvyDtypePromotionError(
-                "these dtypes ({} and {}) are not type promotable, ".format(
-                    type1, type2
-                )
-            )
-    return ret
+    # try:
```
vedpatwardhan, line 2135: let's remove the commented out code. Also, could you please confirm (by just looping over all pairs) to see if all pairs of dtypes are covered in the promotion tables either forward or reverse (both with the precision and non-precision modes)? Thanks!

vedpatwardhan (changes requested):
Just a final couple of changes, thanks @mosesdaudu001 😄

ivy/functional/ivy/data_type.py
vedpatwardhan: You can check with and without the precision mode by once running the loop with `ivy.set_precision_mode(True)` and once with `ivy.set_precision_mode(False)` globally

ivy/functional/ivy/data_type.py
mosesdaudu001: Done

ivy/functional/ivy/data_type.py
mosesdaudu001: Yes I can confirm that all pairs of dtypes are covered in the promotion table. However I use the `ivy.set_precision_mode()` because it threw an error `AttributeError: module 'ivy' has no attribute 'set_precision_mode'`

vedpatwardhan: Oh sorry, I made a mistake here, it should be `ivy.set_precise_mode` 😅

PR #22676, opened 2023-08-28
Reformat function is_int_dtype

### LINKED ISSUE #22675: is_int_dtype

AryanSharma21:
Close #22675

### DIFF
--- a/ivy/functional/ivy/data_type.py
+++ b/ivy/functional/ivy/data_type.py
@@ -1851,12 +1851,9 @@ def is_bool_dtype(
 @handle_exceptions
 @handle_nestable
 @inputs_to_native_arrays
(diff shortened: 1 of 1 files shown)


tomatillos:
Hi @AryanSharma21,

Thanks for the PR!

However, looking more closely, it’s unclear how this PR would meaningfully contribute towards the project, because it only makes minor documentation changes.

Therefore, I’ve simply closed it for now 🙂

Thanks!

AryanSharma21:
No problem, but there are several issues in the to-do list from where I picked this one. Most of those reformatting tasks are already completed so I just took a reference from other PR in the same list to do some changes with this function. Kindly close those issues as completed so that it doesn't create any confusion. Thanks 😊

PR #22691, opened 2023-08-28
Refactored _parse_query of get_item/set_item to fix transpilation issues

AnnaTz:
Making ivy.get_item/set_item more efficient. Will have to see if it is compilable.

### DIFF
--- a/ivy/functional/ivy/general.py
+++ b/ivy/functional/ivy/general.py
@@ -2861,6 +2861,8 @@ def set_item(
     """
     if copy:
         x = ivy.copy_array(x)
+    if not ivy.is_array(val):
+        val = ivy.array(val)
     if 0 in x.shape or 0 in val.shape:
         return x
     if ivy.is_array(query) and ivy.is_bool_dtype(query):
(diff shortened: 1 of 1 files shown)


ivy/functional/ivy/general.py
```diff
@@ -2901,239 +2887,32 @@ def set_item(
...
+    # stack the multi-D indices to bring them to gather_nd/scatter_nd format
+    x_idxs = ivy.stack(x_idxs, axis=-1).astype(ivy.int64)
 
-    return list(_flatten_gen(iterable))
+    return x_idxs, target_shape
 
 
 def _numel(shape):
+    shape = tuple(shape)
     return math.prod(shape) if shape != () else 1
```
Sam-Armstrong, line 2916:
```suggestion
    return ivy.prod(shape) if shape != () else 1
```
This might be better off being `ivy.prod` rather than `math.prod`? Not totally sure though, just a thought

Sam-Armstrong (approved):
This does fix transpiling MobileBert and MobileNetV2 to tensorflow which is great! I got a little confused because transpiling ivy.set_item is still failing in a minimal example, but I think this is just due to our limitations of tracking slices - which isn't a big issue if it isn't affecting these model transpilations. Thanks for all the help on this @AnnaTz, much appreciated😁

PR #22693, opened 2023-08-28
first draft of proposed solution to update test_frontend_method to te…

mosesdaudu001:
# PR Description 
This PR is related to trello card: https://trello.com/c/3q3nizvc/2275-update-testfrontendmethod-to-test-for-frontend-array-arguments. 

The PR is not fully ready yet, as I have run into some issues I hope to resolve. I am also putting this out there incase someone wants to assist.

Thank you

### DIFF
--- a/ivy_tests/test_ivy/helpers/function_testing.py
+++ b/ivy_tests/test_ivy/helpers/function_testing.py
@@ -1628,40 +1628,73 @@ def test_frontend_method(
             test_flags=method_flags,
             on_device=on_device,
         )
+
+        create_frontend_array = importlib.import_module(
+            f"ivy.functional.frontends.{frontend}"
+        )._frontend_array
+
         # End Method #
 
-        args_constructor_ivy, kwargs_constructor_ivy = ivy_backend.args_to_ivy(
-            *args_constructor, **kwargs_constructor
-        )
-        args_method_ivy, kwargs_method_ivy = ivy_backend.args_to_ivy(
-            *args_method, **kwargs_method
-        )
+        if init_flags.generate_frontend_arrays:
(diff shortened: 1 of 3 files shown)


ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -1683,12 +1716,36 @@ def test_frontend_method(
...
+
+        # create NumPy args
+        if method_flags.generate_frontend_arrays:
+            ret_np_flat = flatten_frontend_to_np(
+                ret=ret,
+                frontend_array_fn=create_frontend_array,
+                backend=backend_to_test,
+            )
+        else:
+            ret_np_flat = flatten_and_to_np(ret=ret, backend=backend_to_test)
```
sherry30, line 1748: Dont think this is necessary, `ret_np_flat` is being computed above.
and `ret` will always be an ivy array, since `get_ret_and_flattened_np_array` converts it to an ivy array.

sherry30 (changes requested):
Requested a few changes,
Can you explain what problems are you facing

mosesdaudu001:
> Requested a few changes, Can you explain what problems are you facing

I have been having an undefined error, and I am trying to pinpoint where it's originating from
`[undefined]ivy.utils.exceptions.IvyBackendException: jax: nested_map: jax: nested_map`

mosesdaudu001:
Hi @sherry30 

The aim of the task is to update the `test_frontend_method` to test for frontend array arguments just like the way it is done with `test_frontend_function` 

To do this we need to add a new argument to the method_flags which is basically a bool (True or False) and if True, we then convert the arguments into frontend arrays.

The issue now is with converting the constructors to their corresponding frontends. When this is done I expect evrrything to start working.

PR #22700, opened 2023-08-28
feat: manual dtype casting removal

Madjid-CH:
# PR Description 

This PR is for this [task](https://trello.com/c/kSyDbng5)

I started with the Paddle backend. I need some feedback on my work so I can continue with the other backends.


ivy-leaves:

Thank you :hugs:

ivy/functional/backends/paddle/activations.py
```diff
@@ -13,33 +13,23 @@
...
-        return F.relu(x.cast("float32")).cast(x.dtype)
+    if paddle.is_complex(x):
+        return paddle.complex(F.relu(x.real()), F.relu(x.imag()))
     return F.relu(x)
 
 
-@with_unsupported_device_and_dtypes(
-    {"2.5.1 and below": {"cpu": ("bfloat16",)}}, backend_version
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float16", "uint16", "float32", "float64", "complex")},
```
vedpatwardhan, line 31: instead of `float16`, `float32` and `float64` you could just use `float`, as all float dtypes are supported.
Also, I think [`uint16`](https://github.com/unifyai/ivy/blob/47359a574231bd476f54fea63ee4fd8105d68991/ivy/functional/backends/paddle/__init__.py#L139) isn't supported in `paddle` 
Same applies to other similar instances in the `paddle` backend

ivy/functional/backends/paddle/data_type.py
```diff
@@ -156,15 +173,7 @@ def broadcast_to(
     if x.ndim > len(shape):
         x = x.reshape([-1])
 
-    if x.dtype in [
```
vedpatwardhan, line 159: In this case, given that the number of unsupported dtypes is lesser than the number of supported ones, we can just specify the unsupported ones using `with_unsupported_dtypes`. Same applies to the other similar instances in the `paddle` backend

vedpatwardhan (changes requested):
Hey @Madjid-CH, thanks for looking into this, just requested a couple of changes, but they also apply to all the other functions if applicable 😄

ivy/functional/backends/paddle/activations.py
Madjid-CH: Actually I looked into the functions source code to figure out the supported types, for example: [here](https://github.com/PaddlePaddle/Paddle/blob/daac3829d87d2a96b6015c946bcf6fa82a690fd2/python/paddle/nn/functional/activation.py#L458) I think is where the supported types checks if performed and uint16 is mentioned.

vedpatwardhan: Yeah it seems like they might've started supporting `uint16` in the later versions, @MahmoudAshraf97 what do you think? Thanks!

ivy/functional/backends/numpy/experimental/statistical.py
```diff
@@ -139,12 +139,7 @@ def median(
         keepdims=keepdims,
         out=out,
     )
-    if input.dtype in [np.uint64, np.int64, np.float64]:
```
vedpatwardhan, line 142: was this type-casting unnecessary? ig `numpy.median` was producing output of an incorrect dtype which is why this type casting was added? Not completely sure though

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -125,6 +123,10 @@ def equal(
     )
 
 
+@with_supported_dtypes(
+    {"2.5.1 and below": ("bool", "float32", "float64", "int32", "int64", "complex")},
```
vedpatwardhan, line 127: wouldn't it be easier to mark `int8` and `uint8` as unsupported instead?

vedpatwardhan (changes requested):
Hey @Madjid-CH, just requested a few more changes, the same also apply to other files where applicable. Thanks for diving into this 😄

ivy/functional/backends/numpy/experimental/statistical.py
Madjid-CH: > was this type-casting unnecessary? ig `numpy.median` was producing output of an incorrect dtype which is why this type casting was added? Not completely sure though

when we can say that the return type is incorrect ?

ivy/functional/backends/paddle/activations.py
```diff
@@ -13,35 +13,23 @@
...
-
-def relu(
-    x: paddle.Tensor, /, *, complex_mode="jax", out: Optional[paddle.Tensor] = None
-) -> paddle.Tensor:
-    if x.dtype in unsupported_dtypes:
-        if paddle.is_complex(x):
-            return paddle.complex(F.relu(x.real()), F.relu(x.imag()))
-        return F.relu(x.cast("float32")).cast(x.dtype)
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float", "uint16", "complex")},
```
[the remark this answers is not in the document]
Madjid-CH: sure, 
but I think I'm starting to get confused, what's the criteria for an invalid dtype ?

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: I'm not sure, If paddle actually supports them, shouldn't we prevent any unnecessary type casting ?

RickSanchezStoic: Sneaking in, haha, the role of the decorators is to specify the dtypes not supported by a given function but supported by the backend itself : in the case of `with_unsupported_dtypes`

In the case of `with_supported_dtypes`: the role of the decorators is to specify the ****only**** dtypes supported by a function.

So in this case, since the function doesn't support `int8` and `uint8` we provide that information through the `unsupported` dtype decorators. We don't specify information about the dtypes not supported by the backend at all because this information is redundant and is known already.

ivy/functional/backends/numpy/experimental/statistical.py
vedpatwardhan: It varies from function to function, but I think for `median`, the output `dtype` should intuitively be the same as the input `dtype`?

ivy/functional/backends/paddle/activations.py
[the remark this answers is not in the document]
vedpatwardhan: No worries! So basically there's 2 categories of unsupported dtypes, ones that are unsupported for the whole backend and ones that are unsupported for specific functions. Given that we can't even create a `uint16` array with the `paddle` backend, I think there's issues with using `uint16` with `paddle` which is why it was marked as an invalid `dtype` for the [backend](https://github.com/unifyai/ivy/blob/47359a574231bd476f54fea63ee4fd8105d68991/ivy/functional/backends/paddle/__init__.py#L139) altogether. Happy to clarify further if required 😄

ivy/functional/backends/paddle/elementwise.py
vedpatwardhan: > I'm not sure, If paddle actually supports them, shouldn't we prevent any unnecessary type casting ?

What I meant was that if the number of unsupported dtypes is less than the number of supported dtypes, it might be more convenient to use `with_unsupported_dtypes` instead of `with_supported_dtypes`. Either way would work, but just was curious about which ones more convenient.

vedpatwardhan: > Sneaking in, haha, the role of the decorators is to specify the dtypes not supported by a given function but supported by the backend itself : in the case of `with_unsupported_dtypes`
> 
> In the case of `with_supported_dtypes`: the role of the decorators is to specify the **only** dtypes supported by a function.
> 
> So in this case, since the function doesn't support `int8` and `uint8` we provide that information through the `unsupported` dtype decorators. We don't specify information about the dtypes not supported by the backend at all because this information is redundant and is known already.

That's true, but the point here is a bit different, regarding which one of `with_unsupported_dtypes` or `with_supported_dtypes` is convenient to use based on the number of dtypes that are supported or unsupported 😄

ivy/functional/backends/paddle/activations.py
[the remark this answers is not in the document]
Madjid-CH: Okay so the input of type `uint16` get converted anyway in paddle so we mark is as invalid and do the conversion ourselves if approved by the user.
If this is the case does that mean we should apply this principle to all backends and see which types get converted inside the backend and prevent the conversion ?

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: > wouldn't it be easier to mark `int8` and `uint8` as unsupported instead?

I think here the unsupported dtypes are `int8` `int16` `unsigned` `bfloat16` `float16` 😅

ivy/functional/backends/jax/manipulation.py
```diff
@@ -201,6 +201,7 @@ def tile(
     return jnp.tile(x, repeats)
 
 
+@with_unsupported_dtypes({"0.4.14 and below": "complex"})
```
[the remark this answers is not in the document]
Madjid-CH: I think this PR is starting to get out control 😅, any tips on how to revert this changes quickly ?

ivy/functional/backends/paddle/activations.py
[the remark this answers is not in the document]
vedpatwardhan: Actually, given that `uint16` is marked as the unsupported dtype for a backend altogether, the casting modes wouldn't cast to that `dtype`. The casting modes are only meant for dtypes that are unsupported for a specific function, but are still supported by the backend 😄

ivy/functional/backends/jax/manipulation.py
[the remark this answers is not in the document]
vedpatwardhan: Probably just revert it manually 😅

ivy/functional/backends/paddle/elementwise.py
vedpatwardhan: Oh okay, got it! yeah in this case it's 5 unsupported and 6 supported if we're talking about the number of elements to specify in the tuple which could also indicate specifying unsupported ones are more convenient 😅

ivy/functional/backends/jax/experimental/linear_algebra.py
```diff
@@ -2,6 +2,8 @@
 from typing import Optional, Tuple, Sequence, Union
 import jax.numpy as jnp
 import jax.scipy.linalg as jla
+
+from ivy import with_supported_dtypes
```
vedpatwardhan, line 6: let's probably do `from ivy.func_wrapper import with_supported_dtypes` following the general convention and remove the additional blank line

ivy/functional/backends/paddle/activations.py
```diff
@@ -13,35 +13,23 @@
...
+def relu(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
+    if paddle.is_complex(x):
+        return paddle.complex(F.relu(x.real()), F.relu(x.imag()))
     return F.relu(x)
 
 
-@with_unsupported_device_and_dtypes(
-    {"2.5.1 and below": {"cpu": ("bfloat16",)}}, backend_version
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float", "complex")},
```
vedpatwardhan, line 31: `bfloat16` is unsupported right? We'd have to replace `float` by `float16`, `float32`, `float64` in the decorator if that's the case.

ivy/functional/backends/paddle/activations.py
```diff
@@ -51,18 +39,17 @@ def leaky_relu(
...
+        return paddle.complex(
+            F.leaky_relu(x.real(), negative_slope=alpha),
+            F.leaky_relu(x.imag(), negative_slope=alpha),
+        )
     return F.leaky_relu(x, negative_slope=alpha)
 
 
-@with_unsupported_device_and_dtypes(
-    {"2.5.1 and below": {"cpu": ("bfloat16",)}}, backend_version
+@with_supported_dtypes(
```
vedpatwardhan, line 50: same applies here

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -53,10 +52,9 @@ def bitwise_xor(
     return paddle.bitwise_xor(x1, x2)
 
 
+@with_supported_dtypes({"2.5.1 and below": ("float",)}, backend_version)
 def expm1(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    if x.dtype in [paddle.float16, paddle.float32, paddle.float64]:
-        return paddle.expm1(x)
-    return paddle_backend.subtract(paddle_backend.exp(x), 1.0).astype(x.dtype)
```
vedpatwardhan, line 59: could you please test this out if the `astype` in the output is necessary? Probably we can consider the other dtypes as supported too.

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -126,6 +124,10 @@ def equal(
     )
 
 
+@with_unsupported_dtypes(
+    {"2.5.1 and below": ("int8", "int16", "bfloat16", "unsigned", "float16")},
```
vedpatwardhan, line 128: actually dtypes other then `int8` and `uint8` could be supported by this function as they don't seem to have any manual dtype casting in the implementation

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -156,95 +156,38 @@ def bitwise_and(
...
-)
+@with_supported_dtypes({"2.5.1 and below": "float"}, backend_version)
 def asinh(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    if x.dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.int64,
-        paddle.uint8,
-        paddle.float16,
```
vedpatwardhan, line 231: same as previous function for `float16`

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -254,170 +197,94 @@ def sign(
...
+    backend_version,
+)
 def log10(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    if x.dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.int64,
-        paddle.uint8,
-        paddle.float16,
```
vedpatwardhan, line 353: same doubt for `float16` as above

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -254,170 +197,94 @@ def sign(
...
+    backend_version,
+)
 def log2(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    if x.dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.int64,
-        paddle.uint8,
-        paddle.float16,
```
vedpatwardhan, line 374: `float16` doubt

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -254,170 +197,94 @@ def sign(
...
+    backend_version,
+)
 def log1p(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    if x.dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.int64,
-        paddle.uint8,
-        paddle.float16,
```
vedpatwardhan, line 395: same here for `float16`

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -254,170 +197,94 @@ def sign(
...
-        return paddle.isnan(x.astype("float32"))
+    if paddle.is_complex(x):
+        return paddle.logical_or(paddle.isnan(x.real()), paddle.isnan(x.imag()))
     return paddle.isnan(x)
 
 
+@with_unsupported_dtypes(
+    {
+        "2.5.1 and below": (
+            "int16",
```
vedpatwardhan, line 279: have you tested with `int16` if it's unsupported?
Also, `uint32` and `uint64` don't need to be explicitly specified as [invalid](https://github.com/unifyai/ivy/blob/57a3a50423771c561fa1fbbec7077155a6b4dd2e/ivy/functional/backends/paddle/__init__.py#L141) for the entire backend

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -444,48 +309,31 @@ def multiply(
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
     x1, x2, ret_dtype = _elementwise_helper(x1, x2)
-    if x1.dtype in [paddle.int8, paddle.int16, paddle.uint8, paddle.float16]:
```
vedpatwardhan, line 447: do `int8`, `int16`, `uint8` and `float16` work with this function without any manual type casting?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -444,48 +309,31 @@ def multiply(
...
-        x1, x2 = x1.astype("float32"), x2.astype("float32")
     return paddle.multiply(x1, x2).astype(ret_dtype)
 
 
-@with_unsupported_device_and_dtypes(
-    {"2.5.1 and below": {"cpu": ("bool", "bfloat16")}},
-    backend_version,
-)
+@with_supported_dtypes({"2.5.1 and below": "float"}, backend_version)
```
vedpatwardhan, line 315: as `bfloat16` was previously marked unsupported in the function, could you please check if it's supported now (as we've marked `float` as supported?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -444,48 +309,31 @@ def multiply(
...
 def cos(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    if x.dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.int64,
-        paddle.uint8,
-        paddle.float16,
```
vedpatwardhan, line 463: does `float16` work without any manual type casting (given that `float` is marked supported for the function in the decorator)?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -444,48 +309,31 @@ def multiply(
...
 
 
+@with_unsupported_dtypes(
+    {"2.5.1 and below": ("uint",)},
+    backend_version,
+)
 def logical_not(
     x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None
 ) -> paddle.Tensor:
-    if x.dtype in [paddle.uint8, paddle.float16, paddle.complex64, paddle.complex128]:
```
vedpatwardhan, line 480: does `float16` work without any manual type casting for the function?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -494,15 +342,11 @@ def divide(
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
     x1, x2, ret_dtype = _elementwise_helper(x1, x2)
-    if x1.dtype in [paddle.float16, paddle.bfloat16]:
-        x1, x2 = x1.astype("float32"), x2.astype("float32")
-    if not (ivy.is_float_dtype(ret_dtype) or ivy.is_complex_dtype(ret_dtype)):
-        ret_dtype = ivy.default_float_dtype(as_native=True)
```
vedpatwardhan, line 500: can we get `complex` dtypes to work in the same way as other functions for this function too (i.e. using `is_complex` and then calling the function for `real` and `complex`)?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -494,15 +342,11 @@ def divide(
...
     return (x1 / x2).astype(ret_dtype)
 
 
 @with_supported_dtypes(
-    {"2.5.1 and below": ("float64", "float32", "int64", "int64")},
+    {"2.5.1 and below": ("float", "int32", "int64")},
```
vedpatwardhan, line 349: `int32` wasn't originally present in the decorator, have you tested if it works for this function?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -517,6 +361,18 @@ def fmin(
...
 
 
+@with_supported_dtypes(
+    {
+        "2.5.1 and below": (
+            "bool",
+            "float",
+            "int32",
+            "int64",
+            "complex",
```
vedpatwardhan, line 371: have you tested if `int16` is supported for the function? it doesn't seem to be getting manually type casted in the original implementation

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -525,16 +381,26 @@ def greater(
...
```
vedpatwardhan, line 399: same doubt regarding `int16`

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -543,30 +409,16 @@ def greater_equal(
...
+            imag = paddle.greater_equal(x1.imag(), x2.imag())
+            return paddle.logical_and(real, imag)
     return paddle.greater_equal(x1, x2)
 
 
-@with_unsupported_device_and_dtypes(
-    {"2.5.1 and below": {"cpu": ("bool", "bfloat16")}},
-    backend_version,
-)
+@with_supported_dtypes({"2.5.1 and below": ("float", "complex")}, backend_version)
```
vedpatwardhan, line 420: is `bfloat16` supported by the function?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -578,75 +430,64 @@ def acos(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.
...
-    {"2.5.1 and below": {"cpu": ("complex64", "complex128")}},
+@with_supported_dtypes(
+    {
+        "2.5.1 and below": (
+            "bool",
+            "int8",
+            "int16",
+            "in32",
+            "in64",
+            "float",
```
vedpatwardhan, line 441: is `float16` supported?
also, can we get `complex64` and `complex128` working for this function?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -578,75 +430,64 @@ def acos(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.
...
```
vedpatwardhan, line 461: same question as the previous function

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -658,20 +499,8 @@ def acosh(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle
     return paddle.acosh(x)
 
 
-@with_unsupported_device_and_dtypes(
-    {"2.5.1 and below": {"cpu": ("bool", "bfloat16")}},
-    backend_version,
-)
+@with_supported_dtypes({"2.5.1 and below": ("float", "complex")}, backend_version)
```
vedpatwardhan, line 502: same questions as previous function

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -750,21 +583,8 @@ def bitwise_or(
...
 def sinh(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    if x.dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.int64,
-        paddle.uint8,
-        paddle.float16,
```
vedpatwardhan, line 764: does `float16` work without any manual type casting?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -784,21 +604,26 @@ def positive(
     return x.clone()
 
 
+@with_supported_dtypes(
+    {
+        "2.5.1 and below": (
+            "int32",
+            "int64",
+            "float",
```
vedpatwardhan, line 612: have you tested if `float16` and `bfloat16` work without type casting for this function?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -813,25 +638,22 @@ def pow(
         ret = paddle.pow(x1, x2)
         ret[inf_indices] = _complex_to_inf(ret[inf_indices])
         return ret
-    if x1.dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.uint8,
-        paddle.float16,
```
vedpatwardhan, line 820: does `float16` work?

vedpatwardhan (changes requested):
Hey @Madjid-CH, just went through a few files and requested changes. Given that the issues seem repetitive, I haven't reviewed any code beyond the last function I've commented on to avoid this review from getting too long. Could you please go through all the other functions you've modified and apply the same rationale to make necessary changes? Thanks 😄

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: there's a problem with this casting if the input `dtype `is an `int` because this `paddle_backend.subtract(paddle_backend.exp(x), 1.0)` will return a `float `and the `.astype(x.dtype)` will cause a loss in precision 
image
but if the input is float we get a correct answer:
image

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -156,95 +156,38 @@ def bitwise_and(
...
+    backend_version,
+)
 def ceil(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    x_dtype = x.dtype
-    if x_dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.uint8,
-        paddle.float16,
```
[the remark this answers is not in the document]
Madjid-CH: it seems that the supported dtypes for scalers and tensors are not the same
`float16` is not supported for tensors

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -156,95 +156,38 @@ def bitwise_and(
...
+    backend_version,
+)
 def floor(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    x_dtype = x.dtype
-    if x_dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.uint8,
-        paddle.float16,
```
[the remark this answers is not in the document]
Madjid-CH: yeap, not supported

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -156,95 +156,38 @@ def bitwise_and(
...
-)
+@with_supported_dtypes({"2.5.1 and below": "float"}, backend_version)
 def asin(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    if x.dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.int64,
-        paddle.uint8,
-        paddle.float16,
```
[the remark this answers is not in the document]
Madjid-CH: you're right `float16` is not supported, and I tried `bfloat16` and it's not supported also.

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -156,95 +156,38 @@ def bitwise_and(
...
-        ret_dtype = x.dtype
-        return paddle.asin(x.astype("float32")).astype(ret_dtype)
-    if paddle.is_complex(x):
-        asinh_iz = paddle_backend.asinh(paddle.complex(-x.imag(), x.real()))
-        return paddle.complex(asinh_iz.imag(), -asinh_iz.real())
     return paddle.asin(x)
 
 
-@with_unsupported_device_and_dtypes(
-    {"2.5.1 and below": {"cpu": ("bool", "bfloat16")}},
```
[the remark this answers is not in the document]
Madjid-CH: no `bfloat16` is not supported

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: yeap not supported

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -254,170 +197,94 @@ def sign(
     np_variant: Optional[bool] = True,
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
-    if x.dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.int64,
-        paddle.uint8,
-        paddle.float16,
```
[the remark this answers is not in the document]
Madjid-CH: I revised this function and they're not supported

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -254,170 +197,94 @@ def sign(
...
 def sqrt(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
     """Calculate the square root with type handling."""
-
     if paddle.is_complex(x):
         angle = paddle.angle(x)
         return paddle.complex(
             paddle.cos(angle / 2), paddle.sin(angle / 2)
         ) * paddle.sqrt(paddle.abs(x))
 
-    if x.dtype in {paddle.float32, paddle.float64}:
```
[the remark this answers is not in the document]
Madjid-CH: I tested them and they are not supported

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -254,170 +197,94 @@ def sign(
...
-)
+@with_supported_dtypes({"2.5.1 and below": "float"}, backend_version)
 def cosh(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
-    if x.dtype in [
-        paddle.int8,
-        paddle.int16,
-        paddle.int32,
-        paddle.int64,
-        paddle.uint8,
-        paddle.float16,
```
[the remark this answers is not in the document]
Madjid-CH: no it's not supported

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: nope

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: nope

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: yeah int16 is supported actually

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: yes they don't seem work, I guess I can't depend on the documentations or the source code since some type checking is no explicit.

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: no it's not

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: yeap you're right.

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: no `float16` does not work

Madjid-CH: yeah I can add the complex support.

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -578,75 +430,64 @@ def acos(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.
...
 
+@with_unsupported_dtypes(
+    {"2.5.1 and below": ("uint",)},
+    backend_version,
+)
 def logical_or(
     x1: paddle.Tensor, x2: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None
 ) -> paddle.Tensor:
     x1, x2, ret_dtype = _elementwise_helper(x1, x2)
-    if ret_dtype in [paddle.uint8, paddle.float16, paddle.complex64, paddle.complex128]:
```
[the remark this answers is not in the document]
Madjid-CH: unfortunately not

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -681,15 +510,13 @@ def sin(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.T
     return paddle.sin(x)
 
 
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float32", "float64", "int8", "int16", "int32", "int64")},
```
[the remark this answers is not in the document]
Madjid-CH: I think `int` get replaced with all unsigned types too.
[here](https://github.com/unifyai/ivy/blob/9edecd9a100fd75d1e97eed09cad7de8945cfccf/ivy/func_wrapper.py#L48C1-L48C1) and [here](https://github.com/unifyai/ivy/blob/9edecd9a100fd75d1e97eed09cad7de8945cfccf/ivy/__init__.py#L704C8-L704C8)

ivy/functional/backends/paddle/activations.py
```diff
@@ -186,17 +172,17 @@ def log_softmax(
...
-        return F.mish(x.cast("float32")).cast(x.dtype)
+    if paddle.is_complex(x):
+        return x * paddle_backend.tanh(paddle_backend.log1p(paddle_backend.exp(x)))
     return F.mish(x)
 
 
-@with_unsupported_device_and_dtypes(
-    {"2.5.1 and below": {"cpu": ("float16",)}}, backend_version
-)
+@with_supported_dtypes({"2.5.1 and below": ("float",)}, backend_version)
```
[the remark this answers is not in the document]
vedpatwardhan: I think it should be based on whether our implementation works with those dtypes than the actual framework's implementation, because we might've added it due to tests failing or something which might not exactly be a native unsupported dtype error but some other error message 😄

ivy/functional/backends/paddle/data_type.py
```diff
@@ -5,7 +5,9 @@
 import ivy.functional.backends.paddle as paddle_backend
 import numpy as np
 import ivy
+from ivy import with_unsupported_dtypes
```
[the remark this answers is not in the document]
vedpatwardhan: I don't think we need to force it, but just use one consistent style across the codebase whichever we might choose 😄

ivy/functional/backends/paddle/elementwise.py
vedpatwardhan: As I mentioned in the other discussion, our frame of reference when adding unsupported dtypes for the first time should be the actual framework, but for updating the existing ones in the decorator we should also consider that the dtype might not be actually marked as supported but the tests might've showed that it is and so on. This helps us utilize the work someone else might've put before in adding those dtypes at the start. So if the ivy tests were originally running with `int8` and `uint8` and were passing, then we can consider those dtypes as supported even if that doesn't seem to be marked in the source code. Based on the results of the tests before making these changes, feel free to make the decision of whether or not to add those dtypes. Thanks!

ivy/functional/backends/paddle/elementwise.py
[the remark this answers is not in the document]
vedpatwardhan: sounds good, thanks for updating it! Please also update other similar occurrences of `float16` in unsupported by `float` in the supported decorator, thanks 😄

vedpatwardhan (changes requested):
Hey @Madjid-CH, feel free to request a review once these changes are done, I'll then go through the functions after where I stopped yesterday. Thanks 😄

ivy/functional/backends/paddle/data_type.py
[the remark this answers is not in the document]
Madjid-CH: understood 😄

Madjid-CH:
> Hey @Madjid-CH, just went through a few files and requested changes. Given that the issues seem repetitive, I haven't reviewed any code beyond the last function I've commented on to avoid this review from getting too long. Could you please go through all the other functions you've modified and apply the same rationale to make necessary changes? Thanks 😄

yeah sure, I think next time when a task require a lot of changes in many files we should consider splitting the changes to multiple PRs. what do you think ?

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: understood, I'll keep that in mind.
for this specific example: I tried `int8 `and `uint8` and they are not supported

Madjid-CH:
> > Hey @Madjid-CH, just went through a few files and requested changes. Given that the issues seem repetitive, I haven't reviewed any code beyond the last function I've commented on to avoid this review from getting too long. Could you please go through all the other functions you've modified and apply the same rationale to make necessary changes? Thanks 😄
> 
> yeah sure, I think next time when a task require a lot of changes in many files we should consider splitting the changes to multiple PRs. what do you think ?

hey @vedpatwardhan after I have a thought about it, let's just keep the changes that you already reviewed and revert the rest and will continue the work in another PRs. 
sounds good ?

vedpatwardhan:
> yeah sure, I think next time when a task require a lot of changes in many files we should consider splitting the changes to multiple PRs. what do you think ?

yep that sounds good 👍

vedpatwardhan:
> hey @vedpatwardhan after I have a thought about it, let's just keep the changes that you already reviewed and revert the rest and I will continue the work in another PRs. sounds good ?

yep 👍

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: sure

ivy/functional/backends/paddle/activations.py
```diff
@@ -13,35 +13,23 @@
...
-
-def relu(
-    x: paddle.Tensor, /, *, complex_mode="jax", out: Optional[paddle.Tensor] = None
-) -> paddle.Tensor:
-    if x.dtype in unsupported_dtypes:
-        if paddle.is_complex(x):
-            return paddle.complex(F.relu(x.real()), F.relu(x.imag()))
-        return F.relu(x.cast("float32")).cast(x.dtype)
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float", "complex")},
```
vedpatwardhan, line 21: I think `float16` is also not supported by `F.relu`

ivy/functional/backends/paddle/activations.py
```diff
@@ -73,34 +60,30 @@ def gelu(
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
     if paddle.is_complex(x):
-        sqrt_2_over_pi = 0.7978845608
-        # the other magic number comes directly from the formula in
```
vedpatwardhan, line 77: could you please add this comment back?

ivy/functional/backends/paddle/activations.py
```diff
@@ -73,34 +60,30 @@ def gelu(
...
+                * (1 + paddle_backend.tanh(0.7978845608 * (x + 0.044715 * x * x * x)))
+            )
+        return 0.5 * x * (1 + paddle_backend.erf(x / paddle_backend.sqrt(2)))
     return F.gelu(x, approximate=approximate)
 
 
-@with_unsupported_device_and_dtypes(
-    {"2.5.1 and below": {"cpu": ("bfloat16",)}}, backend_version
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float16", "float32", "float64", "complex")},
```
vedpatwardhan, line 74: `float16` doesn't seem to be supported by the function

ivy/functional/backends/paddle/data_type.py
```diff
@@ -138,6 +140,17 @@ def broadcast_arrays(*arrays: paddle.Tensor) -> List[paddle.Tensor]:
     return result
 
 
+@with_unsupported_dtypes(
+    {
+        "2.5.1 and below": (
+            "uint8",
+            "int8",
+            "int16",
+            "bfloat16",
```
vedpatwardhan, line 149: `float16` is also unsupported by the function

ivy/functional/backends/paddle/elementwise.py
vedpatwardhan: But that's the expected precision anyway right if the output dtype is expected to be `int16` because of the input dtypes that were passed?

ivy/functional/backends/paddle/elementwise.py
vedpatwardhan: sounds good, thanks!

vedpatwardhan (changes requested):
Just requested a few more changes, happy to take another look once these are done, thanks @Madjid-CH 😄

ivy/functional/backends/paddle/activations.py
Madjid-CH: yeah, the paddle docs are misleading, from now on I'll use brute force dtype checking

ivy/functional/backends/paddle/data_type.py
Madjid-CH: that's right

ivy/functional/backends/paddle/activations.py
Madjid-CH: sure

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: I'm not sure if  want to allow this kind of behavior, If you want me to keep these dtypes I'll do it ?

ivy/functional/backends/paddle/data_type.py
```diff
@@ -138,6 +140,17 @@ def broadcast_arrays(*arrays: paddle.Tensor) -> List[paddle.Tensor]:
     return result
 
 
+@with_unsupported_dtypes(
+    {
+        "2.5.1 and below": (
+            "uint8",
+            "int8",
+            "int16",
+            "float16bfloat16",
```
vedpatwardhan, line 149: typo

ivy/functional/backends/paddle/elementwise.py
vedpatwardhan: So just to understand your point, we're basically trying to compute `expm1` with an `int` to start with in the example you shared right? What's the output dtype with the other backends? The array api says that both the inputs and outputs should have a `float` dtype. In the ivy tests we're only testing this function with [float and complex](https://github.com/unifyai/ivy/blob/c26a5c80fd919a241babbe9e69cc5bf9145ed5bd/ivy_tests/test_ivy/test_functional/test_core/test_elementwise.py#L771) numbers which is of course not correct.
Do you think it's better for us to mark `int` as unsupported and cast them to `float` when the casting modes are used? Or do you think we should pass `int` to the function and return a `float` output? Thanks!

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -153,95 +162,55 @@ def bitwise_and(
...
+    if paddle.is_complex(x):
+        return paddle.complex(paddle.floor(x.real()), paddle.floor(x.imag()))
     return paddle.floor(x)
 
 
-@with_unsupported_device_and_dtypes(
-    {"2.5.1 and below": {"cpu": ("bool", "bfloat16")}},
+@with_supported_dtypes(
+    {
+        "2.5.1 and below": (
```
vedpatwardhan, line 187: we could probably format these inputs to a single line like before as there's only 2 dtypes specified. Same applies to other changes where the length of the single line statement is less than `88` characters

ivy/functional/backends/paddle/elementwise.py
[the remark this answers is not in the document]
vedpatwardhan: Ah makes sense, thanks for clarifying!
Also, there seems to be an inconsistency here, the second link you shared considers `int` and `uint` dtypes to fall under the `"int"` category, while in the first link only `int` dtypes fall under the `"int"` category, which one of these do you think is better @RickSanchezStoic?

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -734,11 +565,13 @@ def floor_divide(
...
 ) -> paddle.Tensor:
     x1, x2, ret_dtype = _elementwise_helper(x1, x2)
-    if x1.dtype in [paddle.int32, paddle.int64]:
-        return paddle.floor_divide(x1, x2)
-    return paddle_backend.floor(paddle_backend.divide(x1, x2)).astype(ret_dtype)
+    return paddle.floor_divide(x1, x2)
 
 
+@with_supported_dtypes(
+    {"2.5.1 and below": ("bool", "uint8", "int8", "int16", "int32", "int64")},
```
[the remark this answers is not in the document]
vedpatwardhan: yep, let's continue discussing this in the previous conversation

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -1365,24 +1184,7 @@ def gcd(
...
-                "float16",
-                "float32",
-                "float64",
-                "bool",
-            )
-        }
-    },
-    backend_version,
-)
+@with_supported_dtypes({"2.5.1 and below": "complex"}, backend_version)
```
vedpatwardhan, line 1187: shouldn't this be a `tuple`? Could you please also update similar occurrences in the changes made?

vedpatwardhan (changes requested):
Hey @Madjid-CH, just requested a few changes and raised a couple of points of discussion, thanks 😄

MahmoudAshraf97:
Hi @vedpatwardhan and @Madjid-CH , please note that some of the manual casting found in this PR is not aimed at adding support for unsupported dtypes and hence should be left in place because it has other functionality  (ensuring consistency for between backends for example), most of the manual casting to add extra support is only found in paddle backend

these 2 commits should be reverted and reviewed first:
https://github.com/unifyai/ivy/commit/282bffcfad41dd6a4868375e4d24ec6314fd1d08
https://github.com/unifyai/ivy/commit/cfcfbea2270cd9b3fb54fd16cdfd9118d22cc008

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: understood

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: yes, I'm trying to pass an `int` type as an input.
tensorflow does not support `int` types for this function.
jax and numpy and torch cast the output to `float32`:
```python
import jax.numpy as jnp
x = jnp.array([1, 1, 1], dtype=jnp.int16)
jnp.expm1(x)
Out[11]: Array([1.7182817, 1.7182817, 1.7182817], dtype=float32)
```
```python
x = np.array([1, 1, 1], dtype=np.int16)
np.expm1(x)
Out[16]: array([1.7182817, 1.7182817, 1.7182817], dtype=float32)
```
```python
import torch
x = torch.tensor([1, 1, 1], dtype=torch.int16)
torch.expm1(x)
Out[20]: tensor([1.7183, 1.7183, 1.7183])
_.dtype
Out[21]: torch.float32
```
so the only problem is with tensorflow and paddle(precision problem) which can be resolved by the dtype decorators for tensorflow. And paddle I prefer to raise error or mark them to unsupported instead of the precision loss (ideally we would ask the client what he really wants), because I think the expected return type is `float` even when the input is an `int`. For the other backends that does support an `int` type, I'd say to mark the int as supported since they behave correctly in my view.
what do you think ??

ivy/functional/backends/paddle/elementwise.py
Madjid-CH: I think python consider tuple with 1 element to be the same as 1 element if that makes sense ?

Madjid-CH:
> Hi @vedpatwardhan and @Madjid-CH , please note that some of the manual casting found in this PR is not aimed at adding support for unsupported dtypes and hence should be left in place because it has other functionality (ensuring consistency for between backends for example), most of the manual casting to add extra support is only found in paddle backend
> 
> these 2 commits should be reverted and reviewed first: [282bffc](https://github.com/unifyai/ivy/commit/282bffcfad41dd6a4868375e4d24ec6314fd1d08) [cfcfbea](https://github.com/unifyai/ivy/commit/cfcfbea2270cd9b3fb54fd16cdfd9118d22cc008)

understood

Madjid-CH:
> > Hi @vedpatwardhan and @Madjid-CH , please note that some of the manual casting found in this PR is not aimed at adding support for unsupported dtypes and hence should be left in place because it has other functionality (ensuring consistency for between backends for example), most of the manual casting to add extra support is only found in paddle backend
> > these 2 commits should be reverted and reviewed first: [282bffc](https://github.com/unifyai/ivy/commit/282bffcfad41dd6a4868375e4d24ec6314fd1d08) [cfcfbea](https://github.com/unifyai/ivy/commit/cfcfbea2270cd9b3fb54fd16cdfd9118d22cc008)
> 
> understood

@MahmoudAshraf97 can you please explain how to differentiate between a manual cast for adding support and a manual cast for backends compatibility because my approach is to test the real backend function against all available dtypes  and see which dtypes behave correctly (same input same ouput dtype, not throwing error, ...)

MahmoudAshraf97:
> > > Hi @vedpatwardhan and @Madjid-CH , please note that some of the manual casting found in this PR is not aimed at adding support for unsupported dtypes and hence should be left in place because it has other functionality (ensuring consistency for between backends for example), most of the manual casting to add extra support is only found in paddle backend
> > > these 2 commits should be reverted and reviewed first: [282bffc](https://github.com/unifyai/ivy/commit/282bffcfad41dd6a4868375e4d24ec6314fd1d08) [cfcfbea](https://github.com/unifyai/ivy/commit/cfcfbea2270cd9b3fb54fd16cdfd9118d22cc008)
> > 
> > 
> > understood
> 
> @MahmoudAshraf97 can you please explain how to differentiate between a manual cast for adding support and a manual cast for backends compatibility because my approach is to test the real backend function against all available dtypes and see which dtypes behave correctly (same input same ouput dtype, not throwing error, ...)

there is no solid diffrentiation between the two types, but as I said the casting for compatibility is mostly found in paddle backend

vedpatwardhan:
> Hi @vedpatwardhan and @Madjid-CH , please note that some of the manual casting found in this PR is not aimed at adding support for unsupported dtypes and hence should be left in place because it has other functionality (ensuring consistency for between backends for example), most of the manual casting to add extra support is only found in paddle backend
> 
> these 2 commits should be reverted and reviewed first: [282bffc](https://github.com/unifyai/ivy/commit/282bffcfad41dd6a4868375e4d24ec6314fd1d08) [cfcfbea](https://github.com/unifyai/ivy/commit/cfcfbea2270cd9b3fb54fd16cdfd9118d22cc008)

Yeah you're right! There are cases where we apply casting to the output to ensure consistency, could you please share some examples in those commits where this is being done? Thanks @MahmoudAshraf97 😄

ivy/functional/backends/paddle/elementwise.py
vedpatwardhan: got it, thanks for clarifying! So you mean that `float32` should be the expected `dtype` of the output when an `int` is passed to `expm1`? Sounds good to me, as long as we're consistent and the behaviour is intuitive 👍

ivy/functional/backends/paddle/elementwise.py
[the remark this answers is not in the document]
vedpatwardhan: could you please take a look at this too?

ivy/functional/backends/paddle/elementwise.py
vedpatwardhan: Thanks for fixing this

vedpatwardhan (changes requested):
Hey @Madjid-CH, could you please check through if there's any unresolved conversations in the previous reviews as well? Thanks 😄

Madjid-CH:
> Hey @Madjid-CH, could you please check through if there's any unresolved conversations in the previous reviews as well? Thanks 😄

sure Ved 😃

vedpatwardhan (approved):
lgtm! Let's wait for @MahmoudAshraf97 to give it another look, thanks @Madjid-CH 😄

vedpatwardhan:
ivy-gardener

MahmoudAshraf97 (changes requested):
LGTM, you should keep the dtypes decorator as is because cpu and gpu support are different in paddle

Madjid-CH:
> LGTM, you should keep the dtypes decorator as is because cpu and gpu support are different in paddle

understood

vedpatwardhan:
Hey @MahmoudAshraf97, could you please review this PR so that we can get it merged soon? Thanks 😄

PR #22747, opened 2023-08-29
refactor(ivy): wrapper for dtype handling in paddle backend functions

AnnaTz:
Suggests a wrapper that will handle the dtype "back and forth" that needs to be done in many paddle backend functions. 
This will optimize memory consumption, as it picks the "smallest" possible dtype to cast to rather than always casting to 'float32'.
It will also shorten the code significantly, making it easier to maintain.


AnnaTz:
If the wrapper gets greenlighted I will add a test, and refactor all the functions to use it. Also, it will probably need a better name 😅

AnnaTz:
If the wrapper gets greenlighted, I will add a test for it and refactor all the functions to use it. 
Also, it will probably need a better name 😅

ivy-leaves:

Thank you :hugs:

AnnaTz:
> Hey @AnnaTz, I think the changes you've made should ideally be done through the dtype casting modes anyway. Also, given that we shouldn't have any dtype casting in the backend implementations unless a dtype casting mode was used (to avoid loosing precision or consuming more memory) [this](https://github.com/unifyai/ivy/pull/22700) PR is actually removing all of that anyway 😅 So as a result, after [this](https://github.com/unifyai/ivy/pull/22700) PR is merged, all those dtypes will be marked as unsupported by the function when using the `paddle` backend, and then the users can set particular [casting modes](https://unify.ai/docs/ivy/overview/deep_dive/data_types.html#data-type-casting-modes) if they want to still use those dtypes. Hope that makes sense, thanks for looking into this though 😄

Hi @vedpatwardhan, even better then! Are these casting modes a new feature? I wonder how I missed it 😅

vedpatwardhan:
> Hi @vedpatwardhan, even better then! Are these casting modes a new feature? I wonder how I missed it 😅

It's been there for a while, but probably we need to highlight it in a better way in the docs 😅

PR #22750, opened 2023-08-29
Add center crop

### LINKED ISSUE #22749: center_crop
- [ ]15046

alec-kr:
# PR Description 

This PR adds the `center_crop` function from PaddlePaddle to the Ivy repository frontend and includes the corresponding test.

## Related Issue 

Close #22749

### DIFF
--- a/ivy/functional/frontends/paddle/vision/transforms.py
+++ b/ivy/functional/frontends/paddle/vision/transforms.py
@@ -170,6 +170,21 @@ def normalize(img, mean, std, data_format="CHW", to_rgb=False):
         raise ValueError("Unsupported input format")
 
 
+@with_supported_dtypes({"2.5.1 and below": ("float32", "float64")}, "paddle")
+@to_ivy_arrays_and_back
+def center_crop(img, output_size):
+    h, w = ivy.shape(img)[-2:]
+
+    if isinstance(output_size, int):
+        output_size = (output_size, output_size)
+        
+    th, tw = output_size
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/paddle/vision/transforms.py
```diff
@@ -170,6 +170,21 @@ def normalize(img, mean, std, data_format="CHW", to_rgb=False):
...
+    h, w = ivy.shape(img)[-2:]
+
+    if isinstance(output_size, int):
+        output_size = (output_size, output_size)
+        
+    th, tw = output_size
+    i = (h - th) // 2
+    j = (w - tw) // 2
+
+    return img[..., i:i+th, j:j+tw]
```
vaithak, line 185: HI @alec-kr, you should also handle the case when output_size is greater than shape.

vaithak: Can you verify that this is consistent with paddle's output? Because torch seems to handle this in a different way, by padding the image till the size matches the output. 
Also, one test seems to be failing for such a case because of difference with ground truth output.

Ishticode (changes requested):
Hi @alec-kr 
The tests are failing with error `E   AssertionError: returned shape = (3, 3, 3), ground-truth returned shape = (3, 1, 1)` with numpy backend for example.
Could you look into a fix and request a review when done. Let me know if you are unsure about any aspect. Thanks 🙂️

PR #22753, opened 2023-08-29
feat: implemented "trace" Instance Method for numpy

### LINKED ISSUE #18013: trace

Nwaamaka-Iduwe:
Fixes #18013


YushaArif99 (changes requested):
Hi @Nwaamaka-Iduwe 

The changes look good overall. However, it seems the CI tests are failing [here](https://github.com/unifyai/ivy/actions/runs/6038182486/job/16384000078?pr=22753#step:3:153)
Could you kindly look into this and re-request a review once all relevant tests are passing? Thanks! 🙂

Nwaamaka-Iduwe:
Hi @YushaArif99, I made the changes as requested but I keep getting the error “trace() got an unexpected keyword argument 'dtype'” which is strange because trace indeed has the parameter “dtype” and I made sure to include this when defining the function “trace”.

YushaArif99 (changes requested):
Hi @Nwaamaka-Iduwe 

I think you're misunderstanding the point of the frontends. They serve as an intermediary to map native framework functions to our Ivy API functions. Not *every* argument will **always** be used. Instead, you'll have to work with what is supported by the corresponding Ivy function.

With that in mind, the signature of trace would be exactly that of `numpy.trace` and you're right that it includes the `dtype` argument. However, you're then calling `np_frontend.trace` and if you look at its signature (https://github.com/unifyai/ivy/blob/main/ivy/functional/frontends/numpy/linalg/norms_and_other_numbers.py#L60), its clear that there is no `dtype` argument here. And hence you're getting:
```python
 E   TypeError: trace() got an unexpected keyword argument `dtype`
``` 
Hope that makes sense 🙂

Nwaamaka-Iduwe:
Hi @YushaArif99, thank you for providing this detailed explanation. It seems like I misunderstood a part of the purpose of frontends here. I appreciate your patience in clarifying this and based on your explanation, I was able to refer to the docs and can see why the `dtype` argument was causing an error in this context. I've made the corresponding changes and the test now passes for Numpy specifically.

YushaArif99 (changes requested):
Hi @Nwaamaka-Iduwe 

Thanks for making the requested changes. I am still noticing some failures in the CI [here](https://github.com/unifyai/ivy/actions/runs/6114766034/job/16597016466?pr=22753#step:3:258). Could you kindly try fixing the failing tests before requesting a review since that greatly helps with the reviewing process? Thanks @Nwaamaka-Iduwe 😄 

PS: 
> I've made the corresponding changes and the test now passes for Numpy specifically.

the goal should be to pass **all** tests not just of Numpy 😅

Nwaamaka-Iduwe:
Hi @YushaArif99, I can now confirm that all tests pass on my end, including lint and PR 😊. I’ve also had a detailed look at the GitHub results and none related to the issue I worked on is listed. Looking forward to your review.

ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py
```diff
@@ -3671,3 +3671,50 @@ def test_numpy_ndarray_view(
...
     )
+
+
+# trace
+@handle_frontend_method(
+    class_tree=CLASS_TREE,
+    init_tree="numpy.array",
+    method_name="trace",
+    dtype_x_axis=helpers.dtype_values_axis(
+        available_dtypes=["int64"],
```
YushaArif99, line 3682: Hi @Nwaamaka-Iduwe 

The data generation for the test seems a bit restrictive to me. The `dtype` of the array should be `valid` or `numeric` since `np.trace` **can** accept floating point arrays as well so not sure why you've only limited it to `int64` 😅

YushaArif99 (changes requested):
Hi @Nwaamaka-Iduwe 

Thanks a lot for making the requested changes. Could you kindly address the above-mentioned point and make sure the testing is more rigorous/comprehensive? Once all tests are passed with a broader test set, feel free to re-request a review. I think we can then merge this PR 🙂

ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py
Nwaamaka-Iduwe: Hi, the problem here is that other tests keep failing if I don't restrict it. Only the numpy tests passes using only "numeric" and "valid". I've tried to use lambda to solve this and indexing but none of them work. The best I can get is an 80% pass on tests. If you have any suggestions, that'll be appreciated.

Nwaamaka-Iduwe: Hi, the problem here is that other tests keep failing if I don't restrict it. Only the numpy tests passes using only "numeric" and "valid". I've tried to use lambda to solve this and indexing but none of them work. The best I can get is an 80% pass on tests. If you have any suggestions, that'll be appreciated.

YushaArif99: Do you know the reason why they were failing though? I just looked at the failing tests in a previous commit and the error seems to be a `dtype-mismatch` problem: 
```python
E   AssertionError: returned dtype = int64, ground-truth returned dtype = uint64
E   Falsifying example: test_numpy_trace(
E       backend_fw='jax',
E       frontend='numpy',
E       on_device='cpu',
E       dtype_x_axis=(['uint16'], [array([[0]], dtype=uint16)], (0, 1)),
E       offset=0,
```
Seems like the frontend function is returning an `int64` whereas the original `np.trace` is returning `uint64`. Not sure why `ivy.trace` is returning an incorrect dtype though🤔. Could you try debugging this further to figure out the exact cause?

Nwaamaka-Iduwe: Hi @YushaArif99, this is what I was able to get from debugging the code: 

WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1695749278.850075   35972 tfrt_cpu_pjrt_client.cc:349] TfrtCpuClient created.
No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
/workspaces/ivy/ivy/utils/exceptions.py:390: UserWarning: The current backend: 'jax' does not support inplace updates natively. Ivy would quietly create new arrays when using inplace updates with this backend, leading to memory overhead (same applies for views). If you want to control your memory management, consider doing ivy.set_inplace_mode('strict') which should raise an error whenever an inplace update is attempted with this backend.
  warnings.warn(
/workspaces/ivy/ivy/utils/exceptions.py:390: UserWarning: The current backend: 'tensorflow' does not support inplace updates natively. Ivy would quietly create new arrays when using inplace updates with this backend, leading to memory overhead (same applies for views). If you want to control your memory management, consider doing ivy.set_inplace_mode('strict') which should raise an error whenever an inplace update is attempted with this backend.
  warnings.warn(
/workspaces/ivy/ivy/utils/exceptions.py:390: UserWarning: The current backend: 'paddle' does not support inplace updates natively. Ivy would quietly create new arrays when using inplace updates with this backend, leading to memory overhead (same applies for views). If you want to control your memory management, consider doing ivy.set_inplace_mode('strict') which should raise an error whenever an inplace update is attempted with this backend.
  warnings.warn(
[2023-09-26 17:28:12,182] [ WARNING] data_type.py:159 - This function includes the mixed partial function 'ivy.get_item'. Please note that the returned data types may not be exhaustive. Please check the dtypes of `ivy.get_item` for more details
[2023-09-26 17:28:12,210] [ WARNING] data_type.py:159 - This function includes the mixed partial function 'ivy.get_item'. Please note that the returned data types may not be exhaustive. Please check the dtypes of `ivy.get_item` for more details
[2023-09-26 17:28:12,226] [ WARNING] data_type.py:159 - This function includes the mixed partial function 'ivy.get_item'. Please note that the returned data types may not be exhaustive. Please check the dtypes of `ivy.get_item` for more details
[2023-09-26 17:28:12,577] [ WARNING] data_type.py:159 - This function includes the mixed partial function 'ivy.flatten'. Please note that the returned data types may not be exhaustive. Please check the dtypes of `ivy.flatten` for more details

PR #22802, opened 2023-08-30
fix: resolve device mismatch in where with scalars

jshepherd01:
Previously, the `ivy.where` function would fail on the torch backend when it was called with one (CPU) array input and one scalar, since the scalar would be automatically placed on the GPU by `handle_array_like_without_promotion`. This commit modifies that wrapper so that, if a function is passed at least one array, all new arrays created by the wrapper will be placed on the same device as that array.


vedpatwardhan (commented):
Hey @jshepherd01, thanks for creating the PR. @sherry30 can we handle this through the device handling decorators? Of course, this specific issue wouldn't appear as we change the default device to `cpu` when soft device isn't enabled, but then it would fail if the array was on a `gpu` to start with and we converted the scalar to a `cpu` array. Thanks 😄

jshepherd01:
I think the two places that make sense to handle this are `handle_array_like` and `handle_device_shifting`. The reason I chose the first one:

- `handle_array_like` gets some arrays and some scalars as inputs, so it knows which arguments it should be modifying, and it knows which device is in use because all of the arrays should be on the same one (we leave it to the user to ensure that). `handle_device_shifting` just gets a bunch of arrays as inputs, which may be split across multiple devices, so it doesn't know which ones it needs to change.
- `handle_array_like` creates new arrays, and it makes more sense imo to create them on the correct device immediately rather than creating them in one place and moving them later.
- `handle_device_shifting` already contains logic that raises an error if two arrays are on different devices. That's there for some very good reasons, and if we were going to put this process there we'd be replacing that with just quietly moving the arrays which the user might not expect.

vedpatwardhan:
> I think the two places that make sense to handle this are `handle_array_like` and `handle_device_shifting`. The reason I chose the first one:
> 
> * `handle_array_like` gets some arrays and some scalars as inputs, so it knows which arguments it should be modifying, and it knows which device is in use because all of the arrays should be on the same one (we leave it to the user to ensure that). `handle_device_shifting` just gets a bunch of arrays as inputs, which may be split across multiple devices, so it doesn't know which ones it needs to change.
> * `handle_array_like` creates new arrays, and it makes more sense imo to create them on the correct device immediately rather than creating them in one place and moving them later.
> * `handle_device_shifting` already contains logic that raises an error if two arrays are on different devices. That's there for some very good reasons, and if we were going to put this process there we'd be replacing that with just quietly moving the arrays which the user might not expect.

Yep makes sense! let's see what @sherry30 thinks, thanks for clarifying @jshepherd01 😄

ivy/func_wrapper.py
```diff
@@ -379,6 +379,19 @@ def _handle_array_like_without_promotion(*args, **kwargs):
         parameters = list(type_hints.keys())
         annotations = [param.annotation for param in type_hints.values()]
 
+        # When new arrays are created, they should be created on the same device as
+        # existing array inputs. If a device is specified as a kwarg, create them there
+        # If not, scan for any other inputs which are already arrays and use the device
+        # of the first one found.
+        device = None
+        if "device" in kwargs:
```
sherry30, line 387: should also add `and kwargs["device"] is not None`

sherry30 (changes requested):
I agree that they should be initialized on the correct device to begin with.
But we have to cover all the cases in that case.

for instance when `soft_device` mode is enabled, we would initiate the array on default device if the `device` argument wasn't passed.
for `soft_device` mode disabled, we would initialize it on the device the input arrays are on unless the `device` argument was passed, if there are no arrays and no device argument was passed then we have to do it on the `default_device`

Would be nice to have a private function for determining what is the correct device, can also use that in the `handle_device_shifting` decorator as well.

sherry30:
The priority order for when soft device is enabled is slightly different.
when it is enabled: 
dev arg > default_device
when disabled:
dev arg > device arrays are on > default device

If these things are accounted for then its good to go. Otherwise would be cleaner to have a private function for that

jshepherd01:
Ah, I see. The docs for that will probably need changing then since I think they have the priority list mixed up between the two modes. I'll make the change

ivy/func_wrapper.py
```diff
@@ -293,6 +293,21 @@ def _check_in_nested_sequence(sequence, value=None, _type=None):
...
+def _get_preferred_device(args, kwargs):
+    # When new arrays are created, they should be created on the same device as
+    # existing array inputs. If a device is specified as a kwarg, create them there.
+    # If not, scan for any other inputs which are already arrays and use the device
+    # of the first one found (unless we're in soft device mode).
+    device = None
+    if "device" in kwargs and kwargs["device"] is not None:
+        return device
+    if not ivy.soft_device_mode:
+        for arg in args + list(kwargs.values()):
```
sherry30, line 305: I think a better way to do this is by using `_get_first_array` which uses `nested_argwhere`. Since the input arrays can be passed in a list.
See implementation of `infer_device` for reference.

sherry30 (changes requested):
sorry about the docs, that's my bad 🤣 
Thanks a lot for fixing it.
Requested 1 change, the rest lgtm.
Feel free to merge it after that.

PR #22804, opened 2023-08-30
(fix): found a fix to connect conv2d transpose to backend implementation, removed some unecessary nodes

OmarJabri7:
# PR Description
Fixed issue with frontend conv2d not connecting to backend implementation (torch), fixed some unnecessary nodes being transpired into the graph

## Checklist 

- [x] Did you add a function?
- [ ] Did you add the tests?


ivy/functional/backends/tensorflow/layers.py
```diff
@@ -452,9 +452,7 @@ def conv_general_transpose(
             axis=-1,
         )
     elif dims == 2:
-        res = tf.concat(
-            [
-                conv2d_transpose(
+        res = conv2d_transpose(
```
AnnaTz, line 455: If the test passes this way, you can do the same for the 1D and 3D cases as well.

ivy/functional/frontends/torch/nn/functional/convolution_functions.py
```diff
@@ -78,7 +76,6 @@ def _conv_transpose(
     ret = ret[unpad_slice]
     return ret
```
AnnaTz, line 78: I don't see a new failure on this so I guess it's okay. But is it beneficial in any way? Does `conv_general_transpose` have fewer nodes then `conv_general_dilated`?

ivy/functional/backends/tensorflow/layers.py
```diff
@@ -452,9 +452,7 @@ def conv_general_transpose(
...
                     x[..., j : j + filters.shape[-2] // feature_group_count],
```
AnnaTz, line 456: Actually I see the test failing for the tensorflow backend in the CI checks. Not sure how you can modify this to work correctly.

AnnaTz (changes requested):

ivy/functional/frontends/torch/nn/functional/convolution_functions.py
```diff
@@ -53,21 +53,19 @@ def _conv_transpose(
...
     padding, output_padding = map(
         lambda x: [x] * dims if isinstance(x, int) else x, [padding, output_padding]
     )
-    pad_widths = [(weight.shape[i] - 1,) * 2 for i in range(dims)]
-    ret = ivy.conv_general_dilated(
+    pad_widths = [(0,) * 2 for i in range(dims*2)]
+    ret = ivy.conv_general_transpose(
         input,
         weight,
         1,
```
sherry30, line 63: since we're using `conv_general_transpose` now, the arguments will should be the same as the input now, so should pass `stride` here now.

ivy/functional/frontends/torch/nn/functional/convolution_functions.py
```diff
@@ -53,21 +53,19 @@ def _conv_transpose(
...
         pad_widths,
         dims=dims,
         data_format="channel_first",
         feature_group_count=groups,
-        x_dilations=stride,
+        dilations=stride,
```
sherry30, line 68: dilations should take dilation from the input

ivy/functional/backends/tensorflow/layers.py
sherry30, line 456: If the native tf function supports group convolutions then you would just have to pass in `x` and `filters` without doing any extra indexing.

sherry30 (changes requested):
A few changes requested

ivy/functional/frontends/torch/nn/functional/convolution_functions.py
OmarJabri7: It is just that conv general dilated does not use conv2dtranspose, so want to see if I map, ideally should be one to one mapping (torch.conv2dtranspose to tf.conv2dtranspose)

ivy-seed:

PR #22830, opened 2023-08-31
refactor(All APIs): Added new pre-commit hook for ordering functions

NripeshN:


ivy-leaves:

Thank you :hugs:

ivy/functional/backends/jax/data_type.py
```diff
@@ -9,6 +9,26 @@
 from ivy.functional.backends.jax import JaxArray
 from ivy.functional.ivy.data_type import _handle_nestable_dtype_info
 
+char_rep_dtype_dict = {
```
vedpatwardhan, line 12: I think `char_rep_dtype_dict` is better placed after `ivy_dtype_dict` and `native_dtype_dict`

ivy/functional/backends/jax/data_type.py
```diff
@@ -245,6 +180,45 @@ def as_native_dtype(
         )
 
 
+# Array API Standard #
```
vedpatwardhan, line 183: The `Array API Standard` section should be present before the `Extra` section

ivy/functional/backends/jax/device.py
```diff
@@ -15,54 +15,44 @@
 )
 
 
-# Helpers #
```
vedpatwardhan, line 18: Private helpers should always be at the start of the file

ivy/functional/backends/jax/device.py
```diff
@@ -76,6 +66,10 @@ def _to_device(x, device=None):
     return x
 
 
+# --- Main --- #
```
vedpatwardhan, line 69: we should rename this section to `Extra` if it isn't a part of the Array API Standard

ivy/functional/backends/jax/device.py
```diff
@@ -99,55 +93,65 @@ def as_native_dev(device, /):
...
-    with jax.default_device(device_shifting_dev):
-        return fn(*args, **kwargs)
-
-
 def clear_cached_mem_on_dev(device: str, /):
     return None
 
 
-def _dev_is_available(base_dev):
+# API #
```
vedpatwardhan, line 100: API should instead be `Array API Standard`

ivy/functional/backends/jax/elementwise.py
```diff
@@ -16,6 +16,18 @@
...
 
+# --- Helpers --- #
+# --------------- #
+
+
+def _abs_variant_sign(x):
+    return jnp.where(x != 0, x / jnp.abs(x), 0)
+
+
+# --- Main --- #
```
vedpatwardhan, line 27: `Main` should be replaced by `Array API Standard`

vedpatwardhan (changes requested):
Just came across a few issues while going through the initial files, the same also apply to all files changed. Given that this modifies quite a lot of files, we should take our time and go through all changes made step by step before merging. For a start, could you please try and get the formatter updated according to the suggestions below? Thanks for diving into this @NripeshN, let's see what @KareemMAX thinks 😄

vedpatwardhan:
> @vedpatwardhan, The formator segregates the file into 2 halves. The first half is helpers and the other half would be Main functions. The headers are `helpers` and `main` respectively. We could remove headers like `# Array API Standard #` and `# Extras #` manually. Adding this functionality to the formatter would result in unnecessary complexity to the formatter and I think doing this task once manually would fix the issue as the formatter adds headers automatically(`helpers` and `main`)

I'm couldn't fully follow actually, but the "Array API Standard" header and the "Extras" header are important for us to segregate between functions that are and are not present in the standard as new functions not present in the standard get added on a daily basis. Ideally we shouldn't remove these headers, we should leave them unchanged. So in case we want to sort functions alphabetically, I think we should sort them separately for the functions in the Array API Standard and the Extras sections which would make the structure organized while also retaining the array api and extras distinction. And the Array API Standard header is just an example, we also have different section names like [`Autograd`](https://github.com/unifyai/ivy/blob/2c0bc53770b05c1f91c1ce299f5385be2dfa0543/ivy/functional/ivy/gradients.py#L399), [`Optimizer Steps`](https://github.com/unifyai/ivy/blob/2c0bc53770b05c1f91c1ce299f5385be2dfa0543/ivy/functional/ivy/gradients.py#L591) and [`Optimizer Updates`](https://github.com/unifyai/ivy/blob/2c0bc53770b05c1f91c1ce299f5385be2dfa0543/ivy/functional/ivy/gradients.py#L591) in `gradients.py`, [`Array Printing`](https://github.com/unifyai/ivy/blob/2c0bc53770b05c1f91c1ce299f5385be2dfa0543/ivy/functional/ivy/device.py#L174), [`Retrieval`](https://github.com/unifyai/ivy/blob/2c0bc53770b05c1f91c1ce299f5385be2dfa0543/ivy/functional/ivy/device.py#L344), [`Conversions`](https://github.com/unifyai/ivy/blob/2c0bc53770b05c1f91c1ce299f5385be2dfa0543/ivy/functional/ivy/device.py#L388), [`Memory`](https://github.com/unifyai/ivy/blob/2c0bc53770b05c1f91c1ce299f5385be2dfa0543/ivy/functional/ivy/device.py#L457) and others in `device.py` and similarly in other files. The section headers might not be formatted as correctly as they should be (i.e. with the `-` below the header and so on) but they do convey the information to anyone seeing the code regarding what the functions in that section do, which I think is more beneficial than clubbing all of them into `Main API`

The overall structure I think should be treated as the template while updating the formatter,
```
# global
import ...

# local
import ...


# Global declarations
<global variables like FN_CACHE and INF, global mode stacks and their initializers, postponed evaluation typehints, etc.>


# Helpers #
# -------- #

<all private helpers for the submodule>


# Classes

<any classes in the submodule>


# <function section header 1> --> any arbitrary name

<all functions in section 1 sorted alphabetically>
<could also have any assignment statements dependent on whether that function was defined, same for the following sections>


# <function section header 2> --> any arbitrary name

<all functions in section 2 sorted alphabetically>

.
.
.
.

# <function section header n> --> any arbitrary name

<all functions in section n sorted alphabetically>

```

Happy to know what you think @NripeshN @KareemMAX, thanks 😄

PR #22854, opened 2023-08-31
Update: compile updated to trace in ivy repo

HaiderSultanArc:
We are updating our `ivy.compile( )` function to `ivy.trace( )` function.
This requires updating the Docs, Ivy Repo, Graph Compiler Repo, Demos, and other places to refer to Tracer rather than Graph Compiler!

This PR updates Ivy Repo Code!


mattbarrett98 (commented):
I presume a lot of tests are failing due to what I pointed out, though I can't seem to confirm this with the CI (it just seems to show a long list of what's failing without the error messages).

Also I think you missed this https://github.com/unifyai/ivy/blob/main/ivy/__init__.py#L791 - presume this `try` will fail also when trying to import `compile` which is now `trace`

ivy/compiler/compiler.py
```diff
@@ -32,16 +32,16 @@ def compile(
     kwargs: Optional[dict] = None,
 ) -> Union[Graph, LazyGraph]:
     if python_version[1] == "8":
-        from ._compiler_38 import compile as _compile
+        from ._compiler_38 import trace as _trace
```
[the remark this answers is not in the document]
HaiderSultanArc: Yep. We'll merge all the PRs later. So this will be merged after updating GC and generating the new  `.so` file. For testing, I'll use _compile for now 😄

HaiderSultanArc:
> I presume a lot of tests are failing due to what I pointed out, though I can't seem to confirm this with the CI (it just seems to show a long list of what's failing without the error messages).
> 
> Also I think you missed this https://github.com/unifyai/ivy/blob/main/ivy/__init__.py#L791 - presume this `try` will fail also when trying to import `compile` which is now `trace`

Yeah I missed that 😅. Thanks for pointing out!

guillesanbri:
heads up before leaving a review, all the module stuff is being refactored at the moment afaik, so it will most likely become a nightmare to merge the changes :sweat_smile: I'd say better to wait to change this if the module refactor is going to be merged soon (which I assume it is)

HaiderSultanArc:
> heads up before leaving a review, all the module stuff is being refactored at the moment afaik, so it will most likely become a nightmare to merge the changes 😅 I'd say better to wait to change this if the module refactor is going to be merged soon (which I assume it is)

Yeah, I'll wait for it @guillesanbri 😄.

ivy/compiler/compiler.py
```diff
@@ -32,16 +32,16 @@ def compile(
...
     else:
-        from ._compiler import compile as _compile
+        from ._compiler import trace as _trace
```
guillesanbri, line 37: small heads up: this will need to be changed to `from ._compiler import compile as _trace` as the .so won't be updated yet

guillesanbri (changes requested):

ivy/compiler/compiler.py
HaiderSultanArc: yep!

ivy/utils/_importlib.py
```diff
@@ -120,7 +121,9 @@ def _import_module(name, package=None):
     module = module_from_spec(spec)
     import_cache[absolute_name] = module
     if ivy.is_local():
-        spec.loader.exec_module(module, ivy._compiled_id)
+        spec.loader.exec_module(
+            module, ivy._compiled_id
+        )  # ! Not sure about what this ID is
```
guillesanbri, line 126: I don't think this is related we can remove the comment

ivy/utils/exceptions.py
```diff
@@ -28,22 +28,22 @@ def _remove_so_log(trace):
         if ".pyx" in repr(st):
             continue
         if "<string>" in repr(st):
-            if "compiled_fn" in repr(st) and module_frame:
+            if "traced_fn" in repr(st) and module_frame:
```
guillesanbri, line 31: this probably needs to be `compiled_fn` until the changes are effective in the .so file

ivy/utils/exceptions.py
```diff
@@ -28,22 +28,22 @@ def _remove_so_log(trace):
...
+                traced_lineno = st.lineno
 
         if "<module>" in repr(st):
             module_frame = old_frames[idx]
             module_st = st
         elif (
             transpile_frame is None
             and os.path.join("ivy", "compiler") in st.filename
-            and st.name in ["compile", "transpile"]
+            and st.name in ["trace", "transpile"]
```
guillesanbri, line 41: not entirely sure about this one, so just highlighting in case it also needs to be updated only once the .so file is updated

guillesanbri (changes requested):

ivy/__init__.py
```diff
@@ -812,7 +812,7 @@ class Node(str):
 except:
     pass
 try:
-    from .compiler.compiler import transpile, compile, unify
+    from .compiler.compiler import transpile, trace, unify
```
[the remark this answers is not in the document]
HaiderSultanArc: Oh, I searched for `trace(` LOL. Obviously it would get missed. Sorry for that 😅

ivy/utils/_importlib.py
HaiderSultanArc: Got it!

ivy/utils/exceptions.py
HaiderSultanArc: yeah makes sense!

ivy/utils/exceptions.py
HaiderSultanArc: yeah, it's comming from `.so` file so it would be compile for now

PR #22905, opened 2023-09-01
empirical_covariance implementation (sklearn)

### LINKED ISSUE #22904: empirical_covariance

Anubhav-Goyal01:
Closes #22904

### DIFF
--- a/ivy/functional/frontends/sklearn/covariance/__init__.py
+++ b/ivy/functional/frontends/sklearn/covariance/__init__.py
@@ -1,0 +1,2 @@
+from . import _empirical_covariance
+from ._empirical_covariance import *
--- a/ivy/functional/frontends/sklearn/covariance/_empirical_covariance.py
+++ b/ivy/functional/frontends/sklearn/covariance/_empirical_covariance.py
@@ -11,8 +11,10 @@ def empirical_covariance(X, *, assume_centered = False):
     else:
         covariance = ivy.cov(X.T, bias = 1)
 
-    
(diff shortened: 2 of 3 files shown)


Ishticode (changes requested):
Almost there. I've fixed the "0 not equal 1" issue. which was due to too many or too less dimensions. The sklearn docs specify 2 dimensional input as you might have seen.

There are dtype mismatches still. This mean our test detect the output to have a different dtype when function is called from our sklearn frontend compard against that of the native sklearn front. Ive fixed somewhat. but best maybe you give it a go. Run the test and look at the input to out put dtypes.
An error message like
`AssertionError: returned dtype = float16, ground-truth returned dtype = float64`
means that our frontend sklearn is returning float16 while native sklearn is return float64 for the same input.
The input data can be seen under `Falsifying example ......`
you can also experiment by playing around with native sklearn (`from sklearn.convariance import ...`) or our frontend sklearn ("from ivy.functional.frontends.sklearn.covariance import ..."

Don't hesitate to reach out with questions ☺️

Anubhav-Goyal01:
Hi @Ishticode, thanks for the feedback, I have made a new commit trying to fix those issues

Ishticode (changes requested):
Hi @Anubhav-Goyal01 
It seems like the `test_sklearn_empirical_convariance` has failed due to int8 not being supported by dot function. check the logs for `test_sklearn_empirical_convariance here`

ivy-seed:

PR #22929, opened 2023-09-01
Add test for higher order gradients to ivy.jac

somaia02:
Hello @Ishticode. This PR adds a new test case to test higher-order gradients in ivy.jac as you requested before.

I disabled the new test in the paddle backend because it seems that there are many functions that don't support 2nd order gradient in paddle as in the example below:

```python
import paddle

x=paddle.to_tensor([1.])
def fn(x):
    return paddle.incubate.autograd.Jacobian(paddle.square, x)[:]
g = paddle.incubate.autograd.Jacobian(fn, x)
print(g[:])

# Works: sin, mean, sum
# errors: cos, square, exp
```


Sam-Armstrong:
Closing due to shift of focus to the transpiler, rather than ivy as a framework.

PR #22938, opened 2023-09-01
Completed Max_unpool1d implementation 

mohame54:
# PR Description 

**Completed max_unpool1d function implementation** I didn't add paddle backend implementation cuz paddle.max_unpool1d  had some errors related paddle itself

##

### DIFF
--- a/ivy/functional/frontends/torch/linalg.py
+++ b/ivy/functional/frontends/torch/linalg.py
@@ -76,6 +76,14 @@ def eigh(a, /, UPLO="L", out=None):
     return ivy.eigh(a, UPLO=UPLO, out=out)
 
 
+@with_supported_dtypes(
+    {"2.0.1 and below": ("float32", "float64", "complex32", "complex64", "complex128")},
+    "torch",
+)
+def eigh(A, UPLO="L", *, out=None):
+    return ivy.eigh(A, UPLO=UPLO, out=out)
+
+
 @to_ivy_arrays_and_back
 @with_supported_dtypes(
     {"2.0.1 and below": ("float32", "float64", "complex32", "complex64")}, "torch"
@@ -390,9 +398,3 @@ def vector_norm(input, ord=2, dim=None, keepdim=False, *, dtype=None, out=None):
     return ivy.vector_norm(
         input, axis=dim, keepdims=keepdim, ord=ord, out=out, dtype=dtype
     )
-
-@with_supported_dtypes(
-    {"2.0.1 and below": ("float32", "float64", "complex32", "complex64", "complex128")}, "torch", 
-)
-def eigh(A, UPLO="L", *, out=None):
-    return ivy.eigh(A, UPLO=UPLO, out=out)
--- a/ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
+++ b/ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py
@@ -516,6 +516,50 @@ def test_torch_eigh(
     )
 
 
+@handle_frontend_test(
+    fn_tree="torch.linalg.eigh",
+    dtype_and_x=_get_dtype_and_matrix(dtype="valid", square=True, invertible=True),
+    UPLO=st.sampled_from(("L", "U")),
+)
+def test_torch_eigh(
+    *,
+    dtype_and_x,
+    UPLO,
+    on_device,
+    fn_tree,
+    frontend,
+    test_flags,
+    backend_fw,
+):
+    dtype, x = dtype_and_x
+    x = np.array(x[0], dtype=dtype[0])
+    # make symmetric positive-definite beforehand
+    x = np.matmul(x.T, x) + np.identity(x.shape[0]) * 1e-3
+
+    ret, frontend_ret = helpers.test_frontend_function(
+        input_dtypes=dtype,
+        backend_to_test=backend_fw,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        test_values=False,
+        a=x,
+        UPLO=UPLO,
+    )
+    ret = [ivy.to_numpy(x) for x in ret]
+    frontend_ret = [np.asarray(x) for x in frontend_ret]
+
+    L, Q = ret
+    frontend_L, frontend_Q = frontend_ret
+
+    assert_all_close(
+        ret_np=Q @ np.diag(L) @ Q.T,
+        ret_from_gt_np=frontend_Q @ np.diag(frontend_L) @ frontend_Q.T,
+        atol=1e-02,
+    )
+
+
 # eigvals
 @handle_frontend_test(
     fn_tree="torch.linalg.eigvals",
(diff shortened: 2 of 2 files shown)


ivy-leaves:

Thank you :hugs:

mohame54:
my implementation is close to torch **2.0.1+cu118** version I don't know why gpu version of torch.max_unpool1d is different from the cpu version @AnnaTz

AnnaTz:
> my implementation is close to torch **2.0.1+cu118** version I don't know why gpu version of torch.max_unpool1d is different from the cpu version @AnnaTz

I get this error when I run the test locally:
```
E       AssertionError:  the results from backend torch and ground truth framework jax do not match
E        [[[-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]]
E       
E        [[-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]]
E       
E        [[-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1.  0.  1.  0.]]]!=[[[-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]]
E       
E        [[-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]]
E       
E        [[-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1. -1. -1.  0.]
E         [-1.  0.  2.  0.]]] 
E       
E       
E       Falsifying example: test_max_unpool1d(
E           backend_fw='torch',
E           on_device='cpu',
E           x_k_s_p=(['float32', 'int64'], array([[[-1., -1., -1.],
E                    [-1., -1., -1.],
E                    [-1., -1., -1.],
E                    [-1., -1., -1.]],
E            
E                   [[-1., -1., -1.],
E                    [-1., -1., -1.],
E                    [-1., -1., -1.],
E                    [-1., -1., -1.]],
E            
E                   [[-1., -1., -1.],
E                    [-1., -1., -1.],
E                    [-1., -1., -1.],
E                    [-1.,  1.,  2.]]], dtype=float32), array([[[0, 1, 2],
E                    [0, 1, 2],
E                    [0, 1, 2],
E                    [0, 1, 2]],
E            
E                   [[0, 1, 2],
E                    [0, 1, 2],
E                    [0, 1, 2],
E                    [0, 1, 2]],
E            
E                   [[0, 1, 2],
E                    [0, 1, 2],
E                    [0, 1, 2],
E                    [0, 2, 2]]]), (2,), (1,), 0),
```
It seems the failure happens when running on cpu, so I don't understand what you mean exactly.
Is there a hands-on example that shows the original torch function is inconsistent across versions?

mohame54:
Yeah, I will show you one example.

mohame54:
image

mohame54:
you can see here that's torch **cuda** max_unpool  output from the same input above which is different  from the cpu version

AnnaTz:
Ohh I get what's happening 😅
If you run the specific example many times it will sometimes return the correct value. The result is non-deterministic because the indices contain repeated values.
They actually mention it in their documentation 
image

mohame54:
Yeah, that's a catch I tried before to remove the repeated indices but it was not useful also the tests were giving me errors

AnnaTz:
> Yeah, that's a catch I tried before to remove the repeated indices but it was not useful also the tests were giving me errors

What kind of errors? I think you should be able to force the indices to not include identical values. We should at least know if that's the only problem (if the test passes for all other cases) before we decide what to do about it.

mohame54:
output mismatch errors if tried to exclude the repeated indices there are mismatch errors with torch backend as you can see so I didn't add  the excluding part to the implementation .

mohame54:
>What kind of errors? I think you should be able to force the indices to not include identical values. We should at least know if that's the only problem (if the test passes for all other cases) before we decide what to do about it.
I tried as many tests as I could but the only error that appeared to me were output mismatching errors.

mohame54:
Also I didn't complete the paddle backend implementation cuz the paddle function was giving me errors also  and when I tried it on colab with  `paddle 2.4.0` installed it didn't work for me. it seems that the  function has not been implemented yet.

AnnaTz:
I'm inclined to say we will need a compositional implementation for this 💀

mohame54:
->I'm inclined to say we will need a compositional implementation for this 💀
but Why?

mohame54:
Do you mean compositional implementation using ivy API instead of backend specific API wouldn't this approach affect the performance 😕

AnnaTz:
> ->I'm inclined to say we will need a compositional implementation for this 💀 but Why?

We need an implementation that works consistently for all backends in all cases. The native torch function does not work as expected in the repeated indices case, and the paddle function is not implemented in some versions and giving errors in others. So We need an ivy function implementation to cover these.

AnnaTz:
> Do you mean compositional implementation using ivy API instead of backend specific API, wouldn't that affect the performance 😕

Also, about that, according to what @vedpatwardhan has told me, I think the implementation we have for the numpy, jax and tensorflow backends could be replaced by an ivy composition. It seems that we are doing the exact same operations in all these backends, so it would be better to avoid code repetition.

mohame54:
But I agree with you in order to reduce the code repetition  we need compositional implementation

AnnaTz:
> But I agree with you in order to reduce the code repetition we need compositional implementation

I think it will be quite simple since you have already implemented the backend functions compositionally. I think you will mainly copy that and replace native calls with ivy calls.

mohame54:
sorry I didn't mean to edit your comment so sorry @AnnaTz 😄

ivy/data_classes/array/experimental/layers.py
```diff
@@ -1117,3 +1117,45 @@ def rfftn(
             The result of the RFFT operation.
         """
         return ivy.rfftn(self._data, s=s, axes=axes, norm=norm, out=out)
+
+    def max_unpool1d(
+        self: ivy.Array,
+        indices: ivy.Array,
+        kernel_size: Union[Tuple[int], int],
+        strides: Union[int, Tuple[int]] = None,
```
AnnaTz, line 1125: Just noticed you need to add `/, *,` before the start of the keyword default arguments.

ivy/data_classes/array/experimental/layers.py
```diff
@@ -1117,3 +1117,45 @@ def rfftn(
...
+        padding: Union[int, Tuple[int]] = 0,
```
AnnaTz, line 1126: Also, please remove the `docs/demos` changes.

ivy_tests/test_ivy/test_functional/test_experimental/test_nn/test_layers.py
```diff
@@ -1190,23 +1187,54 @@ def test_max_pool3d(
...
+            data_format="channel_first",
+            **data_gen_kwargs)
+    )
+    dts.extend(["int64"])
+    values = values[0]
+    padding = draw(helpers.ints(min_value=0, max_value=2))
+    if padding > (kernel_size[0] // 2):
+        padding = 0
+
+    values, indices = torch.nn.functional.max_pool1d(
```
AnnaTz, line 1208: @sherry30 are we allowed to use native functions in testing?

ivy/functional/ivy/experimental/layers.py
```diff
@@ -180,9 +205,45 @@ def max_unpool1d(
...
+    f = one_like_mask * feature_range
+    updates_size = input.size
+    indices = ivy.stack([b, f, indices]).reshape(3, updates_size)
+    output = ivy.zeros(output_shape, dtype=input.dtype)
+    indices = tuple(indices)
+    values = input.reshape((-1,))
+    output[indices] = values
+    if revert:
+        output = output.permute_dims([0, 2, 1])
+    return output
```
AnnaTz, line 247: The code looks good overall. The only thing remaining is to add the partial torch backend.

AnnaTz (changes requested):

ivy/functional/ivy/experimental/layers.py
mohame54: What do you mean by partial torch backend? I did not get what you meant here

ivy/data_classes/array/experimental/layers.py
mohame54: If I remove them there will be conflicts with the main branch

ivy/functional/ivy/experimental/layers.py
AnnaTz: https://unify.ai/docs/ivy/overview/deep_dive/function_types.html#partial-mixed-functions

ivy/data_classes/container/experimental/layers.py
```diff
@@ -2280,3 +2280,112 @@ def rfftn(
...
+    @staticmethod
+    def static_max_unpool1d(
+        input: ivy.Container,
+        indices: ivy.Container,
+        kernel_size: Union[Tuple[int], int],
+        strides: Union[int, Tuple[int]] = None,
+        padding: Union[int, Tuple[int]] = 0,
+        data_format: Union[str, ivy.Container] = "NWC",
+        /,
+        *,
```
AnnaTz, line 2293: Misplaced `/, *` here (sorry to nitpick 😅)

AnnaTz (changes requested):

ivy/data_classes/container/experimental/layers.py
mohame54: okay no problem.

PR #22977, opened 2023-09-03
Add broadcast_shape to paddle frontend

### LINKED ISSUE #22919: broadcast_shape

charlie572:
## Related Issue 

Close #22919

## Checklist 

- [x] Did you add a function?
- [x] Did you add the tests?

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
+++ b/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
@@ -1,5 +1,6 @@
 # global
 from hypothesis import strategies as st
+import hypothesis.extra.numpy as nph
 
 # local
 import ivy_tests.test_ivy.helpers as helpers
@@ -471,6 +472,37 @@ def test_paddle_atanh(
         on_device=on_device,
         x=x[0],
     )
+
+
+# broadcast_shape
+@handle_frontend_test(
+    fn_tree="paddle.broadcast_shape",
+    shapes=nph.mutually_broadcastable_shapes(
+        num_shapes=2, min_dims=1, max_dims=5, min_side=1, max_side=5
+    ),
+    test_with_out=st.just(False),
+)
+def test_paddle_broadcast_shape(
+    *,
+    shapes,
+    on_device,
+    fn_tree,
+    frontend,
+    test_flags,
+    backend_fw,
+):
+    ret, frontend_ret = helpers.test_frontend_function(
+        input_dtypes=["int64"],
+        backend_to_test=backend_fw,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        test_values=False,
+        x_shape=shapes.input_shapes[0],
+        y_shape=shapes.input_shapes[1],
+    )
+    assert ret == frontend_ret
 
 
 # ceil
--- a/ivy/functional/frontends/paddle/tensor/math.py
+++ b/ivy/functional/frontends/paddle/tensor/math.py
@@ -110,6 +110,10 @@ def atanh(x, name=None):
     return ivy.atanh(x)
 
 
+def broadcast_shape(x_shape, y_shape):
+    return ivy.broadcast_shapes(x_shape, y_shape)
+
+
 @with_unsupported_dtypes({"2.5.1 and below": ("float16", "bfloat16")}, "paddle")
 @to_ivy_arrays_and_back
 def ceil(x, name=None):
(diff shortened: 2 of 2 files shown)


charlie572:
The test doesn't pass, and I can't figure out why. It says this:

```
Expected :([1],)
Actual   :[1]
```

But when I run the function in the shell, I get the same result from both functions.

```
In [2]: from ivy.functional.frontends import paddle
In [3]: paddle.broadcast_shape((2, 1, 1), (1, 1, 2))
Out[3]: [2, 1, 2]
In [4]: paddle.broadcast_shape([2, 1, 1], [1, 1, 2])
Out[4]: [2, 1, 2]
In [5]: import paddle
In [6]: paddle.broadcast_shape((2, 1, 1), (1, 1, 2))
Out[6]: [2, 1, 2]
In [7]: paddle.broadcast_shape([2, 1, 1], [1, 1, 2])
Out[7]: [2, 1, 2]
```

Also, the function is supposed to allow you to input tuples or lists. I'm not sure how to test both of them in hypothesis. Do I have to sample from (list, tuple) with a strategy, and convert the shapes to that type?

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
```diff
@@ -473,6 +474,37 @@ def test_paddle_atanh(
     )
 
 
+# broadcast_shape
+@handle_frontend_test(
+    fn_tree="paddle.broadcast_shape",
+    shapes=nph.mutually_broadcastable_shapes(
```
MahmoudAshraf97, line 480: use `helpers.mutually_broadcastable_shapes` strategy instead as it contains additional constraints over the strategy you used

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
```diff
@@ -473,6 +474,37 @@ def test_paddle_atanh(
...
+        backend_to_test=backend_fw,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        test_values=False,
+        x_shape=shapes.input_shapes[0],
+        y_shape=shapes.input_shapes[1],
+    )
+    assert ret == frontend_ret
```
MahmoudAshraf97, line 505: I see that you wrote the test differently than other tests, is there a reason for that? because you don't need the separate assertion at the end

MahmoudAshraf97 (changes requested):

charlie572:
test_frontend_function wasn't checking list return values properly. I've tried to fix it.

charlie572:
There are some tests failing because of types e.g. test_paddle_is_complex says. "Cannot convert to Paddle dtype. uint16 is not supported by Paddle." But I checked out previous commits, including the commit from just before I started this branch, and they fail for the same reason. Were these tests failing before, or have I set up my environment incorrectly? Also, I looked at the history of the main branch, and none of the commits in the last few weeks pass CI. So I'm confused about which tests I have to fix.

sherry30 (changes requested):
lets not move the `test_values` if statement, what we should do instead is tuplify `ret` as well so both `ret` and `frontend_ret` are both in tuples. This will make it easy to loop over the returns in the test function (Since some functions can also return tuples, so its easier to always return a tuple).
The other change looks good

Thanks

charlie572:
Ignore those last two commits. I was looking at the wrong function.

charlie572:
test_paddle_tensor_unsqueeze gives the error
`AssertionError: returned dtype = bfloat16, ground-truth returned dtype = uint16`

I noticed that paddle has a problem with setting the dtype as bfloat16. It doesn't work if you use the dtype argument to the Tensor constructor (it sets the dtype to int64 instead). But it does work if you use astype.

```
paddle.Tensor(np.arange(10), dtype="bfloat16")
Out[19]: 
Tensor(shape=[10], dtype=int64, place=Place(cpu), stop_gradient=True,
       [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
paddle.Tensor(np.arange(10), dtype="bfloat16").astype("bfloat16")
Out[20]: 
Tensor(shape=[10], dtype=bfloat16, place=Place(cpu), stop_gradient=True,
       [0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
```

Maybe this has something to do with the error. Should I try to fix it, or should I open an issue in the original paddle repository? I don't know how ivy converts to backend tensors, so I don't know how to fix it in ivy.

MahmoudAshraf97:
Hello, are you still working on this?

charlie572:
Yes. Is it ready to merge?

charlie572:
I opened my pull request first. I even commented on that pull request and pointed out that it was a duplicate. Why didn't you close the other one and focus on mine?

Issue #22978, opened 2023-09-03
complex


dettmerramon:
Implement complex for the torch frontend.
See https://github.com/unifyai/ivy/issues/8 for more information.

PR #22979, opened 2023-09-03
feat(frontend): Add torch.complex, torch.polar and its tests

### LINKED ISSUE #22752: polar
Implement `polar` for the `torch` frontend.
See #8 for more information.

dettmerramon:
# PR Description 

Adds `torch.polar` and `torch.complex` to the torch frontend (see `creation_ops` issue #8), alongside its tests.
Ended up implementing `torch.complex` since it made more sense to be able to use it for `torch.polar`.

I've made some comments on the PR to make some of my decisions more clear, you can probably see them while going through the code or in the PR feed below.

## Related Issue 

Closes #22752 and #22978

## Checklist 

- [X] Did you add a function?
- [X] Did you add the tests?
- [X] Did you follow the steps we provided?

### DIFF
--- a/ivy/functional/frontends/torch/creation_ops.py
+++ b/ivy/functional/frontends/torch/creation_ops.py
@@ -256,6 +256,7 @@ def ones_like_v_0p4p0_and_above(
 
 
 @with_supported_dtypes({"2.0.1 and below": ("float32", "float64")}, "torch")
+@to_ivy_arrays_and_back
 def polar(
     abs,
     angle,
--- a/ivy/functional/frontends/torch/creation_ops.py
+++ b/ivy/functional/frontends/torch/creation_ops.py
@@ -4,7 +4,11 @@ from ivy.functional.frontends.torch.func_wrapper import (
     to_ivy_arrays_and_back,
     to_ivy_shape,
 )
-from ivy.func_wrapper import with_unsupported_dtypes
+from ivy.func_wrapper import (
+    with_unsupported_dtypes,
+    with_supported_dtypes,
+    handle_out_argument,
+)
 import ivy.functional.frontends.torch as torch_frontend
 
 
@@ -308,3 +312,32 @@ def zeros_like(
 ):
     ret = ivy.zeros_like(input, dtype=dtype, device=device)
     return ret
+
+
+@with_supported_dtypes({"2.0.1 and below": ("float32", "float64")}, "torch")
+@to_ivy_arrays_and_back
+@handle_out_argument
+def complex(
+    real,
+    imag,
+    *,
+    out=None,
+):
+    assert real.dtype == imag.dtype, ValueError(
+        f"Expected real and imag to have the same dtype, "
+        " but got real.dtype = {real.dtype} and imag.dtype = {imag.dtype}."
+    )
+
+    complex_dtype = ivy.complex64 if real.dtype != ivy.float64 else ivy.complex128
+    complex_array = real + imag * 1j
+    return complex_array.astype(complex_dtype)
+
+
+@with_supported_dtypes({"2.0.1 and below": ("float32", "float64")}, "torch")
+def polar(
+    abs,
+    angle,
+    *,
+    out=None,
+):
+    return complex(abs * angle.cos(), abs * angle.sin(), out=out)
--- a/ivy/functional/frontends/torch/creation_ops.py
+++ b/ivy/functional/frontends/torch/creation_ops.py
@@ -7,7 +7,6 @@ from ivy.functional.frontends.torch.func_wrapper import (
 from ivy.func_wrapper import (
     with_unsupported_dtypes,
     with_supported_dtypes,
-    handle_out_argument,
 )
 import ivy.functional.frontends.torch as torch_frontend
 
@@ -73,6 +72,24 @@ def asarray(
     copy=None,
 ):
     return ivy.asarray(obj, copy=copy, dtype=dtype, device=device)
+
+
+@with_supported_dtypes({"2.0.1 and below": ("float32", "float64")}, "torch")
+@to_ivy_arrays_and_back
+def complex(
+    real,
+    imag,
+    *,
+    out=None,
+):
+    assert real.dtype == imag.dtype, ValueError(
+        "Expected real and imag to have the same dtype, "
+        f" but got real.dtype = {real.dtype} and imag.dtype = {imag.dtype}."
+    )
+
+    complex_dtype = ivy.complex64 if real.dtype != ivy.float64 else ivy.complex128
+    complex_array = real + imag * 1j
+    return complex_array.astype(complex_dtype, out=out)
 
 
 @to_ivy_arrays_and_back
@@ -238,6 +255,16 @@ def ones_like_v_0p4p0_and_above(
     return ret
 
 
+@with_supported_dtypes({"2.0.1 and below": ("float32", "float64")}, "torch")
+def polar(
+    abs,
+    angle,
+    *,
+    out=None,
+):
+    return complex(abs * angle.cos(), abs * angle.sin(), out=out)
+
+
 @to_ivy_arrays_and_back
 @with_unsupported_dtypes({"2.0.1 and below": ("float16",)}, "torch")
 def range(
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/torch/creation_ops.py
```diff
@@ -297,3 +301,32 @@ def zeros_like(
...
+        " but got real.dtype = {real.dtype} and imag.dtype = {imag.dtype}."
+    )
+
+    complex_dtype = ivy.complex64 if real.dtype != ivy.float64 else ivy.complex128
+    complex_array = real + imag * 1j
+    return complex_array.astype(complex_dtype)
+
+
+@with_supported_dtypes({"2.0.1 and below": ("float32", "float64")}, "torch")
+def polar(
```
dettmerramon, line 326: Not necessary to use `@handle_out_argument` and `@to_ivy_arrays_ and_back` in here since `complex` already outputs in `ivy.frontends.torch.Tensor` and handles `out`

ivy_tests/test_ivy/test_frontends/test_torch/test_creation_ops.py
```diff
@@ -789,3 +789,57 @@ def test_torch_zeros_like(
...
+    input_dtype, input = dtype_and_x
+    helpers.test_frontend_function(
+        input_dtypes=input_dtype,
+        backend_to_test=backend_fw,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        abs=input[0],
+        angle=input[0],
```
dettmerramon, line 844: I'm aware that I'm not testing out in here. I've looked through some of the other tests and didn't seem like it was usually tested. If you guys want me to implement it somehow, just let me know

ivy/functional/frontends/torch/creation_ops.py
```diff
@@ -297,3 +301,32 @@ def zeros_like(
...
+    imag,
+    *,
+    out=None,
+):
+    assert real.dtype == imag.dtype, ValueError(
+        f"Expected real and imag to have the same dtype, "
+        " but got real.dtype = {real.dtype} and imag.dtype = {imag.dtype}."
+    )
+
+    complex_dtype = ivy.complex64 if real.dtype != ivy.float64 else ivy.complex128
```
dettmerramon, line 320: the original torch method returns `complex64` for `float32` and `complex128` for `float64`, so this is just done to align the frontend output to the expected torch output.

ivy/functional/frontends/torch/creation_ops.py
```diff
@@ -297,3 +301,32 @@ def zeros_like(
 ):
     ret = ivy.zeros_like(input, dtype=dtype, device=device)
     return ret
+
+
+@with_supported_dtypes({"2.0.1 and below": ("float32", "float64")}, "torch")
+@to_ivy_arrays_and_back
+@handle_out_argument
```
sherry30, line 308: no need to use handle_out_argument here, this wrapper is for the backends only.
You should use the `out` argument in the ivy functions to inplace update the `out` array.

ivy/functional/frontends/torch/creation_ops.py
```diff
@@ -297,3 +301,32 @@ def zeros_like(
...
+    out=None,
+):
+    assert real.dtype == imag.dtype, ValueError(
+        f"Expected real and imag to have the same dtype, "
+        " but got real.dtype = {real.dtype} and imag.dtype = {imag.dtype}."
+    )
+
+    complex_dtype = ivy.complex64 if real.dtype != ivy.float64 else ivy.complex128
+    complex_array = real + imag * 1j
+    return complex_array.astype(complex_dtype)
```
sherry30, line 322: whichever is the last operation you perform, you should use an ivy function for it and then pass `out` argument in it.
So you can do this as following:
`ivy.astype(complex_array, complex_dtype, out=out)`

ivy/functional/frontends/torch/creation_ops.py
sherry30: You should still use `@to_ivy_arrays_ and_back` since you're doing `angle.sin()` in it.
making sure that everything in this function is is an ivy array is better since native arrays are also expected as an input to this function and doing `angle.sin()` in that case will produce inconsistent results

ivy_tests/test_ivy/test_frontends/test_torch/test_creation_ops.py
sherry30: Its tested automatically internally, so no need to do anything extra

sherry30 (changes requested):
requested a few changes

ivy/functional/frontends/torch/creation_ops.py
dettmerramon: It also seems ok to me to not use `@to_ivy_arrays_ and_back` in here since there's already an implementation for both `sin` and `cos` in the torch tensor frontend (but maybe it's not the best practice (?)).

dettmerramon: Hi @sherry30 , could you give me a resolution on what you think would be the best course of action? given the insights I've described above

sherry30: Right, but native arrays are also expected as inputs.
So if the backend is tensorflow, tf tensors are also expected. But if you're using instance methods, it will give inconsistent results because it will call instance methods of tf tensors.
Maybe @AnnaTz can elaborate here how to proceed, I wasn't aware that `out` argument cannot be an ivy array

AnnaTz: Hi @dettmerramon @sherry30 
The functions of a specific frontend can receive either ivy arrays or its own frontend arrays. I think `polar` would work correctly in both cases without needing `@to_ivy_arrays_and_back`.

dettmerramon: tks for clarifying @AnnaTz .
@sherry30 do you have any requests left? Everything else was already implemented.

AnnaTz: Hi @dettmerramon, I had a talk with @sherry30 and it turns out he was right about frontend functions always needing to carry `@to_ivy_arrays_and_back`. We do need to make sure they can work for native arrays as well, so please ignore my last comment 😅

ivy_tests/test_ivy/test_frontends/test_torch/test_creation_ops.py
```diff
@@ -150,6 +150,60 @@ def _start_stop_step(draw):
 # ------------ #
 
 
+# complex
+@handle_frontend_test(
+    fn_tree="torch.complex",
+    dtype_and_x=helpers.dtype_and_values(available_dtypes=helpers.get_dtypes("float")),
+)
```
AnnaTz, line 157: Now, about the issue with the `out` argument when adding `@to_ivy_arrays_and_back`. I believe there is some extra logic that needs to be added to our frontend design to handle the `out` argument properly, so there is no expectation to have that working in your PR. You can go ahead and add `test_with_out=st.just(False)` here to exclude `out` from the test.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_torch/test_creation_ops.py
dettmerramon: ok, I might go with that. 
Would it be a problem if I implemented `ivy.complex` on this PR? the logic itself is pretty simple, and I personally prefer doing that then "bandaging" the test itself.

dettmerramon: ok, I might go with that. 
Would it be a problem if I implemented `ivy.complex` on this PR? The logic itself is pretty simple, and I personally prefer doing that than "bandaging" the test itself.

dettmerramon: Looking into it a bit more, it seems that the reason is that using `@to_ivy_arrays_and_back` makes it only work with frontend arrays, not working with `ivy.array` either (when using `out`). It seems to be more of a structural problem, since using an already implemented method like `torch.as_array` also yields an error (example below):

```python
import ivy

a = ivy.array([1,2,3], dtype=torch.float64)
b = ivy.array([2,3,4], dtype=torch.float64)
c = torch.as_tensor(a, out=b)
```

The error doesn't happen when using `torch.tensor`. I guess it doesn't make a difference whether `ivy.complex` exists, so I'll go with the more straightforward approach for now. Would it be okay to open an issue for the weird `out` behavior tho?

AnnaTz: `torch.as_tensor` and `torch.tensor` don't have an `out` argument at all actually. I will make a note to move the `out` assertion so that it doesn't come up in such cases before the "argument not supported" error. It is misleading.

AnnaTz: The `out` argument of frontend functions is handled by the ivy function called inside them. 
Essentially `@to_ivy_arrays_and_back` on `torch.complex` will convert the `out` frontend array to an ivy array, that ivy array will get passed to `ivy.complex`, and finally the `@handle_out_argument` on `ivy.complex` will do the inplace update on the `out` array.

dettmerramon: Yeah, I've messed up with that example, my bad.

Dragging this a bit longer tho, the error still happens when using `ivy.array` for `out` in functions that do support `out` (tested it with `torch.full` and `torch.eye`, they both use `ivy` impls, support `out` and raise the same error), just felt it could be good to notify you guys.

Regardless, this seems to be out of the scope of this PR at this point. I've recently updated the PR adding the `@to_ivy_arrays_and_back` to `torch.polar`, so it seems to be good to go.

Issue #22937, opened 2023-09-01
exists


SpyingEnvelope:

SpyingEnvelope:
Reformatted exists and PR was accepted.

PR #22993, opened 2023-09-04
Refactor exists

SpyingEnvelope:
# PR Description 

The docstrings, examples, hint types, and arguments for exists were mostly correct. I turned the passed in parameter to be positional only, as per the recommendations from the Ivy documentation.

I also implemented an ivy.Container instance method for exists that returns an ivy.Container detailing if any of the leaves are None.

## Related Issue 
This should resolve issue #22937 for reformatting exists.

## Checklist 

- [x] Did you add a function?
- [x] Did you add the tests? - I added docstrings tests

## Checklist provided by docs

- [x] Remove all lambda and direct bindings for the backend functions (in ivy.functional.backends), with each function instead defined using def.
- [x] Implement the following if they don’t exist but should do: ivy.Array instance method, ivy.Container instance method, ivy.Array special method, ivy.Array reverse special method, ivy.Container special method, ivy.Container reverse special method.
- [x] Make sure that the aforementioned methods are added into the correct category-specific parent class, such as ivy.ArrayWithElementwise, ivy.ContainerWithManipulation etc.
- [x] Correct all of the Function Arguments and the type hints for every function and its relevant methods, including those you did not implement yourself.
- [x] Add the correct Docstrings to every function and its relevant methods, including those you did not implement yourself.
- [x] Add thorough Docstring Examples for every function and its relevant methods and ensure they pass the docstring tests.

### DIFF
--- a/ivy/data_classes/container/general.py
+++ b/ivy/data_classes/container/general.py
@@ -4242,3 +4242,57 @@ class _ContainerWithGeneral(ContainerBase):
             A tuple containing the strides.
         """
         return self.static_strides(self)
+
+    @staticmethod
+    def static_exists(x: ivy.Container, /) -> ivy.Container:
+        """
+        ivy.Container instance method variant of ivy.exists. This method simply wraps
+        the function, and so the docstring for ivy.exists also applies to this method
+        with minimal changes.
+
+        Parameters
+        ----------
+        x
+            input container.
+
+        Returns
+        -------
+        ret
+            True if x is not None, else False.
+
+        Examples
+        --------
+        >>> x = ivy.Container(a=[1,2,3,4], b=[])
+        >>> y = x.exists()
+        >>> print(y)
+        { a: True, b: True }
+
+        >>> x = ivy.Container(a=None, b=[1,2])
+        >>> y = x.exists()
+        >>> print(y)
+        { a: False, b: True }
+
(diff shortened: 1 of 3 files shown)


ivy/data_classes/container/general.py
```diff
@@ -4242,3 +4242,57 @@ def strides(
...
+        ----------
+        self
+            input container.
+
+        Returns
+        -------
+        ret
+            True if x is not None, else False.
+        """
+        return self.static_exists(self)
```
AnnaTz, line 4298: Hi @SpyingEnvelope, there are some extra arguments we always add to container methods; `key_chains`, `to_apply`, `prune_unapplied` and `map_sequences`. Please have a look at the rest of the methods and define those in the same way. Thanks!

AnnaTz (changes requested):

ivy/data_classes/container/general.py
SpyingEnvelope: Hi @AnnaTz , I pushed the updates you requested. I also renamed static_exists to _static_exists, as that appears to be the more common naming convention. On top of adding the arguments you requested, I also added as well as modified some docstrings to better clarify information.

AnnaTz (approved):
Thanks, looks good!

PR #23003, opened 2023-09-04
feat: add script to automatically infer (un)supported dtypes

jshepherd01:
# PR Description

Adds a (work in progress) script that can be pointed at any `frontend/` or `backend/` file and will automatically add the `with_(un)supported_(device_and_)dtypes` decorators to every function it finds there. It does this by mocking `helpers.get_dtypes` and `helpers.test_function`, then running the test for the relevant function once with a minimal example of each dtype and inspecting any generated errors.

Currently, the new script file contains several `TODO` comments, which I'm currently working through. I'm also planning to prune down the mock implementation of `test_function`, since I'm fairly sure it doesn't really need to e.g. handle the `out` argument for the sake of this script

## Related Issue

The supported dtypes decorators were quite messy and full of shortcuts (like the test uses `get_dtypes('float')` so the `with_unsupported_dtypes` decorator only considers floats) or incorrectness (like `bfloat16` being marked as unsupported because the numpy-based testing pipeline threw an error), and also they had to be updated manually when a new framework version came out which would lead to newly-supported dtypes being missed.

## README

To call the script on e.g. the numpy backend's activation functions you'd do
```
python3 determine_dtypes.py ivy/functional/backends/numpy/activations.py -v
```
you can also pass in multiple files (just give it multiple file paths) or pass it directories and it will recursively search for files. It then gets the frontend or backend name from the file path, searches the file for function definitions (of the form `def func(`, and it ignores private functions), and also finds the relevant test file and test function for each one according to our naming conventions.

For each file, it mocks the `helpers.get_dtypes` function to return a single dtype (iterating over `ivy.valid_dtypes`) and then runs the test function, which will call a mock of either `test_function` or `test_frontend_function`. That mock runs the minimal amount of code needed to produce arguments for either the relevant ivy function (for backend functions) or the relevant function in the gt framework (for frontend functions). It calls it wrapped in a try block, and then there's some error handling with string matching on the error message to identify whether errors were dtype-related or not.

Once it's got all the dtype support information it needs, it then reads the existing dtype decorators from the file, combines what they say with the information from the tests it's just run, and writes the new decorators to the file. It also changes the imports if need be.

The general flow of what it does is in the `run_dtype_setter` function (line 1563), and a lot of the complicated stuff is handled by the `FileTester` classes (lines 700 for base, 1164 for backend, and 1223 for frontend).

The biggest limitations currently are that it doesn't work for methods and that there seems to be a bug with the handling of non-cpu devices in the frontends. Fixing the former probably means classifying the functions as either regular or method functions when they're read from the file, then modifying some of the other logic to use that information; fixing the latter would probably involve changing `FrontendFileTester.setup_test` and `mock_test_frontend_function`.

Since I've now finished my internship I probably won't be able to do major work on this (i.e. new features or very involved fixes) but I'm happy to do minor bug fixes or to explain parts of the code if needed.


vedpatwardhan (commented):
Looks good to me on a high-level, @RickSanchezStoic could you please test this out to see if any changes are required? Ideally we'd want to be able to directly plug it into one of our workflows which would get triggered on every release to update the decorators. Thanks 😄

PR #23012, opened 2023-09-04
add ivy.Array support to jax backend

Killua7362:
Used register_pytree_node in jax backend to add support for ivy.Array class


vedpatwardhan (approved):
lgtm! I haven't tested your changes with the example that uses jax transformations with ivy.Array, but feel free to merge if you've tested those already. Thanks @Killua7362 😄

Killua7362:
I'll test for few more examples and then I will merge
Thank you Ved

PR #23075, opened 2023-09-05
feat: add invert_permutation function to tensorflow frontend

### LINKED ISSUE #22939: invert_permutation

losmi247:
# PR Description 

- Added invert_permutation function at ivy/functional/frontends/tensorflow/math.py
- Added invert_permutation test at ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
- Added a 'for_frontend_test' boolean argument to _invert_permutation_helper experimental Hypothesis helper function at ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_sorting.py so that it can be reused in the test for the new invert_permutation tensorflow frontend function

Regarding the last point - if the 'for_frontend_test' argument is true, _invert_permutation_helper Hypothesis helper only generates permutations as NumPy arrays of type int64. This is required for the new invert_permutation TensorFlow frontend function because all backend implementations of invert_permutation always return results of type int64 (e.g. see ivy/functional/backends/tensorflow/experimental/sorting.py, it casts the resulting tensor to int64), whereas the tf.math.invert_permutation function returns a tensor of same type as the input tensor. Therefore, if we provided a int32 input to the frontend test for this function, our frontend function would give a int64 result, and the TensorFlow ground truth tf.math.invert_permutation would give a int32 result, which means the test always fails.

Also, the line that imports the _invert_permutation_helper in ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py is 89 characters long (1 character longer than black allows), and black can't reformat it in any way, so I turned off formatting for that line as a temporary solution. Moving the _invert_permutation_helper function out of the 'experimental' folder to where regular Hypothesis helpers are could solve this.

## Related Issue 

Close #22939

## Checklist 

- [ ] Did you add a function?
- [ ] Did you add the tests?


losmi247:
@aparajith21

losmi247:
Thank you for taking a look! 

The PR checks say that the new tests test_tensorflow_invert_permutation fail, but they all pass locally. Also, many other tests (test_tensorflow_log_sigmoid, test_tensorflow_logical_not, test_tensorflow_is_nan, etc.) pass locally but are shown as failures in PR checks. I'm not sure why. I ran all tests from test_math.py locally and about 150 fail, but I assume those are old failures, as most of them are not shown in 'new failures' in PR checks.

Anyway, I can not really see any connection between those failures and the changes in this PR (except for the PR checks saying that test_tensorflow_invert_permutation fails, when it actually passes locally), so presumably all is fine locally.

aparajith21:
Thanks for the contribution, LGTM. Failures seem to be due to a dill import that is not related to your PR, so merging

PR #23087, opened 2023-09-05
Migrating all binaries to the binaries repo

vedpatwardhan:
This PR removes the binaries from the `ivy` repo which would instead be stored on a dedicated repo named `binaries` to enable pulling only relevant binaries to a particular user's interpreter, OS and architecture.


vedpatwardhan:
I've done the following,

1. Removed all binaries from the `ivy` repo and added them instead to the [binaries](https://github.com/unifyai/binaries) repo with the naming convention same as for pypi wheels for consistency of naming. Also added a `binaries.json` to `ivy` for the skeleton structure. We'd need to add more binaries for different architectures and also different python versions as the current ones are only for python 3.9 and 3.10, but @vaatsalya123 could you please take a look at this part of the changes?
2. Updated the `setup.py` to download the binaries corresponding to a particular architecture with the same preference list of names of binaries as the names of wheels used by pypi (with `sys.tags()` being that preference list) with a spinner, a progress bar and so on. I'm not sure if it's as clean as it can be, @guillesanbri could you please take a look at the changes to `setup.py`?
3. In the rare case of someone wanting to clone repo, checkout to an older release (through the tag) and pulling the binaries corresponding to that release, I've added a "VERSION" environment variable they can set. But for this to be useful, we'd first need to ensure that we also add tags to the `binaries` repo when we make our releases. @RashulChutani could you please take a look at the changes made to `pypi.yml`?

Happy to know about improvements or clarify about any of these changes, thanks!

setup.py
```diff
@@ -18,13 +18,71 @@
 import setuptools
 from setuptools import setup
 from pathlib import Path
+from packaging import tags
```
guillesanbri, line 21: had to install this manually

setup.py
```diff
@@ -18,13 +18,71 @@
...
+import os
+import json
+import requests
```
guillesanbri, line 24: same with requests

setup.py
```diff
@@ -18,13 +18,71 @@
...
+import itertools
 import re
+from tqdm import tqdm
```
guillesanbri, line 27: and same with tqdm

setup.py
```diff
@@ -18,13 +18,71 @@
...
+        r = requests.get(
+            f"https://github.com/unifyai/binaries/raw/{version}/{search_path}",
+            timeout=40,
+        )
+        if r.status_code == 200:
+            if pbar is None:
+                print()
+                print("Downloading binaries ...")
+                pbar = tqdm(total=len(paths))
+            with open(path, "wb") as f:
```
guillesanbri, line 78: can't pass over this point, I get:
```
          with open(path, "wb") as f:
      FileNotFoundError: [Errno 2] No such file or directory: 'ivy/compiler/utils/III.so'
```

Seems to work if I create the utils dir manually, so I'm assuming we are missing `os.makedirs(os.path.dirname(path), exist_ok=True)` before `open`

setup.py
```diff
@@ -18,13 +18,71 @@
...
+                f.write(r.content)
+            terminate = path == paths[-1]
+            pbar.update(1)
+        else:
```
guillesanbri, line 82: maybe worth adding an error message if the connection hasn't been successful :)

guillesanbri (commented):
Left a few comments, I tried installing in a fresh environment, which I assume is the cause of the missing packages, will need to sort this out nonetheless :sweat_smile: 

I think the `makedirs` call should fix the other problem I was facing. I tried on a fresh conda environment, let me know if something else should be specially tested :+1:

setup.py
vedpatwardhan: done, thanks for pointing that out!

setup.py
vedpatwardhan: replaced it with a built-in alternative, thanks for pointing that out!

vedpatwardhan:
done with the changes, please take a look whenever you're free, thanks!

setup.py
```diff
@@ -18,13 +18,72 @@
...
+            terminate = path == paths[-1]
+        except request.HTTPError:
+            break
+
+for path in paths:
+    if not os.path.exists(path):
+        print(
+            f"Could not download {path}. Check out the unifyai/binaries README for"
+            " the supported configurations!"
+        )
```
guillesanbri, line 84: lgtm!

guillesanbri (approved):

setup.py
guillesanbri: Ah got it yeah, this works fine then, thanks Ved! :+1:

requirements/requirements.txt
```diff
@@ -7,6 +7,8 @@ packaging
 nvidia-ml-py # mod_name=pynvml
 diskcache
 google-auth # mod_name=google.auth
+urllib3<2.0 # for version conflict with google-auth
 requests
 pyvis
 dill
+tqdm
```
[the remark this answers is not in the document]
vedpatwardhan: reverted the changes to `requirements` and `optional` files 👍
actually `tqdm` is being used in the `determine_test_coverage.py` file, probably @RashulChutani could comment on it's usability 😅

vedpatwardhan:
> 1. Removed all binaries from the `ivy` repo and added them instead to the [binaries](https://github.com/unifyai/binaries) repo with the naming convention same as for pypi wheels for consistency of naming. Also added a `binaries.json` to `ivy` for the skeleton structure. We'd need to add more binaries for different architectures and also different python versions as the current ones are only for python 3.9 and 3.10, but @vaatsalya123 could you please take a look at this part of the changes?
> 3. In the rare case of someone wanting to clone repo, checkout to an older release (through the tag) and pulling the binaries corresponding to that release, I've added a "VERSION" environment variable they can set. But for this to be useful, we'd first need to ensure that we also add tags to the `binaries` repo when we make our releases. @RashulChutani could you please take a look at the changes made to `pypi.yml`?

Hey @vaatsalya123 and @RashulChutani, could you please take a look? Given it's directly linked to the binaries, I'm refraining from merging without multiple opinions. Thanks!

requirements/requirements.txt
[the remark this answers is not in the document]
guillesanbri: ah missed that, true!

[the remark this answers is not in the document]
vedpatwardhan: no worries!

vedpatwardhan:
> Overall, it looks good to me. However, the binaries code is better suited to be in the release workflow instead of pypi.yml, as that is dedicated for PyPI deployment.

Thanks for the review, I've basically added a `binaries.yml` for adding the tags to the binaries repo and added it as a job to `release.yml`, happy to know your thoughts 😄

vaatsalya123:
@vedpatwardhan All look's perfect to me , Just so I'm sure did you test the full pipeline in action to confirm it's working or are we yet to do this?

vedpatwardhan:
I'll get this merged for now, let me know if there's anything to be updated with the `binaries.yml` workflow @RashulChutani I'll get that fixed. Ofc given that the `binaries.yml` workflow wouldn't be used until we make a release, it would lead to any immediate problems so merging the PR for now. Thanks all 😄

PR #23136, opened 2023-09-06
Made Mish Activation function support complex dtype

mohame54:
# PR Description 
Made some modification to **mish** backend implementations and their corresponding tests 
For complex dtype support


ivy-leaves:

Thank you :hugs:

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -719,9 +719,12 @@ def tanh(
...
     if paddle.is_complex(x):
-        tanh_a = paddle.tanh(paddle.real(x))
-        tan_b = paddle.tan(paddle.imag(x))
-        return (tanh_a + 1j * tan_b) / (1 + 1j * (tanh_a * tan_b))
+        const = paddle.to_tensor(1j, dtype=x.dtype)
+        tanh_a = paddle.cast(paddle.tanh(x.real()), x.dtype)
+        tan_b = paddle.cast(paddle.tan(x.imag()), x.dtype)
+        term1 = tanh_a + paddle.multiply(tan_b, const)
+        term2 = 1 + paddle.multiply(const, (tanh_a * tan_b))
+        return term1 / term2
```
jshepherd01, line 727: I'm not sure I see the reason for changing this.

Also, rather than casting a float to complex and then multiplying by 1j, it would probably be better to use the `paddle.complex(real, imag)` function as in:
```py
tanh_a = paddle.tanh(paddle.real(x))
tan_b = paddle.tan(paddle.imag(x))
term1 = paddle.complex(tanh_a, tan_b)
term2 = paddle.complex(1, tanh_a * tan_b)
return term1 / term2
```

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
```diff
@@ -136,14 +136,19 @@ def test_log_softmax(*, dtype_and_x, axis, test_flags, backend_fw, fn_name, on_d
...
+        min_dim_size=2,
         safety_factor_scale="log",
     ),
+    ground_truth_backend="jax",
 )
 def test_mish(*, dtype_and_x, test_flags, backend_fw, fn_name, on_device):
     dtype, x = dtype_and_x
+    x = x[0]
+    if len(x.shape) < 2:
+        x = x.reshape(-1, 1)
```
jshepherd01, line 151: Why is this needed? Calling `ivy.mish` on a 1D or even 0D array shouldn't cause any problems

jshepherd01 (commented):
Looks good to me, just a couple of questions to clarify some things

jshepherd01:
You might also want to consider increasing the safety factor even further, it seems to fail tests for any reasonably large value of x (which is reasonable, because it needs to compute numbers on the order of `e^{2*Re[x]}`, so an x value of about 350 would be very near to an overflow for complex128)

ivy/functional/backends/paddle/elementwise.py
mohame54: I have done that cuz paddle was giving me errors related to paddle functional API but I will modify the tanh implementation  as yo suggested.

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
mohame54: it's something related also to paddle backend when I extract the real and imag part of the complex input it transforms them to **0D** although the input is 1D dimension array.

mohame54:
okay on it.

ivy/functional/backends/paddle/elementwise.py
mohame54: it seems there is a problem with paddle.complex not creating output if the real and imaginary part not of the same type
so I kept it the same before .
image

jshepherd01: It's likely the 1, you might need to do `paddle.ones_like(tanh_a)`

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
jshepherd01: Isn't this still doing `x[0]` afterwards, so if the input was 1D then it's still getting a 1D array but only with one element

mohame54: x is initially a list of arrays I took the first array from the list and expand it's dims.

ivy/stateful/module.py
```diff
@@ -6,7 +6,7 @@
 import os
 import abc
 import copy
-import dill
+#import dill
```
jshepherd01, line 9: `dill` seems to be used in this file, you should leave the import in

ivy_tests/test_ivy/test_functional/test_nn/test_activations.py
jshepherd01: Yes, but the test still does `x[0]` afterwards, so for an input of say `x = [ivy.array([1,2,3])]` this process turns it into `x = ivy.array([[1], [2], [3]])` and then the input to `test_function` is `x = x[0] = ivy.array([1])`. I also tried removing this section and re-running it locally and the tests still passed

ivy/functional/backends/torch/activations.py
```diff
@@ -125,13 +125,16 @@ def log_softmax(
 @with_unsupported_dtypes(
     {
         "2.0.1 and below": (
-            "complex",
+            "bfloat16",
```
jshepherd01, line 128: This surprises me, torch is usually fine with `bfloat16` on cpu. Are you sure it's the function that doesn't support it, rather than the testing pipeline?

jshepherd01 (changes requested):

mohame54:
>Yes, but the test still does x[0] afterwards, so for an input of say x = [ivy.array([1,2,3])] this process turns it into x = ivy.array([[1],
>[2], [3]]) and then the input to test_function is x = x[0] = ivy.array([1]). I also tried removing this section and re-running it locally >and the tests still passed
Oh, I totally forgot about it.

jshepherd01:
You might need to run the precommit again, the linter is still failing.

Also about the `x.reshape` thing, I really think that shouldn't be necessary in the test since `ivy.mish(1+2j)` should, in principle, work. If removing that causes tests to fail, it should probably be fixed in the backend functions themselves rather than by modifying the test

mohame54:
it causes errors with paddle backend only

jshepherd01:
> it causes errors with paddle backend only

I haven't been able to reproduce the error, would you be able to provide a screenshot of it?

mohame54:
>I haven't been able to reproduce the error, would you be able to provide a screenshot of it?

No worries managed to solve them.

ivy/functional/ivy/activations.py
```diff
@@ -663,16 +663,27 @@ def softsign(
...
+    *,
+    complex_mode: Literal["split", "magnitude", "jax"] = "jax",
+    out: Optional[ivy.Array] = None,
 ) -> ivy.Array:
     """
     Apply the mish activation function element-wise.
 
+    If the input is complex we apply the mish function to the input
+    ``mish(x) = x * tanh(log(1+ e^x))``.
+
```
jshepherd01, line 679: I think you'd be fine without this part of the docstring, since that's also what happens when the input is real

jshepherd01 (changes requested):
Just a minor docstring change and also the thing about the function signatures that I mentioned on the logit PR applies here too

mohame54:
>Just a minor docstring change and also the thing about the function signatures that I mentioned on the logit PR applies here too
yeah I totally forgot about them.

jshepherd01:
> > Just a minor docstring change and also the thing about the function signatures that I mentioned on the logit PR applies here too
> yeah I totally forgot about them.

No worries, I keep forgetting about them too

PR #23159, opened 2023-09-06
Added ivy.Array support to jax native compile

Killua7362:
Works with jit,lax.map,vmap and jvp.
It fails for [jax.vjp]([https://jax.readthedocs.io/en/latest/_autosummary/jax.vjp.html#jax.vjp).


vedpatwardhan (approved):
lgtm! Let's get this merged as a version 0 as it fixes most of the problems except `vjp`. Thanks @Killua7362 😄

PR #23188, opened 2023-09-07
refactored promote_types_of_torch_inputs to handle an arbitrary numbe…

mosesdaudu001:
…r of input

# Refactored `promote_types_of_torch_inputs` to support an arbitrary number of inputs

## Close https://trello.com/c/G4BMCLwA/2532-refactor-promotetypesoftorchinputs-to-handle-an-arbitrary-number-of-inputs

Close #


ivy-leaves:

Thank you :hugs:

ivy/functional/frontends/torch/__init__.py
```diff
@@ -230,23 +228,31 @@ def promote_types_of_torch_inputs(
...
+
+        def nested_multi(x):
+            x = ivy.asarray(
+                x,
+                dtype=non_empty_array_val.dtype,
+                device=ivy.default_device(item=non_empty_array_val, as_native=False),
+            )
+            return x
+
+        ivy.nested_map(empty_array_vals, nested_multi)
```
AnnaTz, line 250: Does this statement do anything? You are not assigning to any variable 🤔

ivy/functional/frontends/torch/__init__.py
```diff
@@ -230,23 +228,31 @@ def promote_types_of_torch_inputs(
...
+    else:
+        for arg in args:
+            promoted = promote_types_torch(arg.dtype, args[0].dtype)
+            arg = ivy.asarray(arg, dtype=promoted)
```
AnnaTz, line 254: And here I don't think you are updating `args` at all.

ivy/functional/frontends/torch/__init__.py
```diff
@@ -230,23 +228,31 @@ def promote_types_of_torch_inputs(
...
+    return args
```
AnnaTz, line 256: You will need to write a test for this to make sure it works as expected. This will help you in debugging as well.
(I actually just noticed there are functions in the `__init__` files that are not being tested. I will add a card for this.)

AnnaTz (changes requested):

ivy/functional/frontends/torch/__init__.py
mosesdaudu001: Yes the statement recursively applies the `nested_multi` function to the `empty_array_vals`. We do not need to explicitly assign it to another variable. Once the `nested_map` has applied the mapping, the variables maintain their new characteristics. Although I think it's best practice to assign a variable for easy understanding.

ivy/functional/frontends/torch/__init__.py
mosesdaudu001: What happens here is the `else` statement captures occasions where the inputs have different dtypes.

So for these occasions, we want all the inputs to have the same dtypes and thats why we promote all the inputs to have the same dtype as the first input. It actually does update the specifc argument to have the promoted dtype. Again this is for cases where the inputs have different dtypes

ivy/functional/frontends/torch/__init__.py
mosesdaudu001: No problem, I already picked up the card and would start working soon on it.

PR #23207, opened 2023-09-07
scipy test fixes

mobley-trent:
Test fixes for functions in the `ivy.frontends.scipy.linalg` namespace


mobley-trent:
Hey @umairjavaid how do I filter `float16` and `bfloat16` dtypes from the dtype helper. I tried using the filter strategy but it doesn't work for some reason

MahmoudAshraf97 (changes requested):
the only remaining dtypes from this strategy are float32 and 64, so why not pass them directly as the available dtypes? easier and more readible

PR #23213, opened 2023-09-07
Made logit function support complex dtypes

mohame54:
Added the doc string and the sufficient code to make logit function support complex dtypes


ivy/functional/ivy/experimental/activations.py
```diff
@@ -14,21 +14,58 @@
...
+    output = ivy.log(
+        ivy.divide(
+            x,
+            ivy.subtract(
+                ivy.array(1).astype(x.dtype),
+                x,
+            ),
+        ),
+    )
+    return output
```
jshepherd01, line 52: This can most likely be simplified to leverage the existing backend implementations of logit. Why not use something like
```py
real = ivy.real(x)
imag = ivy.imag(x)
if eps is not None:
    real = ivy.clip(real, eps, 1-eps)
z = ivy.add(real, ivy.multiply(ivy.array(1j, dtype=x.dtype), imag))
return fn_original(z, eps=eps, out=out)
```

jshepherd01 (changes requested):
One question about the `jax_like` function. You should also add the `complex_mode="jax"` argument to each of the backend functions, because they should have the same call signature as the function in `ivy/experimental`

ivy/functional/ivy/experimental/activations.py
mohame54: you are always better than me at simplifying code.

mohame54: But I have one question why should we use fn_original my code handles if the real part is bigger than 1 or is smaller than zero  and also handle eps argument so we don't need the fn_original

jshepherd01: Ah, yeah, true. I was thinking to use `fn_original` to put the effort of actually doing the calculation on the backend, since they might have quicker implementations of `logit` than this function does. You make a good point about nans though, and looking at e.g. the paddle implementation it wouldn't really be able to handle that properly.

Okay, fair enough, you've convinced me. I do think it would be best to remove the `.astype`s though, since it should be fine without them if everything else is type safe (inside the subtract you can do `ivy.array(1., dtype=x.dtype)` instead of having a separate call to `.astype`)

jshepherd01:
I meant that the backend functions should have identical signatures to the functional API, so here we have `logit(x, /, *, eps=None, complex_mode="jax", out=None)` and that means each backend implementation should also have `logit(x, /, *, eps=None, complex_mode="jax", out=None)`. I realised I forgot to check on that for softmax (and maybe I forgot to mention it earlier, this came from Ved's comment on a PR of mine quite recently), I can make those changes myself though

mohame54:
Oh I got it i will change modify them at once.

ivy/functional/ivy/experimental/activations.py
mohame54: yeah sure.

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
```diff
@@ -1498,7 +1498,7 @@ def test_paddle_log2(
 @handle_frontend_test(
     fn_tree="paddle.logit",
     dtype_and_x=helpers.dtype_and_values(
-        available_dtypes=helpers.get_dtypes("float"),
+        available_dtypes=helpers.get_dtypes("float_and_complex"),
```
jshepherd01, line 1501: I think the paddle framework itself doesn't support complex so this shouldn't be included (the frontends only need to support the dtypes that the corresponding framework supports)

jshepherd01 (changes requested):
Just the one tiny change to the paddle frontend test, once that's done it's ready to merge

ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_math.py
mohame54: yeah yeah I forgot also about that

PR #23313, opened 2023-09-09
Stateful changes 2

RickSanchezStoic:
feature: added to_device method to Module


ivy/stateful/module.py
```diff
@@ -680,7 +680,6 @@ def build(
         self,
         *args,
         from_call=False,
-        device=None,
```
vedpatwardhan, line 683: I think we can still continue with a `device` argument in the `__init__` that marks some default device which we'd use. 
Also, I think we can have a `model.astype` method as well, just came across [this](https://discuss.pytorch.org/t/is-there-an-efficient-way-to-convert-a-model-with-float32-params-to-float64/130767/2) method in `torch` with the same structure, what do you think?

ivy/stateful/module.py
```diff
@@ -705,7 +704,6 @@ def build(
         ret
             True for successfully built a module.
         """
-        self._dev = ivy.default(device, self._dev)
```
vedpatwardhan, line 708: as mentioned above, we can continue having `_dev` in the class which would have a read-only property corresponding to it. Let's probably rename this to `_device` instead as that's the vocab across the rest of the framework and add a `device` property to `ivy.Module` to access `self._device`

ivy/stateful/module.py
```diff
@@ -832,6 +830,27 @@ def train(self, mode: bool = True):
...
+        # moving weights and buffers to new device
+        for key, obj in self.v.items():
+            if (
+                isinstance(obj, ivy.Module)
+                or ivy.is_ivy_array(obj)
+                or isinstance(obj, ivy.Container)
+            ):
+                obj.to_device(device)
+            else:
+                ivy.to_device(obj, device=device)
```
vedpatwardhan, line 846: Actually, `to_device` isn't inplace by default, so you'd need to either do a `ivy.to_device(obj, device=device, out=obj)` or a `obj._data = ivy.to_device(obj).data`

ivy/stateful/module.py
```diff
@@ -832,6 +830,27 @@ def train(self, mode: bool = True):
...
+        if getattr(self, "buffers", None):
```
vedpatwardhan, line 847: I think we could refactor this further if `self.buffers` was a `{}` by default as we wouldn't need this check and we could also just instead add a `state_dict` and iterate over it instead of having 2 different loops for variables and buffers, what do you think?

vedpatwardhan (changes requested):
Hey @RickSanchezStoic, just requested a few changes, thanks 😄

ivy/stateful/module.py
vedpatwardhan: Thanks for the refactor, could you please also add the `ivy.Module.device` property to allow access to the `_device` private variable?

ivy/stateful/module.py
```diff
@@ -832,6 +833,21 @@ def train(self, mode: bool = True):
             if isinstance(module, ivy.Module):
                 module.train(mode=mode)
 
+    def to_device(self, device, out=None):
```
vedpatwardhan, line 836: let's remove the `out` argument, ig most people would assume `model.to_device` to be inplace by default

vedpatwardhan (changes requested):
Final couple of changes, thanks for making the changes @RickSanchezStoic 😄

vedpatwardhan (approved):

PR #23383, opened 2023-09-10
refactored handle_exceptions decorator

Madjid-CH:
# PR Description

I saw that the handle_exceptions decorator is starting to get very long so I took the liberty in adding some tests to it and refactoring it. 
Of course if this seems unnecessary we can just forget about this idea😄

### DIFF
--- a/ivy/utils/exceptions.py
+++ b/ivy/utils/exceptions.py
@@ -329,6 +329,14 @@ class InplaceUpdateException(IvyException):
         super().__init__(*messages, include_backend=include_backend)
 
 
+non_ivy_exceptions_dict = {
+    IndexError: IvyIndexError,
+    AttributeError: IvyAttributeError,
+    ValueError: IvyValueError,
+    Exception: IvyBackendException,
+}
+
+
 def handle_exceptions(fn: Callable) -> Callable:
     @functools.wraps(fn)
     def _handle_exceptions(*args, **kwargs):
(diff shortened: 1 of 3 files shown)


ivy/utils/exceptions.py
```diff
@@ -349,58 +357,20 @@ def _handle_exceptions(*args, **kwargs):
...
-            raise ivy.utils.exceptions.IvyValueError(
-                fn.__name__, str(e), include_backend=True
-            )
-        except IvyDeviceError as e:
-            _configure_stack_trace(e.__traceback__)
-            raise ivy.utils.exceptions.IvyDeviceError(
-                fn.__name__, str(e), include_backend=True
-            )
-        except InvalidBackendException as e:
+        except (IvyNotImplementedException, InvalidBackendException) as e:
```
vedpatwardhan, line 360: Not sure why we have a separate except block for these 2 exceptions. I think we can just add `IvyNotImplementedException` to the `_non_ivy_exception_mapping` and handle `InvalidBackendException` through the `IvyException` except block

vedpatwardhan (changes requested):
Hey @Madjid-CH, this will be a great addition to the codebase actually, just some minor changes needed, thanks 😄

ivy/utils/exceptions.py
Madjid-CH: okay, I'll do that

vedpatwardhan (approved):
(There's a few tests failing in the CI, but they seem unrelated from what I see)

PR #23389, opened 2023-09-11
fixed wrong returned shape of paddle.mean function

Mghrabi:
This PR is meant to fix the wrong returned shape by mean function `AssertionError: returned shape = (1,), ground-truth returned shape = ()`


mohame54:
Hi @Mghrabi 

Thanks for the PR!

PR #23430, opened 2023-09-11
fix: Take into account frontend supported dtypes in frontend tests

ReneFabricius:
# PR Description
There is a problem with function `helpers.get_dtypes`.

If `prune_function==True` backend and function are taken into account through `_get_supported_devices_dtypes`.

If prune_function==False,
`_get_type_dict` is used as `retrieval_fn`
This `retrieval_fn` is used with parameter `is_frontend_test=False` which leads to considering only ivy framework dtypes,
if `is_frontend_test` was `True`, the passed framework would be considered frontend and only that would be taken into account.

The `prune_function==True` branch seems to miss the frontend dtypes in case we are in a frontend test.
The `prune_function==False` branch, the way it is currently used, seems to miss both frontend and backend dtypes, only ivy dtypes are considered.

[Trello](https://trello.com/c/Iva4muMy/2755-frontend-tests-not-considering-ground-truth-backend-for-helpersgetdtypes)


ivy_tests/test_ivy/helpers/hypothesis_helpers/dtype_helpers.py
```diff
@@ -50,9 +50,13 @@ def _get_type_dict(framework: str, kind: str, is_frontend_test=False):
...
     if is_frontend_test:
         framework_module = get_frontend_config(framework).supported_dtypes
+        return _retrieve_requested_dtypes(framework_module, kind)
     else:
-        framework_module = ivy
+        with WithBackendContext(framework) as ivy_backend:
+            return _retrieve_requested_dtypes(ivy_backend, kind)
 
+
+def _retrieve_requested_dtypes(framework_module, kind):
```
RickSanchezStoic, line 59: we don't need to write a new function for this, you can simply assign the changes to framework_module, the changes made within context will persists.

PR #23433, opened 2023-09-11
removed infer_device and changed handle_device_shifting to handle_device

mosesdaudu001:
# Replaced `handle_device_shifting` with `handle_device` and also removed the `infer_device` decorators where `handle_device` was present.

### DIFF
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -17,7 +17,7 @@ from ivy.utils.exceptions import IvyValueError
 FN_DECORATORS = [
     "handle_complex_input",
     "infer_device",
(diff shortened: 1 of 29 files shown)


mosesdaudu001:
There were some cases where the `infer_device` decorator was solely used without the `handle_device` decorator. In such cases I did not remove the `infer_device` decorator.

Examples of such cases includes:
https://github.com/unifyai/ivy/blob/main/ivy/functional/ivy/experimental/creation.py#L288
https://github.com/unifyai/ivy/blob/main/ivy/functional/ivy/experimental/creation.py#L385

PR #23436, opened 2023-09-11
Complex dtype sigmoid

mosesdaudu001:
# added decorator and argument for complex dtype sigmoid across backend, functional API, data_classes and stateful

### DIFF
--- a/ivy/functional/backends/tensorflow/activations.py
+++ b/ivy/functional/backends/tensorflow/activations.py
@@ -46,7 +46,7 @@ def relu(x: Tensor, /, *, complex_mode="jax", out: Optional[Tensor] = None) -> T
     return tf.nn.relu(x)
 
 
-def sigmoid(x: Tensor, /, *, out: Optional[Tensor] = None) -> Tensor:
+def sigmoid(x: Tensor, /, *, omplex_mode="jax", out: Optional[Tensor] = None) -> Tensor:
     if not ivy.is_array(x):
         x = float(x)
     return tf.nn.sigmoid(x)
--- a/ivy/functional/backends/tensorflow/activations.py
+++ b/ivy/functional/backends/tensorflow/activations.py
@@ -46,7 +46,7 @@ def relu(x: Tensor, /, *, complex_mode="jax", out: Optional[Tensor] = None) -> T
     return tf.nn.relu(x)
 
 
-def sigmoid(x: Tensor, /, *, omplex_mode="jax", out: Optional[Tensor] = None) -> Tensor:
+def sigmoid(x: Tensor, /, *, complex_mode="jax", out: Optional[Tensor] = None) -> Tensor:
     if not ivy.is_array(x):
         x = float(x)
     return tf.nn.sigmoid(x)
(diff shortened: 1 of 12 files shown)


ivy-leaves:

Thank you :hugs:

ivy/functional/backends/tensorflow/activations.py
```diff
@@ -46,7 +46,8 @@ def relu(x: Tensor, /, *, complex_mode="jax", out: Optional[Tensor] = None) -> T
     return tf.nn.relu(x)
 
 
-def sigmoid(x: Tensor, /, *, out: Optional[Tensor] = None) -> Tensor:
+@with_unsupported_dtypes({"2.13.0 and below": ("complex",)}, backend_version)
```
jshepherd01, line 49: Why are we marking complex as unsupported for this?

mosesdaudu001: I think we has this discussion earlier. Even though `tensorflow` does support complex dtypes, We don't get the same results when compared with `jax`. A quick fix for this issue was the `sigmoid_jax_like` function.

The implementation of the sigmoid function in the `tensorflow` backend is correct, but does not match the result from `jax`

jshepherd01: That's a problem with tensorflow (this issue here https://github.com/tensorflow/tensorflow/issues/61800) which will be fixed once they solve it on their end. I think marking it as an unsupported dtype is the wrong move, it might be better to leave it as supported but change the ground truth backend in the test to, for example, jax so that only the tf backend test will fail

ivy/functional/backends/tensorflow/activations.py
```diff
@@ -46,7 +46,7 @@ def relu(x: Tensor, /, *, complex_mode="jax", out: Optional[Tensor] = None) -> T
     return tf.nn.relu(x)
 
 
-def sigmoid(x: Tensor, /, *, out: Optional[Tensor] = None) -> Tensor:
+def sigmoid(x: Tensor, /, *, omplex_mode="jax", out: Optional[Tensor] = None) -> Tensor:
```
jshepherd01, line 49:
```suggestion
def sigmoid(x: Tensor, /, *, complex_mode="jax", out: Optional[Tensor] = None) -> Tensor:
```

ivy/functional/backends/tensorflow/activations.py
```diff
@@ -46,7 +46,7 @@ def relu(x: Tensor, /, *, complex_mode="jax", out: Optional[Tensor] = None) -> T
     return tf.nn.relu(x)
 
 
-def sigmoid(x: Tensor, /, *, out: Optional[Tensor] = None) -> Tensor:
+def sigmoid(x: Tensor, /, *, complex_mode="jax", out: Optional[Tensor] = None) -> Tensor:
```
jshepherd01, line 49:
```suggestion
def sigmoid(
    x: Tensor, /, *, complex_mode="jax", out: Optional[Tensor] = None
) -> Tensor:
```

PR #23439, opened 2023-09-11
feature and fix: added support for casting modes to give the original…

RickSanchezStoic:
added support for casting modes to give the original expected output type, also modified promote_types to accept None


ivy/func_wrapper.py
```diff
@@ -1162,9 +1162,13 @@ def _wrap_function(
     return to_wrap
 
 
-def casting_modes_ops(fn):
+def casting_modes_ops(fn, dtype_determinator=None):
```
vedpatwardhan, line 1165: everything else looks great! let's probably rename the argument to `ret_dtype_target` or something which might be more informative, what do you think?
I'll request @MahmoudAshraf97's review as well on this PR, thanks for making the changes @RickSanchezStoic 💪

vedpatwardhan (changes requested):

ivy/func_wrapper.py
```diff
@@ -1181,11 +1185,23 @@ def method(*args, **kwargs):
...
                 kwargs["dtype"] = ivy.as_native_dtype(dtype)
 
+        if not to_cast and ret_dtype_target:
+            for arg in ret_dtype_target:
+                to_cast = ivy.promote_types(
+                    to_cast,
+                    ivy.dtype(
+                        args[arg_names.index(arg)] if arg not in kwargs else kwargs[arg]
+                    ),
+                )
```
MahmoudAshraf97, line 1203: this does not take into consideration when the inputs are python scalars or other objects that do not have a `dtype` attribute, we should stick to using `promote_types_of_inputs` to guarantee consistent results, also to prevent the existence of two identical promotion codes which will double the maintenance efforts

MahmoudAshraf97: another problem is that this doesn't take into consideration default values that are optional but still affect the output dtype

MahmoudAshraf97 (changes requested):

MahmoudAshraf97:
I think we will need a method to test if this decorator is working as expected or not, my suggestion is to enable testing functions using their unsupported dtypes, this will guarentee consistency with frameworks that do support these dtypes

ivy/func_wrapper.py
```diff
@@ -1196,7 +1225,9 @@ def mini_helper(x):
 
         args = ivy.nested_map(args, mini_helper, include_derived=True)
         kwargs = ivy.nested_map(kwargs, mini_helper)
-        return fn(*args, **kwargs)
+        return (
+            ivy.astype(fn(*args, **kwargs), cast_to) if cast_to else fn(*args, **kwargs)
```
vedpatwardhan, line 1229: we could just do a `ret = fn(*args, **kwargs)` on a previous line instead of calling `fn(*args, **kwargs)` in both conditions, thanks!

vedpatwardhan (changes requested):
Just a minor change, rest looks good to me. I agree with @MahmoudAshraf97, would be great to have tests for these too. Also, does this cover all the other corner cases we were discussing? Thanks 😄

ivy/func_wrapper.py
MahmoudAshraf97: Just asking for the sake of knowledge, does this make a difference in performance or just cleaner code?

vedpatwardhan: yeah just cleaner code 👍

ivy/func_wrapper.py
vedpatwardhan: Hey @MahmoudAshraf97 I remember having this conversation a couple of days back but don't fully remember the specifics, could you please share an example where this would be applicable? Thanks and apologies 😅

MahmoudAshraf97:
Hey @RickSanchezStoic , I guess you reverted my commit when you committed, can you please revise?

ivy/func_wrapper.py
```diff
@@ -1159,7 +1166,36 @@ def mini_helper(x):
...
+                        (
+                            args[arg_names.index(arg)]
+                            if arg not in kwargs
+                            else kwargs[arg]
+                        ),
+                    )
+                    if arg not in kwargs:
+                        args[arg_names.index(arg)] = (
+                            arg_mod
+                            if not ivy.isarray(args[arg_names.index(arg)])
```
vedpatwardhan, line 1184: this should be `ivy.is_array` instead of `ivy.isarray`

ivy/func_wrapper.py
```diff
@@ -1159,7 +1166,36 @@ def mini_helper(x):
...
+                            else args[arg_names.index(arg)]
+                        )
+                    else:
+                        kwargs[arg] = (
+                            arg_mod
+                            if not ivy.isarray(args[arg_names.index(arg)])
```
vedpatwardhan, line 1190: same here

ivy/func_wrapper.py
```diff
@@ -1300,7 +1336,8 @@ def _wrapped(func):
             if "frontends" in func.__module__:
                 # it's a frontend func, no casting modes for this
                 return func
-            return casting_modes_ops(func)
+
+            return casting_modes_ops(func, ret_dtype_target)
```
vedpatwardhan, line 1340: let's remove the blank line above this and pass `ret_dtype_target` as a keyword argument

ivy/functional/ivy/data_type.py
```diff
@@ -2102,6 +2102,9 @@ def promote_types(
     ret
         The type that both input types promote to
     """
+    # in case either is of none type
+    if not (type1 and type2):
+        return type1 if type1 else type2
```
vedpatwardhan, line 2107: wouldn't `return type1 if type1 else type2` on this line be the same as `return` alone?

vedpatwardhan (changes requested):
Just requested a few changes! Let's focus on the functions which use `ivy.promote_types_of_inputs` for now for a version 0 as there isn't really an exhaustive way to test if there are issues in the output dtype for the other functions until we integrate casting modes into the tests (cc @sherry30), I'll also request @MahmoudAshraf97's review on this PR. Thanks @RickSanchezStoic 😄

ivy/func_wrapper.py
MahmoudAshraf97: We do have promotion rules for scalars included in `ivy.promote_types_of_inputs` but not in the promotion table hence they are promoted to the default dtype in `ivy.promote_types` which will lead to an inconsistent promotion results

`ivy.promote_types(float_scalar, float16) = float32`
`ivy.promote_types_of_inputs(float_scalar, float16) = float16`

vedpatwardhan: Actually `ivy.promote_types` should always receive dtypes (`ivy.Dtype` or `ivy.NativeDtype`), so ideally the result from `ivy.promote_types_of_inputs` is the one which would be used anyway. So ideally `float16` would be the promoted dtype as a result 😄

ivy/functional/ivy/data_type.py
RickSanchezStoic: didn't get you, the logic here is to return either of type1 or type2 if either of them are None.

vedpatwardhan: ah yeah, my bad sorry!

PR #23466, opened 2023-09-12
adding missing attributes to some frontend decorators

YushaArif99:


ivy/func_wrapper.py
```diff
@@ -1312,7 +1312,7 @@ def _wrapped(func):
                         # applied to the function, but they are not same
                         # and aren't in conflicting dict either
                         setattr(func, attrib, val)
-                        setattr(func, "dictionary_info", version_dict)
+                        setattr(func, "dictionary_info", (version_dict, version))
```
vedpatwardhan, line 1315: could you please also update other lines of code like [1322](https://github.com/unifyai/ivy/pull/23466/files#diff-e507595f1a13b0ef477a10a0399fa210f0708466a9f19197a27dda65986bbf09R1322) which expect to receive a `dict` but would now receive a `tuple`? I'll also request @RickSanchezStoic's review before merging. Thanks @YushaArif99!

vedpatwardhan (changes requested):

ivy/func_wrapper.py
YushaArif99: oh yeah forgot to update that. Thanks for pointing it out 👍🏼

RickSanchezStoic (approved):

PR #23474, opened 2023-09-12
fix: error when current framework version not found in dtype decorator

jshepherd01:
In the dtype decorators, modify the default value (if the current framework version is not described by the decorator) to be that of the most recent version that was described, rather than an empty tuple. This fixes some errors that occur when new framework versions are released


jshepherd01:
Taking a different approach on this

PR #23489, opened 2023-09-12
feat: throwing a warning when a sub-backend implementation is available but is not being used.

Madjid-CH:
# PR Description
This PR is for this [task](https://trello.com/c/6Bo0xwIw)
my solution is to:
1) during the backend setting `ivy.set_backend`  I check available sub-backend implementations by analyzing the sub backend module and put the result in a dictionary.
2) inside the `_set_backend_as_ivy ` I check if a given symbols has a sub backend implementation from the already populated dictionary in step 1 and if this the case we wrap the function with a wrapper that throws a warning when ever the function is used.

there's an issue which is I need to import the ivy sub backend (for example [this](https://github.com/unifyai/ivy/tree/main/ivy/functional/backends/torch/sub_backends/xformers)) dynamically and analyze it symbols to do step 1 this may lead to some errors if the real sub backend(for example [this](https://github.com/facebookresearch/xformers)) is not installed properly (missing some cuda toolkits or something like that) so either we catch the error and abort the search or we search the sub backend module without importing it (some sort of static analysis) but I didn't find a way to do the later.


ivy/func_wrapper.py
```diff
@@ -1580,6 +1580,38 @@ def func(x):
     return _handle_backend_invalid
 
 
+def _warn_efficient_implementation_available(fn: Callable) -> Callable:
+    @functools.wraps(fn)
+    def _wrapper(*args, **kwargs):
```
vedpatwardhan, line 1585: it would be great to name the wrapper function and the wrapped function in the same way as the other function wrappers

ivy/utils/backend/handler.py
```diff
@@ -243,7 +243,7 @@ def current_backend(*args, **kwargs):
     return importlib.import_module(_backend_dict[implicit_backend])
 
 
-def _set_backend_as_ivy(
```
vedpatwardhan, line 246: I prefer `_set_backend_as_ivy` as it's tends to be more informative than `_modify_module_symbols`

ivy/utils/backend/handler.py
```diff
@@ -252,10 +252,11 @@ def _set_backend_as_ivy(
     backend_str = backend.current_backend_str() if backend_str is None else backend_str
     for k, v in original_dict.items():
         compositional = k not in backend.__dict__
-        if k not in backend.__dict__:
+        if compositional:
             if k in invalid_dtypes and k in target.__dict__:
                 del target.__dict__[k]
                 continue
+            v = _verify_efficient_implementations(v) if v else v
```
vedpatwardhan, line 259: actually we should also be checking for efficient backend implementations for non-compositional functions. 
Ofc as far as `scaled_dot_product_attention` goes there's a compositional implementation and a `xformers` sub-backend implementation with no `torch` backend implementation, but it could be there for some other sub-backend added in the future

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -200,3 +210,65 @@ def find_available_sub_backends(sub_backends_loc):
...
+    for sub in sub_backends:
+        sub_backend = ivy.utils.dynamic_import.import_module(_sub_backend_dict[sub])
+        for k, v in sub_backend.__dict__.items():
+            if isinstance(v, Callable) and not k.startswith("__"):
+                result[k] = result.get(k, []) + [sub]
+
+    return result
+
+
+def available_sub_backend_implementations(obj: Union[Callable, str]) -> list:
```
vedpatwardhan, line 226: I think it would be fine to only accept a `fn_name: str` instead which would be the function name to avoid complicating the implementation

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -200,3 +210,65 @@ def find_available_sub_backends(sub_backends_loc):
...
+    sub_backends = ivy.current_backend().available_sub_backends()
+    result = []
+    if not sub_backends:
+        return result
+    if (
+        ivy.current_backend_str()
+        not in _available_sub_backends_implementations_dict.keys()
+    ):
+        _available_sub_backends_implementations_dict[ivy.current_backend_str()] = (
+            _find_available_sub_backend_implementations(sub_backends)
```
vedpatwardhan, line 260: we could simplify this helper right? I mean we can only look for `fn_name` in the `sub_backend.__dict__` instead of iterating over the dict?

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -200,3 +210,65 @@ def find_available_sub_backends(sub_backends_loc):
...
+        )
+    return _available_sub_backends_implementations_dict[ivy.current_backend_str()].get(
+        obj.__name__, []
+    )
+
+
+def _check_callable(obj):
```
vedpatwardhan, line 267: we wouldn't need this helper after changing `obj` to only be a `str` in the `available_sub_backend_implementations` function

vedpatwardhan (changes requested):
Hey @Madjid-CH, just requested a few changes, thanks 😄

ivy/utils/backend/handler.py
Madjid-CH: since it was used recursively it does not only modify ivy but all the sub modules, so I though that `_set_backend_as_ivy` does not tell the whole purpose of the function.
does that make sense ?

ivy/utils/backend/sub_backend_handler.py
Madjid-CH: I did this to avoid the user passing something like ` ivy.available_sub_backend_implementations("__name__")`

ivy/utils/backend/sub_backend_handler.py
Madjid-CH: Actually, here we check if we already verified for sub backends implementations (i.e. populated the  `_available_sub_backends_implementations_dict`) if not we populate it.

ivy/utils/backend/handler.py
Madjid-CH: sure, I'll do that

ivy/utils/backend/handler.py
vedpatwardhan: Yeah sounds good, but I think `symbols` could also be a bit misleading, also `set` would be a better prefix than `modify`. Probably let's settle on `_set_module_backend_as_ivy`, what do you think?

ivy/utils/backend/handler.py
```diff
@@ -252,11 +252,13 @@ def _set_backend_as_ivy(
...
             backend.__dict__[k] = v
+        v = _verify_efficient_implementations(v) if v else v
```
vedpatwardhan, line 261: probably we could do this before the `if compositional` block to avoid duplication

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -2,60 +2,70 @@
...
+
 
+_detect_sub_backends_dynamically()
 
-_all_sub_backends = []
 
-for v in _backend_to_sub_backends_dict.values():
-    _all_sub_backends.extend(v)
+def _get_all_sub_backends():
+    result = []
```
vedpatwardhan, line 50: probably we can name `result` to `sub_backends`

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -200,3 +210,76 @@ def find_available_sub_backends(sub_backends_loc):
...
+    for sub in sub_backends:
+        sub_backend = ivy.utils.dynamic_import.import_module(_sub_backend_dict[sub])
+        for k, v in sub_backend.__dict__.items():
+            if isinstance(v, Callable) and not k.startswith("__"):
+                result[k] = result.get(k, []) + [sub]
+
+    return result
+
+
+def available_sub_backend_implementations(obj: str) -> list:
```
vedpatwardhan, line 226: let's rename `obj` to `fn_name`

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -200,3 +210,76 @@ def find_available_sub_backends(sub_backends_loc):
...
+    --------
+    >>> import ivy
+    >>> ivy.set_backend('torch')
+    >>> ivy.available_sub_backend_implementations("scaled_dot_product_attention")
+    ['xformers']
+    >>> ivy.set_backend('numpy')
+    >>> ivy.available_sub_backend_implementations("scaled_dot_product_attention")
+    []
+    """
+    obj = _check_callable(obj)
```
vedpatwardhan, line 250: we could just do a `fn = ivy.__dict__[fn_name]` right?
In fact, I think we could simplify this function further so that we don't need `fn` either, we can just import the `sub_backend` dicts and see if `fn_name in sub_backend_dict` for each of the sub-backends rather than iterating over the sub-backend dict which we're doing in `_verify_sub_backends_implementations`.

vedpatwardhan (changes requested):
Just requested a few more changes, thanks @Madjid-CH 😄

ivy/utils/backend/sub_backend_handler.py
vedpatwardhan: I've responded in the new review to this, happy to know your thoughts!

ivy/utils/backend/sub_backend_handler.py
vedpatwardhan: I think we can just rename this to `fn_name` which should tell the user what we expect in that case!

ivy/utils/backend/handler.py
Madjid-CH: I noticed also in the signiture
image
that the second argument is `target` and in the usage:
image
we pass ivy as a second argument so it should be something like: _set the terget as backend_ so the name maybe is `_set_module_as(terget=ivy, backend=backend)` or we can settle on `_set_ivy_as_backend`

ivy/utils/backend/handler.py
Madjid-CH: yeah, right

ivy/utils/backend/sub_backend_handler.py
Madjid-CH: yeah, since we don't check for callable anymore this would be redundant

Madjid-CH: here I'm importing all sub_backends and see if `fn_name in sub_backend_dict` so in the end the `_available_sub_backends_implementations_dict` will be something like
 `{torch: {
        "scaled_dot_product_attention": ["xformers", ...],
        "another_fn": [...],
}`
that's why I'm iterating over all sub backends

vedpatwardhan: Sure, but you're also iterating over the `dict` in the `_find_available_sub_backend_implementations` function on this line, I'm referring to loops like `for k, v in sub_backend.__dict__.items():`

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -200,3 +209,67 @@ def find_available_sub_backends(sub_backends_loc):
...
+
+
+def _find_available_sub_backend_implementations(sub_backends):
+    result = dict()
+    for sub in sub_backends:
+        try:
+            sub_backend = ivy.utils.dynamic_import.import_module(_sub_backend_dict[sub])
+        except ModuleNotFoundError:
+            continue
+        for k, v in sub_backend.__dict__.items():
```
vedpatwardhan, line 221: I think this loop could be avoided with a single line instead, I might be missing something though

vedpatwardhan (changes requested):
Almost there, just a minor change, thanks @Madjid-CH 😄

ivy/utils/backend/sub_backend_handler.py
Madjid-CH: yeah here I'm pre searching all possible efficient implementations not just for `fn_name` but for all functions available in sub backends and put them in `_available_sub_backends_implementations_dict ` and I only do this  thorough search once for every backend so the next time we want to know if there is an efficient implimentation for a given backend we access  `_available_sub_backends_implementations_dict ` directly without executing `_find_available_sub_backend_implementations`

ivy/utils/backend/sub_backend_handler.py
Madjid-CH: yeah, here we search all possible sub backends for efficient implementations so we can Cache them in `_available_sub_backends_implementations_dict  `

ivy/utils/backend/handler.py
Madjid-CH: @vedpatwardhan, a gentle reminder 😄

vedpatwardhan: Thanks for the reminder, sorry missed out on this commit in the last review 😅
Yeah I think `_set_module_backend` would be the semantically preferable name because it indicates that we're setting the backend of the module while `_set_backend_as_ivy` indicates that we're setting `ivy` as a backend which isn't the case.

ivy/utils/backend/sub_backend_handler.py
vedpatwardhan: I think we should only search for one function at a time in the sub_backend dict particularly given that neither solution is sub-optimal in terms of time with the advantage of not having to maintain a cache dict and making debugging easier if a particular sub-backend implementation is retrieved when we try to wrap that particular function in the backend loop along with having a simpler implementation of this logic as well. What do you think?

vedpatwardhan (changes requested):
Hey @Madjid-CH, just a couple more comments, thanks 😄

ivy/utils/backend/sub_backend_handler.py
Madjid-CH: As I understand it: when we set the backend let's say to `torch `we have to check if there's an efficient implementation for every function and let's say that we have 10 sub-backends so if we used the cache we traverse these sub-backends once namely the first time the `_set_backend_as_ivy` tries to assign a function, for the second function we only check `cache["fn_name"] ` to give us all available sub-backends implementations for this function and so on for the rest of the functions.
but if we don't use a cache we have to loop on the 10 sub-backends for every function and I think this would become noticeable as we add more sub-backends.
But maybe there's not significant performance improvement, the only way to know for sure is to measure the two solutions, maybe I can find a way to mock the sub-backends and see the difference. What do you say ?

vedpatwardhan: Yeah I think both approaches are roughly equivalent because you'd still need to loop over all available functions when you're wrapping the first function anyway. It's just about whether we want to iterate over all functions available when we're trying to check for the first function or whether we want to check for sub-backend implementations when we iterate over those functions. Given that we'd be having a relatively small number of sub-backend functions (3 as of now) and sub-backends (2), I think both approaches would roughly be the same in terms of performance because people wouldn't really switch sub-backends that often.
I think for now we can skip the cache approach to simplify the implementation, but surely if we have a large number of sub-backends and sub-backend implementations, we could consider caching for faster retrieval 😄

ivy/utils/backend/handler.py
vedpatwardhan: thanks for refactoring this!

vedpatwardhan (changes requested):

ivy/utils/backend/sub_backend_handler.py
Madjid-CH: understood 😃

ivy/utils/backend/sub_backend_handler.py
```diff
@@ -200,3 +208,39 @@ def find_available_sub_backends(sub_backends_loc):
             available_sub_backends.append(sub_backend)
 
     return available_sub_backends
+
+
+def available_sub_backend_implementations_for(fn_name: str) -> list:
```
vedpatwardhan, line 213: I think the `_for` in the function name would be implicitly assumed if we're accepting `fn_name` and the function name is `available_sub_backend_implementations`

vedpatwardhan (changes requested):
Just a minor change, thanks @Madjid-CH 😄

ivy/utils/backend/sub_backend_handler.py
Madjid-CH: sure, I'll fix it

vedpatwardhan (approved):

PR #23502, opened 2023-09-13
Fix: Added `self` as the first parameter at places where it is missing

Sai-Suraj-27:
# PR Description
Added `self` as the first parameter at 2 places where it is missing.

## Checklist
- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials:


ivy/__init__.py
```diff
@@ -435,11 +435,11 @@ def assert_has_rank(self, rank):
         if self.rank not in (None, rank):
             raise ValueError("Shape %s must have rank %d" % (self, rank))
 
-    def unknown_shape(rank=None, **kwargs):
```
KareemMAX, line 438: I guess it is reasonable to have this function as static. As it produce `unknown_shape` from `rank`

ivy/__init__.py
```diff
@@ -463,7 +463,7 @@ def with_rank_at_most(self, rank):
         else:
             return self
 
-    def as_shape(shape):
```
KareemMAX, line 466: I guess it is reasonable to have this function as static.

KareemMAX (changes requested):
Hey @Sai-Suraj-27,
Thanks for the contribution, just made a few comments.

ivy/__init__.py
Sai-Suraj-27: Ok, sir. I will retrieve these 2 changes. 👍

KareemMAX (approved):

PR #23512, opened 2023-09-13
Ldl ivy

### LINKED ISSUE #23318: ldl_factor

khethan123:
# Initial PR on LDL_factorization

"After a contributor creates a new issue in Ivy, they have 7 days to create a pull request (PR)", this was given in the docs so i created this pull request. But the **main** reason was to get help. This function would be complete if `scipy.linalg.ldl( )` would return pivots as well. I am stuck as i do not have enough experience to implement it right now.

### Currently this `ldl_factor` calls scipy's `ldl` function and returns a compact `LD` matrix similar to troch's `ldl_factor`, but this implementation also has a problem since the output i.e., LD generated by this method is not same as the one generated by torch when the matrix we use is a `non-positive definite` matrix  of order (4 x 4) and above or when all pivots have same value.

Scipy's implementation which was used here gives only LD as o/p
image
☝️ look at the last 2 rows of the first column.

Torch's implementation when used for the same i/p matrix
image
☝️ look at the last 2 rows of the first column, they are not same.
So , if someone would show were to discuss this topic, it would be wonderful!!!!!.

Also i have not added test function yet because their was no `ldl_factor` code present in `ivy/functional/backends/torch/linear_algebra.py`. 

Close #23318 

## Checklist

- [x] Did you add a function?
- [ ] Did you add the tests?


MahmoudAshraf97 (changes requested):
Hello, you should strictly use ivy functions in the frontends, so you'll have to reimplement your code without using SciPy or any other framework

PR #23521, opened 2023-09-13
Version fix in frontends

umairjavaid:
This allows frontends to get version of the selected backend by doing `frontend.__version__`. If the version is not available it'll mention that.

### DIFF
--- a/ivy/functional/frontends/jax/__init__.py
+++ b/ivy/functional/frontends/jax/__init__.py
@@ -15,7 +15,7 @@ from . import random
 from . import _src
 from ._src import tree_util
 
-__version__ = ivy.__version__
+__version__ = ivy.backend_version.get("version", "version not available!")
 
 _frontend_array = numpy.array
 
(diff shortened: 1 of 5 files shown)


ivy/functional/frontends/jax/__init__.py
```diff
@@ -15,6 +15,7 @@
 from . import _src
 from ._src import tree_util
 
+__version__ = ivy.backend_version.get("version", "version not available!")
```
AnnaTz, line 19: The version of the frontend is not necessarily same as the corresponding backend (I think that's what you are doing here). For example, we could use an old pytorch version as frontend and transpile it to a newer pytorch version, in which case frontend and backend would be torch but their versions would differ.
I think we need to get the frontend version from here https://github.com/unifyai/ivy/blob/b54f96889e45c45f22eee75dae79af8b518faa32/ivy/functional/frontends/__init__.py#L88

AnnaTz (changes requested):

ivy/functional/frontends/jax/__init__.py
umairjavaid: No, I don't think so. We need to get the backend version here because all the computations being performed rely on this compatibility.

Issue #15047, opened 2023-05-08
Add `paddle.fft` Fast Fourier Transform Functions to PaddlePaddle Frontend


hmahmood24:
Add [Fast Fourier Transform Functions](https://www.paddlepaddle.org.cn/documentation/docs/en/api/index_en.html) to PaddlePaddle Frontend

> Please keep in mind that the proper way to link an issue to this list is to comment "- [ ] #issue_number" while the issue's title only includes the name of the function you've chosen.

- [ ] fft
- [ ] ifft
- [ ] rfft
- [ ] irfft
- [ ] hfft
- [ ] ihfft
- [ ] fft2
- [ ] ifft2
- [ ] rfft2
- [ ] irfft2
- [ ] hfft2
- [ ] ihfft2
- [ ] fftn
- [ ] ifftn
- [ ] rfftn
- [ ] irfftn
- [ ] hfftn
- [ ] ihfftn
- [ ] fftfreq
- [ ] rfftfreq
- [ ] fftshift
- [ ] ifftshift

KevinUli:
- [ ] #16684

MuhammadNizamani:
- [ ] #17855

WOLFIEEEE:
I want to work on it , is it still open If Yes can you Please assign me this.

Nav21000:
I want to work on it, please assign to me

akgokce:
- [ ] #18285

dilne:
- [ ] #18960

ArsalanAli915:
- [ ] #19052

AJBroad94:
#19379

AwkNinja:
- [ ] #19383

Christimage:
- [ ] #19426

Dhruv-Varshney-developer:
-irfft #19466

Zamokuhle23:
#19814

BrolyDBZ:
#20817

JamesC1305:
- [ ] #21717

catplotlib:
#21817

Seif-Mohamed1:
- [ ] #21863

catplotlib:
#22002

Seif-Mohamed1:
#21863

Aliutman:
#22169

Dhruv-Varshney-developer:
- [ ] #22609

Aliutman:
- [ifftn ] #22651

[...]

PR #23526, opened 2023-09-13
Feat implement inverse fast fourier transform (irfftn) function for paddle frontend

### LINKED ISSUE #22913: irfftn

xingshuodong:
# PR Description 
feat: implement inverse fast Fourier transform (irfftn) function for paddle frontend

Add paddle.fft Fast Fourier Transform Functions to Paddle Frontend #15047 
This pull request is related to resolves [irfftn #22913].
- [x] <irfftn> #22913 
Close #22913 
## Checklist 

- [x] Did you add a function?
- [x] Did you add the tests?

### Socials: 
xingshu@deakin.edu.au
ins: xingshuodong


xingshuodong:
@tomatillos please reply me at this PR, i think the old PR is going to be closed.
As far as the irfftn is working for n-dim case, please let me know if I need to do some improvement.

ivy_tests/test_ivy/test_frontends/test_paddle/test_fft.py
```diff
@@ -242,3 +242,39 @@ def test_paddle_rfftfreq(
...
+    input_dtypes, x, axis = dtype_x_axis
+    helpers.test_frontend_function(
+        input_dtypes=input_dtypes,
+        backend_to_test=backend_fw,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        x=x[0],
+        s=None,
+        axes=[-1],
```
tomatillos, line 278: As I mentioned on the other PR, need to test axes=axis rather than just -1 here.

tomatillos:
Also fix the lint and PR compliance checks please.

ivy/functional/frontends/paddle/fft.py
mobley-trent, line 1: Remember to use pre-commit so that it performs lint checks on your code. As well as the tests

mobley-trent (changes requested):

mobley-trent:
@xingshuodong this is a great PR ⚡ 
Just resolve the minor issues and this should be good to merge. 
Is there a valid reason why you've set `s` to None in the tests ? There might be a bug in your code you aren't aware about

mobley-trent:
As well as `axes`

ivy_tests/test_ivy/test_frontends/test_paddle/test_fft.py
```diff
@@ -242,3 +242,39 @@ def test_paddle_rfftfreq(
...
+        available_dtypes=helpers.get_dtypes("complex"),
+        min_value=-10,
+        max_value=10,
+        min_num_dims=1,
+        max_num_dims=4,
+        min_dim_size=2,
+        valid_axis=True,
+        force_int_axis=True,
+    ),
+    norm=st.sampled_from(["backward"]),
```
mobley-trent, line 259:
```suggestion
    norm=st.sampled_from(["backward", "ortho", "forward"]),
```

ivy_tests/test_ivy/test_frontends/test_paddle/test_fft.py
xingshuodong: hello i have test, the axes can be 0 or 1 or -1 just in the range of the dim of axes, but the frontend is always give 0 in case,
and frontend doesn't support a list of axes to be input, so here it doesn't matter its value for test purpose

ivy_tests/test_ivy/test_frontends/test_paddle/test_fft.py
```diff
@@ -242,3 +242,39 @@ def test_paddle_rfftfreq(
...
+):
+    input_dtypes, x, axis = dtype_x_axis
+    helpers.test_frontend_function(
+        input_dtypes=input_dtypes,
+        backend_to_test=backend_fw,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        x=x[0],
+        s=None,
```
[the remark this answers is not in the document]
xingshuodong: s=none, otherwise the input people should determine the shape of output exactly macthing the result which is an extra step, if input is wrong, then the result is wrong.

a nomarl step taken by tensorflow paddle and numpy is to calculate the shape of x to determine the normal axes and last axis, they similar use this stragety to handle when the s is None, so s is calcualted from the function.

xingshuodong:
> Also fix the lint and PR compliance checks please.

@Hi Mark, i cant pass the lint check becasue the file itself can't pass the lint check specific to rfftfreq,
should i modify it?

ivy/functional/frontends/paddle/fft.py
```diff
@@ -131,3 +131,48 @@ def rfftfreq(n, d=1.0, dtype=None, name=None):
...
+    # Determine the output data type based on the input data type
+    if ivy.is_complex_dtype(x.dtype):
+        output_dtype = "float32" if x.dtype == "complex64" else "float64"
+    else:
+        output_dtype = "float32"
+        
+    target_axes = axes
+    
+    if s is None:
+        s = [x.shape[axis] if axis != axes[-1] else 2 * (x.shape[axis] - 1) for axis in axes]
```
mobley-trent, line 153: I tried running your function against the native version in the paddle docs:
```
import paddle
from ivy.functional.frontends import paddle as ivy_paddle

x = paddle.to_tensor([2.+2.j, 2.+2.j, 3.+3.j]).astype(paddle.complex128)

irfftn_x = paddle.fft.irfftn(x)
print(irfftn_x)

irfftn_x = ivy_paddle.irfftn(x)
print(irfftn_x)
```

Your implementation returns an error that stems from this line

mobley-trent (changes requested):

ivy/functional/frontends/paddle/fft.py
xingshuodong: hi @mobley-trent, i think i need to add  this to handle when axes is not given

if axes is None:
        axes = list(range(len(x.shape)))

xingshuodong: I have update the function to handle mutiple axes case, the key is to isolately reverse each axis in the case, and return them at final step.

xingshuodong: @mobley-trent can we just call backend function from paddle, or other backend in frontend instead of using ivy libaray to develop the irfftn?

xingshuodong: @mobley-trent Hi Eddy, in my old version the bug is it doesn't handle the none axes case, and for mutiple axes, it does not update the local x, which is relative to freq domain if when it goes to mutiple axes.

In this updated irfftn, it will handle the none axes case, it either evidences that the irfftn is working for many axes case,
so rather than test natively which cost extra efforts, i think now it should be okay unless we want to irfftn has ability to cut off or extend the shape of output freq domain which is discussed earlier why s is set to none

ivy/functional/frontends/paddle/fft.py
xingshuodong: yes, i am learning how to use pre-commit

ivy/functional/frontends/paddle/fft.py
mobley-trent: > @mobley-trent can we just call backend function from paddle, or other backend in frontend instead of using ivy libaray to develop the irfftn?

Most frontends are just a simple call to a similar ivy function, or a composition of other ivy functions. This serves as an abstraction to the backends. You are not supposed to implement a frontend function with a native function.

xingshuodong:
@mobley-trent Hi I can not identify what is casuing the issue that after merging the latest branch, the test used to pass the test now cant work,

ivy/functional/frontends/paddle/fft.py
xingshuodong: yea but currently there is no irfftn in ivy backend to handle the case, it is more likely to use ifftn to simulate the irfftn, but for initial purpose the irfftn support for n dim, with just one axis or many specific axes, do you think this PR needs to be further developed?

xingshuodong: so @mobley-trent Eddy which one do you perfer
1. add a irfftn backend function in ivy\functional\ivy\experimental\layers.py, then this serve the purpose of abstract at frontend
2. keep things does not change like now

xingshuodong: @mobley-trent 
image
if test natively, please differectly define function, then when generete a tensor, please directly copy the array into irfftn
becasue the frontend of ivy will return an x tensor, we pass the x[0] which is the array to our function and backend

x = paddle.to_tensor([[2.+2.j, 2.+2.j, 3.+3.j], [2.+2.j, 2.+2.j, 3.+3.j]]).astype(paddle.complex128)
- ivy_test_result=irfftn([[2.+2.j, 2.+2.j, 3.+3.j], [2.+2.j, 2.+2.j, 3.+3.j]])
- paddle_test_result=paddle.fft.irfftn(x)

xingshuodong:
@mobley-trent Hi Eddy I have carefully review the irfftn explnation, and rewrite the function.
1. for nomral axes, irfftn use ifftn
2. for last axis, irfftn use the developed calculation represents irfft
Now this is a ture irfftn function for any dims any size, it enable:
1. input axes like [0, 2, 3, 4] or [int] or None
2. input shape as None, or a customerized shape, the ifftn function used in irfftn is supporting this function

xingshuodong:
image
In Tensorflow, in some cases the assert_all_close function is resulting issue of result doesn't match, this is an backend issue, but may occur during irfftn.

ivy_tests/test_ivy/test_frontends/test_paddle/test_fft.py
```diff
@@ -222,6 +222,43 @@ def test_paddle_irfft(
...
+@handle_frontend_test(
+    fn_tree="paddle.fft.irfftn",
+    dtype_x_axis=helpers.dtype_values_axis(
+        available_dtypes=helpers.get_dtypes("complex"),
+        min_value=-10,
+        max_value=10,
+        min_num_dims=3,
+        max_num_dims=3,
+        min_dim_size=3,
+        max_dim_size=3,
```
mobley-trent, line 234: Why are these dim sizes fixed ?

mobley-trent (changes requested):

ivy_tests/test_ivy/test_frontends/test_paddle/test_fft.py
mobley-trent: Essentially we test for as many input combinations as possible

xingshuodong: it can be as many as input, but in the case, the backend is not developed, and ifftn for backend in tensorflow, jax, and possible some of backends are only support up to 3-D case in N-D use, that's not the frontend issue.
（here maybe the solution is upto the backend ifftn, if go more than 3-D, directly use irfftn，or repeat the ifftn util finish the whole array）
Of course it can be as many as possible if you change the test setting, but it may occur the case I mentioned early in tensorflow (you can check the comparsion of the result), but in general this function is how irfftn works.

Take an example of setting the dim from 5 to 7, in runnning test then you can check that it says current backend of JAX, tensorflow ifftn only support up to 3D.

So please directly point out what irfftn suppose to be, you consider there is no irfftn backend function, and the limitation of backend, in tensorflow, jax.
@mobley-trent  Please suggest ASAP, the test function is just one shot of test, it does not mean it can't pass anyother test.

xingshuodong: You can notice the return datatype enrror in warning, this is definitly what backend should do before calling the function, becasue there is no such backend support, we have to decide what backend it is and convert the input datatype. 
I can only say the frontend function is doing its job, the frontend function is going straight forward.

xingshuodong: image

xingshuodong: If there is no any other concern, please approve the merge, appriciate.

xingshuodong: @mobley-trent

xingshuodong:
> Hey @xingshuodong your function should at least pass for the minimal example I provided earlier from the [paddle docs](https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/fft/irfftn_en.html)

it does pass, you can check the picture, what i mean is the test method fixed doesn't mean it works in limited case, it depends on how the backend developed in other backend, just like in tensorflow there is no nature irfftn function, so it is better to have irfftn special for tensorflow itself, or just rasie the error.

Please check this picture, it works for most cases, not only limit one. @mobley-trent 
image

xingshuodong:
Please pay attention to check, it is tested. @tomatillos

mobley-trent:
Only the paddle backend seems to be running in this case. We can just merge for now
Thanks again for your PR @xingshuodong 🚀

xingshuodong:
The case here is this irfftn is working correctly in numpy backend and paddle backend.
Jax and tensorflow backend with over 3 - D dims, need some specific consideration.

PR #23570, opened 2023-09-13
feat: Update set.py

### LINKED ISSUE #23539: unique_inverse

adlub:
Updated set.py, in particular the unique_inverse function

Close #23539


KareemMAX (changes requested):
Hi, seems like you have changed `docs/demos` which you might have done by mistake. Can you revert that change because we can't merge with that change in place?

adlub:
> Hi, seems like you have changed `docs/demos` which you might have done by mistake. Can you revert that change because we can't merge with that change in place?

@KareemMAX I seem to be having trouble using "git checkout/revert" to revert these changes, I am trying to do it by using a codespace but it does not seem to work. Could i also create a new commit that includes the up to date version of those files that are on the repository (not inlcluding my changes)?

KareemMAX:
> > Hi, seems like you have changed `docs/demos` which you might have done by mistake. Can you revert that change because we can't merge with that change in place?
> 
> @KareemMAX I seem to be having trouble using "git checkout/revert" to revert these changes, I am trying to do it by using a codespace but it does not seem to work. Could i also create a new commit that includes the up to date version of those files that are on the repository (not including my changes)?

I guess that should work, also try merging

adlub:
> > > Hi, seems like you have changed `docs/demos` which you might have done by mistake. Can you revert that change because we can't merge with that change in place?
> > 
> > 
> > @KareemMAX I seem to be having trouble using "git checkout/revert" to revert these changes, I am trying to do it by using a codespace but it does not seem to work. Could i also create a new commit that includes the up to date version of those files that are on the repository (not including my changes)?
> 
> I guess that should work, also try merging

I ended up being able to remove the "docs" changes, is my work ready to merge now? Also, during this process, i accidentally requested 3 other reviewers. @RickSanchezStoic, @sherry30, @CatB1t, sorry about that, are you able to remove yourselves from this?

zaeemansari70 (changes requested):
Hi, you have to handle this in the respective test_function in the test file aswell and add the functionality there in the hypothesis too.
Thanks!

adlub:
> Hi, you have to handle this in the respective `test_function` in the `test file` aswell and add the functionality there in the hypothesis too. Thanks!

Hi @zaeemansari70, I have applied these changes now, thanks

PR #23576, opened 2023-09-14
fix: paddle.median frontend function

Mghrabi:
### PR Description
By Removing `float16` and `bool` all tests pass, the reason I chose to remove `float16` and `bool` even though they are mentioned in the docs, is that paddle actually errors out that they are not supported. And even if we assumed they are supported and the issue is from the testing pipeline, this won't affect the process of testing paddle.median as it always returns `float32` for all given dtypes except for `float64` (and the dtype conversion is done prior function implementation).

### DIFF
--- a/ivy/functional/frontends/paddle/stat.py
+++ b/ivy/functional/frontends/paddle/stat.py
@@ -14,7 +14,7 @@ def mean(input, axis=None, keepdim=False, out=None):
 
 
 @with_supported_dtypes(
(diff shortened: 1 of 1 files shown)


ivy-leaves:

Thank you :hugs:

mohame54:
Hi @Mghrabi , Thanks for your contribution, your PR looks good to me but could you solve the semantic errors .

Mghrabi:
@mohame54 Done

mohame54:
looks good to me.

PR #23582, opened 2023-09-14
fix: Updated set.py

jaskiratsingh2000:
#23539

### DIFF
--- a/ivy/functional/ivy/set.py
+++ b/ivy/functional/ivy/set.py
@@ -158,6 +158,9 @@ def unique_all(
 def unique_inverse(
     x: Union[ivy.Array, ivy.NativeArray],
     /,
+    *,
+    axis: Optional[int] = None,
+    by_value: bool = True,
 ) -> Tuple[Union[ivy.Array, ivy.NativeArray], Union[ivy.Array, ivy.NativeArray]]:
     """
     Return the unique elements of an input array ``x``, and the indices from the set of


jaskiratsingh2000:
@Nightcrab @Daniel4078 
`axis`
It specifies the dimension of x along which the unique elements will be computed.
Explanation:

- If axis is None (the default), the unique elements of the flattened array x are returned. In other words, the function will consider x as a 1D array regardless of its actual shape.
- If axis is an integer, then the unique elements are computed along the specified dimension. This means if x is a 2D array (like a matrix) and axis=0, then the function will find unique rows, and if axis=1, it will find unique columns.

`by_value`
Determines the order in which unique elements are returned.

Explanation:

- If by_value is True (the default), the unique elements are sorted by their value.
- If by_value is False, the unique elements will be sorted in the same order that they occur in x. This means the order of appearance in the original x array is preserved for the unique elements in the output.

adlub:
Hi @jaskiratsingh2000, I submitted a pull request for this issue about 9 hours ago and our commits seem to be identical. Are you able to see this? Thanks

jaskiratsingh2000:
Some of its tests were failing and identically I figured out this can be done towards formatting

Ookamice:
@jaskiratsingh2000 
I believe Nightcrab's point is that `axis` and `by_value` is currently not being used in the function. How the ivy API works is that it calls the respective backend implementation of the ivy function with `ivy.current_backend()..unique_inverse(x)` so even if you have updated  `axis` and `by_value` in the function declaration, it will have no effect unless you've also altered every backend.

Also @adlub makes a great point here. I see that the issue you have linked was opened by him originally. Please could you find another open issue to do as I wil lbe closing this.

adlub:
Thanks @Ookamice. @jaskiratsingh2000, I suggest we move to my pull request page and we can discuss your suggested changes.

PR #23601, opened 2023-09-14
feat: added .diff torch frontend instance method and test

### LINKED ISSUE #23418: diff
adding .diff method to torch frontend

GrigoryMaslov:
# PR Description

A .diff (the n-th forward difference along the given dimension) method added to pyTorch frontend. 
Works well but doesnt't pass all tests: 
1) fails with Paddle backend, because the backend function seems to return a 2-dimensional array while 1-dimensional is required. 
2) fails sometimes with some backedns (JAX, Paddle, rarely Tensorflow) when working with very big or very small complex numbers (dtype complex128), while always passes the tests on numpy and torch backends. This error does not occur if max and min value limits are set in tests, i.e min_value=-1e09, max_value=1e09, as in the last fix.

Would be great to receive a hint or advice on these two! :)

## Related Issue

Close #23418

## Checklist

- [x] Did you add a function?
- [x] Did you add the tests?


ivy/functional/frontends/torch/tensor.py
```diff
@@ -1698,6 +1698,25 @@ def floor_(self):
...
+        {
+            "2.0.1 and below": (
+                "int8",
+                "int16",
+                "uint8",
+                "uint16",
+                "uint32",
+                "uint64",
+                "bfloat16",
+                "float64",
```
Sam-Armstrong, line 1712:
```suggestion
                "bfloat16",
                "complex",
                "float64",
                "int8",
                "int64",
```
Updated the unsupported dtypes

ivy/functional/frontends/torch/tensor.py
```diff
@@ -1698,6 +1698,25 @@ def floor_(self):
         self.ivy_array = self.floor().ivy_array
         return self
 
+    @numpy_to_torch_style_args
```
Sam-Armstrong, line 1701:
```suggestion
```
I don't think we need this.

Sam-Armstrong (changes requested):
Hey @GrigoryMaslov, apologies for taking a while to get to this. I've left a couple of suggested changes, but once you've done these we should be good to merge. Great work! 😁

Sam-Armstrong (approved):
Looks like all the tests are passing, except for paddle, which is probably an issue with the paddle backend. So I think we're good to merge. Thanks for the contribution @GrigoryMaslov!

PR #23603, opened 2023-09-14
Xformers multiversion

RickSanchezStoic:
features: adds multi-version support for sub-backends and added torch multiversion implementation for scaled_dot_product_attention


ivy/functional/backends/torch/layers.py
```diff
@@ -544,3 +544,18 @@ def conv_general_transpose(
     if data_format == "channel_last":
         res = res.permute(0, *range(2, dims + 2), 1)
     return res
+
+
+def scaled_dot_product_attention_v_2p0p1_and_above(
```
vedpatwardhan, line 549: I think `torch.nn.functional.scaled_dot_product_attention` was added in torch version `2.0.0`, we could update the interval here.

Also, could you please rename the `xformers` sub-backend implementation to the version-specific name to be tested for versions before `torch` version `2.0.0`?

vedpatwardhan (changes requested):
Just requested a couple of changes, thanks @RickSanchezStoic 😄

ivy/functional/backends/torch/layers.py
vedpatwardhan: Yep you're right, I was just asking if the changes being made to the sub-backend handler allow that flexibility. 

Secondly, given that the `torch.nn.functional.scaled_dot_product_attention` was added in torch version `2.0.0` instead of `2.0.1`, the version interval in function name should be `2p0p0_and_above`

vedpatwardhan (changes requested):
Just requested another change, thanks @RickSanchezStoic 😄

ivy/functional/backends/torch/layers.py
RickSanchezStoic: Done! Yeah, I tested it out by writing a version specific one for the Xformers it worked!

vedpatwardhan (approved):

PR #23612, opened 2023-09-14
feat(frontends): Creates a `__version__` variable when setting a frontend

AnnaTz:


umairjavaid:
looks good!

umairjavaid (approved):
looks good!

umairjavaid:
Since we're returning the frontend version, we can have this in all the frontends, even for ones whose backend isn't implemented?

AnnaTz:
> looks good!

Did you try it out in practice? I don't have any handy way to test this.

AnnaTz:
> Since we're returning the frontend version, we can have this in all the frontends, even for ones whose backend isn't implemented?

Yes, but I'm not sure we have multi-version support for those new frontends yet.

PR #23650, opened 2023-09-15
Add Dockerfile for Apple Silicon Macs

NripeshN:


vedpatwardhan (changes requested):
Just tried building an image and then a container locally, there seem to be a few issues. 
1. The default `python` version is `2.7`, `python3` is `3.7` but our main focus has been on `3.8` and `3.10`.
2. Secondly, I was able to import ivy in a `3.7` console, but not sure if it's installed correctly because on doing `dir(ivy)`, it returns `['__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__']` meaning none of the functions are accessible.
3. The folder structure in the container contains the `requirement_mappings_apple_silicon.json` file alongside `ivy` in the root folder of the project, we should be cleaning that up at the end of building.

Happy to know what you and the others think, thanks @NripeshN 😄

NripeshN:
> Just tried building an image and then a container locally, there seem to be a few issues.
> 
> 1. The default `python` version is `2.7`, `python3` is `3.7` but our main focus has been on `3.8` and `3.10`.
> 2. Secondly, I was able to import ivy in a `3.7` console, but not sure if it's installed correctly because on doing `dir(ivy)`, it returns `['__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__']` meaning none of the functions are accessible.
> 3. The folder structure in the container contains the `requirement_mappings_apple_silicon.json` file alongside `ivy` in the root folder of the project, we should be cleaning that up at the end of building.
> 
> Happy to know what you and the others think, thanks @NripeshN 😄

Hi @vedpatwardhan, I have fixed all the above mentioned issues

vedpatwardhan (changes requested):
The issue mentioned in the 2nd point of the previous review still seems to be there, just tried building again, the other 2 are resolved. Thanks @NripeshN 😄

RickSanchezStoic:
> The issue mentioned in the 2nd point of the previous review still seems to be there, just tried building again, the other 2 are resolved. Thanks @NripeshN 😄

Just curious are you having issues with the ivy locally installed or the one in the docker image? Because the ivy that is locally on your system should be picked at all times rather than the one in the image. I can't comment further on the Apple image since I don't have an Apple or arm64 device to test this out.

docker/Dockerfile
```diff
@@ -81,3 +78,9 @@ RUN python3 test_dependencies.py -fp requirements.txt,optional.txt && \
     rm -rf optional.txt && \
     rm -rf tmp.txt && \
     rm -rf test_dependencies.py
+
```
RickSanchezStoic, line 81: I don't think we need this, codespaces handle this itself.

docker/DockerfileGPU
```diff
@@ -127,3 +127,9 @@ RUN python3 test_dependencies.py -fp requirements.txt,optional_gpu.txt && \
     rm -rf requirements.txt && \
     rm -rf optional_gpu.txt && \
     rm -rf test_dependencies.py
+
```
RickSanchezStoic, line 130: Same as cpu Dockerfile

RickSanchezStoic (changes requested):
I can't comment on the Apple image, but yeah some changes in the other docker files don't seem to be needed.

docker/DockerfileGPUMultiCuda
```diff
@@ -76,12 +76,16 @@ RUN sed -i '/numpy/d' requirements.txt &&\
...
-
-
-
 COPY run_tests_CLI/test_dependencies.py .
 RUN python3 test_dependencies.py -fp requirements.txt,optional_gpu.txt && \
     rm -rf requirements.txt && \
     rm -rf tmp.txt && \
     rm -rf optional_gpu.txt && \
     rm -rf test_dependencies.py
+
```
RickSanchezStoic, line 85: same changes to be made here too.

docker/DockerfileMultiversion
```diff
@@ -53,7 +53,9 @@ RUN python3 multiversion_framework_directory.py  $fw && \
 
 ENV PATH=/opt/miniconda/envs/multienv/bin:$PATH
 
-
-
-
+# Post installation steps
```
RickSanchezStoic, line 56: same changes to be made here too.

RickSanchezStoic (changes requested):
just some more changes

docker/DockerfileAppleSilicon
```diff
@@ -0,0 +1,242 @@
...
+    rm -rf requirements.txt && \
+    rm -rf optional_apple_silicon_1.txt && \
+    rm -rf optional_apple_silicon_2.txt && \
+    rm -rf tmp.txt && \
+    rm -rf test_dependencies.py && \
+    rm -rf requirement_mappings_apple_silicon.json
+
+# Post installation steps
+COPY .devcontainer/post_create_commands.sh .
+RUN chmod +x post_create_commands.sh && \
```
vedpatwardhan, line 240: I think we should've also updated the other dockerfiles to make use of the `post_create_commands` script and decoupled it from the `.devcontainer` to also run those commands when the user isn't necessarily using a vscode devcontainer as we discussed @NripeshN 😅

PR #23724, opened 2023-09-16
feat: reformatted set_nest_at_indices

### LINKED ISSUE #23723: set_nest_at_indices

Muftawu:
# PR Description
Reformatted function "set_nest_at_indices"

## Related Issue

Close #23723

## Checklist

✅add_reformatting_checklist_nest

### Socials:
https://twitter.com/My_myfpro


ivy/functional/ivy/nest.py
```diff
@@ -458,8 +458,8 @@ def prune_nest_at_indices(nest: Iterable, indices: Tuple, /) -> None:
 @handle_exceptions
 def set_nest_at_indices(
     nest: Union[List, Tuple, Dict, ivy.Array, ivy.NativeArray],
-    indices: Union[List[int], Tuple[int], Iterable[int]],
-    values: Union[List[int], Tuple[int], Iterable[int]],
+    indices: Iterable[Iterable[int]],
```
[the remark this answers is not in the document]
lucasalavapena: Further down there is 
```python
    if not isinstance(values, (list, tuple)):
        values = [values] * len(indices)
```

I think you could maybe get rid of it given your new definition of values? Something that is iterable should be zipable I think. Also could you add an exception that if the len of values and indices do not match, I assume those have to match

lucasalavapena (commented):
Thanks for your PR :smile: , @Muftawu! I made some comments. Also in case you were not aware it seems like this task was already done (see #1954) but was not properly ticked. Nonetheless, that task can still be improved :smile: .

ivy/functional/ivy/nest.py
```diff
@@ -189,14 +189,14 @@ def set_nest_at_index(
         b: ivy.array([3., 4.])
     }
     """
-    is_tuple = isinstance(nest, tuple)
-    nest_type = type(nest) if is_tuple else lambda x: x
+    is_iterable = isinstance(nest, tuple)
```
lucasalavapena, line 192: Why did you update this, if nest' type definition is still a union of a bunch of different types, not iterable, and here it is being specifically checking if it is a tuple?

lucasalavapena: Same goes for the related changes (with the name change)

lucasalavapena (changes requested):
Thanks for the updates! Regarding the assertion I suggested, they seem to be failing tests as they have seem to be some None inputs in the test (See https://github.com/unifyai/ivy/actions/runs/6250286752/job/16980501210?pr=23724 -> Combined Test Result and then search for set_nest or something to that effect).

```
________ test_reptile_step[cpu-paddle-False-False-first-1-True-False-2] ________
ivy/utils/exceptions.py:357: in _handle_exceptions
    return fn(*args, **kwargs)
ivy/functional/ivy/nest.py:533: in set_nest_at_indices
    assert len(indices) == len(values), " 'indices' and 'values' must be of the same length"
E   TypeError: object of type 'NoneType' has no len()
```

Could it be that the old type hints are wrong? Could you please have a look at the tests and figure it out. Also I just realized if we want to be very precise with type hints, is it not guaranteed that iterable have `len`, it could be that `Sequence` ([docs](https://docs.python.org/3/library/stdtypes.html#sequence-types-list-tuple-range)) would be more correct for us in this case, could you check if that is an acceptable change (by checking usage).

lucasalavapena:
I assume you are still working on it since the tests are still failing, please let me know when you are done! Also dont forget to check that you pass the linting check!

Muftawu:
Yes. I am resolving the errors. I will let you know when I'm done.

Muftawu:
Hello lucasalavapena, I have resolved the errors to the best of my abilities.
The remaining errors say "check on failures" and I have no idea where to begin or how 
to resolved them. I would be glad if you could guide me through those ones.

lucasalavapena:
> Hello lucasalavapena, I have resolved the errors to the best of my abilities. The remaining errors say "check on failures" and I have no idea where to begin or how to resolved them. I would be glad if you could guide me through those ones.

Thanks for trying to resolve it. I would suggest running the relevant tests locally or on codespaces (which makes the setup super easy), specifically the tests you changed, I think there are multiple functions you changed but the main one you should test is `set_nest_at_indices`, which you can do as follows `@lucasalavapena ➜ /workspaces/ivy (reformat_nest) $ pytest ivy_tests/test_ivy/test_functional/test_core/test_nest.py::test_set_nest_at_indices`. From there you can do some debugging if the errors are not so clear, there is also the `--pdb` argument for pytest.

Just FYI this is what you get when you run the tests:

```
ivy_tests/test_ivy/test_functional/test_core/test_nest.py:442: AssertionError
==================================================================================== 60.0% of 20 passed ==================================================================================================================================================================== short test summary info =================================================================================
FAILED ivy_tests/test_ivy/test_functional/test_core/test_nest.py::test_set_nest_at_indices[cpu-jax-False-False-False-values0-indices0-nest0] - AssertionError: assert {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}} != {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}}
FAILED ivy_tests/test_ivy/test_functional/test_core/test_nest.py::test_set_nest_at_indices[cpu-jax-False-False-False-values0-indices1-nest0] - AssertionError: assert {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}} != {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}}
FAILED ivy_tests/test_ivy/test_functional/test_core/test_nest.py::test_set_nest_at_indices[cpu-tensorflow-False-False-False-values0-indices0-nest0] - AssertionError: assert {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}} != {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}}
FAILED ivy_tests/test_ivy/test_functional/test_core/test_nest.py::test_set_nest_at_indices[cpu-tensorflow-False-False-False-values0-indices1-nest0] - AssertionError: assert {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}} != {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}}
FAILED ivy_tests/test_ivy/test_functional/test_core/test_nest.py::test_set_nest_at_indices[cpu-torch-False-False-False-values0-indices0-nest0] - AssertionError: assert {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}} != {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}}
FAILED ivy_tests/test_ivy/test_functional/test_core/test_nest.py::test_set_nest_at_indices[cpu-torch-False-False-False-values0-indices1-nest0] - AssertionError: assert {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}} != {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}}
FAILED ivy_tests/test_ivy/test_functional/test_core/test_nest.py::test_set_nest_at_indices[cpu-paddle-False-False-False-values0-indices0-nest0] - AssertionError: assert {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}} != {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}}
FAILED ivy_tests/test_ivy/test_functional/test_core/test_nest.py::test_set_nest_at_indices[cpu-paddle-False-False-False-values0-indices1-nest0] - AssertionError: assert {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}} != {'a': [[1], [2]], 'b': {'c': [1, [2, [8]]]}}
============================================================================== 8 failed, 12 passed in 0.30s ===============================================================================

```

ivy/functional/ivy/nest.py
lucasalavapena: You still need to address these changes. Why did you change the name of the variable if what you are still actually doing is checking if it is a tuple

run_tests_CLI/.ipynb_checkpoints/run_tests-checkpoint.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash -e
+docker run --rm -it -v "$(pwd)":/ivy unifyai/ivy:latest python3 -m pytest ivy_tests/
```
lucasalavapena, line 2: Why was this file added?

lucasalavapena (changes requested):

Muftawu:
hi @lucasalavapena. I was able to run the tests successfully and this is what i got. 
But upon pushing to github, the bot still throws an error. Please any clarification.

image

lucasalavapena:
So looking at it, it is an assertion that they are not equal, the reason being it is checking the shallow argument of the function and per their docstring it says: 

```
    shallow
        Whether to inplace update the input nest or not
        Only works if nest is a mutable type. Default is ``True``.
```

I think the tests are wrong, well at least if you compare it to the docs. The problem is that when shallow is false, it is supposed to create a new dictionary/object and update that one rather than the exisiting object, but the tests are checking that the passed in argument has changed from its initial state via: `assert nest != nest_copy ` which seems wrong to me based the definition and also an equality check is not enough to check if the objected was overwritten, for example an identity transformation could have just occured.  I think in general the shallow logic in the tests is not 100% correct.

Muftawu:
Yes. I just saw that. So what do I do. Should I wait for further instructions?

lucasalavapena:
> Yes. I just saw that. So what do I do. Should I wait for further instructions?

You can fix the tests if you are up for, otherwise I can try to do it by I will make a separate PR for it

Ishticode:
Closing this due to lack of activity but also due to the fact that the issue linked seems to be for reformatting but the changes seem to change code in a non-meaningful way. Plus some of the requests above were left unadressed.

PR #23738, opened 2023-09-17
refactor: Replaced `lambda` expressions with `def` functions.

### LINKED ISSUE #23734: Replace `lambda` expression, use a `def`
## Issue Description
I see that at many places in the codebase, a `lambda` expression is being used instead of a `def`.  I found all the places where this is happening using the [`ruff`](https://github.com/astral-sh/ruff) tool. I think all the `lambda` expressions should be replaced with `def`.
For example in this file: `ivy_tests\test_ivy\test_functional\test_core\test_gradients.py`

Sai-Suraj-27:
# PR Description
I see that at many places in the codebase, a `lambda` expression is being used instead of a `def`.  I found all the places where this is happening using the [`ruff`](https://github.com/astral-sh/ruff) tool, and replaced all of those lambda expressions with a def function.

## Related Issue
Close #23734 

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials:


vedpatwardhan (commented):
I think having lambda functions makes the code more compact in most places where we've used them, but I'll leave this with @KareemMAX to decide whether replacing all lambda functions with `def` is a good idea. Thanks @Sai-Suraj-27 😄

CatB1t (changes requested):
Agree with @vedpatwardhan , however if we decide to go forward with using `def` instead of lambdas, I'd recommend avoiding using the same name for defining two different functions (for example: https://github.com/unifyai/ivy/pull/23738/files#diff-9f4b7017e01842b99516efb0ca6f9430483553a306468f858582c147c0227783R120-R126), let's go for more descriptive names.

Sai-Suraj-27:
> I think having lambda functions makes the code more compact in most places where we've used them, but I'll leave this with @KareemMAX to decide whether replacing all lambda functions with `def` is a good idea. Thanks @Sai-Suraj-27 😄



> Agree with @vedpatwardhan , however if we decide to go forward with using `def` instead of lambdas, I'd recommend avoiding using the same name for defining two different functions (for example: https://github.com/unifyai/ivy/pull/23738/files#diff-9f4b7017e01842b99516efb0ca6f9430483553a306468f858582c147c0227783R120-R126), let's go for more descriptive names.

Sure, I agree. I personally think at some of the places it is better to stay with lambda expressions only, like here:
https://github.com/unifyai/ivy/blob/1e1aaa04f6a2bbf2076afb9a0b569b6ee6cc4bc2/ivy/functional/frontends/torch/func_wrapper.py#L46

https://github.com/unifyai/ivy/blob/1e1aaa04f6a2bbf2076afb9a0b569b6ee6cc4bc2/ivy/functional/frontends/torch/nn/functional/loss_functions.py#L29

But at places like the following I think it is better to use `def` just to make it more readable and easy to understand.
https://github.com/unifyai/ivy/blob/1e1aaa04f6a2bbf2076afb9a0b569b6ee6cc4bc2/ivy/utils/assertions.py#L27

https://github.com/unifyai/ivy/blob/1e1aaa04f6a2bbf2076afb9a0b569b6ee6cc4bc2/ivy/functional/frontends/torch/nn/functional/vision_functions.py#L11

https://github.com/unifyai/ivy/blob/1e1aaa04f6a2bbf2076afb9a0b569b6ee6cc4bc2/ivy/functional/frontends/torch/nn/functional/vision_functions.py#L12

But I changed the lambda expressions to `def` functions at all the places. So, please let me know at what places you think it would be better to revert back to lambda expressions. I will revert the changes made back to lambda expressions at those places, Thank you.

KareemMAX (approved):
Hi @vedpatwardhan @CatB1t @Sai-Suraj-27,
I'd actually disagree with `lambda` being a compact way of writing code. Having regular functions would be better for code readability in my mind. And I think the changes made by @Sai-Suraj-27 are equivalent in AST. So there is no need to change any variable names/function names.

Waiting for your approval to merge @vedpatwardhan @CatB1t

vedpatwardhan:
> Hi @vedpatwardhan @CatB1t @Sai-Suraj-27, I'd actually disagree with `lambda` being a compact way of writing code. Having regular functions would be better for code readability in my mind. And I think the changes made by @Sai-Suraj-27 are equivalent in AST. So there is no need to change any variable names/function names.
> 
> Waiting for your approval to merge @vedpatwardhan @CatB1t

If you think having regular functions are a better choice, then I'm happy for you to merge this PR. Thanks @KareemMAX 😄

Sai-Suraj-27:
@KareemMAX sir, I have resolved the merge conflicts, can you check now, Thank you.

PR #23771, opened 2023-09-17
feat(functional api): super-setting ivy.tensordot to support batched_modes parameter. 

### LINKED ISSUE #23703: tensordot - Add `batched\_modes` parameter to `ivy.tensordot`

Madjid-CH:
## Related Issue
Close #23703


hello-fri-end (changes requested):
Hey @Madjid-CH, thanks for working on this issue. Just went through the changes and it looks like we're adding the same logic for `_tensordot_with_batched_modes` in all the backends. It would be much cleaner to convert this function to a mixed partial function to handle the `batch_modes` parameter. This would mean that we need to add a compositional implementation which handles all the cases in the input and the backend implementations are only used when `batched_modes` parameter is `None`. Here's the section in the deepdive about partial mixed function (https://unify.ai/docs/ivy/overview/deep_dive/function_types.html#partial-mixed-functions). Happy to clarify anything further )

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -17,6 +18,12 @@
 # --------------- #
 
 
+def _avoid_torch_unsupported_cases(x1, x2, dtype):
+    if "64" not in dtype[0]:
+        if x1.ndim == 1 or x2.ndim == 1:
+            dtype[0] = "float64" if "float" in dtype[0] else "int64"
+
+
```
hello-fri-end, line 26: I'm not sure if I follow this. Can you explain why this is needed?

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -1046,6 +1054,33 @@ def test_tensordot(*, dtype_x1_x2_axis, test_flags, backend_fw, fn_name, on_devi
...
+        ([4, 2, 3, 3], [3, 4, 2, 3], [(0, 3), (1, 3)], [1, 2]),
+        ([4, 3, 3], [3, 4, 2], [], []),
+        ([4, 3, 3], [3, 4, 2], [], [(0,), (1,)]),
+    ],
+)
+def test_tensordot_with_batched_modes_parameter(
+    backend_fw, shape1, shape2, axes, batched_modes
+):
+    import ivy
+    from tensorly.tenalg.core_tenalg import tensordot
```
hello-fri-end, line 1071: We shouldn't be adding `Tensorly` as a dependency.

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -1046,6 +1054,33 @@ def test_tensordot(*, dtype_x1_x2_axis, test_flags, backend_fw, fn_name, on_devi
...
+
+    ivy.set_backend(backend_fw)
+    tensor = ivy.random_uniform(shape=shape1)
+    tensor2 = ivy.random_uniform(shape=shape2)
+    res = ivy.tensordot(tensor, tensor2, axes=axes, batched_modes=batched_modes)
+    if backend_fw == "tensorflow":
+        tensor = tensor.numpy()
+        tensor2 = tensor2.numpy()
+    true_res = tensordot(tensor, tensor2, modes=axes, batched_modes=batched_modes)
+    assert np.allclose(res, true_res)
```
hello-fri-end, line 1081: Is this test adapted from TensorLy?

ivy/functional/ivy/linear_algebra.py
```diff
@@ -2426,7 +2436,54 @@ def tensordot(
...
-    return current_backend(x1, x2).tensordot(x1, x2, axes=axes, out=out)
+    if batched_modes is not None:
+        axes = _get_valid_contraction_modes_for_axes(x1.shape, x2.shape, axes)
+        batched_modes = _get_valid_contraction_modes_for_batches(
+            x1.shape, x2.shape, batched_modes
+        )
+        return _tensordot_with_batched_modes(x1, x2, axes, batched_modes)
+    return current_backend(x1, x2).tensordot(
+        x1, x2, axes=axes, batched_modes=batched_modes, out=out
+    )
```
hello-fri-end, line 2447: This creates scalability problems e.g. if we decide to add a new backend in the future that does not have a native `tensordot` function then you would need to implement the function compositionally ( with and without `batch_modes`) for that backend. So, it's best we don't call the backend implementation here and implement the logic of `batch_modes` being None in ivy itself.

ivy/functional/backends/torch/linear_algebra.py
```diff
@@ -425,27 +425,22 @@ def svdvals(x: torch.Tensor, /, *, out: Optional[torch.Tensor] = None) -> torch.
...
 
-    # handle tensordot for axes==0
-    # otherwise call with axes
-    if axes == 0:
-        ret = (x1.reshape(x1.size() + (1,) * x2.dim()) * x2).type(dtype)
-    else:
-        ret = torch.tensordot(x1, x2, dims=axes).type(dtype)
-    return ret
+
+tensordot.partial_mixed_handler = _batched_modes_is_none
```
hello-fri-end, line 443: Let's directly add the lambda function here instead of keeping it in another file. This enhances readability and will be helpful when the user is debugging their code.

hello-fri-end (changes requested):
Hey @Madjid-CH, thanks for looking into this task. Happy to review once again after you have resolved the comments :)

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
Madjid-CH: As discussed in this [message](https://discord.com/channels/799879767196958751/1028303853315162112/1153082103710363839)

```python
x1=torch.tensor([0], dtype=torch.uint8) 
x2 = torch.tensor([1], dtype=torch.uint8)
torch.tensordot(x1, x2, dims=1)
```
```
RuntimeError: expected scalar type Long but found Byte
```
pytorch does not support tensordot with scaler values for with dtypes less than 64 bit so I modified the test to avoid them

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
Madjid-CH: the input in pytest.mark.parametrize are from tensorly

ivy/functional/ivy/linear_algebra.py
Madjid-CH: actually all the backends have a batched_modes parameter but it's ignored so the non None batched_modes param only get handled in ivy without going to the backend

ivy/functional/backends/torch/linear_algebra.py
Madjid-CH: understood

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
hello-fri-end: Got it, thanks for explaining again!

ivy/functional/ivy/linear_algebra.py
hello-fri-end: What I meant was we might add a backend in the future that does not have a `tensordot` function in which case you would need to implement it compositionally in the backend. To avoid that, the compositional implementation should be able to handle both the cases (1) where `batch_modes` is `None` (2) `batch_modes` is not `None` and we shouldn't be deferring to the backend implementation using `current_backend(*args).tensordot(*args,**kwargs)`. Hope it makes sense now. Happy to discuss if you have counter thoughts as well :))

ivy/functional/backends/torch/linear_algebra.py
hello-fri-end: Yeah, I understand your concern as well but having it in the backend implementation will be very useful for debugging.

Madjid-CH: understood

ivy/functional/ivy/linear_algebra.py
Madjid-CH: I see, thanks for clarifying, so do you suggest, should I replace it with `ivy.tensordot(*args, **kwargs)` instead ?
and also I noticed that all other function in this file use `current_backend(*args)` in their implementation.

hello-fri-end: Actually, we shouldn't be using either of those since they'll end up calling the same function again ( for a future backend that does not have a primary implementation as I mentioned earlier). Rather, we should be implementing the logic for `ivy.tensordot` when `batched_modes` is `None` in our ivy implementation. The idea is the the compositional implementation should be able to handle all possible inputs that can be passed to the function. Hope it makes sense now :)

Madjid-CH: but what if the backend actually has a `tensordot` implementation what should we do then ?

Madjid-CH: @hello-fri-end I modified `tensordot `as requested

hello-fri-end: > but what if the backend actually has a `tensordot` implementation what should we do then ?

In that case, `ivy.tensordot` will actually be the backend implementation after the backend is set :) Again, refer to section about partial mixed functions in the deep-dive for more details :smile:

ivy/functional/backends/jax/linear_algebra.py
```diff
@@ -375,19 +373,24 @@ def svdvals(x: JaxArray, /, *, out: Optional[JaxArray] = None) -> JaxArray:
...
 def tensordot(
     x1: JaxArray,
     x2: JaxArray,
     /,
     *,
     axes: Union[int, Tuple[Sequence[int], Sequence[int]]] = 2,
+    batched_modes: Optional[Union[int, Tuple[List[int], List[int]]]] = None,
     out: Optional[JaxArray] = None,
 ) -> JaxArray:
-    x1, x2 = promote_types_of_inputs(x1, x2)
```
hello-fri-end, line 387: We should use `ivy.promote_types` to determine the output datatype and cast the result to that dtype. This is necessary so that all the backends return a consistent dtype for a given input.
```python
ret_dtype = ivy.promote_types(x1.dtype, x2.dtype)
```

ivy/functional/backends/numpy/linear_algebra.py
```diff
@@ -313,19 +312,24 @@ def tensorsolve(
...
 def tensordot(
     x1: np.ndarray,
     x2: np.ndarray,
     /,
     *,
     axes: Union[int, Tuple[List[int], List[int]]] = 2,
+    batched_modes: Optional[Union[int, Tuple[List[int], List[int]]]] = None,
     out: Optional[np.ndarray] = None,
 ) -> np.ndarray:
-    x1, x2 = ivy.promote_types_of_inputs(x1, x2)
```
hello-fri-end, line 325: Same as above, let's detemine the ret_dtype and cast the result to that dtype.

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -1017,12 +1025,13 @@ def test_svdvals(*, dtype_x, test_flags, backend_fw, fn_name, on_device):
...
-        available_dtypes=helpers.get_dtypes("numeric"),
+        available_dtypes=helpers.get_dtypes("valid"),
         min_num_dims=1,
         max_num_dims=5,
         min_dim_size=1,
         max_dim_size=10,
     ),
+    test_gradients=st.just(False),
 )
 def test_tensordot(*, dtype_x1_x2_axis, test_flags, backend_fw, fn_name, on_device):
```
hello-fri-end, line 1036: Given that we converted this function to a partial mixed function, we need to make some changes to the test. This is because the function will now use the primary implementation or the compositional one depending on the input, so we need to make sure that we generate a valid set of dtypes for each example. Here's the section in the [docs](https://unify.ai/docs/ivy/overview/deep_dive/ivy_tests.html#testing-partial-mixed-functions) explaining how to write the tests for partial mixed functions. Happy to help here if you find it a bit confusing :)

hello-fri-end (changes requested):
Hey @Madjid-CH, thanks for taking the time and making the changes I have requested so far. Let's resolve these final comments after which this should be good to merge. Happy to review once again when you are don e:)

PR #23786, opened 2023-09-18
fix: exclude bool dtypes in less backend function for tensorflow

Mghrabi:
According to docs [here](https://www.tensorflow.org/api_docs/python/tf/math/less) `bool` is not of the supported dtypes, and It caused failed tests for frontend functions that supports `bool` dtype


Infrared1029:
Makes sense, Thanks for the fix @Mghrabi!

PR #24170, opened 2023-09-19
fix: added support for bfloat16 representation in torch's to_numpy , …

RickSanchezStoic:
fix: added support for bfloat16 representation in torch's to_numpy ,added ml_dtypes to requirements


vedpatwardhan (changes requested):
As discussed yesterday, let's just add `ml_dtypes` to be the default behaviour of `numpy` in `ivy` rather than it being just for `torch`'s `to_numpy` given how important this is for the testing pipeline as well. Ofc we wouldn't need the try except either as we'd know when `bfloat16` gets supported in `numpy` through `ml_dtypes` anyway. Thanks 😄

ivy/functional/backends/torch/general.py
```diff
@@ -133,12 +133,20 @@ def to_numpy(
...
+                except TypeError:
+                    # although ml_dtypes would enable bfloat16
+                    # behaviour in numpy, but torch is restricting
+                    # it here, so we need to do this
+                    default_dtype = ivy.default_float_dtype(as_native=True)
+                    if default_dtype is torch.bfloat16:
+                        x = x.to(torch.float32)
+                    else:
+                        x = x.to(default_dtype)
+                    from ml_dtypes import bfloat16
```
vedpatwardhan, line 147: The `import ml_dtypes` statement should be done in the `ivy/functional/ivy/general.py` file above the `import numpy as np` import. We wouldn't need to import it again here as a result.
Secondly, I think even without the `try-except` we can just have the logic from line `142` to `149` directly under the `if` block on line `135` and that should work as well.
Also, let's create an issue on the `torch` repo and link it here with a `ToDo` comment.
Thanks @RickSanchezStoic 😄

vedpatwardhan (changes requested):

ivy/functional/backends/torch/general.py
```diff
@@ -133,12 +134,15 @@ def to_numpy(
         x = x.resolve_neg().resolve_conj()
         if copy:
             if x.dtype is torch.bfloat16:
```
vedpatwardhan, line 136: actually we could simplify this to do something like this instead so that we don't need to cast `bfloat16` to `float32`, what do you think?
```
if copy:
    return np.array(x.tolist(), dtype=ivy.as_ivy_dtype(x.dtype))
```

vedpatwardhan (changes requested):
Just requested a minor change, thanks @RickSanchezStoic 😄

vedpatwardhan (approved):
Please revert the changes to the `demos` submodule before merging, thanks @RickSanchezStoic 😄

PR #23659, opened 2023-09-15
feat(torch backend): removed manual dtype casting.

Madjid-CH:
# PR Description
used dtype checker script:
```
import torch


dtypes = [
    torch.int8,
    torch.int16,
    torch.int32,
    torch.int64,
    torch.uint8,
    torch.bfloat16,
    torch.float16,
    torch.float32,
    torch.float64,
    torch.complex64,
    torch.complex128,
    torch.bool,
]


def get_supported_and_unsupported_dtypes():
    supported_dtypes = set()
    unsupported_dtypes = set()

    for dtype in dtypes:
        try:
            x = torch.tensor(
                [
                    [
                        1,
                        2,
                    ],
                    [2, 1],
                ],
                dtype=dtype,
            )
            x1 = torch.tensor([1, 2, 3], dtype=dtype)
            x2 = torch.tensor([1, 2, 3], dtype=dtype)
            print(dtype, cummax(x))
            supported_dtypes.add(dtype)
        except Exception as e:
            print(e)
            unsupported_dtypes.add(dtype)

    return supported_dtypes, unsupported_dtypes


supported_dtypes, unsupported_dtypes = get_supported_and_unsupported_dtypes()

supported_dtypes = sorted(str(d).replace("torch.", "") for d in supported_dtypes)
unsupported_dtypes = sorted(str(d).replace("torch.", "") for d in unsupported_dtypes)

print("Supported Data Types:")
m = f"""@with_supported_dtypes(
        {{"2.0.1 and below": {tuple(supported_dtypes)}}}, backend_version
    )"""
print(m)

print("\nUnsupported Data Types:")
m = f"""@with_unsupported_dtypes(
        {{"2.0.1 and below": {tuple(unsupported_dtypes)}}}, backend_version
    )"""
print(m)

```

PR #24615, opened 2023-09-19
fix(torch backend): reintroduced necessary manual dtype casting.

Madjid-CH:
Revert unnecessary changes unifyai/ivy#23659

### DIFF
--- a/ivy/functional/backends/torch/elementwise.py
+++ b/ivy/functional/backends/torch/elementwise.py
@@ -60,7 +60,8 @@ def imag(
     out: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
     if val.dtype not in (torch.complex64, torch.complex128):
-        return torch.zeros_like(val, dtype=val.dtype)
+        ret = torch.imag(val.to(torch.complex64))
+        return ret.to(val.dtype)
     return torch.imag(val)
 
 
--- a/ivy/functional/backends/torch/elementwise.py
+++ b/ivy/functional/backends/torch/elementwise.py
@@ -59,10 +59,12 @@ def imag(
     *,
     out: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
-    if val.dtype not in (torch.complex64, torch.complex128):
+    if val.dtype in (torch.complex64, torch.complex128):
+        return torch.imag(val)
+    else:
         ret = torch.imag(val.to(torch.complex64))
         return ret.to(val.dtype)
-    return torch.imag(val)
+    
 
 
 imag.support_native_out = False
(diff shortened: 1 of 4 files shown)


ivy/functional/backends/torch/elementwise.py
```diff
@@ -60,7 +60,8 @@ def imag(
     out: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
     if val.dtype not in (torch.complex64, torch.complex128):
-        return torch.zeros_like(val, dtype=val.dtype)
+        ret = torch.imag(val.to(torch.complex64))
```
vedpatwardhan, line 63: I don't think we should cast both `complex64` and `complex128` to `complex64`, we might clip large values

ivy/functional/backends/torch/statistical.py
```diff
@@ -83,6 +83,10 @@ def mean(
             return ivy.inplace_update(out, x)
         else:
             return x
+    if "float" not in str(x.dtype) and "complex" not in str(x.dtype):  # unsupported
+        return torch.mean(x.to(torch.float32), dim=axis, keepdim=keepdims, out=out).to(
```
vedpatwardhan, line 87: Same here, we shouldn't really hardcode type casting to `float32` on this line, it should be based on whatever's the value of `x.dtype`

vedpatwardhan (changes requested):
Just requested a couple of changes, thanks @Madjid-CH 😄

ivy/functional/backends/torch/elementwise.py
Madjid-CH: yeah make sense

Madjid-CH: here we are casting dtypes other than `complex` to `complex64`,
I'll refactor it to make it more readble.

vedpatwardhan: Yeah that's an interesting point tbh, do we want to casting all other dtypes to `complex64` or just mark complex dtypes as supported for this backend if `torch.imag` doesn't support any other dtypes than `complex` ones as mentioned in their [docs](https://pytorch.org/docs/stable/generated/torch.imag.html). What do you think? Thanks @Madjid-CH 😄

vedpatwardhan (changes requested):

ivy/functional/backends/torch/elementwise.py
Madjid-CH: if we don't want any additional memory overhead like we did with the other functions we can mark the only supported types. And I don't see any reason why this function should be an exception. So let's just mark `complex` as supported.
@MahmoudAshraf97 what do you think ?

vedpatwardhan (approved):
lgtm! let's see what @MahmoudAshraf97 thinks, thanks @Madjid-CH 😄

Madjid-CH:
> Hey @Madjid-CH, could you please move this code to a branch on your fork instead of the main repo? You'd need to create a new PR from that branch to `main` instead, would be helpful to keep the main repo clean. Thanks 😄

ooh I didn't notice that, thanks for notifying me @vedpatwardhan

Issue #23044, opened 2023-09-04
[Bug]: Inconsistent `axis` Normalization in `ivy.gather` Across Backends


akshatvishu:
### Bug Explanation

The current implementation of the `gather` function in Ivy normalizes the axis parameter differently than native implementations in frameworks like `TensorFlow` , `PyTorch` or `Paddle`. Specifically, the line `axis = axis % len(params.shape)` may not be in line with how` TensorFlow` , `PyTorch` or `Paddle` handle `axis` normalization.

### Steps to Reproduce Bug

we're setting the value of `axis` to be: `axis = axis % len(params.shape)` across all backend at `ivy`
For `jax` backend:   https://github.com/unifyai/ivy/blob/16d690d318bb0044378be534cf621b2c4bb9160d/ivy/functional/backends/jax/general.py#L123

For `torch` backend: https://github.com/unifyai/ivy/blob/16d690d318bb0044378be534cf621b2c4bb9160d/ivy/functional/backends/torch/general.py#L174
For `tensorflow` backend: https://github.com/unifyai/ivy/blob/16d690d318bb0044378be534cf621b2c4bb9160d/ivy/functional/backends/tensorflow/general.py#L103
For `NumPy` backend:https://github.com/unifyai/ivy/blob/16d690d318bb0044378be534cf621b2c4bb9160d/ivy/functional/backends/numpy/general.py#L75
For `Paddle` backend:https://github.com/unifyai/ivy/blob/16d690d318bb0044378be534cf621b2c4bb9160d/ivy/functional/backends/paddle/general.py#L149

Now, this is how  `axis` value is being set at various framework native APIs:

for `torch`:
At `Pytorch` [gather](https://github.com/pytorch/pytorch/blob/4146be192ead477360a2763c5005e46a9485c3bf/torch/_inductor/lowering.py#L2472) we use a helper called : [_validate_dim](https://github.com/pytorch/pytorch/blob/4146be192ead477360a2763c5005e46a9485c3bf/torch/_inductor/lowering.py#L1100C4-L1100C18)

for `TensorFlow`:
for [tf.gather](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/ragged/ragged_gather_ops.py#L99) we use a helper called [get_positive_axis](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/array_ops.py#L6659)

[...]

PR #25447, opened 2023-09-20
fix bug  #23044

Aryan8912:
adding a get_positive_axis_for_gather function to normalize the axis parameter, and then using it in the ivy.gather function in the Jax backend


Aryan8912:
Fix the bug #23044

PR #20907, opened 2023-07-27
Implement Multinomial

AwkNinja:
Close #19456

PR #25492, opened 2023-09-20
feat: Implement frontend for paddle.multinomial and tests

### LINKED ISSUE #19456: multinomial

AwkNinja:
# PR Description
Implemented frontend for paddle.random.multionomial with tests.

## Related Issue

Close #19456

## Checklist

- [x] Did you add a function?
- [x] Did you add the tests?


AwkNinja:
Hi @MuhammedAshraf2020 can you please review my PR?

MuhammedAshraf2020:
Hey @AwkNinja!
Thanks for your contribution, will review this PR soon 🙏🏻

MuhammedAshraf2020:
Hello @AwkNinja!
Great job! However, there is just one little thing that needs to be fixed, as I will demonstrate below.

ivy_tests/test_ivy/test_frontends/test_paddle/test_random.py
```diff
@@ -7,6 +7,45 @@
...
+    fn_tree="paddle.tensor.random.multinomial",
+    input_dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("valid"),
+        min_value=0.25,
+        max_value=0.25,
+        min_dim_size=4,
+        max_dim_size=4,
+        shape=helpers.get_shape(min_num_dims=1, max_num_dims=2, min_dim_size=2),
+    ),
+    num_samples=st.integers(min_value=1, max_value=2),
```
MuhammedAshraf2020, line 21: You're not covering the entire space here. For instance, you should expand the x-range beyond just 0.25. Additionally, there's no need to set the min dimensions and maximum dimension size to 4.

AwkNinja: Actually there is a reason for it I informed Haris as you can see here. https://github.com/unifyai/ivy/pull/20907#discussion_r1306372771 
Basically numpy and jax needs the array so that the sum of all elements always adds up to 1. I think other framework normalize the array themselves. So that is the reason tuned the parameters so that the array is always added upto 1. But it works with other values as well for the other frameworks that normalize it.

AwkNinja:
Hi @MuhammedAshraf2020, I just made a composite strategy to test this function as it makes more sense to normalize the array than to always create a static array of the same size and values. There was also a case that when `num_samples` is greater than the shape dimensions the `replacement` parameter should be set to `True` otherwise the function throws an error as it is not allowed, so I also handled that case in the composite strategy. Please have look and let me what you think.

ivy_tests/test_ivy/test_frontends/test_paddle/test_random.py
```diff
@@ -7,6 +7,64 @@
...
+        helpers.dtype_and_values(
+            available_dtypes=helpers.get_dtypes("valid"),
+            shape=helpers.get_shape(min_num_dims=1, max_num_dims=2, min_dim_size=2),
+        )
+    )
+    num_samples = draw(st.integers(min_value=1, max_value=10))
+    if num_samples > 2:
+        replacement = True
+    else:
+        replacement = draw(st.booleans())
```
MuhammedAshraf2020, line 26: Can you explain why the replacement has to be True for values bigger than 2 ?

ivy_tests/test_ivy/test_frontends/test_paddle/test_random.py
```diff
@@ -7,6 +7,64 @@
...
+
+    input_dtype, x = input_dtype_and_x
+
+    total = sum(x)
+    x = [arr / total for arr in x]
```
MuhammedAshraf2020, line 31: If the goal is to normalize the x values, I would recommend using min_value and max_value in  `helpers.dtype_and_values` method instead .

ivy_tests/test_ivy/test_frontends/test_paddle/test_random.py
AwkNinja: The number of dimensions of the input probability distribution (`x` array in this case) should be <= 2, this is the max all of the frameworks allow. Which is the reason I set `max_num_dims=2`. So if we have 2 dims we can't take a larger sample than population when `replacement=False`, hence whenever `num_samples > 2` we will set `replacement=True`.

ivy_tests/test_ivy/test_frontends/test_paddle/test_random.py
AwkNinja: That is what I did before but in order for the array to always sum to 1. We need to set `min_value` and `max_value` to `0.25` and `min_dim_size` and `max_dim_size` to 4. This means that the array will always be `[0.25 0.25 0.25 0.25]`. I think this isn't a good way to test the function as the array is static and limits it. So after generating the array with `helpers.dtype_and_values` I normalize it so the sum is 1.

MuhammedAshraf2020: make sense to me!

AwkNinja:
Hi @MuhammedAshraf2020 I see that you have resolved the conversations, so is the PR good to merge?

MuhammedAshraf2020:
Yeah, I just wanted to test it locally. 
LGTM! Thanks

PR #25657, opened 2023-09-20
fix(ivy): Fixes misleading ivy array attribute error

AnnaTz:
```
x = ivy.array([[1, 2]])
x.unsqueeze(1)
```
Before the fix: `AttributeError: 'numpy.ndarray' object has not attribute '__getattr__'. Did you mean: 'squeeze'?`
After the fix: `AttributeError: 'numpy.ndarray' object has no attribute 'unsqueeze'. Did you mean: 'squeeze'?`

### DIFF
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -431,8 +431,11 @@ class Array(
     def __getattr__(self, item):
         try:
             attr = self._data.__getattribute__(item)
-        except AttributeError:
-            attr = self._data.__getattr__(item)
+        except AttributeError as e:
+            if "has no attribute '__getattribute__'" in str(e):
+                attr = self._data.__getattr__(item)
+            else:
+                raise e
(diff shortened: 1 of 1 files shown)


ivy/data_classes/array/array.py
```diff
@@ -431,8 +431,11 @@ def __getattribute__(self, item):
     def __getattr__(self, item):
         try:
             attr = self._data.__getattribute__(item)
-        except AttributeError:
-            attr = self._data.__getattr__(item)
+        except AttributeError as e:
```
vedpatwardhan, line 434: actually this is something that was being worked on in [this](https://github.com/unifyai/ivy/pull/16339) PR but ofc we could close that one now. @Infrared1029 could you please review this PR as the author of that task? Thanks @AnnaTz 😄

vedpatwardhan (changes requested):

PR #25678, opened 2023-09-20
paddle frontend Tensor.__add__ method

Sam-Armstrong:
Hey @AnnaTz, not exactly sure who's working most on the paddle frontends right now so tagging you, apologies! paddle.Tensor.__add__ is one of the missing frontends that is affecting our paddle transpilation tests, so implementing it here. The method seems to be working fine - the transpilation tests pass now - but the frontend test I wrote is failing with a segmentation fault (at least locally) and really not sure what might be going wrong. 

As I side note, I'm not sure special instance methods like this currently have a todo list issue for the paddle frontend? Might have missed it though?

Thanks!

### DIFF
--- a/ivy/functional/frontends/paddle/tensor/tensor.py
+++ b/ivy/functional/frontends/paddle/tensor/tensor.py
@@ -61,6 +61,13 @@ class Tensor:
     # Special Methods #
     # -------------------#
 
+    @with_unsupported_dtypes(
+        {"2.5.1 and below": ("bool", "unsigned", "int8", "float16", "bfloat16")},
+        "paddle",
+    )
+    def __add__(self, y, name=None):
+        return paddle_frontend.add(self, y)
+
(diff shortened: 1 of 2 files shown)


AnnaTz:
ivy-gardener

AnnaTz:
Yeah you are right. These are not in any list 🤔. The [paddle api documentation](https://www.paddlepaddle.org.cn/documentation/docs/zh/api/paddle/Tensor_cn.html#tensor) doesn't mention these at all, so we have no way of knowing which inplace and/or dunder methods are supported in practice.

AnnaTz (approved):
There is an issue in `test_frontend_method`. I made a quick hack to run the test locally and it passes, so the code is fine. I'd say let's merge this. 
I have let the testing team know about the bug.

Sam-Armstrong:
Ok sure, thanks for looking into it 😁

PR #25702, opened 2023-09-20
Stateful repr

RickSanchezStoic:
adds `__repr__` to stateful module, also adds extra repr to linear layer, extra_repr needs to be added to other stateful layers too.


vedpatwardhan (changes requested):
The changes look great, let's also add it for the other stateful classes, thanks @RickSanchezStoic 😄

PR #25713, opened 2023-09-20
fixed the torch.nn.Param being passed in torch.nn.Param, which changes the requires_grad flag to True

umairjavaid:
If you pass a `torch.nn.Parameter` with `requires_grad = False`, to a `torch.nn.Parameter`, it'll change the `requires_grad` to `True`. This will raise a torch error, `RuntimeError: Only Tensors of floating point and complex dtype can require gradients`, if the passed tensor is of type int. 
This error is raised when we try to `ivy.unify` the following torch model without the proposed changes. 

In my solution, I've added an if condition in the `_TorchIvyModule._replace_update_v` function to check if the passed torch tensor is of type `torch.nn.Parameter` or not. If it is, we don't need to pass it to `torch.nn.Parameter`, we can set it as it is!

Run the following test case without and with the changes to see the issue and fix!
```
import ivy
import torch


class torchmodel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.indices = torch.nn.Parameter(torch.tensor([1,2,3]).to(torch.int64), requires_grad=False)
    
    def forward(self):
        return self.indices
    

model = torchmodel()
print(model.forward())

model = ivy.unify(model, source="torch")
print(model._forward())
```

### DIFF
--- a/ivy/stateful/module.py
+++ b/ivy/stateful/module.py
@@ -1210,7 +1210,10 @@ class _TorchIvyModule(Module):
                 native.__setattr__(k, v)
             elif isinstance(v, torch.Tensor):
                 # noinspection PyProtectedMember
-                native.__setattr__(k, torch.nn.Parameter(v))
+                if (isinstance(v,torch.nn.Parameter)):
+                    native.__setattr__(k, v)
+                else:
+                    native.__setattr__(k, torch.nn.Parameter(v))
             else:
                 raise ivy.utils.exceptions.IvyException(
                     f"found item in variable container {v} which was neither a sub"
--- a/ivy/stateful/module.py
+++ b/ivy/stateful/module.py
@@ -1210,10 +1210,7 @@ class _TorchIvyModule(Module):
                 native.__setattr__(k, v)
             elif isinstance(v, torch.Tensor):
                 # noinspection PyProtectedMember
-                if (isinstance(v,torch.nn.Parameter)):
-                    native.__setattr__(k, v)
-                else:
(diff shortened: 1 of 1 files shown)


ivy/stateful/module.py
```diff
@@ -1210,7 +1210,10 @@ def _replace_update_v(self, new_v, native=None):
                 native.__setattr__(k, v)
             elif isinstance(v, torch.Tensor):
                 # noinspection PyProtectedMember
-                native.__setattr__(k, torch.nn.Parameter(v))
```
vedpatwardhan, line 1213: I think we could just replace [this](https://github.com/unifyai/ivy/pull/25713/files#diff-9d3cd8360ac6368fb7cfebdd75ef04230b2963749af122e8515e44ea6ddf3971L1213) line with `native.__setattr__(k, torch.nn.Parameter(v, requires_grad=v.requires_grad))` instead and remove the `if-else` added.
@hmahmood24 and @mattbarrett98 could you please confirm if this change to the converters causes any issues in the transpiler? Thanks!

vedpatwardhan (changes requested):
Just requested a minor change, let's see what @hmahmood24 and @mattbarrett98 think. Thanks 😄

ivy/stateful/module.py
umairjavaid: changed it. That also works!

hmahmood24 (approved):
Change works fine with the tracer/transpiler 👍🏼

mattbarrett98 (approved):

PR #25758, opened 2023-09-21
feat: added paddle.kthvalue frontend function

### LINKED ISSUE #25753: kthvalue

Mghrabi:
Closes #25753

### DIFF
--- a/ivy/functional/frontends/paddle/search.py
+++ b/ivy/functional/frontends/paddle/search.py
@@ -33,6 +33,28 @@ def argsort(x, /, *, axis=-1, descending=False, name=None):
     return ivy.argsort(x, axis=axis, descending=descending)
 
 
+# kthvalue
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float32", "float64", "int32", "int64")}, "paddle"
+)
+@to_ivy_arrays_and_back
+def kthvalue(x, k, axis=None, keepdim=False, name=None):
+    if axis is None:
+        axis = -1
+    sorted_input = ivy.sort(x, axis=axis)
+    sort_indices = ivy.argsort(x, axis=axis)
+
+    values = ivy.gather(sorted_input, ivy.array(k - 1), axis=axis)
+    indices = ivy.gather(sort_indices, ivy.array(k - 1), axis=axis)
+
+    if keepdim:
+        values = ivy.expand_dims(values, axis=axis)
+        indices = ivy.expand_dims(indices, axis=axis)
+
+    ret = (values, indices)
+    return ret
+
+
 @with_supported_dtypes(
     {"2.5.1 and below": ("float32", "float64", "int32", "int64")},
     "paddle",
--- a/ivy_tests/test_ivy/test_frontends/test_paddle/test_search.py
+++ b/ivy_tests/test_ivy/test_frontends/test_paddle/test_search.py
@@ -1,6 +1,7 @@
 # global
 import numpy as np
-from hypothesis import strategies as st
+from hypothesis import assume, strategies as st
+
 
 # local
 import ivy_tests.test_ivy.helpers as helpers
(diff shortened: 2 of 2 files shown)


ivy/functional/frontends/paddle/search.py
```diff
@@ -33,6 +33,32 @@ def argsort(x, /, *, axis=-1, descending=False, name=None):
     return ivy.argsort(x, axis=axis, descending=descending)
 
 
+# kthvalue
+@with_supported_dtypes(
+    {"2.5.1 and below": ("float32", "float64", "int32", "int64")}, "paddle"
+)
+@to_ivy_arrays_and_back
+def kthvalue(x, k, axis=-1, keepdim=False, *, out=None):
```
Infrared1029, line 41: paddle.kthvalue does not have an `out` argument, just a `name`, default value for axis is `None`

ivy_tests/test_ivy/test_frontends/test_paddle/test_search.py
```diff
@@ -148,6 +149,45 @@ def test_paddle_argsort(
...
 
+# kthvalue
+@handle_frontend_test(
+    fn_tree="paddle.kthvalue",
+    dtype_input_axis=helpers.dtype_values_axis(
+        available_dtypes=helpers.get_dtypes("valid"),
+        min_num_dims=2,
+        valid_axis=True,
+        force_int_axis=True,
+    ).filter(lambda v: len(np.unique(v[1][0])) == len(np.ravel(v[1][0]))),
```
Infrared1029, line 160: can you elaborate more on the filtering condition here?

Infrared1029:
Thanks for the PR @Mghrabi, left a few comments!

ivy_tests/test_ivy/test_frontends/test_paddle/test_search.py
Mghrabi: Sure!, The whole condition filters out arrays that have only unique values. The indexing part `v[1][0]` is to return the test values array, `v` is a value similar to `(['int64'], [array([[0],[2])], 0)` which is a tuple contains the values array in the index = `1`, so `v[1]` returns `[array([[0],[2])]` in the above example and then `v[1][0]` extract the array itself returning `array([[0],[2])`. the `len(np.unique(v[1][0])) == len(np.ravel(v[1][0]))` condition assures only arrays with unique values to be in the test set.

**Reason for having that filter**:
I will be using torch code to demonstrate examples since paddle and torch share the same unexpected behavior in `kthvalue` and torch in their docs mentioned that behavior [here](https://pytorch.org/docs/stable/generated/torch.kthvalue.html) so this will help in elaboration. 
For the following examples `k=1` means to return the smallest value in the array:
```
>>> tens = torch.tensor([1,2,3])
>>> torch.kthvalue(tens, k=1) [//]:
torch.return_types.kthvalue(
values=tensor(1),
indices=tensor(0))
```
In the above example, it runs as expected, it returns the smallest value which is `1` and it's index `0` , everything looks fine, but the unexpected behavior occurs with duplicates, for the following example :
```
>>> tens = torch.tensor([0,0,0,0])
>>> torch.kthvalue(tens, k=1)
torch.return_types.kthvalue(
values=tensor(0),
indices=tensor(1))
```
In the above example, the smallest value is `0`, and there are three zeros in the array. The unexpected behavior is that torch returned the index `1` instead of `0` (which was expected to return the first index in which the smallest value occurred), paddle has a similar behavior but it returns the index `2` instead of `1` in that particular example, both frameworks seem to not care about returning the first index the smallest value occurred as long as the index correctly points to the smallest value (which we can't say it's not a correct implementation, but not expected), while our backends in ivy return the first index the smallest value occurred which is `0` and that mismatch results into failing tests when the arrays are not unique like that one I got with paddle when not including the filter:
```
 AssertionError:  the results from backend jax and ground truth framework paddle do not match
E        [0]!=[2] 
E       
E       
E       Falsifying example: test_paddle_kthvalue(
E           on_device='cpu',
E           frontend='paddle',
E           backend_fw='jax',
E           dtype_input_axis=(['int32'], [array([[0],
E                    [0],
E                    [0],
E                    [0]], dtype=int32)], 0),
E           k=1,
...
```
and with `torch.kthvalue` frontend function I get the exact same error but with `1` instead of `2` in `[0]!=[2]` error.

So I found it reasonable to skip testing for duplicated arrays as they don't affect the correctness of the result, though make tests fail.

ivy/functional/frontends/paddle/search.py
Mghrabi: For the `axis` argument I found in the docs that if `axis` is `None` then the default value of `axis` will be -1, that's why I directly assigned the default value to be -1, but you are CORRECT I should follow the signature anyway sorry for that. And for the `out` yeah my bad :(.

Infrared1029:
comments makes sense, merging!

PR #25784, opened 2023-09-21
paddle frontend Tensor.numpy()

Sam-Armstrong:
Add missing paddle.Tensor.numpy() frontend method which is required for some of the PaddlePaddle transpilation tests.

There is a problem with the test here with a numpy backend, because it complains about the frontend returning a native array, but I think the functionality here is correct. The test passes for the other backends.


ivy-leaves:

Thank you :hugs:

AnnaTz:
ivy-gardener

AnnaTz (approved):
I don't see any error in the test now. And since it was only in one backend it probably wasn't caused by the frontend code anyway. I'd say we can merge this.

PR #22501, opened 2023-08-24
feat(loss-functions): add poisson_nll_loss function

akshatvishu:
Close #21727

PR #25872, opened 2023-09-21
fix(ivy-backends): fix import issue for losses at ivy backends

akshatvishu:
# PR Description

This commit adds missing support for losses across all Ivy backends (TensorFlow, Paddle, Jax, NumPy, and Torch).

# Reviewer Request
@HaiderSultanArc

### DIFF
--- a/ivy/functional/backends/jax/experimental/__init__.py
+++ b/ivy/functional/backends/jax/experimental/__init__.py
@@ -14,6 +14,7 @@ from .general import *
 from .gradients import *
 from .layers import *
 from .linear_algebra import *
+from .losses import *
 from .manipulation import *
 from .norms import *
 from .random import *
--- a/ivy/functional/backends/jax/experimental/__init__.py
+++ b/ivy/functional/backends/jax/experimental/__init__.py
@@ -24,3 +24,26 @@ from .sorting import *
 from .sparse_array import *
 from .statistical import *
 from .utility import *
+
+del (
+    activations,
+    converters,
+    creation,
+    data_type,
+    device,
+    elementwise,
+    general,
+    gradients,
+    layers,
+    linear_algebra,
+    losses,
+    manipulation,
+    norms,
+    random,
+    searching,
+    set,
+    sorting,
+    sparse_array,
+    statistical,
+    utility,
+)
--- a/ivy/functional/backends/jax/experimental/__init__.py
+++ b/ivy/functional/backends/jax/experimental/__init__.py
@@ -24,26 +24,3 @@ from .sorting import *
 from .sparse_array import *
 from .statistical import *
 from .utility import *
-
-del (
-    activations,
-    converters,
-    creation,
-    data_type,
-    device,
-    elementwise,
-    general,
-    gradients,
-    layers,
-    linear_algebra,
-    losses,
-    manipulation,
-    norms,
-    random,
-    searching,
-    set,
-    sorting,
-    sparse_array,
-    statistical,
-    utility,
-)
--- a/ivy/functional/backends/numpy/experimental/__init__.py
+++ b/ivy/functional/backends/numpy/experimental/__init__.py
@@ -13,6 +13,7 @@ from .general import *
 from .gradients import *
 from .layers import *
 from .linear_algebra import *
+from .losses import *
 from .manipulation import *
 from .norms import *
 from .random import *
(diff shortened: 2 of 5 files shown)


ivy-leaves:

Thank you :hugs:

akshatvishu:
Hey @HaiderSultanArc  I added the `del` functionality as discussed in our discussion with Ved. 

refs: [DiscordLink](https://discord.com/channels/799879767196958751/1000044049333485648/1154460521089740950)

ivy/functional/backends/jax/experimental/__init__.py
```diff
@@ -23,3 +24,26 @@
 from .sparse_array import *
 from .statistical import *
 from .utility import *
+
+del (
```
HaiderSultanArc, line 28: What does this `del` do here exactly? 🤔

akshatvishu: The `del` keyword is used to delete references to certain sub-modules that were previously imported. These sub-modules, such as `activations`, `converters`, `creation`, etc., are first imported from the current package (denoted by `.`  ) and used for various tasks in the code.

Once these tasks are completed and the sub-modules are no longer needed, they are deleted from the namespace using the `del` keyword. This is done to keep the namespace clean and manageable, avoiding potential naming conflicts and making the set of available names more concise.

To summarize :  First we  imports several sub-modules, uses them for certain tasks, and then removes them(using `del`) from the namespace to keep the code clean and efficient  This practice is particularly useful in larger code bases where careful management of memory and namespaces is important.

You can see a live example of it at <https://github.com/tensorflow/tensorflow/blob/v2.13.0/tensorflow/__init__.py#L31>

- The code imports the Python package, which in turn brings in modules from `tensorflow.core` and `tensorflow.python`. This has the side effect of introducing the symbols` python` and `core` into the current namespace. To keep the namespace clean and minimize the risk of naming conflicts, these symbols are explicitly removed using the `del` statement.In my limited knowledge,  It's a strategy often used in larger, complex codebases like `TensorFlow` to manage the namespace effectively and prevent potential issues down the line. By ensuring that the names `python` and `core` are not accessible after the import, the code avoids any ambiguities or conflicts that could arise later.

HaiderSultanArc: Yes but we need to import the packages that you are deleting 😄. That's the entire purpose of `from .losses *`.
It's ok if we can `ivy.functional.losses.<loss_fn>` in our namespace. It's consistent with rest of the `__init__.py` files. I think just `from .losses import *` is enough to solve our problem.

akshatvishu: what I meant is :
```py
# Let assume  losses.py contains a function called poisson_nll_loss
from losses import *

# poisson_nll_loss is now available in the current namespace
print(poisson_nll_loss(...))

# Deleting the 'losses' name from the namespace
del losses

# poisson_nll_loss is STILL available in the current namespace
print(poisson_nll_loss(...))

# Trying to access 'losses' will result in a NameError
try:
    print(losses.poisson_nll_loss(...))  # This will raise a NameError
except NameError as e:
    print(f"Caught an error: {e}")

```

HaiderSultanArc: Yes, I know what you are trying to do. But our namespace is not structured like that. It will also cause an issue with the setting of the backend. I tried to do this before that's why I know. 😄

akshatvishu: Thanks for the answer. I will remove it from the pr !

HaiderSultanArc:
@akshatvishu it seems it's causing some issues with setting of Backend as I suspected 😅
image

HaiderSultanArc:
Actually, it seems like a CI issue 🤔 It's from `run_tests no 78 and 115`. @RashulChutani maybe you should have a look at this.
image

akshatvishu:
> Actually, it seems like a CI issue 🤔 It's from `run_tests no 78 and 115`. @RashulChutani maybe you should have a look at this. image

Is it still showing these failures after I remove the `del` ?

HaiderSultanArc:
> Is it still showing these failures after I remove the `del` @HaiderSultanArc ?

These failures are due to some issues on our Testing side. Let's wait for it to be fixed. I think there are no issues in the updates you made but lets just wait to be clear! 😄

akshatvishu:
@HaiderSultanArc  Shall I close this `PR` as these imports are already added as a part of my [poission_nll_loss PR](https://github.com/unifyai/ivy/pull/22501/files) ?

PR #25873, opened 2023-09-21
fix(paddle_backend.round): align rounding behavior with `np.round`

### LINKED ISSUE #25865: [Bug]: `ivy_paddle_backend.round()` function inconsistent with `numpy.round()`
### Bug Explanation

When using the [ivy_paddle_backend.round()](https://github.com/unifyai/ivy/blob/main/ivy/functional/backends/paddle/elementwise.py#L833) function, I found that its results are inconsistent with those of [numpy.round()](https://numpy.org/doc/stable/reference/generated/numpy.round.html) for several input numbers. I ran a test comparing the two and discovered that they produce different results in `7` cases.

### Steps to Reproduce Bug

```py
import numpy as np
import paddle
import ivy
import ivy.functional.frontends.paddle as ivy_paddle

def test_rounding_against_numpy():
    test_numbers = [
        0.1, 0.5, 0.9, # generic decimals
        1.5, 2.5, 3.5, 4.5, # even .5s
        1.4, 2.6, 3.7, 4.8, # other decimals
        -1.5, -2.5, -3.5, -4.5, # negative even .5s
        -1.4, -2.6, -3.7, -4.8, # negative other decimals
    ]

    errors = 0

    for num in test_numbers:
        numpy_res = np.round(num)
        ivy.set_paddle_backend()
        paddle_input = paddle.to_tensor(num, dtype=paddle.float32)
        paddle_res = ivy.round(paddle_input)
        paddle_res = ivy.to_numpy(paddle_res)
        if numpy_res != paddle_res:
            print(f"Error: For input {num}, numpy gives {numpy_res} but paddle gives {paddle_res}.")
            errors += 1

    if errors == 0:
        print("All tests passed for ivy_paddle.round()!")
    else:
        print(f"{errors} tests failed.")

test_rounding_against_numpy()

"""
Error: For input 0.9, numpy gives 1.0 but paddle gives 0.0.
Error: For input 1.4, numpy gives 1.0 but paddle gives 2.0.
Error: For input 2.6, numpy gives 3.0 but paddle gives 2.0.
Error: For input 4.8, numpy gives 5.0 but paddle gives 4.0.
Error: For input -1.4, numpy gives -1.0 but paddle gives -2.0.
Error: For input -2.6, numpy gives -3.0 but paddle gives -2.0.
Error: For input -4.8, numpy gives -5.0 but paddle gives -4.0.
7 tests failed.
"""
```

### Environment

linux+docker+vscode

### Ivy Version

v1.1.9

### Backend

- [ ] NumPy
- [ ] TensorFlow
- [ ] PyTorch
- [ ] JAX
- [x] Paddle

### Device

CPU

akshatvishu:
# PR Description

The rounding function is now aligned to behave similarly to np.round.

## Related Issue

Close #25865

## Checklist

- [x] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/backends/paddle/elementwise.py
+++ b/ivy/functional/backends/paddle/elementwise.py
@@ -830,39 +830,45 @@ def pow(
     return paddle.pow(x1, x2)
 
 
+# Implementation based on TensorFlow's scalar_round_half_to_even_op logic
+# Reference: https://github.com/tensorflow/tensorflow/blob/7f1050a6976d11bfb0bb37bdfc82350c0a238faa/tensorflow/core/kernels/cwise_ops.h#L510  # noqa: E501
+def _round_half_to_even(x):
+    round_val = paddle.floor(x + 0.5)
+    fraction = round_val - x
+
+    # Identify elements with a fractional part of 0.5
+    mask = paddle.equal(fraction, paddle.to_tensor(0.5, dtype=fraction.dtype))
+
+    # Round to the nearest even number if the fraction is 0.5
+    even_round_val = 2 * paddle.floor(0.5 * x + 0.5)
+
+    # Combine the results
+    return paddle.where(mask, even_round_val, round_val)
+
+
+# This function aims to mimic the behavior of np.round similar to how tf.experimental.numpy.round does # noqa: E501
+# Reference for tf.experimental.numpy.round:https://github.com/tensorflow/tensorflow/blob/v2.13.0/tensorflow/python/ops/numpy_ops/np_array_ops.py#L724 # noqa: E501
 def round(
     x: paddle.Tensor, /, *, decimals: int = 0, out: Optional[paddle.Tensor] = None
 ) -> paddle.Tensor:
-    def _np_round(x):
-        # this is a logic to mimic np.round behaviour
-        # which rounds odd numbers up and even numbers down at limits like 0.5
+    x = paddle.to_tensor(x, dtype=x.dtype)
+    dtype_ = x.dtype
+    factor = math.pow(10, decimals)
 
-        one = paddle.to_tensor(1, dtype="int64")
+    # Handle floating point and complex numbers
+    if paddle.is_floating_point(x) or paddle.is_complex(x):
+        factor = paddle.to_tensor(factor)
+        factor = paddle.cast(factor, dtype_)
+    else:
+        float_dtype_ = paddle.float32  # paddle.get_default_dtype()
+        x = x.astype(float_dtype_)
+        factor = paddle.cast(factor, float_dtype_)
 
-        # check if the number is even or odd
-        is_even = paddle.bitwise_and(paddle_backend.trunc(x).astype("int64"), one) == 0
+    x = paddle.multiply(x, factor)
+    x = _round_half_to_even(x)
+    x = paddle.divide(x, factor)
 
-        # round the number to the nearest integer
-        round_x = paddle.sign(x) * paddle.where(
-            is_even, paddle.floor(x.abs()), paddle.ceil(x.abs())
-        )
-
-        # if the number was rounded up from an even number
-        #   round the number down to the nearest even number
-        return paddle.where(
-            paddle.logical_and(
-                paddle.bitwise_and(round_x.astype("int64"), one) == 1.0,
-                is_even,
-            ),
-            round_x - 1.0,
-            round_x,
-        )
-
-    if x.dtype not in [paddle.float32, paddle.float64]:
-        if paddle.is_complex(x):
-            return paddle.complex(_np_round(x.real()), _np_round(x.imag()))
-        return _np_round(x.astype("float32")).astype(x.dtype)
-    return _np_round(x).astype(x.dtype)
+    return x.astype(dtype_)
 
 
 def trunc(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
--- a/ivy/functional/backends/paddle/elementwise.py
+++ b/ivy/functional/backends/paddle/elementwise.py
@@ -833,14 +833,14 @@ def pow(
 # Implementation based on TensorFlow's scalar_round_half_to_even_op logic
 # Reference: https://github.com/tensorflow/tensorflow/blob/7f1050a6976d11bfb0bb37bdfc82350c0a238faa/tensorflow/core/kernels/cwise_ops.h#L510  # noqa: E501
 def _round_half_to_even(x):
-    round_val = paddle.floor(x + 0.5)
+    round_val = paddle_backend.floor(x + 0.5)
     fraction = round_val - x
 
     # Identify elements with a fractional part of 0.5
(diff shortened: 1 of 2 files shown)


bipinKrishnan:
Hey @akshatvishu, thanks for noting the issue. But we shouldn't be comparing paddle backend with numpy, instead the outputs can be checked against the output of native paddle operations. Also while working on issues that are not related to any open tasks, please get approval from the issue assignee before making a PR. Hope that makes sense 🙂.

akshatvishu:
> Hey @akshatvishu, thanks for noting the issue. But we shouldn't be comparing paddle backend with numpy, instead the outputs can be checked against the output of native paddle operations.

But I saw the comment at `paddle_backend` and it says that we're trying to mimic to `np.round`:<https://github.com/unifyai/ivy/blob/main/ivy/functional/backends/paddle/elementwise.py#L838>

AnnaTz:
Hi @bipinKrishnan, even though the PR's title is a bit misleading, I think this is a valid PR to fix `test_round`. `test_round` is failing in the paddle backend currently, which is in turn causing failures in the frontends as well (as @akshatvishu pointed out above). 

> we shouldn't be comparing paddle backend with numpy, instead the outputs can be checked against the output of native paddle operations

Basically, `ivy.round` is supposed to mimic numpy's rounding operation but the paddle backend is not consistent with that. What you are saying about comparing to native paddle operations is true for paddle frontend functions, not ivy functions. For ivy functions we compare to the approach which we have chosen to follow (could be any framework's, usually it's the approach followed by most frameworks or what the array api standard defines).

bipinKrishnan:
> Hi @bipinKrishnan, even though the PR's title is a bit misleading, I think this is a valid PR to fix `test_round`. `test_round` is failing in the paddle backend currently, which is in turn causing failures in the frontends as well (as @akshatvishu pointed out above).
> 
> > we shouldn't be comparing paddle backend with numpy, instead the outputs can be checked against the output of native paddle operations
> 
> Basically, `ivy.round` is supposed to mimic numpy's rounding operation but the paddle backend is not consistent with that. What you are saying about comparing to native paddle operations is true for paddle frontend functions, not ivy functions. For ivy functions we compare to the approach which we have chosen to follow (could be any framework's, usually it's the approach followed by most frameworks or what the array api standard defines).

Yeah makes sense 🙂. Thanks for the clarification!

bipinKrishnan:
Some of the rounding functions raise unsupported dtype issue, other than that the changes LGTM!

akshatvishu:
I've been thinking about what @bipinKrishnan said "But we shouldn't be comparing paddle backend with numpy, instead the outputs can be checked against the output of native paddle operations." and which was further corrected by   @AnnaTz that "We want to mimic paddle.round only for Ivy paddle frontends but not for ivy.round calls in-general which should follow numpy.round" .

So, I was thinking that maybe we can  of add a flag called `numpy_behaviour`, set to `True` by default at paddle_backend round implementation . If someone wants the `paddle.round` behavior, they can just flip the flag to `False ` (I add a `return paddle.round ` call at paddle_backend). This way, users can pick whatever  behavior they want and may save unnecessary large frontend composition that may have to be added in the future to mimic/get `paddle.round`  behavior for paddle frontends . 

What do you guys think about this?

AnnaTz:
> I've been thinking about what @bipinKrishnan said "But we shouldn't be comparing paddle backend with numpy, instead the outputs can be checked against the output of native paddle operations." and which was further corrected by @AnnaTz that "We want to mimic paddle.round only for Ivy paddle frontends but not for ivy.round calls in-general which should follow numpy.round" .
> 
> So, I was thinking that maybe we can of add a flag called `numpy_behaviour`, set to `True` by default at `paddle_backend .round` implementation . If someone wants the `paddle.round` behavior, they can just flip the flag to `False ` (I add a `return paddle.round ` call at paddle_backend). This way, users can pick whatever behavior they want and may save unnecessary large frontend composition that may have to be added in the future to mimic/get `paddle.round` behavior for paddle frontends .
> 
> What do you guys think about this?

I agree as this would 1) keep the paddle frontend implementation short, and 2) allow the use of the native paddle function in some cases for efficiency (`ivy.round` will have to become a partial mixed function).
Instead of `numpy_behaviour`, I would suggest naming the argument `nearest_even`, which I think is more intuitive.

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -830,39 +830,45 @@ def pow(
...
+    else:
+        float_dtype_ = paddle.float32  # paddle.get_default_dtype()
+        x = x.astype(float_dtype_)
+        factor = paddle.cast(factor, float_dtype_)
+
+    x = paddle.multiply(x, factor)
+    x = _round_half_to_even(x)
+    x = paddle.divide(x, factor)
+
+    return x.astype(dtype_)
```
AnnaTz, line 872: The frontend changes are great. Ran the test locally and it passes now. 
But `ivy.round`'s backend is still failing.  I'm getting `ivy.utils.exceptions.IvyValueError: paddle: round: (InvalidArgument) cast(): argument 'x' (position 0) must be Tensor, but got float`.

AnnaTz (changes requested):

ivy/functional/backends/paddle/elementwise.py
akshatvishu: > The frontend changes are great. Ran the test locally and it passes now. But `ivy.round`'s backend is still failing. I'm getting `ivy.utils.exceptions.IvyValueError: paddle: round: (InvalidArgument) cast(): argument 'x' (position 0) must be Tensor, but got float`.

Let me take a look, it might due to `factor`

ivy/functional/backends/paddle/elementwise.py
```diff
@@ -830,39 +830,48 @@ def pow(
...
 
-    if x.dtype not in [paddle.float32, paddle.float64]:
-        if paddle.is_complex(x):
-            return paddle.complex(_np_round(x.real()), _np_round(x.imag()))
-        return _np_round(x.astype("float32")).astype(x.dtype)
-    return _np_round(x).astype(x.dtype)
+# This function aims to mimic the behavior of np.round similar to how tf.experimental.numpy.round does # noqa: E501
+# Reference for tf.experimental.numpy.round:https://github.com/tensorflow/tensorflow/blob/v2.13.0/tensorflow/python/ops/numpy_ops/np_array_ops.py#L724 # noqa: E501
+@with_unsupported_device_and_dtypes(
+    {"2.5.1 and below": {"cpu": ("bfloat16", "float16")}}, backend_version
```
AnnaTz, line 852: You need to add `"complex"` to the unsupported dtypes. After that the test will be passing and we can merge.

AnnaTz (changes requested):

ivy/functional/backends/paddle/elementwise.py
akshatvishu: Thanks for pointing it out @AnnaTz , I made the suggested changes !

PR #25917, opened 2023-09-22
Ivy experimental API: `solve_triangular`

### LINKED ISSUE #25883: solve_triangular
Reserve subtask to work on. Required for efficient implementation of #25882

bridgekat:
Close #25883


bridgekat:
Some considerations that might go against "superset behaviour":

- Some backends support parameters `left`, `adjoint` or `transpose` (indicating whether to solve `XA = B` instead of `AX = B`) but the conventions are different. I remove it since it does not provide a consistent way of improving performance on all backends (the functionality can be achieved by calling `adjoint()`).
- The parameter `unit_diagonal` (or `unitriangular`, indicating whether the diagonal is guaranteed to be all `1`'s) is not supported on all backends, but affects gradient output on the supported ones. I remove it in favour of consistent behaviours on all backends.

bridgekat:
Thank you for reviewing @nassimberrada, I have adjusted the implementation and test to include those arguments (just realised a way to stop the gradients on the diagonal).

Also merged with upstream and made sure `test_solve_triangular` is passing (locally with default configuration i.e. numpy results checked against torch/jax/tensorflow).

bridgekat:
Added. Just note that I don't believe it is a bug in paddlepaddle either:

```python
import paddle

a = paddle.to_tensor([[-1.]], dtype=paddle.float32)
a.stop_gradient = False
b = paddle.to_tensor([[-1.]], dtype=paddle.float32)
b.stop_gradient = False
c = paddle.linalg.triangular_solve(a, b, upper=False, transpose=False, unitriangular=False)

paddle.grad(outputs=[c], inputs=[a], create_graph=False, retain_graph=True)
```

This piece of code replicating the failed example correctly outputs `1`.

bridgekat:
Thanks, this seems to be a bug with paddlepaddle indeed.

Screenshot 2023-10-02 222559

I tried multiple times and got different outputs each time 😂

PR #25936, opened 2023-09-23
feat : added paddle instance method not_equal with test

### LINKED ISSUE #25934: not_equal

AryanSharma21:
close #25934


AryanSharma21:
Hey @Daniel4078. Please help me understand the issue with the test failure here, I am not sure what is it with 

`_ ERROR collecting ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_tensor.py _`

while I have tested it over the GitHub codespace and the function test is working fine. 
having Same issue with local env also . . . 


Here is the test result from the GitHub codespace

Screenshot 2023-09-23 at 1 35 20 PM

Daniel4078:
> Hey @Daniel4078. Please help me understand the issue with the test failure here, I am not sure what is it with
> 
> `_ ERROR collecting ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_tensor.py _`
> 
> while I have tested it over the GitHub codespace and the function test is working fine. having Same issue with local env also . . .
> 
> Here is the test result from the GitHub codespace
> 
> Screenshot 2023-09-23 at 1 35 20 PM

I suppose that error is about you do not have all the needed libraries to be imported in your local env. Can I have more context about that error though?

AryanSharma21:
> > Hey @Daniel4078. Please help me understand the issue with the test failure here, I am not sure what is it with
> > `_ ERROR collecting ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_tensor.py _`
> > while I have tested it over the GitHub codespace and the function test is working fine. having Same issue with local env also . . .
> > Here is the test result from the GitHub codespace
> > Screenshot 2023-09-23 at 1 35 20 PM
> 
> I suppose that error is about you do not have all the needed libraries to be imported in your local env. Can I have more context about that error though?

It was working fine like 3-4 days ago (locally) when a previous PR was merge closed, so today I pulled upstream changes and am having this issue with local env

AryanSharma21:
Hey @Daniel4078 , I have looked more into this error actually what I could get is that there is a dependency conflict between tensorflow-macos and numpy versions on my local system, unlike tensorflow nvidia which runs on codespace. I have tried to align them but for now, it fails. Other failing tests seem to be working fine now.

Daniel4078:
So now, all your tests are failing here due to a strange error: 
```
  File "/ivy/ivy/utils/backend/ast_helpers.py", line 279, in exec_module
    exec(compiled_obj, module.__dict__)
  File "/ivy/ivy/__init__.py", line 1, in <module>
    # global
KeyError: 139774142937936
```
I will try ask other Ivy members about this as I havn't seen any thing like this before.

AryanSharma21:
> So now, all your tests are failing here due to a strange error:
> 
> ```
> Traceback (most recent call last):
>   File "/ivy/ivy/utils/backend/ast_helpers.py", line 279, in exec_module
>     exec(compiled_obj, module.__dict__)
>   File "/ivy/ivy/__init__.py", line 1, in <module>
>     # global
> KeyError: 139774142937936
> ```
> 
> I will try ask other Ivy members about this as I havn't seen any thing like this before.

Sure 🙂

Daniel4078:
> > So now, all your tests are failing here due to a strange error:
> > ```
> > Traceback (most recent call last):
> >   File "/ivy/ivy/utils/backend/ast_helpers.py", line 279, in exec_module
> >     exec(compiled_obj, module.__dict__)
> >   File "/ivy/ivy/__init__.py", line 1, in <module>
> >     # global
> > KeyError: 139774142937936
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I will try ask other Ivy members about this as I havn't seen any thing like this before.
> 
> Sure 🙂

It's an issue with `with_backend` not working with the newly added sub-backends. The number is actually the `id` of the `ivy` imported which is why it varies. Someone is currently looking into this as it is a wide spread problem now.

AryanSharma21:
@Daniel4078 OK, that seems like something to be fixed internally. Does that mean this pull request would be merged only after the case is resolved?

Daniel4078:
> @Daniel4078 OK, that seems like something to be fixed internally. Does that mean this pull request would be merged only after the case is resolved? Also, thank you for resolving the merge conflict 🙂

It was fixed earlier today, I will check if there is any other problems left in your code.

Daniel4078 (changes requested):
your tests currently fails for errors like this:
```
E   AssertionError:  the results from backend tensorflow and ground truth framework paddle do not match
E    True!=False 
E   
E   
E   Falsifying example: test_paddle_tensor_not_equal(
E       # The test always failed when commented parts were varied together.
E       frontend='paddle',
E       on_device='cpu',
E       backend_fw='tensorflow',
E       dtype_and_x=(['float32', 'float32'],
E        [array(-7.450581e-09, dtype=float32),
E         array(-1.1175871e-08, dtype=float32)]),
```
I suppose it is due to a dtype precision problem, maybe you should change the supported dtype for the function to not let too precise dtype to be allowed.

AryanSharma21:
> your tests currently fails for errors like this:
> 
> ```
> E   AssertionError:  the results from backend tensorflow and ground truth framework paddle do not match
> E    True!=False 
> E   
> E   
> E   Falsifying example: test_paddle_tensor_not_equal(
> E       # The test always failed when commented parts were varied together.
> E       frontend='paddle',
> E       on_device='cpu',
> E       backend_fw='tensorflow',
> E       dtype_and_x=(['float32', 'float32'],
> E        [array(-7.450581e-09, dtype=float32),
> E         array(-1.1175871e-08, dtype=float32)]),
> ```
> 
> I suppose it is due to a dtype precision problem, maybe you should change the supported dtype for the function to not let too precise dtype to be allowed.

Will look into it

AryanSharma21:
Hey, @Daniel4078 , we had some discussion on the discord server regarding the failures on this tensor instance method and 
found that the test failure here is related to backend function behavior.  @akshatvishu has created an issue for the same as mentioned here. Please look into it.🙂

Daniel4078 (commented):
I think that the implementation here(https://github.com/unifyai/ivy/blob/a843cfbbdfc5b20cea07938b22ab151dced1751c/ivy/functional/frontends/paddle/logic.py#L286) needs to be changed as the ivy.not_equal does not behave exactly the same as paddle's version.

AryanSharma21:
Hi @Daniel4078, What I have understood by now is that while testing the frontend instance method for
`paddle_frontend.not_equal` the results are compared to the native `paddle.not_equal` which causes failures `only for close floating point values`. But when using the function with ivy, we end up using the `ivy.not_equal` which is a modified version and works fine for all cases. So wouldn't that be possible if we limit the testing of this frontend instance method to only `integer` values? Because the behavior of ivy.not_equal is fine as verified [here](https://github.com/unifyai/ivy/issues/26234#:~:text=PaddlePaddle%20not_equal%20result%3A%20False) ?

Daniel4078:
> Hi @Daniel4078, What I have understood by now is that while testing the frontend instance method for `paddle_frontend.not_equal` the results are compared to the native `paddle.not_equal` which causes failures `only for close floating point values`. But when using the function with ivy, we end up using the `ivy.not_equal` which is a modified version and works fine for all cases. So wouldn't that be possible if we limit the testing of this frontend instance method to only `integer` values? Because the behavior of ivy.not_equal is fine as verified [here](https://github.com/unifyai/ivy/issues/26234#:~:text=PaddlePaddle%20not_equal%20result%3A%20False%0AIVY%20modified%20Paddle%20not_equal%20result%3A%20True) ?

Well, we need to replicate paddle.not_equal's behavior in the paddle_frontend.not_equal, so we can not directly use ivy.not_equal there, the test fails just because our frontend function does not match the paddle version's behavior as suggested by another ivy member in the bug report issue.

AryanSharma21:
Please review @Daniel4078

Daniel4078 (approved):
Looks good to me, thank you!

PR #25976, opened 2023-09-23
feat: setting sub_backends by default if they are available

ShreyanshBardia:
part of [task](https://trello.com/c/ovTBv2KG/2766-set-a-sub-backend-by-default-if-available)

With these changes I think this [function](https://github.com/unifyai/ivy/blob/5b873228a2ff2efbf05ce944e27fc7fdbfc52066/ivy/utils/backend/handler.py#L406) becomes redundant and should be removed. 

@vedpatwardhan


vedpatwardhan (changes requested):
I'd suggest checking through the failing tests to see if any of the backend handler tests are related to the change made (which it most probably won't be).
Secondly, that line wouldn't be redundant but I suppose all the changes added in [this](https://github.com/unifyai/ivy/pull/23489) PR would be (sorry @Madjid-CH 😅), because if a sub-backend wasn't set then it means either it wasn't installed or the user opted out. In either case we shouldn't be throwing the warning as a result of these changes. Could you please also revert those changes? Thanks @ShreyanshBardia 😄

Madjid-CH:
> I'd suggest checking through the failing tests to see if any of the backend handler tests are related to the change made (which it most probably won't be). Secondly, that line wouldn't be redundant but I suppose all the changes added in [this](https://github.com/unifyai/ivy/pull/23489) PR would be (sorry @Madjid-CH 😅), because if a sub-backend wasn't set then it means either it wasn't installed or the user opted out. In either case we shouldn't be throwing the warning as a result of these changes. Could you please also revert those changes? Thanks @ShreyanshBardia 😄

do we have to consider if multiple sub-backends are available or we can ignore that for the moment since we currently only have one sub-backend ?
for example maybe we have an efficient implementation in another sub-backend a warning should be raised, right ?
or when setting a sub-backend by default, what's the criteria for such setting do we set the first one in the list or we should consider another factors for choosing a sub-backend.  
Thanks @vedpatwardhan

ShreyanshBardia:
Yes realised it later, we still would need to update `original_backend_dict`. It looks like you have already reverted the changes, let me know if anything else is needed. Also thhe test errors seem to be unrelated. Thanks @vedpatwardhan

vedpatwardhan:
> do we have to consider if multiple sub-backends are available or we can ignore that for the moment since we currently only have one sub-backend ? for example maybe we have an efficient implementation in another sub-backend a warning should be raised, right ? or when setting a sub-backend by default, what's the criteria for such setting do we set the first one in the list or we should consider another factors for choosing a sub-backend. Thanks @vedpatwardhan

Hey @Madjid-CH, I think the possibility of there being an implementation of multiple sub-backends for the same function is a bit rare, so we're not considering that for now. Basically for all functions if there's a sub-backend implementation it would be used by default on setting that backend if the library was installed. Hope that makes sense 😄

vedpatwardhan:
> Yes realised it later, we still would need to update `original_backend_dict`. It looks like you have already reverted the changes, let me know if anything else is needed. Also thhe test errors seem to be unrelated. Thanks @vedpatwardhan

Yep I've already reverted those changes, I think the PR is good to merge. Thanks @ShreyanshBardia 😄

PR #25982, opened 2023-09-23
update(torch-frontend): Scalars to be converted to default 0d torch Tensors

hirwa-nshuti:
As some `torch` functions supports the operation on scalar values this PR adds this functionality as there were mismatch due to how ivy handles the default dtypes.

Example `torch`:
```python
import torch

print(torch.add(torch.tensor([True]), 2).dtype) # torch.int64
```
With Ivy we were getting 

```python

print(torch_frontend.add(torch_frontend.tensor([True]), 2).dtype) # int32

```

This PR updates this behaviour to behave as the torch

- [x] Did you add a function?
- [ ] Did you add the tests?


ivy-leaves:

Thank you :hugs:

AnnaTz:
Could this issue be accounted for in `promote_types_of_torch_inputs` rather than `_to_ivy_array`? I think we need more examples where this occurs. I am a bit hesitant to make such a generalized change. `to_ivy_arrays_and_back` is used in **all** frontend functions, but I don't think the error appears in all tests.

hirwa-nshuti:
> Could this issue be accounted for in `promote_types_of_torch_inputs` rather than `_to_ivy_array`? I think we need more examples where this occurs. I am a bit hesitant to make such a generalized change. `to_ivy_arrays_and_back` is used in **all** frontend functions, but I don't think the error appears in all tests.

Yeah this can also be handled in `promote_types` and it makes sense, I'll update my implementation sorry for responding late😅

AnnaTz (approved):

PR #26000, opened 2023-09-24
fix: kl_div and superset for torch backend

ShreyanshBardia:
The previous implementation of `kl_div` was buggy and didn't follow input style of native backends i.e. `torch` and `paddle`. 
This PR aims at adding a mixed implementation of `kl_div`, earlier redundant code was present in backends (`numpy`, `jax`, `tensorflow`) which has been removed in favour of a compositional implementation added at `ivy/experimental/losses.py`. The previous implementation also didn't include a parameter `log_target` which is present in `torch` backend, this has been added following `Superset Approach`


ivy-leaves:

Thank you :hugs:

ivy/data_classes/array/experimental/losses.py
```diff
@@ -258,7 +258,7 @@ def kl_div(
         /,
         *,
         reduction: Optional[str] = "mean",
-        out: Optional[ivy.Array] = None,
+        log_target=False,
```
HaiderSultanArc, line 261: I don't know about this change 🤔 why did you remove the `out` parameter and introduce a new parameter in the functional API of ivy?

HaiderSultanArc: You can not remove `out` parameter. It's part of the standard for API that we are following all across the Functional API

ivy/data_classes/array/experimental/losses.py
```diff
@@ -294,7 +294,9 @@ def kl_div(
         >>> print(output_array)
         ivy.array(0.0916)
         """
-        return ivy.kl_div(self._data, target, reduction=reduction, out=out)
```
HaiderSultanArc, line 297: Same here, we are not supposed to remove `out` parameter from Functional API! 😄

ivy/functional/backends/jax/experimental/losses.py
```diff
@@ -64,26 +64,6 @@ def soft_margin_loss(
         return loss
 
 
-def kl_div(
```
HaiderSultanArc, line 67: You just removed the backend implementation from jax? 🤔 I am not following what you are trying to do. Can you elaborate more by providing description in the PR?

ivy/functional/backends/numpy/experimental/losses.py
```diff
@@ -75,29 +75,6 @@ def soft_margin_loss(
         return loss
 
 
-@with_unsupported_dtypes({"1.26.0 and below": ("bool", "bfloat16")}, backend_version)
```
HaiderSultanArc, line 78: Same here. You can't remove backend implementation from any backend unless you are creating a new implementation that is better

ivy/functional/ivy/experimental/losses.py
```diff
@@ -421,7 +421,7 @@ def kl_div(
     /,
     *,
     reduction: Optional[str] = "mean",
-    out: Optional[ivy.Array] = None,
```
HaiderSultanArc, line 424: Again, we can't just introduce and remove parameters from the Functional API of ivy without any explaination.

ShreyanshBardia: Is it necessary for every function to have `out` parameter. I was not aware of that, will make the required changes

ivy/functional/backends/jax/experimental/losses.py
ShreyanshBardia: We don't need a backend specific in these cases because there is no efficient implementation of it, so even if we define a compositional implementation, the function will fall back on to that (the one defined in `ivy/experimental/losses.py`). If you'll take a look at the previous implementations, they were using a compositional approach only, there is no primary function that exists for `numpy`, `jax`, `tensorflow`

ShreyanshBardia: We can define a compositional implementation in `ivy/experimental/losses.py`, so that for all the backends not having a primary or efficient implementation, this compositional implementation is called and we don't have to write the same function in each backend

ivy/data_classes/array/experimental/losses.py
ShreyanshBardia: I'll add back the `out` parameter, but as for the `log_target`, aren't we trying to capture the superset behaviour? If so then I think it is important to include it, given that `torch` implements it

ShreyanshBardia:
@HaiderSultanArc I have added back the mistakenly removed `out` argument, also I have updated the PR description to be more elaborate. Can you please take a look now? (Apologies for being ignorant about the `out` param) Thanks!

ivy/functional/backends/jax/experimental/losses.py
HaiderSultanArc: We still need to implement backend-specific implementation 😅 . Even if a backend doesn't provide the implementation we still provide that (as the composition of other ivy functions) because other backends are providing it.

ivy/functional/ivy/experimental/losses.py
HaiderSultanArc: I'll have a look at `poisson_nll_loss`

HaiderSultanArc: But yes it is necessary

ivy/functional/backends/jax/experimental/losses.py
HaiderSultanArc: Got it! So only Torch has an implementation of k1_div in its functional API. 😄 Looks good to me.

ivy_tests/test_ivy/test_functional/test_experimental/test_nn/test_losses.py
```diff
@@ -77,20 +78,23 @@ def test_huber_loss(
...
     dtype_and_target,
     reduction,
+    log_target,
     test_flags,
     backend_fw,
     fn_name,
     on_device,
 ):
     input_dtype, input = dtype_and_input
+    input[0] = np.log(input[0])
```
HaiderSultanArc, line 94: Why are you taking log of the input here? 🤔

ivy_tests/test_ivy/test_functional/test_experimental/test_nn/test_losses.py
```diff
@@ -77,20 +78,23 @@ def test_huber_loss(
...
     target_dtype, target = dtype_and_target
-
+    if log_target:
+        target[0] = np.log(target[0])
```
HaiderSultanArc, line 97: if `log_target` is true then the user is expected to input a log of their target? 🤔

ShreyanshBardia: Yes that's correct. This is again from [`torch`](https://pytorch.org/docs/stable/generated/torch.nn.functional.kl_div.html) documentation  `A flag indicating whether target is passed in the log space. It is recommended to pass certain distributions (like softmax) in the log space to avoid numerical issues caused by explicit log. Default: False`

HaiderSultanArc:
@ShreyanshBardia merge main into your branch and let the CI run again.

ShreyanshBardia:
@HaiderSultanArc not sure about the errors on CI, they seem to run find locally

HaiderSultanArc:
@ShreyanshBardia are all the tests passing locally? 🤔

HaiderSultanArc:
Let's just ask for a final review from @vedpatwardhan 😄

ivy/functional/backends/torch/experimental/losses.py
```diff
@@ -102,18 +103,8 @@ def soft_margin_loss(
...
-            "uint8",
-            "int8",
-            "int16",
-            "int32",
-            "int64",
-            "bool",
-        )
-    },
+@with_supported_dtypes(
+    {"2.0.1 and below": "float"},
```
vedpatwardhan, line 107: The error message in [this](https://github.com/unifyai/ivy/actions/runs/6301635175/job/17107481998?pr=26000#step:3:30698) test failure is coming from here, basically it should be `("float",)` instead of `"float"`

vedpatwardhan (changes requested):
Just a minor change, thanks @ShreyanshBardia @HaiderSultanArc 😄

ivy/functional/backends/torch/experimental/losses.py
ShreyanshBardia: updated it, but it looks like the tests are not running now due to `dill` not being available. Thanks!

HaiderSultanArc:
@ShreyanshBardia I think you should run the tests locally and see if they pass 🤔 if `dill` missing error appears just do `pip install dill`

ShreyanshBardia:
Hi @HaiderSultanArc, like I mentioned the tests were passing for me in local earlier also (now as well), it was the CI which was throwing error. Also reference to `dill` being missing was also with respect to `CI`.

PR #26065, opened 2023-09-25
fix: Fix class method `from_CPTensor()` to use `cls` as the first parameter

### LINKED ISSUE #26063: Class method `from_CPTensor()` should use `cls` as the first parameter.
The class method `from_CPTensor()` in the `Parafac2Tensor` class currently uses `self` as the first parameter. However, it should use `cls` as the first parameter, as it is a class method and using `cls` is generally recommended.

https://github.com/unifyai/ivy/blob/f45b3c232e97fb53f661b633f7590f50d53b4196/ivy/data_classes/factorized_tensor/parafac2_tensor.py#L92

Sai-Suraj-27:
# PR Description
The class method `from_CPTensor()` in the `Parafac2Tensor` class currently uses `self` as the first parameter. However, it should use `cls` as the first parameter, as it is a class method.

https://github.com/unifyai/ivy/blob/f45b3c232e97fb53f661b633f7590f50d53b4196/ivy/data_classes/factorized_tensor/parafac2_tensor.py#L92

This PR changes the first parameter of the `from_CPTensor(`) method from self to `cls`. This makes the method more idiomatic and easier to read and understand. It also ensures that the method is consistent with other class methods in Python.

## Related Issue
Close #26063

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials:


MuhammedAshraf2020:
Hi @Sai-Suraj-27 
We wanted to express our gratitude for your contribution. Your efforts are greatly appreciated.

MuhammedAshraf2020:
LGTM!

PR #26187, opened 2023-09-27
Chore: Cache the ivy docker files

NripeshN:
Cache the docker image created in https://github.com/unifyai/ivy/blob/main/.github/workflows/dockerfile-push.yml to be cached for the whole organisation to use.

### DIFF
--- a/.github/workflows/dockerfile-push.yml
+++ b/.github/workflows/dockerfile-push.yml
@@ -36,7 +36,7 @@ jobs:
 
     # Build and push Docker image
     - name: Build and push Dockerfile
-      uses: docker/build-push-action@v3
+      uses: docker/build-push-action@v5
       with:
         context: .
         file: docker/Dockerfile
--- a/.github/workflows/dockerfile-push.yml
+++ b/.github/workflows/dockerfile-push.yml
@@ -2,14 +2,12 @@ name: Dockerfile Push
 
 on:
   schedule:
-    # run at everyday at 00:00 UTC
+    # run everyday at 00:00 UTC
     - cron:  '0 0 * * *'
   workflow_dispatch:
 
 jobs:
-
   build:
-
     runs-on: ubuntu-20.04
 
     steps:
@@ -24,8 +22,25 @@ jobs:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
 
+    # Cache Docker layers
+    - name: Cache Docker layers
+      uses: actions/cache@v3
+      with:
+        path: /tmp/.buildx-cache
+        key: ${{ runner.os }}-buildx-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-buildx-
+
+    # Build and push Docker image
     - name: Build and push Dockerfile
-      run: |
-        docker build . --file docker/Dockerfile --tag unifyai/ivy:latest
-        docker push unifyai/ivy:latest
+      uses: docker/build-push-action@v3
+      with:
+        context: .
+        file: docker/Dockerfile
+        push: true
+        tags: unifyai/ivy:latest
+        cache-from: type=local,src=/tmp/.buildx-cache
+        cache-to: type=local,dest=/tmp/.buildx-cache


.github/workflows/dockerfile-push.yml
```diff
@@ -24,8 +22,25 @@ jobs:
...
+        key: ${{ runner.os }}-buildx-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-buildx-
+
+    # Build and push Docker image
     - name: Build and push Dockerfile
-      run: |
-        docker build . --file docker/Dockerfile --tag unifyai/ivy:latest
-        docker push unifyai/ivy:latest
+      uses: docker/build-push-action@v3
```
Dsantra92, line 39: Action v5 is now available.

.github/workflows/dockerfile-push.yml
```diff
@@ -24,8 +22,25 @@ jobs:
...
+      with:
+        context: .
+        file: docker/Dockerfile
+        push: true
+        tags: unifyai/ivy:latest
+        cache-from: type=local,src=/tmp/.buildx-cache
```
Dsantra92, line 45: Difference between gha cache and local cache. Can you explain the difference between them? The original doc doesn't seem to have lot of information on how to use these variables.

.github/workflows/dockerfile-push.yml
```diff
@@ -24,8 +22,25 @@ jobs:
...
 
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
 
+    # Cache Docker layers
+    - name: Cache Docker layers
+      uses: actions/cache@v3
+      with:
+        path: /tmp/.buildx-cache
+        key: ${{ runner.os }}-buildx-${{ github.sha }}
```
Dsantra92, line 33: This creates individual docker caches over time. Better to use cache/restore and cache/save manually rather than relying on this.

Dsantra92 (changes requested):

.github/workflows/dockerfile-push.yml
NripeshN: GitHub Actions will periodically clean up older cache entries based on usage and the retention policy, so it's not an infinite accumulation.

.github/workflows/dockerfile-push.yml
NripeshN: GitHub Actions Cache (GHA Cache)
What is it? It's a storage space that GitHub provides. We use it to keep things we frequently use, so we don't have to fetch or recreate them every time our automation scripts (or "workflows") run.
Where is it? This cache is on GitHub's servers, which means it's remote and not on the same machine where our tasks actually run.
When do we use it? Mostly for things that don't change often but take time to fetch or build, like certain software packages or, in our case, parts of Docker images.
Any limits? Yes, it's like a storage locker – we can't store more than a certain amount, and if we don’t access it for a while, GitHub may clear it.

Local Cache
What is it? It's also a storage space but on the machine (or "runner") that's doing our tasks, not on GitHub's servers.
Where is it? Right on the same computer or "runner" where our GitHub Action is happening.
When do we use it? It's useful if we're doing several tasks in a row in the same workflow, and we don't want to redo the same steps over and over.
Any limits? The main thing is, it's temporary. Imagine it like RAM on a computer. If the machine restarts (which GitHub does often with their machines), everything in this cache is gone.

.github/workflows/dockerfile-push.yml
Dsantra92: Yep just create one key for cache for main. I think that will do the trick.

.github/workflows/dockerfile-push.yml
Dsantra92: We have to use github cache, we want this cached image to be available in all github workflows. Just to confirm here: we are caching the dockerfile right?

Dsantra92:
Hi @NripeshN did you have a chance to go through the last few comments I made?

.github/workflows/dockerfile-push.yml
KareemMAX: Yup, completely agree with @Dsantra92 on that. Although in @NripeshN defense: This cache will only work during weekends, where we don't push commits to github. Basically building the image immediately.

KareemMAX (changes requested):
Hey @NripeshN, great work.

I've some issues about the motivation around this PR. Ideally in most production scenarios it's [discouraged to use docker cache](https://pythonspeed.com/articles/disabling-docker-caching/#:~:text=When%20building%20Docker%20images%2C%20caching,not%20getting%20new%20security%20updates!). And we don't really get cache benefits, as this workflow is being run once everyday. Which is quite negligible.

Also be aware that there is a server-side cache on docker hub. The current scenario is:

1. We build without cache with the latest updates
2. Docker calculate layer hashes
3. If a layer exist on docker hub, docker simply doesn't upload it

KareemMAX:
> Hi @KareemMAX @Dsantra92,
> What changes do I have to make now?

@NripeshN do you still think the motivation for it stands?

Dsantra92:
Hey @KareemMAX this pr was meant to make the pulling of ivy docker image cached rather than building.

KareemMAX:
Hey @NripeshN @Dsantra92,
As per our meeting, I guess this PR should be closed. Is that correct?

PR #26197, opened 2023-09-27
Sort Function is already formatted correctly

### LINKED ISSUE #26044: Sort

GOW7527:
# PR Description

I only made a switch between two numbers in the docstrings just to flag that the sort function is already formatted correctly, or at least I believe it is formatted correctly.

## Related Issue
I don't understand what I should do, it seems like the function does not require any more formatting . sort #26044

Close #26044

## Checklist

- [x] Did you add a function? There doesn't seem to be any functions that need to be added
- [x] Did you add the tests? I did not add any tests

From my understanding I should not address this issues:
Implement the hypothesis testing for the function
Get the tests passing for your function, if they are failing before you start

Close #26044


ivy/functional/ivy/sorting.py
```diff
@@ -202,7 +202,7 @@ def sort(
     --------
     With :class:`ivy.Array` input:
 
-    >>> x = ivy.array([7, 8, 6])
```
Sarvesh-Kesharwani, line 205: Hey @GOW7527,
I don't know if it's an oversight,
but the order [7, 8, 6] is already in the unsorted order,

henceforth, this change is not necessary.
Thanks anyway.

PR #26199, opened 2023-09-27
(fix): for older torch versions without mps backend

mattbarrett98:
small fix for old torch versions which don't have an mps backend, was this picked up by the multiversion tests @vedpatwardhan ?

### DIFF
--- a/ivy/functional/backends/torch/device.py
+++ b/ivy/functional/backends/torch/device.py
@@ -96,7 +96,9 @@ def num_gpus() -> int:
 
 
 def gpu_is_available() -> bool:
-    return torch.backends.mps.is_available() or torch.cuda.is_available()
+    return (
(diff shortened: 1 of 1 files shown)


vedpatwardhan (approved):
lgtm, Feel free to merge, thanks @mattbarrett98 😄 
Actually the multi-version tests are still being integrated into the CI, so we haven't tested it yet basically (I know it's been a while since we've been trying to integrate it 😅)
(Given that the tests aren't running in the CI due to a docker issue, I'd suggest running a few ivy tests with the torch backend and ensuring that they're passing before you merge)

vedpatwardhan:
Actually a similar change would be required [here](https://github.com/unifyai/ivy/pull/26199/files#diff-95514c5ee4e99eb2323adc3c1c52251c6576d8e9393b882cd09eede86ba543f9R99) as well ig, thanks 😄

PR #26279, opened 2023-09-28
refactor(torch-frontend & testing): Updates the handling of `inplace` so that functions always return frontend arrays even when the inputs are ivy arrays

AnnaTz:
There are some frontend functions that are supposed to do an inplace update. In case the input is a frontend array, we update it inplace and the function returns the same frontend array. In case the input is an ivy array, we currently update it inplace and return that ivy array. However, it is counterintuitive for a frontend function to return an ivy array. 
This PR changes the above approach to: updating the input ivy array inplace but returning a frontend array (which contains that ivy array). This new approach also simplifies testing quite a bit because it removes the conditional handling of whether the return is frontend or ivy.


AnnaTz:
Not yet ready. Need to run more tests and also look at the rest of the frontends.

ivy-leaves:

Thank you :hugs:

PR #26292, opened 2023-09-28
feat: added TensorArray class to tensorflow

ShreyanshBardia:
close [task](https://trello.com/c/q8qpnQ0C/2819-implement-tftensorarray-class-in-the-tensorflow-frontend)

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
+++ b/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
@@ -6,10 +6,7 @@ import tensorflow as tf
 # local
 import ivy
 import ivy_tests.test_ivy.helpers as helpers
-from ivy_tests.test_ivy.helpers import (
-    handle_frontend_method,
-    BackendHandler,
-)
+from ivy_tests.test_ivy.helpers import handle_frontend_method, BackendHandler
 from ivy_tests.test_ivy.test_frontends.test_tensorflow.test_raw_ops import (
     _pow_helper_shared_dtype,
 )
--- a/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
+++ b/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
@@ -109,13 +109,21 @@ def _helper_random_tensorarray(draw, fn=None):
     element_shape = draw(helpers.get_shape())
     element_shape = draw(st.one_of(st.just(None), st.just(element_shape)))
     shape = None
-    if infer_shape or element_shape is not None or fn in ["scatter", "stack", "gather"]:
-        if element_shape is None:
+    if (
+        infer_shape
+        or element_shape is not None
+        or fn in ["scatter", "stack", "gather", "concat"]
+    ):
+        if fn == "concat":
+            element_shape = None
+            infer_shape = False
+            shape = list(draw(helpers.get_shape(min_num_dims=1)))
+        elif element_shape is None:
             shape = draw(helpers.get_shape())
         else:
             shape = element_shape
     dtype = draw(helpers.get_dtypes(full=False, prune_function=False))[0]
-    if fn == "stack":
+    if fn in ["stack", "concat"]:
         ids_to_write = [True for i in range(size)]
     else:
         ids_to_write = [draw(st.booleans()) for i in range(size)]
@@ -129,6 +137,8 @@ def _helper_random_tensorarray(draw, fn=None):
     }
     id_write = []
     for id, flag in enumerate(ids_to_write):
+        if fn == "concat":
+            shape[0] = draw(st.integers(1, 10))
         if flag:
             write = np.array(
                 draw(
@@ -1745,6 +1755,19 @@ def test_tesorarray_close(
     ta.close()
     assert np.array(ta_frontend.size()) == 0
     assert np.array(ta.size()) == 0
+
+
+@given(l_kwargs=_helper_random_tensorarray(fn="concat"))
+def test_tesorarray_concat(
+    l_kwargs,
+    backend_fw,
+):
+    ta, ta_frontend = _helper_init_tensorarray(backend_fw, l_kwargs)
+    helpers.value_test(
+        ret_np_from_gt_flat=ta.concat().numpy().flatten(),
+        ret_np_flat=np.array(ta_frontend.concat()).flatten(),
+        backend=backend_fw,
+    )
 
 
 @given(l_kwargs=_helper_random_tensorarray())
--- a/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
+++ b/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
@@ -88,6 +88,11 @@ def _helper_init_tensorarray(backend_fw, l_kwargs, fn=None):
         elif fn == "split":
             ta = ta.split(**id_write)
             ta_frontend = ta_frontend.split(**id_write)
+        elif fn == "scatter":
+            indices, value = [*zip(*id_write)]
+            ta = ta.scatter(indices, tf.convert_to_tensor(value, dtype=ta.dtype))
+            value = function_module.stack(list(map(function_module.constant, value)))
+            ta_frontend = ta_frontend.scatter(indices, value)
         else:
             for id, write in id_write:
                 ta = ta.write(id, tf.constant(write))
@@ -104,7 +109,7 @@ def _helper_random_tensorarray(draw, fn=None):
     element_shape = draw(helpers.get_shape())
     element_shape = draw(st.one_of(st.just(None), st.just(element_shape)))
     shape = None
-    if infer_shape or element_shape is not None or fn == "stack":
+    if infer_shape or element_shape is not None or fn in ["scatter", "stack"]:
         if element_shape is None:
             shape = draw(helpers.get_shape())
         else:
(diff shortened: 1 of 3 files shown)


ShreyanshBardia:
Hi @AnnaTz, I have added the class and will now start adding the tests, raised a PR earlier to keep you in the loop and any suggestions you might have

ivy-leaves:

Thank you :hugs:

ShreyanshBardia:
I have added the first test for `dtype` property. We can't go by usual testing pipeline as there are various `write` actions that need to be performed after which we should begin testing. Let me know if you have some ideas on how this can be improved.

AnnaTz:
> I have added the first test for `dtype` property. We can't go by usual testing pipeline as there are various `write` actions that need to be performed after which we should begin testing. Let me know if you have some ideas on how this can be improved.

You are on the right track about testing array properties. We basically have to manually test these as we've done in the rest of the frontends, e.g. [jax](https://github.com/unifyai/ivy/blob/main/ivy_tests/test_ivy/test_frontends/test_jax/test_array.py) [numpy](https://github.com/unifyai/ivy/blob/main/ivy_tests/test_ivy/test_frontends/test_numpy/test_ndarray/test_ndarray.py) [tf](https://github.com/unifyai/ivy/blob/main/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py) [torch](https://github.com/unifyai/ivy/blob/main/ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py#L2161). You can use `@given`, `helpers.value_test` and plain assertions.
The methods however can be tested in the usual way, right?

ShreyanshBardia:
>The methods however can be tested in the usual way, right?

Not really, because if you take a look at [TensorArray](https://www.tensorflow.org/api_docs/python/tf/TensorArray), when we initialise it, it's basically an empty list of `None` on which tensors will be overwritten. So ideally there should be some `write` operations before we start testing the methods. That's why I think we'll have to go by the same way as for the properties

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
```diff
@@ -70,10 +73,266 @@ def _check_query(query):
...
+        )
+
+
+@given(l_kwargs=_helper_random_tensorarray())
+def test_tensorarray_size(
+    l_kwargs,
+    backend_fw,
+):
+    ta, ta_frontend = _helper_init_tensorarray(backend_fw, l_kwargs)
+    id_read, _ = l_kwargs
```
AnnaTz, line 289: I think we don't need `id_read`, right? I don't see it used anywhere in the test.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
```diff
@@ -70,10 +73,266 @@ def _check_query(query):
...
+    l_kwargs,
+    backend_fw,
+):
+    ta, ta_frontend = _helper_init_tensorarray(backend_fw, l_kwargs, "unstack")
+    helpers.value_test(
+        ret_np_from_gt_flat=ta.stack().numpy().flatten(),
+        ret_np_flat=np.array(ta_frontend.stack()).flatten(),
+        backend=backend_fw,
+    )
+
```
AnnaTz, line 334: I know we are already using it everywhere, but still it would be nice to add a dedicated test for `write`.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
```diff
@@ -70,10 +73,266 @@ def _check_query(query):
...
+
```
AnnaTz, line 335: And also tests for the rest of the properties. Would be very simplistic, like the one you wrote for `dtype`.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
```diff
@@ -70,10 +73,266 @@ def _check_query(query):
...
+        if fn == "unstack":
+            ta = ta.unstack(tf.constant(id_write))
+            ta_frontend = ta_frontend.unstack(function_module.constant(id_write))
+        elif fn == "split":
+            ta = ta.split(**id_write)
+            ta_frontend = ta_frontend.split(**id_write)
+        elif fn == "scatter":
+            indices, value = [*zip(*id_write)]
+            ta = ta.scatter(indices, tf.cast(tf.stack(value), dtype=ta.dtype))
+            value = function_module.stack(list(map(function_module.constant, value)))
```
AnnaTz, line 94: Why do we generate `value` differently for our frontend? Since we are doing `tf.cast(tf.stack(value), dtype=ta.dtype)`, I would then expect to do `function_module.cast(function_module.stack(value), dtype=ta.dtype)`.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
```diff
@@ -70,10 +73,266 @@ def _check_query(query):
     )
 
 
+def _helper_init_tensorarray(backend_fw, l_kwargs, fn=None):
+    id_write, kwargs = l_kwargs
+    with BackendHandler.update_backend(backend_fw) as ivy_backend:
+        local_importer = ivy_backend.utils.dynamic_import
+        function_module = local_importer.import_module(
```
AnnaTz, line 80: I think naming this `tf_frontend`, and then naming the tensor array variables `ta` and `ta_gt` would be more comprehensible. But that's subjective I guess.

AnnaTz (changes requested):
Great work. This class and its testing required a very niche approach.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
ShreyanshBardia: Yes that's not required, I had done some refactoring and somehow missed removing it. Thanks for catching it.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
ShreyanshBardia: Yes so actually, `read` and `write` both will have exact same tests, I guess it would be better to add a comment rather than copying the same `read` test or maybe changing the name of test to `test_tensorarray_write_read`

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
ShreyanshBardia: Yes so the reason for doing that was, 
```
import tensorflow as tf
import ivy
import ivy.functional.frontends.tensorflow as tf_frontend
import numpy as np

arr = [np.array([1,2])]*2

tf.stack(arr)  #works

ivy.set_backend("torch")
tf_frontend.stack(arr) #raises error
tf_frontend.stack(list(map(tf_frontend.constant, arr))) # this gives same result
```

ShreyanshBardia: From a functional prespective they both are achieving the same thing, though I guess we can still add `function_module.cast`

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
ShreyanshBardia: Yes I guess `tf_frontend` is more clear and using `gt` rather than `frontend` is more in line with convention being followed in the testing pipeline, will make changes

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
ShreyanshBardia: Done

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
```diff
@@ -70,10 +70,302 @@ def _check_query(query):
...
+    helpers.value_test(
+        ret_np_from_gt_flat=ta_gt.stack().numpy().flatten(),
+        ret_np_flat=np.array(ta.stack()).flatten(),
+        backend=backend_fw,
+    )
+
+
+# test for read and write methods
+@given(l_kwargs=_helper_random_tensorarray())
+def test_tensorarray_write_read(
```
AnnaTz, line 355: Actually we need the test names to follow the format "test_[frontend name]_[function/method name]". This is important for our AI developer team. So, the comment you left will suffice. No need to change to "write_read".

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
```diff
@@ -70,10 +70,302 @@ def _check_query(query):
...
+    l_kwargs,
```
AnnaTz, line 356: I just realized the test names don't follow the correct format. Could you please update them? Basically replacing "tensorarray" with tensorflow".

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_tensor.py
ShreyanshBardia: Looking at the format you mentioned above, I guess it should be `test_tensorflow_tensorarray_read`

AnnaTz (approved):
Great, thanks!

PR #26340, opened 2023-09-29
Camp ione patch 2

Camp-ione:
#fixed func_wrapper.py

## Related Issue

Close #26340

## Checklist

- [ ] Did you add a text? yes

- [ ] Did you add the tests? yes 
- [ ] Did you run your tests and are your tests passing? yes
- [ ] Did pre-commit not fail on any check? yes
- [ ] Did you follow the steps we provided? yes


vedpatwardhan (commented):
Hey @Camp-ione, thanks for creating this PR. Actually in cases where the function has the array typehint in lowercase like in [`numpy`](https://github.com/unifyai/ivy/blob/b9b24b4df8bf51d51974a1a0aa5ed5f69db8140c/ivy/functional/backends/numpy/activations.py#L35) the changes would fail, which is why we were looking for "rray" in the annotation string instead of "Array"
So as a result of the changes made the following example fails,
```
import ivy

ivy.set_backend('numpy')
x = [1., 2., 3.]
print(ivy.gelu(x))
```
I'll close this PR for now as a result, feel free to create an issue if there's other bugs you'd like to report or features you'd like to request 😄

PR #26367, opened 2023-09-30
feat(gradients): add ivy.vjp and ivy.jvp

yopknopixx:
# PR Description

This PR adds support for `ivy.vjp` and `ivy.jvp` per [this Trello task](https://trello.com/c/JJDL7MYZ/1485-implement-ivyjvp-and-ivyjvp).

## Checklist

- [x] Implement `ivy.jvp` for [`jax`, `torch`, `tensorflow`, `mxnet`, `paddle`]
- [x] Implement `ivy.vjp` for [`jax`, `torch`, `tensorflow`]
- [x] Add tests
- [ ] Refactor to support containers

### DIFF
--- a/ivy/functional/backends/torch/experimental/gradients.py
+++ b/ivy/functional/backends/torch/experimental/gradients.py
@@ -1,5 +1,6 @@
 # global
 import torch
+from typing import Callable
 
 # local
 import ivy
@@ -25,3 +26,36 @@ def bind_custom_gradient_function(func, custom_grad_fn):
 
     custom_module = _CustomModule.apply
     return inputs_to_native_arrays(custom_module)
+
+
+def vjp(func: Callable, *primals):
+    def grad_fn(*x_in):
+        return ivy.to_native(
+            func(*ivy.to_ivy(x_in, nested=True)), nested=True, include_derived=True
+        )
+
+    primals_out, _vjpfun = ivy.outputs_to_ivy_arrays(torch.func.vjp)(
+        grad_fn, *ivy.to_native(primals, nested=True)
+    )
+
+    def vjpfun(x_in):
+        return ivy.to_ivy(
+            _vjpfun(ivy.to_native(x_in, nested=True)), nested=True, include_derived=True
+        )
+
+    return (primals_out, vjpfun)
+
+
+def jvp(func: Callable, primals, tangents):
+    def grad_fn(x_in):
+        return ivy.to_native(
+            func(ivy.to_ivy(x_in, nested=True)), nested=True, include_derived=True
+        )
+
+    primals_out, tangents_out = ivy.outputs_to_ivy_arrays(torch.func.jvp)(
+        grad_fn,
+        ivy.to_native(primals, nested=True),
+        ivy.to_native(tangents, nested=True),
+    )
+
+    return (primals_out, tangents_out)
--- a/ivy/functional/backends/jax/experimental/gradients.py
+++ b/ivy/functional/backends/jax/experimental/gradients.py
@@ -1,5 +1,6 @@
 # global
 import jax
+from typing import Callable
 
 # local
 import ivy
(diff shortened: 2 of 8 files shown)


ivy-leaves:

Thank you :hugs:

ivy/functional/backends/torch/experimental/gradients.py
```diff
@@ -25,3 +26,36 @@ def backward(ctx, upstream):
 
     custom_module = _CustomModule.apply
     return inputs_to_native_arrays(custom_module)
+
+
+def vjp(func: Callable, *primals):
```
vedpatwardhan, line 31: For `torch` do you think we could use something like [`torch.autograd.functional.vjp`](https://pytorch.org/docs/stable/generated/torch.autograd.functional.vjp.html)?

vedpatwardhan: same with `torch.autograd.functional.jvp`

vedpatwardhan (changes requested):
Hey @yopknopixx, just requested a minor change, rest seems good to merge. Thanks 😄

ivy/functional/backends/torch/experimental/gradients.py
yopknopixx: That would've been ideal, but `torch.autograd.functional.vjp` requires the `v` (`tangents` in `jax`) argument to be passed for `size` > 1. That doesn't align with the `jax` API where `tangents` is passed to a callback function returned by `vjp`.
We could do an explicit computation for `func_output` then call `vjp` in the callback function if we wanted to call from `torch.autograd.functional`. But that will have us computing `func_output` twice.
What do you think?

vedpatwardhan: actually `jax.vjp` returns a callback whereas `torch.autograd.functional.vjp` returns the product computed. So given that we need to pass `v` to the callback anyway for computing the product, an ideal comparison would be between `jax.vjp(func, x)(v)` with `torch.autograd.functional.vjp(func, x, v)` instead of `jax.vjp(func, x)`. 
Based on this I think both `jax.vjp` and `torch.autograd.functional.vjp` require `v` anyway, so we could just wrap this logic into a function and return that function from `ivy.vjp`
e.g. for the `torch` backend, it could be something like,
```
def vjp(func, *inputs):
    def vjp_func(v):
        return torch.autograd.functional.vjp(func, inputs, v)
    return vjp_func
```
and then for `jax`, we'd just return `jax.vjp(func, *inputs)` if that makes sense?
Happy to clarify further if required, I could be missing something.

yopknopixx: That makes sense, but both `torch` and `jax` return the function outputs when calling `vjp`. So what it would look like in the `torch` backend is:
```
def vjp(func, *inputs):
    outputs = func(*inputs)
    def vjp_func(v):
        outputs, vjp = torch.autograd.functional.vjp(func, inputs, v)
        return vjp
    return (outputs, vjp_func)

```
As you can see, we'd be calculating the outputs twice. If that is fine, we can go ahead with `torch.autograd.functional.vjp`.

vedpatwardhan: Ah yeah you're right, sorry missed that part! What do you think about adding `v` as an argument to `ivy.vjp` then? Probably that's the only way we can get all backends work consistently without doing the computation twice

yopknopixx: That would work for consistency (with the current `paddle` implementation we are computing the outputs twice) but the problem I see with that is this:
When a user wants to do backward mode autodiff, they expect to do something like this:
```
x1, back_x1 = vjp(layer1, x)
x2, back_x2 = vjp(layer2, x1)
# calculate gradients (assuming x2 is scalar)
grad = back_x1(back_x2(tensor(1.)))
```
So, they won't have the `v` argument ready till the next computation is done. This would force them to compute the outputs twice, once during the forward propagation and again when calling `vjp` to compute gradients.

vedpatwardhan: Yeah but if `v` is an argument accepted in the `vjp` function anyway, wouldn't we simply be returning the product instead of a callback? I mean if we have everything we need to compute the product inside the `ivy.vjp` function itself, it wouldn't really be required to return a callback that accepts the `v` (just like `torch.autograd.functional.vjp`). Let me know what you think 😄

yopknopixx: I understand your point, and it makes total sense from `ivy` PoV. But, from user PoV, consider the original use case of `thinc` that needed this function:
```
    def begin_update(self, inputs: ArgsKwargs):
    output, vjp_fn = ivy.vjp(
        lambda v, x: self._model(x, v=v),
        self._model.v,
        inputs.args[0],
    )

    def backprop(external_grads):
        self.param_grads = vjp_fn(external_grads)
        return self.param_grads

    return output, backprop
```
In this case, and many other cases like this, the user won't have the `v` argument ready. And if they have to use `ivy.vjp(func, inputs, v)` it would go something like this:
```
def begin_update(self, inputs: ArgsKwargs):
    output = self._model(*inputs.args, **inputs.kwargs) # 1st computation of outputs

    def backprop(external_grads):
        outputs, self.param_grads = ivy.vjp(self._model, inputs, external_grads) # 2nd computation of outputs
        return self.param_grads

    return output, backprop
```
So, the user would have to calculate the outputs twice. That's why I think having a callback is necessary. Let me know what you think about that.
Regardless, if using `torch.autograd.functional.vjp` is important from the compilation perspective, We can include the `v` argument in `vjp` directly 😊.

vedpatwardhan: Yeah you're right, that makes sense! Given that our gradients API is functional we can't really track operations as they're executed with `torch` as a result of which `torch.autograd.functional.vjp` in `ivy.vjp` by adding the `v` parameter doesn't make sense unless we do some weird wrapping of the product itself (`lambda: product`) or something when calling `ivy.vjp` and then use that callback inside the function defined in your example to avoid computing twice. 
Yeah let's not use `torch.autograd.function.vjp` and mention the reason behind doing that as a comment in the `torch` backend implementation for future reference. 
Thanks for the deep dive @yopknopixx, I'll resolve this conversation now 😄

vedpatwardhan (approved):
lgtm! Feel free to merge, thanks @yopknopixx 😄

PR #26521, opened 2023-10-03
feat(testing): allow testing functions to support dtypes other than targe and …

abdulasiraj:
…gtt backends in frontend and Ivy API tests respectively

# PR Description
The main idea is:

1.  Get all the supported dtypes for the frontend framework function.
2.  As we have to check all the above dtypes, it's passible that these dtypes may not be supported by the targat backends. so, we've to types cast the dtype to the nearest supported dtype if is not supported by target backend.
3.  I'm using Jax data type casting semantics which are safe and predictable (means we always get the dtype or None).
4.  If the dtype is None, we skip that particular example (right now the test case for specific backend, logic needs to be implemented for just skip single example)
5.  things work 98% of time without testing skipping, because the chance is very rare that gt supports `complex128` while the target not. also it doesn't skip the whole test.
6.  reverse cast the dtype for comparison. (not implemented yet)

## Related Issue

Close #

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy_tests/test_ivy/helpers/function_testing.py
+++ b/ivy_tests/test_ivy/helpers/function_testing.py
@@ -680,22 +680,73 @@ def test_frontend_function(
     args_np, kwargs_np = kwargs_to_args_n_kwargs(
         num_positional_args=test_flags.num_positional_args, kwargs=all_as_kwargs_np
     )
-
-    # extract all arrays from the arguments and keyword arguments
-    arg_np_vals, args_idxs, c_arg_vals = _get_nested_np_arrays(args_np)
-    kwarg_np_vals, kwargs_idxs, c_kwarg_vals = _get_nested_np_arrays(kwargs_np)
-    # make all lists equal in length
-    num_arrays = c_arg_vals + c_kwarg_vals
-    if len(input_dtypes) < num_arrays:
-        input_dtypes = [input_dtypes[0] for _ in range(num_arrays)]
-    if len(test_flags.as_variable) < num_arrays:
-        test_flags.as_variable = [test_flags.as_variable[0] for _ in range(num_arrays)]
-    if len(test_flags.native_arrays) < num_arrays:
-        test_flags.native_arrays = [
-            test_flags.native_arrays[0] for _ in range(num_arrays)
-        ]
+    # list and dict
+    # have to do type cast stuff here
 
     with BackendHandler.update_backend(backend_to_test) as ivy_backend:
+        target_args_np = copy.deepcopy(args_np)
+        target_kwargs_np = copy.deepcopy(kwargs_np)
+        backend_supported_dtypes = (
+            t_globals.CURRENT_RUNNING_TEST.supported_device_dtypes[backend_to_test][
+                on_device
+            ]
+        )
+        if input_dtypes[0] not in backend_supported_dtypes["valid"]:
+            # cast dtype to some supported dtype
+            temp = get_nearest_castable_dtype(
+                input_dtypes[0], backend_supported_dtypes["valid"]
+            )
+            if temp is not None:
+                input_dtypes[:] = [temp] * len(temp)
+                # change dtype for args_for_test and kwargs_for_test
+                target_args_np = [
+                    (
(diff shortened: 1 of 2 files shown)


ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -678,22 +680,62 @@ def test_frontend_function(
...
 
     with BackendHandler.update_backend(backend_to_test) as ivy_backend:
+        target_args_np = copy.deepcopy(args_np)
+        target_kwargs_np = copy.deepcopy(kwargs_np)
+        backend_supported_dtypes = (
+            t_globals.CURRENT_RUNNING_TEST.supported_device_dtypes[backend_to_test][
+                on_device
+            ]
+        )
+        if input_dtypes[0] not in backend_supported_dtypes["valid"]:
```
vedpatwardhan, line 694: I don't think we should be having a block for this as discussed, it's much better to just set the casting modes and the input would be casted to a necessary dtype

ivy_tests/test_ivy/helpers/hypothesis_helpers/dtype_helpers.py
```diff
@@ -228,20 +246,32 @@ def get_dtypes(
     # FN_DTYPES & BACKEND_DTYPES & FRONTEND_DTYPES & GROUND_TRUTH_DTYPES
 
     # If being called from a frontend test
-    if test_globals.CURRENT_FRONTEND is not test_globals._Notsetval:
-        frontend_dtypes = _get_type_dict_helper(
-            test_globals.CURRENT_FRONTEND, kind, True
-        )
-        valid_dtypes = valid_dtypes.intersection(frontend_dtypes)
+    # if test_globals.CURRENT_FRONTEND is not test_globals._Notsetval:
```
vedpatwardhan, line 249: ig we should either remove or uncomment this

ivy_tests/test_ivy/helpers/hypothesis_helpers/dtype_helpers.py
```diff
@@ -448,3 +478,154 @@ def bound_dtype_bits(d):
             )
             and (min_x > 0 or not ivy_backend.is_uint_dtype(dtype))
         )
+
+
+possible_castable_dtypes = {
```
vedpatwardhan, line 483: as mentioned before, we shouldn't need this code as the casting modes should be able to cover this behaviour. If there's issues with the casting modes, we should be fixing the casting modes instead (cc @RickSanchezStoic)

vedpatwardhan (changes requested):
Hey @abdulasiraj, just requested a few changes. Thanks @abdulasiraj 😄

ivy_tests/test_ivy/helpers/hypothesis_helpers/dtype_helpers.py
abdulasiraj: Hey @vedpatwardhan  I discussed it with @RickSanchezStoic. he suggested me to use different logic. Also, there's couple of restrictions with casting modes predictability but here, we even need the castable dtype for generating data arrays. like If the casting mode will handle the dtype under the hood than how we can restrict data to that dtype, we don't know which dtype it'll casted to. I think if @RickSanchezStoic confirm this thing again here, It'll be helpful. thanks!

vedpatwardhan: Happy to know what you think @RickSanchezStoic, but I think if we still need to duplicate the casting logic such that we do the casting before passing the input to the function instead of using the casting modes kinda defeats the purpose of the casting modes because then we won't be able to test if the casting modes are working as expected. I think fixing the casting modes to capture all cases which lead to this duplication should be the way to go 😅

vedpatwardhan:
As mentioned in one of the previous reviews, we'd only need to enable the casting modes at the start of the tests and ideally that should work out of the box. As a result, I'll close this PR for now and get that working once the testing refactor is completed. Thanks everyone 😄

PR #26572, opened 2023-10-04
refactor: Extract `Duplicate Code` into Helper Function in `ivy\func_wrapper.py`

Sai-Suraj-27:
# PR Description
I identified duplicate sections of code in the following functions:
https://github.com/unifyai/ivy/blob/2bfb19a2ca3b277e95cba987353aa349ef28015e/ivy/func_wrapper.py#L94
https://github.com/unifyai/ivy/blob/2bfb19a2ca3b277e95cba987353aa349ef28015e/ivy/func_wrapper.py#L137

So, I extracted these into their own method/function to increase the readability and decrease the amount of duplicate/repeating code in the file. This refactoring simplifies the code, and reduces redundancy.

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials

### DIFF
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -94,88 +94,47 @@ def caster(dtype, intersect):
 def upcaster(dtype, intersect):
     # upcasting is enabled, we upcast to the highest
     if "uint" in str(dtype):
-        index = casting_modes_dict["uint"]().index(dtype) + 1
-        result = ""
-        while index < len(casting_modes_dict["uint"]()):
-            if casting_modes_dict["uint"]()[index] not in intersect:
-                result = casting_modes_dict["uint"]()[index]
-                break
-            index += 1
-        return result
-
+        return upcast_helper("uint", dtype, intersect)
     if "int" in dtype:
-        index = casting_modes_dict["int"]().index(dtype) + 1
-        result = ""
-        while index < len(casting_modes_dict["int"]()):
-            if casting_modes_dict["int"]()[index] not in intersect:
-                result = casting_modes_dict["int"]()[index]
-                break
-            index += 1
-        return result
-
+        return upcast_helper("int", dtype, intersect)
     if "float" in dtype:
-        index = casting_modes_dict["float"]().index(dtype) + 1
-        result = ""
-        while index < len(casting_modes_dict["float"]()):
-            if casting_modes_dict["float"]()[index] not in intersect:
-                result = casting_modes_dict["float"]()[index]
-                break
-            index += 1
-        return result
-
+        return upcast_helper("float", dtype, intersect)
     if "complex" in dtype:
-        index = casting_modes_dict["complex"]().index(dtype) + 1
-        result = ""
-        while index < len(casting_modes_dict["complex"]()):
-            if casting_modes_dict["complex"]()[index] not in intersect:
-                result = casting_modes_dict["complex"]()[index]
-                break
-            index += 1
-        return result
+        return upcast_helper("complex", dtype, intersect)
+
+
+def upcast_helper(arg0, dtype, intersect):
+    index = casting_modes_dict[arg0]().index(dtype) + 1
+    result = ""
+    while index < len(casting_modes_dict[arg0]()):
+        if casting_modes_dict[arg0]()[index] not in intersect:
+            result = casting_modes_dict[arg0]()[index]
+            break
+        index += 1
+    return result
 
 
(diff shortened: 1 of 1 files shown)


vedpatwardhan (commented):
The changes look good to me, but I'll request @RickSanchezStoic's review just in case this breaks the casting modes. Thanks @Sai-Suraj-27 😄

ivy/func_wrapper.py
```diff
@@ -94,88 +94,47 @@ def caster(dtype, intersect):
...
-        while index >= 0:
-            if casting_modes_dict["complex"]()[index] not in intersect:
-                result = casting_modes_dict["complex"]()[index]
-                break
-            index -= 1
-        return result
+        return downcast_helper("complex", dtype, "complex", intersect)
+
+
+def downcast_helper(arg0, dtype, arg2, intersect):
```
RickSanchezStoic, line 129: let's change this to `arg1`

ivy/func_wrapper.py
```diff
@@ -94,88 +94,47 @@ def caster(dtype, intersect):
...
+    index = casting_modes_dict[arg0]().index(dtype) - 1
```
RickSanchezStoic, line 130: same here

ivy/func_wrapper.py
Sai-Suraj-27: @RickSanchezStoic sir, I made the requested changes. Please check now, Thank you.

ivy/func_wrapper.py
Sai-Suraj-27: Done!

Sai-Suraj-27:
@RickSanchezStoic sir, its been a week since I made the requested changes. Please look into the PR when you are free, Thank you.

vedpatwardhan:
Hey @Sai-Suraj-27, can we refactor this further to have a single helper which receives a flag about whether to use the upcast or downcast logic? Seems like the `>=0` and `<len(casting_modes_dict[arg0]())` condition in the while loop and the `index +=1` and `index -= 1` are the only 2 things different in the 2 helpers. What do you think? Thanks 😄

vedpatwardhan:
I think that's just a mistake, can't really think of a reason why it would be `int` instead of `uint` on 
https://github.com/unifyai/ivy/blob/19cb1a48f27cf4b3ed27b07fb81b7d7a10d52fba/ivy/func_wrapper.py#L143
Good catch @Sai-Suraj-27 😄

Sai-Suraj-27:
@vedpatwardhan Sorry for the late response, I have refactored further to use only a single helper function. Please check now.

vedpatwardhan:
lgtm! I'll get this PR merged now, thanks @Sai-Suraj-27 😄

ivy/func_wrapper.py
```diff
@@ -91,91 +91,48 @@ def caster(dtype, intersect):
                 return ret_dtype
 
 
+def cast_helper(arg, dtype, intersect, is_upcast=True):
+    step = 1 if is_upcast else -1
+    index = casting_modes_dict[arg]().index(dtype) + step
+    result = ""
+    if is_upcast:
+        while index < len(casting_modes_dict[arg]()):
```
vedpatwardhan, line 99: ig we could replace this with a single loop instead of 2 where the condition would change based on the value of `is_upcast`. What do you think?

Sai-Suraj-27: I thought of the same but I wasn't sure if `index` will be always >= 0. If the index value is always >= 0, then we can do the following:
```Python
result = ""
while 0 <= index < len(casting_modes_dict[t]()):
   if casting_modes_dict[t]()[index] not in intersect:
       result = casting_modes_dict[t]()[index]
       break
   index += index_step
return result
```
Please tell me your view, Thank you.

vedpatwardhan: I think it's safe to assume that `index` will always be `>=0` 😄

vedpatwardhan (approved):
lgtm! Feel free to merge the PR @Dsantra92, thanks @Sai-Suraj-27 😄

Sai-Suraj-27:
Hey, @Dsantra92 this PR got approval almost a month back, can you please merge this, Thank you.

Sai-Suraj-27:
Hey, @vedpatwardhan You approved this PR almost a month back but no one merged it. Can you please merge this, thank you.

vedpatwardhan:
Yep merging the PR now, thanks @Sai-Suraj-27 😄

PR #26718, opened 2023-10-06
docs: add failing test docs

sherry30:
# PR Description

## Related Issue

Close #

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?


a0m0rajab (commented):
Thank you for this Sherry,  I just have few thoughts to share, no need to apply them now since It will be needed to go through the guide again next few days: 

- the walkthrough which explained everything and showed an example is really great! 
- When I used our doc building pipeline, I could not see the page on the menu 
- Could we have a docker-free instructions? sometimes docker could be a bit heavy for some machines.
- it would be great if we could have beginner friendly and experts friendly guide at the same time, how to do that? we could hide beginners details or link the original related docs. 
- some formatting would be great, I will get into those in details later. 

just wanted to keep this comments right now and will visit it later.

PR #25717, opened 2023-09-20
feat: Added the `isort` hook and modified all the files accordingly.

Sai-Suraj-27:
# PR Description
Updated the `.pre-commit-config.yaml` file by adding one more check (with automated fix). I found that this check is very common and used in most of the big `Python` based organizations including [scrapy](https://github.com/scrapy/scrapy/blob/master/.pre-commit-config.yaml), [openai/evals](https://github.com/openai/evals/blob/main/.pre-commit-config.yaml) ,[dask](https://github.com/dask/dask/blob/main/.pre-commit-config.yaml), [jina](https://github.com/jina-ai/jina/blob/master/.pre-commit-config.yaml), [mindsbd](https://github.com/Sai-Suraj-27/mindsdb/blob/staging/.pre-commit-config.yaml), [pandas](https://github.com/Sai-Suraj-27/pandas/blob/main/.pre-commit-config.yaml#L72), etc. So, it will be nice to have it in our codebase.

### Added the following pre-commit hook
1. [`isort`](https://github.com/PyCQA/isort)
 
**Benefits**
It automatically rearranges imports in a Python file based on the specified rules and settings to ensure consistent and standardized import organization.

Additionally, `isort` hook follows specific guidelines for line length and line wrapping to keep imports within a specified character limit, promoting readability and maintainability. By following these rules, `isort` helps maintain a clean and standardized import layout across your Python codebase, making it easier to read, and understand.

## Related Issue
Close #25714 

## Checklist
- [ ] Did you add a function?
- [ ] Did you add the tests?
- [ ] Did you run your tests and are your tests passing?
- [x] Did pre-commit not fail on any check?
- [ ] Did you follow the steps we provided?

### Socials: 
https://twitter.com/Sai_Suraj_27

PR #26723, opened 2023-10-07
Add(formatter): Add reorder-python-imports to pre-commit hooks

NripeshN:


KareemMAX (commented):
Hey @NripeshN,
This looks good, but I'm afraid of some complications that might occur. This PR is similar to https://github.com/unifyai/ivy/pull/25717.

After we merged we discovered that it was incompatible with our linting suite, as the linting was always failing due to conflicts.

Also another concern is `reorder-python-imports` is not as popular as `isort`, so I'm afraid that it doesn't become as supported as time goes by.

vedpatwardhan:
Also, not sure if it's related but just came across [this](https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement#comment128571625_58804315)

PR #26737, opened 2023-10-07
fix: Update module converters to correctly handle tuple outputs in their `_forward` methods

hmahmood24:
Fixes an issue where if a module has a `tuple` instance as the output of its forward call, we don't unnecessarily append an extra `{}` to the return statement by doing `return ivy.args_to_native(*ret)` even though we aren't passing any mapping as a second argument to `args_to_native`, the return has a `{}` appended to it which is expected since python automatically creates a `{}` for the `**kwargs` passed to a function. See the example below:
```python
def dummy_args_to_native(*args, abc=None, **kwargs):
    return args, kwargs

# Call `dummy_args_to_native` but don't pass in any mapping as the second argument
dummy_args_to_native(*(1, 2, 3))  # ((1, 2, 3), {})
dummy_args_to_native(*(1, 2, 3))[0]  # (1, 2, 3)
```
This fixes tracing the `torch.nn.LSTM` module as well.


ivy/stateful/module.py
```diff
@@ -1079,7 +1079,7 @@ def _forward(self, *a, **kw):
         params_hk = self._dict_to_hk_flat_map(self.v.cont_to_dict())
         ret = self._native_module.apply(params_hk, 0, *a, **kw)
         if isinstance(ret, tuple):
-            return ivy.args_to_native(*ret)
+            return ivy.args_to_native(*ret)[0]
```
vedpatwardhan, line 1082: just out of curiosity, why are we even checking if `ret` is a `tuple` and using `args_to_native` if that's the case? Ideally `ivy.to_native` should be able to correctly convert arrays in the tuple to native,
```
import ivy

ivy.set_backend("numpy")
x = ivy.array([1., 2., 3.])
print(ivy.to_native((x,), nested=True))
```

vedpatwardhan (commented):
The changes look good, just wanted to see if we can remove that `if` block entirely. Thanks @hmahmood24 😄

ivy/stateful/module.py
hmahmood24: @vedpatwardhan Yeah I mean I didn't add that logic 😅 Although my thinking was the same on first glace. Probably look at the line history to see who authored this originally

RickSanchezStoic:
Looks good to me too, however since these classes were originally by @YushaArif99 (not sure) , so best to request a review from him too @hmahmood24

ivy/stateful/module.py
YushaArif99: It's been quite some time since I last reviewed this code, and I don't recall every minor detail. However, upon examining the implementation, it appears to be an extraneous addition. As @vedpatwardhan suggested, we can effectively substitute this logic with `ivy.to_native(ret, nested=True)`.

YushaArif99 (approved):
lgtm! I think we are safe to merge if the converter tests pass 👍🏼 

PS: Looks like our CI tests are on vacation! 😂 They're not running the right tests. Could you do me a favor and check if the tests in [test_converters](https://github.com/unifyai/ivy/blob/main/ivy_tests/test_ivy/test_stateful/test_converters.py) are passing @hmahmood24?

hmahmood24:
> lgtm! I think we are safe to merge if the converter tests pass 👍🏼
> 
> PS: Looks like our CI tests are on vacation!😂. Could you do me a favor and check if the tests in [test_converters](https://github.com/unifyai/ivy/blob/main/ivy_tests/test_ivy/test_stateful/test_converters.py) are passing @hmahmood24?

I am having [trouble](https://discordapp.com/channels/799879767196958751/1028303853315162112/1160960097195921479) setting up the testing in my codespace since a week myself 😅 Can any one of you guys test and see if the tests run fine after these changes? @YushaArif99 @vedpatwardhan @RickSanchezStoic

HaiderSultanArc:
> > lgtm! I think we are safe to merge if the converter tests pass 👍🏼
> > PS: Looks like our CI tests are on vacation!😂. Could you do me a favor and check if the tests in [test_converters](https://github.com/unifyai/ivy/blob/main/ivy_tests/test_ivy/test_stateful/test_converters.py) are passing @hmahmood24?
> 
> I am having [trouble](https://discordapp.com/channels/799879767196958751/1028303853315162112/1160960097195921479) setting up the testing in my codespace since a week myself 😅 Can any one of you guys test and see if the tests run fine after these changes? @YushaArif99 @vedpatwardhan @RickSanchezStoic

I ran the tests locally. All the tests in [test_converters](https://github.com/unifyai/ivy/blob/main/ivy_tests/test_ivy/test_stateful/test_converters.py) are failing with `KeyError: ''` both on main and PR. 🤔

ivy/stateful/module.py
vedpatwardhan: Sounds good! Thanks for making the changes @hmahmood24 and the helpful suggestions @RickSanchezStoic @YushaArif99. I'll resolve this now

vedpatwardhan (approved):
lgtm! Thanks @hmahmood24 😄
(I've fixed the converters tests on the repo `main` so I'd suggest pulling from `main`, running the converters tests and verifying if everything continues to pass before merging)

hmahmood24:
Just re-ran the tests locally after setting up the codespace again and they now seem to run fine with the changes. Thanks for the help @vedpatwardhan 😄

PR #26746, opened 2023-10-07
fix: Replace use of `default mutable arguments` in a function.

### LINKED ISSUE #26745: Replace use of `default mutable arguments` in a function.
In some of the files default mutable arguments are being used they should be replaced properly to avoid unintended behaviour.
Examples:
In file-1:
https://github.com/unifyai/ivy/blob/373b033695b5413045aa58bf9d841567b8bbfee8/ivy/functional/backends/jax/gradients.py#L77
https://github.com/unifyai/ivy/blob/373b033695b5413045aa58bf9d841567b8bbfee8/ivy/functional/backends/jax/gradients.py#L78
In another file:
https://github.com/unifyai/ivy/blob/373b033695b5413045aa58bf9d841567b8bbfee8/ivy/functional/backends/mxnet/gradients.py#L29
https://github.com/unifyai/ivy/blob/373b033695b5413045aa58bf9d841567b8bbfee8/ivy/functional/backends/mxnet/gradients.py#L30

Sai-Suraj-27:
# PR Description
Default argument values which are mutable are `evaluated only once at function definition time`, so only one list, set or dictionary instance will be created. This means that if this list is mutated in one call to a function, those changes will show up in subsequent calls of that function (`With a mutable default value, with each call the default value is modified`), This is usually `unintended behaviour`.

https://github.com/unifyai/ivy/blob/373b033695b5413045aa58bf9d841567b8bbfee8/ivy/functional/backends/jax/gradients.py#L77
https://github.com/unifyai/ivy/blob/373b033695b5413045aa58bf9d841567b8bbfee8/ivy/functional/backends/jax/gradients.py#L78

In some of the files default mutable arguments are being used they should be replaced properly to avoid unintended behaviour. I found most of the places where this is happening and adjusted the functions properly to avoid this.

## Related Issue
Close #26745

## Checklist
- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials


sherry30:
I'll leave this to @vedpatwardhan since its a change in the gradients file

vedpatwardhan (changes requested):
Hey @Sai-Suraj-27, I don't really think we should be making this change, the default argument approach seems better to me instead of the conditional ([reference](https://github.com/zedr/clean-code-python#use-default-arguments-instead-of-short-circuiting-or-conditionals)). For the mutable default argument issue, let's probable make it a `tuple` instead of a `list`. Thanks for creating the PR though, I'd suggest instead removing the `Optional` typehint from `xs_grad_idxs` and `ret_grad_idxs` because any argument with a non-`None` default value shouldn't be typed as `Optional`. We can get that change merged instead if that works for you 😄

Sai-Suraj-27:
> Hey @Sai-Suraj-27, I don't really think we should be making this change, the default argument approach seems better to me instead of the conditional ([reference](https://github.com/zedr/clean-code-python#use-default-arguments-instead-of-short-circuiting-or-conditionals)). For the mutable default argument issue, let's probable make it a `tuple` instead of a `list`. Thanks for creating the PR though, I'd suggest instead removing the `Optional` typehint from `xs_grad_idxs` and `ret_grad_idxs` because any argument with a non-`None` default value shouldn't be typed as `Optional`. We can get that change merged instead if that works for you 😄

@vedpatwardhan sure👍, I have made the requested changes. Please take a look now, Thank you.

vedpatwardhan:
> > Hey @Sai-Suraj-27, I don't really think we should be making this change, the default argument approach seems better to me instead of the conditional ([reference](https://github.com/zedr/clean-code-python#use-default-arguments-instead-of-short-circuiting-or-conditionals)). For the mutable default argument issue, let's probable make it a `tuple` instead of a `list`. Thanks for creating the PR though, I'd suggest instead removing the `Optional` typehint from `xs_grad_idxs` and `ret_grad_idxs` because any argument with a non-`None` default value shouldn't be typed as `Optional`. We can get that change merged instead if that works for you 😄
> 
> @vedpatwardhan sure👍, I have made the requested changes. Please take a look now, Thank you.

Hey @Sai-Suraj-27, actually you'd need to replace `((0))` by `((0,),)` as `((0))` is essentially an `int`. Similarly in the assertions, you'd need to replace `(0)` by `(0,)`. Thanks 😄

Sai-Suraj-27:
> Hey @Sai-Suraj-27, actually you'd need to replace `((0))` by `((0,),)` as `((0))` is essentially an `int`. Similarly in the assertions, you'd need to replace `(0)` by `(0,)`. Thanks 😄

@vedpatwardhan Yup sorry for the mistake. I have replaced `((0))` by `((0,),)` in all the files please check now, Thank you.

ivy/functional/backends/jax/gradients.py
```diff
@@ -74,8 +74,8 @@ def execute_with_gradients(
     /,
     *,
     retain_grads: bool = False,
-    xs_grad_idxs: Optional[Sequence[Sequence[Union[str, int]]]] = [[0]],
-    ret_grad_idxs: Optional[Sequence[Sequence[Union[str, int]]]] = [[0]],
+    xs_grad_idxs: Sequence[Sequence[Union[str, int]]] = ((0,),),
```
vedpatwardhan, line 77: could you please also update line [51](https://github.com/unifyai/ivy/pull/26746/files#diff-2ad043743db916943bf7613c8da8747238a2cd3b9682557ee9b30408360c097dR51) to `xs_grad_arr_idxs.append(list(grad_idx) + idx)`? After the `tuple` change, this change is also required. Thanks @Sai-Suraj-27 😄

vedpatwardhan (changes requested):

ivy/functional/backends/jax/gradients.py
Sai-Suraj-27: Done!

vedpatwardhan: thanks!

vedpatwardhan (approved):
lgtm! Thanks @Sai-Suraj-27 😄

PR #26751, opened 2023-10-07
Add rfft2 to numpy frontend

### LINKED ISSUE #26706: rfft2

samir-fama:
close #26706

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_numpy/test_fft/test_discrete_fourier_transform.py
+++ b/ivy_tests/test_ivy/test_frontends/test_numpy/test_fft/test_discrete_fourier_transform.py
@@ -254,3 +254,42 @@ def test_numpy_rfftn(dtype_and_x, frontend, backend_fw, test_flags, fn_tree, on_
         axes=axes,
         norm=norm,
     )
+
+
+@handle_frontend_test(
+    fn_tree="numpy.fft.rfft2",
+    dtype_values=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("complex"),
+        num_arrays=1,
+        min_value=-1e5,
+        max_value=1e5,
+        min_num_dims=2,
+        max_num_dims=5,
+        min_dim_size=2,
+        max_dim_size=5,
+        allow_inf=False,
+        large_abs_safety_factor=2.5,
+        small_abs_safety_factor=2.5,
+        safety_factor_scale="log",
+    ),
+    axes=st.sampled_from([(0, 1), (-1, -2), (1, 0)]),
+    s=st.tuples(
+        st.integers(min_value=2, max_value=256), st.integers(min_value=2, max_value=256)
+    ),
+    norm=st.sampled_from(["backward", "ortho", "forward", None]),
+)
+def test_numpy_rfft2(dtype_values, axes, s, norm, frontend, backend_fw, test_flags, fn_tree, on_device):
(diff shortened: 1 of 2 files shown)


ivy_tests/test_ivy/test_frontends/test_numpy/test_fft/test_discrete_fourier_transform.py
```diff
@@ -213,6 +213,47 @@ def test_numpy_rfft(
     )
 
 
+@handle_frontend_test(
+    fn_tree="numpy.fft.rfft2",
+    dtype_values=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("complex"),
```
YushaArif99, line 219: `numpy.rfftn` expects real-value arrays. why are you creating complex arrays as inputs?

ivy_tests/test_ivy/test_frontends/test_numpy/test_fft/test_discrete_fourier_transform.py
```diff
@@ -213,6 +213,47 @@ def test_numpy_rfft(
...
+        min_num_dims=2,
+        max_num_dims=5,
+        min_dim_size=2,
+        max_dim_size=5,
+        allow_inf=False,
+        large_abs_safety_factor=2.5,
+        small_abs_safety_factor=2.5,
+        safety_factor_scale="log",
+    ),
+    axes=st.sampled_from([(0, 1), (-1, -2), (1, 0)]),
```
YushaArif99, line 232: rather than hard-coding a few `axes` values, it's better to make generate `axes` values using hypothesis strategies.

ivy_tests/test_ivy/test_frontends/test_numpy/test_fft/test_discrete_fourier_transform.py
```diff
@@ -213,6 +213,47 @@ def test_numpy_rfft(
...
+    s=st.tuples(
```
YushaArif99, line 233: `s` shouldnt always be a tuple of length 2. Instead it should be either `None`, or a tuple of length of `dim(a)`

YushaArif99 (changes requested):
Hi @ShreKrente 

The implementation looks good overall. The data generation strategy though seems a little restrictive to me and could potentially be better generalized. For instance, you could generate an n-dim array `a` based on which you could define a new helper which would then correspondingly generate valid values for `s` and `axes`. 
Here is a possible implementation for the helper: 

```python
def generate_s_and_axes(a):
    ndim = a.ndim  

    # Strategy for generating 's'
    s_strategy = st.one_of(
        st.none(),  # 's' can be None
        st.tuples(*(st.integers(min_value=0, max_value=a.shape[i]) for i in range(ndim))) # Each element of 's' is an integer between 0 and the size of the corresponding dimension of 'a'
    )

    # Strategy for generating 'axes'
    axes_strategy = st.one_of(
        st.none(),  # 'axes' can be None
        st.lists(
            st.integers(min_value=0, max_value=ndim-1), 
            min_size=1, max_size=ndim  
        ) #Each element of 'axes' is an integer between 0 and the number of dimensions of 'a' - 1
    )

    return s_strategy, axes_strategy
```

Sarvesh-Kesharwani:
Closing the PR for the time being, since there has been no activity for more than 2 weeks.

PR #26839, opened 2023-10-09
feat: added function matrix_power to frontend paddle tensor

### LINKED ISSUE #26431: matrix_power

binael:
# PR Description

Created the function matrix_power for paddle frontend tensor. This function takes two values, a square matrix of float dtype and the power to raise the matrix, having integer dtype

## Related Issue
None

Close #26431

## Checklist

- [x] Did you add a function?
- [x] Did you add the tests?

### Socials
twitter.com/binaelnchekwube

### DIFF
--- a/ivy/functional/frontends/paddle/tensor/tensor.py
+++ b/ivy/functional/frontends/paddle/tensor/tensor.py
@@ -908,3 +908,7 @@ class Tensor:
     def gather_(self, y, name=None):
         res = self.gather(self, y)
         return ivy.inplace_update(self, res)
+
+    @with_unsupported_dtypes({"2.5.1 and below": ("float16", "bfloat16")}, "paddle")
+    def matrix_power(self, n, name=None):
+        return paddle_frontend.matrix_power(self, n)
--- a/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_tensor.py
+++ b/ivy_tests/test_ivy/test_frontends/test_paddle/test_tensor/test_tensor.py
@@ -3042,6 +3042,44 @@ def test_paddle_tensor_logical_xor(
     )
 
 
+# matrix_power
+@handle_frontend_method(
+    class_tree=CLASS_TREE,
+    init_tree="paddle.to_tensor",
+    method_name="matrix_power",
+    dtype_and_x=helpers.dtype_and_values(
+        available_dtypes=helpers.get_dtypes("float"),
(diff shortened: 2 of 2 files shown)


binael:
> Hi @binael, thank you for your contribution! This looks good overall, could you make sure the tests pass locally as well, the intelligent tests are currently unstable and in the process of being fixed. Let me know when you'd like another review, thanks!

Hi @nassimberrada 
Thank you for your response.
I would like another review. I have implemented the tests and found that the test only supports dtypes of "float32". With this the all test (5) passed
I will like to effect this correction

This is my second PR, the first has been merged. 
Trying the first PR now, I saw it passes sometimes and fails sometimes due to the dtypes (supports only float32 and float64)
Can I also correct this data type on the first PR which has been merged?

binael:
> I'm not sure I fully follow, are both PRs for the same function and you've opened this second PR to rectify the first ? Also, from your screenshot it seems you're allowing the search to be performed on both `float32` and `float64` but you say the tests only pass on `float32`, am I missing something ? Either way, if you would like to adjust an issue with your merged PR, feel free to make a PR that specifically makes that change, explaining the purpose and the rationale behind it. Thanks!



> I'm not sure I fully follow, are both PRs for the same function and you've opened this second PR to rectify the first ? Also, from your screenshot it seems you're allowing the search to be performed on both `float32` and `float64` but you say the tests only pass on `float32`, am I missing something ? Either way, if you would like to adjust an issue with your merged PR, feel free to make a PR that specifically makes that change, explaining the purpose and the rationale behind it. Thanks!

Oh, Sorry for the confusing statements.
You answered my question perfectly with your last comment. 
The screenshot I previously posted was for the already merged PR for bmm function.
I have now added the specified dtype in this. The screenshot is below

**Please, you can review now**

Screenshot 2023-10-14 161243

binael:
> Gotcha, the tests pass for `float32` but I believe the [native function](https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/linalg/matrix_power_en.html) should be able to take `float64` as well. Do you know why the tests fail for `float64` ? Is it an issue with the backend implementation and can you mitigate the issue within the frontend ? Otherwise the tests look good but since we want the frontend to be as faithful to the native function as possible we probably want to tackle the `float64` issue, hope this makes sense. Thanks!

IMG-20231019-WA0013
Float64 giving the above Error
I am looking at the backend to know where the fault is coming from

PR #26853, opened 2023-10-10
add(pre-commit hook): mypy

NripeshN:


KareemMAX (commented):
Have you run this check?

PR #26860, opened 2023-10-10
fix(test): changes to test_trace and the binaries section in the docs

vedpatwardhan:
1. changed the test_trace value to be generated through a strategy instead of being hardcoded to False
2. updated the binaries section in the docs asking people to use --skip-trace-testing when running the tests for their open tasks

Let me know what you think @sherry30, unless there's a specific reason to disable the tracer tests by default, we should always run the tracer tests on every ivy or frontend test. 
Thanks 😄


vedpatwardhan:
> Hey Ved, We disabled it because there was no way of checking if the user has key.pem which is required for tracing functions so it threw errors instead of moving on. I think @Aarsh2001 requested the tracer team to add a function to validate if the key exists so we can skip in the case it doesn't. I'm not sure if it has been added yet. For now I guess we can go with this and update it later when we have that validation function. Maybe you can add a ToDo here

Yeah I remember, something to do with a `validate` function, I'll get in touch with everyone to get this done, let's add a `ToDo` until then. Thanks @sherry30 😄

CatB1t (commented):
Agree with @sherry30, mainly we should have the tests in a state that is running for everyone, I'd suggest we can have a simple configuration check in `conftest` to check if tracer is configured properly, if so the flag will be set to True.

PR #26867, opened 2023-10-10
fix: Add missing job name for `.github\workflows\binaries.yml`

### LINKED ISSUE #26856: Missing `job name` in the `.github\workflows\binaries.yml` file
Missing job name for the `.github\workflows\binaries.yml` Workflow. The workflow is missing a name field, which is important as it is good to have a meaningful and descriptive name for the workflow.

https://github.com/unifyai/ivy/blob/8608b7f9cda83f78b3409eaa33a47701c76bfa21/.github/workflows/binaries.yml#L4
I'm not sure but a name like `release_binaries:` or `tag_binary_release:` could be added.

```YAML
jobs:
  release_binaries:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout 🛎️Binaries
        uses: actions/checkout@v2
        with:
          repository: unifyai/binaries
          path: binaries
          persist-credentials: false

      - name: Add Tag to Binaries
        run: |
          cd binaries
          git tag ${{ github.ref_name }}
          git push origin ${{ github.ref_name }}
```

Sai-Suraj-27:
# PR Description
Added the missing job name for the `.github\workflows\binaries.yml` Workflow. 
https://github.com/unifyai/ivy/blob/8608b7f9cda83f78b3409eaa33a47701c76bfa21/.github/workflows/binaries.yml#L4
I'm not sure about the name but I guess something like `release_binaries:` or `tag_binary_release:` could be added.

## Related Issue
Closes #26856

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials


Sai-Suraj-27:
@vedpatwardhan sir, I have made the PR adding the missing job name (#26867). Please look into this PR when you are free, If you think there should be a better `job name` other than `release_binaries:` please tell me here, I will modify accordingly, Thank you.

.github/workflows/binaries.yml
```diff
@@ -2,17 +2,18 @@ name: pypi
...
 jobs:
-  runs-on: ubuntu-latest
-  steps:
-    - name: Checkout 🛎️Binaries
-      uses: actions/checkout@v2
-      with:
-        repository: unifyai/binaries
-        path: binaries
-        persist-credentials: false
+  release_binaries:
```
vedpatwardhan, line 5: I think we should instead add this name aside the `name` field on line [1](https://github.com/unifyai/ivy/pull/26867/files#diff-bdfe89ee8491e51e3f81ada685c51d24188416d8b529b443ee6085a0e0fd4a4aR1). I think given that this workflow's name is the same as that of `pypi.yml`, we have this problem where the `binaries` workflow doesn't get named correctly. 
And yeah the `release-binaries` title sounds good to me
Thanks @Sai-Suraj-27 😄

vedpatwardhan (changes requested):

.github/workflows/binaries.yml
Sai-Suraj-27: @vedpatwardhan sure, I added it as the name on line-1, but I think it should be added under `jobs` also? I'm not sure but in all the other `.yml` files there is a name for the job under `jobs:` like `run_tests:`, `fix-linting:`, `build:`, `publish-release-docs:` , etc. So, I guess we should add the `release-binaries` name both on line 1, and immediately under the `jobs:` on line 4.

vedpatwardhan: yep sounds good! Thanks @Sai-Suraj-27 😄

PR #26946, opened 2023-10-12
fix: Functions in the statistical folder

Aaryan562:
image
@vedpatwardhan @AnnaTz  Every function in `ivy_tests/test_ivy/test_functional/test_core/test_statistical.py` is failing the test cases. I was assigned the trello card https://trello.com/c/71MXGtPF. But I feel before adding additional arguments to the functions we should make these existing functions pass the test cases. 

The error can be reproduced by running: `pytest ivy_tests/test_ivy/test_functional/test_core/test_statistical.py -s --disable-pytest-warnings --backend all `


ivy/functional/backends/paddle/statistical.py
```diff
@@ -89,8 +114,21 @@ def max(
     return ret.astype(ret_dtype)
 
 
-@with_supported_dtypes(
```
vedpatwardhan, line 92: There's a current problem with `with_supported_dtypes`, but that should be fixed soon. I don't think we should stop using `with_supported_dtypes` or `with_unsupported_dtypes` in favour of `with_supported_device_and_dtypes` and `with_unsupported_device_and_dtypes` as both types of decorators are supposed to do the same thing based on convenience 😄

Aaryan562: > There's a current problem with `with_supported_dtypes`, but that should be fixed soon. I don't think we should stop using `with_supported_dtypes` or `with_unsupported_dtypes` in favour of `with_supported_device_and_dtypes` and `with_unsupported_device_and_dtypes` as both types of decorators are supposed to do the same thing based on convenience 😄

@vedpatwardhan Okay got it. For the time being I have shifted to these decorators as there are functions failing even after correcting the decorators. For example mean gives assertion error as backend values and grouth truth values dosent match. Will look into each function and notify you accordingly!

vedpatwardhan: this issue has been fixed on `main` now, we can revert back to using `with_supported_dtypes` 😄

Aaryan562:
Closing this PR

PR #27003, opened 2023-10-14
feat: Implemented Jax numpy einsum_path

### LINKED ISSUE #26432: einsum_path

rohitkg83:
Implemented einsum_path function and associated tests in jax numpy front end function. 

#26432 

Closes #26432 

## Checklist

- [x] Did you add a function?
- [x] Did you add the tests?

### Socials

Twitter - @rkg830926


ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_mathematical_functions.py
```diff
@@ -1062,6 +1062,41 @@ def test_jax_ediff1d(
     )
 
 
+# einsum_path
+@handle_frontend_test(
+    fn_tree="jax.numpy.einsum_path",
+    eq_n_op_n_shp=helpers.einsum_helper(),
+    dtype=helpers.get_dtypes("numeric", full=False),
+    test_with_out=st.just(False),
+)
```
AnnaTz, line 1071: Hi @rohitkg83, thanks for the PR. You should also generate values to test the `optimize` argument.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_mathematical_functions.py
rohitkg83: Hey @AnnaTz - Thanks for reviewing this. I have now added optimize parameter to the tests. I haven't added boolean values to the samples as it seems that Jax einsum_path function currently fails when True or False is passed as optimize values. Jax einsum_path function calls opt_einsum.contract_path function and it seems that there is an open [bug](https://github.com/dgasmith/opt_einsum/issues/219) on their repository for boolean values. Hope this makes sense.

ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -2311,6 +2311,7 @@ def flatten(*, backend: str, ret):
                     x, dtype=ivy_backend.Dtype(str(np.asarray(x).dtype))
                 )
                 for x in ret_flat
+                if ivy.is_ivy_array(x)
```
AnnaTz, line 2314: Hi @rohitkg83, these changes you made in `function_testing.py` result in functions not being tested properly. Please remove them.
Because `einsum_path`'s return is unusual, we will have to manually check it's correct.

ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_mathematical_functions.py
```diff
@@ -1062,6 +1062,44 @@ def test_jax_ediff1d(
...
+        input_dtypes=dtypes,
+        backend_to_test=backend_fw,
+        frontend=frontend,
+        test_flags=test_flags,
+        fn_tree=fn_tree,
+        on_device=on_device,
+        subscripts=eq,
+        **kw,
+        optimize=optimize,
+    )
```
AnnaTz, line 1100: You will need to set `test_values` to False and get the frontend and ground truth return from `test_frontend_function` as is done [here](https://github.com/unifyai/ivy/blob/91478fff35902604e3c64cac7310d545319c0b50/ivy_tests/test_ivy/test_frontends/test_numpy/test_linalg/test_norms_and_other_numbers.py#L192). 
Then you can check the returns are the same like this:
```
len(ret[0]) == len(ret_gt[0])
all(x==y for x, y in zip(ret[0], ret_gt[0]))
ret[1] == str(ret_gt[1])
```

ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_mathematical_functions.py
AnnaTz: Please add a comment above the function's test, explaining the issue and including the link.

ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_mathematical_functions.py
AnnaTz: Currently these checks will fail due to the length not being equal. There is a part missing in your implementation's return.

AnnaTz (changes requested):

ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_mathematical_functions.py
rohitkg83: Done.

ivy_tests/test_ivy/helpers/function_testing.py
rohitkg83: Hey @AnnaTz  - I have removed these changes but the tests were still failing as these were trying to flatten the return with one return being string. I have made some changes to function testing file wherein it will only flatten the return if test values is true. The reason being when it is false we are returning non-flattened ret and front-end ret. Hope it makes sense!

ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_mathematical_functions.py
rohitkg83: @AnnaTz - Added these assertions to the test and all tests are now getting passed!

AnnaTz (approved):

PR #27004, opened 2023-10-14
feat: implementation of torch.cdist function to pytorch frontend.

### LINKED ISSUE #26797: cdist

joyios1:
# PR Description

Hello,

feat: implementation of torch.cdist function to pytorch frontend.

The shape of the two input arrays 'x1' and 'x2' is three dimensional tensors (BxRxM), so I check if the shape of both tensors is 3 otherwise an exception occurs.

the torch.cdist function does not accept float16 as a type therefore float32 and float 64 types were used.

the parameter 'p' recieves values from 0 to inf so I set minimum number 0 and maximum number 1 million. Is that correct? Maybe I need to use math.inf.

The parameter 'compute mode' is a tricky one. It handles the special case of p=2 (euclidean distance) and it specifies whether or not matrix multiplication (mm) will be used for the calculation of the distances . Therefore, I implemented both ways, the mm way with ivy.vector_norm and the not_mm way with 3 for loops.  I am sure if this is the correct way but the tests pass for all five (5) backends.

Thank you for your time, I am looking forward for your response.

Close #26797

# Checklist

- [x] Did you add a function?
- [x] Did you add the tests?

# Socials

Linkdin - linkedin.com/in/zois-tsakiris-759062286


ivy/functional/frontends/torch/miscellaneous_ops.py
```diff
@@ -92,6 +92,46 @@ def cartesian_prod(*tensors):
     return ret
 
 
+@with_unsupported_dtypes({"2.1.0 and below": "float16"}, "torch")
+@to_ivy_arrays_and_back
+def cdist(x1, x2, p, compute_mode="use_mm_for_euclid_dist_if_necessary"):
```
Sam-Armstrong, line 97:
```suggestion
def cdist(x1, x2, p=2.0, compute_mode="use_mm_for_euclid_dist_if_necessary"):
```
following the function signature from the docs https://pytorch.org/docs/stable/generated/torch.cdist.html

ivy_tests/test_ivy/test_frontends/test_torch/test_miscellaneous_ops.py
```diff
@@ -510,6 +510,50 @@ def test_torch_cartesian_prod(
...
 
 
+@handle_frontend_test(
+    fn_tree="torch.cdist",
+    dtypes_and_x=helpers.dtype_and_values(
+        shape=st.shared(helpers.get_shape(min_num_dims=3, max_num_dims=3), key="shape"),
+        shared_dtype=True,
+        num_arrays=2,
+        allow_inf=False,
+        dtype=["float32", "float32", "float64", "float64"],
```
Sam-Armstrong, line 520:
```suggestion
        available_dtypes=["float32", "float64"],
```

Sam-Armstrong (changes requested):
@joyios1 Thanks for making the PR! I've left a couple of minor comments but otherwise this looks good, so if we can get these addressed then we should be good to merge.

PR #27084, opened 2023-10-20
fix(jax-backend): fix failing test for mean

Aaryan562:
fix failing test for mean


ivy/functional/backends/jax/statistical.py
```diff
@@ -37,7 +37,7 @@ def max(
     return jnp.max(a=jnp.asarray(x), axis=axis, keepdims=keepdims)
 
 
-@with_unsupported_dtypes({"0.4.14 and below": "bfloat16"}, backend_version)
+@with_unsupported_dtypes({"0.4.19 and below": "bfloat16"}, backend_version)
```
vedpatwardhan, line 40: I think this should be `("bfloat16",)` instead of `"bfloat16"` for it to be considered a `tuple`

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
```diff
@@ -45,7 +45,7 @@ def _statistical_dtype_values(draw, *, function, min_value=None, max_value=None)
         small_abs_safety_factor = 24
     dtype, values, axis = draw(
         helpers.dtype_values_axis(
-            available_dtypes=helpers.get_dtypes("valid"),
+            available_dtypes=helpers.get_dtypes("float"),
```
vedpatwardhan, line 48: recently we've moved towards generating `"valid"` dtypes instead of `"float"` for every test. Ideally the unsupported dtype decorator should prevent all unintended dtypes from being generated (like `bool` or `int`)

vedpatwardhan (changes requested):
Hey @Aaryan562, just requested a couple of changes, thanks 😄

ivy/functional/backends/jax/statistical.py
```diff
@@ -37,7 +37,22 @@ def max(
     return jnp.max(a=jnp.asarray(x), axis=axis, keepdims=keepdims)
 
 
-@with_unsupported_dtypes({"0.4.14 and below": "bfloat16"}, backend_version)
+@with_unsupported_dtypes(
+    {
+        "0.4.19 and below": (
+            "bfloat16",
+            "int8",
```
vedpatwardhan, line 44: I think we can just use the `int` and `uint` typesets instead of explicitly specifying all signed and unsigned integer dtypes.
So for e.g. 
```
"bfloat16",
"int",
"uint"
```
would be equivalent to the `tuple` you've specified.
Also, I think ideally `mean` should support integers, are there any errors when passing ints to `mean`?

vedpatwardhan (changes requested):
Just requested a minor change, thanks @Aaryan562 😄

ivy/functional/backends/jax/statistical.py
vedpatwardhan: This error has been fixed on [`main`](https://github.com/unifyai/ivy/commit/d038b486dac1b8ad04bfaff90f99a5cae31a6449) now, could you please pull from `main` and try again? Thanks @Aaryan562 😄

Aaryan562: > I think we can just use the `int` and `uint` typesets instead of explicitly specifying all signed and unsigned integer dtypes. So for e.g.
> 
> ```
> "bfloat16",
> "int",
> "uint"
> ```
> 
> would be equivalent to the `tuple` you've specified. Also, I think ideally `mean` should support integers, are there any errors when passing ints to `mean`?

@vedpatwardhan If I don't  explicitly mention the datatypes then it generated those `int` and `uint` datatypes and gives the `AssertionError: returned dtype = float32, ground-truth returned dtype = int8`

Aaryan562: > > I think we can just use the `int` and `uint` typesets instead of explicitly specifying all signed and unsigned integer dtypes. So for e.g.
> > ```
> > "bfloat16",
> > "int",
> > "uint"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > would be equivalent to the `tuple` you've specified. Also, I think ideally `mean` should support integers, are there any errors when passing ints to `mean`?
> 
> @vedpatwardhan If I don't explicitly mention the datatypes then it generated those `int` and `uint` datatypes and gives the `AssertionError: returned dtype = float32, ground-truth returned dtype = int8`

Does that mean I need to make sure that the returned dtype should be of `int8` as it is not case of unsupported datatype. Also as discussed even though the python array api docs suggest that for mean we should have floating datatypes we should go beyond that as it is minimum standard.

Aaryan562: > I think we can just use the `int` and `uint` typesets instead of explicitly specifying all signed and unsigned integer dtypes. So for e.g.
> 
> ```
> "bfloat16",
> "int",
> "uint"
> ```
> 
> would be equivalent to the `tuple` you've specified. Also, I think ideally `mean` should support integers, are there any errors when passing ints to `mean`?

It is `integer` and `unsigned` which is equivalent to all integer and unsigned integers

vedpatwardhan (changes requested):
> > I think we can just use the `int` and `uint` typesets instead of explicitly specifying all signed and unsigned integer dtypes. So for e.g.
> > ```
> > "bfloat16",
> > "int",
> > "uint"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > would be equivalent to the `tuple` you've specified. Also, I think ideally `mean` should support integers, are there any errors when passing ints to `mean`?
> 
> It is `integer` and `unsigned` which is equivalent to all integer and unsigned integers

Ah sorry I got the names wrong, yeah it should be `integer` and `unsigned` instead of `int` and `uint`. Thanks @Aaryan562 😄

vedpatwardhan:
I'll quickly close and reopen this PR to see if the intelligent tests get triggered

vedpatwardhan:
nvm, but the tests seem to getting triggered properly for newly created PRs e.g [here](https://github.com/unifyai/ivy/pull/27137) and [here](https://github.com/unifyai/ivy/pull/27139) after the fix yesterday 😄

Aaryan562:
> nvm, but the tests seem to getting triggered properly for newly created PRs e.g [here](https://github.com/unifyai/ivy/pull/27137) and [here](https://github.com/unifyai/ivy/pull/27139) after the fix yesterday 😄

Ohh okay so should I close this PR or still there are some changes required?

vedpatwardhan:
> > nvm, but the tests seem to getting triggered properly for newly created PRs e.g [here](https://github.com/unifyai/ivy/pull/27137) and [here](https://github.com/unifyai/ivy/pull/27139) after the fix yesterday 😄
> 
> @vedpatwardhan Ohh okay so should I merge this PR or still there are some changes required?

I think we should get `mean` to work with integer and unsigned dtypes rather than marking them as unsupported as that seems a common usecase, e.g 
```
import jax.numpy as jnp
print(jnp.mean(jnp.array([1, 2, 3], dtype='int32'), dtype='int32'))
```
the above works which should also then work when using `ivy` with a `jax` backend. Thanks @Aaryan562 😄

Aaryan562:
> import jax.numpy as jnp
> print(jnp.mean(jnp.array([1, 2, 3], dtype='int32'), dtype='int32'))

@vedpatwardhan The `mean` function defined does not take `dtype` argument due to which this example will fail while using ivy. Now what do you suggest should I first add these extra argument to support superset behavior across all backends and ivy? Otherwise the thing is the output `dtype` is different for different backends as of now for example:
a = ivy.array([1,2,3], dtype='int32')
`numpy`->`float64`
`jax`->`float32`
`tensorflow`->`int32`

vedpatwardhan (changes requested):
> @vedpatwardhan The `mean` function defined does not take `dtype` argument due to which this example will fail while using ivy. Now what do you suggest should I first add these extra argument to support superset behavior across all backends and ivy? Otherwise the thing is the output `dtype` is different for different backends as of now for example:
> 
> ```
> for backend in ["numpy","jax","tensorflow"]:
>     ivy.set_backend(backend)
>     for dtype in ["int32"]:
>         a = ivy.array([1,2,3], dtype=dtype)
>         try:
>             print(
>                 "Value from"
>                 f" {backend} and its type is={ivy_back.mean(a)} and its dtype is {ivy_back.mean(a).dtype}"
>             )
>         except Exception as e:
>             print(e)
>     ivy.unset_backend()
> ```
> 
> `numpy`->`float64` `jax`->`float32` `tensorflow`->`int32`

Yeah I think we should add the `dtype` argument to `mean` just like we have for `sum` and then pass that argument to `jnp.mean` to ensure dtype consistency. In fact, we could also just pass `dtype = x.dtype` atm to `jnp.mean` without the `dtype` argument and that would work too, but I think adding the `dtype` argument is certainly an intuitive thing to do. Thanks @Aaryan562 😄

Aaryan562:
> > @vedpatwardhan The `mean` function defined does not take `dtype` argument due to which this example will fail while using ivy. Now what do you suggest should I first add these extra argument to support superset behavior across all backends and ivy? Otherwise the thing is the output `dtype` is different for different backends as of now for example:
> > ```
> > for backend in ["numpy","jax","tensorflow"]:
> >     ivy.set_backend(backend)
> >     for dtype in ["int32"]:
> >         a = ivy.array([1,2,3], dtype=dtype)
> >         try:
> >             print(
> >                 "Value from"
> >                 f" {backend} and its type is={ivy_back.mean(a)} and its dtype is {ivy_back.mean(a).dtype}"
> >             )
> >         except Exception as e:
> >             print(e)
> >     ivy.unset_backend()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `numpy`->`float64` `jax`->`float32` `tensorflow`->`int32`
> 
> Yeah I think we should add the `dtype` argument to `mean` just like we have for `sum` and then pass that argument to `jnp.mean` to ensure dtype consistency. In fact, we could also just pass `dtype = x.dtype` atm to `jnp.mean` without the `dtype` argument and that would work too, but I think adding the `dtype` argument is certainly an intuitive thing to do. Thanks @Aaryan562 😄

@vedpatwardhan Atm, I decided to go with `dtype = x.dtype` as anyways I am tasked with adding extra arguments for the statistical folder, I need to add `where` along with `dtype` which will do in separate PR.

ivy/functional/backends/jax/statistical.py
```diff
@@ -37,7 +37,10 @@ def max(
     return jnp.max(a=jnp.asarray(x), axis=axis, keepdims=keepdims)
 
 
-@with_unsupported_dtypes({"0.4.14 and below": "bfloat16"}, backend_version)
+@with_unsupported_dtypes(
```
vedpatwardhan, line 40: probably we can revert this change, seems like the old format of passing the decorator also passes the lint tests

vedpatwardhan (approved):
lgtm! Feel free to merge the PR without another review after making the below change, thanks @Aaryan562 😄

PR #27087, opened 2023-10-21
fix(jax-backend): fix failing test for sum

Aaryan562:
fix failing test for sum


AnnaTz:
Hi @Aaryan562, I'm not sure we should declare `complex` as unsupported for the jax backend in general, because the error you are trying to fix only occurs when `test_gradients=False`. 
@vedpatwardhan, what would you suggest in this case?
Fot the record, I ran the test and the full error is this: 
```
ivy.utils.exceptions.IvyBackendException: jax: execute_with_gradients: grad requires real-valued outputs (output dtype that is a sub-dtype of np.floating), but got complex128. For holomorphic differentiation, pass holomorphic=True. For differentiation of non-holomorphic functions involving complex outputs, use jax.vjp directly.
    ```

**Comment by vedpatwardhan** [comment]
> @vedpatwardhan, what would you suggest in this case? Fot the record, I ran the test and the full error is this:
> 
> ```
> ivy.utils.exceptions.IvyBackendException: jax: execute_with_gradients: grad requires real-valued outputs (output dtype that is a sub-dtype of np.floating), but got complex128. For holomorphic differentiation, pass holomorphic=True. For differentiation of non-holomorphic functions involving complex outputs, use jax.vjp directly.
> ```

Ideally we shouldn't be running gradient tests when the input is a complex dtype. We also have a [check](https://github.com/unifyai/ivy/blob/bb0b201805bf86bc66db65ea1ca05a1eab252960/ivy_tests/test_ivy/helpers/function_testing.py#L601) to avoid that from happening, could you please look into why the check isn't working for this test @Aaryan562? Thanks 😄

**Comment by Aaryan562** [comment]
> > @vedpatwardhan, what would you suggest in this case? Fot the record, I ran the test and the full error is this:
> > ```
> > ivy.utils.exceptions.IvyBackendException: jax: execute_with_gradients: grad requires real-valued outputs (output dtype that is a sub-dtype of np.floating), but got complex128. For holomorphic differentiation, pass holomorphic=True. For differentiation of non-holomorphic functions involving complex outputs, use jax.vjp directly.
> > ```
> 
> Ideally we shouldn't be running gradient tests when the input is a complex dtype. We also have a [check](https://github.com/unifyai/ivy/blob/bb0b201805bf86bc66db65ea1ca05a1eab252960/ivy_tests/test_ivy/helpers/function_testing.py#L601) to avoid that from happening, could you please look into why the check isn't working for this test @Aaryan562? Thanks 😄

@vedpatwardhan @AnnaTz The paddle version has been changed to `2.5.2`, which might create error in the `paddle` backend. The `jax` backend for the complex datatypes has been resolved. Kindly check!

File: ivy_tests/test_ivy/test_functional/test_core/test_statistical.py [hunk 1]
```diff
@@ -339,7 +339,7 @@ def test_sum(
     input_dtype, x, axis, castable_dtype = dtype_x_axis_castable
     # ToDo: set as_variable_flags as the parameter generated by test_sum once
     # this issue is marked as completed https://github.com/pytorch/pytorch/issues/75733
-    if "torch" in backend_fw:
+    if "torch" or "jax" in backend_fw:
```
vedpatwardhan, line 342: > @vedpatwardhan @AnnaTz The paddle version has been changed to `2.5.2`, which might create error in the `paddle` backend. The `jax` backend for the complex datatypes has been resolved. Kindly check!

Yeah it might create an issue with the `paddle` backend but for the `jax` backend we seem to be skipping the gradient tests entirely, but we should only be doing so if we've generated complex dtypes. Thanks @Aaryan562 😄

Aaryan562: > > @vedpatwardhan @AnnaTz The paddle version has been changed to `2.5.2`, which might create error in the `paddle` backend. The `jax` backend for the complex datatypes has been resolved. Kindly check!
> 
> Yeah it might create an issue with the `paddle` backend but for the `jax` backend we seem to be skipping the gradient tests entirely, but we should only be doing so if we've generated complex dtypes. Thanks @Aaryan562 😄

In my latest commit I have target `jax` backend and its `complex` dtypes.

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
```diff
@@ -342,6 +342,9 @@ def test_sum(
     if "torch" in backend_fw:
         assume(not test_flags.as_variable[0])
         assume(not test_flags.test_gradients)
+    if "jax" in backend_fw and castable_dtype in ["complex32", "complex64"]:
```
vedpatwardhan, line 345: I think this should be `"complex32", "complex128"` instead, what do you think?

vedpatwardhan (changes requested):
Hey @Aaryan562, just requested a final change, thanks 😄

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
Aaryan562: > I think this should be `"complex32", "complex128"` instead, what do you think?

Ivy support `complex64` and `complex128` so maybe we should cater to that? Also the paddle backend gave this error so changed its decorator `ivy.utils.exceptions.IvyBackendException: paddle: execute_with_gradients: paddle: nested_map: paddle: mean: (NotFound) The kernel with key (CPU, Undefined(AnyLayout), int32) of kernel `mean` is not registered. Selected wrong DataType `int32`. Paddle support following DataTypes: float64, complex128, float32, complex64, bool`
I dont know why `mean` is getting triggered even when we are dealing with `sum` function.

ivy/functional/backends/paddle/statistical.py
```diff
@@ -8,7 +8,6 @@
 import ivy
 from ivy.func_wrapper import (
     with_supported_dtypes,
-    with_unsupported_dtypes,
```
Aaryan562, line 11: I may have mistakenly deleted this line but  there is no function in this file which is using `@with_unsupported_dtypes` decorator, so ig its fine.

ivy/functional/backends/paddle/statistical.py
```diff
@@ -167,7 +166,10 @@ def std(
     return _std(x, axis, correction, keepdims).cast(x.dtype)
 
 
-@with_unsupported_dtypes({"2.5.2 and below": ("int8", "uint8")}, backend_version)
+@with_supported_dtypes(
```
Aaryan562, line 169: Also the paddle backend gave this error so changed its decorator `ivy.utils.exceptions.IvyBackendException: paddle: execute_with_gradients: paddle: nested_map: paddle: mean: (NotFound) The kernel with key (CPU, Undefined(AnyLayout), int32) of kernel mean is not registered. Selected wrong DataType int32. Paddle support following DataTypes: float64, complex128, float32, complex64, bool`
I dont know why `mean` is getting triggered even when we are dealing with `sum` function.

vedpatwardhan: `mean` is actually applied for the gradient tests [here](https://github.com/unifyai/ivy/blob/9d6dcbd0dabce5371f1973823c5291ab7002bbdb/ivy_tests/test_ivy/helpers/function_testing.py#L1123), so both `mean` and `sum` are called when running the gradient tests for `sum`. We should probably we fixing `mean` first if there's an issue with `mean`. Thanks @Aaryan562 😄
Also, does paddle not natively support `int` dtypes for its `mean` and `sum` functions (cc @MahmoudAshraf97)?

vedpatwardhan (changes requested):

ivy/functional/backends/paddle/statistical.py
vedpatwardhan: yep sounds good!

ivy/functional/backends/paddle/statistical.py
MahmoudAshraf97: [`int8`, `int16`, `uint8`] are not supported for `mean` and [`int8`, `uint8`] for `sum`

vedpatwardhan: oh okay, ig then `int32` and `int64` can also be marked as supported @Aaryan562 😄
(probably specifying the unsupported dtypes is more convenient that specifying the supported dtypes for this function)

Aaryan562: `ivy.utils.exceptions.IvyBackendException: paddle: execute_with_gradients: paddle: nested_map: paddle: mean: (NotFound) The kernel with key (CPU, Undefined(AnyLayout), int32) of kernel mean is not registered. Selected wrong DataType int32. Paddle support following DataTypes: float64, complex128, float32, complex64, bool`

@vedpatwardhan But then when supporting `int32` and `int64` i am getting the above error

Aaryan562: I further debugged I got to know it was when my `test_gradients=True`. So I changes my test_sum to make it false and all the test cases passed succesfully. But what could be the reason for it?

vedpatwardhan (approved):

PR #27096, opened 2023-10-22
Update compiler to a775614

Dsantra92:
# PR Description

Update the compiler.py and binaries.json for the autotuner.


vedpatwardhan:
https://github.com/unifyai/ivy/blob/7985b70e8780b33ff301d0073bb1adef61aa8a84/binaries.json#L7-L15

Actually in a git diff, `C` appears after `CX` rather than at the start (as visible [here](https://github.com/unifyai/binaries/pull/7/files)) so it becomes easier to look through the binaries in a PR and those in the `json` parallely. Let's probably move `C.so`, `D.so`, etc. back to the end of their corresponding series. Secondly, let's also keep the blank line between 2 series (e.g. `C` and `D`) as before as it becomes easier to cross-check. What do you think? Thanks @Dsantra92 😄

Dsantra92:
This is the first attempt at automating the binaries.json through the release pipeline. Sorting them would make the git diff easier to track when new files are added. I wasn't focusing on readability as it is now not supposed to be manually tweaked.

vedpatwardhan:
> This is the first attempt at automating the binaries.json through the release pipeline. Sorting them would make the git diff easier to track when new files are added. I wasn't focusing on readability as it is now not supposed to be manually tweaked.

Got it, thanks for clarifying @Dsantra92! Actually we'd still want it to be readable because on the off chance where there's some issue with the automated pipeline we might still need to debug if all binaries in the json are present. But yeah given that this is just a first attempt, I shouldn't have commented on it until it's finalized, sorry about that haha 😅

PR #27113, opened 2023-10-24
feat: Add TensorShape class to the TensorFlow frontend

Sam-Armstrong:
I still need to make some more changes here and add the rest of the tests, but just checking if you think the test I've added looks ok @AnnaTz? And how could I improve it? Thanks!

### DIFF
--- a/ivy/functional/frontends/tensorflow/tensor.py
+++ b/ivy/functional/frontends/tensorflow/tensor.py
@@ -8,9 +8,6 @@ from ivy.functional.frontends.tensorflow.func_wrapper import _to_ivy_array
 from ivy.functional.frontends.numpy.creation_routines.from_existing_data import array
 
 
-# Main
-
-
 class EagerTensor:
     def __init__(self, array):
         self._ivy_array = (
@@ -235,8 +232,8 @@ class TensorShape:
     # TODO: there are still some methods that may need implementing
 
     def __init__(self, dims):
-        self._dims = tuple(dims)
         self._ivy_shape = dims if isinstance(dims, ivy.Shape) else ivy.shape(dims)
+        self._dims = self._ivy_shape._shape
 
     def __repr__(self):
         if self._dims is not None:
--- a/ivy/functional/frontends/tensorflow/tensor.py
+++ b/ivy/functional/frontends/tensorflow/tensor.py
@@ -232,8 +232,7 @@ class TensorShape:
     # TODO: there are still some methods that may need implementing
 
     def __init__(self, dims):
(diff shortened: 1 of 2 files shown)


ivy/functional/frontends/tensorflow/tensor.py
```diff
@@ -228,6 +228,70 @@ def __iter__(self):
...
 
 
+class TensorShape:
+    # TODO: there are still some methods that may need implementing
+
+    def __init__(self, dims):
+        self._dims = tuple(dims)
+
+    def __repr__(self):
+        return str(self._dims)
```
AnnaTz, line 238: Is this really the original implementation? I think we should be printing something like this 
image

AnnaTz (changes requested):

ivy/functional/frontends/tensorflow/tensor.py
Sam-Armstrong: Yeah you're right, just looked at the source code. For some reason when I ran it in Python locally it returned the shape in the tuple format which is why I did it like this, not sure what was causing that

Sam-Armstrong:
Strangely there doesn't seem to be a property of method for directly accessing the value of `_dims`, other than one which is deprecated so probably not worth including https://github.com/tensorflow/tensorflow/blob/v2.14.0/tensorflow/python/framework/tensor_shape.py

AnnaTz:
> Strangely there doesn't seem to be a property of method for directly accessing the value of `_dims`, other than one which is deprecated so probably not worth including https://github.com/tensorflow/tensorflow/blob/v2.14.0/tensorflow/python/framework/tensor_shape.py

I'm guessing we will need it at some point but no reason to add it until we do.

ivy/functional/frontends/tensorflow/tensor.py
```diff
@@ -228,6 +228,115 @@ def __iter__(self):
             yield self[i]
 
 
+class TensorShape:
+    # TODO: there are still some methods that may need implementing
+
+    def __init__(self, dims):
+        self._ivy_shape = dims if isinstance(dims, ivy.Shape) else ivy.shape(dims)
+        self._dims = self._ivy_shape._shape
```
AnnaTz, line 236: I would suggest keeping only `_dims` here and inferring the ivy shape in the corresponding property.
Other than that, looks good!

AnnaTz (changes requested):

ivy/functional/frontends/tensorflow/tensor.py
Sam-Armstrong: Sure, thanks Anna!

PR #27118, opened 2023-10-25
fix: multiversion fix for torch==1.11.0 which doesnt have complex32…

mattbarrett98:


vedpatwardhan (changes requested):
Thanks for making the changes @mattbarrett98, I think we'd need to do the same for the `native_dtype_dict` in this file and change [these](https://github.com/unifyai/ivy/blob/85b428c0f6be816e8476614a49d04819e7b7d3f8/ivy/functional/backends/torch/__init__.py#L131) lists as well (same applies to the other backends). Seems like the author who added `complex32` a few weeks ago only updated this dict and not the ones in the `__init__.py` or even update the `ivy/__init__.py` and the other backends to add that `dtype`.
If that ends up being a lot of work, I'd say we can just revert [this](https://github.com/unifyai/ivy/commit/fc30de39e6269e58ab031e1d8fe806bba51d0e21) commit for now and add `complex32` back in sometime in the future. 
Works for me either way, let me know what you think. Thanks @mattbarrett98 😄

mattbarrett98:
> Thanks for making the changes @mattbarrett98, I think we'd need to do the same for the `native_dtype_dict` in this file and change [these](https://github.com/unifyai/ivy/blob/85b428c0f6be816e8476614a49d04819e7b7d3f8/ivy/functional/backends/torch/__init__.py#L131) lists as well (same applies to the other backends). Seems like the author who added `complex32` a few weeks ago only updated this dict and not the ones in the `__init__.py` or even update the `ivy/__init__.py` and the other backends to add that `dtype`.
> If that ends up being a lot of work, I'd say we can just revert [this](https://github.com/unifyai/ivy/commit/fc30de39e6269e58ab031e1d8fe806bba51d0e21) commit for now and add `complex32` back in sometime in the future.
> Works for me either way, let me know what you think. Thanks @mattbarrett98 😄

ah I see, seems quite strange such an incomplete change was able to be committed unnoticed, maybe we need to be stricter with who can push to main? or maybe this isn't as relevant anymore after recent changes? but yea I'd say we can just revert it for now

vedpatwardhan:
> ah I see, seems quite strange such an incomplete change was able to be committed unnoticed, maybe we need to be stricter with who can push to main? or maybe this isn't as relevant anymore after recent changes? but yea I'd say we can just revert it for now

Yeah I agree, that PR was merged without a review which ideally shouldn't happen. I've reverted the commit now 😄

mattbarrett98:
> > ah I see, seems quite strange such an incomplete change was able to be committed unnoticed, maybe we need to be stricter with who can push to main? or maybe this isn't as relevant anymore after recent changes? but yea I'd say we can just revert it for now
> 
> Yeah I agree, that PR was merged without a review which ideally shouldn't happen. I've reverted the commit now 😄

awesome, thanks ved!

PR #27125, opened 2023-10-26
fix: bug where fn_unsupported_dnd is a tuple on the latest paddle ver…

Sam-Armstrong:
@vedpatwardhan On the latest paddle version (2.5.2) `fn_unsupported_dnd` is a (empty) tuple, which is causing quite a lot of the checks in the tracer-transpiler to fail when .items() is attempted on it. This simple solution does fix the problem, but  there may well be a better fix.

### DIFF
--- a/ivy/functional/ivy/general.py
+++ b/ivy/functional/ivy/general.py
@@ -4069,8 +4069,9 @@ def _get_devices_and_dtypes(fn, recurse=False, complement=True):
                 list(fn_supported_dnd.values())[0], tuple
             )
 
-        for device, dtypes in fn_supported_dnd.items():
-            fn_supported_dnd[device] = tuple(_expand_typesets(dtypes))
+        if isinstance(fn_supported_dnd, dict):
+            for device, dtypes in fn_supported_dnd.items():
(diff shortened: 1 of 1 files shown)


ivy/functional/ivy/general.py
```diff
@@ -4086,8 +4086,9 @@ def _get_devices_and_dtypes(fn, recurse=False, complement=True):
                 list(fn_unsupported_dnd.values())[0], tuple
             )
 
-        for device, dtypes in fn_unsupported_dnd.items():
-            fn_unsupported_dnd[device] = tuple(_expand_typesets(dtypes))
+        if isinstance(fn_unsupported_dnd, dict):
```
vedpatwardhan, line 4089: Great spot, let's also make the same change on line [`4072`](https://github.com/unifyai/ivy/pull/27125/files#diff-f6556baf0d4607e70527506a4c173c78f8e3452e3b1c10d4d356210c527b41c3R4072) so that it works with the supported dtypes decorators too. Thanks @Sam-Armstrong 😄

vedpatwardhan (changes requested):

Sam-Armstrong:
All done, thanks for the review Ved!

PR #27152, opened 2023-10-27
feat: Add extra arguments to min

Aaryan562:
@vedpatwardhan As discussed in the trello card where I need to add additional arguments to the statistical functions, I decided to create individual PR for each function as it was becoming overwhelming for me to do all the changes in single PR and also it would be easy for you to debug and review.


Aaryan562:
image
I have tested with 100 examples. The failing tests in CI might be due to paddle new version

ivy/data_classes/array/statistical.py
```diff
@@ -14,7 +14,9 @@ def min(
         /,
         *,
         axis: Optional[Union[int, Sequence[int]]] = None,
-        keepdims: bool = False,
+        keepdims: Optional[bool] = False,
```
vedpatwardhan, line 17: actually the `Optional` typehint should only be used when the default value of the argument is `None`. Same applies to the container methods, ivy function and backend functions 😄

vedpatwardhan (changes requested):
Hey @Aaryan562, just requested a minor change. As far as the test failures go, most of them seem to be stemming from the new `paddle` release which is unrelated to your PR. The tests for `test_min` seem to be [passing](https://github.com/unifyai/ivy/actions/runs/6671646993/job/18134810889?pr=27152#step:3:8522). Thanks 😄

ivy/data_classes/array/statistical.py
Aaryan562: @vedpatwardhan Done the required changes. One thing I want to ask if you see how i have handled the various datatypes using if else, is that the only option I have if I want to incorporate them? I cant use `inf` for `integer` and `unsigned` as it gives `EagerTensor cannot convert `inf` to int32` or like that. So I chose this to replace the `max` value.

vedpatwardhan: I think this could be simplified further as mentioned in the review comment, thanks!

ivy/functional/backends/tensorflow/statistical.py
```diff
@@ -12,17 +12,44 @@
...
     axis: Optional[Union[int, Sequence[int]]] = None,
     keepdims: bool = False,
+    initial: Optional[Union[int, float, complex]] = None,
+    where: Optional[Union[tf.Tensor, tf.Variable]] = None,
     out: Optional[Union[tf.Tensor, tf.Variable]] = None,
 ) -> Union[tf.Tensor, tf.Variable]:
     axis = tuple(axis) if isinstance(axis, list) else axis
-    return tf.math.reduce_min(x, axis=axis, keepdims=keepdims)
+    if where is not None:
+        if x.dtype == tf.int8:
```
vedpatwardhan, line 30: I think we could simplify lines 30-49 with 
```
max_val = ivy.iinfo(x.dtype).max if ivy.is_int_dtype(x.dtype) else ivy.finfo(x.dtype).max
```
Same applies to the `paddle` backend implementation

vedpatwardhan (changes requested):
Hey @Aaryan562, just requested another change, thanks 😄

ivy/functional/backends/tensorflow/statistical.py
Aaryan562: Made the changes. Thanks!

vedpatwardhan: awesome!

ivy/functional/backends/numpy/statistical.py
```diff
@@ -20,10 +20,18 @@ def min(
     *,
     axis: Optional[Union[int, Sequence[int]]] = None,
     keepdims: bool = False,
+    initial: Optional[Union[int, float, complex]] = None,
+    where: Optional[np.ndarray] = None,
     out: Optional[np.ndarray] = None,
 ) -> np.ndarray:
+    if where is None:
```
vedpatwardhan, line 27: Not sure why we're doing this, doesn't `where=True` and `where=None` have the opposite behaviour?
Rest looks good to me, thanks @Aaryan562 😄

vedpatwardhan (changes requested):

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
```diff
@@ -82,7 +82,10 @@ def _statistical_dtype_values(draw, *, function, min_value=None, max_value=None)
                 | helpers.floats(min_value=0, max_value=max_correction - 1)
             )
         return dtype, values, axis, correction
-    return dtype, values, axis
+    dtype3, where = draw(
+        helpers.dtype_and_values(available_dtypes=["bool"], shape=shape)
```
vedpatwardhan, line 86: Another point, probably we can generate the `where` argument to have all broadcastable shapes of `shape` instead of `shape` directly

ivy/functional/backends/numpy/statistical.py
Aaryan562: Yeah right. I dont remember why I did maybe because I though the default value of where taken by numpy is True but its not the case so yeah reverted this change. Thanks

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
Aaryan562: @vedpatwardhan Could you pls specify the actual code to do this as I am unaware of this method and not able to get from the docs of Ivy.

ivy/functional/backends/numpy/statistical.py
vedpatwardhan: awesome, thanks!

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
vedpatwardhan: basically you could use something like [broadcastable_arrays](https://github.com/unifyai/ivy/blob/6e1f438cbe6d3cf7077f805648920074a8f15792/ivy_tests/test_ivy/test_functional/test_core/test_dtype.py#L75) to generate a boolean array for `where` along with the `values` if that makes sense?

vedpatwardhan: Apart from this, rest seems good to merge. Thanks @Aaryan562 😄

vedpatwardhan:
Happy to take a look again once you're done with the changes 😄

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
Aaryan562: > basically you could use something like [broadcastable_arrays](https://github.com/unifyai/ivy/blob/6e1f438cbe6d3cf7077f805648920074a8f15792/ivy_tests/test_ivy/test_functional/test_core/test_dtype.py#L75) to generate a boolean array for `where` along with the `values` if that makes sense?


@vedpatwardhan I'm sorry, but I didn't quite grasp that.😅 There is no explanation present to use it so I am finding it hard to decode the function and how to use it in my usecase.

vedpatwardhan: No worries, basically what we'd do is,
```
where_shape = helpers.mutually_broadcastable_shapes(num_shapes=1, base_shape=shape)
dtype3, where = draw(
        helpers.dtype_and_values(available_dtypes=["bool"], shape=where_shape)
)
```
Let me know if that works, thanks @Aaryan562 😄

vedpatwardhan (changes requested):

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
Aaryan562: @vedpatwardhan Hey actually the above code did not work as we need to draw the values of `where_shape` like [this ](https://github.com/Aaryan562/ivy/blob/min_stat_arg/ivy_tests/test_ivy/test_frontends/test_numpy/test_broadcast/test_methods.py#L19). The code which worked was:
```
    where_shape = draw(helpers.mutually_broadcastable_shapes(num_shapes=1, base_shape=shape))
    dtype3, where = draw(
            helpers.dtype_and_values(available_dtypes=["bool"], shape=where_shape[0])
    )
```
The above code is creating errors as the `where_shape` sometimes generates more dimensions than the `axis` which thereby gives the error: 
```
E       ValueError: input operand has more dimensions than allowed by the axis remapping
```
The falsifying example is:
```
Trying example: test_min(
    backend_fw='numpy',
    on_device='cpu',
    dtype_and_x=(['uint32'],
     [array([ 2561, 32768,     0, 22144,  1537, 49163, 58670], dtype=uint32)],
     (0,),
     ['bool'],
     [array([[False]])]),
    keep_dims=False,
    initial=0,
    test_flags=FunctionTestFlags(
        ground_truth_backend='tensorflow',
        num_positional_args=1,
        with_out=False,
        with_copy=False,
        instance_method=False,
        test_gradients=False,
        test_trace=False,
        transpile=False,
        as_variable=[False],
        native_arrays=[True],
        container=[True],
        precision_mode=False,
    ),
    fn_name='min',
    ```

File: ivy_tests/test_ivy/test_functional/test_core/test_statistical.py [hunk 5]
```diff
@@ -82,7 +82,20 @@ def _statistical_dtype_values(draw, *, function, min_value=None, max_value=None)
...
             )
         return dtype, values, axis, correction
-    return dtype, values, axis
+
+    if isinstance(axis, tuple):
+        axis = axis[0]
+
+    where_shape = draw(
+        helpers.mutually_broadcastable_shapes(
+            num_shapes=1, base_shape=shape, min_dims=0, max_dims=axis
```
Aaryan562, line 91: I have used `min_dims=0` as the default value defined is 1 which was causing errors as the axis would take 0 value. [this](https://github.com/Aaryan562/ivy/blob/min_stat_arg/ivy_tests/test_ivy/helpers/hypothesis_helpers/array_helpers.py#L1703).

ivy/functional/backends/tensorflow/statistical.py
```diff
@@ -12,17 +12,31 @@
 # -------------------#
 
 
-@with_unsupported_dtypes({"2.14.0 and below": ("complex",)}, backend_version)
+@with_unsupported_dtypes(
+    {"2.14.0 and below": ("complex", "bool", "uint64")}, backend_version
```
vedpatwardhan, line 16: not sure why `tf.mathreduce_min` doesn't support `uint64`, their docstring doesn't seem to indicate so, but it seems to be something that should ideally work, have you tested it with an example?

ivy/functional/backends/torch/statistical.py
```diff
@@ -14,23 +14,39 @@
 # -------------------#
 
 
-@with_unsupported_dtypes({"2.1.0 and below": ("complex",)}, backend_version)
+@with_unsupported_dtypes({"2.1.0 and below": ("complex", "bool")}, backend_version)
```
vedpatwardhan, line 17: Same here, `torch.min` should ideally be working with `bool`, could you please test that as well?

vedpatwardhan (changes requested):
Just requested a final couple of changes, rest looks good to merge. Thanks @Aaryan562 😄

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
vedpatwardhan: Ah makes sense, probably we should generate both shapes from this helper, does passing `num_shapes=2` and then using one for `x` and the other for `where` work?

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
vedpatwardhan: As mentioned in the comment above, let's try generating the shapes for both `x` and `where` through the `mutually_broadcastable_shapes` helper. Thanks @Aaryan562 😄

vedpatwardhan (commented):
Just requested a change in one of the comments from the previous review, rest looks good. thanks @Aaryan562 😄

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
Aaryan562: @vedpatwardhan It generates different shape like [(1,), (3,)] which then gives 
```ValueError: non-broadcastable operand with shape (1,) doesn't match the broadcast shape (3,)```

File: ivy/functional/backends/torch/statistical.py [hunk 8]
```diff
@@ -21,16 +21,32 @@ def min(
...
+    initial: Optional[Union[int, float, complex]] = None,
+    where: Optional[torch.Tensor] = None,
     out: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
     if axis == ():
         if ivy.exists(out):
             return ivy.inplace_update(out, x)
         else:
             return x
+
```
vedpatwardhan, line 33: could you please remove this blank line?

vedpatwardhan (changes requested):
The changes look good, but the test seems to be failing with the `jax` and numpy` backends. For the `jax` ofc, the test is failing due to the version number, but for the `numpy` backend, the [failure](https://github.com/unifyai/ivy/actions/runs/6833401165/job/18586027984?pr=27152#step:3:4754) seems to be related to the changes made. Could you please look into it? Thanks @Aaryan562 😄

Aaryan562:
> The changes look good, but the test seems to be failing with the `jax` and numpy`backends. For the`jax`ofc, the test is failing due to the version number, but for the`numpy` backend, the [failure](https://github.com/unifyai/ivy/actions/runs/6833401165/job/18586027984?pr=27152#step:3:4754) seems to be related to the changes made. Could you please look into it? Thanks @Aaryan562 😄

@vedpatwardhan I did the changes as you said in the last commit. But it did gave the error and I have mentioned it in the above msg as well which I think got missed.

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
Aaryan562: > @vedpatwardhan It generates different shape like [(1,), (3,)] which then gives `ValueError: non-broadcastable operand with shape (1,) doesn't match the broadcast shape (3,)`

@vedpatwardhan This was the msg which got overlooked.

vedpatwardhan: Oh sorry about missing this message, yeah let's just then skip using broadcasting here and revert back to what you were doing before where we used the same shape as `x` for generating the shape of `where`. Thanks @Aaryan562 😄

vedpatwardhan:
> @vedpatwardhan I did the changes as you said in the last commit. But it did gave the error and I have mentioned it in the above msg as well which I think got missed.

Sorry about that, just responded to that message, feel free to request a review once you're done with those changes. Thanks!

ivy_tests/test_ivy/test_functional/test_core/test_statistical.py
Aaryan562: > Oh sorry about missing this message, yeah let's just then skip using broadcasting here and revert back to what you were doing before where we used the same shape as `x` for generating the shape of `where`. Thanks @Aaryan562 😄

No worries, reverted the changes. Passed all the test cases successfully with 100 examples.

vedpatwardhan: sounds good!

vedpatwardhan (approved):

Issue #12656, opened 2023-03-18
Fix elementwise.test_sin


ivy-leaves:
|   |   |
|---|---|
|jax|<a href="https://github.com/unifyai/ivy/actions/runs/6048902031"><img src=https://img.shields.io/badge/-failure-red></a>
|numpy|<a href="https://github.com/unifyai/ivy/actions/runs/6048902031"><img src=https://img.shields.io/badge/-success-success></a>
|tensorflow|<a href="https://github.com/unifyai/ivy/actions/runs/6017745032/job/16324574226"><img src=https://img.shields.io/badge/-success-success></a>
|torch|<a href="https://github.com/unifyai/ivy/actions/runs/6048902031"><img src=https://img.shields.io/badge/-failure-red></a>
|paddle|<a href="https://github.com/unifyai/ivy/actions/runs/6059097283"><img src=https://img.shields.io/badge/-success-success></a>

PR #27156, opened 2023-10-28
fix: failing test for sin 

Aaryan562:
failing test for sin (shared priority task) #12656


hello-fri-end (changes requested):
Hey @Aaryan562, thanks a lot for looking into this failing test. Could you please link [this](https://github.com/unifyai/ivy/issues/12656) issue in the PR title. Also, I'm not sure if I follow why we shouldn't be testing for gradients in this function, could you please explain why it is necessary to set `test_gradients` as `False` in this?

Aaryan562:
> Hey @Aaryan562, thanks a lot for looking into this failing test. Could you please link [this](https://github.com/unifyai/ivy/issues/12656) issue in the PR title. Also, I'm not sure if I follow why we shouldn't be testing for gradients in this function, could you please explain why it is necessary to set `test_gradients` as `False` in this?

```
ivy.utils.exceptions.IvyBackendException: paddle: mean: (NotFound) The kernel with key (CPU, Undefined(AnyLayout), float16) of kernel `mean` is not registered. Selected wrong DataType `float16`. Paddle support following DataTypes: float64, complex128, float32, complex64, bool.
E         [Hint: Expected kernel_iter == iter->second.end() && kernel_key.backend() == Backend::CPU != true, but received kernel_iter == iter->second.end() && kernel_key.backend() == Backend::CPU:1 == true:1.] (at /paddle/paddle/phi/core/kernel_factory.cc:199)
```
@hello-fri-end Basically when `paddle` was taking `float16` dtype and  `test_gradients` flag was set to `True` it gave the above error. So in my latest commit I have focused on that but I am not quite sure why is error for this condition. Maybe your inputs could help.

Aaryan562:
@vedpatwardhan Since @hello-fri-end would be inactive this week, could you pls review my PR

vedpatwardhan (approved):

PR #27165, opened 2023-10-29
fix: Fixed a type annotation of `Tuple`.

### LINKED ISSUE #27164: Using PEP 585 annotations without a `from __future__ import annotations` import
In the following line:
https://github.com/unifyai/ivy/blob/ea0eaad440ff7f953c2fb4c621eccfe4a6fb2ecd/ivy/functional/backends/paddle/experimental/statistical.py#L422
It should be `Tuple[Any, ...]:` because we are not using the `from __future__ import annotations` import.
Using `tuple[Any, ...]:` without the `from __future__ import annotations` import will cause runtime errors on Python versions prior to 3.9 and 3.10, respectively.
So, it should be fixed.

Sai-Suraj-27:
# PR Description
In the following line:
https://github.com/unifyai/ivy/blob/ea0eaad440ff7f953c2fb4c621eccfe4a6fb2ecd/ivy/functional/backends/paddle/experimental/statistical.py#L422
It should be `Tuple[Any, ...]:` because we are not using the `from __future__ import annotations` import.
Using `tuple[Any, ...]:` without the `from __future__ import annotations` import will cause runtime errors on Python versions prior to 3.9 and 3.10, respectively.
So, i fixed it by changing to `Tuple[Any, ...]:`.

## Related Issue
Closes #27164

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials

### DIFF
--- a/ivy/functional/backends/paddle/experimental/statistical.py
+++ b/ivy/functional/backends/paddle/experimental/statistical.py
@@ -419,7 +419,7 @@ def unravel_index(
     /,
     *,
     out: Optional[paddle.Tensor] = None,
-) -> tuple[Any, ...]:
+) -> Tuple[Any, ...]:
     if indices.ndim == 0:
         indices = indices.unsqueeze(0)
     coord = []


rishabgit:
Great catch @Sai-Suraj-27! 😄

PR #27191, opened 2023-10-31
refactor dynamic backend to make use of DLPack

MahmoudAshraf97:
# PR Description
this PR integrates dlpack into the dynamic backend setting process to avoid using numpy, it also fixes minor issues listed below:
1. `ivy.Array.backend` now returns the actual backend of the array, not the current global backend
2. in implicit backend, it's treated as actual numpy backend, you can differentiate between numpy backend and implicit backend by checking the length of `ivy.backend_stack`

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -441,7 +441,7 @@ class Shape(Sequence):
 
     def with_rank(self, rank):
         try:
-            return self.merge_with(unknown_shape(rank=rank))
+            return self.merge_with(self.unknown_shape(rank=rank))
         except ValueError:
             raise ValueError(f"Shape {self} must have rank {rank}")
 
@@ -978,7 +978,7 @@ def del_global_attr(attr_name):
     delattr(globals_vars, attr_name)
 
 
-backend = ""
+backend = ivy.utils.backend.handler.implicit_backend
 backend_version = {}
 
 native_inplace_support = None
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -978,7 +978,7 @@ def del_global_attr(attr_name):
     delattr(globals_vars, attr_name)
 
 
-backend = ivy.utils.backend.handler.implicit_backend
+backend = ""
 backend_version = {}
 
 native_inplace_support = None
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -401,13 +401,7 @@ class Array(
                 self._post_repr = ")"
         sig_fig = ivy.array_significant_figures
         dec_vals = ivy.array_decimal_values
-        if self.backend == "" or ivy.is_local():
-            # If the array was constructed using implicit backend
-            backend = ivy.current_backend()
-        else:
-            # Requirerd in the case that backend is different
-            # from the currently set backend
-            backend = ivy.with_backend(self.backend)
+        backend = ivy.with_backend(self.backend)
         arr_np = backend.to_numpy(self._data)
         rep = (
             np.array(ivy.vec_sig_fig(arr_np, sig_fig))
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -160,7 +160,7 @@ class Array(
         self._dev_str = None
         self._pre_repr = None
         self._post_repr = None
-        self._backend = ivy.backend
+        self._backend = ivy.current_backend(self._data).backend
         if dynamic_backend is not None:
             self._dynamic_backend = dynamic_backend
         else:
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -51,7 +51,6 @@ from .experimental import (
     _ArrayWithStatisticalExperimental,
     _ArrayWithUtilityExperimental,
 )
-from ivy.utils.backend.handler import _get_backend_for_arg
 
 
 class Array(
@@ -161,7 +160,7 @@ class Array(
         self._dev_str = None
         self._pre_repr = None
         self._post_repr = None
-        self._backend = _get_backend_for_arg(self.data.__class__.__module__).backend
+        self._backend = ivy.backend
         if dynamic_backend is not None:
             self._dynamic_backend = dynamic_backend
         else:
@@ -190,7 +189,7 @@ class Array(
     @dynamic_backend.setter
     def dynamic_backend(self, value):
         from ivy.functional.ivy.gradients import _variable
-        from ivy.utils.backend.handler import _data_to_new_backend
+        from ivy.utils.backend.handler import _data_to_new_backend, _get_backend_for_arg
 
         if value:
             ivy_backend = ivy.with_backend(self._backend)
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -664,10 +664,10 @@ class Array(
         return ivy.remainder(self._data, other)
 
     def __divmod__(self, other):
-        return tuple([ivy.divide(self._data, other), ivy.remainder(self._data, other)])
+        return ivy.divide(self._data, other), ivy.remainder(self._data, other)
 
     def __rdivmod__(self, other):
-        return tuple([ivy.divide(other, self._data), ivy.remainder(other, self._data)])
+        return ivy.divide(other, self._data), ivy.remainder(other, self._data)
 
     def __truediv__(self, other):
         """
--- a/ivy/data_classes/array/array.py
+++ b/ivy/data_classes/array/array.py
@@ -51,6 +51,7 @@ from .experimental import (
     _ArrayWithStatisticalExperimental,
     _ArrayWithUtilityExperimental,
 )
+from ivy.utils.backend.handler import _get_backend_for_arg
 
 
 class Array(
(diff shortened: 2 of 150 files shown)


ivy/__init__.py
```diff
@@ -978,7 +978,7 @@ def del_global_attr(attr_name):
     delattr(globals_vars, attr_name)
 
 
-backend = ""
+backend = ivy.utils.backend.handler.implicit_backend
```
vedpatwardhan, line 981: I don't think we should be doing this, because as a result of this change `ivy.backend` returns `numpy` and so does `ivy.current_backend_str()`. However, if you haven't set a backend and still pass a `torch.Tensor` to an `ivy` function, it would defer to the `torch` backend implementation, which isn't the same as when the `numpy` backend is actually set. 
We should ideally retain the current behaviour when a backend isn't set, what do you think?

ivy/data_classes/array/array.py
```diff
@@ -160,7 +161,7 @@ def _init(self, data, dynamic_backend=None):
         self._dev_str = None
         self._pre_repr = None
         self._post_repr = None
-        self._backend = ivy.backend
+        self._backend = _get_backend_for_arg(self.data.__class__.__module__).backend
```
vedpatwardhan, line 164: I can't really think of a case when an `ivy.Array` is created with a different backend than the one set currently, not sure if that's possible either tbh (as ideally it would throw an error), could you please share an example of the issue being fixed? Thanks!

ivy/data_classes/array/array.py
```diff
@@ -188,28 +189,27 @@ def dynamic_backend(self):
...
-                native_data = _variable_data(self.data)
-                np_data = to_numpy(native_data)
-                new_arr = ivy.array(np_data)
-                self._data = _variable(new_arr).data
+            if ivy_backend.gradients._is_variable(self.data):
+                native_var = ivy_backend.gradients._variable_data(
+                    self,
+                )
+                data = _data_to_new_backend(native_var, ivy_backend).data
+                self._data = _variable(data).data
```
vedpatwardhan, line 203: ig this could also be `ivy_backend.gradients._variable` instead?

ivy/functional/backends/paddle/gradients.py
```diff
@@ -21,8 +21,6 @@
 
 
 def variable(x, /):
-    if ivy.is_int_dtype(x.dtype):
```
vedpatwardhan, line 24: I'm not sure why we're changing this, this is actually quite useful to avoid code from breaking when someone incorrectly passes an integer array to execute_with_gradients and misses out on specifying that it should be ignored in `xs_grad_idxs`

ivy/functional/ivy/general.py
```diff
@@ -1826,7 +1826,7 @@ def current_backend_str() -> Union[str, None]:
     """
     fw = current_backend()
     if not backend_stack:
-        return ""
+        return ivy.utils.backend.handler.implicit_backend
```
vedpatwardhan, line 1829: ig we shouldn't make this change for the reason mentioned in the `__init__.py` comment

ivy/utils/backend/handler.py
```diff
@@ -236,9 +236,24 @@ def _handle_backend_specific_vars(target, backend):
         target.set_global_attr("RNG", target.functional.backends.jax.random.RNG)
 
 
-def convert_from_source_backend_to_numpy(variable_ids, numpy_objs, devices):
-    # Dynamic Backend
-    from ivy.functional.ivy.gradients import _is_variable, _variable_data
+def _data_to_new_backend(x, previous_backend):
```
vedpatwardhan, line 239: I'm not sure if I'm testing the wrong example, but if I run the following,
```
import ivy

ivy.set_backend("torch")
x = ivy.array([1., 2., 3.])
ivy.set_backend("tensorflow")
print(ivy.utils.backend.handler._data_to_new_backend(x, ivy.functional.backends.torch))
```
It seems to throw an error `ivy.utils.exceptions.IvyBackendException: tensorflow: dev: 'torch.device' object is not subscriptable`. Probably we could separate out the conversion to and from dlpack in the way we were doing it for conversion to and from numpy in 2 different steps, what do you think?

ivy/utils/backend/handler.py
```diff
@@ -236,9 +236,24 @@ def _handle_backend_specific_vars(target, backend):
...
+    except Exception:
+        np_res = previous_backend.to_numpy(x.data)
+        result = ivy.asarray(np_res, device=device)
+    return result
+
+
+def dynamic_backend_converter(backend_stack):
+    from ivy.functional.ivy.gradients import _variable
+
+    if len(backend_stack) < 2:
```
vedpatwardhan, line 253: I think it's safe to assume that dynamic backend setting would only be used when one backend was set, a few arrays were created and then the user wants to switch to the second backend (i.e. `len(backend_stack)` would always be `>=2`).
Cuz we can't guarantee the backend of the array created when no backend was set (because any type of native array can be passed with implicit backend). And given that implicit backend is only supposed to be used for testing something out quickly, it makes our job much easier when all the arrays created were guaranteed to have the same backend.
Secondly, there might be some issues using a local `ivy` version (from `with_ backend`) and the global one, because the `ivy.Array` in the local version is different from the `ivy.Array` class in the global one, have you tested if the code works as expected?

ivy_tests/test_ivy/test_misc/test_backend_utils/test_backend_handler.py
```diff
@@ -88,32 +88,33 @@ def test_current_backend(backend, array_type):
...
-    "middle_backend,end_backend", [(a, b) for a in backends for b in backends if a != b]
+    "middle_backend,end_backend",
+    [(a, b) for a in backends for b in backends if (a != b and "mxnet" not in [a, b])],
 )
 def test_dynamic_backend_all_combos(middle_backend, end_backend):
     # create an ivy array, container and native container
     a = ivy.array([1, 2, 3])
     b = ivy.array([4, 5, 6])
     ivy_cont = ivy.Container({"w": a, "b": b})
-    nativ_cont = ivy.Container(
```
vedpatwardhan, line 98: are we skipping this test now (with native array containers)?

vedpatwardhan (changes requested):
Hey @MahmoudAshraf97, just requested a few changes, thanks 😄

ivy/data_classes/array/array.py
MahmoudAshraf97: `ivy_backend` is the old backend, but we need to call this function using `ivy` which uses the new backend

ivy_tests/test_ivy/test_misc/test_backend_utils/test_backend_handler.py
MahmoudAshraf97: I gave this a lot of thought, and I couldn't see how the dynamic backend switch should change a non-ivy native variable to the new backend, this is a very invasive behaviour and shouldn't be expected, instead I added a new test case for ivy-wrapped variable container

ivy/__init__.py
MahmoudAshraf97: we can still differentiate between implicit and numpy backend as I mentioned above by `len(ivy.backend_stack)`, it's demonstrated in the change in `sub_backend_handler.py`, aside from that everything is working as expected
image

ivy/data_classes/array/array.py
MahmoudAshraf97: it's possible to create an array with another backend than the set one using `with_backend`, but you are right, it still returns the correct backend then, reverting this

ivy/__init__.py
vedpatwardhan: I'm not sure if it will be used as an indicator through, `ivy.backend` being `numpy` would end up being assumed as though the backend set is `numpy`. I think the difference in `ivy.backend` with and without setting the backend is a much clearer indicator that `numpy` would only be used if the input passed to the function doesn't belong to any specific backend (as `current_backend()` returns `numpy` in this case, but the appropriate backend if the input belongs to another backend). I think the previous setup of `numpy` being the "default" backend rather than the "implicit" one is less ambiguous about the usability of `ivy` without setting a backend

ivy/data_classes/array/array.py
vedpatwardhan: sounds good, thanks!

ivy/data_classes/array/array.py
vedpatwardhan: got it! Thanks for clarifying

ivy_tests/test_ivy/test_misc/test_backend_utils/test_backend_handler.py
vedpatwardhan: yeah sounds good!

ivy/__init__.py
vedpatwardhan: .

ivy/data_classes/array/array.py
```diff
@@ -401,13 +400,7 @@ def __repr__(self):
                 self._post_repr = ")"
         sig_fig = ivy.array_significant_figures
         dec_vals = ivy.array_decimal_values
-        if self.backend == "" or ivy.is_local():
```
vedpatwardhan, line 404: not sure why we're changing the `__repr__`, ig related to the same point about `numpy` being the implicit backend? This is something which could break if the array was belonging to a different backend than `numpy` right?

ivy/functional/backends/paddle/gradients.py
vedpatwardhan: .

vedpatwardhan (commented):
Just responded to some conversations and took another look, let's continue on the unresolved conversations, thanks @MahmoudAshraf97 😄

PR #27195, opened 2023-11-01
style: Consistent Style for `pytest decorators`

### LINKED ISSUE #27194: Follow a consistent Style for `pytest decorators`
If a `@pytest.mark.<marker>()` doesn't take any arguments, the parentheses are optional.
Either removing those unnecessary parentheses or requiring them for all fixtures is fine, but it's best to be consistent.
Similarly in the case of `@pytest.fixture()` where the parentheses are optional.
Either removing those unnecessary parentheses or requiring them for all fixtures is fine, but it's best to be consistent.

https://docs.astral.sh/ruff/rules/pytest-incorrect-mark-parentheses-style/
https://docs.astral.sh/ruff/rules/pytest-fixture-incorrect-parentheses-style/

Sai-Suraj-27:
# PR Description
If a `@pytest.mark.<marker>()` doesn't take any arguments, the parentheses are optional.
Either removing those unnecessary parentheses or requiring them for all fixtures is fine, but it's best to be consistent.
Similarly in the case of `@pytest.fixture()` where the parentheses are optional.
Either removing those unnecessary parentheses or requiring them for all fixtures is fine, but it's best to be consistent.

https://docs.astral.sh/ruff/rules/pytest-incorrect-mark-parentheses-style/
https://docs.astral.sh/ruff/rules/pytest-fixture-incorrect-parentheses-style/

I made changes all over the codebase (wherever required) according to these 2 rules, to increase readability and to adhere to best practices.

## Related Issue
Closes #27194

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
+++ b/ivy_tests/test_ivy/test_frontends/test_jax/test_random.py
@@ -115,7 +115,7 @@ def get_shape_and_arrays(draw):
 # ------------ #
 
 
-@pytest.mark.xfail
+@pytest.mark.xfail()
 @handle_frontend_test(
     fn_tree="jax.random.ball",
     dtype_key=helpers.dtype_and_values(
@@ -179,7 +179,7 @@ def test_jax_ball(
         assert u.shape == v.shape
 
 
-@pytest.mark.xfail
+@pytest.mark.xfail()
 @handle_frontend_test(
     fn_tree="jax.random.bernoulli",
     dtype_key=helpers.dtype_and_values(
@@ -234,7 +234,7 @@ def test_jax_bernoulli(
         assert u.shape == v.shape
 
 
-@pytest.mark.xfail
+@pytest.mark.xfail()
 @handle_frontend_test(
     fn_tree="jax.random.beta",
     dtype_key=helpers.dtype_and_values(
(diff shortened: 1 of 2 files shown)


vedpatwardhan:
> Not sure if that is correct, maybe @vedpatwardhan can you take a look?

Yep I agree, given that we aren't passing any arguments, we can follow the same syntax as the [docs](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail), what do you think @Sai-Suraj-27? Thanks 😄

Sai-Suraj-27:
> > Not sure if that is correct, maybe @vedpatwardhan can you take a look?
> 
> Yep I agree, given that we aren't passing any arguments, we can follow the same syntax as the [docs](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail), what do you think @Sai-Suraj-27? Thanks 😄

Anything is okay! Both the decorators can be used with or without the parenthesis. As i mentioned, pytest itself is using parenthesis in its codebase for [`@pytest.fixture()`](https://github.com/pytest-dev/pytest/blob/8fb7e8b31efaa55e760c142e26eb82b42081ca28/testing/test_python_path.py#L11). So, Its ultimately our choice to follow which style 😄.

vedpatwardhan:
I prefer the current approach tbh, what do you think @KareemMAX 😄

PR #27216, opened 2023-11-06
fix: some fixes to torch.autograd.backward

vedpatwardhan:
1. Updated the torch frontend GradFn to allow more than 2 inputs and also keyword arguments to be leaf nodes in the graph based on the function being called.
2. Made changes to the logic used for computing jacobians in the chain where the current computed jacobian is of shape `(layer_out_dim, layer_in_dim)` and the prev_grads is of shape `(out_dim, layer_out_dim)` and also added missing checks to avoid generating a ivy.jac callback for inputs which don’t require grad.

### DIFF
--- a/ivy/functional/frontends/torch/func_wrapper.py
+++ b/ivy/functional/frontends/torch/func_wrapper.py
@@ -29,8 +29,6 @@ class AccumulateGrad:
         return self.__name__ == __value
 
     def __call__(self, grads):
-        for i in range(self.__self__.ndim):
-            grads = grads.sum(-1)
         self.__self__._grads = grads
         return None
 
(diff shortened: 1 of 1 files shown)


aparajith21:
Sorry Ved I refuse to review this PR :joy:

vedpatwardhan:
> Sorry Ved I refuse to review this PR 😂

That's unacceptable haha 😂

hmahmood24 (approved):
This fixes the manual setting of `requires_grad=False` on the torch.Tensors. Thanks so much Ved! :D

vedpatwardhan:
great, will merge the PR now 😄

PR #27218, opened 2023-11-06
refactor: Refactored few files to make the code better.

### LINKED ISSUE #27217: Remove unnecessary-double-cast iterables.
The following optimizations can be made to make the code better:
It's unnecessary to double-cast or double-process iterables by wrapping the listed functions within an additional `list`, `set`, `sorted`, or `tuple` call. Doing so is redundant and can be confusing for readers.

There is no reason to use a `list comprehension` if the result is `immediately unpacked`. Instead, use a `generator` expression, which is more efficient as it avoids allocating an intermediary list.

I used [`ruff`](https://docs.astral.sh/ruff/) linter to identify these.

Sai-Suraj-27:
# PR Description
The following optimizations can be made to make the code better:
It's unnecessary to double-cast or double-process iterables by wrapping the listed functions within an additional `list`, `set`, `sorted`, or `tuple` call. Doing so is redundant and can be confusing for readers.

There is no reason to use a `list comprehension` if the result is `immediately unpacked`. Instead, use a `generator` expression, which is more efficient as it avoids allocating an intermediary list.

I used [`ruff`](https://docs.astral.sh/ruff/) linter to identify and fix these.

## Related Issue
Closes #27217

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials

### DIFF
--- a/ivy_tests/test_ivy/test_functional/test_core/test_device.py
+++ b/ivy_tests/test_ivy/test_functional/test_core/test_device.py
@@ -289,7 +289,7 @@ def test_function_supported_devices(
         res = ivy_backend.function_supported_devices(func)
         exp = set(expected)
 
-        assert sorted(tuple(exp)) == sorted(res)
+        assert sorted(exp) == sorted(res)
 
 
 # function_unsupported_devices
@@ -308,7 +308,7 @@ def test_function_unsupported_devices(
         res = ivy_backend.function_unsupported_devices(func)
         exp = set(expected)
 
-        assert sorted(tuple(exp)) == sorted(res)
+        assert sorted(exp) == sorted(res)
 
 
 @handle_test(
--- a/ivy_tests/test_ivy/test_functional/test_core/test_dtype.py
+++ b/ivy_tests/test_ivy/test_functional/test_core/test_dtype.py
@@ -655,7 +655,7 @@ def test_function_supported_dtypes(*, func, backend_fw):
         exp = set(ivy_backend.all_dtypes).difference(
             set(func.test_unsupported_dtypes[backend_fw])
         )
-        assert set(tuple(exp)) == set(res)
+        assert set(exp) == set(res)
 
 
 # function_unsupported_dtypes
@@ -667,7 +667,7 @@ def test_function_unsupported_dtypes(*, func, backend_fw):
     with BackendHandler.update_backend(backend_fw) as ivy_backend:
         res = ivy_backend.function_unsupported_dtypes(func)
         exp = func.test_unsupported_dtypes[backend_fw]
-        assert set(tuple(exp)) == set(res)
+        assert set(exp) == set(res)
 
 
 # iinfo
--- a/ivy/func_wrapper.py
+++ b/ivy/func_wrapper.py
@@ -46,7 +46,7 @@ FN_DECORATORS = [
 casting_modes_dict = {
     "uint": lambda: ivy.valid_uint_dtypes,
     "int": lambda: sorted(
-        tuple(set(ivy.valid_int_dtypes).difference(set(ivy.valid_uint_dtypes)))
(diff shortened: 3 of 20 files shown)


ivy_tests/test_ivy/test_functional/test_core/test_device.py
```diff
@@ -289,7 +289,7 @@ def test_function_supported_devices(
         res = ivy_backend.function_supported_devices(func)
         exp = set(expected)
 
-        assert sorted(tuple(exp)) == sorted(res)
```
KareemMAX, line 292: Are you sure this will pass the equality test?

ivy_tests/test_ivy/test_functional/test_core/test_device.py
```diff
@@ -308,7 +308,7 @@ def test_function_unsupported_devices(
         res = ivy_backend.function_unsupported_devices(func)
         exp = set(expected)
 
-        assert sorted(tuple(exp)) == sorted(res)
```
KareemMAX, line 311: Are you sure this will pass the equality test?

ivy_tests/test_ivy/test_functional/test_core/test_dtype.py
```diff
@@ -655,7 +655,7 @@ def test_function_supported_dtypes(*, func, backend_fw):
         exp = set(ivy_backend.all_dtypes).difference(
             set(func.test_unsupported_dtypes[backend_fw])
         )
-        assert set(tuple(exp)) == set(res)
```
KareemMAX, line 658: Are you sure this will pass the equality test?

ivy_tests/test_ivy/test_functional/test_core/test_dtype.py
```diff
@@ -667,7 +667,7 @@ def test_function_unsupported_dtypes(*, func, backend_fw):
     with BackendHandler.update_backend(backend_fw) as ivy_backend:
         res = ivy_backend.function_unsupported_dtypes(func)
         exp = func.test_unsupported_dtypes[backend_fw]
-        assert set(tuple(exp)) == set(res)
```
KareemMAX, line 670: Are you sure this will pass the equality test?

KareemMAX (changes requested):
Hey @Sai-Suraj-27,

Pointed out some small issues that I'm not certain of. Also @vedpatwardhan can you take a look before merging?

ivy_tests/test_ivy/test_functional/test_core/test_device.py
Sai-Suraj-27: Yes, `exp` and `res` are already iterable objects, removing the `tuple()` should not change anything because the `sorted` function acts the same on any iterable. So, I think it will pass the equality test.

AnnaTz:
Hi @Sai-Suraj-27, I don't see any issue with these changes, but I think we should test everything to make sure no failures are caused in practice.
I think @vedpatwardhan should take a look into this as well before we approve the merge.

vedpatwardhan (approved):
lgtm! Feel free to merge, thanks @Sai-Suraj-27 😄

Sai-Suraj-27:
> Hi @Sai-Suraj-27, I don't see any issue with these changes, but I think we should test everything to make sure no failures are caused in practice. Have you tried running the tests of the functions you modified? I think @vedpatwardhan should take a look into this as well before we approve the merge.

Yes, I ran tests for these 2 files `ivy_tests\test_ivy\test_functional\test_core\test_dtype.py`, `ivy_tests/test_ivy/test_functional/test_core/test_device.py`, and tests of this function that I modified `def zero_fraction`.

**None of the tests where I made changes are failing.** So, I guess there is no problem 👍.


image


image

PR #27221, opened 2023-11-07
feat: add segment_sum function to Ivy TensorFlow frontend

### LINKED ISSUE #27220: segment_sum

pixel3user:
# add segment_sum function to Ivy TensorFlow frontend under math 

## 27220

Closes #27220

## Checklist

- [x] Did you add a function?
- [x] Did you add the tests?


hello-fri-end (changes requested):
Hey @pixel3user , thanks for working on this issue. Could you please undo the changes in the `.idea` folder and the `requirements.txt` file, we shouldn't be pushing any local changes to these files to the ivy repo. Happy to review once again after you're done :)

pixel3user (commented):
Hi, @hello-fri-end . I have removed all of the local changes. Kindly review :)

ivy/functional/frontends/tensorflow/math.py
```diff
@@ -612,6 +614,12 @@ def scalar_mul(scalar, x, name="scalar_mul"):
     return ivy.multiply(x, scalar).astype(x.dtype)
 
 
+@to_ivy_arrays_and_back
+def segment_sum(data, segment_ids, name="segment_sum"):
+    num_segments = np.array(np.max(segment_ids) + 1)
```
hello-fri-end, line 619: Why are we using numpy here, we can use ivy, right?

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
```diff
@@ -2552,6 +2552,40 @@ def test_tensorflow_scalar_mul(
...
 
+# segment_sum
+@handle_frontend_test(
+    fn_tree="tensorflow.math.segment_sum",
+    data=helpers.array_values(dtype=ivy.int32, shape=(5, 6), min_value=1, max_value=9),
+    segment_ids=helpers.array_values(
+        dtype=ivy.int32,
+        shape=(5,),
+        min_value=0,
+        max_value=4,
```
hello-fri-end, line 2563: Why are we hardcoding the dtype and shape here? Ideally, you should look into the `helpers.get_dtypes_and_values` function for generating the data for the test.

hello-fri-end (changes requested):

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
pixel3user: According to the tensorflow docs https://www.tensorflow.org/api_docs/python/tf/math/segment_sum . segment_ids must be one of the following types: int32, int64. A 1-D tensor whose size is equal to the size of data's first dimension. That's why i hardcoded the shape and dtype here. Same goes for the min and max values of segment_ids, must be in a range not maximum than the shape of the segement_ids itself. The data can be of any dtype but for simplicity and to understand the data matrix while i was working i have choosen int.

ivy/functional/frontends/tensorflow/math.py
pixel3user: Yes, i should have used ivy here, i just realized and i am trying to change from numpy to ivy but i am failing some test cases for tensorflow when using ivy instead of numpy because of the item method. Do ivy.array have any item() method when using with tensorflow? i will update as soon as i get my test cases completed.

pixel3user:
I am getting an exception from the back-end "handle back-end" and "another exception occur while handling the exception". 
The test cases are stuck in an infinite loop and is taking too long to process.

hello-fri-end (changes requested):
Hey @pixel3user, could you please remove all the changes from the `.idea.` folder, we shouldn't be pushing any local changes to the files inside this folder.

ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
hello-fri-end: Thanks for explaning @pixel3user. `helpers.get_dtypes_and_values` has a `max_dims` parameter which you can set to 1 to generate only 1-dimensional arrays. Moreover, you can specify `dtype = ['int32', 'int64']` so that only these datatypes are used to generate the values inside the arrays. 

About the data, we should be as exhaustive as possible in the tests, so we should be generating all the dtypes and arbitrary shapes in our tests. Happy to help out if you run into any issues after updating these. :)

pixel3user:
I have removed all local changes and all the test cases are passing. Kindly review the lasted commit.

hello-fri-end (changes requested):
Hey @pixel3user, could you please undo all the changes inside the `.idea` folder, they're still there. Happy to review once again after you;re done.

PR #27243, opened 2023-11-11
feat: update jax vversion to 0.4.20

Aaryan562:
new version of Jax

PR #27247, opened 2023-11-11
Fix: torch pointwise operations

niixxaaa:
Fixed import issue with hypothesis_helpers. All tests pass for most functions except for jax always and paddle sometimes. 
Getting the following error:
E       KeyError: 'cpu'
E       
E       You can reproduce this example by temporarily adding @reproduce_failure('6.88.3', b'AA==') as a decorator on your test case

This is a task in the priority dashboard

@vedpatwardhan

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_torch/test_pointwise_ops.py
+++ b/ivy_tests/test_ivy/test_frontends/test_torch/test_pointwise_ops.py
@@ -5,7 +5,7 @@ from hypothesis import strategies as st, assume
 # local
 import ivy
 import ivy_tests.test_ivy.helpers as helpers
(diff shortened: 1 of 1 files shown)


vedpatwardhan:
Hey @niixxaaa, thanks for creating the PR, @MahmoudAshraf97 would take a look soon.  As for the failures with `jax` and `paddle`, those were due to the new releases with `jax` and `paddle` that happened recently. The `paddle` one has been fixed on `main`, and the `jax` one is on a [PR](https://github.com/unifyai/ivy/pull/27243) which should be merged soon. So if the test failure is of the form `KeyError` or `Cannot sample from a zero length sequence`, then that failure can be attributed to those issues 😄

PR #27285, opened 2023-11-15
feat: Add Apple Silicon venv installation

NripeshN:


vedpatwardhan:
Really sorry for the delay, I'll take a look first thing in the morning tomorrow. Thanks @NripeshN 😄

vedpatwardhan (commented):
Despite the error message when doing the installation, the imports of `tensorflow`, `jax` and `torchvision` seem to be correctly working with python 3.8, 3.9 3.10 and 3.11 on my machine (M2), are there any issues with the usage of those libraries if we ignore the error? Thanks @NripeshN 😄

PR #27403, opened 2023-11-27
refactor(stateful): Refactor `torch.nn.Module` and `ivy.Module` classes

hmahmood24:
# PR Description

- Refactored the `torch.nn.Module` class to use `ivy.Module` methods where possible.
- Removed bloatware from the code e.g. hooks etc. that are not needed
- Rewrote the `torch.nn.Module` `_create_variables` and `_build` methods to correctly find variables and build submodules recursively
- Modified the `ivy.Module.__setattr__` method to also handle variables and modules passed in such that they are also registered in the `_v` and `_module_dict` respectively to mimic what `torch` essentially does as well.
- Fixed the infinite recursion issue in `ivy.Module.build` where while trying to access `v`, we build the module but while building the module, we again try to access `v` which was leading to a recursion loop.
- Tested the changes with the tracer-transpiler. All tests passing
- Tested the `test_converter`, `test_modules` and `test_sequential` locally in `ivy`. All tests passing

## Related Issue

Closes #

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?


ivy/stateful/helpers.py
```diff
@@ -298,10 +309,11 @@ def _rebuild(self):
         self.build(*self._args, **self._kwargs)
 
     def _compute_module_dict(self):
-        self._module_dict = Container()
```
vedpatwardhan, line 301: I think this might end up being a problem if we delete a module from a model, because that module would continue to be in the `_module_dict` even if we did `del model.module` or `model.module = new_module`, ensuring it's empty everytime we compute the dict (which is everytime we build). 
What do you think?

ivy/stateful/module.py
```diff
@@ -516,7 +520,7 @@ def __call__(
 
     def __getattribute__(self, name):
         if name == "v":
-            if super().__getattribute__("v") is None and not self.built:
```
vedpatwardhan, line 519: This check has been there for a while (like a year or so), any reasons for removing it (just out of curiosity)?

ivy/stateful/module.py
```diff
@@ -525,16 +529,33 @@ def __getattribute__(self, name):
     def __setattr__(self, name, value):
         if name in ["v", "buffers"]:
             name = "_" + name
-        if isinstance(value, Module):
-            ret = super().__setattr__(name, value)
+
+        from ivy.functional.ivy.gradients import _is_variable
+
+        if (
+            name != "_v" and _is_variable(value.ivy_array)
```
vedpatwardhan, line 536: actually the array being passed doesn't necessarily need to be an array with gradients enabled, because otherwise the same line of code would add the array to `_v` of the model for `jax` but not for `torch` because `_is_variable` would always be `True` for `jax` 😅
Secondly, this could also entail that every array added to an `ivy.Module` with the `jax` backend would be added to the `_v` even if it's intended to be a non-parameter and non-buffer array (like the intermediate ones created in the methods, etc.). Happy to look into the issues which resulted in this change though 😄

ivy/stateful/module.py
```diff
@@ -525,16 +529,33 @@ def __getattribute__(self, name):
...
+            if hasattr(self, "_v"):
+                self._v[name] = value
+            if (
+                hasattr(self, "_build_mode")
+                and self.build_mode == "on_init"
+                and self.built
+            ):
+                self._rebuild()
+
+        elif name != "_module_dict" and isinstance(value, Module):
```
vedpatwardhan, line 549: Also a bit regarding this, I don't see any harm in adding this check but if we do end up with a case where the `name` is `_module_dict` and the `value` is an `ivy.Module`, then I think there's a bug somewhere else to be fixed because ideally `_module_dict` should always be a `Container` 😅

vedpatwardhan (changes requested):
Hey @hmahmood24, just made a few suggestions, thanks for creating the PR 😄

hmahmood24:
@vedpatwardhan, great points! This is gonna be stupid but many of the changes you suggested were removed without breaking anything. I added some of these when I was experimenting with things and forgot to clean them up. Turns out the last minute changes I made to the frontend `Module` class fixed a lot of the problems. The `__setattr__` method was also reverted since now the `_module_dict` is gonna be emptied anyways and `_v` is gonna be emptied and populated on repeated `explicit` build calls in the frontend anyways. I resolved the comments that were resolved/reverted. Left a comment underneath another one of your suggested change that I feel needs to stay. Let me know if that makes sense. Pushing the updated changes here

ivy/stateful/helpers.py
```diff
@@ -301,7 +302,9 @@ def _compute_module_dict(self):
         self._module_dict = Container()
         for key, value in self.__dict__.items():
             if isinstance(value, ivy.Module):
-                if "stateful" in value.__module__:
+                if key in self._module_dict:
```
vedpatwardhan, line 305: we don't need this `if` check either right? Given that `self._module_dict` is empty before iterating and that the `key` would be present in the `self.__dict__` only once?

ivy/stateful/module.py
```diff
@@ -362,6 +364,7 @@ def register_parameter(self, name, value):
             Value of the buffer
         """
         self._v.update({name: value})
+        super().__setattr__(name, value)
```
vedpatwardhan, line 367: I had intentionally avoided doing a `__setattr__` here actually so that we don't end up with a module and a parameter having the same name, leading to confusion. So for e.g. if you set something as a module `model.mod1 = <some ivy module>` and then do a `register_parameter("mod1", <array>)` then `model.mod1` becomes the parameter rather than the module itself 😅
I know that `torch` does it that way but with `torch` they have an explicit [`add_module`](https://github.com/pytorch/pytorch/blob/8933ff35953908545cb410d736a6c4a3a7761c54/torch/nn/modules/module.py#L596) method, but adding it and all the other validation for checking if the same name exists in the module, etc. might end up taking more time than adding somewhat a restriction where `model.mod1` will always be a module and `model.v.mod1` will always be a variable if that makes sense? Same applies to buffers as well

ivy/stateful/module.py
vedpatwardhan: Yep sounds good, in which case I think we can probably add back `if not super().__getattribute__("_v") and not self.built:`, given that an empty container is a boolean `False`. What do you think?

vedpatwardhan (changes requested):
> @vedpatwardhan, great points! This is gonna be stupid but many of the changes you suggested were removed without breaking anything. I added some of these when I was experimenting with things and forgot to clean them up. Turns out the last minute changes I made to the frontend `Module` class fixed a lot of the problems. The `__setattr__` method was also reverted since now the `_module_dict` is gonna be emptied anyways and `_v` is gonna be emptied and populated on repeated `explicit` build calls in the frontend anyways. I resolved the comments that were resolved/reverted. Left a comment underneath another one of your suggested change that I feel needs to stay. Let me know if that makes sense. Pushing the updated changes here

No worries at all haha! Responded to one of the previous conversations and made a couple more suggestions. Thanks for the effort uptil now @hmahmood24 💪

ivy/stateful/module.py
hmahmood24: > I had intentionally avoided doing a `__setattr__` here actually so that we don't end up with a module and a parameter having the same name, leading to confusion. So for e.g. if you set something as a module `model.mod1 = <some ivy module>` and then do a `register_parameter("mod1", <array>)` then `model.mod1` becomes the parameter rather than the module itself 😅 I know that `torch` does it that way but with `torch` they have an explicit [`add_module`](https://github.com/pytorch/pytorch/blob/8933ff35953908545cb410d736a6c4a3a7761c54/torch/nn/modules/module.py#L596) method, but adding it and all the other validation for checking if the same name exists in the module, etc. might end up taking more time than adding somewhat a restriction where `model.mod1` will always be a module and `model.v.mod1` will always be a variable if that makes sense? Same applies to buffers as well

There are two ways we can deal with this then. One was this approach which I agree might not be the most robust one to go for. We can however do what `torch` [does](https://github.com/pytorch/pytorch/blob/8933ff35953908545cb410d736a6c4a3a7761c54/torch/nn/modules/module.py#L1676-L1687) by modifying our `__getattribute__` method such that whenever someone registers a parameter like `register_parameter("bias", ....)` and then subsequently fetches that using `self.bias`, we should still be able to retrieve that paramter from the module object. `torch` does this explicitly for parameters, buffers and modules. It is pertinent that if a parameter was registered i.e. added to `._v`, we still should be able to retrieve it by simply doing `model.bias` rather than `model.v.bias` if we want to superset `torch`'s behaviour here. What do you think? @vedpatwardhan

hmahmood24: Same is the [case](https://github.com/pytorch/pytorch/blob/8933ff35953908545cb410d736a6c4a3a7761c54/torch/nn/modules/module.py#L1750-L1758) with `delattr` actually

vedpatwardhan: Yeah exactly, the main difference between our design and theirs is that with `torch`,
```
import torch
from torch import nn


class Model(nn.Module):
    def __init__(self):
        super().__init__()
        self.mod1 = nn.Linear(10, 10)
        self.mod2 = nn.Linear(10, 10)
    
    def forward(self, x):
        return self.mod2(self.mod1(x))


model = Model()
print(model._parameters)
print(model._modules)
print(list(model.parameters()))
```
you'd see that `mod1` isn't a part of the `_parameters` but is a part of the `_modules`, but with our design with an equivalent `ivy.Module`, `mod1` would be a part of both the `_v` and the `_module_dict`, so we can't really distinguish between them when doing a `__getattr__` or a `__setattr__` or a `__delattr__`. 
So in order to replicate their design we'd need a `Parameter` class for e.g. to distinguish between a regular array and a parameter, build modules and find variables from the `_module_dict` everytime we try to access the parameter dict (if our `self._v` is also empty like theirs) and so on which is why I think the current minimal structure is better for now. 
Having said that, I do think we can still implement something like that in the frontends because we do mimic their design there 😄

hmahmood24: @vedpatwardhan, that's the reason why I came to you asking the other day if there's a way we can distinguish between regular arrays and model parameters in `ivy` 😄

vedpatwardhan: Yeah so in `ivy` you could distinguish them based on whether or not they are a part of the `model._v`, if an array's not a part of the `model._v` then it's a regular array 😄
So basically whenever a `torch.nn.Parameter` is created in the frontend, the corresponding `ivy.Array` would be a part of the `model._v` of the `ivy.Module` if that makes sense?

hmahmood24: @vedpatwardhan Also, all this was implemented in the frontends [previously](https://github.com/unifyai/ivy/blob/d9e8e49e16965a7c0dff22c0cfaad4f4422f6013/ivy/functional/frontends/torch/nn/modules/module.py#L717-L799) but it doesn't make sense to keep such lengthy implementations in the frontends if we want to eventually design `ivy.Module` to superset `torch.nn.Module` and other stateful classes anyways. I'm not sure what's the best way to go about this in the interest of time. Also that the work I'm doing with the transformations relies heavily on the changes in this PR

vedpatwardhan: Actually we are already supersetting the stuff being added to the `torch` frontend module in the sense that the `._v` is present and the `_module_dict` is present in the `ivy.Module` and we can add a new module or a new parameter, etc. The main difference is the way we use it. I do get the point that it might be a bit cluttered in the frontends, but unfortunately if the `torch.nn.Module` itself has those condition checks and so on, which creates the need to have those in the frontend 😅

hmahmood24: @vedpatwardhan Thing I'm trying to understand is it makes perfect sense if someone does `self.register_paramter("bias", ..) and then subsequently use `self.bias` simply afterwards even if it is just `ivy.Module` and not `torch.nn.Module`. Wouldn't you think?

vedpatwardhan: > Actually I didn't understand. Why would `mod1` be a part of both `_v` and `_module_dict` here? I've reverted the `__setattr__` method entirely so not sure why this would still happen? We are not calling `register_parameter` here. This is a simply `__setattr__` call, no? @vedpatwardhan

Probably I'm missing something but the `__setattr__` call seems to be there on [this](https://github.com/unifyai/ivy/pull/27403/files#diff-9d3cd8360ac6368fb7cfebdd75ef04230b2963749af122e8515e44ea6ddf3971R367) line which would set the name of the variable as an attribute of the object ig 😅

vedpatwardhan: > @vedpatwardhan Thing I'm trying to understand is it makes perfect sense if someone does `self.register_paramter("bias", ..)` and then subsequently use `self.bias` simply afterwards even if it is just `ivy.Module` and not `torch.nn.Module`. Wouldn't you think?

Yeah I do think that would be great (and more intuitive) but the problem is we need to do a fairly large refactor into the `Module` to get something close to it working, like changing our design such that `self._v` only contains the parameters that are registered using `register_parameter` and not the ones in the modules, then we'd need to incorporate this into the `model.v` property so that we `_find_variables` everytime the property is called, and add all the checks to avoid duplicates, etc. Basically we'd need to get our `ivy.Module` restructured quite a bit so that closely aligns with `torch.nn.Module` 😅

PR #27441, opened 2023-12-05
fix: Replace project.toml to pyproject.toml

NripeshN:


vedpatwardhan (commented):
Thanks for getting the CI working with pyproject.toml @NripeshN, let's verify if using `sudo` is necessary on linux systems for this to work or whether there. Does it work with `pip install -e --user .` instead of `sudo pip install -e .` on linux machines as well?

vedpatwardhan:
Hey @NripeshN, just wanted to confirm if you've tested that `sudo` isn't required for `pip install -e .` for installing `ivy` on linux systems and this is specific to github workflows before merging the PR? Also, could you please also update all other github workflows which do a `pip install -e` as well (like the manual-tests workflow, the submodule-specific workflows, etc.)? Thanks 😄

PR #27734, opened 2023-12-11
refactor: Refactored and reformatted for better readability and efficiency

Sai-Suraj-27:
# PR Description
I used [`ruff`](https://docs.astral.sh/ruff/), [`Pylint`](https://pylint.readthedocs.io/en/stable/) linters to identify and fix some of the changes.

## Related Issue
Closes #

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials


ivy/functional/backends/numpy/random.py
```diff
@@ -113,7 +113,7 @@ def randint(
 
 def seed(*, seed_value: int = 0) -> None:
```
KareemMAX, line 114: I think this should be a void function?

ivy/functional/backends/paddle/experimental/elementwise.py
```diff
@@ -751,9 +751,11 @@ def _is_scalar(x):
     """Determines if the given tensor is a scalar.
 
     Args:
```
KareemMAX, line 753:
```suggestion
    Args
```

ivy/functional/backends/paddle/experimental/elementwise.py
```diff
@@ -751,9 +751,11 @@ def _is_scalar(x):
...
+    ----
     - x (paddle.Tensor): Input tensor.
 
     Returns:
```
KareemMAX, line 757:
```suggestion
    Returns
```

ivy/functional/backends/paddle/random.py
```diff
@@ -142,7 +142,7 @@ def randint(
 
 def seed(*, seed_value: int = 0) -> None:
```
KareemMAX, line 143: I think this should be a void function? Maybe @vedpatwardhan can chime in here as looks like this is a systematic error (which I'm not sure what are the consequences of removing it)

ivy/functional/backends/tensorflow/random.py
```diff
@@ -146,7 +146,7 @@ def randint(
 
 def seed(*, seed_value: int = 0) -> None:
```
KareemMAX, line 147: Same issue with void function here @vedpatwardhan

ivy/functional/backends/torch/random.py
```diff
@@ -124,7 +124,7 @@ def seed(*, seed_value: int = 0) -> None:
             from torch import mps
 
             mps.manual_seed(seed_value)
-    return
+    return None
```
KareemMAX, line 127: Same issue with void function here @vedpatwardhan

ivy/functional/frontends/numpy/ndarray/ndarray.py
```diff
@@ -329,7 +329,7 @@ def flatten(self, order="C"):
 
     def fill(self, num, /):
         self.ivy_array = np_frontend.full(self.shape, num).ivy_array
-        return
+        return None
```
KareemMAX, line 332: Same issue with the void function @vedpatwardhan

ivy/functional/frontends/torch/dtype.py
```diff
@@ -44,4 +44,4 @@ def set_default_dtype(d):
     )
     global _default_dtype
     _default_dtype = d
-    return
+    return None
```
KareemMAX, line 47: Same issue with void function here @vedpatwardhan

KareemMAX (commented):
LGTM! There are some comments that I'm thinking they should be resolved.

Also @vedpatwardhan, I found some issues with function typehints where they are annotated with `def -> None`, I think that is a mistake. And it causes `ruff` to force the function to `return None`. I'm not really sure if that is intentional, if it was please let me know (I've mentioned you in all of those instances below). If it's not intentional feel free to let me know (no need to rereview the PR, just this part)

vedpatwardhan (commented):
> Also @vedpatwardhan, I found some issues with function typehints where they are annotated with `def -> None`, I think that is a mistake. And it causes `ruff` to force the function to `return None`. I'm not really sure if that is intentional, if it was please let me know (I've mentioned you in all of those instances below). If it's not intentional feel free to let me know (no need to rereview the PR, just this part)

Yep those should be void functions, don't think there'll be any issues removing the `-> None`, so feel free to remove it, thanks @KareemMAX @Sai-Suraj-27 😄

ivy/functional/backends/numpy/random.py
Sai-Suraj-27: @KareemMAX, I have made the changes as per the reply from Ved. Please look into it now. Thank you.

KareemMAX (approved):

PR #27738, opened 2023-12-12
refactor: Replaced `lambda` expressions with `def` functions

Sai-Suraj-27:
# PR Description
I see that, at few places in the codebase, a `lambda` expression is being used instead of a `def`.  I found all the places where this is happening using the [`ruff`](https://github.com/astral-sh/ruff) tool, and replaced all of those lambda expressions with a `def` function.

## Related Issue
Closes #

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -1211,7 +1211,10 @@ current_sub_backends = []
 downcast_dtypes = False
 upcast_dtypes = False
 crosscast_dtypes = False
-cast_dtypes = lambda: downcast_dtypes and upcast_dtypes and crosscast_dtypes
+
+
+def cast_dtypes():
+    return downcast_dtypes and upcast_dtypes and crosscast_dtypes
 
 
 def downcast_data_types(val=True):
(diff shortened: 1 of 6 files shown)


KareemMAX (changes requested):
Seems like there are some lint failures that are happening, can you fix those?

vedpatwardhan (commented):
I'm not fully sure if we should be replacing all lambda functions by `def` (espectially the ones that are simple like `lambda x: isinstance(x, bool)`), what do you think @KareemMAX? Thanks for creating the PR @Sai-Suraj-27 😄

KareemMAX:
> I'm not fully sure if we should be replacing all lambda functions by `def` (espectially the ones that are simple like `lambda x: isinstance(x, bool)`), what do you think @KareemMAX? Thanks for creating the PR @Sai-Suraj-27 😄

I don't think it will hurt, the functions will have equivalent AST after all. Unless you prefer lambdas from a readability/writability point of view

PR #27770, opened 2023-12-18
feat(framework): add cython wrappers mode

yopknopixx:
# PR Description

Integrate cython-wrappers in the Ivy framework


ivy/__init__.py
```diff
@@ -591,6 +591,7 @@ class Node(str):
 nan_policy_stack = []
 dynamic_backend_stack = []
 warn_to_regex = {"all": "!.*", "ivy_only": "^(?!.*ivy).*$", "none": ".*"}
+cython_wrappers_mode = False
```
vedpatwardhan, line 594: Following the general structure this could be `cython_wrappers_stack` instead

ivy/__init__.py
```diff
@@ -961,6 +963,7 @@ def __deepcopy__(self, memo):
     "default_uint_dtype_stack": data_type.default_uint_dtype_stack,
     "nan_policy_stack": nan_policy_stack,
     "dynamic_backend_stack": dynamic_backend_stack,
+    "cython_wrappers_mode": cython_wrappers_mode,
```
vedpatwardhan, line 966: would be stack here as well

ivy/func_wrapper.py
```diff
@@ -38,6 +38,10 @@
     "handle_nans",
 ]
 
+CYTHON_WRAPPERS = [
```
vedpatwardhan, line 41: is there a way we can avoid hardcoding a list here? Like is there a way we can just infer the wrappers based on the .so files available in the system and their names?

ivy/utils/backend/handler.py
```diff
@@ -358,7 +358,8 @@ def set_backend(backend: str, dynamic: bool = False):
         global ivy_original_dict
         if not backend_stack:
             ivy_original_dict = ivy.__dict__.copy()
-
+        # copy current cython_wrappers_mode to the original dict
+        ivy_original_dict["cython_wrappers_mode"] = ivy.cython_wrappers_mode
```
vedpatwardhan, line 362: I think the value of the cython wrappers mode would be preserved across backends, similar to the other global modes, so we can probably revert the changes to this file

vedpatwardhan (changes requested):
Hey @yopknopixx, just requested a few changes, thanks 😄
(apologies if the review request wasn't intended and you were still working on this 😅)

ivy/func_wrapper.py
yopknopixx: That is the intended way, the hardcoded list is only here for now for ease of testing. I'll update it soon. The logic for available `.so` files is already implemented, but not all of them work hence the hard-coded list.

wrappers.json
```diff
@@ -0,0 +1,200 @@
+{
+    "ivy": {
+        "functional": ["negative.so",
```
vedpatwardhan, line 3: As discussed before, we shouldn't be hardcoding the names of the wrappers but instead have some way to pull those binaries based on whether they exist programmatically, as it becomes much more easier to manage the usage 😅

vedpatwardhan (changes requested):
Just requested another change, feel free to re-request a review once you're done with this and the changes from the previous review. Rest of the code looks good to me, thanks @yopknopixx 😄

wrappers.json
yopknopixx: The reason for this is the automation building the binaries can't be sure which ones are working (at least in its current state). So, there has to be some way to specify which ones are working correctly (identified with manual testing for now). We could move this to the `cython-wrappers` repo till we get it automated. Please let me know what you think. I do agree it'll be better to have only working binaries on the binaries repo.

vedpatwardhan: Yeah exactly, ig we can have some check to ensure that we'd only have the working binaries on the binaries repo 😅
Ofc even the current approach works but it isn't as scalable, so probably we can just assume that all binaries on the binaries repo are working for now. This would lead to a few failures with the cython wrappers mode, but we've got tests to identify that anyway, what do you think?

yopknopixx: One more problem 😅. I'm using the JSON file to make sure I only mark those functions that have a wrapper implementation with the `download_cython_wrappers_wrapper`. The alternative would be to wrap every function (but then we'd end up re-wrapping it with the Python wrappers once the download fails (for functions where the wrapper doesn't exist). As for scalability, I think for the time-being we can manually update the JSON file with working wrappers.. What do you think?

vedpatwardhan: Yeah let's do the manual json for now, but eventually we'd need the programmatic approach as you mentioned where we re-wrap with the python wrappers once the download fails. Thanks!

ivy/utils/backend/handler.py
vedpatwardhan: I think the code will continue to work as expected even if we remove this line as we're using a stack now

ivy_tests/test_ivy/helpers/test_parameter_flags.py
```diff
@@ -48,6 +48,7 @@ def _as_varaible_strategy(draw):
 BuiltFrontendArrayStrategy = DynamicFlag(st.booleans())
 BuiltTranspileStrategy = DynamicFlag(st.just(False))
 BuiltPrecisionModeStrategy = DynamicFlag(st.booleans())
+BuiltCythonWrapperStrategy = DynamicFlag(st.booleans())
```
vedpatwardhan, line 51: let's set `BuiltCythonWrapperStrategy` to be `DynamicFlag(st.just(False))` for now, we need to figure out a way to update the database with those failures separate to the ones with the python function wrappers. That's because the cython wrapper code isn't public and so the failures emerging from the cython wrappers can't be fixed by the os community

vedpatwardhan (changes requested):
Thanks for making the changes @yopknopixx, just a final couple of requests. Feel free to re-request my review once you're done with these, we can get the PR merged after that 😄

ivy_tests/test_ivy/helpers/test_parameter_flags.py
yopknopixx: 2. Got it!
1. For some reason it does not work as expected without that line. I did verify it when you first mentioned it. 😅

vedpatwardhan: ig that might be because the mode is missing from [this](https://github.com/unifyai/ivy/pull/27770/files#diff-8f7902fadeb2dd53265eb3cfbeb649d12f543d777ebd7f85225d4431b76f067cR1450) list? I'd suggest searching for all occurrences of `array_mode` or some other mode and verifying if the equivalents are present for the `cython_wrappers_mode`, as ideally we wouldn't need to  do any specific assignment statement for a mode in the backend handler 😄

yopknopixx: Thanks for pointing it out @vedpatwardhan! I'd completely missed that 🤦

vedpatwardhan (approved):
Final thing, can we please also add a command line argument like [this](https://github.com/unifyai/ivy/blob/b59d1ee6ecd92919121a5d2812f179c995fd064c/ivy_tests/test_ivy/conftest.py#L352) for `with-cython-wrapper-testing` and `skip-cython-wrapper-testing`? Probably just do a search in files for the `with-trace-testing` and `skip-trace-testing` and replicate that for `with-cython-wrapper-testing` and `skip-cython-wrapper-testing`.
Feel free to merge the PR once that's done without a review, thanks @yopknopixx 😄

PR #27779, opened 2023-12-21
feat: Replaced `flake8` with `ruff` and removed `setup.cfg` by merging the configurations into `pyproject.toml`

### LINKED ISSUE #27775: Merge `setup.cfg` into `pyproject.toml`
Starting with [PEP 621](https://peps.python.org/pep-0621/), the Python community selected `pyproject.toml` as a standard way of specifying project metadata. `Setuptools` has adopted this standard and will use the information contained in this file as an input in the build process.

Also, I see that none of the big Python organizations ([jax](https://github.com/google/jax/blob/main/pyproject.toml), [Numpy](https://github.com/numpy/numpy/blob/main/pyproject.toml), [pandas](https://github.com/pandas-dev/pandas/blob/main/pyproject.toml), [mlflow](https://github.com/mlflow/mlflow/blob/master/pyproject.toml),...etc.) are using `setup.cfg` now especially when they are already using a `pyproject.toml` file.

`pyproject.toml` became the modern standard for build configuration in Python projects. Our current [`setup.cfg`](https://github.com/unifyai/ivy/blob/main/setup.cfg) also can be easily merged into the `pyproject.toml` file.
So, I think It is better to keep everything in one file `pyproject.toml`.

Sai-Suraj-27:
# PR Description
Replaced `flake8` with the following 2 major checks from `ruff`
1. https://docs.astral.sh/ruff/rules/#pyflakes-f
2. https://docs.astral.sh/ruff/rules/#pycodestyle-e-w

Removed `setup.cfg` by merging the required configurations into `pyproject.toml`

**Note**
1. Fixed the `E721` check : Do not compare types, use `isinstance()` rule wherever required.
2. Changed the position of `ruff` in pre-commit hooks because sometimes, the fixes might modify files/lines that needs to be formatted by `black` again. So, placed ruff hook before the black hook.
4. Added new `noqa` comments for 3 rules (`E501` : line-to-long, `E722` : Do not use bare except, and `E741` : ambiguous-variable-name) and removed the old ones as `ruff` requires `noqa` comments to be added in a particular way.

## Related Issue
Closes #27775 

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?


setup.cfg
```diff
@@ -1,25 +0,0 @@
-[flake8]
-max-line-length = 88
-ignore = E203, E402, E731, E704, W503, W504, W291, W293
```
vedpatwardhan, line 3: I don't fully know how this works, but shouldn't we be ignoring all of these in the `pyproject.toml`? In the `pyproject.toml` it seems to only be `E203`, `E402` and `E731`

vedpatwardhan (commented):
Just a minor point, rest of the changes look good. Let's see what @KareemMAX thinks. Thanks @Sai-Suraj-27 😄

KareemMAX (approved):
LGTM! I'll continue monitoring lint behavior to be sure

PR #27810, opened 2023-12-28
fix paddle problem with zipping iterables of tensors

Sam-Armstrong:
Fixes some failing checks on the tracer repo caused by changes in paddle 2.6.0. I've raised an issue on the paddle repo, but this should serve as a temporary solution.

### DIFF
--- a/ivy/functional/backends/paddle/general.py
+++ b/ivy/functional/backends/paddle/general.py
@@ -227,6 +227,10 @@ def gather(
     return _gather(params)
 
 
+@with_unsupported_device_and_dtypes(
+    {"2.5.2 and below": {"cpu": ("bfloat16", "float16")}},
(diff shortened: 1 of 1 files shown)


vedpatwardhan (changes requested):
The tests for gather_nd seem to be [failing](https://github.com/unifyai/ivy/actions/runs/7350999115/job/20013910597?pr=27810#step:4:1146) due to an unsupported dtype issue, could you please add `bfloat16` and `float16` as the unsupported dtypes for this backend implementation to see if the test passes? Rest of the changes seem good, thanks @Sam-Armstrong 😄

Sam-Armstrong:
Thanks @vedpatwardhan, looks like the test is passing now 😁 I'll go ahead and merge

PR #27886, opened 2024-01-09
fix jax backend set item with scalar values

mattbarrett98:
this fixes the jax backend for `set_item` when scalar values are used i.e.

```python
import ivy
ivy.set_jax_backend()
x = ivy.array([[1., 2.], [3., 4.]])
query = ivy.array([[True, False], [True, True]])
val = 5.
print(ivy.set_item(x, query, val))
```

### DIFF
--- a/ivy/functional/backends/jax/general.py
+++ b/ivy/functional/backends/jax/general.py
@@ -91,6 +91,7 @@ def set_item(
 ) -> JaxArray:
     if ivy.is_array(query) and ivy.is_bool_dtype(query):
         query, expected_shape = _mask_to_index(query, x)
+    if ivy.is_array(val):
         val = _broadcast_to(val, expected_shape)._data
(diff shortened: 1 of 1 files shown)


Ishticode:
hey @mattbarrett98 . I think there are cases where `expected_shape` is undefined. lmc if can be fixed unless you have something

Ishticode (commented):
leads to `expected_shape` being undefined in one test case (https://github.com/unifyai/ivy/actions/runs/7463291702/job/20307666791?pr=27886).
Indenting the new conditional seems to work but not sure entirely.
```python
  if ivy.is_array(query) and ivy.is_bool_dtype(query):
      query, expected_shape = _mask_to_index(query, x)
      if ivy.is_array(val):
          val = _broadcast_to(val, expected_shape)._data
```

PR #27890, opened 2024-01-09
feat: handle case for float values with int dtype specified in tf backend

Ishticode:
Previously

```python
import torch
import ivy
print(torch.tensor(0.5, dtype=torch.int64)) # works - creates tensor(0)

ivy.set_tensorflow_backend()
print(torch_frontend.tensor(0.5, dtype='int64')) # throws error
```
But after this the frontend torch will create `ivy.frontends.torch.Tensor(0)`

### DIFF
--- a/ivy/functional/backends/tensorflow/creation.py
+++ b/ivy/functional/backends/tensorflow/creation.py
@@ -93,10 +93,10 @@ def asarray(
     with tf.device(device):
         if tf.is_tensor(obj):
             ret = tf.cast(obj, dtype) if obj.dtype != dtype else obj
-        # elif dtype.is_integer and np.issubdtype(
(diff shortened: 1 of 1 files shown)


vedpatwardhan (changes requested):
Hey @Ishticode, thanks for creating the PR 😄
I think this is something we should do in the frontends because passing in a float input and specifying the int dtype is quite strange behaviour which should be treated as framework-specific and dealt with in the frontends. What do you think?

Ishticode (commented):
best handled in backend bcz all frontends support in native. e.g.
jnp.array(0.5, dtype=jnp.int64) -> array(0)
np.array(0.5, dtype=np.int64) -> array(0)
:smile:

PR #27921, opened 2024-01-15
fix tf get item with non 0 dim tensor queries

mattbarrett98:
fixes something like the following:
```python
import ivy
ivy.set_tensorflow_backend()
x = ivy.array([[0.]])
print(x[(slice(None, None, None), ivy.array([0]))])
```
where non 0 dimensional tensors are used in the query. there was a check filtering out non 0 dim boolean tensors, but i think this needs to be extensed to any non 0 dim tensor, as i've done. I also believe the fix you made @Ishticode is no longer necessary after this fix, so i removed this as well. Do you think we can add some of these cases to the get item tests @Ishticode ?


vedpatwardhan:

PR #27955, opened 2024-01-18
feat(stateful): Add `NativeModule` to all backends with the implementation of a custom `tensorflow.keras.Model` subclass for the `tensorflow` backend to follow `ivy.Module`'s API

hmahmood24:
# PR Description
Adds `tensorflow.keras.Model` stateful class that inherits directly from `tf.keras.Model` and also implements the front-facing API following `ivy.Module`. Also adds an additional `call` method to the frontend `nn.Module` class.

## Related Issue

Closes #

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?


hmahmood24:
Hey @vedpatwardhan, could you take a look? I didn't remove some more attributes, properties, arguments which I think can be safely removed for our use case but for the time being this version of the class works. We can then revisit and remove any unneeded code from the file

ivy/stateful/__init__.py
```diff
@@ -1,3 +1,4 @@
+from . import backends
```
vedpatwardhan, line 1: The only concern here is that we'd import `tensorflow` on importing `ivy`, thus adding it as a dependency. Let's probably do something like create a submodule named `module.py` in all the backends and add the keras model implementation in the `tensorflow` backend. After that we can just add a `NativeModule` to all the backend `__init__.py` files so that based on the backend, that particular model would be contained in `ivy.NativeModule`. After that we can just do the same thing as we're doing now except for the namespace change
Apologies for the delayed review @hmahmood24, let me know what you think, thanks 😄

vedpatwardhan (changes requested):

ivy/stateful/__init__.py
hmahmood24: Ah didn't think of it from that perspective. Thanks for the suggestion! 👍🏼

hmahmood24:
@vedpatwardhan For the time being, I've kept the `module.py` sub-module in all the backends empty except for `tensorflow`. For all the other backends, the `NativeModule` points to the backend native modules e.g. for `torch` points to `torch.nn.Module`, for `paddle` points to `paddle.nn.Layer` and so on. Let me know if this makes sense? Thanks!

vedpatwardhan (approved):
> @vedpatwardhan For the time being, I've kept the `module.py` sub-module in all the backends empty except for `tensorflow`. For all the other backends, the `NativeModule` points to the backend native modules e.g. for `torch` points to `torch.nn.Module`, for `paddle` points to `paddle.nn.Layer` and so on. Let me know if this makes sense? Thanks!

Yep sounds good, tbh even if we had set `NativeModule` as `None` for those backends it wouldn't be a problem but setting it as the native module of the framework by default sounds good! Feel free to merge the PR, thanks @hmahmood24 😄

PR #28037, opened 2024-01-24
fix: Fix Tests - test_function_backend_computation

### LINKED ISSUE #28036: [Bug]: Incorrect test handling when `with_copy=True`
### Bug Explanation

Many sub-tasks in the "Priority Open" section are failing and they all have one common error:
`RuntimeError: Can't call numpy() on Tensor that requires grad. Use tensor.detach().numpy() instead.`

Now, this is not an issue with the actual function implementation, but the test suite. Whenever the `requires_grad` attribute is `True`, then this is bound to happen, as they are passed through `np.may_share_memory` without detaching from the computation graph. I'm not sure about other backends, but it is happening in `torch` backend for sure.

### Steps to Reproduce Bug

On running `pytest`, which requires `with_copy=True` with `torch` backend. 

For example,
`pytest ivy_tests/test_ivy/test_functional/test_core/test_manipulation.py::test_squeeze --backend torch --verbose`

### Environment

OS: Pop!_OS 22.04 LTS x86_64 
Kernel: 6.6.6-76060606-generic
Environment: conda
Code Editor: VS Code
Python Version: 3.10.0

### Ivy Version

0.0.6.2

### Backend

- [ ] NumPy
- [ ] TensorFlow
- [X] PyTorch
- [ ] JAX

### Device

CPU

sgalpha01:
# PR Description

This is my attempt to address issue #28036. I welcome any suggestions for improvement. I replaced `np.may_share_memory` with `np.shares_memory` because the former has more false positives. There is a nice [SO discussion](https://stackoverflow.com/questions/44865261/what-is-the-difference-between-numpy-shares-memory-and-numpy-may-share-memory) on this topic.
 
But the main changes are manually handling the cases where `dtype=bfloat16` and where it was necessary to `detach` the tensors before check for sharing memory.

## Related Issue

Closes #28036

## Checklist

- [ ] Did you add a function?
- [x] Did you add the tests?

### DIFF
--- a/ivy_tests/test_ivy/helpers/function_testing.py
+++ b/ivy_tests/test_ivy/helpers/function_testing.py
@@ -1,13 +1,13 @@
 # global
 import copy
-import time
-from typing import Union, List, Optional
-import numpy as np
-import types
 import importlib
 import inspect
+import time
+import types
 from collections import OrderedDict
+from typing import List, Optional, Union
 
+import numpy as np
 
 from .globals import mod_backend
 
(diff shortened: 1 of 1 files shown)


Ishticode:
is this still draft?

PR #28079, opened 2024-01-27
feat: added Ivy.unflatten

### LINKED ISSUE #28054: ivy.unflatten
Continuation of the work made at https://github.com/unifyai/ivy/pull/27416 .

Kacper-W-Kozdon:
# PR Description
Adds support for unflatten to the methods of data classes (ivy.Container, ivy.Array).

## Related Issue

Closes #28054

## Checklist

- [x] Did you add a function?
- [x] Did you add the tests?


Kacper-W-Kozdon:
Hi @joaozenobio , would you mind taking a look if this is how it's meant to be set up? I also noticed that a similar function (take) has some extra arguments (mode, fill_value) to deal with some special cases of inputs and outputs. But it seems to come from one of the backends or at least from its implementation in ivy.take(), so it's probably not necessary currently in the unflatten() function? 

I've also noticed that `ivy_tests/array_api_testing/test_array_api/array_api_tests/test_manipulation_functions.py` there are not many tests implemented- does this one should have tests for containers and arrays as well or just `ivy_tests/test_ivy/test_functional/test_experimental/test_core/test_manipulation.py` (it's not yet implemented but it should be nearly identical to the frontend one)?

Kacper-W-Kozdon:
Hey, @KareemMAX , it seems that I have some problems with making it work. Pytest throws errors both on the _static_unflatten and unflatten for containers. In the traceback, the error says that there is something wrong with shape kwarg (I think it was written that it was too long?)- I'm rerunning tests to give more detailed description of the issue.

ivy/data_classes/array/experimental/manipulation.py
```diff
@@ -1134,6 +1134,62 @@ def take(
             self, indices, axis=axis, mode=mode, fill_value=fill_value, out=out
         )
 
+    def unflatten(
+        self: Union[int, ivy.Array, ivy.NativeArray],
```
joaozenobio, line 1138: In this case, the array object is calling this function therefore, "self" can only be of type "Ivy.Array" to fit accordingly to the logic of the Array class. Please refer to the previous code you can find in this file and the documentation.

ivy/data_classes/container/experimental/manipulation.py
```diff
@@ -4065,6 +4065,180 @@ def trim_zeros(
...
+            dim=dim,
+            out=out,
+            key_chains=key_chains,
+            to_apply=to_apply,
+            prune_unapplied=prune_unapplied,
+            map_sequences=map_sequences,
+        )
+
+    def unflatten(
+        self: Union[int, ivy.Array, ivy.NativeArray, ivy.Container],
```
joaozenobio, line 4157: Same error as in the array class, "self" can only be "ivy.container".

joaozenobio (commented):
Hi @Kacper-W-Kozdon ! Thanks for your new PR! 🚀 The functions look great, I've noted down some of the corrections in the code that should be done. One thing you need to be sure of is if the dim argument is Optional or not. If one of the backends provides the an optional dim argument, all your functions should provide support for an optional dim too so we can establish the superset behavior.  Please check this for each backend and pin me whenever you ready 😄

ivy/data_classes/array/experimental/manipulation.py
Kacper-W-Kozdon: Corrected 👍

ivy/data_classes/container/experimental/manipulation.py
Kacper-W-Kozdon: Corrected 👍

Kacper-W-Kozdon:
> Hi @Kacper-W-Kozdon ! Thanks for your new PR! 🚀 The functions look great, I've noted down some of the corrections in the code that should be done. One thing you need to be sure of is if the dim argument is Optional or not. If one of the backends provides the an optional dim argument, all your functions should provide support for an optional dim too so we can establish the superset behavior. Please check this for each backend and pin me whenever you ready 😄

Ok, got it- I'll go through the code to ensure that dim is correctly defined.

Kacper-W-Kozdon:
> Hi @Kacper-W-Kozdon ! Thanks for your new PR! 🚀 The functions look great, I've noted down some of the corrections in the code that should be done. One thing you need to be sure of is if the dim argument is Optional or not. If one of the backends provides the an optional dim argument, all your functions should provide support for an optional dim too so we can establish the superset behavior. Please check this for each backend and pin me whenever you ready 😄

Only paddle and torch have natively implemented unflatten(), in neither of them dim is optional, so I fixed that. Shape is required as well. Since both are as positional/keyword and both are required, I removed default values from dim and shape. I did that for all the backgrounds covered by pytest.

joaozenobio:
Hi @Kacper-W-Kozdon ! Perfect, all tests are passing and the code looks in good shape :smile: I'll merge your PR now. Thanks for the contribution and keep up with the hard work :rocket:

PR #28122, opened 2024-01-30
fix: use tf numpy api instead for broader superset dtype support

mattbarrett98:
I've changed many of the tf elementwise functions that have been problematic for the integrations (as well as some other similar functions which have the same problem) to use the `experimental.numpy` api instead, which mirrors numpy, and hence has wider dtype support.

The basic problem being that tf has stricter dtype support, and so often when transpiling from any other framework to tf we run into issues. for example, all other backends support integer inputs in `exp` but tf, so if `torch.exp` is used on an integer input in some function, and this is transpiled to tf, it will throw an exception as integers aren't supported in `tf.math.exp`.

I think this is a pretty fundamental hole in our 'superset' behaviour which we claim to strive for. It also required enabling some flag in the tf backend init, which hopefully doesn't have any unintended consequences in the tests @vedpatwardhan

### DIFF
--- a/ivy/functional/backends/tensorflow/__init__.py
+++ b/ivy/functional/backends/tensorflow/__init__.py
@@ -3,7 +3,7 @@ import sys
 import logging
 import tensorflow as tf
 
-tf.experimental.numpy.experimental_enable_numpy_behavior(prefer_float32=True) 
+tf.experimental.numpy.experimental_enable_numpy_behavior(prefer_float32=True)
 
 for device in tf.config.experimental.list_physical_devices("GPU"):
     try:
--- a/ivy/functional/backends/tensorflow/__init__.py
+++ b/ivy/functional/backends/tensorflow/__init__.py
@@ -3,6 +3,8 @@ import sys
 import logging
 import tensorflow as tf
 
+tf.experimental.numpy.experimental_enable_numpy_behavior(prefer_float32=True)
+
 for device in tf.config.experimental.list_physical_devices("GPU"):
     try:
         tf.config.experimental.set_memory_growth(device, True)
--- a/ivy/functional/backends/tensorflow/__init__.py
+++ b/ivy/functional/backends/tensorflow/__init__.py
@@ -3,7 +3,7 @@ import sys
 import logging
 import tensorflow as tf
 
-tf.experimental.numpy.experimental_enable_numpy_behavior(prefer_float32=True)
+tf.experimental.numpy.experimental_enable_numpy_behavior(prefer_float32=True) 
 
 for device in tf.config.experimental.list_physical_devices("GPU"):
     try:
--- a/ivy/functional/backends/tensorflow/elementwise.py
+++ b/ivy/functional/backends/tensorflow/elementwise.py
@@ -50,12 +50,10 @@ def add(
     out: Optional[Union[tf.Tensor, tf.Variable]] = None,
 ) -> Union[tf.Tensor, tf.Variable]:
     x1, x2 = ivy.promote_types_of_inputs(x1, x2)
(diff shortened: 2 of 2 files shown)


vedpatwardhan:
ivy-gardener

vedpatwardhan:
Hey @mattbarrett98 could you please try a few examples using `ivy.crosscast_data_types` or `ivy.cast_data_types`?
So for e.g.,
```
import ivy

ivy.set_backend("tensorflow")
# can also use ivy.cast_data_types() instead which enables upcasting, downcasting as well as cross-casting
ivy.crosscast_data_types()  

x = ivy.array([1, 2, 3])
print(ivy.expm1(x))
```
gets `ivy.expm1` working with the `tensorflow` backend. I suppose there was some issue with the unsupported dtypes functions which is why this wasn't working. I had pushed a fix for the unsupported dtypes stuff last week after which this might've started working. Thanks 😄

vedpatwardhan (commented):
Hey @mattbarrett98 could you please try a few examples using `ivy.crosscast_data_types` or `ivy.cast_data_types`?
So for e.g.,
```
import ivy

ivy.set_backend("tensorflow")
# can also use ivy.cast_data_types() instead which enables upcasting, downcasting as well as cross-casting
ivy.crosscast_data_types()  

x = ivy.array([1, 2, 3])
print(ivy.expm1(x))
```
gets `ivy.expm1` working with the `tensorflow` backend. I suppose there was some issue with the unsupported dtypes functions which is why this wasn't working. I had pushed a fix for the unsupported dtypes stuff last week after which this might've started working. Thanks 😄

mattbarrett98:
> Hey @mattbarrett98 could you please try a few examples using `ivy.crosscast_data_types` or `ivy.cast_data_types`? So for e.g.,
> 
> ```
> import ivy
> 
> # can also use ivy.cast_data_types() instead which enables upcasting, downcasting as well as cross-casting
> ivy.crosscast_data_types()  
> 
> x = ivy.array([1, 2, 3])
> print(ivy.expm1(x))
> ```
> 
> gets `ivy.expm1` working with the `tensorflow` backend. I suppose there was some issue with the unsupported dtypes functions which is why this wasn't working. I had pushed a fix for the unsupported dtypes stuff last week after which this might've started working. Thanks 😄

seems to work yea, could you make the cross casting mode default then if we are going with this approach @vedpatwardhan ?

vedpatwardhan:
> > Hey @mattbarrett98 could you please try a few examples using `ivy.crosscast_data_types` or `ivy.cast_data_types`? So for e.g.,
> > ```
> > import ivy
> > 
> > ivy.set_backend("tensorflow")
> > # can also use ivy.cast_data_types() instead which enables upcasting, downcasting as well as cross-casting
> > ivy.crosscast_data_types()  
> > 
> > x = ivy.array([1, 2, 3])
> > print(ivy.expm1(x))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > gets `ivy.expm1` working with the `tensorflow` backend. I suppose there was some issue with the unsupported dtypes functions which is why this wasn't working. I had pushed a fix for the unsupported dtypes stuff last week after which this might've started working. Thanks 😄
> 
> seems to work yea, could you make the cross casting mode default then if we are going with this approach @vedpatwardhan ?

Yep, I'll create a bounty task for this as we also need to enable the casting modes by default in the tests, etc. In the meantime, feel free to enable the casting modes in the tracer so that this doesn't become a blocker on your work. Thanks @mattbarrett98 👍

mattbarrett98:
okay nice, will close this pr then thanks @vedpatwardhan

PR #28150, opened 2024-02-01
Fix the name changes required for windows binaries support

Dsantra92:
Remove the hardcoded .so to use `so` or `pyd` based on os.


vedpatwardhan (changes requested):
Hey @Dsantra92, could you please also update the `setup.py` with these changes? Thanks!

ivy/utils/binaries.py
```diff
@@ -9,8 +9,9 @@
 def _get_paths_from_binaries(binaries, root_dir=""):
```
vedpatwardhan, line 9: The changes made to this helper would also need to be replicated [here](https://github.com/unifyai/ivy/pull/28150/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R27)

vedpatwardhan (changes requested):
Hey @Dsantra92, just requested a final change, rest seems good to merge. Thanks 😄

PR #28181, opened 2024-02-05
fix: broadcasting issue in compositional set_item

mattbarrett98:
This fixes the following for tf and jax (already works for torch/numpy):

```python
import ivy

query = (slice(None, None, None), slice(None, None, None), slice(None, None, None), 1)

ivy.set_tensorflow_backend()
x = ivy.ones((1, 3, 3, 3))
val = ivy.zeros((3, 1))
x[query] = val
print(x)

ivy.set_jax_backend()
x = ivy.ones((1, 3, 3, 3))
val = ivy.zeros((3, 1))
x[query] = val
print(x)
```
I suspect the expanding dims logic maybe isn't necessary before the broadcasting but not sure.

Any ideas on if it's possibe with hypothesis to add a very specifc example to a test, like my example. I imagine it being a lot more complicated to get examples like this to appear in a property based way, but i don't think property based testing is really that important here and just testing one specific example would be sufficient. Is this possible @vedpatwardhan @Ishticode ?

### DIFF
--- a/ivy/functional/ivy/general.py
+++ b/ivy/functional/ivy/general.py
@@ -2932,16 +2932,6 @@ def _broadcast_to(input, target_shape):
         return ivy.reshape(input, target_shape)
     else:
         input = input if len(input.shape) else ivy.expand_dims(input, axis=0)
-        new_dims = ()
-        i_i = len(input.shape) - 1
-        for i_t in range(len(target_shape) - 1, -1, -1):
-            if len(input.shape) + len(new_dims) >= len(target_shape):
-                break
-            if i_i < 0 or target_shape[i_t] != input.shape[i_i]:
-                new_dims += (i_t,)
-            else:
-                i_i -= 1
-        input = ivy.expand_dims(input, axis=new_dims)
         return ivy.broadcast_to(input, target_shape)
 
 
--- a/ivy_tests/test_ivy/test_functional/test_core/test_general.py
+++ b/ivy_tests/test_ivy/test_functional/test_core/test_general.py
@@ -6,7 +6,7 @@ import math
 from types import SimpleNamespace
 
 import pytest
-from hypothesis import given, assume, strategies as st
+from hypothesis import given, assume, example, strategies as st
 import numpy as np
 from collections.abc import Sequence
 
@@ -15,7 +15,11 @@ import threading
 import ivy
 
 import ivy_tests.test_ivy.helpers as helpers
-from ivy_tests.test_ivy.helpers import handle_test, BackendHandler
+from ivy_tests.test_ivy.helpers import (
+    handle_test,
+    BackendHandler,
+    test_parameter_flags as pf,
+)
 from ivy_tests.test_ivy.helpers.assertions import assert_all_close
 from ivy_tests.test_ivy.test_functional.test_core.test_elementwise import pow_helper
 
@@ -1642,6 +1646,31 @@ def test_set_inplace_mode(mode):
     test_instance_method=st.just(False),
     container_flags=st.just([False]),
     test_with_copy=st.just(True),
+)
+@example(
+    dtypes_x_query_val=(
+        ["int32", "int32"],
+        np.ones((1, 3, 3, 3)),
+        (slice(None, None, None), slice(None, None, None), slice(None, None, None), 1),
+        np.zeros((3, 1)),
+    ),
+    copy=False,
+    fn_name="set_item",
+    test_flags=pf.FunctionTestFlags(
+        ground_truth_backend="numpy",
+        num_positional_args=3,
+        instance_method=False,
+        with_out=False,
+        with_copy=False,
+        test_gradients=False,
+        test_trace=False,
+        transpile=False,
+        as_variable=[False],
+        native_arrays=[False],
+        container=[False],
+        precision_mode=False,
+        test_cython_wrapper=False,
+    ),
 )
 def test_set_item(
     dtypes_x_query_val,


vedpatwardhan (commented):
> Any ideas on if it's possibe with hypothesis to add a very specifc example to a test, like my example. I imagine it being a lot more complicated to get examples like this to appear in a property based way, but i don't think property based testing is really that important here and just testing one specific example would be sufficient. Is this possible @vedpatwardhan @Ishticode ?

One thing we can do is use [`hypothesis.example`](https://hypothesis.readthedocs.io/en/latest/reproducing.html#hypothesis.example) which can be used alongside `@given` in our tests. Could you please check if this works with the current tests as well? Also, what do you think should be the criteria to decide when to add a strategy and when to add an explicit example for a particular example? I'll add that as an explanation to the docs as well. 
As for the changes to `set_item`, I'll leave those with @Ishticode to review as he's already working on some other issues with `get_item` and `set_item`. Thanks @mattbarrett98 😄

Ishticode (approved):
looks good

mattbarrett98:
> One thing we can do is use [`hypothesis.example`](https://hypothesis.readthedocs.io/en/latest/reproducing.html#hypothesis.example) which can be used alongside `@given` in our tests. Could you please check if this works with the current tests as well? Also, what do you think should be the criteria to decide when to add a strategy and when to add an explicit example for a particular example? I'll add that as an explanation to the docs as well. As for the changes to `set_item`, I'll leave those with @Ishticode to review as he's already working on some other issues with `get_item` and `set_item`. Thanks @mattbarrett98 😄

Thanks for the suggestion @vedpatwardhan, i've added a `hypothesis.example` as you said and it's correctly failing without my fix to `set_item`. 

As for criteria i'm not too sure. I think in this scenario it's just far simpler to add an example, and also I think accomplishes everything a full-blown strategy approach would. Since the problem in question here was some broadcasting issue, which occurs in a fairly specific situation of the shapes being a certain way, with the dtypes and values of the arrays not being relevant at all here.

This way seems far more efficient to me, as trying to change the strategies to include examples like this would probably expand the search space more than necessary for this specific example. Whereas with this approach the exact problematic example will be tested each time, for a more efficient, deterministic CI.

Let me know if i'm good to merge @vedpatwardhan 😁

vedpatwardhan:
> Thanks for the suggestion @vedpatwardhan, i've added a `hypothesis.example` as you said and it's correctly failing without my fix to `set_item`.
> 
> As for criteria i'm not too sure. I think in this scenario it's just far simpler to add an example, and also I think accomplishes everything a full-blown strategy approach would. Since the problem in question here was some broadcasting issue, which occurs in a fairly specific situation of the shapes being a certain way, with the dtypes and values of the arrays not being relevant at all here.
> 
> This way seems far more efficient to me, as trying to change the strategies to include examples like this would probably expand the search space more than necessary for this specific example. Whereas with this approach the exact problematic example will be tested each time, for a more efficient, deterministic CI.
> 
> Let me know if i'm good to merge @vedpatwardhan 😁

Yep sounds good! I'll create a task for adding a new decorator that uses `example` so that one doesn't need to specify all the test flags that should be used in the test even if they are the default values of those flags. Feel free to merge the PR, thanks @mattbarrett98 😄

mattbarrett98:
> > Thanks for the suggestion @vedpatwardhan, i've added a `hypothesis.example` as you said and it's correctly failing without my fix to `set_item`.
> > As for criteria i'm not too sure. I think in this scenario it's just far simpler to add an example, and also I think accomplishes everything a full-blown strategy approach would. Since the problem in question here was some broadcasting issue, which occurs in a fairly specific situation of the shapes being a certain way, with the dtypes and values of the arrays not being relevant at all here.
> > This way seems far more efficient to me, as trying to change the strategies to include examples like this would probably expand the search space more than necessary for this specific example. Whereas with this approach the exact problematic example will be tested each time, for a more efficient, deterministic CI.
> > Let me know if i'm good to merge @vedpatwardhan 😁
> 
> Yep sounds good! I'll create a task for adding a new decorator that uses `example` so that one doesn't need to specify all the test flags that should be used in the test even if they are the default values of those flags. There's a few errors in the tests such as [this](https://github.com/unifyai/ivy/actions/runs/7802697271/job/21282114876?pr=28181#step:4:29173), [this](https://github.com/unifyai/ivy/actions/runs/7802697271/job/21282114876?pr=28181#step:4:27600), etc., could you please confirm if the test failures aren't related to your changes? Thanks @mattbarrett98 😄

Sounds good 😁 I think you linked the same failure twice, but i checked this one and it is indeed unrelated and fails with the same without my change (although it does look very related). Also went through the some other failures that look set_item related and none were due to this change. Think we're good to merge?

mattbarrett98:
> > > Thanks for the suggestion @vedpatwardhan, i've added a `hypothesis.example` as you said and it's correctly failing without my fix to `set_item`.
> > > As for criteria i'm not too sure. I think in this scenario it's just far simpler to add an example, and also I think accomplishes everything a full-blown strategy approach would. Since the problem in question here was some broadcasting issue, which occurs in a fairly specific situation of the shapes being a certain way, with the dtypes and values of the arrays not being relevant at all here.
> > > This way seems far more efficient to me, as trying to change the strategies to include examples like this would probably expand the search space more than necessary for this specific example. Whereas with this approach the exact problematic example will be tested each time, for a more efficient, deterministic CI.
> > > Let me know if i'm good to merge @vedpatwardhan 😁
> > 
> > 
> > Yep sounds good! I'll create a task for adding a new decorator that uses `example` so that one doesn't need to specify all the test flags that should be used in the test even if they are the default values of those flags. There's a few errors in the tests such as [this](https://github.com/unifyai/ivy/actions/runs/7802697271/job/21282114876?pr=28181#step:4:29173), [this](https://github.com/unifyai/ivy/actions/runs/7802697271/job/21282114876?pr=28181#step:4:27600), etc., could you please confirm if the test failures aren't related to your changes? Thanks @mattbarrett98 😄
> 
> Sounds good 😁 I think you linked the same failure twice, but i checked this one and it is indeed unrelated and fails with the same without my change (although it does look very related). Also went through the some other failures that look set_item related and none were due to this change. Think we're good to merge @vedpatwardhan ?

vedpatwardhan:
> Sounds good 😁 I think you linked the same failure twice, but i checked this one and it is indeed unrelated and fails with the same without my change (although it does look very related). Also went through the some other failures that look set_item related and none were due to this change. Think we're good to merge @vedpatwardhan ?

Yep sure, let's get the PR merged, thanks @mattbarrett98 😄

PR #28224, opened 2024-02-08
fix: use of ivy.Shape in conditions

mattbarrett98:
fixing the `__bool__` method of `ivy.Shape`. problem is lists/tuples don't have a `__bool__` method, instead we need to call `bool()` on them for it to work. this breaks the use of an `ivy.Shape` object in conditions, and added a test which shows what this fixes @vedpatwardhan

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -294,7 +294,7 @@ class Shape(Sequence):
         return self
 
     def __bool__(self):
-        return self._shape.__bool__()
+        return builtins.bool(self._shape)
 
     def __div__(self, other):
         return self._shape // other
--- a/ivy_tests/test_ivy/test_misc/test_shape.py
+++ b/ivy_tests/test_ivy/test_misc/test_shape.py
@@ -1,6 +1,7 @@
 from hypothesis import assume, strategies as st
 import numpy as np
 
+import ivy
 import ivy_tests.test_ivy.helpers as helpers
 
 from ivy_tests.test_ivy.helpers import handle_method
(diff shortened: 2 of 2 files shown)


vedpatwardhan (approved):

mattbarrett98:
@vedpatwardhan there is a recent pr from @fnhirwa on the same method: https://github.com/unifyai/ivy/commit/9789ea5392d9585d65acfa308a9907fb50a120d5.  I've overridden with my change as I don't believe this was the right fix. The method now tries to extract the first element out of the shape, and check the truth value of this element, which i don't think is quite right.

I think it would coincidentally give the right answer most of the time, but one example where it fails is the shape of a 0 dimensional array would be `ivy.Shape(())` ie an empty tuple. Trying to extract the 0th element from this would fail.

fnhirwa:
> @vedpatwardhan there is a recent pr from @fnhirwa on the same method: [9789ea5](https://github.com/unifyai/ivy/commit/9789ea5392d9585d65acfa308a9907fb50a120d5). I've overridden with my change as I don't believe this was the right fix. The method now tries to extract the first element out of the shape, and check the truth value of this element, which i don't think is quite right.
> 
> I think it would coincidentally give the right answer most of the time, but one example where it fails is the shape of a 0 dimensional array would be `ivy.Shape(())` ie an empty tuple. Trying to extract the 0th element from this would fail.

Hey @mattbarrett98 this makes sense, I'm going to revert this commit🤞

vedpatwardhan:
> @vedpatwardhan there is a recent pr from @fnhirwa on the same method: [9789ea5](https://github.com/unifyai/ivy/commit/9789ea5392d9585d65acfa308a9907fb50a120d5). I've overridden with my change as I don't believe this was the right fix. The method now tries to extract the first element out of the shape, and check the truth value of this element, which i don't think is quite right.
> 
> I think it would coincidentally give the right answer most of the time, but one example where it fails is the shape of a 0 dimensional array would be `ivy.Shape(())` ie an empty tuple. Trying to extract the 0th element from this would fail.

Yeah makes sense, thanks for pointing that out @mattbarrett98 😄

Issue #28232, opened 2024-02-09
 Fix Ivy Failing Test: paddle - activations.relu


fleventy-5:
To-do List:https://github.com/unifyai/ivy/issues/27501

PR #28233, opened 2024-02-09
Fix Ivy Failing Test: paddle - activations.relu for all backend

fleventy-5:
# PR Description

500 test cases passed locally for all backend
Screenshot from 2024-02-09 18-10-42

## Related Issue

Closes https://github.com/unifyai/ivy/issues/28232

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?


ivy/functional/backends/paddle/activations.py
```diff
@@ -22,7 +22,12 @@
 
 
 @with_supported_dtypes(
-    {"2.6.0 and below": ("float32", "float64", "complex")},
+    {
+        "2.6.0 and below": (
+            "float32",
+            "float64",
+        )
```
samthakur587, line 29:
```suggestion
        the test is failed at paddle on complex dtype. so we should try to pass the test on complex dtype. so in paddle supported dtype we should have the complex dtype.
```

samthakur587 (changes requested):
handle the complex dtype at paddle backend.

fleventy-5:
Did as you've asked and tested again found out that  the error for complex numbers happen due to cases where `complex_mode=st.sampled_from(["jax","split","magnitude"]),` was set to jax so should I add a check at the paddle backend or should i just removed jax from the list of stategy to sample from @samthakur587

samthakur587:
> Did as you've asked and tested again found out that the error for complex numbers happen due to cases where `complex_mode=st.sampled_from(["jax","split","magnitude"]),` was set to jax at the main ivy test_activation.py file so should I add a check at the paddle backend or should i just removed jax from the list of stategy to sample from @samthakur587

No we should not remove the jax from `complex_mode=st.sampled_from(["jax","split","magnitude"])` according to the `_handle_complex_input` handling the complex data. if it's 'jax' then  it's calling the function based on the backend like here the `paddle` and call the relu impl. for complex dtype. as i have seen the we don't have the correct implementation for complex dtype at paddle backend. so we should try to fix this. @fleventy-5

fleventy-5:
Screenshot from 2024-02-10 21-08-54 @samthakur587 just to make sure the implementation to handle complex dataype should follow this comment  which I found in the ivy main activation.py file

fleventy-5:
@samthakur587  did the fix and found out the root cause of the problem complex128 dtype which is not supported in paddle I think
Pass 500 cases tested locally for paddle backend
Screenshot from 2024-02-10 21-24-04
Also tested 200 cases  for all backend just to be sure that the fix didn't break other backend
Screenshot from 2024-02-10 21-31-16

samthakur587:
Hii @fleventy-5 can you please solve the merge conflict to run the CI.

fleventy-5:
> Hii @fleventy-5 can you please solve the merge conflict to run the CI.

@samthakur587 done

ivy/functional/backends/paddle/activations.py
```diff
@@ -22,10 +22,18 @@
...
+            "complex64",
+        )
+    },
     backend_version,
 )
-def relu(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.Tensor:
+def relu(
+    x: paddle.Tensor, /, *, complex_mode="jax", out: Optional[paddle.Tensor] = None
+) -> paddle.Tensor:
     if paddle.is_complex(x):
```
samthakur587, line 37:
```python
import ivy
import paddle
import jax
x = paddle.to_tensor([1-1j, 1+1j])
y = jax.numpy.array([1-1j, 1+1j])
#paddle output: ivy.array([1.+0.j, 1.+1.j])
print(ivy.relu(x))
#jax output: ivy.array([1.-1.j, 1.+1.j])
print(ivy.relu(y))
```
as the logic for the complex dtype at  paddle is `return paddle.complex(F.relu(x.real()), F.relu(x.imag()))` it's dosen't pass this case bcz the x.imag() part is -ve and if we apply the `relu` activation function then it returns the `zero` that is the wrong logic here for paddle backend for complex dtype.

samthakur587 (changes requested):
Hii @fleventy-5 sorry for the late reply. i have added the comment what changes are required.

fleventy-5:
@samthakur587  complemented the implementation at the paddle back-end in the below code 
```
def relu(
    x: paddle.Tensor, /, *, complex_mode="jax", out: Optional[paddle.Tensor] = None
) -> paddle.Tensor:
    x_dtype = str(x.dtype).split('.')[1]
    if paddle.is_complex(x):
        if complex_mode=="jax":
            x_real = paddle.real(x)
            x_imag = paddle.imag(x)
    
            condition = paddle.logical_or(
               paddle.logical_and(x_real <= 0, x_imag < 0)
               , x_real < 0)
            
            x_imag = paddle.where(condition, paddle.zeros_like(x_imag), x_imag)  
            return paddle.complex(F.relu(x_real),x_imag)
        
        return paddle.complex(F.relu(x.real()), F.relu(x.imag()))
    return F.relu(x)

```
but there is still issue with complex128 
Screenshot from 2024-02-18 17-29-21
 and when I tried the function all by itself on isolation using the above function definition it was working fine as shown below
Screenshot from 2024-02-18 17-26-48

Ishticode (commented):
Hi @fleventy-5 
Thanks for the PR. Merging this now. 
@samthakur587 Could you fix the value mismatches for some of the complex vlue cases that you outlined in paddle :)
Thank you both

PR #28235, opened 2024-02-10
Update ndarray.py

ramyamanasa:
I have added the setfield method in the ndarray section of Numpy Frontend.
Please do review it and let me t\know if should make any appropriate changes.


NiteshK84:
Hi @ramyamanasa,
along with the function also please add the unit test in corresponding test file and let me know afterwords.
Thanks

Issue #27501, opened 2023-12-10
Failing Tests - Ivy Functional API


ivy-leaves:
# Failing Tests
> Please see the failing tests divided into priority sections below. Click on each section to expand. Feel free to claim an issue to fix by commenting `@ivy-leaves assign me` on the linked issue.

niixxaaa:
#27838

NripeshN:
- [ ] #27902

samthakur587:
#27930

samthakur587:
- [ ] #27930

samthakur587:
- [ ] #27932

MuhammadNizamani:
#27935

samthakur587:
- [ ] #27936

samthakur587:
- [ ] #27947

samthakur587:
- [ ] #27957

samthakur587:
- [ ] #27959

samthakur587:
- [ ] #27961

samthakur587:
- [ ] #27970

samthakur587:
- [ ] #27969

samthakur587:
- [ ] #27968

samthakur587:
- [ ] #27967

samthakur587:
- [ ] #27966

samthakur587:
- [ ] #27972

Sai-Suraj-27:
- [ ] https://github.com/unifyai/ivy/issues/27979

Sai-Suraj-27:
https://github.com/unifyai/ivy/issues/27980

Sai-Suraj-27:
- [ ] https://github.com/unifyai/ivy/issues/27981

Sai-Suraj-27:
- [ ] https://github.com/unifyai/ivy/issues/27982

Sai-Suraj-27:
- [ ] https://github.com/unifyai/ivy/issues/27983

Sai-Suraj-27:
- [ ] https://github.com/unifyai/ivy/issues/27987

Vismay-dev:
- [ ] #27993

MuhammadNizamani:
#27999

NripeshN:
- [ ] #28001

MuhammadNizamani:
#28002

Vismay-dev:
- [ ] #28006

Vismay-dev:
- [ ] #28007

Vismay-dev:
[...]

PR #28251, opened 2024-02-12
feat: Added a decorator function, which will be used to test `explicit examples`

### LINKED ISSUE #28207: Adding a way to enable explicit `@example` in our tests
Currently most tests in `ivy` and the frontends are property-based, wherein we define [strategies](https://github.com/unifyai/ivy/blob/2c465d6de04d04bdee0d7b07ede0bbbeb66b69b0/ivy_tests/test_ivy/helpers/hypothesis_helpers/array_helpers.py#L287) for generating data inside our tests (more explanations [here](https://unify.ai/docs/ivy/overview/deep_dive/ivy_tests.html)).
Now, in certain cases where we'd like to test certain examples explicitly which are outliers and it isn't feasible to define them as a strategy. One such example is [this](https://github.com/unifyai/ivy/pull/28181/files#diff-a13fff36fcac31f1f8fa76cb78fceb815348b58f0b03358bbf471bd355416ee8R1651) where we need to test `ivy.set_item` with `slice` objects. 

`hypothesis` allows us to test with an explicit example deterministically using the [`@example`](https://hypothesis.readthedocs.io/en/latest/reproducing.html#hypothesis.example) decorator. One of the challenges currently is that we need to define all the [flags](https://github.com/unifyai/ivy/pull/28181/files#diff-a13fff36fcac31f1f8fa76cb78fceb815348b58f0b03358bbf471bd355416ee8R1659) as well even if we're using the default values for them. What we should instead have is a way to define those flags only if their values are different from a default we set for all these flags. 

This task would involve adding a new decorator to the tests (similar to how we have [`handle_test`](https://github.com/unifyai/ivy/blob/2c465d6de04d04bdee0d7b07ede0bbbeb66b69b0/ivy_tests/test_ivy/helpers/testing_helpers.py#L326), [`handle_frontend_test`](https://github.com/unifyai/ivy/blob/2c465d6de04d04bdee0d7b07ede0bbbeb66b69b0/ivy_tests/test_ivy/helpers/testing_helpers.py#L470), etc.) which would just wrap the function with `@example` but would use the default values of these flags similar to what `handle_test` and `handle_frontend_test` do (ofc in this case the default value would be a python str, list, bool, etc. instead of a hypothesis strategy that generates these values). Also, necessary generalizations should be made to the newly created decorator so that a single decorator with the tests of functions and methods both on ivy and the frontends when trying to specify explicit examples.

You should finally update [this](https://github.com/unifyai/ivy/pull/28181/files#diff-a13fff36fcac31f1f8fa76cb78fceb815348b58f0b03358bbf471bd355416ee8R1650) implementation using the new decorator and write up a sub-section under [this](https://unify.ai/docs/ivy/overview/deep_dive/ivy_tests.html#id2) section of the docs explaining how this decorator should be used.

Feel free to reach out for any questions!

Sai-Suraj-27:
# PR Description
Added a decorator function, which will be used to test `explicit examples` that are difficult/not feasible to define as strategies.
The decorator is also generalized so that we can test all 4 `ivy tests`, `frontend tests`, `methods`, `frontend methods` with explicit examples.

## Related Issue
Closes #28207

## Checklist

- [x] Did you add a function?
- [ ] Did you add the tests?


vedpatwardhan (commented):
Hey @Sai-Suraj-27, could you please test this out for frontend tests and methods as well? Probably just add the decorator to a function from those categories temporarily as well so that we can confirm in the CI if that example ran correctly?
The changes look great to me otherwise, thanks for looking into this 😄

Sai-Suraj-27:
> Hey @Sai-Suraj-27, could you please test this out for frontend tests and methods as well? Probably just add the decorator to a function from those categories temporarily as well so that we can confirm in the CI if that example ran correctly? The changes look great to me otherwise, thanks for looking into this 😄

Hey, @vedpatwardhan. I actually did that already, but didn't committed those changes. Sure, I will add few more example decorators 👍🏻.

Sai-Suraj-27:
Hey, @vedpatwardhan. I see that the following 2 tests are mentioned as passing in main (main: pass) but failing after these changes, in the CI.
1. https://github.com/unifyai/ivy/actions/runs/7884526822/job/21515049892?pr=28251#step:6:34
2. https://github.com/unifyai/ivy/actions/runs/7884526822/job/21515049892?pr=28251#step:6:127

But when I tested locally, the first 2nd one is passing all the backends successfully. 

image

The first one is already failing on main (present [here](https://github.com/unifyai/ivy/issues/27501) as currently one of the failing tests). So, I'm not sure, why they are shown in the CI. Just thought of clarifying. Thank you.

vedpatwardhan (approved):
lgtm! Feel free to revert the changes with the dummy `handle_example` and merge the PR (the 2 test failures that are passing on `main` seem unrelated to the changes made)
Thanks @Sai-Suraj-27 😄

PR #28252, opened 2024-02-12
fix: undocumented args for torch frontend- added `axis` support for `stack`, and `value` support for `scatter_` + extended tests

mattbarrett98:
This adds support for behaviour which works for native torch i.e.  `torch.stack` allows the `dim` argument to also be passed using keyword `axis`, although it isn't in the documentation it does support this and so occasionally pops up in botorch.

### DIFF
--- a/ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
+++ b/ivy/functional/frontends/torch/indexing_slicing_joining_mutating_ops.py
@@ -398,7 +398,9 @@ def squeeze(input, dim=None):
 
 
 @to_ivy_arrays_and_back
-def stack(tensors, dim=0, *, out=None):
+def stack(tensors, dim=0, *, out=None, axis=None):
+    if axis is not None:
+        dim = axis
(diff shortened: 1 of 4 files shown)


vedpatwardhan:
Oh sorry, just noticed that `test_torch_scatter_` seems to be failing in the [CI](https://github.com/unifyai/ivy/actions/runs/7873981485/job/21484066426?pr=28252#step:4:5447). Could you please confirm if the failures are related to the changes made? Thanks!

hmahmood24:
> This adds support for behaviour which works for native torch i.e. `torch.stack` allows the `dim` argument to also be passed using keyword `axis`, although it isn't in the documentation it does support this and so occasionally pops up in botorch.

@mattbarrett98 There's a [numpy_to_torch_style_args](https://github.com/unifyai/ivy/blob/8dcc33b895240395686db165c710ac31708aa691/ivy/functional/frontends/torch/func_wrapper.py#L154) decorator in the `torch` frontend for this purpose I suppose

PR #28296, opened 2024-02-15
fix: torch frontend linspace working with `dtype=None` + array inputs. and extended test

mattbarrett98:
fixes frontend linspace to use the default dtype when no dtype is specified, rather than getting the dtype of one of the input arrays, which isn't the desired behaviour.


vedpatwardhan (approved):

PR #28298, opened 2024-02-15
fix: frontend searchsorted should only fail when `side` is explicitly set to `'left'` + generalised test

mattbarrett98:


vedpatwardhan (approved):

PR #28394, opened 2024-02-22
fix: removed code modifying ndarray/torch tensor methods (caused an issue with torch.compile recognising dunder methods). but added the same functionality using `__torch_function__`/`__array_ufunc__` which are the proper mechanisms to enable this behaviour. also simplified the test

mattbarrett98:
I think this code in all the backend init files is very hacky and should be avoided. the torch code specifically was causing an issue with torch.compile so needed to remove this, and also removed for numpy. There were much simpler ways to get the desired behaviour for torch/numpy using `__torch_function`/`__array_ufunc__`, though i don't think the equivalent mechanisms exist in the other frameworks so i have left them for now


vedpatwardhan (approved):

PR #28470, opened 2024-03-02
fix: fixing the tests for `ivy.Shape` method

fnhirwa:
# PR Description

Hi, @vedpatwardhan and @Ishticode I walked through the `ivy.Shape` class and given that shape implements the `Sequence` like behaviour. I believe that some of the magic methods implemented for this class always raise `TypeError` and I decided to remove them as they are not needed for the implementation.
Removed methods
* `__mod__`
*  `__rmod__`
*  `__rdiv__`
*  `__div__`
*  `__sub__`
*  `__rsub__`
*  `__int__`
*  `__floordiv__`

After removing these methods I refactored all unit tests of the `ivy.Shape` methods as most of them were failing due to wrong passing of parameters.

## Related Issue

Closes #

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -308,12 +308,6 @@ class Shape(Sequence):
             return builtins.bool(builtins.tuple(self._shape))
         return builtins.bool(self._shape)
 
-    def __div__(self, other):
-        return self._shape // other
-
-    def __floordiv__(self, other):
-        return self._shape // other
-
     def __reduce__(self):
         return (self.__class__, (self._shape,))
 
--- a/ivy/__init__.py
+++ b/ivy/__init__.py
@@ -314,9 +314,6 @@ class Shape(Sequence):
     def __floordiv__(self, other):
         return self._shape // other
 
-    def __rmod__(self, other):
-        return other % self._shape
-
     def __reduce__(self):
         return (self.__class__, (self._shape,))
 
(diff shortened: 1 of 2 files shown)


vedpatwardhan (commented):
Hey @fnhirwa, thanks for creating the PR!
Actually these methods were primarily added because the methods exist in `torch.Size`, `tf.TensorShape`, etc. Could you please check if those classes have these methods? If they don't then ofc we can remove them from the `ivy.Shape` class 😄

fnhirwa:
> Hey @fnhirwa, thanks for creating the PR! Actually these methods were primarily added because the methods exist in `torch.Size`, `tf.TensorShape`, etc. Could you please check if those classes have these methods? If they don't then ofc we can remove them from the `ivy.Shape` class 😄

Yeah makes sense,
I did a quick walkthrough and found that the methods in `torch.Size` that are not implemented are 
`numel` and `__getnewargs__` and for the `tf.TensorShape` are `most_specific_common_supertype`, `cast`, `__concat__`, `is_subtype_of`, `_tf_api_names_v1`, `experimental_from_proto`, `flatten`, `_v2_behavior`, `experimental_as_proto`, `assert_is_compatible_with`, `_dims`, `merge_with`, `experimental_type_proto`, `__nonzero__`, `as_proto`, `most_specific_compatible_shape`, `placeholder_value`, `to_tensors`, `from_tensors`,  and `_tf_api_names`

So I think these are the ones we can add superset behaviour for🙄

vedpatwardhan (approved):
lgtm! Feel free to merge, thanks @fnhirwa 😄

PR #28473, opened 2024-03-03
fix: setting torch frontend tensor 'data' attribute

Sam-Armstrong:
Fixes setting the `data` attribute of a frontend tensor to match the behaviour of native torch, such as this minimal example which prior to this fix would throw `AttributeError: can't set attribute 'data'` for the frontend:
```
import torch

x = torch.randn((1, 3, 3))
y = torch.randn((1, 3, 3))
x.__setattr__('data', y)
print(x)

x = torch_frontend.randn((1, 3, 3))
y = torch_frontend.randn((1, 3, 3))
x.__setattr__('data', y)
print(x)
```
Does this look ok to you @Ishticode?

### DIFF
--- a/ivy/functional/frontends/torch/tensor.py
+++ b/ivy/functional/frontends/torch/tensor.py
@@ -49,7 +49,7 @@ class Tensor:
 
     def __setattr__(self, name, value):
         if name == "data":
-            self = value
+            pass
         else:
             super().__setattr__(name, value)
 
--- a/ivy/functional/frontends/torch/tensor.py
+++ b/ivy/functional/frontends/torch/tensor.py
@@ -47,6 +47,12 @@ class Tensor:
     def __hash__(self):
         return id(self)
 
+    def __setattr__(self, name, value):
+        if name == "data":
+            self = value
+        else:
+            super().__setattr__(name, value)
+
     # Properties #
     # ---------- #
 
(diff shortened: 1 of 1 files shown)


Sam-Armstrong:
@Ishticode also, if this looks ok is there a way to stop the linter changing the code? 😅

Ishticode (commented):
was just playing with this. 
One question i have is why do we want to just not do anything when data attribute is being updated. torch's native behaviour shows that the data changes. You can slightly modify this script to see. Whereas torch frontend data remains same before and after this.

```python
print("native torch \n")
x = torch.randn((1, 3, 3))
y = torch.randn((1, 3, 3))
print("x.data before setattr:\n", x.data)
x.__setattr__('data', y)
print("x.data after setattr:\n", x.data)

print("ivy torch frontend\n")
x = torch_frontend.randn((1, 3, 3))
y = torch_frontend.randn((1, 3, 3))
print("x.data before setattr:\n")
print(x.data)
x.__setattr__('data', y)
print("x.data after setattr:\n")
print(x.data)
```

hmmm it does fix the issues you mentioned but it leads to this new subtle issue which might be more troublesome elsewhere

Sam-Armstrong:
@Ishticode my fix didn't have the `pass`, the linter added that for some reason

PR #28477, opened 2024-03-04
fix: torch.Tensor.to frontend failing with certain arg/kwarg combinations

Sam-Armstrong:
The frontend `to` method currently fails when passed the args (not kwargs) (None, dtype) which works fine for native torch. Here I generalise the torch.Tensor.to frontend to fix this problem, and also fix the tests for all backends aside paddle.


Ishticode (approved):
lgtm :)

PR #28481, opened 2024-03-04
fix: tf.function working with set item

mattbarrett98:


ivy_tests/test_ivy/test_functional/test_core/test_general.py
```diff
@@ -1875,6 +1875,21 @@ def test_tensorflow_get_item_condition():
     assert _get_item_condition(None, query)
 
 
+def test_tf_function_set_item():
```
vedpatwardhan, line 1878: let's probably add the input data to the `handle_example` decorator on `test_set_item`, this (along with all other generated examples from the strategy) would get tested with the tracer in the tracer tests workflow. Rest looks good to me, thanks @mattbarrett98 😄

vedpatwardhan (changes requested):

vedpatwardhan (approved):

PR #28494, opened 2024-03-06
fix: ensure traced function return is turned to ivy arrays

Ishticode:
This fixes the issues with `is_ivy_array` check in test_functions flattening process


ivy_tests/test_ivy/helpers/function_testing.py
```diff
@@ -2580,6 +2580,8 @@ def get_ret_and_flattened_np_array(
     with BackendHandler.update_backend(backend_to_test) as ivy_backend:
         with ivy_backend.PreciseMode(precision_mode):
             ret = fn(*args, **kwargs)
+            if test_trace or test_trace_each:
+                ret = ivy.to_ivy(ret, True)
```
vedpatwardhan, line 2584: this should be `ivy_backend` instead of `ivy`, rest looks good to merge, thanks @Ishticode 😄

Issue #28495, opened 2024-03-06
Fix Frontend Failing Test: paddle - math.paddle.diff


ZJay07:
To-do List: https://github.com/unifyai/ivy/issues/27500

PR #28497, opened 2024-03-06
Fix Frontend Failing Test: paddle - math.paddle.diff

ZJay07:
# PR Description
2 errors:
1. dtype error, append, prepend and x are different dtypes hence causing an error, fix by casting append and preppend to the same dtype as x
2. dim error x, append and prepend should be the same dim, fixed by making max_dim of x to be the same as the other 2

## Related Issue

Close https://github.com/unifyai/ivy/issues/28495
Close https://github.com/unifyai/ivy/issues/28527
Close https://github.com/unifyai/ivy/issues/28526
Close https://github.com/unifyai/ivy/issues/28525
Close https://github.com/unifyai/ivy/issues/28524

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_paddle/test_math.py
+++ b/ivy_tests/test_ivy/test_frontends/test_paddle/test_math.py
@@ -838,46 +838,29 @@ def test_paddle_diagonal(
 # diff
 @handle_frontend_test(
     fn_tree="paddle.diff",
-    dtype_n_x_n_axis=helpers.dtype_values_axis(
-        available_dtypes=st.shared(helpers.get_dtypes("valid"), key="dtype"),
+    dtype_n_x_prepend_append=helpers.dtype_values_axis(
+        available_dtypes=helpers.get_dtypes("valid"),
+        num_arrays=3,
         min_num_dims=1,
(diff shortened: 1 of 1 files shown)


Ishticode (changes requested):
Hi @ZJay07 
I think restriciting max number of dims to be also 1 like min num of dims means that we are testing for only 1d arrays while paddles diff function should also be able to run multidimensional arrays. So can we put a higher limit on if it helps maybe 5 or 6?

ZJay07:
@Ishticode it should worked now for multi dimensions :)

Ishticode (commented):
We should not be needing to do so much to get dim and dtype right for append and prepend after they are generated. Can you see if you can instead generate them right to begin with? like in the helpers used. There is a shared strategy used here but that doesn't seem to get the same dtype? maybe use a single array generate with num_arrays equal to two or something.
Let me know if something is not clear

ZJay07:
Just fixed it to generate an array, let me know if this is the right fix!

Ishticode (approved):
That looks good. Thank you very much :) @ZJay07

PR #28323, opened 2024-02-18
fix: Frontend Failing Test: torch - linalg.torch.linalg.matrix_norm

Kacper-W-Kozdon:
# PR Description
- Added type-casting to native dtypes in backends.
- Added casting to the argument provided in `dtype` to backends to reproduce the behaviour of torch frontend.
- Changed `min_value` and `max_value` in tests to `large_abs_safety_factor=1.5, small_abs_safety_factor=1.5, safety_factor_scale="log"`.

## Related Issue

Closes #28314

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?
- [x] Did you run your tests and are your tests passing?
- [x] Did pre-commit not fail on any check?
- [x] Did you follow the steps we provided?

PR #28500, opened 2024-03-07
fix: ivy.linalg.matrix_norm() for paddle backend

### LINKED ISSUE #28314: Fix Frontend Failing Test: torch - linalg.torch.linalg.matrix_norm
At the first glance, the possible reasons for errors:
- lack of `@with_supported_dtypes`/type-casting in some backends or unnecessary type-casting in the tests,
- lack of set tolerances and/or no bounds on values, possibly leading to the generation of values too big for the dtype,
- errors in the backend implementations (?) or some backends having limited implementations (not working at all for some dtypes).

Kacper-W-Kozdon:
# PR Description
- Type-casting to the `dtype` argument through the `@infer_dtype` decorator and `ivy.astype(x, dtype).as_native()`.
- Added `dtype` argument to array and container methods.
- Added type-casting to `float32` for the `out` argument passed down to torch backend- `out` argument has to be of the same type as the return of `torch.linalg.matrix_norm()` (`float32`).
- Test coverage for all valid dtypes.

**Current issues**:
- Gradient tests for paddle have problems with `max()` and `min()` functions (I will write just for `max()`) due to different implementations between backends (look up `paddle.max()` vs `paddle.amax()` or `torch.max()` vs `torch.amax()`). Tensorflow, numpy, torch and jax pass all the tests alike. 

## Related Issue
Continuation of the issue https://github.com/unifyai/ivy/issues/28314 and PR https://github.com/unifyai/ivy/pull/28323

Closes #28314 

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?


Kacper-W-Kozdon:
@vedpatwardhan @Ishticode it might be easier to work with the code when there is a pull request (I'm not sure what kind of issue it should be opened as though, as it is not any of the open sub tasks).

ivy/functional/backends/numpy/linear_algebra.py
```diff
@@ -152,10 +152,13 @@ def matrix_norm(
     out: Optional[np.ndarray] = None,
 ) -> np.ndarray:
     if dtype is not None:
-        x = ivy.astype(x, dtype)
+        x = ivy.astype(x, dtype).to_native()
     if not isinstance(axis, tuple):
         axis = tuple(axis)
-    return np.linalg.norm(x, ord=ord, axis=axis, keepdims=keepdims)
+    ret = np.linalg.norm(x, ord=ord, axis=axis, keepdims=keepdims)
+    if ivy.exists(out):
```
vedpatwardhan, line 159: I think this should get handled automatically through the `handle_out_argument` decorator, could you please check if that's the case? Same applies to the other backends

ivy/functional/ivy/linear_algebra.py
```diff
@@ -1129,14 +1130,15 @@ def matmul(
...
 
-@handle_exceptions
-@handle_backend_invalid
-@handle_nestable
-@handle_array_like_without_promotion
-@handle_out_argument
-@to_native_arrays_and_back
-@handle_array_function
 @handle_device
+@infer_dtype
```
vedpatwardhan, line 1134: The order of the relevant decorators applied to the function should be the reverse of the order [here](https://github.com/unifyai/ivy/blob/36a36bf1da65fa6f6b1c4077b2a64d5e0904ac8a/ivy/func_wrapper.py#L17)

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -643,27 +643,39 @@ def test_matmul(*, x, y, test_flags, backend_fw, fn_name, on_device):
...
         min_num_dims=2,
         valid_axis=True,
         min_axes_size=2,
         max_axes_size=2,
+        max_value=1e4,
+        min_value=-1e4,
+        abs_smallest_val=1e-4,
         force_tuple_axis=True,
         allow_neg_axes=False,
+        safety_factor_scale="log",
```
vedpatwardhan, line 659: as `min_value`, `max_value`, etc. are specified, the `safety_factor` arguments isn't required

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
```diff
@@ -643,27 +643,39 @@ def test_matmul(*, x, y, test_flags, backend_fw, fn_name, on_device):
...
 def test_matrix_norm(
-    *, dtype_value_axis, kd, ord, test_flags, backend_fw, fn_name, on_device
+    *, dtype_value_axis, kd, ord, dtypes, test_flags, backend_fw, fn_name, on_device
 ):
-    dtype, x, axis = dtype_value_axis
+    input_dtype, x, axis = dtype_value_axis
+    if dtypes[0] is not None:
+        dtypes[0] = input_dtype[0][0:-2] + max([input_dtype[0][-2:], dtypes[0][-2:]])
+
+    test_flags.test_gradients = True
```
vedpatwardhan, line 674: let's not hardcode `test_gradients` to `True` for each test and instead pass the one that was generated by the strategy

vedpatwardhan (commented):
Hey @Kacper-W-Kozdon, thanks for creating the PR, just requested a few changes. Let's see what @Ishticode thinks 😄

ivy_tests/test_ivy/test_functional/test_core/test_linalg.py
Kacper-W-Kozdon: Fixed 👍

ivy/functional/ivy/linear_algebra.py
Kacper-W-Kozdon: I took the order from here: https://unify.ai/docs/ivy/overview/deep_dive/function_wrapping.html so I'm guessing that deep-dive might need an update. Either way- fixed!

ivy/functional/backends/numpy/linear_algebra.py
Kacper-W-Kozdon: You're 100% correct, I wasn't completely sure about that. Fixed! There are some small changes in torch backend related to the `out` argument as well- with said changes, now all the supported dtypes work correctly (not just float).

Kacper-W-Kozdon:
I think other frontends had some issues as well (they were failing the tests), so if this PR is ready to be merged, I can make the necessary changes there as well. This particular PR solves the issues of the core ivy, ivy.Array, ivy.Container and torch frontend.

vedpatwardhan (commented):
let's try the logic you mentioned (with using `max` and then dividing it by the size of the last dim) and see if it solves the gradient problem with `paddle`. Thanks @Kacper-W-Kozdon 😄

Kacper-W-Kozdon:
> let's try the logic you mentioned (with using `max` and then dividing it by the size of the last dim) and see if it solves the gradient problem with `paddle`. Thanks @Kacper-W-Kozdon 😄

From https://unify.ai/docs/ivy/overview/deep_dive/gradients.html

> JAX treats duplicate arrays as distinct while computing gradients, so we need additional logic to replicate gradients computed w.r.t one array over all its duplicates.

is this already implemented? If so, in which function is it? Considering the differences in the implementation of how paddle vs other backends handle `amax` for some functions, the solution from JAX might help, too.

Kacper-W-Kozdon:
Hey @vedpatwardhan , new observation (which might be a coincidence but might also point to the issue with paddle backend)- whenever `ivy.linalg.matrix_norm()` for `ord in [-float("inf"), -2, -1, 1, 2, float("inf")]` returns rubbish gradients when the backend uses `amax`, it returns the correct results with `max`; and the other way around is true (and happens) as well- when `max` returns wrong results, `amax` returns the correct ones. I have *zero* clue what is wrong with paddle but it's not the fault of ivy's implementation of gradients or anything else- I have been checking it with raw paddle to compare the results to ivy's ones on the side as a sanity check. My suggestion would be either opening an issue directly in the paddle github or modifying how `amax` is handled in the ivy backend. The fact that when the gradient of `amax` returns rubbish results, `max` works, suggests, that the problem occurs when there are no multiples in the outputs of `max`.

vedpatwardhan:
> Hey @vedpatwardhan , new observation (which might be a coincidence but might also point to the issue with paddle backend)- whenever `ivy.linalg.matrix_norm()` for `ord in [-float("inf"), -2, -1, 1, 2, float("inf")]` returns rubbish gradients when the backend uses `amax`, it returns the correct results with `max`; and the other way around is true (and happens) as well- when `max` returns wrong results, `amax` returns the correct ones. I have _zero_ clue what is wrong with paddle but it's not the fault of ivy's implementation of gradients or anything else- I have been checking it with raw paddle to compare the results to ivy's ones on the side as a sanity check. My suggestion would be either opening an issue directly in the paddle github or modifying how `amax` is handled in the ivy backend. The fact that when the gradient of `amax` returns rubbish results, `max` works, suggests, that the problem occurs when there are no multiples in the outputs of `max`.
> 
> **EDIT:** If you wish, I can prepare a jupyter notebook with an explanation, because it seems that the results might be salvageable. In fact, maybe even without touching gradients, but that would require noticable changes in the paddle backend, either directly in the `min()`/`max()` functions or just in the `matrix_norm()`.

I'd say let's create a minimal example of the incorrect gradients and create an issue on the paddle repo regarding this, but probably what we'll get as the reply is that `inf` is the problem. Thanks for the efforts so far @Kacper-W-Kozdon 😄

Kacper-W-Kozdon:
Hey @vedpatwardhan , I implemented the workaround, it works- with `--num-examples=500` all of the tests passed (including gradient tests) 😄 Could you double-check it?

vedpatwardhan (changes requested):
Hey @Kacper-W-Kozdon, seems like there's a few other test failures that have arised as a result of these changes (e.g. [this](https://github.com/unifyai/ivy/actions/runs/8249748795/job/22564884184?pr=28500#step:4:22654)), could you please take a look at this and other failures that are pointed in the workflow if any of them are related? Thanks 😄

Kacper-W-Kozdon:
> Hey @Kacper-W-Kozdon, seems like there's a few other test failures that have arised as a result of these changes (e.g. [this](https://github.com/unifyai/ivy/actions/runs/8249748795/job/22564884184?pr=28500#step:4:22654)), could you please take a look at this and other failures that are pointed in the workflow if any of them are related? Thanks 😄

Oh! My bad, thanks for checking that thoroughly- the changes to `min()` and `max()` caused problems in other functions because I did not consider `NoneType` input for the `axis` argument 👍should be fixed now but we'll see what the workflow shows.

vedpatwardhan (approved):
lgtm, there don't seem to be any additional failures introduced. Feel free to merge the PR @Ishticode, thanks @Kacper-W-Kozdon 😄

Issue #28541, opened 2024-03-11
Fix Frontend Failing Test: jax - math.tensorflow.math.reduce_prod


ZJay07:
To-do List: https://github.com/unifyai/ivy/issues/27496

Issue #28540, opened 2024-03-11
Fix Frontend Failing Test: numpy - math.tensorflow.math.reduce_prod


ZJay07:
To-do List: https://github.com/unifyai/ivy/issues/27497

Issue #28539, opened 2024-03-11
Fix Frontend Failing Test: torch - math.tensorflow.math.reduce_prod


ZJay07:
To-do List: https://github.com/unifyai/ivy/issues/27498

PR #28542, opened 2024-03-11
Fix Frontend Failing Test: jax, numpy, torch - math.tensorflow.math.reduce_prod

ZJay07:
# PR Description
Test was failing because of small floating point rounding difference and too small/large values causing the results to be `inf` in some frameworks.
Fixed by adding tolerance and min max x values.

## Related Issue

Close https://github.com/unifyai/ivy/issues/28541
Close https://github.com/unifyai/ivy/issues/28540
Close https://github.com/unifyai/ivy/issues/28539

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
+++ b/ivy_tests/test_ivy/test_frontends/test_tensorflow/test_math.py
@@ -2432,6 +2432,8 @@ def test_tensorflow_reduce_min(
     fn_tree="tensorflow.math.reduce_prod",
     dtype_and_x=helpers.dtype_and_values(
         available_dtypes=helpers.get_dtypes("numeric"),
+        min_value=-5,
+        max_value=5,
     ),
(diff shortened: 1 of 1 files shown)


Ishticode (changes requested):
Hi @ZJay07
Maybe we should use safety factors instead of specifying min max values

ZJay07:
Oh, what do you mean by a safety factor, do you mind giving me an example?

Ishticode (commented):
safety factor ensures values are bounded based on dtype instead of just hardcoded min or max. You can see it used with `test_tensorflow_reduce_std` in the same file.

ZJay07:
Updated with safety factors similar to `reduce_std`. Let me know if this is correct.

Ishticode (approved):
Thanks @ZJay07 for the fix :)

PR #28543, opened 2024-03-11
fix: for jax backend type hint, `NativeArray` shouldn't be a tuple

mattbarrett98:
fix for `NativeArray` which was a tuple but this breaks things when it is used as a type hint. this fixes the issue for the jax readme demo


vedpatwardhan (approved):

PR #28553, opened 2024-03-12
fix: enable tensorflow numpy api indexing in get_item for tf backend 

Ishticode:
Also removes the _get_item_condition as we perhaps don't need the compositional version for tf. Will restore in case there are scenarios where that is still required.


vedpatwardhan (commented):
let's probably close this PR and continue on [this](https://github.com/unifyai/ivy/pull/28556) one. Thanks @Ishticode 😄

PR #28578, opened 2024-03-13
fix: downgraded tensorflow to 2.15.1 in the GPU docker image as 2.16.1 can't detect a GPU

vedpatwardhan:
This is being tracked in https://github.com/tensorflow/tensorflow/issues/63362


YuvarajP:
Yes, this helps!

PR #28582, opened 2024-03-13
fix: updated _parse_query to use ivy's set_item to use get_item as it's optimized now

vedpatwardhan:


Ishticode:
I've removed to_scalar call in _numel. This should be mostly fine but I am expecting `ivy.pul_along_axis` to fail as it uses _numel expecting it to be scalar for all backends. But will fix any other related changes. @vedpatwardhan

PR #28586, opened 2024-03-13
fix: haiku ivy module build working without needing backend set + extended test

mattbarrett98:
fixes the failing haiku perceiver demo


vedpatwardhan (commented):
Probably would be better to use `with_backend` instead of `set_backend` to avoid updating the dict globally. Would also be interesting to see why the perceiver code snippet worked before (if it worked before) as I don't think we had backend setting in the native modules, what do you think? Thanks @mattbarrett98 😄

Issue #28605, opened 2024-03-14
Fix Frontend Failing Test: paddle - tensor.torch.Tensor.fix_


ZJay07:
To-do List: https://github.com/unifyai/ivy/issues/27500

Issue #28604, opened 2024-03-14
Fix Frontend Failing Test: paddle - tensor.torch.Tensor.fix


ZJay07:
To-do List: https://github.com/unifyai/ivy/issues/27500

PR #28611, opened 2024-03-15
Fix Frontend Failing Test: paddle - tensor.torch.Tensor.fix

ZJay07:
# PR Description
Some dtypes aren't supported according to the logs, updated the supported dtype handle.
tensor fix

## Related Issue

Close https://github.com/unifyai/ivy/issues/28604
Close https://github.com/unifyai/ivy/issues/28605

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/backends/paddle/experimental/elementwise.py
+++ b/ivy/functional/backends/paddle/experimental/elementwise.py
@@ -272,7 +272,9 @@ def allclose(
     return paddle.allclose(x1, x2, rtol=rtol, atol=atol, equal_nan=equal_nan).squeeze(0)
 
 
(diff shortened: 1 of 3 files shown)


Ishticode (changes requested):
Hi @ZJay07 
Thank you very much for the contribution. As this is an issue with paddle backend not being able to handle certain dtypes, it should be specified in the specific ivy function for it is not supported in paddle backend no torch frotnend as native torch itself is able to support `uint8` for `fix` for example. Does that make sense?

ZJay07:
Hey @Ishticode thanks for the feedback! Just updated the fix, not sure if it's the right way to fix it base on paddle specific failures.

Ishticode (approved):
Lgtm. Thank @ZJay07

PR #28631, opened 2024-03-18
fix: fixed `test_torch___getitem__` for paddle backend

### LINKED ISSUE #28630: Fix Frontend Failing Test: paddle - tensor.torch.Tensor.__getitem__

ZenithFlux:
# PR Description

- Fixed the syntax for checking if x is True.
- Added a function to remove negative values in arrays inside `query` of `paddle.get_item` because paddle currently throws an error if any of the value in the arrays inside the query is negative.
- paddle does not support cloning tensors with a 0 length dimension, hence fixed this by creating a new tensor with same shape when asked for a copy.

## Related Issue

Closes #28630

## Checklist

- [x] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/backends/paddle/data_type.py
+++ b/ivy/functional/backends/paddle/data_type.py
@@ -118,6 +118,10 @@ def astype(
     out: Optional[paddle.Tensor] = None,
 ) -> paddle.Tensor:
     dtype = ivy.as_native_dtype(dtype)
+
+    if copy and 0 in x.shape:
+        return paddle.empty(x.shape, dtype=dtype)
+
     if x.dtype == dtype:
         return x.clone() if copy else x
     return x.clone().cast(dtype) if copy else x.cast(dtype)
--- a/ivy/functional/backends/paddle/general.py
+++ b/ivy/functional/backends/paddle/general.py
@@ -87,6 +87,58 @@ def _squeeze_helper(query, x_ndim):
     return squeeze_indices
 
 
+def _make_non_negative(query, x):
+    """Converts negative values inside the tensors in the query to their
+    positive form.
+
+    Returns ``query`` unmodified if it is not a ``list``, ``tuple``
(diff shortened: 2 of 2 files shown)


Ishticode (changes requested):
Hi @ZenithFlux 
The changes here are critical and seem to have lead to many many failures for paddle tests that you can check under logs here at https://github.com/unifyai/ivy/actions/runs/8326992315/job/22798877642?pr=28631
Check Display Test Failures for a list of functions failing.

ZenithFlux:
@Ishticode I don't think these failures are due to my changes.

I confirmed this by reverting all my changes and adding only a comment to both the functions (to trigger the same tests). Even without any changes to the code, there were still about a 100 tests failing. I also ran the tests for the some other failing functions on my local machine and I couldn't find any failure that was newly introduced by this branch. On the contrary, the tests I ran for `unfold` function passed successfully with no errors. I think this is because this change triggers many failing tests whose issues may be unrelated.

If there's is still something I am missing I will be happy to fix that.

ZenithFlux:
@Ishticode I don't think these failures are due to changes on this branch.

I confirmed this by reverting all my changes and adding only a comment to both the functions (to trigger the same tests). Even without any changes to the code, there were still about a 100 tests failing. I also ran the tests for the some other failing functions on my local machine and I couldn't find any failure that was newly introduced by this branch. On the contrary, the tests I ran for `unfold` function passed successfully with no errors. I think this is because this change triggers many failing tests whose issues may be unrelated.

If there's is still something I am missing I will be happy to fix that.

Ishticode (commented):
Right, otherwise looks good. I will merge it after quick few checks later today :)

Ishticode (approved):
tests seems fine and i've checked with some custom examples. Looks good. I will merge now. Thank you very much for the effort put into this contribution @ZenithFlux :)

PR #28665, opened 2024-03-22
fix: for tf.function compatibility in tf backend

mattbarrett98:
Feel free to merge this yourself @vedpatwardhan if everything is all good, as i'm OOO soon for a week. If there's issues will fix when i'm back


vedpatwardhan (commented):
Probably @hmahmood24 would be better suited to review this PR, could you please take a look? Thanks @mattbarrett98 😄

Issue #28664, opened 2024-03-22
Fix Frontend Failing Test: paddle - general_functions.tensorflow.clip_by_value


ZJay07:
To-do List: https://github.com/unifyai/ivy/issues/27500

PR #28668, opened 2024-03-23
Fix Frontend Failing Test: paddle - general_functions.tensorflow.clip_by_value

ZJay07:
# PR Description
generalfunction clip_by_value
clip_by_value paddl

Paddle does not support some dtypes so handled that, not sure why `complex` is not supported, test were kinda flaky where sometimes `jax` passes and `numpy` fails or `tensorflow` passes and `torch` fails, all of them threw the same error (in the pic) saying `complex` is not supported, so added that to the unsupported dtypes.

## Related Issue

Close https://github.com/unifyai/ivy/issues/28664

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/frontends/tensorflow/general_functions.py
+++ b/ivy/functional/frontends/tensorflow/general_functions.py
@@ -94,7 +94,7 @@ def clip_by_norm(t, clip_norm, axes=None):
 
 
 @to_ivy_arrays_and_back
-@with_unsupported_dtypes({"2.15.0 and below": ("float16",)}, "tensorflow")
+@with_unsupported_dtypes({"2.15.0 and below": ("float16", "complex")}, "tensorflow")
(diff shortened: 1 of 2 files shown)


Ishticode (changes requested):
Hi @ZJay07 
Thanks for the PR.
I know you didn't specify float16 as unsupported in tf frontend `clip_by_value` funtion as it was already there but it does support float16
image

So maybe remove that. Otherwise all is good

Ishticode:
but that says "complex64" is not in supported dtypes. And it then lists the supported dtypes. You can see "half" listed in there which does refer to float16.

Ishticode (approved):
lgtm. Thanks for the fix.
I have seen this before. Although `complex` should automatically include both but there might be a bug that we should look into in the un/supported dtypes decorator. :)

Issue #28676, opened 2024-03-24
Fix Frontend Failing Test: tensorflow - mathematical_functions.jax.numpy.sinc


ZJay07:
To-do List: https://github.com/unifyai/ivy/issues/27499

PR #28677, opened 2024-03-24
Fix Frontend Failing Test: tensorflow - mathematical_functions.jax.numpy.sinc

ZJay07:
# PR Description
image
`bfloat16` seems to be not supported

## Related Issue

Close https://github.com/unifyai/ivy/issues/28676

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/backends/tensorflow/experimental/elementwise.py
+++ b/ivy/functional/backends/tensorflow/experimental/elementwise.py
@@ -66,6 +66,10 @@ def lgamma(
     return tf.math.lgamma(x)
 
 
+@with_unsupported_dtypes(
+    {"2.15.0 and below": ("bfloat16",)},
+    backend_version,
+)
 def sinc(
     x: Union[tf.Tensor, tf.Variable],
     /,


Ishticode (changes requested):
Thank you very much @ZJay07 for the PR
You are right in recognizing that ivy.sinc with tf backend throws an error but this looks like an ivy issue. Becaue the code in the actual implementation works with native tensorflow as in
```python
import tensorflow as tf
x = tf.constant([1, 2, 3.], dtype=tf.bfloat16)
x = x*3.14
print(tf.cast(tf.where(x == 0, 1, tf.math.sin(x) / x), x.dtype))
```

which means it does support bfloat16 natively.
Are you able to figure out where our code creates an issue

ZJay07:
Hi @Ishticode, it should work now, I'm using the `tf` version of `sinc` and cast it back to the same `dtype` of `x` which follows the characteristic of `jax.numpy.sinc` which mentions `out – sinc(x), which has the same shape as the input.`

ref: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.sinc.html

ZJay07:
Hey @Ishticode, not sure if I am reading the right action logs, seems like there's no more failures with my changes

Ishticode (commented):
sure

Ishticode:
any updates @ZJay07 :)

PR #28688, opened 2024-03-26
Fix Frontend Failing Test: paddle - comparison.numpy.less_equal

ZJay07:
# PR Description
https://www.paddlepaddle.org.cn/documentation/docs/en/2.2/api/paddle/less_equal_en.html#less-equal
Specify dtypes based on `paddlepaddle docs

## Related Issue

Close https://github.com/unifyai/ivy/issues/28687

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/backends/paddle/elementwise.py
+++ b/ivy/functional/backends/paddle/elementwise.py
@@ -136,8 +136,8 @@ def equal(
     return paddle.equal(x1, x2)
 
 
-@with_unsupported_dtypes(
-    {"2.6.0 and below": ("bool",)},
+@with_supported_dtypes(
+    {"2.6.0 and below": ("bool", "float32", "float64", "int32", "int64")},
     backend_version,
 )
 def less_equal(


Ishticode (changes requested):
Hi @ZJay07 
Thanks for the PR
The implementation here should be able to handle more than just ones specified as supported. For example int8 works but is not included in the supoorted types
you can check by 
```python
ivy.set_paddle_backend()
x = ivy.array(2, dtype="int8")
print(ivy.less_equal(x,x))
```
In addition, the function also seems to be explicitly handling complex types.

ZJay07:
> In addition, the function also seems to be explicitly handling complex types.

Hi @Ishticode, thanks for the review, should the function be allowed to handle `complex` type or should I remove the handling of those since less_equals does not support complex

Ishticode:
> > In addition, the function also seems to be explicitly handling complex types.
> 
> Hi @Ishticode, thanks for the review, should the function be allowed to handle `complex` type, is it mainly to accommodate the `jax` implementation?

Yes. Does it get the same results as jax?

ZJay07:
Hey @Ishticode managed to fix it seems like it was missing some equal signs since it is comparing `less_equal`

Ishticode (changes requested):
That looks right but the int8 dtype is still not speciifed as supported. As you can see it does work with native paddle in the example given above. Could you include that? Thanks :)

Ishticode:
any updates @ZJay07 on this one? :)

ZJay07:
Hey @Ishticode, currently busy with exams but will be back to fixing it later this month!

Issue #28700, opened 2024-03-28
 Fix Frontend Failing Test: paddle - comparison.numpy.greater_equal


ZJay07:
To-do List: https://github.com/unifyai/ivy/issues/27500

PR #28701, opened 2024-03-28
Fix Frontend Failing Test: paddle - comparison.numpy.greater_equal

ZJay07:
# PR Description
Error thrown due to complex comparison, `paddle` does not support complex for `greater equal` not sure why the function was trying to do so, removing the comparison works, but not sure if it was essential to make a "`complex` acceptable" `paddle.greater_equal`

ref: https://www.paddlepaddle.org.cn/documentation/docs/en/2.2/api/paddle/greater_equal_en.html#greater-equal

## Related Issue

Close https://github.com/unifyai/ivy/issues/28700

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/backends/paddle/elementwise.py
+++ b/ivy/functional/backends/paddle/elementwise.py
@@ -470,6 +470,7 @@ def greater(
             "int16",
             "int32",
             "int64",
+            "complex",
         )
     },
     backend_version,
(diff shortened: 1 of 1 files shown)


Ishticode (commented):
I think this was based on our goal to superset the functions. So our backends are as comprehensive as possible. I think numpy support complex comparison so perhaps its trying to copy that. 
image

Could you possibly check that out if it matches with np. Thank you very much :)

ZJay07:
Similar fix to https://github.com/unifyai/ivy/pull/28688

Ishticode:
> Similar fix to #28688

Thanks. Similar to #28688 I would say supported dtypes can be extended here too like int8. But I can leave that for another PR. As you changes here don't touch the suppported dtypes decorators etc. As opposed to #28688.

Ishticode (approved):
Thanks @ZJay07 for the effort put into this fix :)

PR #28721, opened 2024-04-01
Updated inv_ex function in torch frontend

rohitsalla:
the current inv_ex function had a TODO task as  TODO: info should return an array containing the diagonal element of the . i have rewritten the task


Ishticode:
closing this as the changes at the moment are the same as #28724 with the same undefined variable errors. Please feel free to reopen if there is a distinct goal here. Thanks :)

Issue #28727, opened 2024-04-03
Fix Frontend Failing Test: paddle - operators.jax.lax.ne


ZJay07:
To-do List: https://github.com/unifyai/ivy/issues/27500

PR #28728, opened 2024-04-03
Fix Frontend Failing Test: paddle - operators.jax.lax.ne

ZJay07:
# PR Description
Error was due to very small abs differences, where jax rounds down to `0` but `paddlepaddle` takes the value. Fix by limiting the smallest abs value that can be generated for testing

eg: `-2.23517418e-08` and `-1.49011612e-08`, is labelled as `False` in `paddle` but `True` in `jax` 

## Related Issue

Close https://github.com/unifyai/ivy/issues/28727

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?


Ishticode (approved):
Thanks @ZJay07 for the fix :)

PR #28738, opened 2024-04-15
fix: failing test test_jax_min

### LINKED ISSUE #28737: Fix Frontend Failing Test: numpy - operators.jax.lax.min
Working on it

tanmaymunjal:
# PR Description

The test test_jax_min failing seems to be an issue with the paddle real function(https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/real_en.html). When it converts a float64 object, it turns it into float32, losing precision and often rounding numbers close to each other to the same number. Setting the supported dtype of minimum function to only complex64 solves this

## Related Issue

Closes #28737 

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### Socials

https://twitter.com/free_thinker_9

### DIFF
--- a/ivy/functional/backends/paddle/elementwise.py
+++ b/ivy/functional/backends/paddle/elementwise.py
@@ -1067,7 +1067,7 @@ def erf(x: paddle.Tensor, /, *, out: Optional[paddle.Tensor] = None) -> paddle.T
 
 
 @with_supported_dtypes(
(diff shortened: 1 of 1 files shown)


Ishticode (commented):
Hi. Thanks for the PR @tanmaymunjal 
looks good. I guess it doesn't support complex128 either. usually "complex" should include both but there is some underlying bug with the decorator yet to be fixed. Maybe add complex128 and will get this merged. THanks for looking into this. :).

PR #28740, opened 2024-04-17
fix: torch.masked_fill

### LINKED ISSUE #28437: Fix Frontend Failing Test: paddle - tensor.torch.Tensor.masked_fill
To-do list: https://github.com/unifyai/ivy/issues/27500

Kacper-W-Kozdon:
# PR Description
Changed the way type-casting was done in the frontend. Previous implementation did not enforce keeping the input's dtype and paddle backend utilises type promotion in the function called by torch's frontend `.masked_fill()`. This resulted in the failure of dtype assertions and a failed paddle pytest. All of the backends use `.where()` to get the same results as `torch.masked_fill()`, with paddle's solution being probably a bit overcomplicated- but the tests for `.where()` are passing, so I did not want to alter that.

Potentially to-do:
- if it's not too redundant: adding `.masked_fill()` for the superset behaviour (even torch backend uses `.where()` when `.masked_fill()` is called);
- reformatting paddle's backend implementation of `.where()`.

## Related Issue

Closes #28437

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/frontends/torch/tensor.py
+++ b/ivy/functional/frontends/torch/tensor.py
@@ -1062,8 +1062,8 @@ class Tensor:
         return torch_frontend.acosh(self)
 
     def masked_fill(self, mask, value):
-        dtype = self.dtype
-        return torch_frontend.where(mask, value, self).to(dtype)
+        dtype = ivy.as_native_dtype(self.dtype)
+        return ivy.astype(torch_frontend.where(mask, value, self), dtype)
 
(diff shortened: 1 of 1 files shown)


Ishticode (commented):
lgtm. I will run a further check on this later today. Thanks for the effort and patience @Kacper-W-Kozdon :)

Ishticode (approved):
Thanks @Kacper-W-Kozdon :)

PR #28767, opened 2024-06-19
fix torch frontend blas_and_lapack_ops cholesky

### LINKED ISSUE #28766: Fix Frontend Failing Test: torch - blas_and_lapack_ops.cholesky and torch - linalg.cholesky

Daniel4078:
# PR Description

## Related Issue

Closes #28766

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/frontends/torch/blas_and_lapack_ops.py
+++ b/ivy/functional/frontends/torch/blas_and_lapack_ops.py
@@ -91,7 +91,7 @@ def chain_matmul(*matrices, out=None):
 
 @to_ivy_arrays_and_back
 def cholesky(input, upper=False, *, out=None):
-    return ivy.cholesky(input, upper=upper, out=out)
+    return ivy.cholesky(input, upper=upper, out=out).astype(ivy.float64)
 
 
 @to_ivy_arrays_and_back
--- a/ivy/functional/frontends/torch/blas_and_lapack_ops.py
+++ b/ivy/functional/frontends/torch/blas_and_lapack_ops.py
@@ -91,7 +91,10 @@ def chain_matmul(*matrices, out=None):
 
 @to_ivy_arrays_and_back
 def cholesky(input, upper=False, *, out=None):
-    return ivy.cholesky(input, upper=upper, out=out).astype(ivy.float64)
+    temp = ivy.cholesky(input, upper=upper, out=out)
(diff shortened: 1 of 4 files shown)


Daniel4078:
now for the 5 frontend tests of test_torch_cholesky, only one of paddle fails with RuntimeError('(NotFound) The kernel with key (CPU, Undefined(AnyLayout), bfloat16) of kernel `cholesky` is not registe...cond.end() && kernel_key.backend() == Backend::CPU:1 == true:1.] (at /paddle/paddle/phi/core/kernel_factory.cc:262)\n')

Daniel4078:
Now it seems like the target test_torch_cholesky is no longer failing

ivy/functional/frontends/torch/blas_and_lapack_ops.py
```diff
@@ -91,7 +91,10 @@ def chain_matmul(*matrices, out=None):
 
 @to_ivy_arrays_and_back
 def cholesky(input, upper=False, *, out=None):
-    return ivy.cholesky(input, upper=upper, out=out)
+    temp = ivy.cholesky(input, upper=upper, out=out)
```
Sam-Armstrong, line 94: For future reference, a better naming scheme here would be something like `out` or `ret`.
```suggestion
    ret = ivy.cholesky(input, upper=upper, out=out)
```

Sam-Armstrong (changes requested):
Thanks for the contribution @Daniel4078!

I've just had a look at the failing test for `torch.linalg.cholesky`, and it seems the problem is actually with the test itself. I was able to fix it for all backends by changing this line:

`x = np.matmul(x.T, x) + np.identity(x.shape[0])`

to this:

`x = np.matmul(np.conjugate(x.T), x) + np.identity(x.shape[0], dtype=dtype[0])`

Would you be able to make this change, and see if it also works for the version in blas_and_lapack_ops? (you can remove all these current changes you made) Thanks!

Daniel4078:
@Sam-Armstrong Thank you! I have checked, your changes also work in this blas_and_lapack_ops version. And I think this PR should be ready to be merged.

Sam-Armstrong (approved):
Looks good to me. Thanks a lot @Daniel4078!

PR #28770, opened 2024-06-20
Fix torch svd

### LINKED ISSUE #28769: Fix Frontend Failing Test: torch - blas_and_lapack_ops.svd

Daniel4078:
# PR Description

## Related Issue

Closes #28769

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/backends/jax/linear_algebra.py
+++ b/ivy/functional/backends/jax/linear_algebra.py
@@ -361,7 +361,7 @@ def svd(
     x: JaxArray, /, *, compute_uv: bool = True, full_matrices: bool = True
 ) -> Union[JaxArray, Tuple[JaxArray, ...]]:
     if compute_uv:
-        results = namedtuple("svd", "U S Vh")
+        results = namedtuple("svd", ['U', 'S', 'Vh'])
         U, D, VT = jnp.linalg.svd(x, full_matrices=full_matrices, compute_uv=compute_uv)
         return results(U, D, VT)
     else:
--- a/ivy/functional/backends/numpy/linear_algebra.py
+++ b/ivy/functional/backends/numpy/linear_algebra.py
@@ -291,7 +291,7 @@ def svd(
     x: np.ndarray, /, *, compute_uv: bool = True, full_matrices: bool = True
 ) -> Union[np.ndarray, Tuple[np.ndarray, ...]]:
     if compute_uv:
-        results = namedtuple("svd", "U S Vh")
+        results = namedtuple("svd", ['U', 'S', 'Vh'])
         U, D, VT = np.linalg.svd(x, full_matrices=full_matrices, compute_uv=compute_uv)
         return results(U, D, VT)
     else:
--- a/ivy/functional/backends/paddle/linear_algebra.py
+++ b/ivy/functional/backends/paddle/linear_algebra.py
@@ -521,7 +521,7 @@ def svd(
 ) -> Union[paddle.Tensor, Tuple[paddle.Tensor, ...]]:
     ret = paddle.linalg.svd(x, full_matrices=full_matrices)
     if compute_uv:
-        results = namedtuple("svd", "U S Vh")
+        results = namedtuple("svd", ['U', 'S', 'Vh'])
         return results(*ret)
     else:
         results = namedtuple("svd", "S")
--- a/ivy/functional/backends/tensorflow/linear_algebra.py
+++ b/ivy/functional/backends/tensorflow/linear_algebra.py
@@ -545,8 +545,7 @@ def svd(
     compute_uv: bool = True,
 ) -> Union[Union[tf.Tensor, tf.Variable], Tuple[Union[tf.Tensor, tf.Variable], ...]]:
     if compute_uv:
-        results = namedtuple("svd", "U S Vh")
-
+        results = namedtuple("svd", ['U', 'S', 'Vh'])
         batch_shape = tf.shape(x)[:-2]
         num_batch_dims = len(batch_shape)
         transpose_dims = list(range(num_batch_dims)) + [
--- a/ivy/functional/backends/torch/linear_algebra.py
+++ b/ivy/functional/backends/torch/linear_algebra.py
@@ -414,16 +414,13 @@ def svd(
     x: torch.Tensor, /, *, full_matrices: bool = True, compute_uv: bool = True
 ) -> Union[torch.Tensor, Tuple[torch.Tensor, ...]]:
     if compute_uv:
-        results = namedtuple("svd", "U S Vh")
-
+        results = namedtuple("svd", ['U', 'S', 'Vh'])
         U, D, VT = torch.linalg.svd(x, full_matrices=full_matrices)
         return results(U, D, VT)
     else:
         results = namedtuple("svd", "S")
-        svd = torch.linalg.svd(x, full_matrices=full_matrices)
-        # torch.linalg.svd returns a tuple with U, S, and Vh
-        D = svd[1]
-        return results(D)
+        s = torch.linalg.svdvals(x)
+        return results(s)
 
 
 @with_unsupported_dtypes({"2.2 and below": ("float16", "bfloat16")}, backend_version)
--- a/ivy/functional/ivy/linear_algebra.py
+++ b/ivy/functional/ivy/linear_algebra.py
@@ -2130,15 +2130,12 @@ def svd(
         If ``True`` then left and right singular vectors will be computed and returned
         in ``U`` and ``Vh``, respectively. Otherwise, only the singular values will be
         computed, which can be significantly faster.
-    .. note::
-        with backend set as torch, svd with still compute left and right singular
-        vectors irrespective of the value of compute_uv, however Ivy will still
-        only return the singular values.
 
     Returns
     -------
     .. note::
         once complex numbers are supported, each square matrix must be Hermitian.
+        In addition, the return will be a namedtuple ``(S)`` when compute_uv is ``False``
 
     ret
         a namedtuple ``(U, S, Vh)`` whose


Daniel4078:
@Sam-Armstrong Can you tell me about how to get the device of the input received by the torch frontend function svd? Because the documentation requires the zero-filled matrixes in the output should be on the same device as input (https://pytorch.org/docs/stable/generated/torch.svd.html)

Sam-Armstrong:
@Daniel4078 you should just be able to do `x.device` right? I'm not sure about the ground truth, but it should be produced by the native torch function, so you should be able to reproduce the results manually in a python script

Daniel4078:
@Sam-Armstrong I found something strange when I run the test locally. The test's error is like this 
```
values = [[array([[0., 0.],
       [0., 0.]], dtype=float32), array([[4.89516249e-308, 0.00000000e+000],
       [4.89516249e-30...    [4.89516249e-308, 0.00000000e+000]]), array([[[0., 0.],
        [0., 0.]],

       [[0., 0.],
        [0., 0.]]])]]
this_key_chain = None

    def assert_same_type_and_shape(values, this_key_chain=None):
        x_, y_ = values
        for x, y in zip(x_, y_):
            if isinstance(x, np.ndarray):
                x_d = str(x.dtype).replace("longlong", "int64")
                y_d = str(y.dtype).replace("longlong", "int64")
>               assert (
                    x.shape == y.shape
                ), f"returned shape = {x.shape}, ground-truth returned shape = {y.shape}"
E               AssertionError: returned shape = (2, 2), ground-truth returned shape = (2, 2, 2)
E               Falsifying example: test_torch_svd(
E                   on_device='cpu',
E                   frontend='torch',
E                   backend_fw='jax',
E                   dtype_and_x=(['float64'], [array([[[-2.44758124e-308, -2.44758124e-308],
E                             [-2.44758124e-308, -2.44758124e-308]],
E                     
E                            [[-2.44758124e-308, -2.44758124e-308],
E                             [-2.44758124e-308, -2.44758124e-308]]])]),
E                   some=False,
E                   compute=False,
E                   test_flags=FrontendFunctionTestFlags(
E                       num_positional_args=0,
E                       with_out=False,
E                       with_copy=False,
E                       inplace=False,
E                       as_variable=[False],
E                       native_arrays=[False],
E                       test_trace=False,
E                       test_trace_each=False,
E                       generate_frontend_arrays=False,
E                       transpile=False,
E                       precision_mode=True,
E                   ),
E                   fn_tree='ivy.functional.frontends.torch.svd',```
However, the return from both the ground truth function and ourfrontend are named tuples (U,S,V) where U and V should be zero matrixs. It seems like the assertion check somehow pick the first two part of one returned tuple and consider them as the frontend-groundtruth pair. I think maybe we should investigate this further.

**Comment by Sam-Armstrong** [comment]
@Daniel4078 isn't the problem here that the shape of the tensors in the named tuple is different to the ground truth? Like the first element U of the tuple is 3d in the ground truth but 2d in the function return

**Comment by Daniel4078** [comment]
> @Daniel4078 isn't the problem here that the shape of the tensors in the named tuple is different to the ground truth? Like the first element U of the tuple is 3d in the ground truth but 2d in the function return

But according to (https://pytorch.org/docs/stable/generated/torch.svd.html), If compute_uv is False, the returned U and V will be zero-filled matrices of shape (m, m) and (n, n) respectively, and the same device as input. The argument some has no effect when compute_uv is False. Then the shape of U should be 2D, not 3D like S right?

**Comment by Daniel4078** [comment]
@Sam-Armstrong the tests now fails due to the fact that the correct output of svd is not unique and I am thinking of copying the test for ivy.linalg.svd(https://github.com/ivy-llc/ivy/blob/72fa1698afd8e262a5c35c0567bbaa8d3873006b/ivy_tests/test_ivy/test_functional/test_core/test_linalg.py#L944) instead as it test values based on recreating the original matrix and testing the consistency.

**Comment by Sam-Armstrong** [comment]
@Daniel4078 sure, think that makes sense. yeah that would be great if you can fix the other failing frontend svd tests in this PR, thanks!

**Comment by Daniel4078** [comment]
> @Daniel4078 sure, think that makes sense. yeah that would be great if you can fix the other failing frontend svd tests in this PR, thanks!

It is quite surprising that now torch.blas_and_lapack_ops.svd, torch.tensor.svd, jax.lax.linalg.svd, jax.numpy.linalg.svd, numpy.linalg.decompositions.svd are failing; while torch.linalg.svd , tensorflow.linalg.svd and tensorflow.raw_ops.svd are skipped.

**Comment by Sam-Armstrong** [comment]
@Daniel4078 you haven't made any changes to the tests here, did you force-push over them?

**Comment by Daniel4078** [comment]
> @Daniel4078 you haven't made any changes to the tests here, did you force-push over them?

I accidententally commited some other changes intended in another branch in here, so I thought it would be better to remove that commit.

**Comment by Daniel4078** [comment]
@Sam-Armstrong Why the torch.svd frontend tests are skipping themselves when I run them locally, showing that "Skipped: Function not implemented in backend." What should I do in this case?

File: ivy_tests/test_ivy/test_frontends/test_torch/test_linalg.py [hunk 1]
```diff
@@ -1272,25 +1280,30 @@ def test_torch_svd(
         fn_tree=fn_tree,
         on_device=on_device,
         test_values=False,
-        atol=1e-03,
-        rtol=1e-05,
         A=x,
         full_matrices=full_matrices,
     )
     ret = [ivy.to_numpy(x) for x in ret]
```
Sam-Armstrong, line 1286: I think the problem is caused by this line, maybe change it to `ret = [np.asarray(x) for x in ret]` and see if that works?

Daniel4078:
trying to solve the TypeError: iteration over a 0-d array error when comparing two (somehow empty) result from the matric calculation

Daniel4078:
@Sam-Armstrong I am stuck trying to fix this final error of jax.numpy.linalg.svd with torch backend:
```
test_linalg.py:855: 
test_linalg.py:915: in test_jax_svd
    assert_all_close(
../../../helpers/assertions.py:56: in assert_all_close
    np.nan_to_num(ret_np), np.nan_to_num(ret_from_gt_np), rtol=rtol, atol=atol
/opt/fw/numpy/numpy/lib/type_check.py:497: in nan_to_num
    x = _nx.array(x, subok=True, copy=copy)
/opt/project/ivy/data_classes/array/array.py:366: in __array__
    return self._data.__array__(*args, dtype=self.dtype, **kwargs)

self = tensor([8.6646e+01, 1.0000e-03], dtype=torch.float64,
       grad_fn=<ToCopyBackward0>)
dtype = 'float64'

    def __array__(self, dtype=None):
        if has_torch_function_unary(self):
            return handle_torch_function(Tensor.__array__, (self,), self, dtype=dtype)
        if dtype is None:
            return self.numpy()
        else:
>           return self.numpy().astype(dtype, copy=False)
E           RuntimeError: Can't call numpy() on Tensor that requires grad. Use tensor.detach().numpy() instead.
```
But I can't find in ivy repo the line that cause the problem. Should I just leave it like this and fix other problems?

ivy_tests/test_ivy/test_frontends/test_torch/test_blas_and_lapack_ops.py
```diff
@@ -848,37 +849,62 @@ def test_torch_qr(
...
+            ground_truth_backend=frontend,
+        )
+    else:
+        helpers.assert_all_close(
+            ret_np=np.asarray(frontend_u[...,:frontend_s.shape[0]] @ np.diag(frontend_s) @ frontend_v.T, dtype=np.float32),
+            ret_from_gt_np=np.asarray(u[...,:s.shape[0]] @ np.diag(s) @ v.T, dtype=np.float32),
+            atol=1e-04,
+            backend=backend_fw,
+            ground_truth_backend=frontend,
+        )
```
Sam-Armstrong, line 907: Why have you changed from using test_frontend_function (with test_values=True) to helpers.assert_all_close?

Sam-Armstrong (changes requested):
The reason the test is failing is you're trying to convert a torch tensor to a numpy array, without detaching the gradients (using `tensor.detach().numpy()`). I don't really understand the reasoning behind changing the format of the tests, such as the comment I left below - can't we just change the `atol_`/`rtol_` values with test_frontend_function, rather than adding these extra assertions? Or are there other reasons you did this?

ivy_tests/test_ivy/test_frontends/test_torch/test_blas_and_lapack_ops.py
Daniel4078: because singular value decomposition A=UΣVT may not be unique and so should check computation result to see if the return is valid

Daniel4078:
> The reason the test is failing is you're trying to convert a torch tensor to a numpy array, without detaching the gradients (using `tensor.detach().numpy()`). I don't really understand the reasoning behind changing the format of the tests, such as the comment I left below - can't we just change the `atol_`/`rtol_` values with test_frontend_function, rather than adding these extra assertions? Or are there other reasons you did this?

But the part that convert a torch tensor to a numpy array is not implemented by me and the same process works for tests of tenserflow.linalg.svd so I did not think I should not do the convertion. Or is there some way to do the same computation without converting? Thank you!

ivy_tests/test_ivy/test_frontends/test_torch/test_blas_and_lapack_ops.py
```diff
@@ -848,37 +849,62 @@ def test_torch_qr(
...
         fn_tree=fn_tree,
         on_device=on_device,
-        input=x[0],
+        test_values=False,
+        input = x,
         some=some,
-        compute_uv=compute,
+        compute_uv=compute_uv,
     )
+    ret = [np.asarray(x) for x in ret]
```
Sam-Armstrong, line 886:
```suggestion
    if backend_fw == "torch":
        ret = [x.detach() for x in ret]
    ret = [np.asarray(x) for x in ret]
```
You can fix the error you're seeing with something like this. You'll need to implement this for the other tests as well.

ivy/functional/backends/jax/linear_algebra.py
```diff
@@ -361,7 +361,7 @@ def svd(
     x: JaxArray, /, *, compute_uv: bool = True, full_matrices: bool = True
 ) -> Union[JaxArray, Tuple[JaxArray, ...]]:
     if compute_uv:
-        results = namedtuple("svd", "U S Vh")
+        results = namedtuple("svd", ['U', 'S', 'Vh'])
```
Sam-Armstrong, line 364: I think changing this from a string to a list is unnecessary, right? Can you change this back for all the functions you've changed it in, unless there's a reason?

Sam-Armstrong (changes requested):
Looks like these svd tests are still failing, could you try to resolve these as well? But otherwise that makes sense 👍 thanks! 😊

- `ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py::test_torch_svd`
- `ivy_tests/test_ivy/test_frontends/test_numpy/test_linalg/test_decompositions.py::test_numpy_svd`
- `ivy_tests/test_ivy/test_frontends/test_jax/test_numpy/test_linalg.py::test_jax_svd`
- `ivy_tests/test_ivy/test_frontends/test_jax/test_lax/test_linalg.py::test_jax_svd`
- `ivy_tests/test_ivy/test_frontends/test_tensorflow/test_raw_ops.py::test_tensorflow_Svd`

ivy_tests/test_ivy/test_frontends/test_torch/test_blas_and_lapack_ops.py
Daniel4078: Thanks a lot!

ivy/functional/backends/jax/linear_algebra.py
Daniel4078: sure, will do!

Daniel4078:
it is strange that with the same calculation to reconstruct the decomposited matrix (used in all other relevant tests), current jax.numpy.linalg.svd passes all tests while current jax.lax.linalg.svd fails with significantly different values produced. May be the decomposition of jax.lax.linalg.svd is different from the common way?

Daniel4078:
found that the svd function of tenserflow backend does not support complex input, this may be the reason why some of the frontend tests for svd returns wrong results in tensorflow backend for complex value inputs. Also, I want to ask is the first component of the return of the helpers.test_frontend_function representing the groundtruth return?

Daniel4078:
found that the svd function of tenserflow backend does not support complex input, this may be the reason why some of the frontend tests for svd returns wrong results in tensorflow backend for complex value inputs. Also, I want to ask is the first component of the return of the helpers.test_frontend_function representing the groundtruth return?

PR #28773, opened 2024-06-21
Fix torch linalg cholesky_ex

### LINKED ISSUE #28772: Fix Frontend Failing Test: torch - linalg.cholesky_ex

Daniel4078:
# PR Description

## Related Issue

Closes #28772

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/frontends/torch/linalg.py
+++ b/ivy/functional/frontends/torch/linalg.py
@@ -16,18 +16,23 @@ def cholesky(input, *, upper=False, out=None):
 
 
 @to_ivy_arrays_and_back
+@with_supported_dtypes(
+    {"2.2 and below": ("float32", "float64", "complex32", "complex64")}, "torch"
+)
 def cholesky_ex(input, *, upper=False, check_errors=False, out=None):
     try:
+        results = namedtuple("cholesky_ex", ['L', 'info'])
         matrix = ivy.cholesky(input, upper=upper, out=out)
         info = ivy.zeros(input.shape[:-2], dtype=ivy.int32)
-        return matrix, info
+        return results(matrix, info)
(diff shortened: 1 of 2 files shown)


Daniel4078:
@Sam-Armstrong This experimental function is supposed to quckly and directly returning the LAPACK error codes as its return when the normal cholesky would fail, by directly checking for the error before calcuating the cholesky value. Though I don't know how to implement and test for that behavior. Though the tests are passing for now it seems.

Sam-Armstrong (changes requested):
I think we can just return a regular tuple rather than a named tuple, I know the torch function returns a named tuple, but practically I don't think there would really be a difference and I'm concerned that the named tuple could possibly cause problems for the transpiler. Otherwise the changes look good. Thanks @Daniel4078!

Sam-Armstrong:
I don't think you need to implement/test the LAPACK error codes right now, that's probably out of scope for the moment.

Daniel4078:
> I think we can just return a regular tuple rather than a named tuple, I know the torch function returns a named tuple, but practically I don't think there would really be a difference and I'm concerned that the named tuple could possibly cause problems for the transpiler. Otherwise the changes look good. Thanks @Daniel4078!

The reason why I use named tuple is that some other backend functions like ivy.svd did already return named tuple (https://github.com/ivy-llc/ivy/blob/5665437a25b2ffec563229fd9eecd67fb870ac3d/ivy/functional/ivy/linear_algebra.py#L2143) and they seems to work fine.

Sam-Armstrong (approved):
Ah yeah, I think this is fine on second thought. Thanks a lot @Daniel4078!

PR #28780, opened 2024-06-29
add torch.tensor.masked_select

Daniel4078:
# PR Description

## Related Issue

None, add a missing frontend function of torch

## Checklist

- [x] Did you add a function?
- [x] Did you add the tests?

### DIFF
--- a/ivy/functional/frontends/torch/tensor.py
+++ b/ivy/functional/frontends/torch/tensor.py
@@ -1896,7 +1896,9 @@ class Tensor:
         ).ivy_array
         return self
 
-    @with_unsupported_dtypes({"2.2 and below": ("bfloat16", "float16")}, "torch")
+    @with_supported_dtypes(
+        {"2.2 and below": ("float32", "float64", "complex32", "complex64")}, "torch"
+    )
     def cholesky(self, upper=False):
         return torch_frontend.cholesky(self, upper=upper)
 
--- a/ivy/functional/frontends/torch/tensor.py
+++ b/ivy/functional/frontends/torch/tensor.py
@@ -1071,6 +1071,11 @@ class Tensor:
         self.ivy_array = self.masked_fill(mask, value).ivy_array
         return self
 
+    def masked_select(self, mask):
+        x, mask = torch_frontend.broadcast_tensors(self, mask)
+        out = x.reshape(-1).index_select(0, mask.reshape(-1).nonzero().flatten())
+        return torch_frontend.tensor(out)
+
     @with_unsupported_dtypes({"2.2 and below": ("float16", "bfloat16")}, "torch")
     def index_add_(self, dim, index, source, *, alpha=1):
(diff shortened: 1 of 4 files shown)


Daniel4078:
@Sam-Armstrong Please check if the new function is added to the correct place. And I want to ask do I need to add a ivy version and a torch backend version of this function as well?

ivy/functional/frontends/torch/tensor.py
```diff
@@ -1071,6 +1071,11 @@ def masked_fill_(self, mask, value):
         self.ivy_array = self.masked_fill(mask, value).ivy_array
         return self
 
+    def masked_select(self, mask):
+        x, mask = torch_frontend.broadcast_tensors(self, mask)
+        out = x.reshape(-1).index_select(0, mask.reshape(-1).nonzero().flatten())
+        return torch_frontend.tensor(out)
```
Sam-Armstrong, line 1077:
```suggestion
        return torch_frontend.masked_select(self, mask)
```
Couldn't we just call into the existing torch frontend `masked_select` function with something like this?

Sam-Armstrong (changes requested):

ivy/functional/frontends/torch/tensor.py
```diff
@@ -1891,7 +1894,9 @@ def addcdiv_(self, tensor1, tensor2, *, value=1):
         ).ivy_array
         return self
 
-    @with_unsupported_dtypes({"2.2 and below": ("bfloat16", "float16")}, "torch")
+    @with_supported_dtypes(
+        {"2.2 and below": ("float32", "float64", "complex32", "complex64")}, "torch"
+    )
```
Sam-Armstrong, line 1899: Changing this seems to cause the torch.Tensor.cholesky test to fail with a tensorflow backend? Can we just remove this from the PR? Otherwise looks good 😁

Sam-Armstrong (changes requested):

ivy/functional/frontends/torch/tensor.py
Daniel4078: But the origenal version would let the test generate boolean values for cholesky, which is not right. So can I have more details on the failure caused by this? Thanks

Sam-Armstrong:
```
torch._C._LinAlgError: cholesky: The factorization could not be completed because the input is not positive-definite (the leading minor of order 1 is not positive-definite)
```
Looks like you need to ensure that the input matrix is positive-definite.

You should be able to reproduce the error yourself with:
```
pytest ivy_tests/test_ivy/test_frontends/test_torch/test_tensor.py::test_torch_cholesky -p no:warnings --backend tensorflow
```

Also I think `"complex128"` is a supported dtype.

Sam-Armstrong (approved):
lgtm! thanks @Daniel4078 😁

PR #28803, opened 2024-08-11
fix: tf backend set_item

Sam-Armstrong:


Sam-Armstrong:
This still isn't compatible with `tf.function`, as new `tf.Variable`s can't be created within the compiled function - closing

PR #28823, opened 2024-09-15
fix: tf backend get_item and add set_item

Sam-Armstrong:
Fixes ivy.get_item with a tensorflow backend in the case that the query passed is a multi-dimensional tensor.


ivy/utils/decorator_utils.py
```diff
@@ -183,8 +183,6 @@ def handle_get_item(fn):
     def wrapper(inp, query, **kwargs):
         try:
             res = inp.__getitem__(query)
-        except IndexError:
-            raise
```
Sam-Armstrong, line 187: need to remove this so the backend implementation can be called, which handles an IndexError when the query is a multi-dimensional tensor

Sam-Armstrong:
@hmahmood24 does this look ok?

hmahmood24 (approved):
@Sam-Armstrong Looks good! `get_item` is a critical function used everywhere so might be good to keep this on a separate branch and rerunning the integration tests from this branch to see this change doesn't break any other case and then merge into `main`. What do you think?

PR #28824, opened 2024-09-17
simplify requirements.txt

Sam-Armstrong:
Remove unnecessary dependencies from the requirements.txt file.

### DIFF
--- a/docs/overview/design/building_blocks.rst
+++ b/docs/overview/design/building_blocks.rst
@@ -354,15 +354,11 @@ The contents of this function are as follows:
         # if a global backend has been set with set_backend then this will be returned
         if backend_stack:
             f = backend_stack[-1]
-            if verbosity.level > 0:
-                verbosity.cprint(f"Using backend from stack: {f}")
             return f
 
         # if no global backend exists, we try to infer the backend from the arguments
         f = _determine_backend_from_args(list(args) + list(kwargs.values()))
         if f is not None:
-            if verbosity.level > 0:
-                verbosity.cprint(f"Using backend from type: {f}")
             implicit_backend = f.current_backend_str()
             return f
         return importlib.import_module(_backend_dict[implicit_backend])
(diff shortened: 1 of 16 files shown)


Sam-Armstrong:
@hmahmood24 @YushaArif99 am I ok to merge this? seems to be fine with local testing, but I think the only way to test it out properly will be as part of the 'full ivy test from fresh dev perspective' which we can do soon.

YushaArif99 (approved):
lgtm at a first glance! But i guess we'll need to do some e2e testing to verify everything works smoothly from an end user's perspective.

hmahmood24 (changes requested):
@Sam-Armstrong Thanks for working on this. The first question/observation that comes to mind is that `requirements.txt` is supposed to represent the minimal requirements in order to use `ivy` out of the box without minimal effort and since we are launching `ivy` with the purpose of using it as a transpiler with support for going from `torch` to `tensorflow`, do you think it makes sense to have these deps in the regular `requirements.txt` file because if someone does `pip install ivy`, we'd only install `numpy` and that's no good (it was good when it was to be used as a framework but is useless if we are marketing `ivy` as a transpiler with no support for transpiling to/fro `numpy` at launch). If you look at [transformers](https://github.com/huggingface/transformers/blob/3476c19e91a882fab8c76970a8798890a3f0f299/setup.py#L97) for example, they have `tensorflow`, `torch`, `jax` and `numpy` all listed down directly in their `setup.py` rather than have `optional` requirements files that the users themselves have to install regardless of whether [they're installing from source or from pip](https://ivy.dev/docs/overview/get_started.html#installing-using-pip)

hmahmood24:
Also would be good to pin the versions as part of this PR @Sam-Armstrong

Sam-Armstrong:
I'll make a new PR for pinning the frameworks, so we can check we agree on the pinned versions.

PR #28826, opened 2024-09-18
fix: pin framework versions in optional.txt

Sam-Armstrong:
I guess we should just pin the latest version like this, right @hmahmood24?


hmahmood24 (commented):
@Sam-Armstrong I think this implies that if any framework is installed to a less than the number specified then we won't override that installation. Is this intended?

Issue #28848, opened 2024-12-17
Add Pytorch Backend Support to Ivy Transpiler 


YushaArif99:
**Description**:  
The current implementation of `ivy.transpile` supports `"jax"`,`"numpy"` and `"tensorflow"` as valid `target` argument. This allows transpiling framework-specific code such as PyTorch functions or classes to target frameworks like TensorFlow, JAX, or NumPy. This task aims to extend the functionality by adding Pytorch as a valid `target`, enabling code to be transpiled to Pytorch.

For example, after completing this task, we should be able to transpile TensorFlow code using:  
```python
ivy.transpile(func, source="tensorflow", target="torch")
```  

### Goals:  
The main objective is to implement the final stage of the transpilation pipeline:  
1. **Lower Ivy IR representation to native Pytorch code.**  

Once this stage is complete, the rest of the pipeline can be reused to target other frameworks like JAX, PyTorch, or NumPy. The steps would look as follows:  
```text
source='tensorflow' → target='tensorflow_frontend'  
source='tensorflow_frontend' → target='ivy'  
source='ivy' → target='torch'  
```  

This mirrors the existing pipeline for eg: PyTorch -> TF:  
```text
source='torch' → target='torch_frontend'  
source='torch_frontend' → target='ivy'  
source='ivy' → target='tensorflow'  
```  

### Key Tasks:  
1. **Add Native Framework-Specific Implementations for Core Transformation Passes:**  
   - For example, implement the `ivy_to_torch_postprocessing_transformer.py` for doing postprocessing on the Ivy code when lowering it to Pytorch (eg: mapping ivy dtypes (eg: `ivy.float32`) to torch dtypes(`torch.float32`)
   - Use `ivy_to_jax_postprocessing_transformer.py` as a reference ([example here](https://github.com/ivy-llc/ivy/blob/open-source/ivy/transpiler/transformations/transformers/postprocessing_transformer/ivy_to_jax_postprocessing_transformer.py#L24))

2 **Define the Transformation Pipeline for Ivy IR to Torch**  
   - Create a new pipeline in `ivy_to_source_translator_config.py` to handle the stage `source='ivy', target='torch'` ([example here](https://github.com/ivy-llc/ivy/blob/open-source/ivy/transpiler/translations/configurations/ivy_to_source_translator_config.py#L92)).  

3. **Add Stateful Classes for `torch.nn` APIs:**  
[...]

PR #28860, opened 2025-03-20
Torch Backend in Transpiler

yugborana:
# PR Description

## Related Issue
#28848 

## Checklist

- [ ] Did you add a function?
- [ ] Did you add the tests?

### DIFF
--- a/ivy/functional/backends/torch/__init__.py
+++ b/ivy/functional/backends/torch/__init__.py
@@ -6,6 +6,21 @@ import torch as torch
 import ivy
 from ivy.func_wrapper import _dtype_from_version
 
+all_dtypes = (
+    ivy.int8,
+    ivy.int16,
+    ivy.int32,
+    ivy.int64,
+    ivy.uint8,
+    ivy.bfloat16,
+    ivy.float16,
+    ivy.float32,
+    ivy.float64,
+    ivy.complex64,
+    ivy.complex128,
+    ivy.bool,
+)
+
 backend_version = {"version": torch.__version__.split("+")[0]}
 
 # Registering ivy.Array as trackable submodule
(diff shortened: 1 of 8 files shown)


yugborana:
@Sam-Armstrong @YushaArif99 Good Afternoon. Would like to request for a review?

Sam-Armstrong (changes requested):
@yugborana Thanks for making the PR!

Looks like there's a few more things that will need changing before this works correctly. For example, you'll need to add `torch` as one of the `SUPPORTED_S2S_TARGETS` in ivy/transpiler/main_config.py.

I'd recommend using the following minimal example to check your solution is working correctly:
```
import kornia
import torch
import ivy

new_kornia = ivy.transpile(kornia, target="torch", reuse_existing=False)
rgb_image = torch.randn((1, 3, 224, 224))
gray_image = new_kornia.color.rgb_to_grayscale(rgb_image)
```

You can see this currently fails with the error `ivy.transpiler.exceptions.exceptions.InvalidTargetException: 'target' must be one of ['tensorflow', 'jax', 'numpy', 'ivy', 'torch_frontend'].` - but it looks like there are still other problems after fixing this.

yugborana:
@Sam-Armstrong i have changed the targets list, what changes must be done? New to this codebase so if you could guide a bit more..

yugborana:
@Sam-Armstrong these are the errors, could you please guide more, where should I do changes?

Sam-Armstrong:
@yugborana Can you push that change to the targets list? For this fix, it looks like you'll need to add another elif block in `ivy/transpiler/utils/source_utils.py::_maybe_create_stateful_layers_module` for torch where you set the `module_name` and `source` right?

yugborana:
@Sam-Armstrong I have pushed the changes. But now, I think we have to do something with torch stateful layers?

PR #28926, opened 2025-07-23
Fix Ivy Failing Test: jax - norms.batch_norm

Ajay6601:


Sam-Armstrong:
@Ajay6601 we can fix this more simply like this: https://github.com/ivy-llc/ivy/pull/28927

Issue #28936, opened 2026-02-24
resolve_neg


mustafaa7med:
Solving subtask resolve_neg related to #16

PR #28937, opened 2026-02-24
feat: Added torch resolve_neg frontend func

mustafaa7med:
Note: This issue is a sub-issue derived from #16 with an issue id of #28936

Added a resolve_neg function to the frontend and applied test cases.
this function was dependant on another function called negative() and has an alias of neg()


mustafaa7med:
@Sam-Armstrong I believe those failed tests are dependecies?

